PR #19, opened 2022-01-06
authenticate_request example should be async

billcrook:
Getting a mypy error otherwise:
`Return type "AuthenticationResult" of "authenticate_request" incompatible with return type "Coroutine[Any, Any, AuthenticationResult]" in supertype "AbstractAuthenticationMiddleware"
`


Goldziher:
Thanks

PR #28, opened 2022-01-14
Auto tablenames

Butch78:
Using as_declarative & declared_attr allows you to modify the Base class. You can then simply generate the table name Automatically using cls. I also added some tests :)


Goldziher (approved):
Cool, thanks.

PR #34, opened 2022-01-16
Add DTOFactory

Goldziher:
This PR adds a DTOFactory class.

This class can be used to create DTOs from pydantic models, dataclasses and any other class supported via plugins:

Given a supported model class - either pydantic, dataclass or a class supported via plugins, create a DTO pydantic model class.

An instance of the factory must first be created, passing any plugins to it. It can then be used to create a DTO by calling the instance like a function. Additionally, it can exclude (drop) attributes specifies in the 'exclude' list and remap field names and/or field types.

For example, given a pydantic model
```python
class MyClass(BaseModel):
    first: int
    second: int

MyClassDTO = DTOFactory()(MyClass, exclude=["first"], field_mapping={"second": ("third", float)})
```

`MyClassDTO` is now equal to this:

```python
class MyClassDTO(BaseModel):
    third: float
````

It can be used as a regular pydantic model:

```python            
@post(path="/my-path)
def create_obj(data: MyClassDTO) -> MyClass:
    ...
```
                
This will affect parsing, validation and how OpenAPI schema is generated exactly like when using a pydantic model.

Note: Although the value generated is a pydantic factory, because it is being generated programmaticaly, 
it's currently not possible to extend editor auto-complete for the DTO properties - it will be typed as a 
Pydantic BaseModel, but no attributes will be inferred in the editor.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass as vanilla_dataclass
 from enum import Enum
 from typing import Dict, List, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from pydantic.dataclasses import dataclass as pydantic_dataclass
 from pydantic_factories import ModelFactory
 
@@ -16,7 +16,7 @@ class Species(str, Enum):
 
 class Pet(BaseModel):
     name: str
-    species: Species
+    species: Species = Field(default=Species.MONKEY)
     age: float
 
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -468,7 +468,7 @@ email = ["email-validator (>=1.0.3)"]
 
 [[package]]
 name = "pydantic-factories"
-version = "1.2.0"
+version = "1.2.1"
 description = "Mock data generation for pydantic based models"
 category = "main"
 optional = false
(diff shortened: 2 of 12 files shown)


tests/__init__.py
```diff
@@ -16,7 +16,7 @@ class Species(str, Enum):
 
 class Pet(BaseModel):
     name: str
-    species: Species
+    species: Species = Field(default=Species.MONKEY)
```
yudjinn, line 19: do you not have to give it an ENUM type in the Field() def? unsure

yudjinn (approved):

tests/__init__.py
Goldziher: not at all, but i wanted to check that some lines got covered - and this was required.

sondrelg:
Could you elaborate on the naming, and the motivation? Is the idea that you might create a DTO from a DB-model or the like?

And are you sure about having this be a class? Writing a class with only an `__init__` and a `__call__` seems redundant. Why not make it a function? Syntactically, I'm not sure I prefer this
```python
DTOFactory()(MyClass, exclude=["first"], field_mapping={"second": ("third", float)})
```

to, e.g., this
```python
create_dto(MyClass, exclude=["first"], field_mapping={"second": ("third", float)})
```

PR #38, opened 2022-01-17
allow empty path for controller

### LINKED ISSUE #31: allow empty `""` path in Controller ?
https://github.com/starlite-api/starlite/blob/965d97620774991bd302a3a227bd3ebe6236ae96/starlite/controller.py#L27-L28

### Context
In some of our project (built with FastAPI) we use **Factory** classes that helps us creating bunch of endpoints for given user inputs and dependencies (https://developmentseed.org/titiler/advanced/tiler_factories/ https://developmentseed.org/titiler/api/titiler/core/factory/#tilerfactory). This is why I was really interested in the `Controller` class to replace our `Factories`. 

One problem I'm facing right now is that Controller `Path` has to be non null. 

```python
from starlite.controller import Controller


class Endpoints(Controller):
    """TileMatrixSets endpoints."""

    path = "/"

    @get(path="/yo")
    def hey(self, request: Request) -> str:
        return "yipi"

    @get(path="ye")
    def hey2(self, request: Request) -> str:
        return "what"


app = Starlite(route_handlers=[Endpoints])
```
schema

<details>

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Starlite API",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "/"
    }
  ],
  "paths": {
    "//yo": {
      "get": {
        "operationId": "hey",
        "responses": {
          "200": {
            "headers": {},
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        },
        "deprecated": false
      }
    },
    "//ye": {
      "get": {
        "operationId": "hey2",
        "responses": {
          "200": {
            "headers": {},
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        },
        "deprecated": false
      }
    }
  }
}
```

</details>

as we can see in the schema ☝️, both endpoints are `//` prefixed because Starlite is normalizing path (I think).  I can agree that I want to use the Controller class for something it wasn't designed for, but it could be nice if the `path` could be set to `""`.

### Potential fix

By allowing path to be all but not `None` it seem to fix my issue, but I'm not quite sure what it means for the rest of the application.

```python
class Endpoints(Controller):
    """TileMatrixSets endpoints."""

    path = ""

    def __init__(self, owner: "Router"):
        if not hasattr(self, "path") or self.path is None:
            raise ImproperlyConfiguredException("Controller subclasses must set a path attribute")

        for key in ["dependencies", "response_headers", "response_class", "guards"]:
            if not hasattr(self, key):
                setattr(self, key, None)

        self.path = self.path
        self.owner = owner
        for route_handler in self.get_route_handlers():
            route_handler.owner = self

    @get(path="/yo")
    def hey(self, request: Request) -> str:
        print(request)
        return "yipi"

    @get(path="/ye")
    def hey2(self, request: Request) -> str:
        print(request)
        return "what"
```

again, I will understand that I want to use the Controller class outside of its scope 🙏

vincentsarago:
closes #31


starlite/controller.py
```diff
@@ -24,14 +25,21 @@ class Controller:
...
             raise ImproperlyConfiguredException("Controller subclasses must set a path attribute")
 
         for key in ["dependencies", "response_headers", "response_class", "guards"]:
             if not hasattr(self, key):
                 setattr(self, key, None)
 
-        self.path = normalize_path(self.path)
+        if self.path == "":
+            warnings.warn(
+                "Empty controller's path. Routes will be added to the application without prefix.",
```
vincentsarago, line 37: feel free to update the wording here!

starlite/controller.py
```diff
@@ -24,14 +25,21 @@ class Controller:
...
+                UserWarning,
+            )
+        else:
+            self.path = normalize_path(self.path)
```
vincentsarago, line 41: If the path is empty we don't need to use `normalize_path`. (in `normalize_path` it will first add `/` and then remove `/`).

tests/test_controller.py
```diff
@@ -275,3 +275,29 @@ async def ws(self, socket: WebSocket) -> None:
...
+
+    class MyController(Controller):
+        path = test_path
+
+        @decorator(path="/")
+        def test_method(self) -> Person:
+            return person_instance
+
+    with pytest.warns(UserWarning), create_test_client(MyController) as client:
+        response = client.request(http_method, "/")
```
vincentsarago, line 301: Seems that it could led to another issue here. if `path=""` then we can't have empty route path.

cc @Goldziher, maybe we can add a second validation here https://github.com/starlite-api/starlite/blob/main/starlite/controller.py#L36-L37 in case `path == ""` and `route_handler.path == ""` 🤷

Goldziher:
Code looks good, but we also need some docs 😃

vincentsarago:
🙏 sure @Goldziher just wanted a first review :D

starlite/controller.py
ashwinvin: I think "controller" would be enough.

vincentsarago, line 37:
```suggestion
                "Empty controller path. Routes will be added to the application without prefix.",
```

starlite/controller.py
```diff
@@ -24,14 +25,21 @@ class Controller:
...
+        if not hasattr(self, "path") or self.path is None:
             raise ImproperlyConfiguredException("Controller subclasses must set a path attribute")
 
         for key in ["dependencies", "response_headers", "response_class", "guards"]:
             if not hasattr(self, key):
                 setattr(self, key, None)
 
-        self.path = normalize_path(self.path)
+        if self.path == "":
+            warnings.warn(
```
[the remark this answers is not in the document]
vincentsarago: Not sure, will remove!

tests/test_controller.py
vincentsarago: 🤦 well handler's path can't be empty https://github.com/starlite-api/starlite/blob/main/starlite/routing.py#L62-L63
so we are fine!

tests/test_controller.py
```diff
@@ -275,3 +275,29 @@ async def ws(self, socket: WebSocket) -> None:
...
+    [
+        (get, HttpMethod.GET, HTTP_200_OK),
+        (post, HttpMethod.POST, HTTP_201_CREATED),
+        (put, HttpMethod.PUT, HTTP_200_OK),
+        (patch, HttpMethod.PATCH, HTTP_200_OK),
+        (delete, HttpMethod.DELETE, HTTP_204_NO_CONTENT),
+    ],
+)
+def test_controller_empty_path(decorator, http_method, expected_status_code):
+    test_path = ""
```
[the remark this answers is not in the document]
vincentsarago:
```python
from starlite import create_test_client
from starlite.controller import Controller


    path = "/"

    @get(path="/something")
    def test_method(self) -> str:
        return "yo"

with create_test_client(MyController) as client:
    print(client.get("//something"))
>>> <Response [404]>

[r.path for r in Starlite(route_handlers=[MyController]).routes]

>>> ['//something',
 '/schema/redoc',
 '/schema/openapi.json',
 '/schema/openapi.yaml']
```

@Goldziher, this doesn't seem to work 🤷‍♂️

tests/test_controller.py
```diff
@@ -287,17 +317,18 @@ async def ws(self, socket: WebSocket) -> None:
...
             return person_instance
 
-    with pytest.warns(UserWarning), create_test_client(MyController) as client:
+    with create_test_client(MyController) as client:
         response = client.request(http_method, "/")
         assert response.status_code == expected_status_code
         assert response.json() == person_instance.dict()
+
+        response = client.request(http_method, "")
+        assert response.status_code == HTTP_404_NOT_FOUND
```
vincentsarago, line 334: this returns 200 for now, but I believe it should return 404 🤷

tests/test_controller.py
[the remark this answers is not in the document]
vincentsarago: TBH, I think we should just disable `path="/"` for controllers (I think other framework works the same)

PR #39, opened 2022-01-17
Add support for "before_request" and "after_request" hooks

Goldziher:
This PR adds ''before_request" and "after_request" hooks. 

These are directly borrowed from flask, with two differences:

1. Async support
2. You can define these functions on all layers - app, router,  controller, method/function, and the lowest level takes precedence.

### DIFF
--- a/starlite/controller.py
+++ b/starlite/controller.py
@@ -4,7 +4,12 @@ from typing_extensions import Type
 
 from starlite.exceptions import ImproperlyConfiguredException
 from starlite.response import Response
-from starlite.types import Guard, ResponseHeader
+from starlite.types import (
+    AFTER_REQUEST_HANDLER,
+    BEFORE_REQUEST_HANDLER,
+    Guard,
+    ResponseHeader,
+)
 from starlite.utils import normalize_path
 
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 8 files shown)


sondrelg:
Could consider adding something like [this](https://github.com/ariebovenberg/slotscheck) as a pre-commit hook to help ensure the `__slots__` definitions don't get messed up.

starlite/handlers.py
```diff
@@ -220,20 +256,58 @@ def resolve_response_headers(self) -> Dict[str, ResponseHeader]:
...
-                cur = cur.owner
             self.resolved_headers = headers
         return cast(Dict[str, ResponseHeader], self.resolved_headers)
 
+    def resolve_before_request(self) -> Optional[BEFORE_REQUEST_HANDLER]:
+        """
+        Resolves the before_handler handler by starting from the handler and moving up.
+
+        If a handler is found it is returned, otherwise None is set.
+        This mehtod is memoized so the computation occurs only once
```
yudjinn, line 271: small type, method

yudjinn (changes requested):

starlite/controller.py
```diff
@@ -14,20 +19,38 @@
 
 
 class Controller:
-    __slots__ = ("dependencies", "owner", "path", "response_headers", "response_class")
+    __slots__ = (
+        "dependencies",
+        "owner",
+        "path",
+        "response_headers",
+        "response_class",
```
ashwinvin, line 27:
```suggestion
        "response_class",
        "guards",
```

ashwinvin (changes requested):

PR #46, opened 2022-01-19
updated application lifecycle handling

Goldziher:
This PR updates how lifecycle is handled-

1. functions set to `on_startup` and `on_shutdown` can now receive the application `state` object.
2. the state object is now imported from `starlite.datastructures.State` instead of from Starlette, and it has a custom __copy__ method which ensures correct shallow copy
3. the state object can be injected to route handlers and dependencies by specifying the `state` kwarg.
4. documentation has also been updated.


ashwinvin (approved):

vincentsarago:
> functions set to on_startup and on_shutdown can now receive the application state object.

@Goldziher, To be honest I'm not quite sure this is the best way to go. Changing the `on_startup` / `on_shutdown` definition (with this PR, callable could receive the app.state object) seem kinda weird and will create a big difference between Starlite and starlette which I'm not sure it's something we really want. 

I was more thinking that maybe we could have done:

### Surface `Starlette.on_event`  (`Starlite.asgi_router.on_event`)

First and simple solution would be to surface `Starlette.on_event` method in `Starlite` Object to do something like:

```python
app = Starlite()

@app.on_event("startup")
async def startup_event():
    """Application startup"""
    await startup(app)


@app.on_event("shutdown")
async def shutdown_event():
    """Application shutdown."""
    await shutdown(app)
```

### Surface `Starlette.lifespan`

```python
async def lifespan(app: Starlite):
    assert isintance(Starlite, app) # Not sure where the magic happens here but I confirmed that app is the Starlite application here
    print("Hello there! Let's create a Pool, hope you took your swimsuit!")
    app.state.pool = await get_db_con()
    yield
    await app.state.pool.close()
    print("I'm done here. Pool closed.")

app = Starlite(lifespan=lifespan)
```

I've tested ☝️ locally and by forwarding `lifespan` to the starlette.Router here https://github.com/starlite-api/starlite/blob/main/starlite/app.py#L93-L98 It works fine. TBH I'm not sure how it works (I can't figure how passing lifespan to the `asgi_router` wi

Goldziher:
The kind of async lifespan handler you show above is deprecated in Starlette. They recommend using an async context manager for this purpose. 

This works because of how Starlette handles lifecycles. Look in the Starlette router lifecycle method to see this.


Can I ask, why is passing state to lifecycle functions weird? 

I don't have a problem with having a different logic from Starlette, so for me that is not a blocker.

vincentsarago:
> The kind of async lifespan handler you show above is deprecated in Starlette. They recommend using an async context manager for this purpose.

Ok 🤷, I've tried to use `asynccontextmanager` but I got a weird issue. 

> This works because of how Starlette handles lifecycles. Look in the Starlette router lifecycle method to see this.

👍 

> Can I ask, why is passing state to lifecycle functions weird?

It's more that we force the user to either provide `Callable[]` or `Callable[[Scope]]`. But it you are ok with that 👍 

> I don't have a problem with having a different logic from Starlette, so for me that is not a blocker.

👍

vincentsarago (approved):

PR #47, opened 2022-01-20
Handler multi path support

### LINKED ISSUE #44: [feature request] Multiple path per handler
I sometimes want to use optional `path` parameter. For this I need to be able to set more than one path per route: 

```python
@get(path="/image")
@get(path="/image.{format:str}")
def get_image(format: str = "jpeg"):
    ...
```
see: https://github.com/developmentseed/titiler/blob/master/src/titiler/core/titiler/core/factory.py#L436-L450

Goldziher:
This PR adds multi-paths support.

Closes #44


vincentsarago (approved):
Looks good to me!

starlite/handlers.py
```diff
@@ -274,10 +276,8 @@ def __init__(
             self.http_method = HttpMethod.from_str(http_method)  # type: ignore
         if status_code:
             self.status_code = status_code
-        elif isinstance(self.http_method, list):
-            raise ImproperlyConfiguredException(
-                "When defining multiple methods for a given path, a status_code is required"
-            )
+        if isinstance(self.http_method, list):
+            self.status_code = HTTP_200_OK
```
ashwinvin, line 280: Whats the reason to change this behaviour?

ashwinvin (approved):

starlite/handlers.py
Goldziher: I had doubts about the original impl. for a while. I should've separated this to its own PR though.

PR #51, opened 2022-01-21
SQL Alchemy Plugin, iteration 2

Goldziher:
- updated SQLAlchemy plugin- all relationships are now typed as `Optional`
- updated SQLAlchemy plugin- all back-references are now typed as `Any`

### DIFF
--- a/starlite/plugins/sql_alchemy.py
+++ b/starlite/plugins/sql_alchemy.py
@@ -2,7 +2,7 @@ from datetime import date, datetime, time, timedelta
 from decimal import Decimal
 from inspect import isclass
 from ipaddress import IPv4Network, IPv6Network
-from typing import Any, Callable, Dict, List, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 from uuid import UUID
 
 from pydantic import BaseModel, Json, conint, constr, create_model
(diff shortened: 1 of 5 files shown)


starlite/plugins/sql_alchemy.py
```diff
@@ -284,15 +284,19 @@ def to_pydantic_model_class(self, model_class: DeclarativeMeta, **kwargs: Any) -
...
                 for name, relationship_property in mapper.relationships.items():
-                    related_entity_class = relationship_property.mapper.class_
-                    related_model_name = related_entity_class.__qualname__
-                    if relationship_property.uselist:
-                        field_definitions[name] = (List[related_model_name], ...)  # type: ignore
+                    if not relationship_property.back_populates and not relationship_property.backref:
+                        related_entity_class = relationship_property.mapper.class_
+                        related_model_name = related_entity_class.__qualname__
+                        if relationship_property.uselist:
+                            field_definitions[name] = (Optional[List[related_model_name]], None)  # type: ignore
```
ashwinvin, line 291: Wouldn't this just simplify into `field_definitions[name] = Optional[List[related_model_name]]` ?

Goldziher: in this case though pydantic will fail if None value is given.

PR #58, opened 2022-01-24
Add templating

ashwinvin:
This pr aims to bring templating support for Starlite by leveraging already existing template engine. Currently `mako` and `jinja2` are supported. 

### Basic Example
```py 
from starlite import Starlite, TemplateConfig, Template, get
from starlite.template import JinjaTemplateEngine


@get(path="/")
def index() -> Template:
    return Template(name="index.html", context={"test": "test"})


app = Starlite(
    route_handlers=[heath_check], template_config=TemplateConfig(directory=["templates"], engine=JinjaTemplateEngine)
)
```
#### Stuff to do
- write docs
- ~~write tests~~
-  OpenAPI integration

### DIFF
--- a/docs/usage/5-responses.md
+++ b/docs/usage/5-responses.md
@@ -62,17 +62,17 @@ used:
 - GET, PATCH, PUT: 200 (Ok)
 
 !!! note
-    When using the `route` decorator with multiple http methods, the default status code is `200`.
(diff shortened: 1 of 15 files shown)


Goldziher:
Code looks pretty good.

Testing though is not sufficent :

1. 100% coverage
2. Test both templating engines
3. Test several scenarios

If you need inspiration, look at the django and Starlette tests.

PR #59, opened 2022-01-25
Remove second faulty "]" from README

lucasconnellm:
Sorry y'all, found a separate case from the one I found yesterday.. Wish I'd noticed on the first PR.


Goldziher:
Thanks 😃

PR #79, opened 2022-02-05
fix: proper Partial typing

to-ph:
Using `TypeVar("T", bound=Type[BaseModel])` as the type variable in `Partial` does not type-check. 

For example:

```python
class Person(BaseModel):
    name: str

PartialPerson = Partial[Person]  # does NOT type-check
PartialPerson = Partial[Type[Person]]  # type-checks
```

This PR changes the type variable used in `Partial` from `TypeVar("T", bound=Type[BaseModel])` to `TypeVar("T", bound=BaseModel)` and subsequent types in `Partial` as needed to allow `PartialPerson = Partial[Person]` to type-check 🤡 

For context, I enjoy using Pyright in its strictest configuration. If somehow this was ok by MyPy, then please ignore this PR.
BTW, `Partial` is brilliant! Typescript has a similar idea:

```typescript
type Partial<T> = {
    [P in keyof T]?: T[P];
};
```

It's unfortunate Python can't express it in this way, but we make do!

### DIFF
--- a/starlite/types.py
+++ b/starlite/types.py
@@ -87,16 +87,17 @@ class MiddlewareProtocol(Protocol):
         ...
 
 
-T = TypeVar("T", bound=Type[BaseModel])
+T = TypeVar("T", bound=BaseModel)
 
 
 class Partial(Generic[T]):
-    _models: Dict[T, Any] = {}
+    _models: Dict[Type[T], Any] = {}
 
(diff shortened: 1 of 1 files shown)


Goldziher:
Thanks 😃,

Yes I am a TypeScript developer as well. So thats the source of inspiration here.

Goldziher:
Released as v1.0.5

PR #80, opened 2022-02-06
implemented response caching

Goldziher:
This PR implements reponse caching:

1. add a `CacheBackendProtocol`, which specifies a simple cache backend protocol. This should allow users to use whatever implementation they prefer, including stuff like Redis and Memcached.
2. add a `CacheConfig`, which sets the cache backend for the app and the default expiration.
3. update handlers to have the cache kwarg - this value can be either a boolean (default False), or an integer (expiration in seconds).
4. update the HTTPRoute.handle method to handle the caching of responses. The Response instance is "pickled" - which is somewhat security sensitive but I think its OK in this context - the only vector for code injection is if an attacker controls the cache backend already, which basically means he or she has access to the environment.
5. the caching keys are derived by default from the path + query params. The user can set a function to create the cache key both in the global config as well as per handler. 

P.S. This PR also includes splitting the Router from the Routes - i.e. move Router out of routes.py to router.py. In hindsight it pollutes the DIFF and makes review harder, so apologies. 

TODOs:

1. update CHANGELOG
2. update documentation
3. optional - more tests against other backends. This needs some consideration because testing against redis requires some work.


starlite/routes.py
```diff
@@ -0,0 +1,309 @@
...
+            raise MethodNotAllowedException()
+        request: Request[Any, Any] = Request(scope=scope, receive=receive, send=send)
+        route_handler, parameter_model = self.route_handler_map[request.method]
+        if route_handler.resolve_guards():
+            await route_handler.authorize_connection(connection=request)
+
+        caching_enabled = request.method == "GET" and route_handler.cache
+        response: Optional[Union[Response, StarletteResponse]] = None
+        if caching_enabled:
+            response = self.get_cached_response(request=request)
```
ashwinvin, line 134: We should also consider supporting async cache backends.

Goldziher: done

starlite/__init__.py
```diff
@@ -45,7 +45,8 @@
 from .plugins import PluginProtocol
 from .provide import Provide
 from .response import Response
-from .routing import BaseRoute, HTTPRoute, Router, WebSocketRoute
+from .router import Router
+from .routes import BaseRoute, HTTPRoute, WebSocketRoute
```
ashwinvin, line 49: Missing `CacheConfig`

ashwinvin:
I think adding a config to cache content for each IP address would be great too.

Goldziher:
> I think adding a config to cache content for each IP address would be great too.

I don't follow, can you explain?

starlite/__init__.py
Goldziher: fixed

PR #84, opened 2022-02-09
Updated logging config to be non-blocking by adding queue handler.

### LINKED ISSUE #77: [enhancement] logging setup to be non-blocking
The default logging config, which is just a simple abstraction on top of DictConfig, creates a standard python logger. The issue with this setup is that python logging is sync and blocking. This means that the logging config needs to be updated to setup a QueueHandler. 

Here are some resources regarding this:

1. https://rob-blackbourn.medium.com/how-to-use-python-logging-queuehandler-with-dictconfig-1e8b1284e27a
2. https://stackoverflow.com/questions/25585518/python-logging-logutils-with-queuehandler-and-queuelistener
3. https://betterprogramming.pub/power-up-your-python-logging-6dd3ed38f322

madlad33:
Fixes #77

### DIFF
--- a/starlite/__init__.py
+++ b/starlite/__init__.py
@@ -44,6 +44,7 @@ from .openapi.controller import OpenAPIController
 from .params import Body, Parameter
 from .plugins import PluginProtocol
 from .provide import Provide
+from .queue_listener_handler import QueueListenerHandler
 from .response import Response
 from .routing import BaseRoute, HTTPRoute, Router, WebSocketRoute
 from .testing import TestClient, create_test_client, create_test_request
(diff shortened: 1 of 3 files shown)


Goldziher:
Ok, I went through the code.

All in all its quite good 😃. Left some comments.

As for tests -

1. You can unit test the queue handler itself
2. You can test logging using pytest.  Purest has a fixture for capturing logging output. Basically you will need to capture this and test.

I will have time on the weekend to help you with this, if you need help still. I am also pinging here @ashwinvin , who might be able to give a hand. 

Thanks for contributing!

PR #91, opened 2022-02-19
fix: 🐛 fixing request json method for empty payload

t1waz:
fixing request json method for empty payload


t1waz:
hope You guys will find it helpfull, cheers

Issue #92, opened 2022-02-19
Version >0.6.0 breaks query parameters


i404788:
I haven't done extensive testing yet, but it seems like version >0.6.0 breaks query parameters as described in the documentation (https://starlite-api.github.io/starlite/usage/3-parameters/#query-parameters). ~~It gives 404 without every executing the function.~~

```
class ApiV1Controller(Controller):
    path = "/api/v1"

    @get(path="/product/{isin:str}/rawdata")
    async def product_rawdata(self, isin: str, dataType:  Literal['profile', 'etc']) -> Any:
        raise KeyError("Test")

with create_test_client(ApiV1Controller) as client:
        print(client.get("/api/v1/product/test/rawdata").text)
        print(client.get("/api/v1/product/test/rawdata?dataType=profile").text)
```
Output (1.2.0):
```
{"detail":"Missing required parameter dataType for url http://testserver/api/v1/product/test/rawdata","extra":null}
{"detail":"Not Found","extra":null}
```

Output (0.6.0):
```
{"detail":"Validation failed for GET http://testserver/api/v1/product/test/rawdata:\n\ndataType\n  none is not an allowed value (type=type_error.none.not_allowed)","extra":null}
Traceback (most recent call last):
...
    raise KeyError("test")
KeyError: 'test'
```

UPDATE: seems like it does run the function, but fails with 404 if there is a KeyError; I've updated the example. This might be intended behavior but it's very confusing since it also doesn't log anything on the server-side.

Goldziher:
Hi, it should return a status 400 not a status 404. Can you add a test marked by @pytest.mark.xfail?

Goldziher:
thank you @i404788 . V1.2.3 has been released with a fix. Please see if it resolves your issue and if so close this issue.

PR #93, opened 2022-02-20
Add test for issue #92

i404788:


Goldziher:
Well, this doesn't actually test the issue - if a handler raises a keyerror like this it will be a status 500 - internal server error. The issue that should be tested is declaring a query parameter in the arguments, and then not passing it. If I understand you correctly this should return a 404, which is wrong.

i404788:
In my testing a KeyError gives a 404 even though it should be 500 so I used that is the final reduction. I'm not sure if this depends on the query parameter but that's how I found it; I can do further testing.

i404788:
Without a query parameter the issue still occurs, so it's independent of parameters:
```
class ApiV1Controller(Controller):
    path = "/api/v1"

    @get(path="/product/{isin:str}/rawdata")
    async def product_rawdata(self, isin: str) -> Any:
        raise KeyError("Test")

with create_test_client(ApiV1Controller) as client:
        resp = client.get("/api/v1/product/test/rawdata")
        print(resp.text, resp.status_code)
 ```
 Ouput (1.2.0):
 ```
 {"detail":"Not Found","extra":null} 404
 ```

PR #94, opened 2022-02-20
Feat/optional extras sqlalchemy

tclasen:


Goldziher:
Ok, this is stale I think, no?

PR #95, opened 2022-02-21
Ts logging non blocking

### LINKED ISSUE #77: [enhancement] logging setup to be non-blocking

madlad33:
Fixes #77 
- Added tests for logging and the resolve handler.

### DIFF
--- a/starlite/__init__.py
+++ b/starlite/__init__.py
@@ -44,6 +44,7 @@ from .openapi.controller import OpenAPIController
 from .params import Body, Parameter
 from .plugins import PluginProtocol
 from .provide import Provide
+from .queue_listener_handler import QueueListenerHandler
 from .response import Response
 from .routing import BaseRoute, HTTPRoute, Router, WebSocketRoute
 from .testing import TestClient, create_test_client, create_test_request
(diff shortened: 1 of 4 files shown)


Goldziher:
Very nice!

Goldziher:
There are validation issues. Namely missing types etc. Please install `pre-commit` on your machine, and then do:

1. pre-commit install
2. pre-commit run --all-files

Then fix all issues..

madlad33:
> There are validation issues. Namely missing types etc. Please install `pre-commit` on your machine, and then do:
> 
> 1. pre-commit install
> 2. pre-commit run --all-files
> 
> Then fix all issues..

I ran the pre-commit locally but I don't why it didn't detect those errors.
Thanks for merging.

PR #105, opened 2022-04-21
changed middleware call order

slavugan:
Currently middlewares provided by user are called in reverse order, my changes make them to be called in order provided by user.
Also changed addition of `TrustedHostMiddleware` and `CORSMiddleware` so that they called before custom middlewares.

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -257,7 +257,7 @@ class Starlite(Router):
         Builds the middleware stack by passing middlewares in a specific order
         """
         current_app: ASGIApp = self.asgi_router
-        # last added middleware will be on top of stack hense it will be called first
+        # last added middleware will be on the top of stack and it will therefore be called first.
         # we therefore need to reverse the middlewares to keep the call order according to
         # middlewares list provided by a user
         for middleware in reversed(user_middleware):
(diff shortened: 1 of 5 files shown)


Goldziher (commented):
Nice 😃. Please fix tests.

starlite/app.py
```diff
@@ -257,15 +257,18 @@ def build_middleware_stack(
         Builds the middleware stack by passing middlewares in a specific order
         """
         current_app: ASGIApp = self.asgi_router
-        if allowed_hosts:
-            current_app = TrustedHostMiddleware(app=current_app, allowed_hosts=allowed_hosts)
-        if cors_config:
-            current_app = CORSMiddleware(app=current_app, **cors_config.dict())
-        for middleware in user_middleware:
+        # last added middleware will be on top of stack hense it will be called first
```
Goldziher, line 260:
```suggestion
        # last added middleware will be on the top of stack and it will therefore be called first.
```

starlite/app.py
```diff
@@ -257,15 +257,18 @@ def build_middleware_stack(
...
+        # so need to reverse middlewares to keep call order according to
```
Goldziher, line 261:
```suggestion
        # we therefore need to reverse the middlewares to keep the call order according to
```

starlite/app.py
```diff
@@ -257,15 +257,18 @@ def build_middleware_stack(
...
+        # middlewares list provided by a user
```
Goldziher, line 262:
```suggestion
        # the middlewares' list provided by the user
```

PR #114, opened 2022-05-24
DTOFactory.__call__(): `...` equivalent to `PydanticUndefined` in checking for default

peterschutt:
- fields in `DTOFactory.__call__()` either come from a pydantic model, a dataclass or a plugin
- where fields come from a plugin, they most likely come from a model built with `create_model()`, and no-default defined with `...`

This patch uses `...` as well as `PydanticUndefined` to check whether a field has a default value.


peterschutt:
For #114

Goldziher:
Nice

PR #116, opened 2022-05-24
Pass `FieldInfo` straight through to `create_model()` if no need to change

peterschutt:
Where a field is neither excluded, nor mapped the least surprising thing would be that the field stays the same.

This PR puts the key, value pair `field_name` and `(field_type, model_field.field_info)` into the `field_definitions` dict which is later unpacked into `Pydantic.create_model()`.

### DIFF
--- a/starlite/dto.py
+++ b/starlite/dto.py
@@ -137,12 +137,16 @@ class DTOFactory:
                         field_name, field_type = mapping
                     else:
                         field_name = mapping
-                if model_field.field_info.default not in (Undefined, None, ...):
-                    field_definitions[field_name] = (field_type, model_field.default)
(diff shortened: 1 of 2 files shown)


peterschutt:
For #116

Goldziher:
Please rebase

Goldziher:
Ok, so apparently both this PR and main now fail on lower versions of python. Can you check what's what?

Goldziher:
ok, i fixed development - i basically skipped the test in lower python versions because the issue is dataclass behaviour in py 3.7/8, which is not worthwhile to worry about IMO. So pls simply rebase and lets get this merged and released.

PR #124, opened 2022-05-31
Take resolved dependencies into account when generating OpenAPI schema

timwedde:
Instead of simply ignoring dependencies that are mentioned in the method definition, this change resolves them and adds the dependency's parameters to the schema. IMHO this behavior makes more sense, as any time a dependency is active, its own path/header/query parameters are taken into account when invoking the endpoint, so they should also show up in the OpenAPI schema.

Since this is my first PR for this project i'm not a 100% on all the things I need to do to upstream something (nor am I that familiar with the codebase), so this is a tentative start to test the waters and to gauge whether this even makes any sense.

### DIFF
--- a/starlite/openapi/parameters.py
+++ b/starlite/openapi/parameters.py
@@ -1,6 +1,7 @@
-from typing import Any, Dict, List
+from typing import Any, Dict, List, cast
 
 from openapi_schema_pydantic import Parameter, Schema
+from pydantic import BaseModel
 from pydantic.fields import ModelField
 
 from starlite.constants import RESERVED_KWARGS
@@ -26,11 +27,13 @@ def create_parameters(
     """
     path_parameter_names = [path_param["name"] for path_param in path_parameters]
     parameters: List[Parameter] = []
-    ignored_fields = [
-        *RESERVED_KWARGS,
-        *list(route_handler.resolve_dependencies().keys()),
-    ]
+    ignored_fields = [*RESERVED_KWARGS]
+    dependencies = route_handler.resolve_dependencies()
     for f_name, field in handler_fields.items():
+        if f_name in dependencies:
+            dependency_fields = cast(BaseModel, dependencies[f_name].signature_model).__fields__
+            parameters.extend(create_parameters(route_handler, dependency_fields, path_parameters, generate_examples))
+            continue
         if f_name not in ignored_fields:
             schema = None
             param_in = "query"
--- a/starlite/openapi/parameters.py
+++ b/starlite/openapi/parameters.py
@@ -27,14 +27,13 @@ def create_parameters(
     """
     path_parameter_names = [path_param["name"] for path_param in path_parameters]
     parameters: List[Parameter] = []
-    ignored_fields = [*RESERVED_KWARGS]
     dependencies = route_handler.resolve_dependencies()
     for f_name, field in handler_fields.items():
         if f_name in dependencies:
             dependency_fields = cast(BaseModel, dependencies[f_name].signature_model).__fields__
             parameters.extend(create_parameters(route_handler, dependency_fields, path_parameters, generate_examples))
             continue
-        if f_name not in ignored_fields:
+        if f_name not in RESERVED_KWARGS:
             schema = None
             param_in = "query"
             required = field.required
--- a/tests/openapi/test_schema.py
+++ b/tests/openapi/test_schema.py
@@ -1,7 +1,19 @@
+from typing import cast
+
 from openapi_schema_pydantic import Example
 from openapi_schema_pydantic.v3.v3_1_0.schema import Schema
 from pydantic.fields import FieldInfo
 
+from starlite import (
+    Controller,
+    MediaType,
+    Parameter,
+    Provide,
+    Starlite,
+    create_test_client,
+    get,
+)
+from starlite.app import DEFAULT_OPENAPI_CONFIG
 from starlite.openapi.constants import (
     EXTRA_TO_OPENAPI_PROPERTY_MAP,
     PYDANTIC_TO_OPENAPI_PROPERTY_MAP,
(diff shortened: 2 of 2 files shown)


Goldziher:
I'm curious to see the result

timwedde:
Alright, I dug around the code a little more and realized that my initial implementation did not consider transitive dependencies properly (and was also a little messy). This new version has a *much* smaller diff and recursively resolves dependencies so that all transitive ones also show up with the proper parameters.

This can take a rather unsuspecting doc entry like this:
before

to its real nature, which is this:
after

The example is of course hyperbolic to show the effect that omitting these dependency parameters has, but I guess the point comes across.

The example app used to produce this result is basically an extended version of the sample from the documentation about dependencies, I merely added a couple transitive ones and a few optional values to see the effect more clearly.
```py
from starlite import Controller, Router, Starlite, Provide, Parameter, get


def app_dep(app_dep_int: int) -> str:
    return "app_dep"


def app_dep_trans2(
    router_trans2_dep_int: int,
    router_trans2_dep_str_opt: str = Parameter(header="router_trans2_dep_str_opt", required=False),
) -> str:
    return "app_dep_trans2"


def app_dep_trans1(
    router_trans1_dep_int: int, app_dependency_transitive_2: str, router_trans1_dep_str_opt: str = None
) -> str:
    return "app_dep_trans1"


def router_dep(router_dep_int: int, app_dependency_transitive_1: str) -> str:
    return "router_dep"


def controller_dep(controller_dep_int: int) -> str:
    return "controller_dep"


def handler_dep(handler_dep_int: int) -> str:
    return "handler_dep"


    path = "/controller"
    dependencies = {"controller_dependency": Provide(controller_dep)}

    @get(path="/handler", dependencies={"local_dependency": Provide(handler_dep)})
    def example_handler(
        self,
        app_dependency: str,
        router_dependency: str,
        controller_dependency: str,
        local_dependency: str,
        local_int: int,
    ) -> dict[str, str]:
        return {
            "app_dependency": app_dependency,
            "router_dependency": router_dependency,
            "controller_dependency": controller_dependency,
            "local_dependency": local_dependency,
        }


my_router = Router(
    path="/router",
    dependencies={"router_dependency": Provide(router_dep)},
    route_handlers=[MyController],
)

app = Starlite(
    route_handlers=[
        my_router,
    ],
    dependencies={
        "app_dependency": Provide(app_dep),
        "app_dependency_transitive_1": Provide(app_dep_trans1),
        "app_dependency_transitive_2": Provide(app_dep_trans2),
    },
)
```

Do you think that this is a good change to introduce to Starlite? Imho this would make the schema generation more correct (the parameters are in fact required anyway), but of course I don't have full visibility on the project, so there may be some reasons to block this. Let me know your thoughts.

Goldziher:
@timwedde, so IMO dependency parameters should not appear in the docs because they are internal to the API. 

URL, query, header and cookie parameters, as well as a request body, are external inputs. 

That's the reason they were ignored to begin with.

timwedde:
I'm wondering if they really *are* internal to the API. After all they do manifest in the external invocation of the handler, right?
If I define a dependency which mandates a path parameter, this parameter must then be given when invoking the handler at that path.
FastAPI [also handles it this way](https://fastapi.tiangolo.com/tutorial/dependencies/#integrated-with-openapi_1), as a point of reference.

Imho FastAPI's behavior makes a bit more sense here than Starlite's, since the path parameter from the dependency is required one way or the other, and I struggle to see a use case where hiding parameters for a route by default is preferable to showing them, especially when they're mandatory.

In what way would a mandatory parameter in a dependency be internal or rather: What exactly does "internal" mean in this context when it also affects the external API interface?

starlite/openapi/parameters.py
```diff
@@ -26,11 +27,13 @@ def create_parameters(
     """
     path_parameter_names = [path_param["name"] for path_param in path_parameters]
     parameters: List[Parameter] = []
-    ignored_fields = [
-        *RESERVED_KWARGS,
-        *list(route_handler.resolve_dependencies().keys()),
-    ]
+    ignored_fields = [*RESERVED_KWARGS]
```
Goldziher, line 30: no need to unpack this, or to aliad it as `ignored_fields` if its only this iterable.

timwedde: Ah, good catch, will make a small change to improve this.

timwedde:
Fully agreed on all counts, making truly internal parameters available in the schema was indeed never my intention.
When it comes to your example: This should actually already be handled implicitly with this PR.
In fact, this example models what I believe you wanted to convey in your last comment and actually already behaves properly when running on this PR:
```py
from starlite import Controller, Router, Starlite, Provide, get


def get_user(user_id: int) -> str:
    return f"User {user_id}"


def authenticate_user(user: str) -> str:
    return f"{user} - Authenticated"


    path = "/controller"

    @get(
        path="/handler",
        dependencies={"auth_user": Provide(authenticate_user)},
    )
    def example_handler(
        self,
        auth_user: str,
    ) -> dict[str, str]:
        return {
            "auth_user": auth_user,
        }


my_router = Router(path="/router", route_handlers=[MyController])

app = Starlite(
    route_handlers=[my_router],
    dependencies={
        "user": Provide(get_user),
    },
)
```

This is because my implementation recursively resolves parameters. Due to this, it recurses into other dependencies and as such, the actual dependency itself is never written into the schema. The only things that will get written are method parameters that are:
- not mentioned in `RESERVED_KWARGS`
- not a dependency (as figured out via `route_handler.resolve_dependencies()`)

This effectively achieves the exact effect we're after, where only method parameters that actually translate into OpenAPI parameters are written into the schema. The only time where this can fail is when something declares the same parameter twice, which I believe is already checked for by Starlite and should be prevented anyway.

The three points you listed also make sense to me. Towards the first one I would add that, if the schema is indeed supposed to be able to be used for client code generation, then including path/query/header etc parameters from dependencies is actually a mandatory feature, as otherwise the generated code can not be complete. This is in fact a similar issue to what I'm running into currently and which is what prompted me to create this PR in the first place, heh.

If this implementation makes sense to you (including the part where it specifically recurses into transitive dependencies instead of listing them directly in the schema, which would be incorrect), I can write some tests to validate this programmatically.

peterschutt:
I modified the example app to highlight the issue as it's a nicer pattern than I had for reusing query parameter configs. Using `1.3.7` docs display like this:

image

And from @timwedde's branch:

image

Nice PR. LMK If there's anything I can do to help you get this across the line @timwedde.

timwedde:
A test for this functionality has been added. I wasn't a 100% sure where I should put this, so I figured the test/openapi/schema file was closest to the concerns this PR touches. Let me know if this should go somewhere else or be pulled into a new test file.

This test creates an example app with multiple dependencies that each introduce various parameters in the header, query or path and then tests for their availability in the generated schema. It will ensure that all the parameters that are active for a given handler (including dependencies) are given in the schema as well. Disabling my PR makes it fail, as it should.

If this should be expanded some more, please let me know. I'm not yet fully familiar with how testing is done in this project so I figured adding a basic test would be a good starting point.

PR #125, opened 2022-06-02
Fix `Router.tags` ignored.

peterschutt:
PR #89 added the `tags` slot to `Router` in ef02e8e015b1f42e3390073f0785305a4a1c5838, but `openapi.utils.extract_tags_from_route_handler()` only checks the handler and it's immediate owner (from 7e5610825377c8299845d4429bf669a9ca667d78).

This implementation returns the union of the route handler's tags, and any ancestor tags.

I also made a few changes to the docs:
- `docs/usage/12-openapi.md` - `tags` documented as accepting `Tag` objects but is typed as `Optional[List[str]]`.
- `docs/usage/1-routers-and-controllers` - add `tags` to controller class variables.
- `docs/usage/1-routers-and-controllers` - add `tags` to router kwargs.

### DIFF
--- a/docs/usage/1-routers-and-controllers.md
+++ b/docs/usage/1-routers-and-controllers.md
@@ -61,6 +61,7 @@ Aside from the `path` class variable, you can also set the following optional cl
   will be used.
 - `after_request`: a sync or async function to execute before the `Response` is returned. This function receives the
   `Respose` object and it must return a `Response` object.
+- `tags`: a list of `str`, which correlate to the [tag specification](https://spec.openapis.org/oas/latest.html#tag-object).
 
 ## Routers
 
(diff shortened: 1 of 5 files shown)


peterschutt:
I added a test module for openapi tag collection but I was getting a stack of errors to do with the `disallow_untyped_decorators` flag and the `pytest.fixture` decorator so I disabled it for that module only in `pytest.ini`.

I see that `pytest.fixture` isn't used anywhere else in the test suite. Is this by design? I can refactor so that I don't use `pytest.fixture` at all, or can just do `type:ignore[misc]` on each line, or just leave it as is:)

Goldziher:
> I added a test module for openapi tag collection but I was getting a stack of errors to do with the `disallow_untyped_decorators` flag and the `pytest.fixture` decorator so I disabled it for that module only in `pytest.ini`.

Both are fine for me - as is easier for you.

peterschutt:
OK, I'll just leave as is for now. No idea about the CI sorry.

Goldziher:
try rebasing

PR #136, opened 2022-06-11
Improper config when user-defined default type provided as parameter

peterschutt:
Pydantic's `field.SHAPE_GENERIC` doesn't exist in `PYDANTIC_FIELD_SHAPE_MAP`.


peterschutt:
I have a service object that looks a bit like:

```python
class Service(Generic[T_model, T_repository, T_schema]):

    def __init__(
        self,
        *,
        id_filter: CollectionFilter[UUID] | None = None,
    ) -> None:
        ...
```
The service object is provided as a dependency at the router level, and injected into each handler function.

The `id_filter` dependency is only provided for collection GET routes, so when it isn't provided the framework attempts to render it as a parameter which is when we get the issue:

```
  File "/home/peter/.cache/pypoetry/virtualenvs/provider-integrations-sr_TuWLX-py3.10/lib/python3.10/site-packages/starlite/openapi/schema.py", line 212, in create_schema
    openapi_type = PYDANTIC_FIELD_SHAPE_MAP[field.shape]
KeyError: 10
```

This exact case will be fixed by the `Dependency` function once it is available, as I'll declare it `id_filter: CollectionFilter[UUID] | None = Dependency(None)` and so it will be automatically excluded from parameter documentation.

However, we should probably eagerly check for the case where a user provides a parameter to a handler function, that is a user-defined generic, and the dependency is not provided to the handler, and is not marked specifically as a dependency with the `Dependency` function. We can then raise an exception for improper configuration and give a better error message, as the error message above is quite cryptic.

Goldziher (approved):

Goldziher (approved):

Issue #128, opened 2022-06-04
Mark that a dependency should never be a query parameter.


peterschutt:
This is a side-effect of the change in `1.3.9` and the way that I've structured the consumption of dependencies in the example app.

The `__init__()` method of the repository looks like this:

```python
    def __init__(
        self,
        limit_offset: LimitOffset | None = None,
        updated_filter: BeforeAfter | None = None,
    ) -> None:
        ...
        if limit_offset is not None:
            self.apply_limit_offset_pagination(limit_offset)
        if updated_filter is not None:
            self.filter_on_datetime_field(updated_filter)
```

If the repository is included as a dependency and if the the `limit_offset` and `updated_filter` dependencies are available those dependencies get consumed by the repository, and their dependencies end up documented as query params. However, for routes where the `limit_offset` and `updated_filter` dependencies aren't available to the route, they themselves are shown as query params:

![image](https://user-images.githubusercontent.com/20659309/171970784-ecf53ca7-4905-4ef1-9668-20d785a80645.png)

The reason I use this pattern is that without it, every list route, e.g., `/users` or `/users/{user_id}/items` has to do something like this:

```python
    @get(
        dependencies={
            "limit_offset": Provide(limit_offset_pagination),
            "updated_filter": Provide(filter_for_updated),
        }
    )
    async def get(
        self,
        repository: UserRepository,
        is_active: bool = Parameter(query="is-active", default=True),
        limit_offset: LimitOffset,
        updated_filter: BeforeAfter,
    ) -> list[UserModel]:
        repository.apply_limit_offset_pagination(limit_offset)
        repository.filter_on_datetime_field(updated_filter)
        return await repository.get_many(is_active=is_active)
```

which reduces down to this if the repository can be left to optionally consume the dependency if it is available:

```python
    @get(
        dependencies={
            "limit_offset": Provide(limit_offset_pagination),
            "updated_filter": Provide(filter_for_updated),
        }
    )
    async def get(
        self,
        repository: UserRepository,
        is_active: bool = Parameter(query="is-active", default=True),
    ) -> list[UserModel]:
        return await repository.get_many(is_active=is_active)
```

peterschutt:
What about a user-land configuration that is similar to the `RESERVED_KWARGS` constant, and if a field name appears in there, it is ignored for doc-generation purposes?

```python
def create_parameters(...) -> List[Parameter]:
    ...
    for f_name, field in handler_fields.items():
        ...
[...]

PR #137, opened 2022-06-11
Initial implementation.

peterschutt:
- defines `Dependency` in `params.py`
- `Dependency` inserts `is_dependency=True` into `field.extra`
- excludes explicit dependencies from parameter documentation.


peterschutt:
Closes #128 
Related #136

peterschutt:
- [x] Implement `Dependency` function - consider the params that we expose from `Field`.
- [ ] Implement skip declared dependencies for parameter documentation if the dependency not provided.
- [ ] Build logic for checking that all declared dependencies are either provided or have a default value.
- [ ] Docs for `Dependency`, include example.

Test cases:
- [ ] In/out tests for `Dependency` function that tests the stuff we rely on in `extra` is there.
- [ ] declared optional type and undefined default, e.g., `dep: int | None = Dependency()`
- [ ] non-optional type and defined default, e.g., `dep: int = Dependency(default=30)`
- [ ] error case, e.g., `dep: int = Dependency()` and `dep` not provided
- [ ] test that parameters for dependencies reach the docs if the dependency is provided.
- [ ] test that the default values reach the dependency if not provided.
- [ ] some tests of validation `Field` parameters.
- [ ] test that any validation failures of dependency fields are 500s not 400s - if a dependency fails validation it isn't the client's fault.

Goldziher:
Please check sonar - also coverage is below 100%.

peterschutt:
Yeh this isn't ready yet. 

I pushed up the impl so that I could open a PR to move my checklist from the issue into here.

Working on the tests now.

Goldziher:
> Yeh this isn't ready yet. 
> 
> I pushed up the impl so that I could open a PR to move my checklist from the issue into here.
> 
> Working on the tests now.

Sorry. Will hold back next time. To avoid confusion, best to use a draft PR

peterschutt:
It is draft ;) although I marked it draft a few seconds after I created it. So Sonar should run on drafts?

Goldziher:
> It is draft ;) although I marked it draft a few seconds after I created it. So Sonar should not run on drafts?

It does. Anyhow, sorry

peterschutt:
No problems at all.

Does it make too much noise for you if I work directly in the repo? I can be doing a lot of this in my fork if that's better?

Goldziher:
Not at all. 

I get a lot of notifications usually - work stuff.

peterschutt:
I've cooled on allowing validation arguments to `Dependency` for now. 

For one thing it isn't clear what error should be returned to the client. Probably should be a 500 error by default if an explicit dependency fails validation, however that might need to be configurable, maybe `Dependency(lt=13, invalid_status=400)`.

It would also be possible to have both a client parameter validation failure and a dependency validation failure on the same signature model for a given request. In that case we'd probably have to return the higher status code of 400 or the status for the dependency validation failure.

By excluding the validation kwargs from `Dependency` altogether for now it means we can add them in later on when someone requests them with a concrete use case and not break anything. It would be much harder to take them away or even worse to have to maintain support with edge cases everywhere forever.

There is also the issue of identifying that the field that raised the `ValidationError` is a `Dependency` field in `SignatureModel.parse_values_from_connection_kwargs()`. I'm sure it is possible, but case where the dependency is buried a few levels below the surface of the dependency tree of the handler might be be tricky.

peterschutt:
@Goldziher I still have to do the docs, but code is ready for review.

Goldziher:
> @Goldziher I still have to do the docs, but code is ready for review.

Ok, will review shortly. Please look in sonar and increase coverage to 100%.

starlite/app.py
```diff
@@ -241,11 +241,17 @@ def create_handler_signature_model(self, route_handler: BaseRouteHandler) -> Non
         """
         if not route_handler.signature_model:
             route_handler.signature_model = model_function_signature(
-                fn=cast(AnyCallable, route_handler.fn), plugins=self.plugins
+                fn=cast(AnyCallable, route_handler.fn),
+                plugins=self.plugins,
+                provided_dependencies=(route_handler.dependencies or {}).keys(),
```
Goldziher, line 246: this value can be placed in a variable and then used in both calls.

starlite/signature.py
```diff
@@ -58,7 +66,54 @@ def parse_values_from_connection_kwargs(
             ) from e
 
 
-def model_function_signature(fn: AnyCallable, plugins: List[PluginProtocol]) -> Type[SignatureModel]:
+def detect_optional_union(annotation: Any) -> bool:
```
Goldziher, line 69: it might be better to move these two helper functions into a utils module.

starlite/signature.py
```diff
@@ -58,7 +66,54 @@ def parse_values_from_connection_kwargs(
...
+    >>> get_origin(Union[int, None]) is Union
+    True
+    >>> get_args(Optional[int])
+    (<class 'int'>, <class 'NoneType'>)
+    >>> get_args(int | None)
+    (<class 'int'>, <class 'NoneType'>)
+    >>> get_args(Union[int, None])
+    (<class 'int'>, <class 'NoneType'>)
+    """
+    return get_origin(annotation) in UNION_TYPES and type(None) in get_args(annotation)
```
Goldziher, line 87: clever.

Goldziher (approved):
Looks good 🚀

peterschutt:
> Please look in sonar and increase coverage to 100%.

The one line that has no coverage in sonar is the else block for `if sys.version_info >= (3, 10):`. Sonar must be running on 3.10. Is it coverage.py under the hood (I.e, `# pragma: no cover` to exclude the line from coverage)?

Goldziher:
Yup

peterschutt:
I ran into a chicken and egg scenario where I need the name of all provided dependencies to check for the error condition prior to the signature model being constructed (because I need them to check the signature model themselves). That prevents me from using `route_handler.resolve_dependencies()` to get the names of the dependencies. So I added to `BaseRouteHandler`:

```python
    @property
    def dependency_name_set(self) -> Set[str]:
        """
        The set of all dependency names provided in the handler's ownership layers.

        Intended as a fast to compute set of the names of dependencies provided to the handler, and
        available at the time that the handler's signature model is generated. Full resolution of
        dependencies requires that the signature model is already generated and is performed in
        ``BaseRouteHandler.resolve_dependencies()``.
        """
        if self.resolved_dependency_name_set is BaseRouteHandler.empty:
            self.resolved_dependency_name_set = {
                k
                for k in itertools.chain.from_iterable(
                    (layer.dependencies or {}).keys() for layer in self.ownership_layers()
                )
            }
        return cast(Set[str], self.resolved_dependency_name_set)
```

Also added a section to the bottom of the usage docs in `6-dependency-injection.md`.

LMK any changes you'd like and I'll try to close this one out so I can clean up the other couple of PRs that are there.

Would you prefer a squash and merge for a PR like this one?

starlite/handlers/base.py
```diff
@@ -63,9 +66,29 @@ def __init__(
...
+        Intended as a fast to compute set of the names of dependencies provided to the handler, and
+        available at the time that the handler's signature model is generated. Full resolution of
+        dependencies requires that the signature model is already generated and is performed in
+        ``BaseRouteHandler.resolve_dependencies()``.
+        """
+        if self.resolved_dependency_name_set is BaseRouteHandler.empty:
+            self.resolved_dependency_name_set = {
+                k
+                for k in itertools.chain.from_iterable(
+                    (layer.dependencies or {}).keys() for layer in self.ownership_layers()
```
Goldziher, line 87: Please split this into variable rather than nest it

Goldziher (approved):
Great stuff. Merge when ready .

PR #145, opened 2022-06-17
Simple tox setup for testing across multiple py versions.

peterschutt:
Instructions in `CONTRIBUTING.md`

PR #146, opened 2022-06-17
Tests allow untyped decorators

peterschutt:
We were individually ignoring lots of decorated functions because of this so better to ignore at the config level.

This is based off #145 so needs to merge after that.


Goldziher (approved):
Ok. Next time simply point the other PR at this one, then there will be a smaller diff.

Issue #142, opened 2022-06-15
`Provide` and `@classmethod`


peterschutt:
`Provide(Something.new)` where `new()` is an `@classmethod` doesn't work.

repro:

```python-console
asyncio REPL 3.10.2 (main, Jan 26 2022, 20:19:57) [GCC 10.2.1 20210110] on linux
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from app.domain.providers.entities.service import Service
>>> Service.new
<bound method Service.new of <class 'app.domain.providers.entities.service.Service'>>
>>> await Service.new()
3
>>> from starlite import Provide
>>> await Provide(Service.new).dependency()
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 446, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
TypeError: Service.new() takes 1 positional argument but 2 were given
```

Stems from here:
https://github.com/starlite-api/starlite/blob/4c411b4e260c1db0d1e8f9beba3e888e7f1e7a89/starlite/provide.py#L22-L24

I think the current pattern is to prevent the callable becoming an instance method of the `Provide` instance, which perhaps would be better done with `self.dependency = staticmethod(dependency)`... but I haven't tested it or anything yet.

Goldziher:
Well, the issue is that if self is used inside the method, making it static will break the code. This is why this is in place. But there is a test for it - 

```python
def test_provide_method() -> None:
    class MyClass:
        def my_method(self) -> None:
            assert self.__class__ is MyClass

    provider = Provide(dependency=MyClass().my_method)
    assert isinstance(provider.dependency, partial)
    assert isinstance(provider.dependency.args[0], MyClass)
```

The issue above seems to be you are using a static method, which is still being forwarded `self`. So the bug here is that static methods are not handled correctly by provide. Am I correct?

peterschutt:
> The issue above seems to be you are using a static method, which is still being forwarded self. So the bug here is that static methods are not handled correctly by provide. Am I correct?

In my case, `Service.new` is a `@classmethod`.

At least in 3.10, assigning an instance method to another class instance doesn't seem to overwrite `self`:

```python-console
>>> class C1:
...   def method_1(self):
...     return self.__class__
... 
[...]

PR #147, opened 2022-06-17
Fix `Provide` for `@classmethod`

peterschutt:
Fix for #142 where `Provide` wasn't working for `@classmethod`s.

Adds tests for each type of callable that might be given to provide.

Based off #146 so will need to merge after that.

### DIFF
--- a/starlite/provide.py
+++ b/starlite/provide.py
@@ -11,7 +11,7 @@ from starlite.signature import SignatureModel
 
 
 class Provide:
-    __slots__ = ("dependency", "use_cache", "value", "signature_model", "sync_to_thread")
+    __slots__ = ("dependency", "use_cache", "value", "signature_model", "sync_to_thread", "is_coro")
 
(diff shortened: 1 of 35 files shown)


starlite/provide.py
```diff
@@ -11,17 +11,19 @@
...
         self.use_cache = use_cache
         self.value: Any = Undefined
         self.signature_model: Optional[Type[SignatureModel]] = None
         self.sync_to_thread = sync_to_thread
-        if ismethod(dependency) and hasattr(dependency, "__self__"):
-            # ensure that the method's self argument is preserved
-            self.dependency = partial(dependency, dependency.__self__)
+        # manage async partial objects in 3.7 (https://stackoverflow.com/a/52422903/6560549)
+        if isinstance(dependency, partial):
+            self.is_coro = iscoroutinefunction(dependency.func)
```
Goldziher, line 24: I prefer being explicit rather than using aliases, so `is_coroutine` instead of `is_coro`

Goldziher (approved):
Great stuff.

PR #152, opened 2022-06-17
Add pydantic-factories testing examples

### LINKED ISSUE #143: pydantic-factories testing examples
Testing page in docs has a tip that `pydantic-factories` package is bundled with Starlite. The problem is that there’s no example why it is useful. As I can see Starlite uses the package mostly for generating OpenAPI examples. I think we should either don’t mention `pydantic-factories` on testing page or add an example how to use it.

vrslev:
In this PR I added mini-tutorial for using pydantic-factories in tests.
Also corrected markup and made subtitles on Testing page more consistent.
Fixes #143

### DIFF
--- a/docs/usage/14-testing.md
+++ b/docs/usage/14-testing.md
@@ -214,7 +214,6 @@ def test_get_item(item: Item):
         assert response.json() == item.dict()
 ```
 
-That's better! Now we don't have to think about this not important thing.
 
 ## Creating a Test Request
 
--- a/docs/usage/14-testing.md
+++ b/docs/usage/14-testing.md
@@ -2,11 +2,6 @@
 
 Testing is a first class citizen in Starlite, which offers several powerful testing utilities out of the box.
 
-<!-- prettier-ignore -->
-!!! tip
-    Starlite bundles the library [pydantic-factories](https://github.com/Goldziher/pydantic-factories), which offers an
-    easy and powerful way to generate mock data from pydantic models and dataclasses.
-
 ## Test Client
 
 Starlite extends the Starlette testing client, which in turn is built using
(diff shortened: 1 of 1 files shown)


docs/usage/14-testing.md
```diff
@@ -116,7 +111,112 @@ def test_health_check():
         assert response.text == "healthy"
 ```
 
-## Create a Test Request
+## Using pydantic-factories
+
+Starlite bundles the library [pydantic-factories](https://github.com/Goldziher/pydantic-factories), which offers an easy and powerful way to generate mock data from pydantic models and dataclasses.
+
+Let's say we have an API that talks to external service (that we don't care about in unit tests) and retrieves some item:
```
Goldziher, line 118:
```suggestion
Let's say we have an API that talks to an external service and retrieves some data:
```

docs/usage/14-testing.md
```diff
@@ -116,7 +111,112 @@ def test_health_check():
...
+    def get(self) -> Item:
+        ...
+
+
+@get(path="/item")
+def get_item(service: Service) -> Item:
+    return service.get()
+```
+
+Then we can add a test for `/item` route:
```
Goldziher, line 144:
```suggestion
We could test the `/item` route like so:
```

docs/usage/14-testing.md
```diff
@@ -116,7 +111,112 @@ def test_health_check():
...
+    with create_test_client(
+        route_handlers=get_item,
+        dependencies={"service": Provide(lambda: MyService()),
+    }) as client:
+        response = client.get("/item")
+        assert response.status_code == HTTP_200_OK
+        assert response.json() == item.dict()
+```
+
+As you can see, we came up with a test product by ourselves. It is fine for small applications, but can be quite a pain in larger ones. That's where [pydantic-factories](https://github.com/Goldziher/pydantic-factories) library comes in. It generates mock data for pydantic datastructures based on type annotations. Let's rewrite our example using it:
```
Goldziher, line 166:
```suggestion
While we can define the test data manually, as is done in the above, this can be quite cumbersome. That's where [pydantic-factories](https://github.com/Goldziher/pydantic-factories) library comes in. It generates mock data for pydantic models and dataclasses based on type annotations. With it, we could rewrite the above example like so:
```

docs/usage/14-testing.md
```diff
@@ -116,7 +111,112 @@ def test_health_check():
...
+    with create_test_client(
+        route_handlers=get_item,
+        dependencies={"service": Provide(lambda: MyService()),
+    }) as client:
+        response = client.get("/item")
+        assert response.status_code == HTTP_200_OK
+        assert response.json() == item.dict()
+```
+
+That's better! Now we don't have to think about this not important thing.
```
Goldziher, line 217:
```suggestion
```

Goldziher (changes requested):
Looks good, and thanks! Some suggested rewrites.

vrslev:
Great, thanks!

Issue #154, opened 2022-06-18
204 status code not handled properly


vrslev:
Exception raises when route returns response with 204 No Content status.
<details>

```console
INFO:     127.0.0.1:58657 - "DELETE / HTTP/1.1" 204 No Content
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/starlite/app.py", line 149, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/starlite/asgi.py", line 70, in __call__
    await route.handle(scope=scope, receive=receive, send=send)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/starlite/routes.py", line 161, in handle
    await response(scope, receive, send)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/starlette/responses.py", line 167, in __call__
    await send({"type": "http.response.body", "body": self.body})
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 462, in send
    output = self.conn.send(event)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/h11/_connection.py", line 510, in send
    data_list = self.send_with_data_passthrough(event)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/h11/_connection.py", line 543, in send_with_data_passthrough
    writer(event, data_list.append)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/h11/_writers.py", line 65, in __call__
    self.send_data(event.data, write)
  File "/private/tmp/starlite-fetch-error/venv/lib/python3.10/site-packages/h11/_writers.py", line 91, in send_data
    raise LocalProtocolError("Too much data for declared Content-Length")
h11._util.LocalProtocolError: Too much data for declared Content-Length

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #155, opened 2022-06-19
Empty response content for 204 status and `None` handler return value

peterschutt:
Where a handler returns `None` and the status is 204 we assume the response content should be empty, `b''` rather than JSON `null`.


peterschutt:
Is this a fair enough premise @Goldziher?

peterschutt:
Closes #154

PR #160, opened 2022-06-20
Issue 159 exception handlers

### LINKED ISSUE #159: passing validation error to frontend
Similar to previous issue [#102](https://github.com/starlite-api/starlite/issues/102)

It never reaches the "VALIDATION FAILED" as its catching the validation error and returning a JSON response.

How can I pass it through?

```python
class User(BaseModel):
    username: str
    password: str

class LoginController(Controller):
    path = ""

    @get(path="/login")
    async def login(self) -> Template:
        return Template(name="login.html")

    @post(path="/login")
    async def post_login(
        self,
        data: User = Body(media_type=RequestEncodingType.URL_ENCODED),
    ) -> None:
        if "VALIDATION FAILED":
            return Template(name="login.html", context="VALIDATION ERRORS")

        response = RedirectResponse(url=f"/profile", status_code=303)
        response.set_cookie("MY_COOKIE", "1")

        return response
```

Goldziher:
This PR updates exception handling by allowing users to define exception handlers on the `Router`, `Controller` or `HTTPRouteHandler` (e.g. `get`, `delete` etc.). This closes #159.

I decided to restrict this to http route handlers because both `@websocket` and `@asgi` do not fall into the same kind of category.

Please let me know what you think of the implementation.

### DIFF
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -100,6 +100,9 @@ class HTTPRouteHandler(BaseRouteHandler):
         status_code: Optional[int] = None,
         cache: Union[bool, int] = False,
         cache_key_builder: Optional[CacheKeyBuilder] = None,
+        exception_handlers: Optional[Dict[Union[int, Type[Exception]], ExceptionHandler]] = None,
+        # sync only
+        sync_to_thread: bool = False,
         # OpenAPI related attributes
         content_encoding: Optional[str] = None,
         content_media_type: Optional[str] = None,
@@ -111,9 +114,6 @@ class HTTPRouteHandler(BaseRouteHandler):
         response_description: Optional[str] = None,
         summary: Optional[str] = None,
         tags: Optional[List[str]] = None,
-        # sync only
-        sync_to_thread: bool = False,
-        exception_handlers: Optional[Dict[Union[int, Type[Exception]], ExceptionHandler]] = None,
     ):
         if not http_method:
             raise ImproperlyConfiguredException("An http_method kwarg is required")
@@ -376,6 +376,9 @@ class get(HTTPRouteHandler):
         status_code: Optional[int] = None,
         cache: Union[bool, int] = False,
         cache_key_builder: Optional[CacheKeyBuilder] = None,
+        exception_handlers: Optional[Dict[Union[int, Type[Exception]], ExceptionHandler]] = None,
+        # sync only
+        sync_to_thread: bool = False,
         # OpenAPI related attributes
         content_encoding: Optional[str] = None,
         content_media_type: Optional[str] = None,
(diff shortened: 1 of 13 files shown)


starlite/utils/exception.py
```diff
@@ -0,0 +1,30 @@
...
+) -> Optional[ExceptionHandler]:
+    """
+    Given a dictionary that maps exceptions and status codes to handler functions,
+    and an exception, returns the appropriate handler if existing.
+    """
+    if not exception_handlers:
+        return None
+    if isinstance(exc, HTTPException) and exc.status_code in exception_handlers:
+        return exception_handlers[exc.status_code]
+    if exc.__class__ in exception_handlers:
```
peterschutt, line 21: Could also be `type(exc) in exception_handlers`

starlite/utils/exception.py
```diff
@@ -0,0 +1,30 @@
...
+        return exception_handlers[exc.__class__]
+    if HTTP_500_INTERNAL_SERVER_ERROR in exception_handlers:
+        return exception_handlers[HTTP_500_INTERNAL_SERVER_ERROR]
+    for key, value in exception_handlers.items():
```
peterschutt, line 25: Another way to do this could be to walk the mro of the the exception type and check for an entry in `exception_handlers` for the class and each of it's bases, e.g., 

```python
def get_exception_handler(...) -> ...:
    if not exception_handlers:
        return None
    if isinstance(exc, HTTPException) and exc.status_code in exception_handlers:
        return exception_handlers[exc.status_code]
    for cls in type(exc).__mro__:
        if cls not in exception_handlers:
            continue
        return exception_handlers[cls]
    return None
```
I haven't tested though.

starlite/handlers/http.py
```diff
@@ -239,6 +253,22 @@ def resolve_after_request(self) -> Optional[AfterRequestHandler]:
...
 
+    def resolve_exception_handlers(self) -> Dict[Union[int, Type[Exception]], ExceptionHandler]:
+        """
+        Resolves the exception_handlers by starting from the route handler and moving up.
+
+        This method is memoized so the computation occurs only once.
+        """
+        if self.resolved_exception_handlers is BaseRouteHandler.empty:
+            exception_handlers: Dict[Union[int, Type[Exception]], ExceptionHandler] = {}
+            for layer in self.ownership_layers():
```
peterschutt, line 264: If you reverse the ownership layers you can skip the checks for whether the key exists or not as each layer will overwrite the previous, e.g.:

```python
...
for layer in reversed(self.ownership_layers):
    exception_handlers.update(layer.exception_handlers or {})
self.resolved_exception_handlers = exception_handlers
...
```

peterschutt (approved):
LGTM. A couple of comments there that don't change any behavior (I think!) and simplify the code a little. Feel free to adopt them, or not.

peterschutt:
Once you merge this is, I'll address any review points you have for me in #162 and then I'll cut the point release this evening my time.

Goldziher:
> Once you merge this is, I'll address any review points you have for me in #162 and then I'll cut the point release this evening my time.

excellent review, thanks. I applied all your suggestions.

PR #162, opened 2022-06-21
Adds `utils.is_async_callable()` `tests.test_utils` module.

peterschutt:
`is_async_callable()` is a replacement for `iscoroutinefunction()` that detects more cases and handles partial objects.


Goldziher (approved):
Nice

PR #168, opened 2022-06-23
Conform exceptions to the existing OpenAPI schema

### LINKED ISSUE #167: OpenAPI schema generation includes `status_code` but default exception handler never outputs that attribute
I was doing some endpoint schema validation for my project and ran into what seems like an inconsistency between the generated OpenAPI schema and reality.

Specifically, `HTTPException` has an (optional) `status_code` attribute which is typically set to the response status code when something exceptional happens.
The fact that this field should always be present in the output is documented in the generated schema, but in the code it is omitted in all cases in the default exception handler:
```py
if isinstance(exc, HTTPException):
    content = {"detail": exc.detail, "extra": exc.extra}
```

This currently makes my tests fail because the schema says status_code is required, but starlite will never output it.

There are two possible fixes for this:
1. Fix the OpenAPI schema generator not to include the field as required (or at all)
2. Update the default exception handler to include the field when it exists

timwedde:
Fixes #167 by adding `status_code` to the response body when dealing with exceptions in the default exception handler.

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -283,7 +283,7 @@ class Starlite(Router):
             server_middleware = ServerErrorMiddleware(app=self)
             return server_middleware.debug_response(request=request, exc=exc)
         if isinstance(exc, HTTPException):
-            content = {"detail": exc.detail, "extra": exc.extra}
+            content = {"detail": exc.detail, "extra": exc.extra, "status_code": exc.status_code}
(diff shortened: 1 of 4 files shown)


timwedde:
Ah, apologies, I see that the test failed. I forgot to run the test suite before pushing, will fix this right now.

PR #172, opened 2022-06-24
Provides a callback interface for modifying the instantiated template engine

### LINKED ISSUE #170: template additional config
How can this be achieved with the current way the template config is used?

```python
templates = Jinja2Templates(directory="templates")
templates.env.globals[’flashed_messages’] = flashed_messages
```

peterschutt:
We cannot really have a consistent api for passing kwargs to the underlying engine type due to supporting any template engine, so a callback allows the user to receive the instantiated template engine instance to modify it.

Refactor of template tests.

Closes #170


Goldziher (approved):
Ok, I must say this is very "node" and less pythonic than I would like. On the other hand it's a working solution, and I'm all for pragmatism. 

So I'm approving, but I leave you with a question - isn't it simpler to modify the 'TemplateEngineProtocol' to accept either a class or an instance and then allow users to simply instantiate on their own and pass the instance to the engine constructor if they want to instantiate if with custom logic?

I leave it to you to decide how to proceed here, both are acceptable solutions in my book. Merge when you feel ready.

peterschutt:
The `**kwargs` pass-through approach doesn't fix the issue of setting jinja global env as the `Environment` constructor doesn't pass any args through to their global state.

So my thinking here was that this solution changes 4 lines of production code, doesn't modify any protocols and provides access to the engine instance that we instantiate, which at least solves the OPs issue, but should be pretty flexible too. 

It certainly doesn't close the door for more work on this interface if people end up hating the pattern, but I'd rather put my time toward #173 then refactor this - at least for now.

BTW, the test refactoring is just my way of getting to know the code, so hope you don't mind that stuff.

Goldziher:
All good. You can merge when ready

PR #183, opened 2022-06-27
Discourage testing import from package, fix star import from `starlite`

### LINKED ISSUE #181: `from starlite import *` broken if `testing` extra not installed
This is only an issue on main, not in any release.

When I want to try a library out, I'll install it into a fresh env, run python repl and do `from lib import *` and have a play around. If just doing that raised an error it would freak me out a little about the lib.

Possible solution:
- remove `.testing` imports from `starlite.__all__`
- add deprecation warning for top-level `.testing` imports
- remove `if TYPE_CHECKING` too? May as well if we are doing the above, I think?

Refs: #174 #130

Bobronium:
Fixes #181

I'm hesitant on adding `DeprecationWarning` just yet. 

As I see it, we can deal with deprecation in tho fases:
1. Static: encourage importing from `starlite.testing` by removing testing names it from `__all__`, IDEs will pick this up
2. Runtime: emit warning on import

This PR is executing the 1st phase. This should give people some time to get used to importing it from `starlite.testing`, and then we can explicitly deprecate importing it in runtime as well.

### DIFF
--- a/starlite/__init__.py
+++ b/starlite/__init__.py
@@ -120,15 +120,23 @@ __all__ = [
     "websocket",
 ]
 
-# these imports are still allowed, but importing from starlite.testing instead is encouraged
-_dynamic_imports = {"TestClient", "create_test_client", "create_test_request"}
+
+_deprecated_imports = {"TestClient", "create_test_client", "create_test_request"}
 
 
 # pylint: disable=import-outside-toplevel
 def __getattr__(name: str) -> Any:
     """Provide lazy importing as per https://peps.python.org/pep-0562/"""
-    if name not in _dynamic_imports:
+    if name not in _deprecated_imports:
         raise AttributeError(f"Module {__package__} has no attribute {name}")
+
+    import warnings
+
+    warnings.warn(
+        f"Importing {name} from {__package__} is deprecated, use `from startlite.testing import {name}` instead",
+        DeprecationWarning,
+        stacklevel=2,
+    )
 
     from . import testing
 
--- a/starlite/__init__.py
+++ b/starlite/__init__.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
 from starlite.datastructures import File, Redirect, State, Stream, Template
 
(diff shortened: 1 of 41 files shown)


peterschutt:
Thanks for rounding this out!

I wouldn't mind if the deprecation was included here, at least that would explain the IDE/linting changes but up to you. How long should we wait to add the deprecation in your view?

Bobronium:
Trick I've used to change imports (requires PyCharm/other refactoring tools):

1. Put this in `starlite/__init__.py`:
  ```py
  class TestClient: ...
  def create_test_client(): ...
  def create_test_request(): ...
  ```
2. remove these names from `starlite/testing.py`
3. Select code from step 1, Refactor -> Move... -> `starlite/testing.py`
4. Restore content of `starlite.testing,py`

For those who use starlite as a dependency, equivalent for 1. and 2. would be:
  ```shell
  mkdir starlite && touch starlite/testing.py && echo "class TestClient: ...\ndef create_test_client(): ...\ndef create_test_request(): ...\n" > starlite/__init__.py
  ```
Then do `3.` and after that just `rm -rf starlite`

Bobronium:
> I wouldn't mind if the deprecation was included here, at least that would explain the IDE/linting changes but up to you. 

Actually, it makes more sense, as I think about it now. I'll add it here.

peterschutt:
> Trick I've used to change imports (requires PyCharm/other refactoring tools):
> 
>     1. Put this in `starlite/__init__.py`:
> 
> 
> ```python
> class TestClient: ...
> def create_test_client(): ...
> def create_test_request(): ...
> ```
> 
>     2. remove these names from `starlite/testing.py`
> 
>     3. Select code from step 1, Refactor -> Move... -> `starlite/testing.py`
> 
>     4. Restore content of `starlite.testing,py`
> 
> 
> For those who use starlite as a dependency, equivalent for 1. and 2. would be:
> 
> ```shell
> mkdir starlite && touch starlite/testing.py && echo "class TestClient: ...\ndef create_test_client(): ...\ndef create_test_request(): ...\n" > starlite/__init__.py
> ```
> 
> Then do `3.` and after that just `rm -rf starlite`

Neat!

Bobronium:
BTW, found a way to do it without pycharm

Prequisites:
- [ropecli](https://github.com/aevri/ropecli)
    ```shell
    pip install git+https://github.com/aevri/ropecli.git
    ```
Command:
```shell
mkdir starlite \
  && touch starlite/testing.py \
  && echo "class TestClient: ...\ndef create_test_client(): ...\ndef create_test_request(): ..." >starlite/__init__.py \
  && rope move "starlite/__init__.py::*" starlite/testing.py \
  && rm -rf starlite
```

This however will turn usages of `TestClient`, etc. into `testing.TestClient`.

Goldziher (approved):
Very nice

PR #186, opened 2022-06-27
Change to order of exception handler resolution.

### LINKED ISSUE #176: `get_exception_handler()` util resolve ordering.
https://github.com/starlite-api/starlite/blob/da79c068c906740a4375d431953f1c74fdea8067/starlite/utils/exception.py#L10-L28

I've registered an exception handler to `500_INTERNAL_SERVER_ERROR` that logs the exception but it is also catching and logging `HTTPException` instances. Even if I register an exception handler to `HTTPException` to try to avoid that, because I also have a handler registered for `500` that conditional is always hit and the handler returned before it has the chance to do the walk MRO thing.

So I think the mro walk should be above the `500` fallback, and then I think we should think of whether we'd actually want the `HTTPException` subclasses to be pulled in with the `500` handler if there isn't a narrower mapping.

This is the registration mapping:

```python
    exception_handlers={
        HTTP_500_INTERNAL_SERVER_ERROR: logging_exception_handler,
        HTTPException: passthrough_exception_handler,  # type:ignore[dict-item]
    },
```

But `HTTPException` sub-classes are getting sucked up by this:

```python
     if HTTP_500_INTERNAL_SERVER_ERROR in exception_handlers: 
         return exception_handlers[HTTP_500_INTERNAL_SERVER_ERROR] 
```

... before the mro check can find the base class mapping.

peterschutt:
- check the MRO before checking for a 500 handler
- only use the 500 handler for non-HTTPException exceptions
- description of algo in docstring
- add tests with multiple handlers in `exception_handlers` mapping

Closes #176

### DIFF
--- a/starlite/utils/exception.py
+++ b/starlite/utils/exception.py
@@ -1,5 +1,5 @@
 from inspect import getmro
-from typing import Dict, Optional, Type, Union, cast
+from typing import Dict, Optional, Type, Union
 
 from starlette.exceptions import HTTPException
 from starlette.status import HTTP_500_INTERNAL_SERVER_ERROR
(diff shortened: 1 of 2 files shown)


Goldziher (approved):
🔥

PR #190, opened 2022-06-28
Run middleware on exceptions

### LINKED ISSUE #175: Middleware are not executed on Starlite error
When route is not found, method is not allowed, or some similar error, then middleware are not executed although and it is unexpected behaviour.

Starlite, second test fails:
```python
from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
from starlette.responses import Response
from starlite import Starlite, TestClient, get


class ServerHeaderMiddleware(BaseHTTPMiddleware):
    async def dispatch(
        self, request: Request, call_next: RequestResponseEndpoint
    ) -> Response:
        response = await call_next(request)
        response.headers["Server"] = "NULL"
        return response


@get()
async def route() -> str:
    return ""


app = Starlite(
    debug=True,
    route_handlers=[route],
    middleware=[ServerHeaderMiddleware],
)


def test_middleware_is_executed_when_route_found():
        response = client.get("/")
        assert response.headers["Server"] == "NULL"


def test_middleware_is_executed_when_route_not_found():
        response = client.get("/non-existing-route")
        assert response.headers["Server"] == "NULL"
```

Starlette, tests pass:
```python
from starlette.applications import Starlette
from starlette.middleware import Middleware
from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
from starlette.responses import PlainTextResponse, Response
from starlette.routing import Route
from starlette.testclient import TestClient


class ServerHeaderMiddleware(BaseHTTPMiddleware):
    async def dispatch(
        self, request: Request, call_next: RequestResponseEndpoint
    ) -> Response:
        response = await call_next(request)
        response.headers["Server"] = "NULL"
        return response


async def route(request: Request):
    return PlainTextResponse("")


app = Starlette(
    debug=True,
    routes=[Route("/", route, methods=["GET"])],
    middleware=[Middleware(ServerHeaderMiddleware)],
)


def test_middleware_is_executed_when_route_found():
        response = client.get("/")
        assert response.headers["Server"] == "NULL"


def test_middleware_is_executed_when_route_not_found():
        response = client.get("/non-existing-route")
        assert response.headers["Server"] == "NULL"
```

vrslev:
This PR is here to simplify discussion around code @peterschutt proposed to *potentially* resolve #175.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
             "flake8-simplify",
           ]
   - repo: https://github.com/ariebovenberg/slotscheck
-    rev: v0.14.0
+    rev: v0.14.1
     hooks:
       - id: slotscheck
         exclude: "^(?!starlite/)"
--- a/docs/usage/15-templating.md
+++ b/docs/usage/15-templating.md
@@ -68,3 +68,22 @@ class TemplateEngineProtocol(Protocol[T]):
 ```
 
 Once you have your custom engine you can regiter it as you would the built-in engines.
+
+## Modifying the Template Engine Instance
+
+`TemplateConfig` accepts the `engine_callback` keyword arg which provides a way to modify the instantiated
+template engine instance. For example:
+
+```python
+def engine_callback(jinja_engine: JinjaTemplateEngine) -> JinjaTemplateEngine:
+    jinja_engine.engine.globals["key"] = "value"
+    return jinja_engine
+
+
+template_config = TemplateConfig(
+    directory="templates", engine=JinjaTemplateEngine, engine_callback=engine_callback
+)
+```
+
+The callback should receive a single argument which will be the instantiated engine, and must
+return the instantiated engine.
--- a/docs/usage/7-middleware.md
+++ b/docs/usage/7-middleware.md
@@ -1,15 +1,14 @@
 # Middleware
 
 Middlewares are mini ASGI apps that receive the raw request object and validate or transform it in some manner.
-Middlewares are useful when you need to operate on all incoming requests on the app level.
-
-Starlite builds on top of the [Starlette middleware architecture](https://www.starlette.io/middleware/) and is 100%
-compatible with it - and any 3rd party middlewares created for it.
 
 ## The Middleware Protocol
 
-You can build your own middleware by either subclassing the `starlette.middleware.base.BaseHTTPMiddleware` class (see the
-starlette documentation), or by creating a class that implements the Starlite `MiddlewareProtocol`.
+!!! important Starlite allows users to use [Starlette Middleware](https://www.starlette.io/middleware/) and any 3rd party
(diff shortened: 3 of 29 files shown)


vrslev:
In example I suggested (https://github.com/starlite-api/starlite/issues/175#issue-1284457869), test still fails here:
https://github.com/starlite-api/starlite/blob/bdb47e12891877d0084ae3677149df9ebf87c34d/starlite/asgi.py#L53

If `StarliteASGIRouter` fails to resolve an `ASGIApp` from route handler—i. e. matching handler not found—it raises `NotFoundException`. The thing is that all middleware sit in `ASGIApp` in route handlers. If we can't resolve a route handler, router doesn't know about _any_ middleware, including `ExceptionMiddleware`. So the issue is still present in this case.

vrslev:
The only way we can deal with this, I suppose, is to make app-wide middleware stack and don't include that middleware in route handers' stacks.

Goldziher:
> The only way we can deal with this, I suppose, is to make app-wide middleware stack and don't include that middleware in route handers' stacks.

Thanks @vrslev. I don't think this is the solution to the problem though. We will go with middleware declaration on multiplelayers and with per route handler ASGIApp computation. 

I am thinking that we also need to pre-compute, similar to the way that route handler ASGIApps are computed, ASGIApps for exceptions handlers.

You missed one part in your PR --> see the code snipper below from `starlite/route.py#L147`, which includes the route handler per route. This is only for http routes mind you - although we could consider expanding this to the other routes as well. 

```python
    async def call_handler(
        self, scope: Scope, request: Request, parameter_model: KwargsModel, route_handler: HTTPRouteHandler
    ) -> Union[Response, StarletteResponse]:
        """
        Calls the before request handlers, retrieves any data required for the route handler,
        and calls the route handler's to_response method.

        This is wrapped in a try except block - and if an exception is raised,
        it tries to pass it to an appropriate exception handler - if defined.
        """
        try:
            response_data = None
            before_request_handler = route_handler.resolve_before_request()
            # run the before_request hook handler
            if before_request_handler:
                if is_async_callable(before_request_handler):
                    response_data = await before_request_handler(request)
                else:
                    response_data = await run_sync(before_request_handler, request)
            if not response_data:
                response_data = await self.get_response_data(
                    route_handler=route_handler, parameter_model=parameter_model, request=request
                )
            return await route_handler.to_response(
                app=scope["app"],
                data=response_data,
                plugins=request.app.plugins,
            )
        except Exception as e:
            handler = get_exception_handler(route_handler.resolve_exception_handlers(), e)
            if handler:
                return handler(request, e)
            raise e
```

vrslev:
> You missed one part in your PR --> see the code snipper below from starlite/route.py#L147, which includes the route handler per route. This is only for http routes mind you - although we could consider expanding this to the other routes as well.

@Goldziher I just added the PR, code is @peterschutt's 😊

Issue #189, opened 2022-06-28
Feature: Support for middleware parameter specification


peterschutt:
Thinking about ways to make parameter specification and parsing / validation available to the middleware stack.

Perhaps something like:

```python
@runtime_checkable
class MiddlewareProtocol(Protocol):
    parameters: dict[str, tuple] = {"middleware_required_header": (int, Parameter(header="required-header"))}

    def __init__(self, app: ASGIApp):  # pragma: no cover
        ...

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:  # pragma: no cover
        # this obv. not actually defined in the protocol, but just for demo
        if self.validated_paraeters.get("middleware_required_header") != "13":
            ...
        else:
            ...
    
    @property
    def validated_parameters(self) -> dict[str, Any]:
        # this mapping actually constructed dynamically
        return {"middleware_required_header": 13}
```

Lets discuss:)

peterschutt:
Any thoughts on how you'd like this implemented @Goldziher

Goldziher:
So, there is a problem with defining concrete attributes on protocols - pydantic doesn't play nice with this. 

How about adding a container type for Middleware that allows defining stuff like parameters, descriptions etc? 

Otherwise we might need to use an abstract class instead of protocol.

peterschutt:
> So, there is a problem with defining concrete attributes on protocols - pydantic doesn't play nice with this.

OK cool, Also there is the nature of a protocol being that it's just an interface, I.e, shouldn't need to extend from the base to fulfil the protocol.

> How about adding a container type for Middleware that allows defining stuff like parameters, descriptions etc?

Can you give example of how this looks?

> Otherwise we might need to use an abstract class instead of protocol.

This is intuitive to me, but we shouldn't need to do away with the protocol altogether, we could just  have the base class that implements the protocol that users could extend for their custom middleware that allows us to handle parameter validation/parsing/docs for the parameters required by the middleware?

Goldziher:
So, as a container I'd try to do something like this probably:

```python
from pydantic import validate_arguments
from starlite.params import Parameter
from starlette.middleware.base import BaseHTTPMiddleware
from starlette.middleware import Middleware as StarletteMiddleware

class DefinedMiddleware:
    @validate_arguments(config={"arbitrary_types_allowed": True})
    def __init__(
            self,
[...]

PR #197, opened 2022-06-29
update exception handling

### LINKED ISSUE #175: Middleware are not executed on Starlite error

Goldziher:
This PR updates error handling by wrapping an asgi app 3 times inside error handlers.  Resolves #175 

Starlite - Route Handler Call Process


MrAwesomeRocks:
Why is the innermost exception handler (surrounding dependencies & route handlers) needed?

If we just had the two outer ones, middlewares such as https://github.com/bigbag/starlette-request-id should just work on everything but routing errors, which is an easier problem to solve.

Goldziher:
> Why is the innermost exception handler (surrounding dependencies & route handlers) needed?
> 
> If we just had the two outer ones, middlewares such as https://github.com/bigbag/starlette-request-id should just work on everything but routing errors, which is an easier problem to solve.

not at all. Without the innermost exception handler, exceptions will not be caught before entering middleware, and hence middlewares will not operate on exceptions. This is the core of the issue.

Goldziher:
> Oh whoops, sorry, I viewed the middleware stack backwards. My bad, I'm still learning the codebase.
> 
> Here's an idea I just had, it may or may not work, please feel free to correct me if you think of a better way to do this. There are only two routing errors, 404 and 405. Why not make the default exception handlers for those two exceptions then jump into the middleware stack somehow? Or somehow hardcode those exceptions to running the middleware before calling the exception handler? I think it could work, but it wouldn't play nicely with #189, so there would have to be a way of skipping middleware. But this should at least let us run the ASGI middleware from Starlette.
> 

I don't quite understand what you are proposing 😀

MrAwesomeRocks:
Essentially, on 404/405, have the default exception handler run the middleware stack.

starlite/middleware.py
```diff
@@ -61,3 +72,53 @@ async def authenticate_request(self, request: HTTPConnection) -> AuthenticationR
...
+            content = {"detail": exc.detail, "extra": exc.extra, "status_code": exc.status_code}
+        elif isinstance(exc, StarletteHTTPException):
+            content = {"detail": exc.detail}
+        else:
+            content = {"detail": repr(exc)}
+        return Response(
+            media_type=MediaType.JSON,
+            content=content,
+            status_code=status_code,
+        )
```
peterschutt, line 124: Feels like a nice separation of concerns to have been able to pull this stuff out of the `Starlite` class. +1.

tests/middleware/test_exception_handler_middleware.py
```diff
@@ -0,0 +1,76 @@
...
+        "detail": "starlite_exception",
+    }
+
+    response = middleware.default_http_exception_handler(
+        Request(scope={"type": "http", "method": "GET"}), AttributeError("oops")
+    )
+    assert response.status_code == HTTP_500_INTERNAL_SERVER_ERROR
+    assert json.loads(response.body) == {
+        "detail": repr(AttributeError("oops")),
+    }
```
peterschutt, line 76: Just a comment. You may have seen I like to split my unit tests up so that one function tests one thing, and I give the function a descriptive name.

When tests are structured like this, in order to work out what has been tested I need to go through each logical test within that single function and mentally parse it. If there were 6 separate functions there each named something like `test_default_http_exception_handling_ensure_extra_in_json_response()`, it would be easier to get a fly-over of what has been tested.

starlite/app.py
```diff
@@ -138,28 +134,20 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
-        try:
-            scope["state"] = {}
-            await self.asgi_router(scope, receive, send)
-        except Exception as e:  # pylint: disable=broad-except
-            await self.handle_exception(scope=scope, receive=receive, send=send, exc=e)
+        scope["state"] = {}
+        await self.wrap_in_exception_handler(
+            app=self.asgi_router,
+            exception_handlers=self.exception_handlers,
+        )(scope, receive, send)
```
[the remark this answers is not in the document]
Goldziher: Sure

tests/middleware/test_exception_handler_middleware.py
Goldziher: Will change

Goldziher:
> Essentially, on 404/405, have the default exception handler run the middleware stack.

hmm, you might have a point there regarding CORS at least.

starlite/app.py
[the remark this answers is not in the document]
Goldziher: done

PR #200, opened 2022-06-30
Fix mypy errors after openapi-schema-pydantic update.

peterschutt:
Tidy up unnecessary `# type:ignore`'s and add in a few new ones.


peterschutt:
Issue was mine.

PR #201, opened 2022-06-30
Refactor of `model_function_signature()` to address complexity limits.

peterschutt:
- Should be no logical changes.
- Breaks out detailed logic into methods.
- High-level business logic in `.model()` method.
- Added a type to hold interesting properties of the parameters as we iterate over them.
- Docstrings everywhere.

### DIFF
--- a/starlite/signature.py
+++ b/starlite/signature.py
@@ -27,6 +27,8 @@ from starlite.plugins.base import PluginMapping, PluginProtocol, get_plugin_for_
 from starlite.utils.dependency import is_dependency_field
 from starlite.utils.typing import detect_optional_union
 
+UNDEFINED_SENTINELS = {Undefined, Signature.empty}
+
 
 class SignatureModel(BaseModel):
     class Config(BaseConfig):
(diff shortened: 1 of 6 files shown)


starlite/signature.py
```diff
@@ -60,63 +76,247 @@ def parse_values_from_connection_kwargs(
...
+    @property
+    def default_defined(self) -> bool:
+        """
+        `True` if `self.default` is not one of the undefined sentinel types.
+
+        Returns
+        -------
+        bool
+        """
+        return self.default not in {Signature.empty, Undefined}
```
Goldziher, line 99: Predeclare the set outside the class, it will make things faster.

starlite/signature.py
```diff
@@ -60,63 +76,247 @@ def parse_values_from_connection_kwargs(
...
-            if default not in [signature.empty, Undefined]:
-                field_definitions[kwarg] = (type_annotation, default)
-                defaults[kwarg] = default
-            elif not type_optional:
-                field_definitions[kwarg] = (type_annotation, ...)
-            else:
-                field_definitions[kwarg] = (type_annotation, None)
-        model: Type[SignatureModel] = create_model(
-            fn_name + "_signature_model", __base__=SignatureModel, **field_definitions
+    name: str
```
Goldziher, line 85: Please add a `__slots__` dunder

starlite/signature.py
```diff
@@ -60,63 +76,247 @@ def parse_values_from_connection_kwargs(
...
+    field_definitions : dict[str, Tuple[Any, Any]
+        Maps parameter name to the `(<type>, <default>)` tuple passed to `pydantic.create_model()`.
+    defaults : dict[str, Any]
+        Maps parameter name to default value, if one defined.
+    dependency_name_set : set[str]
+        The names of all known dependency parameters.
+    """
+
+    # names of fn params not included in signature model.
+    SKIP_NAMES = {"self", "cls"}
```
Goldziher, line 159: Also here add a __slots__ dunder please

starlite/signature.py
```diff
@@ -60,63 +76,247 @@ def parse_values_from_connection_kwargs(
...
+        signature that should be included in the `SignatureModel` type.
+
+        Returns
+        -------
+        Generator[SignatureParameter, None, None]
+        """
+        for name, parameter in self.signature.parameters.items():
+            if name in self.SKIP_NAMES:
+                continue
+            yield SignatureParameter.new(self.fn_name, name, parameter)
```
Goldziher, line 284: Why use new rather than a standard constructor?

Goldziher (approved):

starlite/signature.py
peterschutt: If I think of a good reason I'll let you know

PR #205, opened 2022-06-30
Raise exception on routes with duplicate path parameters (#204)

### LINKED ISSUE #204: Duplicate path parameters should raise exception
~Just beginning to learn the codebase, but excited to contribute something substantial in the future 😄 

Currently, routes with duplicate path parameters can be registered to a handler.  The conflict is resolved by using the value associated to the last occurrence of the duplicated parameters (see [here](https://github.com/starlite-api/starlite/blob/c7057dc59298eb0bcdc9acc325157a2d0a4addd1/starlite/routes.py#L76-L83)).

I would think an exception should be raised when registering the route, or at least that's the behavior I'm used to seeing.  The test below currently passes, as a demonstration.

### Current behavior
```python
def test_that_currently_passes() -> None:
    @get(path=["/{some_id:int}/foo/{some_id:int}"], media_type=MediaType.TEXT)
    def handler_fn(some_id: int) -> str:
        return str(some_id)

    with create_test_client(handler_fn) as client:
        resp = client.get("/1/foo/2")
        assert resp.status_code == HTTP_200_OK
        assert resp.text == "2"
```

### Proposed behavior
```python
def test_duplicate_path_param_validation() -> None:
    @get(path="/{param:int}/foo/{param:int}")
    def handler_fn(param: int) -> None:
        pass

        Starlite(route_handlers=[handler_fn])
```

### Possible fix
A simple change when creating the kwargs model that checks for repeated params.  Currently param names are added to a set without checking for repeats.
```python
def create_handler_kwargs_model(self, route_handler: BaseRouteHandler) -> KwargsModel:
    """
    Method to create a KwargsModel for a given route handler
    """
    dependencies = route_handler.resolve_dependencies()
    signature_model = get_signature_model(route_handler)
    path_parameters = set()
    for param in self.path_parameters:
        param_name = param["name"]
        if param_name in path_parameters:
                f"Duplicate path parameters should not be used. Duplicate parameter {{{param_name}}} found in path {self.path}"
            )
        path_parameters.add(param_name)
    return KwargsModel.create_for_signature_model(
        signature_model=signature_model, dependencies=dependencies, path_parameters=path_parameters
    )
```
<br>

Happy to push this, I just still have a very shallow grasp of the codebase so am probably not doing this the best way!

danesolberg:
Fixes #204 

Proposed changes to handling duplicate / repeated path parameters in routes.  Raises `ImproperlyConfiguredException` while building the kwargs model for the route that contains the repeated params.

### DIFF
--- a/starlite/routes.py
+++ b/starlite/routes.py
@@ -93,9 +93,7 @@ class BaseRoute:
         for param in self.path_parameters:
             param_name = param["name"]
             if param_name in path_parameters:
-                raise ImproperlyConfiguredException(
-                    f"Duplicate path parameters should not be used. Duplicate parameter {{{param_name}}} found in path {self.path}"
-                )
+                raise ImproperlyConfiguredException(f"Duplicate parameter '{param_name}' detected in '{self.path}'.")
             path_parameters.add(param_name)
         return KwargsModel.create_for_signature_model(
             signature_model=signature_model, dependencies=dependencies, path_parameters=path_parameters
--- a/starlite/routes.py
+++ b/starlite/routes.py
@@ -89,7 +89,14 @@ class BaseRoute:
         """
         dependencies = route_handler.resolve_dependencies()
         signature_model = get_signature_model(route_handler)
-        path_parameters = {p["name"] for p in self.path_parameters}
+        path_parameters = set()
+        for param in self.path_parameters:
+            param_name = param["name"]
+            if param_name in path_parameters:
+                raise ImproperlyConfiguredException(
+                    f"Duplicate path parameters should not be used. Duplicate parameter {{{param_name}}} found in path {self.path}"
+                )
+            path_parameters.add(param_name)
         return KwargsModel.create_for_signature_model(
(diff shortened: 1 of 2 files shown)


starlite/routes.py
```diff
@@ -89,7 +89,14 @@ def create_handler_kwargs_model(self, route_handler: BaseRouteHandler) -> Kwargs
         """
         dependencies = route_handler.resolve_dependencies()
         signature_model = get_signature_model(route_handler)
-        path_parameters = {p["name"] for p in self.path_parameters}
+        path_parameters = set()
+        for param in self.path_parameters:
+            param_name = param["name"]
+            if param_name in path_parameters:
+                raise ImproperlyConfiguredException(
+                    f"Duplicate path parameters should not be used. Duplicate parameter {{{param_name}}} found in path {self.path}"
```
peterschutt, line 97: Is the `{{{...}}}` so that it is formatted like a path param?

How about:
`f"Duplicate parameter '{param_name}' detected in '{self.path}'."`

peterschutt (commented):
Thanks for this.

Issue #131, opened 2022-06-07
Dark docs theme is challenging to read


peterschutt:
.. at least for me.

![image](https://user-images.githubusercontent.com/20659309/172288722-4cee6933-d28b-4f13-b748-c69f23599bce.png)

The dark text on dark background is the issue.

It looks like this changed recently?

Goldziher:
Perhaps due to the recent updates to mkdocs. Let's ask @dkress59 to assist with improving the docs.

peterschutt:
Thanks @dkress59!

peterschutt:
Whatever the cause, it was introduced in 1.3.9 with mkdocs-material bumped from 8.2.16 to 8.3..

Maybe we should pin to 8.2 for now?

Goldziher:
Go for it

dkress59:
sorry guys I'm in the middle of moving 📋 🚚 📦 📦

Goldziher:
So, I reverted the pin - it didn't resole anything. We have an actual bug here I think, and we need an update to the CSS etc.

peterschutt:
Hmm, ok. Definitely resolved it locally with `mkdocs build` but I don't know enough about mkdocs and the custom css to know if there was a difference between my local environment and the production one.

Goldziher:
Well, maybe Im wrong. I repinned it and am deploying. Lemme know if its resolved -- i am myself color blind and cannot be trusted :)

peterschutt:
Version pinned in `.github/workflows/docs.yml` as well as in `pyproject.toml`.

dkress59:
> @dkress59 any update on this? We currently have to pin the mkdocs versions.

For me it's working now – I'll open a PR, please verify on your end @Goldziher

Goldziher:
Fixed! thanks @dkress59

PR #206, opened 2022-07-01
Fix the color scheme for the documentation

dkress59:
#131

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -14,5 +14,5 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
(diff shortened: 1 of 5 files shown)


Goldziher:
Ok, but the mkdocs versions are still pinned. 

Please unpinn them in `pyproject.toml` and `.github/workflows/docs.yaml`. the issue was with the latest version of mkdocs material.

peterschutt (approved):
Toggling between main and this branch locally shows the issue appear on main and go away on this branch - so LGTM.

I pushed in an update to `poetry.lock` as mkdocs material was still locked to 8.2.

Appreciate the docs addition too, cheers.

PR #207, opened 2022-07-01
Add docs addressing known limitation (404, 405) about middleware on exceptions

vrslev:
https://github.com/starlite-api/starlite/issues/175#issuecomment-1172601182


Goldziher:
I will make this update, because some fixes in the docs are also required. Thanks though!

Issue #203, opened 2022-06-30
Refactor: Reduce Cognitive Complexity


Goldziher:
Sonar complains about several functions having too high "cognitive complexity", see --> https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&id=starlite-api_starlite. It would be nice to refactor these to be both clearer and simpler - probably by splitting them into functions, or refactoring them into classes, using functional composition etc. 

This is a good issue for those who like to contribute - because it will accept multiple small PRs. If you work on this, please make sure to have unit tests in place to ensure implementation is not broken. Also please ensure that coverage remains at 100% for the affected code. 

<a href="https://polar.sh/litestar-org/litestar/issues/203">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/203/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/203/pledge.svg">
</picture>
</a>

provinzkraut:
> Fixed in #501

Not really. That was one PR that addressed this, but as you can see from the top comment and the provided link therewithin, there are still 19 open reports from SonarCloud regarding this.

Issue #191, opened 2022-06-28
Code quality: more functional approach


vrslev:
I'm excited about functional programming. I think it makes code more clean and self-explanatory, especially when it is split into small functions. Starlite has code that is much better and clean that you can see in other ambitious projects. Still, it has a lot of quite big functions. They are hard to test in isolation and potentially can lead to bugs.
What do you think about taking more of FP approach? Writing more pure functions, avoiding object oriented design when possible, keep classes small and easy to understand.
Since Starlite is thoroughly tested, we can be confident that nothing important will break on refactor. Of course, since it is my initiative, I will take part in the refactoring process.

Goldziher:
Go ahead. Please go small so I can review PRs granularly. Also, testing is good, but it should be increased if you want to do unit testing.

vrslev:
I’m struggling with big project that should be done first. I added it in my action list and will make first steps in couple of weeks or months (programming is currently just a hobby for me, so I do it in my spare time)

Goldziher:
>  I’m struggling with big project that should be done first. I added it in my action list and will make first steps in couple of weeks or months (programming is currently just a hobby for me, so I do it in my spare time)

Ok. So I will close this issue for now. Reopen if when you want to do the work.

PR #208, opened 2022-07-01
Refactor `SignatureModel.parse_values_from_connection_kwargs`

vrslev:
This is a draft attempt to start refactoring process (#203, #191). 
After reading through the code base, I decided to start with something small—function signature.

I have couple of questions about continuing the work:
- If there's no unit tests, should I add them? Also, there's a lot of—what can be considered—integration tests. If I add an isolated test that duplicates what already has been tested in an integration test, should I remove previous one?
- Is `SignatureModel` a public API?
- Do you see the benifit of proposed changes?


Goldziher:
I won't review this now, but -

1. Yes, add full unit tests to ensure code is not broken

2. Don't remove any existing test unless there is  compelling reason to do so. 

3. SignatureModel is completely internal. As a rule if thumb, anything not in the docs is internal.

starlite/signature.py
```diff
@@ -90,32 +100,26 @@ def construct_exception(
...
+
         for error in exc.errors():
-            if error["loc"][-1] in cls.factory.dependency_name_set:
-                server_errors.append(error)
-            else:
-                client_errors.append(error)
-        method = connection.method if isinstance(connection, Request) else "websocket"
+            is_server_error = error["loc"][-1] in cls.factory.dependency_name_set
+            lst = server_errors if is_server_error else client_errors
+            lst.append(error)
```
peterschutt, line 117: I find this somewhat harder to parse than the original.

Totally agree that `error["loc"][-1] in cls.factory.dependency_name_set` should be abstracted away.

Maybe we should have a utility staticmethod `is_server_error()` so then the block would become:

```python
if self.is_server_error(error):
    server_errors.append(error)
...
```
Not a big deal either way, but we don't have to avoid simple `if`/`else` statements, do we?

starlite/signature.py
```diff
@@ -31,9 +31,31 @@
...
 from starlite.utils.typing import detect_optional_union
 
+if TYPE_CHECKING:
+    from pydantic.error_wrappers import ErrorDict
+
+
 UNDEFINED_SENTINELS = {Undefined, Signature.empty}
 
 
+def get_value_using_plugin_mapping(
```
peterschutt, line 41: Perhaps this could be a `@staticmethod` on `SignatureModel` if it is not used anywhere else?

starlite/kwargs.py
```diff
@@ -307,16 +308,14 @@ async def get_request_data(self, request: Request) -> Any:
             return parse_form_data(media_type=media_type, form_data=form_data, field=model_field)
         return await request.json()
 
-    async def resolve_dependency(
-        self, dependency: Dependency, connection: Union[WebSocket, Request], **kwargs: Any
-    ) -> Any:
+    async def resolve_dependency(self, dependency: Dependency, method: str, url: URL, **kwargs: Any) -> Any:
```
peterschutt, line 311: Can you talk about why it is better to pass the method and url around instead of the connection? I have not problems with it but just curious.

starlite/signature.py
```diff
@@ -45,43 +67,31 @@ class Config(BaseConfig):
...
+        except ValidationError as exc:
+            raise cls.construct_exception(method=method, url=url, exc=exc) from exc
+
+    def resolve_field_value(self, key: str) -> Any:
+        """
+        Return value for a given field key using plugin mapping if available.
+        """
+        value = self.__getattribute__(key)  # pylint: disable=unnecessary-dunder-call
+        mapping = self.field_plugin_mappings.get(key)
+        return get_value_using_plugin_mapping(mapping, value) if mapping else value
```
peterschutt, line 90: much nicer

peterschutt (commented):
All good from me, a few discussion points but nothing big. `SignatureModel.parse_values_from_connection_kwargs()` is a nice improvement.

Was this method in the sonar report?

starlite/signature.py
vrslev: It depends on what philosophy we want to use regarding namespaces. `signature.py` has 3 global classes and 1 function. It would be easier to parse and understand a little standalone function that is not a part of a class, but harder to get that it helps with this and only this class—so it breaks “structurality” of the code.

Let’s get back to our case. There are three viable solutions:
- keep the function the way I refactored it (with the caveat from above),
- make it a method of `PlugginMapping`,
- split `signature.py` file up into `signature_model.py` and `signature_model_factory.py`.

starlite/signature.py
vrslev: You’re right. I will change it.

starlite/signature.py
peterschutt: What ever you think is the best one.

starlite/kwargs.py
vrslev: - It is an abstraction. Soon, I will add unit tests that will test every little detail and case. If the function takes both `Request` or `Websocket`, I’d have to double all tests against this method.
- Separation of concerns: why `KwargsModel` has to know about connection classes?
- Separation of concerns, again.  When we pass `Request` or `Websocket` to `KwargsModel`, it does nothing with it but extract necessary data. If we give the method just simple _data_ values—method and URL—mental load goes down when you read the code.

vrslev:
> Was this method in the sonar report?

@peterschutt Just checked it up: no.

starlite/kwargs.py
peterschutt: >  I’d have to double all tests against this method.

Something easily done with pytest parametrization.

I see all your points and we are getting nitty gritty here, so no wrong answers. I see two parameters to every function instead of one, and lost context (i.e., there's nothing we cannot do with the existing `connection: ...` arg, but plenty we cannot do without it - but hypothetical of course. 

We are in the weeds here really, so I don't care overall.. but I think it borders on changes for the sake of changes. If this was new code I wouldn't bat an eyelid.

peterschutt:
| @peterschutt Just checked it up: no.

All good, I went looking to see what the complexity rating was and couldn't see it.

Don't worry about breaking off one of the bigger changes if you are worried about it. The ones in the report are the low hanging fruit, so best value for our time!

starlite/signature.py
vrslev: Done.

starlite/signature.py
vrslev: I decided to go with new method in `PluginMapping`.

starlite/kwargs.py
vrslev: > Something easily done with pytest parametrization.

Of course it is. I don't debate that it is. It just can feel like a burden to maintain and adds cognitive load.

I reverted this change and added `SignatureModel.get_connection_method_and_url` static method. Now I'm not sure it belongs in `SignatureModel` 🤔

starlite/signature.py
peterschutt: Good call!

starlite/kwargs.py
peterschutt: > Now I'm not sure it belongs in SignatureModel

Haha, well maybe not - make the call if you think it needs to move - but don't let perfect be the enemy of good enough:)

For the record, I was going to be happy if you left it, or changed it back - just good to talk about these things.

Good stuff.

vrslev: > For the record, I was going to be happy if you left it, or changed it back - just good to talk about these things.

I understand. Changed my mind—that's all. The goal of this PR is to reduce code complexity, not increase it 😂

starlite/signature.py
```diff
@@ -97,25 +92,32 @@ def construct_exception(
...
+
+        return InternalServerException(detail=f"A dependency failed validation for {method} {url}", extra=server_errors)
 
-        return InternalServerException(
-            detail=f"A dependency failed validation for {method} {connection.url}",
-            extra=server_errors,
-        )
+    @classmethod
+    def is_server_error(cls, error: "ErrorDict") -> bool:
+        """Check whether validation error is server error"""
```
Goldziher: "is a..."

starlite/plugins/base.py
```diff
@@ -54,3 +54,20 @@ def get_plugin_for_value(value: Any, plugins: List[PluginProtocol]) -> Optional[
...
+        Given some generated value by plugin,
+        return instance of original class.
+
+        Also excepts can accept list or tuple of values.
+        """
+
+        def get_instance(value: BaseModel) -> Any:
+            return self.plugin.from_pydantic_model_instance(self.model_class, pydantic_model_instance=value)
+
+        if isinstance(value, (list, tuple)):
```
Goldziher, line 71: Not sure if we have a util already, you can create a util to check if value is a non string iterable. 

Optional.

Goldziher (approved):
Nice work

starlite/signature.py
```diff
@@ -97,25 +92,32 @@ def construct_exception(
...
+        return error["loc"][-1] in cls.factory.dependency_name_set
+
+    @staticmethod
+    def get_connection_method_and_url(connection: Union[Request, WebSocket]) -> Tuple[str, URL]:
+        """Extract method and URL from Request or WebSocket"""
+        method = "websocket" if isinstance(connection, WebSocket) else connection.method
```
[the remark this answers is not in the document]
vrslev: > I don't like this part - somewhat hacky.

Yep. You have a better solution?

> Please instead of websocket use the enum member ScopeType.WebSocket

Done.

starlite/plugins/base.py
```diff
@@ -54,3 +54,20 @@ def get_plugin_for_value(value: Any, plugins: List[PluginProtocol]) -> Optional[
...
     model_class: Any
+
+    def get_value_converted_to_model_class(
+        self, value: Union[BaseModel, List[BaseModel], Tuple[BaseModel, ...]]
+    ) -> Any:
+        """
+        Given some generated value by plugin,
+        return instance of original class.
+
+        Also excepts can accept list or tuple of values.
```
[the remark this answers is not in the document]
vrslev: Done. "Also can accept list or tuple of values."

starlite/plugins/base.py
```diff
@@ -54,3 +54,20 @@ def get_plugin_for_value(value: Any, plugins: List[PluginProtocol]) -> Optional[
 class PluginMapping(NamedTuple):
     plugin: PluginProtocol
     model_class: Any
+
+    def get_value_converted_to_model_class(
```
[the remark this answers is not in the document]
vrslev: > Name: you're converting a value to ?

Arbitrary value is converted to `PluginMapping.model_class`

> Also, typing is Any.

I can make `PluginMapping` a generic with `PluginProtocol`'s `ModelT` and add several overloads, but prefer to keep it simple since the function is used only with dynamic values.

starlite/signature.py
```diff
@@ -55,29 +60,19 @@ def parse_values_from_connection_kwargs(
...
-            raise cls.construct_exception(connection, e) from e
-        return output
+            signature = cls(**kwargs)
+            return {key: signature.resolve_field_value(key) for key in cls.__fields__}
+        except ValidationError as exc:
+            raise cls.construct_exception(connection, exc) from exc
+
+    def resolve_field_value(self, key: str) -> Any:
+        """
+        Return value for a given field key using plugin mapping if available.
```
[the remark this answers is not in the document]
vrslev: Done.

starlite/signature.py
vrslev: Done.

starlite/plugins/base.py
vrslev: I'm not sure it is possible from typing perspective. This won't work with if block:

```python
@overload
def is_nonstring_iterable(value: list[T]) -> TypeGuard[list[T]]:
    ...


@overload
def is_nonstring_iterable(value: tuple[T, ...]) -> TypeGuard[tuple[T, ...]]:
    ...


@overload
def is_nonstring_iterable(value: Any) -> Any:
    ...


def is_nonstring_iterable(value: Any | list[T] | tuple[T, ...]) -> Any | TypeGuard[list[T]] | TypeGuard[tuple[T, ...]]:
    return isinstance(value, (list, tuple))

```

starlite/plugins/base.py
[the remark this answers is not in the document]
Goldziher: Simple is fine. My point though is that it's a model instance not model class, and the name should reflect it.

starlite/plugins/base.py
Goldziher: Use the `Sequence` type?

starlite/plugins/base.py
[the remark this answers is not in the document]
vrslev: Will `get_value_converted_to_model` do?

starlite/plugins/base.py
vrslev: `str` is `Sequence[str]` and the function is named `is_nonstring_iterable` 🤔

Goldziher: Nvm, not worth to waste time on.

vrslev: I agree.

starlite/plugins/base.py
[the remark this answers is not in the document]
Goldziher: `value_to_model_instsnce`?

peterschutt (approved):

peterschutt:
Thanks @vrslev!

PR #244, opened 2022-07-04
Extract `HTTPRouteHandler.to_response`'s ifs battery and after_request hook into two different methods

185504a9:
simple refactor which was mentioned as needed by #203 

i've run sonar locally and it doesn't scream anymore on that function

i've extracted the first and outermost `if` since it had the biggest `ifs` battery but i don't think the second one -data that is not starlite-like- should be also extracted, but i will  if you think i should


peterschutt (approved):

peterschutt:
Thanks!

@all-contributors add @185504a9 for code

vrslev:
What about using [`functools.singledispatchmethod`](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod)?

Issue #151, opened 2022-06-17
Enhancement: Add a CLI


Goldziher:
We should consider adding a CLI that will allow users to quickly scaffold a project with a default folder / file structure, readme, tooling etc. 

A good example is the NestJS CLI. 

I'd like to discuss this proposal in this issue.

peterschutt:
`poetry new` is the first example that stands out to me. They have [cleo](https://github.com/python-poetry/cleo.git).

vrslev:
> We should consider adding a CLI that will allow users to quickly scaffold a project with a default folder / file structure, readme, tooling etc.

I think it is a great idea in context of Starlite project.

> A good example is the NestJS CLI.

I really like the one that Vue 3 has: https://vuejs.org/guide/quick-start.html#local. Very powerful yet approachable.

Goldziher:
Ok. So, let's talk about structure-

Given that you probably want to install a CLI as as a global executable outside your project, I think it might be a good idea to separate it out into its own package and publish it separately. I know sometimes the CLi is a part of the app, but I think here it's going to be an issue - the CLI should be independent from the starlite version IMO. 

We should consider whether we want to split it to a different repo, or simply a folder in this repo. 

Pros of different repo:

1. Separation of concerns
2. No shared dev dependencies
3. Simpler release process
4. No need for a monorepo or similar construct

Cons:

1. A separate code base might make it harder to contribute, e.g. issues etc. 
2. Stars are spread across repos.  
3. No code sharing is possible (no biggy on this case)

vrslev:
Is there anything similar to [`npm init`](https://docs.npmjs.com/cli/v8/commands/npm-init#examples) in Python world? Maybe just [`cookiecutter`](https://github.com/cookiecutter/cookiecutter)?

JeromeK13:
What about using cookiecutter together with clap?
So you could create a separate repo for the template but still integrate the cli into the "main" repo while using cookiecutter as a base.

Goldziher:
I'm open to this idea. I have no experience working with these tools tbh.

JeromeK13:
Alright I could do a proposal. In general do you care about dependencies in this project?
For example is it a problem for using Clap instead of the default arg-parser?

peterschutt:
Would it also mean `cookiecutter` was a dependency of the standard install?

Goldziher:
It will be an `extra` dependency I assume, so it shouldnt be a problem.

[...]

PR #251, opened 2022-07-05
Support for Starlite-CLI

JeromeK13:
# Starlite-CLI
(#151)
The Starlite-CLI should give developers a smooth starter experience with different templates to chose from.

## Current Features
* Creation of predefined templates
* Help for every command
* Version display
* Auto completion for commands (not fully functional, could be deactivated) 

## Thoughts behind current implementation
As a base I used [typer](https://typer.tiangolo.com/). Its a nice wrapper around [click](https://click.palletsprojects.com/en/8.1.x/) to reduce boilerplate. For the Project generation I've used [cookiecutter](https://cookiecutter.readthedocs.io/en/stable/)

## Project implementation
With the poetry-script we can directly call the cli like this when starlite is installed.
```starlite-cli -h                                                                
Usage: starlite-cli [OPTIONS] COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]...

Options:
  -v, --version                   Get current CLI version
  --install-completion [bash|zsh|fish|powershell|pwsh]
                                  Install completion for the specified shell.
  --show-completion [bash|zsh|fish|powershell|pwsh]
                                  Show completion for the specified shell, to
                                  copy it or customize the installation.
  -h, --help                      Show this message and exit.

Commands:
  create  Generates project from cookiecutter template
```

### Poetry script
```toml
[tool.poetry.scripts]
starlite-cli = "starlite.cli:cli"
```

## TODO

- [ ] Add menu selection for presets - maybe find a better mapping than Enum
- [ ] Maybe wrap some uvicorn commands for running dev (hotreload) and prod

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,8 @@ pyyaml = "*"
 starlette = "*"
 typing-extensions = "*"
 requests = { version = "*", optional = true }
+typer = "^0.4.2"
+cookiecutter = "^2.1.1"
 
 [tool.poetry.dev-dependencies]
 hypothesis = { extras = ["cli"], version = "*" }
@@ -110,3 +112,6 @@ omit = ["*/tests/*"]
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+
+[tool.poetry.scripts]
+starlite-cli = "starlite.cli:cli"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,6 @@ typing-extensions = "*"
 requests = { version = "*", optional = true }
 typer = "^0.4.2"
 cookiecutter = "^2.1.1"
-pick = "^1.2.0"
 
 [tool.poetry.dev-dependencies]
 hypothesis = { extras = ["cli"], version = "*" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ typing-extensions = "*"
 requests = { version = "*", optional = true }
 typer = "^0.4.2"
 cookiecutter = "^2.1.1"
+pick = "^1.2.0"
 
 [tool.poetry.dev-dependencies]
 hypothesis = { extras = ["cli"], version = "*" }
--- a/poetry.lock
+++ b/poetry.lock
@@ -17,6 +17,18 @@ test = ["coverage[toml] (>=4.5)", "hypothesis (>=4.0)", "pytest (>=7.0)", "pytes
 trio = ["trio (>=0.16)"]
 
 [[package]]
+name = "arrow"
+version = "1.2.2"
+description = "Better dates & times for Python"
+category = "main"
+optional = false
+python-versions = ">=3.6"
+
+[package.dependencies]
+python-dateutil = ">=2.7.0"
+typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
+
+[[package]]
 name = "atomicwrites"
 version = "1.4.0"
 description = "Atomic file writes."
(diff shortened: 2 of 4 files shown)


JeromeK13:
https://github.com/JeromeK13/starlite-minimal-starter
This is my early example for a template minimal starter repo. In general every cookiecutter template can be supported by adding it to the Enum

```python
class ProjectTemplates(str, Enum):
    minimal = "https://github.com/JeromeK13/starlite-minimal-starter.git"
```

Goldziher:
Thanks for this. 

I like the idea of supporting multiple templates, that's good. 

Ieft a couple of comments.

vrslev:
**Suggestions for [the template](https://github.com/JeromeK13/starlite-minimal-starter/blob/main/%7B%7Bcookiecutter.app_name%7D%7D/pyproject.toml):**

- Pin Starlite's dependencies,
- Add testing dependencies to development requirements,
- Change default description from `Light-weight and flexible ASGI API Framework 💫`,
- Get default author name and mail from git config (I saw project generators doing that),
- Change code so `OpenAPIConfig` is imported in `main.py`,
- Remove unrelated things from `.gitignore` (my preference is to keep is simple),
- Remove top `__init__.py`: it's not a part of the package.

Also that would be pleasant to see the generated code nicely formatted with—de facto standard—`black`.

**Concerning the CLI itself:**
- Since it has only one feature, maybe simplify it a little bit and don't use commands? I mean, calling `starlite-cli` is enough, you don't need `starlite-cli create`.
- Another thought: change the name from abstract `starlite-cli` to just `create-starlite-app`. Is is straightforward and self-explanatory ([`create-react-app`](https://github.com/facebook/create-react-app)).
- A CLI with one command doesn't need shell completion. And I suspect, the vast majority of the people using it will do it only once.
- `-o`, `--output-dir PATH` option should be the first and only parameter _(not a flag)_.

vrslev:
> Maybe wrap some uvicorn commands for running dev (hotreload) and prod

What do you think about using concept of scripts like Node has in `package.json`? You'd put `dev` and `run` to `[tool.poetry.scripts]` and use it with `poetry run`

pyproject.toml
```diff
@@ -42,6 +42,8 @@ pyyaml = "*"
 starlette = "*"
 typing-extensions = "*"
 requests = { version = "*", optional = true }
+typer = "^0.4.2"
+cookiecutter = "^2.1.1"
```
peterschutt, line 46: I believe we were going to make these optional? Also do they need to be pinned so aggressively, or at all?

pyproject.toml
```diff
@@ -110,3 +112,6 @@ omit = ["*/tests/*"]
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+
+[tool.poetry.scripts]
+starlite-cli = "starlite.cli:cli"
```
peterschutt, line 117: Can you run pre-commit over all this? Instructions in `CONTRIBUTING.md`.

peterschutt (commented):
A couple of points, also when you choose to remove draft status it would be good to squash all the commits down.

Cheers!

peterschutt:
> What do you think about using concept of scripts like Node has in package.json? You'd put dev and run to [tool.poetry.scripts] and use it with poetry run

Makes sense to use these IMO

> Since it has only one feature, maybe simplify it a little bit and don't use commands? I mean, calling starlite-cli is enough, you don't need starlite-cli create.
Another thought: change the name from abstract starlite-cli to just create-starlite-app. Is is straightforward and self-explanatory ([create-react-app](https://github.com/facebook/create-react-app)).

This would limit any future additions to the cli though. Perhaps we could use something like [click-default-group](https://github.com/click-contrib/click-default-group) so that creating apps from template is a default command? If we are sure that we'll never use the cli for anything other than this use case, then I'd be supportive of `create-starlite-app`.

> Pin Starlite's dependencies

Do you mean include a `poetry.lock`? Do we really even want to pin the `starlite` version in the app? My instinct is to make it `starlite = "*"` so that users get the latest version of starlite and deps when they run `poetry install ` after generating the app. Also then it means we don't have to update the templates every time we make a release.

> A CLI with one command doesn't need shell completion. And I suspect, the vast majority of the people using it will do it only once.

Agree here, especially given OPs original comment:

> Auto completion for commands (not fully functional, could be deactivated)

We can always add stuff in later.

vrslev:
> Do you mean include a poetry.lock? Do we really even want to pin the starlite version in the app? My instinct is to make it starlite = "*" so that users get the latest version of starlite and deps when they run poetry install after generating the app. Also then it means we don't have to update the templates every time we make a release.

I like to strictly pin top dependencies in `pyproject.toml` and make Dependabot send updates. This way I can determine when an update would break existing functionality. There’s no robust way to make poetry update only a dependency that one of top dependencies depend on.

pyproject.toml
JeromeK13: Yes sure, forgot that sorry - will be done when I develop further

JeromeK13:
Hey Guys

I really like the Ideas. If we really go the way with "only" supporting the creation of the commands, then I would go with only poetry commands like @vrslev mentioned. The we would spare us the hustle of maintaining this wrapper cli. Cookiecutter could then be optional to the once who want that.

Goldziher:
any update?

JeromeK13:
Hey
Sorry was not able to work on that due to vacations. But I'm currently also not very satisfied with the solution. I think the best way to reduce work / be the most efficient is to create a good cookiecutter template and then use it with the default cookiecutter package

Open to do a CLI implementation but currently dont find the time to create a good template

Issue #132, opened 2022-06-07
Documentation: Add API documentation using Doctstring Annotations. 


peterschutt:
I really miss having API docs generally available.. this isn't just a starlite issue, but in general I notice that developer docs are more tailored to the beginner by being mostly prose and metaphor.

I actually agree with that though, ease of uptake is important, but that doesn't mean api docs can't be buried somewhere via some tiny obscure link at the bottom of the page somewhere.

https://github.com/mkdocstrings/mkdocstrings looks like it might be a good plugin to use.

Goldziher:
Go ahead, I just got exhausted from writing docs 😀

peterschutt:
Haha, I can totally understand that, and I hope my comment didn't come across as diminishing in any way to the work you've done!

@Goldziher do you have a preference for docstring style? I prefer the numpy style myself.

@dkress59 I'm only asking your opinion here as I see you were tagged into the other docs issue, so assume you are across the mkdocs setup? I've only used readthedocs and sphinx in the past. Do you have any opinion on this, or happy to let me run with it?

Goldziher:
Not really, as long as the comment style is consistent I'm good.

peterschutt:
Great, thanks.

dkress59:
I'm not exactly sure what you question is but I'll try to answer 😄 
I've only ever used mkdocs and docsify, so I have no opinion of  sphinx or readthedocs.
I'm mostly responsible for the appearance of the starlite docs and I've always wanted to go over it once more to improve readability/ease-of-use: making paragraphs easier to skim by making keywords bold, introducing more paragraphs, splitting long pages into shorter sub-pages, plus maybe some re-arranging.
If you're asking to contribute to the docs – please go ahead! @Goldziher and me would both be happy to review your PR.

peterschutt:
Thanks for your reply @dkress59.

This issue is specifically about adding api reference docs. I linked a tool that I would probably try to use if I was to undertake the task, and I was trying to see if you had some opinion as @Goldziher linked you into the other docs issue that I raised.

Sorry if I was unclear.

dkress59:
yeah that mcdocstrings plugin looks awesome!

peterschutt:
Yeah, out of all the plugins listed for autogen api docs in the mkdocs wiki, it seems to be the most supported/active.

There's another issue I'm working on that will have a bit of a docs component to it. I'll hopefully find my way alright but if I need any assistance would you mind if I ping you?

[...]

PR #252, opened 2022-07-05
Enhancement: Brotli compression middleware

cofin:
Hey all, spent some time with the framework over the past few days, and things look great!  

On other Starlette projects, I have used a custom Brotli compression middleware instead of GZIP, and I wanted to propose adding this middleware as an optional installation component.

I've included the base changes required to implement this enhancement, and it would be great to get some feedback and potentially get this integrated as a new feature.

There are two things to note in this branch:
- I renamed the Starlite `gzip_config` to `compression_config` as it now takes a `BrotliConfig` or `GzipConfig`.  The correct backend will load based on what config type is passed.
- I have created a `contrib` folder for future "optional extras".  I considered adding this into the `plugins` folder, but decided against that since it exactly a plugin


docs/usage/7-middleware.md
```diff
@@ -148,17 +148,37 @@ You can enable gzip compression of responses by passing an instance of `starlite
...
+
+You can enable brotli compression of responses by passing an instance of `starlite.config.BrotliConfig`:
+
+```python
+from starlite import Starlite, BrotliConfig
+
+app = Starlite(request_handlers=[...], compression_config=BrotliConfig())
+```
+
+You can configure two values:
```
Goldziher, line 173: I see 4 values below 😉

pyproject.toml
```diff
@@ -42,6 +42,7 @@ pyyaml = "*"
 starlette = "*"
 typing-extensions = "*"
 requests = { version = "*", optional = true }
+Brotli = { version = "*", optional = true }
```
Goldziher, line 45: No need to capitalize the dependency name, pypi is case insensitive. Please keep consistent lower case.

starlite/contrib/brotli.py
```diff
@@ -0,0 +1,136 @@
...
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+from starlette.middleware.gzip import GZipResponder, unattached_send
+
+try:
+    import brotli
+except ImportError as exc:  # pragma: no cover
+    raise MissingDependencyException("brotli is not installed") from exc
+
+
+class BrotliMode:
```
Goldziher, line 13: How about using an `enum.Enum` here?

starlite/contrib/brotli.py
```diff
@@ -0,0 +1,136 @@
...
+    """Brotli available modes."""
+
+    generic = brotli.MODE_GENERIC
```
Goldziher, line 16: By convention enumerations should be SCREAMING_SNAKE_CASE

starlite/contrib/brotli.py
```diff
@@ -0,0 +1,136 @@
...
+    text = brotli.MODE_TEXT
+    font = brotli.MODE_FONT
+
+
+class BrotliMiddleware:
+    def __init__(
+        self,
+        app: ASGIApp,
+        quality: int = 4,
+        mode: str = "text",
```
Goldziher, line 26: Should use an enum value everywhere

starlite/contrib/brotli.py
```diff
@@ -0,0 +1,136 @@
...
+        self.app = app
+        self.quality = quality
+        self.mode = getattr(BrotliMode, mode)
+        self.minimum_size = minimum_size
+        self.lgwin = lgwin
+        self.lgblock = lgblock
+        self.gzip_fallback = gzip_fallback
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] == "http":
```
Goldziher, line 41: Use the enum `ScopeType.HTTP`

starlite/plugins/sql_alchemy.py
```diff
@@ -147,7 +147,7 @@ def providers_map(self) -> Dict[Type[TypeEngine], Callable[[Union[TypeEngine, Ty
             mssql.DATETIMEOFFSET: lambda x: datetime,
             mssql.IMAGE: self.handle_string_type,
             mssql.MONEY: lambda x: Decimal,
-            mssql.NTEXT: self.handle_string_type,
+            mssql.NEXT: self.handle_string_type,
```
Goldziher, line 150: A bit off the reservation here. Have you absolutely verified this?

tests/contrib/test_brotli.py
```diff
@@ -0,0 +1,34 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@get(path="/")
+def handler() -> None:
+    ...
+
+
+def test_brotli_middleware() -> None:
```
Goldziher, line 19: So, for the amount of code you are adding this test is really not sufficient. I would expect a more or less full suite of unit tests + some integration tests to ensure this is working properly.

Goldziher (changes requested):
So, first off - thanks a lot for the contribution. This is high quality work and it's appreciated. 

I left several comments that should be addressed. 

A few other points that are important - 

1. I would like to consider a way to wrap both brotli and gzip in a single interface. Perhaps the pattern you have there is the simplest, but perhaps we can use a single compression config for both and have some generalization. 

2. I don't like the branching logic when setting the app. I would love to see some splitting there. 

3. I'm not sure about contrib. What's the argument against putting this under the Middleware namespace?

starlite/plugins/sql_alchemy.py
cofin: I'll respond to this one first.  It is certainly off topic, and wasn't supposed to be included.  I'll remove this edit from the PR.

tests/contrib/test_brotli.py
cofin: No issue.  I agree here.  I'll add additional test cases to properly check for compression headers and validate sizing.

starlite/app.py
```diff
@@ -138,11 +137,13 @@ def create_asgi_handler(self) -> ASGIApp:
         """
         Creates an ASGIApp that wraps the ASGI router inside an exception handler.
 
-        If CORS or TruseedHost configs are provided to the constructor, they will wrap the router as well.
+        If CORS or TrustedHost configs are provided to the constructor, they will wrap the router as well.
         """
         asgi_handler: ASGIApp = self.asgi_router
-        if self.gzip_config:
-            asgi_handler = GZipMiddleware(app=asgi_handler, **self.gzip_config.dict())
+        if self.compression_config and self.compression_config.backend:
```
Goldziher, line 143: Why would the `and` be required here? If it's. Pydantic model you should already have validation.

starlite/config.py
```diff
@@ -33,9 +34,47 @@ class CORSConfig(BaseModel):
...
+    backend: Optional[CompressionBackend] = None
     minimum_size: int = 500
-    compresslevel: int = 9
+    gzip_compress_level: int = 9
+    brotli_quality: int = 5
+    brotli_mode: BrotliMode = BrotliMode.TEXT
+    brotli_lgwin: int = 22
+    brotli_lgblock: int = 0
+    brotli_gzip_fallback: bool = True
+    CompressionMiddleware: PyObject = None
```
Goldziher, line 57: So this part and the init method I'd dump. I would instead use a `property` that has this logic OR I would move this logic into the consumer.

starlite/config.py
```diff
@@ -33,9 +34,47 @@ class CORSConfig(BaseModel):
...
+
+    def __init__(__pydantic_self__, **data: Any) -> None:
+        backend = data.get("backend", None)
```
Goldziher, line 60: No need for the None default, because it is the default.

starlite/config.py
```diff
@@ -33,9 +34,47 @@ class CORSConfig(BaseModel):
...
+        if backend == CompressionBackend.GZIP:
+            data["CompressionMiddleware"] = "starlite.middleware.GZipMiddleware"
+        elif backend == CompressionBackend.BROTLI:
+            data["CompressionMiddleware"] = "starlite.middleware.brotli.BrotliMiddleware"
+        super().__init__(**data)
+
+    def dict(self) -> Dict[str, Any]:
```
Goldziher, line 67: Clever

starlite/middleware/__init__.py
```diff
@@ -0,0 +1,6 @@
+from starlite.middleware.exceptions import ExceptionHandlerMiddleware
+from starlite.middleware.authentication import AuthenticationResult, AbstractAuthenticationMiddleware
+from starlite.middleware.gzip import GZipMiddleware
+
+
+__all__ = ["ExceptionHandlerMiddleware", "AuthenticationResult", "AbstractAuthenticationMiddleware", "GZipMiddleware"]
```
Goldziher, line 6: Not sure we need gzip here, since our usage was not meant for user consumption

Goldziher (commented):
Very good progress. Please finalize the PR and make sure documentation is up to date.

starlite/__init__.py
```diff
@@ -135,7 +135,7 @@ def __getattr__(name: str) -> Any:
     import warnings
 
     warnings.warn(
-        f"Importing {name} from {__package__} is deprecated, use `from startlite.testing import {name}` instead",
```
peterschutt, line 138: Good catch

starlite/config.py
cofin: Removed the `None` after setting backend to required.

starlite/middleware/__init__.py
cofin: I agree after sleeping on it.  I'm not sure why I thought it was a good idea to include it there last night.

peterschutt:
@cofin my only comment here beyond what @Goldziher has raised is doc strings on your contributions - we have #132 lined up to go over all the doc strings in the code base so it would be good for new work to have all that done upfront. Numpy style would be good.

peterschutt:
Would you mind squashing down all the commits once this is ready to merge @cofin?

cofin:
Sure thing.  I'll also be sure to take care of the docstrings.

starlite/app.py
```diff
@@ -50,6 +49,7 @@
 
 DEFAULT_OPENAPI_CONFIG = OpenAPIConfig(title="Starlite API", version="1.0.0")
 DEFAULT_CACHE_CONFIG = CacheConfig()
+DEFAULT_COMPRESSION_CONFIG = None
```
Goldziher, line 52: this is redundant - no need to have a variable here.

starlite/middleware/compression/brotli.py
```diff
@@ -0,0 +1,147 @@
...
+                    app=self.app,
+                    minimum_size=self.minimum_size,
+                    quality=self.quality,
+                    mode=self.mode,
+                    lgwin=self.lgwin,
+                    lgblock=self.lgblock,
+                )
+                await responder(scope, receive, send)
+                return
+            elif self.gzip_fallback and "gzip" in headers.get("Accept-Encoding", ""):
```
Goldziher, line 51: no need for elif since you return above.
```suggestion
            if self.gzip_fallback and "gzip" in headers.get("Accept-Encoding", ""):
```

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
...
+    last name in the path. Raise ImportError if the import failed.
+
+    Based on Django's module_loading code
+
+    """
+
+    def cached_import(module_path, class_name):
+        # Check whether module is loaded and fully initialized.
+        if not (
+            (module := sys.modules.get(module_path))
```
Goldziher, line 17: walrus operator is not supported in python 3.7.

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
+import sys
+from importlib import import_module
+
+
+def import_string(dotted_path):
+    """
+    Import a dotted module path and return the attribute/class designated by the
+    last name in the path. Raise ImportError if the import failed.
+
+    Based on Django's module_loading code
```
Goldziher, line 10:
```suggestion
```

Goldziher: unless you literally ported the code from django, in which case you need a comment that references the original. Otherwise, if its just inspiration, no need for this comment.

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
...
+            (module := sys.modules.get(module_path))
+            and (spec := getattr(module, "__spec__", None))
+            and getattr(spec, "_initializing", False) is False
+        ):
+            module = import_module(module_path)
+        return getattr(module, class_name)
+
+    try:
+        module_path, class_name = dotted_path.rsplit(".", 1)
+    except ValueError as err:
```
Goldziher, line 26:
```suggestion
    except ValueError as e:
```

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
...
+        raise ImportError("%s doesn't look like a module path" % dotted_path) from err
```
Goldziher, line 27:
```suggestion
        raise ImportError("%s doesn't look like a module path" % dotted_path) from e
```

Also the error message here is not quite clear enough. Please describe the issue in the error message.

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
...
+
+    try:
+        return cached_import(module_path, class_name)
+    except AttributeError as err:
```
Goldziher, line 31:
```suggestion
    except AttributeError as e:
```

Very javascripty. Please stick to `e` or `exc`

starlite/utils/module_loading.py
```diff
@@ -0,0 +1,32 @@
...
+        raise ImportError('Module "%s" does not define a "%s" attribute/class' % (module_path, class_name)) from err
```
Goldziher, line 32:
```suggestion
        raise ImportError('Module "%s" does not define a "%s" attribute/class' % (module_path, class_name)) e err
```

Goldziher: Also please use format strings.

starlite/utils/module_loading.py
Goldziher: Also please use format strings.

starlite/utils/module_loading.py
Goldziher: also, its best to break this into multiple variables and be more descriptive. This code is not quite clear enough.

starlite/middleware/compression/brotli.py
```diff
@@ -0,0 +1,147 @@
...
+        self.mode = mode
+        self.lgwin = lgwin
+        self.lgblock = lgblock
+        self.br_file = brotli.Compressor(quality=self.quality, mode=self.mode, lgwin=self.lgwin, lgblock=self.lgblock)
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        self.send = send
+        await self.app(scope, receive, self.send_with_brotli)
+
+    async def send_with_brotli(self, message: Message) -> None:
```
Goldziher, line 84: this method is too complex, please refactor it by splitting its logic.

starlite/middleware/compression/brotli.py
```diff
@@ -0,0 +1,147 @@
...
+                headers["Content-Length"] = str(len(body))
+                headers.add_vary_header("Accept-Encoding")
+                message["body"] = body
+
+                await self.send(self.initial_message)
+                await self.send(message)
+            else:
+                # Initial body in streaming Brotli response.
+                headers = MutableHeaders(raw=self.initial_message["headers"])
+                headers["Content-Encoding"] = "br"
```
Goldziher, line 113: "br" should be a constants at the top of the file, and also be used in the tests.

starlite/middleware/compression/brotli.py
```diff
@@ -0,0 +1,147 @@
...
+            body = message.get("body", b"")
+            more_body = message.get("more_body", False)
+            if len(body) < self.minimum_size and not more_body:
+                # Don't apply Brotli to small outgoing responses.
+                await self.send(self.initial_message)
+                await self.send(message)
+            elif not more_body:
+                # Standard Brotli response.
+                body = self.br_file.compress(body) + self.br_file.finish()
+                headers = MutableHeaders(raw=self.initial_message["headers"])
```
Goldziher, line 102: there is code redundancy between the different clauses that can be generalized.

Goldziher (commented):
Ok, stuff looks good. Most of my comments are about code style etc., a couple are more serious. 

Please make sure to add doc strings to all functions / methods. And for the components that are supposed to be user facing or could be user facing, make sure to document that API as well (no need to do this for private methods and stuff that is meant to be internal).

starlite/utils/module_loading.py
cofin: It's very close to what Django has.  Once I remove the Walrus operator, it maybe sufficiently different though.

Goldziher (approved):
Great stuff

Goldziher:
@all-contributors add @cofin code docs

PR #258, opened 2022-07-07
Documentation: Integrates `mkdocstrings` into docs config.

peterschutt:
- adds `api.md` to `docs/` with the `Starlite` class included in documentation.
- incomplete class doc string for `Starlite`
- include `mkdocstrings` plugin in `mkdocs.yml`
- adds `mkdocstrings[python]` dependency


peterschutt:
This is just intended as a starting point so I can get more experienced eyes on the config.

Should the plugin be included in the root level `plugins` key in `mkdocs.yml` - I see we have other plugins included in `theme.plugins`.

- if I include the plugin in `theme.plugins` using the docker instructions included in `CONTRIBUTING.md` works , but the plugin doesn't appear to work.
- if I include the plugin in the `plugins` at the root level (as I've committed) trying to run the docs through the docker command fails saying that the plugin isn't installed.
- if I include the plugin in `plugins` at the root level, do I also need to declare the `search` plugin there, or is it enough that it is declared in `theme.plugins`?

pyproject.toml
```diff
@@ -44,19 +44,20 @@ typing-extensions = "*"
 requests = { version = "*", optional = true }
 
 [tool.poetry.dev-dependencies]
+Jinja2 = "*"
```
Goldziher, line 47: Lower case for deps

Goldziher (approved):

Goldziher:
Merge when ready. We should probably mark the API reference as WIP

PR #261, opened 2022-07-07
Issue 187 layered parameters

Goldziher:
This PR adds the ability to declare parameters on all layers of the app.


starlite/kwargs.py
```diff
@@ -114,40 +164,38 @@ def create_for_signature_model(
             for key in dependencies
             if key in signature_model.__fields__
         }
-        expected_path_parameters: Set[ParameterDefinition] = set()
-        expected_header_parameters: Set[ParameterDefinition] = set()
-        expected_cookie_parameters: Set[ParameterDefinition] = set()
-        expected_query_parameters: Set[ParameterDefinition] = set()
 
         ignored_keys = {*RESERVED_KWARGS, *[dependency.key for dependency in expected_dependencies]}
```
peterschutt, line 168: This can be done without having to build the list:

```python
ignored_keys = {*RESERVED_KWARGS, *(dependency.key for dependency in expected_dependencies)}
```

starlite/routes.py
```diff
@@ -107,14 +108,25 @@ def create_handler_kwargs_model(self, route_handler: BaseRouteHandler) -> Kwargs
...
             param_name = param["name"]
             if param_name in path_parameters:
                 raise ImproperlyConfiguredException(f"Duplicate parameter '{param_name}' detected in '{self.path}'.")
             path_parameters.add(param_name)
+
+        layered_parameters: Dict[str, FieldInfo] = {}
+        for layer in route_handler.ownership_layers():
+            if getattr(layer, "parameters", None):
+                for key, field_info in cast(Dict[str, FieldInfo], layer.parameters).items():  # type: ignore
+                    layered_parameters[key] = field_info
```
peterschutt, line 123:
```python
layered_parameters: Dict[str, FieldInfo] = {}
for layer in route_handler.ownership_layers():
    layered_parameters.update(getattr(layer, "parameters", {})
```

peterschutt: The getattr is because the handlers don't have the `parameters` attribute right?

starlite/controller.py
```diff
@@ -48,22 +50,14 @@ class Controller:
     guards: Optional[List["Guard"]]
     middleware: Optional[List[Union["Middleware", "Type[BaseHTTPMiddleware]", "Type[MiddlewareProtocol]"]]]
     owner: "Router"
+    parameters: Optional[Dict[str, Tuple[Any, "FieldInfo"]]]
```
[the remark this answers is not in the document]
Goldziher: Yes

starlite/handlers/base.py
```diff
@@ -156,10 +153,12 @@ def resolve_exception_handlers(self) -> Dict[Union[int, Type[Exception]], Except
...
-            exception_handlers: Dict[Union[int, Type[Exception]], ExceptionHandler] = {}
-            for layer in reversed(list(self.ownership_layers())):
-                exception_handlers = {**exception_handlers, **(layer.exception_handlers or {})}
-            self.resolved_exception_handlers = exception_handlers
+            self.resolved_exception_handlers = {}
+            for layer in self.ownership_layers():
+                self.resolved_exception_handlers = {
+                    **self.resolved_exception_handlers,
+                    **(layer.exception_handlers or {}),
+                }
```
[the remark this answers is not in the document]
Goldziher: It's an assignment to a variable name. I could use update, but I like destructing.

starlite/handlers/http.py
```diff
@@ -211,10 +210,9 @@ def resolve_response_headers(self) -> Dict[str, ResponseHeader]:
         This method is memoized so the computation occurs only once.
         """
         if self.resolved_headers is BaseRouteHandler.empty:
-            headers: Dict[str, ResponseHeader] = {}
-            for layer in reversed(list(self.ownership_layers())):
-                headers = {**headers, **(layer.response_headers or {})}
-            self.resolved_headers = headers
+            self.resolved_headers = {}
+            for layer in self.ownership_layers():
+                self.resolved_headers = {**self.resolved_headers, **(layer.response_headers or {})}
```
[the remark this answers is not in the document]
Goldziher: Aight as well

starlite/kwargs.py
```diff
@@ -48,11 +49,12 @@ def merge_parameter_sets(first: Set[ParameterDefinition], second: Set[ParameterD
 
 class KwargsModel:
     """
-    This class is used to model the required kwargs for a given handler and its dependencies. This done during
-    application bootstrap, to reduce the computation required during run-time.
+    This class is used to create_signature_model the required kwargs for a given handler and its dependencies.
```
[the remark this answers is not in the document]
Goldziher: Yup

starlite/routes.py
Goldziher: Yup

starlite/handlers/base.py
```diff
@@ -98,25 +98,25 @@ def dependency_name_set(self) -> Set[str]:
...
+        layers = []
+
         cur: Any = self
         while cur:
-            value = cur
+            layers.append(cur)
             cur = cur.owner
-            yield value
+
+        yield from reversed(layers)
```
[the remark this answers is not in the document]
Goldziher: Aight

starlite/kwargs.py
Goldziher: Fair point

starlite/app.py
```diff
@@ -47,7 +47,7 @@
 from starlite.utils import normalize_path
 from starlite.utils.templates import create_template_engine
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
```
peterschutt, line 50: We could add this to `pyproject.toml`:

```toml
[tool.coverage.report]
exclude_lines = [
    'if TYPE_CHECKING:',
    'pragma: no cover',
]
```

starlite/openapi/parameters.py
```diff
@@ -65,66 +72,127 @@ def list(self) -> List[Parameter]:
...
+        field_info = layer_field_info
+        # allow users to manually override Parameter definition using Parameter
+        if signature_field_info.extra.get(EXTRA_KEY_IS_PARAMETER):
+            field_info = signature_field_info
+
+        field_info.default = (
+            signature_field_info.default
+            if signature_field_info.default not in [Undefined, Ellipsis]
+            else layer_field_info.default
+        )
```
peterschutt, line 180:
```python

    parameters = {"param": Parameter(value_type=int, default=5)}
    route = "/"

    @get("override-default")
    def override_default(self, param: int = Parameter(default=...)) -> dict[str, Any]:
        """GET /with-default"""
        # returns {"param": 5}
        return {"param": param}
```
Just questioning whether `Parameter(default=...)` should be treated different to overriding the parameter with something like `Parameter(default=6)`? I think it would have less WTF factor if the ellipsis thing worked.

peterschutt (approved):
On board overall, most of the things I tested worked the way I expected they might except for not allowing overriding the default with `...`.

Do you think we should try to police that the types in the handler signature are the same as the parameter type declared in the layer?

starlite/openapi/parameters.py
Goldziher: So, what should default ellipsis do?

PR #262, opened 2022-07-07
Issue #255: allow single element query param in sequence types

### LINKED ISSUE #255: Query parameter arrays with 1 element aren't parsed as arrays
I've been reading through the docs and so far I'm loving how domain modelling is built in with pydantic! Experimenting with query params I noticed that query parameter arrays with 1 element end up with type `T` rather than `List[T]`.  See the example below:

```
from typing import List

@get("/test")
async def test(myArray: List[int]) -> int:
    return 0

app = Starlite(route_handlers=[test])
```

Running 
- `curl "localhost:8000/test?myArray=1"` gives a 400 with `value is not a valid list`. 
- `curl "localhost:8000/test?myArray=1&myArray=2` is fine.  

The workaround isn't too bad - declare myArray as `Union[int, List[int]]` and handle the different types in the body of the function, but I wonder if from an ergonomics perspective its nicer to wrap that logic into the library itself? Something like

```
from pydantic.generics import GenericModel
from typing import List, Generic, TypeVar

DataT = TypeVar('DataT')

class QueryParamArray(GenericModel, Generic[DataT]):
    __root__: Union[DataT, List[DataT]] 

    @property
    def array(self):
        return self.__root__ if isinstance(self.__root__, list) else [self.__root__]
```

Then it would be a simple matter of declaring `myArray` as a `QueryParamArray`. Happy to raise a PR to include this but am not sure where the best place to do so is (types.py?). Also happy to be told this is a stupid idea and its up to the users to declare this utility themselves.

Thanks in advance

josepdaniel:
Resolves #255 

Query parameters declared as `List[T]` will be parsed in as arrays even if they only have one element.

### DIFF
--- a/starlite/kwargs.py
+++ b/starlite/kwargs.py
@@ -1,16 +1,14 @@
-from array import array
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Type, Union, cast
 
-
 from pydantic.fields import (
-    FieldInfo,
-    SHAPE_LIST,
-    SHAPE_SET,
-    SHAPE_SEQUENCE,
-    SHAPE_TUPLE,
-    SHAPE_TUPLE_ELLIPSIS,
     SHAPE_DEQUE,
     SHAPE_FROZENSET,
+    SHAPE_LIST,
+    SHAPE_SEQUENCE,
+    SHAPE_SET,
+    SHAPE_TUPLE,
+    SHAPE_TUPLE_ELLIPSIS,
+    FieldInfo,
     ModelField,
     Undefined,
 )
@@ -28,6 +26,17 @@ from starlite.provide import Provide
 from starlite.signature import SignatureModel, get_signature_model
 from starlite.types import ReservedKwargs
 
+# Shapes corresponding to sequences
+SEQ_SHAPES = {
+    SHAPE_LIST,
+    SHAPE_SET,
+    SHAPE_SEQUENCE,
+    SHAPE_TUPLE,
+    SHAPE_TUPLE_ELLIPSIS,
+    SHAPE_DEQUE,
+    SHAPE_FROZENSET,
+}
+
 
 class ParameterDefinition(NamedTuple):
     param_type: ParamType
@@ -35,7 +44,7 @@ class ParameterDefinition(NamedTuple):
     field_alias: str
     is_required: bool
     default_value: Any
-    is_array: bool
+    is_sequence: bool
 
 
 class Dependency:
@@ -64,18 +73,6 @@ def merge_parameter_sets(first: Set[ParameterDefinition], second: Set[ParameterD
     return result
 
 
-def is_array(model_field: ModelField):
-    return model_field.shape in [
-        SHAPE_LIST,
-        SHAPE_SET,
-        SHAPE_SEQUENCE,
-        SHAPE_TUPLE,
-        SHAPE_TUPLE_ELLIPSIS,
-        SHAPE_DEQUE,
-        SHAPE_FROZENSET,
-    ]
-
-
 class KwargsModel:
     """
     This class is used to model the required kwargs for a given handler and its dependencies.
@@ -91,7 +88,7 @@ class KwargsModel:
         "expected_path_params",
         "expected_query_params",
         "expected_reserved_kwargs",
-        "array_query_parameter_names",
+        "sequence_query_parameter_names",
     )
 
     def __init__(
@@ -104,7 +101,7 @@ class KwargsModel:
         expected_path_params: Set[ParameterDefinition],
         expected_query_params: Set[ParameterDefinition],
         expected_reserved_kwargs: Set[ReservedKwargs],
-        array_query_parameter_names: Set[str],
+        sequence_query_parameter_names: Set[str],
     ) -> None:
         self.expected_cookie_params = expected_cookie_params
         self.expected_dependencies = expected_dependencies
@@ -113,7 +110,7 @@ class KwargsModel:
         self.expected_path_params = expected_path_params
         self.expected_query_params = expected_query_params
         self.expected_reserved_kwargs = expected_reserved_kwargs
-        self.array_query_parameter_names = array_query_parameter_names
+        self.sequence_query_parameter_names = sequence_query_parameter_names
         self.has_kwargs = (
             expected_cookie_params
             or expected_dependencies
@@ -139,7 +136,7 @@ class KwargsModel:
 
     @staticmethod
     def create_parameter_definition(
-        allow_none: bool, field_info: FieldInfo, field_name: str, path_parameters: Set[str], is_array: bool
+        allow_none: bool, field_info: FieldInfo, field_name: str, path_parameters: Set[str], is_sequence: bool
     ) -> ParameterDefinition:
         """
         Creates a ParameterDefition for the given pydantic FieldInfo instance and inserts it into the correct parameter set
@@ -167,7 +164,7 @@ class KwargsModel:
             field_alias=field_alias,
             default_value=default_value,
             is_required=is_required and (default_value is None and not allow_none),
-            is_array=is_array,
+            is_sequence=is_sequence,
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


Goldziher:
Hi there,

Thanks for contributing! Please add tests.

josepdaniel:
Sure thing @Goldziher, just added some and also added some missing logic.

josepdaniel:
@Goldziher made the requested changes - the array handling stuff now lives in `parsers.py` by adding an optional argument to `parse_query_params` representing keys where an array is expected. Please take another look and let me know what you think.

peterschutt:
Can you rebase please @josepdaniel?

starlite/kwargs.py
```diff
@@ -267,30 +269,33 @@ def to_kwargs(self, connection: Union[WebSocket, Request]) -> Dict[str, Any]:
                 reserved_kwargs["socket"] = connection
             if "data" in self.expected_reserved_kwargs:
                 reserved_kwargs["data"] = self.get_request_data(request=cast(Request, connection))
+        connection_query_params = parse_query_params(
+            connection, {p.field_name for p in self.expected_query_params if p.is_array}
+        )
```
peterschutt, line 274: Given that `to_kwargs()` is called in line with the request/response, I think it would be better to aggregate the set of array type query parameter names in `create_for_signature_model()` which is called during bootstrap. So instead of adding `is_array` to `ParameterDefinition` you could add `array_query_parameter_names` to `__slots__`, aggregate the set while iterating over the signature model fields in `create_for_signature_model()` and store it on the instance. Then, pass that already compiled set into `parse_query_params()` here.

Also, perhaps this should be evaluated before the `if self.expected_reserved_kwargs:` block, so that the mapping of query params injected if the `query` reserved kwarg is specified is also accurate for this case.

josepdaniel: Great point, thanks for the feedback. I completely missed the `expected_reserved_kwargs` - I added an additional test case to check the `query` kwarg is in line with expected parameters.

starlite/kwargs.py
```diff
@@ -252,6 +259,7 @@ def to_kwargs(self, connection: Union[WebSocket, Request]) -> Dict[str, Any]:
         Return a dictionary of kwargs. Async values, i.e. CoRoutines, are not resolved to ensure this function is sync.
         """
         reserved_kwargs: Dict[str, Any] = {}
+        connection_query_params = parse_query_params(connection, self.array_query_parameter_names)
```
Goldziher, line 262: this shouldn't be here and the parse_query_params interface should not be modified (i.e. it shouldn't receive any other parameters), because it should have an identical output when called as `request.query` and here.

starlite/kwargs.py
```diff
@@ -138,6 +142,8 @@ def create_for_signature_model(
             else:
                 parameter_set = expected_query_parameters
                 field_alias = model_info.extra.get("query") or field_name
+                if model_field.shape == SHAPE_LIST:
```
Goldziher, line 145: What about:

```python
    SHAPE_SEQUENCE,
    SHAPE_SET,
    SHAPE_TUPLE
```

How will this behave with them? And what is the expected behaviour? I would submit that this should be more generic than a list - it should work on all sequence types. And the tests should cover this.

starlite/parsers.py
```diff
@@ -18,39 +29,45 @@
...
+        acc[key] = [value] if is_array else value
     elif isinstance(param, str):
         acc[key] = [param, value]
     else:
         acc[key].append(value)  # type: ignore
     return acc
 
 
-def parse_query_params(connection: "HTTPConnection") -> Dict[str, Any]:
+def parse_query_params(connection: "HTTPConnection", array_parameters: Optional[Set[str]] = None) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
josepdaniel: 🤦 ok, thanks for the feedback! I misunderstood your original comments. I'll fix this up.

starlite/kwargs.py
```diff
@@ -50,6 +64,18 @@ def merge_parameter_sets(first: Set[ParameterDefinition], second: Set[ParameterD
...
+def is_array(model_field: ModelField):
+    return model_field.shape in [
+        SHAPE_LIST,
+        SHAPE_SET,
+        SHAPE_SEQUENCE,
+        SHAPE_TUPLE,
+        SHAPE_TUPLE_ELLIPSIS,
+        SHAPE_DEQUE,
+        SHAPE_FROZENSET,
+    ]
```
peterschutt, line 76: I was going to suggest defining a set in the module scope to speed up the lookup times, but I felt like getting my nerd on:)

This profiles constructing a literal list inside a function and checking if the first element is in the list:
```
$ python -m timeit -s "def f(): 0 in [0,1,2,3,4,5,6]" "f()"
5000000 loops, best of 5: 60.3 nsec per loop
```
This profiles constructing the list as a global and checking for the first element (slower!):
```
$ python -m timeit -s "L = list(range(7))" -s "def f(): 0 in L" "f()"
5000000 loops, best of 5: 69.1 nsec per loop
```
So the construction of the list literal must get optimized away and its more expensive to look up a value from the global scope than the function local scope.

Where the list falls over is for elements other than the first:
```
$ python -m timeit -s "def f(): 6 in [0,1,2,3,4,5,6]" "f()"
2000000 loops, best of 5: 110 nsec per loop
```

I suppose the right thing to do is to use a set literal inside the function as it will have faster lookup times for all but the first element of the list:
```
$ python -m timeit -s "def f(): 6 in {0,1,2,3,4,5,6}" "f()"
5000000 loops, best of 5: 65 nsec per loop
```

... but these are very small numbers indeed.

starlite/parsers.py
```diff
@@ -1,6 +1,17 @@
 from contextlib import suppress
 from functools import reduce
-from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Type, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    Optional,
+    Set,
```
peterschutt, line 9: Unused imports?

peterschutt (changes requested):

starlite/kwargs.py
```diff
@@ -50,6 +64,18 @@ def merge_parameter_sets(first: Set[ParameterDefinition], second: Set[ParameterD
     return result
 
 
+def is_array(model_field: ModelField):
```
Goldziher, line 67: `is_sequence`

starlite/kwargs.py
Goldziher: A set in the global scope sounds best tbh.

starlite/kwargs.py
```diff
@@ -110,10 +139,7 @@ def create_dependency_graph(cls, key: str, dependencies: Dict[str, Provide]) ->
 
     @staticmethod
     def create_parameter_definition(
-        allow_none: bool,
-        field_info: FieldInfo,
-        field_name: str,
-        path_parameters: Set[str],
+        allow_none: bool, field_info: FieldInfo, field_name: str, path_parameters: Set[str], is_array: bool
```
Goldziher, line 142: This should be formatted, can you run `pre-commit run --all-files`?

starlite/kwargs.py
```diff
@@ -220,13 +249,17 @@ def create_for_signature_model(
...
                     path_parameters=path_parameters,
+                    is_array=is_array(model_field),
                 )
             )
 
         expected_path_parameters = {p for p in param_definitions if p.param_type == ParamType.PATH}
         expected_header_parameters = {p for p in param_definitions if p.param_type == ParamType.HEADER}
         expected_cookie_parameters = {p for p in param_definitions if p.param_type == ParamType.COOKIE}
         expected_query_parameters = {p for p in param_definitions if p.param_type == ParamType.QUERY}
+        array_query_parameter_names = set(
```
Goldziher, line 260: Change `array` to `sequence` everywhere please

Goldziher (commented):
Looks good. We left a few comments. Please address, and when the pipeline is green, we'll merge.

Goldziher:
@all-contributors add @josepdaniel code

peterschutt:
Cheers @josepdaniel!

PR #263, opened 2022-07-07
Refactor `Starlite.construct_route_map`

danesolberg:
Addresses #203 

- Quick refactor of `construct_route_map` that extracts the logic into three separate methods.
- Fixes the sonar issue related to code complexity.
- Only real logic change is checking for conflicting path parameters (e.g. /foo/{a:int} vs /foo/{b:int}) by comparing the preexisting `_path_parameters` key on the terminating trie node of a route path to the `route.path_parameters` attribute on the route being added to the trie.  Previously used a set of previously seen paths.

Clearly there's more that can be done here, but still more context to learn before I attempt deeper changes! 

Seems to be an opportunity to take a class-based approach whereby we have a RouteMap and PathSegmentNode's that could simplify some of this logic, but I'm not sure it's worth doing given that the route_map is only used by `StarliteASGIRouter.traverse_route_map`?

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -171,55 +171,69 @@ class Starlite(Router):
 
         return ExceptionHandlerMiddleware(app=app, exception_handlers=exception_handlers, debug=self.debug)
 
-    def construct_route_map(self) -> None:  # noqa: C901 # pylint: disable=R0912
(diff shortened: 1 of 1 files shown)


Goldziher:
Hi, thanks for this. Please add tests.

PR #264, opened 2022-07-08
Enhancement: Move optional dependencies to extras 

### LINKED ISSUE #185: Refactor: Move optional dependencies to extras
Starlite has several potentially optional dependencies: `pydantic-factories` for generating examples in OpenAPI schema, `pyyaml` for `/openapi.yaml` path and—of course—`requests` for testing. If we decide that `requests` is optional, we should consider to make `pydantic-factories` and `pyyaml` not required as well.

_Originally posted by @vrslev in https://github.com/starlite-api/starlite/issues/174#issuecomment-1166215699_

Bobronium:
Closes #185

TODO:
- [x] handle missing optional dependencies
- [ ] create extras in pyproject.toml
- [ ] stop importing these libs at package level
- [ ] change docs accordingly

Context manager was chosen over try/except for three reasons:
1. It looks much cleaner and requires less writing
2. All error-handling is happening in once place
3. It's testable and doesn't require `# pragma: no cover`

### DIFF
--- a/starlite/exceptions.py
+++ b/starlite/exceptions.py
@@ -25,7 +25,29 @@ class StarLiteException(Exception):
 
 
 class MissingDependencyException(StarLiteException, ImportError):
-    pass
+    def __init__(self, feature: str, extras: tuple[str, ...]) -> None:
+        self.feature = feature
+        self.extras = extras
+
+        if len(extras) > 1:
+            _extras_repr = ", ".join(map(repr, extras[:-1])) + " or " + repr(extras[-1])
+        elif extras:
+            _extras_repr = repr(extras[0])
+        else:
+            raise RuntimeError(f"`extras` argument of {self.__class__.__name__} cannot be empty.")
+
+        detail = (
+            f"To use {feature}, install starlite with {_extras_repr} extra:\n"
+            f"e.g. `pip install starlite[{extras[0]}]`\n"
+            f"or `poetry add starlite --extras {extras[0]}`"
+        )
+        super().__init__(detail)
+
+        # this and custom Exception.__repr__ below is a workaround
+        # for https://github.com/starlite-api/starlite/issues/188
+        Exception.__init__(self, detail)
+
+    __repr__ = Exception.__repr__
 
 
 class HTTPException(StarLiteException, StarletteHTTPException):
--- a/starlite/exceptions.py
+++ b/starlite/exceptions.py
@@ -1,7 +1,5 @@
-from __future__ import annotations
-
 from http import HTTPStatus
-from typing import Any
+from typing import Any, Dict, List, Optional, Union
 
 from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.status import (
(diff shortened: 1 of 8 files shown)


starlite/exceptions.py
```diff
@@ -25,18 +27,40 @@ def __repr__(self) -> str:
...
+        # for https://github.com/starlite-api/starlite/issues/188
+        Exception.__init__(self, detail)
+
+    __repr__ = Exception.__repr__
 
 
 class HTTPException(StarLiteException, StarletteHTTPException):
     status_code = HTTP_500_INTERNAL_SERVER_ERROR
-    extra: Optional[Union[Dict[str, Any], List[Any]]] = None
+    extra: dict[str, Any] | list[Any] | None = None
```
Goldziher, line 57: Syntax is not backwards compatible here, use `Union` instead throughout

Goldziher (commented):
@peterschutt can you review this one?

starlite/exceptions.py
Bobronium: It is syntactically backwards compatible though: CI passes for every supported version.

I agree however to change it to Union to comply with the rest of the code base. 

Though I think it's worth to discuss perspective of upgrading annotations in the entire project somewhere.

Goldziher: Interesting that it passes, but this is PEP 604--> https://peps.python.org/pep-0604/, which is python 3.10 only.

peterschutt: It's the postponed annotations that make this correct though.. and these would be the only two modules in the whole project that introduce those.

> Though I think it's worth to discuss perspective of upgrading annotations in the entire project somewhere.

Yep, consistency is key.

peterschutt:
Do you want a review on this yet @Bobronium? 

Pydantic factories seems to be used in quite a few places beyond the sqlalchemy plugin module: https://github.com/starlite-api/starlite/search?q=pydantic_factories

Bobronium:
@peterschutt, not sure about review. A quick overall feedback about chosen path would be perfect.

The main reason I opened this PR is to keep progress visible for others.

peterschutt:
I like the context manager idea :+1:

starlite/plugins/sql_alchemy.py
```diff
@@ -6,16 +6,16 @@
...
-    ImproperlyConfiguredException,
-    MissingDependencyException,
-)
+from starlite.exceptions import ImproperlyConfiguredException
+from starlite.extras import PYDANTIC_FACTORIES, SQLALCHEMY
 from starlite.plugins.base import PluginProtocol
 
-try:
+with PYDANTIC_FACTORIES:
+    from pydantic_factories import ModelFactory
```
peterschutt, line 16: Perhaps these context managers can be named so that their function is more apparent at the call site, maybe `PYDANTIC_FACTORIES_AVAILABLE` or `..._INSTALLED`..

Bobronium: How about a method on Feature?

```py
with PYDANTIC_FACTORIES.ensure_installed():

```

peterschutt: Sure, anything along those lines.

Goldziher:
any update on this?

Bobronium:
> any update on this?

No updates yet, no estimations either, unfortunately. 

I was enjoying my funemployment when I've started this PR, but now I'm back in the business, so don't have as much free time and urge to write code as I had.

Goldziher:
reopen if you work on this.

PR #265, opened 2022-07-08
Add `TYPE_CHECKING` blocks to all required files

### LINKED ISSUE #253: Refactor: Consistent Type Checking Blocks
A lot of the files do not use a type checking block, and thus imports only for typing are imported at runtime - increasing memory usage unnecessarily. The task would be to a flake8 type checking plugin (for example [this one](https://github.com/snok/flake8-type-checking)) to the `.pre-commit-config.yaml` file, run `pre-commit run --allf-files`, and fix all reported issues.

Harry-Lees:
Closes #253 

### Changes

* Add `flake8-type-checking` extension to pre-commit
* Enable `pydantic` and `FastAPI` configuration options for `flake8-type-checking` extension
* Fixed all `flake8-type-checking` linting errors.

There were a couple of things I wanted to note:

#### 1

In `starlite/types.py`, there was some imports which were converted to `Any` when not `TYPE_CHECKING`. I left these as they were and added `noqa: TC004` to prevent the linter from erroring those lines.

#### 2

In `starlite/routes.py` there were several imports which raised errors when placed in the `TYPE_CHECKING` block.

```python
from starlite.handlers import ASGIRouteHandler  # noqa: TC001
from starlite.handlers import BaseRouteHandler  # noqa: TC001
from starlite.handlers import HTTPRouteHandler  # noqa: TC001
from starlite.handlers import WebsocketRouteHandler  # noqa: TC001
```

I believe this is due to the fact that the pydantic `@validate_arguments` decorator does some magic under the hood which the flake8 extension doesn't detect. AFAIK this is the only instance in the codebase where this interaction occurs, so I just marked all the imports as `noqa: TC001`.

#### 3
After I noticed point 2 happening, I enabled the `type-checking-fastapi-enabled = true` option in `.flake8`. This didn't actually fix 2, but the docs say

> This will treat the annotations of any decorated function as needed at runtime.

which I thought could be helpful in avoiding other issues with decorators confusing the type checker. Disabling this option does actually recommend quite a few more types be moved into `TYPE_CHECKING` blocks, I didn't test to see how many (if any) of these new recommendations caused errors.


Goldziher:
@all-contributors add @Harry-Lees code

Issue #177, opened 2022-06-25
Rust, C or C++ bindings for path resolution


Goldziher:
Starlite currently resolves paths using a tree structure, but in python. This is quite slow. It would be very helpful if we could create custom bindings in a low level and substantially faster language to implement the tree and its required parsing methods. 

The data structure that is often used for this purpose is called a "radix prefix tree". This is not necessarily the only option we have - but its a strong one. 

PRs are welcome, as are discussions.

Bobronium:
Some links to the existing code in starlite would be helpful to start.

Also, have you considered/tried compiling starlite/parts of it with mypyc?

peterschutt:
> Also, have you considered/tried compiling starlite/parts of it with mypyc?

100% this. I'd love to see what the solution looks like in modern cython (i say modern because the last time I built anything of consequence with cython was before typing was really a thing) and mypyc.

Thinking about this issue on the next meta level up, an overall efficiency analysis of the two main things `Starlite` does is something that I'm keen to do. From my understanding so far, these two main things are:
1. serve requests, sounds obvious, the route from request entry to response return would benefit from some diagrams and docs, and even the smallest of optimizations along that path are +1 from me.
2. the handler -> controller -> router -> app walk on startup - the whole library is pretty much built on this and so it should be pretty optimal I think. I reckon it also might be neat to make this walk a feature of library startup and let users register hooks onto it. For example, I might want to register a hook onto the walk that logs the number of routes that I have nested under each router/controller and how many of them are unauthenticated and authenticated. Just an example, but it would be a neat and I think unique feature that we could promote.

peterschutt:
Don't get me wrong, I'd also like to see what the solution looks like with py03 and rust too though :laughing: 

It would be good for developer docs to have a diagram of that "radix prefix tree" pattern included. I'm glad you mentioned that @Goldziher as the code is a bit hard to follow (not a criticism but just by nature) - is that the algo in `StarliteASGIRouter.__call__()`?

Goldziher:
I did some trials with cython and mypyc. Mypyc Didn't like async and some other stuff, so it could handle some parts of the codebase and not others. Cython required quite a few changes to the code to get the compiled code to actually work. 

I'm not against this, but for money these solutions are a compromise - offering a performance boost, but less so than native code. 

As to code snippets, diagrams etc. Sure, I'll do this later today.

peterschutt:
[...]

PR #270, opened 2022-07-09
Issue #177: Add a baseline for Rust path resolution bindings

nramos0:
Addresses #177.

Baseline implementation very closely mimics the original Python trie implementation for path resolution, but uses Rust types for native performance.

A native extension wheel can be built using `maturin build` in a Poetry shell, which will output to `./target/wheels`, but the packaging of the wheels into the Starlite package has not been done. I think this should be doable by calling a build script from Poetry and including the wheels somehow, if it is not otherwise possible to directly include the wheel directory's wheels.

The native extension has to be built with `maturin develop` before running `pytest` in order to run tests locally.


starlite/extensions/rust/route_map.rs
```diff
@@ -0,0 +1,410 @@
...
+        Ok(())
+    }
+
+    /// Checks if a given path refers to a static path
+    pub fn is_static_path(&self, path: &str) -> PyResult<bool> {
+        Ok(self.static_paths.contains(path))
+    }
+
+    /// Removes a path from the static path set
+    pub fn remove_static_path(&mut self, path: &str) -> PyResult<bool> {
```
Goldziher, line 117: Interesting, but do we need this? Or is this future proofing?

starlite/extensions/rust/route_map.rs
```diff
@@ -0,0 +1,410 @@
...
+            let path_parameters: Vec<HashMap<String, Py<PyAny>>> =
+                route.getattr("path_parameters")?.extract()?;
+
+            let cur = self.add_node_to_route_map(&ctx, route, path, &path_parameters[..])?;
+
+            let cur_path_parameters = cur.path_parameters.as_ref().unwrap();
+
+            if !path_parameters_eq(cur_path_parameters, &path_parameters, py)? {
+                return Err(ImproperlyConfiguredException::new_err(
+                    "Should not use routes with conflicting path parameters",
```
Goldziher, line 165: Please use the existing errors if possible

starlite/route_map.py
```diff
@@ -0,0 +1,62 @@
+from typing import Any, Callable, Collection, Dict, List, Tuple
```
Goldziher, line 1: Please add a comment on the class explaining it's a proxy for rust bindings. Also add #pragma: no cover top level, because this file is basically a shim.

Goldziher (commented):
First off, very nice stuff and thank you. It's a great first step. I left some comments - 

1. Implementation is not identical I think, and I did not see a 405 being raised anywhere. I would double check and compare the python code to the rest code. 

2. I'm missing unit tests on the rest code. How involved is adding this and updating CI?

3. Whats needed for a setup script that builds the rust component correctly before publishing? 

4. How can we run integration tests locally?

5. What can of tooling should he added to `pre-commit`?

I'm tagging @peterschutt here so he gives input

Dr-Emann:
I also have been working on an [implementation](https://github.com/Dr-Emann/starlite_rust_router) as well, currently out-of-tree, I think we should work together to try to take the best of both worlds.

My implementation deviates from the original python one a fair bit in order to be more type-safe, and less stringly-typed. This looks more integrated with the existing repository though.

nramos0:
> First off, very nice stuff and thank you. It's a great first step. I left some comments -
> 
>     1. Implementation is not identical I think, and I did not see a 405 being raised anywhere. I would double check and compare the python code to the rest code.
> 
>     2. I'm missing unit tests on the rest code. How involved is adding this and updating CI?
> 
>     3. Whats needed for a setup script that builds the rust component correctly before publishing?
> 
>     4. How can we run integration tests locally?
> 
>     5. What can of tooling should he added to `pre-commit`?
> 
> 
> I'm tagging @peterschutt here so he gives input

Thank you for the comments, I will take a look at these later in the day and get back to you with some changes.

nramos0:
> I also have been working on an [implementation](https://github.com/Dr-Emann/starlite_rust_router) as well, currently out-of-tree, I think we should work together to try to take the best of both worlds.
> 
> My implementation deviates from the original python one a fair bit in order to be more type-safe, and less stringly-typed. This looks more integrated with the existing repository though.

Absolutely. I was also wondering if there was a way to do this in a more type-safe way, and I'd be happy to collaborate on the implementation. I'll take a look at your code later today as well and see if we can talk about next steps.

build.py
```diff
@@ -0,0 +1,4 @@
+
+
+def build():
```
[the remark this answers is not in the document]
nramos0: In retrospect I'm not sure why I included this in the end commit 😂 I originally intended to write the script to run `maturin` here and then include the produced wheels into the package, but I'm not familiar with how to do that. Will update this in the future if this is the proper method to handle the setup script

pyproject.toml
```diff
@@ -42,6 +42,7 @@ pyyaml = "*"
 starlette = "*"
 typing-extensions = "*"
 requests = { version = "*", optional = true }
+maturin = "^0.12.20"
```
vrslev, line 45: Shouldn’t it be a development dependency?

starlite/app.py
```diff
@@ -12,6 +12,8 @@
 from starlette.types import ASGIApp, Receive, Scope, Send
 from typing_extensions import Type
 
+from rust_backend import RouteMap as RouteMapInit
```
vrslev, line 15: Maybe,  change the module name from `rust_backend` to something like `starlite._rust` or `starlite._route_map`?

starlite/extensions/rust/route_map.rs
nramos0: Future proofing, wasn't sure if something like this might be necessary, since Python cannot directly access the Rust static path/plain route sets so it doesn't have full control over their data. If you all think this won't be needed I'll get rid of it

pyproject.toml
Goldziher: it should

nramos0:
> First off, very nice stuff and thank you. It's a great first step. I left some comments -
> 
>     1. Implementation is not identical I think, and I did not see a 405 being raised anywhere. I would double check and compare the python code to the rest code.
> 
>     2. I'm missing unit tests on the rest code. How involved is adding this and updating CI?
> 
>     3. Whats needed for a setup script that builds the rust component correctly before publishing?
> 
>     4. How can we run integration tests locally?
> 
>     5. What can of tooling should he added to `pre-commit`?
> 
> 
> I'm tagging @peterschutt here so he gives input

To reply to some of your points the best I can here:

1. I noticed that in the original code, `MethodNotAllowedException` is raised by `StarliteASGIRouter.resolve_asgi_app`, which I did not port to Rust and is still being called by `StarliteASGIRouter.__call__`. Should the resolution of the actual handler as done in `resolve_asgi_app` also be handled by the Rust code? When writing the extension, I wasn't sure to what extent I should move tasks out of the Python code and into the Rust code, so I only replaced up to `StarliteASGIRouter.parse_scope_to_route`.

2. As the code stands, I'm not so certain about how we might go about writing pure Rust unit tests for the Rust component, since we would need to somehow mock the Starlite instance and others using PyO3. That is, the Rust code needs pieces from the Python environment that normal Rust unit tests wouldn't have access to. But we could probably do it either with maybe a Python wrapper (it seems possible that this could just be more tests in the `./tests/` directory that are directed at the Rust code?), or I think perhaps this might be possible with directly evaluating Python code (which PyO3 supports I believe, see [here](https://pyo3.rs/v0.16.4/python_from_rust.html#include-multiple-python-files)) to import and instantiate Starlite in Rust native tests before making calls to the instance.

3. As far as I'm aware, a script running in the project's Poetry-instantiated virtual environment with all `pyproject.toml` dependencies installed should be able to run `maturin build` to generate native extension wheels. Then perhaps in the `tool.poetry` section of `pyproject.toml` there could be something like 
```toml
include = [
    {path = "target/wheels/*.whl", format = "wheel"}
]
```
But when trying this and building the Starlite wheel with `poetry build`, and then installing the wheel in a new virtual environment, I get that the `rust_backend` module (the native extension's module name) cannot be found. Does anyone know what the problem might be here? Poetry doesn't seem to include the native extension wheel in the end Starlite wheel. The same happens trying to install the .tar.gz package. Will do more research on what is wrong with this.

4. Are the integration tests the ones that run with `pytest`? If so, in a fresh install of my forked branch, the following works:
- `poetry shell`
- `poetry install`
- `maturin develop`
- `pytest`

5. Perhaps we can look at adding something that uses `cargo fmt` or `cargo clippy`, which are a Rust formatter and linter respectively. I'm not familiar with `pre-commit` and its config file

Goldziher:
> > First off, very nice stuff and thank you. It's a great first step. I left some comments -
> > ```
> > 1. Implementation is not identical I think, and I did not see a 405 being raised anywhere. I would double check and compare the python code to the rest code.
> > 
> > 2. I'm missing unit tests on the rest code. How involved is adding this and updating CI?
> > 
> > 3. Whats needed for a setup script that builds the rust component correctly before publishing?
> > 
> > 4. How can we run integration tests locally?
> > 
> > 5. What can of tooling should he added to `pre-commit`?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm tagging @peterschutt here so he gives input
> 
> To reply to some of your points the best I can here:
> 
> 1. I noticed that in the original code, `MethodNotAllowedException` is raised by `StarliteASGIRouter.resolve_asgi_app`, which I did not port to Rust and is still being called by `StarliteASGIRouter.__call__`. Should the resolution of the actual handler as done in `resolve_asgi_app` also be handled by the Rust code? When writing the extension, I wasn't sure to what extent I should move tasks out of the Python code and into the Rust code, so I only replaced up to `StarliteASGIRouter.parse_scope_to_route`.
> 2. As the code stands, I'm not so certain about how we might go about writing pure Rust unit tests for the Rust component, since we would need to somehow mock the Starlite instance and others using PyO3. That is, the Rust code needs pieces from the Python environment that normal Rust unit tests wouldn't have access to. But we could probably do it either with maybe a Python wrapper (it seems possible that this could just be more tests in the `./tests/` directory that are directed at the Rust code?), or I think perhaps this might be possible with directly evaluating Python code (which PyO3 supports I believe, see [here](https://pyo3.rs/v0.16.4/python_from_rust.html#include-multiple-python-files)) to import and instantiate Starlite in Rust native tests before making calls to the instance.
> 3. As far as I'm aware, a script running in the project's Poetry-instantiated virtual environment with all `pyproject.toml` dependencies installed should be able to run `maturin build` to generate native extension wheels. Then perhaps in the `tool.poetry` section of `pyproject.toml` there could be something like
> 
> ```toml
> include = [
>     {path = "target/wheels/*.whl", format = "wheel"}
> ]
> ```
> 
> But when trying this and building the Starlite wheel with `poetry build`, and then installing the wheel in a new virtual environment, I get that the `rust_backend` module (the native extension's module name) cannot be found. Does anyone know what the problem might be here? Poetry doesn't seem to include the native extension wheel in the end Starlite wheel. The same happens trying to install the .tar.gz package. Will do more research on what is wrong with this.
> 
> 4. Are the integration tests the ones that run with `pytest`? If so, in a fresh install of my forked branch, the following works:
> 
> * `poetry shell`
> * `poetry install`
> * `maturin develop`
> * `pytest`
> 
> 5. Perhaps we can look at adding something that uses `cargo fmt` or `cargo clippy`, which are a Rust formatter and linter respectively. I'm not familiar with `pre-commit` and its config file

Hi, so as to 1 - yes, there is little benefit in having the route map logic in rust and then do the parsing of routes in the `ASGIRouter` in python - this IS the bottleneck.

pyproject.toml
nramos0: fixed

starlite/extensions/rust/route_map.rs
```diff
@@ -0,0 +1,410 @@
...
+
+impl Node {
+    /// Creates a new trie node
+    pub fn new() -> Self {
+        Self {
+            components: HashSet::new(),
+            children: HashMap::new(),
+            path_parameters: None,
+            asgi_handlers: None,
+            is_asgi: None,
```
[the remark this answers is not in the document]
nramos0: fixed

starlite/app.py
vrslev: I guess we don’t have to if there are stubs

Dr-Emann:
I think I figured out building using poetry: https://github.com/nramos0/starlite/pull/1

nramos0:
> I think I figured out building using poetry: [nramos0#1](https://github.com/nramos0/starlite/pull/1)

Nice, I will take a look!

nramos0:
> I think I figured out building using poetry: [nramos0#1](https://github.com/nramos0/starlite/pull/1)

Seems to work out all fine for me, thanks, will merge now. Though I discovered I had a strange issue where I had to initialize the virtual environment with upgraded dependencies, since the default `setuptools` version was too old and didn't contain modules required by `setuptools_rust`, not sure if anyone else will run into this.

nramos0:
Packaging now works thanks to @Dr-Emann's PR (and I also borrowed some of the external python type/function handle import logic from your implementation). Building is no longer done with `maturin`, but instead with `setuptools_rust`. Building the Rust extension for development needs a `poetry install` to rebuild every time there is a change in the Rust code. For packaging for release, `poetry build` as normal will do.

Additionally, I've renamed module to `route_map_rs`, which causes the imports of the proxy type and actual extension respectively to look like this
```python
from starlite.route_map import RouteMap
from starlite.route_map_rs import RouteMap as RouteMapInit
```
Not sure if this would be a preferred naming convention / if there is a clearer way to add typings to the extension module. @Goldziher @vrslev

nramos0:
Currently we are still missing Rust unit tests.

As for the running the `pytest` integration tests, the following now works:
1. `poetry shell`
2. `poetry install`
3. `pytest`

However, if you get the same issue as me where `setuptools.command.build` is not found due to an old `setuptools` version installed in new virtual environments by default, running `pip install --upgrade setuptools` directly before `poetry install` works.

Goldziher:
Great stuff guys!

Re `setuptools` - cant you simply add it as a dev dependency? 

Re naming - looks good. 

@peterschutt any input?

extensions/rust/src/route_map.rs
```diff
@@ -0,0 +1,454 @@
...
+    starlite: Py<PyAny>,
+    ctx: StarliteContext,
+}
+
+// The functions below are available to Python code
+#[pymethods]
+impl RouteMap {
+    /// Creates an empty `RouteMap`
+    #[new]
+    pub fn new(starlite: Py<PyAny>) -> PyResult<Self> {
```
Dr-Emann, line 95: The first non-self argument to `pyo3` methods can be a `Python<'_>`, so you don't have to go through `Python::acquire_gil().python()`

extensions/rust/src/route_map.rs
```diff
@@ -0,0 +1,454 @@
...
+                http_route,
+                web_socket_route,
+                asgi_route,
+                parse_path_params,
+            },
+        })
+    }
+
+    /// Adds a new static path by path name
+    pub fn add_static_path(&mut self, path: &str) -> PyResult<()> {
```
Dr-Emann, line 128: If the method can't fail, you don't have to return PyResult

nramos0:
> Great stuff guys!
> 
> Re `setuptools` - cant you simply add it as a dev dependency?
> 
> Re naming - looks good.
> 
> @peterschutt any input?

Strangely enough, I can't seem to get this to work. Including a newer `setuptools` version as a dev dependency and then creating a new virtual env, activating it and then running `poetry install` just doesn't update the default `setuptools` (I think it was version 58 or so). My current workaround for the problem is to just use `virtualenv` instead of `python -m venv`, which defaults to a much newer version of `setuptools` (62) for me.

extensions/rust/src/route_map.rs
nramos0: Fixed, thanks

extensions/rust/src/route_map.rs
nramos0: Fixed this for all of the other methods too

peterschutt:
> Strangely enough, I can't seem to get this to work. Including a newer setuptools version as a dev dependency and then creating a new virtual env, activating it and then running poetry install just doesn't update the default setuptools (I think it was version 58 or so). My current workaround for the problem is to just use virtualenv instead of python -m venv, which defaults to a much newer version of setuptools (62) for me.

Same for me.

I ran these commands in a fresh shell:

1. Make sure latest and greatest: `$ pyenv local 3.10.5 && pyenv shell 3.10.5`
3. Make sure setuptools and pip up-to-date: `$ python3.10 -m pip install --upgrade setuptools pip`
```sh
Requirement already satisfied: setuptools in /home/peter/.pyenv/versions/3.10.5/lib/python3.10/site-packages (63.2.0)
Requirement already satisfied: pip in /home/peter/.pyenv/versions/3.10.5/lib/python3.10/site-packages (22.1.2)
```
4. Tell poetry to use same: `$ poetry env use python3.10`
5. But pip and setuptools are still an old version :man_shrugging: `$ poetry run python -m pip install --upgrade setuptools pip`:
```sh
/home/peter/.pyenv/versions/3.10.5/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
Requirement already satisfied: setuptools in /home/peter/.cache/pypoetry/virtualenvs/starlite-Zk1-j2lI-py3.10/lib/python3.10/site-packages (58.3.0)
Collecting setuptools
  Using cached setuptools-63.2.0-py3-none-any.whl (1.2 MB)
Requirement already satisfied: pip in /home/peter/.cache/pypoetry/virtualenvs/starlite-Zk1-j2lI-py3.10/lib/python3.10/site-packages (21.3.1)
Collecting pip
  Using cached pip-22.1.2-py3-none-any.whl (2.1 MB)
Installing collected packages: setuptools, pip
  Attempting uninstall: setuptools
    Found existing installation: setuptools 58.3.0
    Uninstalling setuptools-58.3.0:
      Successfully uninstalled setuptools-58.3.0
  Attempting uninstall: pip
    Found existing installation: pip 21.3.1
    Uninstalling pip-21.3.1:
      Successfully uninstalled pip-21.3.1
Successfully installed pip-22.1.2 setuptools-63.2.0
```
6. All good after that `$ poetry install`

I also had to update my rust toolchain as I was hitting https://github.com/rust-lang/rust/issues/71126 during the build.

peterschutt:
Here's the fix for the poetry setuptools/pip business:

1. Poetry uses `virtualenv` to create the env so thats where the setuptools/pip wheels come from: `$ python3.10 -m virtualenv --upgrade-embed-wheels`
2. Make sure poetry version is `1.1.14`: `$ poetry self update`
3. Get rid of existing poetry environment: `poetry env remove python`
4. Tell poetry to use the python version where you know `virtualenv` wheels are up-to-date: `$ poetry env use python3.10`
5. `poetry install`

peterschutt:
Is the interface in `route_map.py` settled, and should we start porting the existing code to that interface?

Dr-Emann:
Can we reduce the interface more, do we need to expose e.g. `traverse_to_dict`? It looks like `add_plain_route` is never called from outside, (and it feels like it never should be). It also feels awkward to me that `add_routes` takes no arguments.

nramos0:
> Can we reduce the interface more, do we need to expose e.g. `traverse_to_dict`? It looks like `add_plain_route` is never called from outside, (and it feels like it never should be). It also feels awkward to me that `add_routes` takes no arguments.

For `add_routes` I totally agree, just changed to allow the routes collection to just be a parameter, and pass in `self.routes` from the `Starlite` instance in the `construct_route_map` method. I also hid the plain route methods from the python interface.

nramos0:
> Here's the fix for the poetry setuptools/pip business:
> 
>     1. Poetry uses `virtualenv` to create the env so thats where the setuptools/pip wheels come from: `$ python3.10 -m virtualenv --upgrade-embed-wheels`
> 
>     2. Make sure poetry version is `1.1.14`: `$ poetry self update`
> 
>     3. Get rid of existing poetry environment: `poetry env remove python`
> 
>     4. Tell poetry to use the python version where you know `virtualenv` wheels are up-to-date: `$ poetry env use python3.10`
> 
>     5. `poetry install`

Worked like a charm!

nramos0:
> Can we reduce the interface more, do we need to expose e.g. `traverse_to_dict`? It looks like `add_plain_route` is never called from outside, (and it feels like it never should be). It also feels awkward to me that `add_routes` takes no arguments.

I wrote `traverse_to_dict` to allow inspection of `Node`s in python tests, like in `./tests/middleware/test_exception_handler_middleware.py`, though on that note I'm trying to come up with a good way to reduce code duplication between `traverse_to_dict` and `traverse_to_node`, which I'm having trouble with since `traverse_to_node` creates side-effects in `scope` and needs to return `path_params`. I'm getting the feeling that a solution to this that I could come up with (with macros or otherwise) would not be clean enough to actually use.

Goldziher:
> > Can we reduce the interface more, do we need to expose e.g. `traverse_to_dict`? It looks like `add_plain_route` is never called from outside, (and it feels like it never should be). It also feels awkward to me that `add_routes` takes no arguments.
> 
> I wrote `traverse_to_dict` to allow inspection of `Node`s in python tests, like in `./tests/middleware/test_exception_handler_middleware.py`, though on that note I'm trying to come up with a good way to reduce code duplication between `traverse_to_dict` and `traverse_to_node`, which I'm having trouble with since `traverse_to_node` creates side-effects in `scope` and needs to return `path_params`. I'm getting the feeling that a solution to this that I could come up with (with macros or otherwise) would not be clean enough to actually use.

We don't need to fanatically dry the code. If it becomes an issue, I'd say - let it be. Also, refactors can always come after. That's why I'd suggest to focus on tests, since they enable easier refactors.

Goldziher:
Any progress on the tests?

nramos0:
> Any progress on the tests?

Working on this, should have an update within the next day

peterschutt:
Is something like https://github.com/sumerc/yappi what we'd be looking for as a profiling tool?

nramos0:
So the last few commits added a bit more refactoring to allow for easier testing, along with a very minimal test that can be fairly easily copied to directly add new tests. I ported `build_route_middleware_stack` and `wrap_in_exception_handler` to Rust functions to remove dependence on the Starlite instance (I discovered this might cause some headaches with testing, it feels cleaner without the dependence as well).

There are a few things to mention for the testing. 
1. The current python environment must contain the libpython shared library in order for PyO3 to dynamically embed a Python interpreter into Rust, but this is not the default for my OSX installations. I did this by reinstalling all of my python versions using `pyenv` with the command `env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install x.y.z` as described [here](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#special-environment-variables).
2. While PyO3 can identify the current virtual environment, it appears that it will not select the proper `lib` folder from the virtual env itself, but from the python installation that created the virtual env, so this causes the Rust tests to not be able to find any of Starlite's dependencies when it tries to import. This can be solved by using the `PYTHONPATH` env variable and setting it to the virtual environment's `lib` and then running tests with `cargo` like so: 
```env PYTHONPATH="/Users/haxi/Documents/code/starlite/nramos-starlite/.venv/lib/python3.10/site-packages" cargo test --no-default-features```
4. Testing must be run with `cargo test --no-default-features`, because tests will fail with some errors if the `extension-module` PyO3 feature is enabled, which is the same feature needed to compile the extension in the first place. So I set up a default feature `extension-module` for the Rust extension crate that just maps to enabling PyO3's `extension-module`, but when we test we can just disable it. Testing itself requires dynamically embedding a Python interpreter, which is done with the `auto-initializer` PyO3 feature. This feature is enabled unconditionally and has no issues if it's still on when compiling the extension.
5. Another caveat is that the `lib` folder of the virtual env will not actually contain Starlite's source code as a package, so Rust tests will also not be able to find it. I got the unit tests up and running by running 
```cp -r starlite ./.venv/lib/python3.10/site-packages/starlite``` 
to just directly make the code available, but I think this should be done in a Cargo build script instead. I will look at doing this next, not sure if there is a better way, but I'm certain at the moment that this works.

I have confirmed that after porting `build_route_middleware_stack`, all `pytest` integration tests still pass, please confirm that it is still working for you too.

Goldziher:
> So the last few commits added a bit more refactoring to allow for easier testing, along with a very minimal test that can be fairly easily copied to directly add new tests. I ported `build_route_middleware_stack` and `wrap_in_exception_handler` to Rust functions to remove dependence on the Starlite instance (I discovered this might cause some headaches with testing, it feels cleaner without the dependence as well).
> 
> There are a few things to mention for the testing.
> 
> 1. The current python environment must contain the libpython shared library in order for PyO3 to dynamically embed a Python interpreter into Rust, but this is not the default for my OSX installations. I did this by reinstalling all of my python versions using `pyenv` with the command `env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install x.y.z` as described [here](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#special-environment-variables).
> 2. While PyO3 can identify the current virtual environment, it appears that it will not select the proper `lib` folder from the virtual env itself, but from the python installation that created the virtual env, so this causes the Rust tests to not be able to find any of Starlite's dependencies when it tries to import. This can be solved by using the `PYTHONPATH` env variable and setting it to the virtual environment's `lib` and then running tests with `cargo` like so:
>    `env PYTHONPATH="/Users/haxi/Documents/code/starlite/nramos-starlite/.venv/lib/python3.10/site-packages" cargo test --no-default-features` Here `nramos-starlite` is the root directory of my Starlite fork. This issue is described in the PyO3 repo at [Can not link to virtual environment on OS X PyO3/pyo3#1741](https://github.com/PyO3/pyo3/issues/1741).
> 3. Testing must be run with `cargo test --no-default-features`, because tests will fail with some errors if the `extension-module` PyO3 feature is enabled, which is the same feature needed to compile the extension in the first place. So I set up a default feature `extension-module` for the Rust extension crate that just maps to enabling PyO3's `extension-module`, but when we test we can just disable it (PyO3 recommends this workaround [here](https://pyo3.rs/v0.16.4/faq.html#i-cant-run-cargo-test-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror)). Testing itself requires dynamically embedding a Python interpreter, which is done with the `auto-initializer` PyO3 feature. This feature is enabled unconditionally and has no issues if it's still on when compiling the extension.
> 4. Another caveat is that the `lib` folder of the virtual env will not actually contain Starlite's source code as a package, so Rust tests will also not be able to find it. I got the unit tests up and running by running
>    `cp -r starlite ./.venv/lib/python3.10/site-packages/starlite`
>    to just directly make the code available, but I think this should be done in a Cargo build script instead, which would do the copying when running `poetry install`, but I'm not quite sure how to handle the virtual env python version and what not. I will look at doing this next, not sure if there is a better way, but I'm certain at the moment that this works.
> 
> I have confirmed that after porting `build_route_middleware_stack`, all `pytest` integration tests still pass, please confirm that it is still working for you too.

Ok, great. Can you perhaps join the discord server? then inform us whenever there is an update so we can easily jump in and test.

Please rebase - and then we can merge your work into a temporary branch in main repo and test it.

Goldziher:
@nramos0 why did you close this issue and the PR? Was this a mistake?

nramos0:
> @nramos0 why did you close this issue and the PR? Was this a mistake?

@Goldziher Sorry about that, rebasing right now. I thought I'd move the commits to another branch, sync the fork's main branch and rebase within the fork. I haven't merged the branch back into this main branch yet so it appears github has automatically closed the PR

PR #271, opened 2022-07-09
Enable postponed evaluation and upgrade typehints

Bobronium:
Oh, I initially forgot about tests. When I upgraded tests as well, many failed. 

Mostly because starlite currently doesn't work with ForwardRefs at all.


Goldziher:
So @Bobronium . I like you take initiative, but I don't understand the reasoning for this PR. Why would we want to do this? We will eventually update our syntax when we only support v3.10+ .

Bobronium:
This IMO improves readability and requires less effort to write annotations.

I don't see what holding us back from using the new syntax inside the library since it's compatible with python 3.7+ when enabling postponed evaluation for typehints. Locally all tests have passed.

CI presumably fails because of a bug in pydantic that I haven't yet looked into much.

As for why I opened PR instead of an issue/discussion: save time and effort. These changes were mostly done automatically, so I'm ok with throwing this PR away :)

Goldziher:
Aight. Lets close this for now.

Bobronium:
Sure. Could you tell why though?

Goldziher:
> Sure. Could you tell why though?

I think its an unnecessary change - its syntactic sugar. And - python 3.11 will not include postponed evaluation is originally intended, see: https://mail.python.org/archives/list/python-dev@python.org/message/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/

PR #273, opened 2022-07-12
Add tests for `Starlite.construct_route_map`

danesolberg:
Adds tests previously requested in #263 

Feel free to merge these or not, given that a new implementation of route resolution is likely forthcoming (Rust or otherwise)


Goldziher (approved):
Great, thanks!

Issue #275, opened 2022-07-12
Hypothesis test case failure in CI


peterschutt:
```
___________________________ test_add_route_map_path ____________________________
    @given(test_paths=route_test_paths())
>   @settings(
        max_examples=5,
        deadline=None,
    )
tests/route/test_route_map.py:61: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test_paths = [('/A', '/A', set()), ('/A/{e:int}/{a:int}', '/A/1/1', {'a', 'e'}), ('/{a:int}/A', '/1/A', {'a'}), ('/A/{a:int}', '/A/1', {'a'}), ('/B', '/B', set()), ('/B/{a:int}', '/B/1', {'a'}), ...]
    @given(test_paths=route_test_paths())
    @settings(
        max_examples=5,
        deadline=None,
    )
    def test_add_route_map_path(test_paths: List[RouteMapTestCase]) -> None:
        @get(path=[])
        def handler_fn(a: int = 0, b: int = 0, c: int = 0, d: int = 0, e: int = 0) -> None:
            ...
        client = create_test_client(handler_fn)
        app = client.app
        route_map = app.route_map
        for router_path, _, path_params in test_paths:
>           assert is_path_in_route_map(route_map, router_path, path_params) is False
E           AssertionError: assert True is False
E            +  where True = is_path_in_route_map({'/': {'*': {'A': {'_asgi_handlers': {}, '_components': set(), '_is_asgi': False, '_path_parameters': [{'full': 'a:int...}, '/A': {'_asgi_handlers': {}, '_components': set(), '_is_asgi': False, '_path_parameters': []}, '_components': {'/'}}, '/A/{a:int}', {'a'})
tests/route/test_route_map.py:74: AssertionError
---------------------------------- Hypothesis ----------------------------------
Falsifying example: test_add_route_map_path(
    test_paths=[('/A', '/A', set()),
     ('/A/{e:int}/{a:int}', '/A/1/1', {'a', 'e'}),
     ('/{a:int}/A', '/1/A', {'a'}),
     ('/A/{a:int}', '/A/1', {'a'}),
     ('/B', '/B', set()),
     ('/B/{a:int}', '/B/1', {'a'}),
     ('/C', '/C', set()),
     ('/C/{a:int}', '/C/1', {'a'}),
[...]

PR #277, opened 2022-07-14
Fix route-map path existence test logic (#275)

danesolberg:
This fixes the acute issue raised in #275.   I would also love to hear any tips on forcing hypothesis to run a strategy continuously locally in order to find problems like these ahead of time.  I like the library, but it seems to fight you (even if just from a performance standpoint) if you want to run large test sets.


Goldziher:
I have similar experiences with hypothesis

Issue #250, opened 2022-07-05
Feature: Extend lifecycle hooks


Goldziher:
We currently support two lifecycle hooks for HTTP handlers, it would be good to extend and systematize these because it can be very useful for users. I would like to open this to discussion. 

So questions- 
1. at what junctures does it make sense to add a lifecycle hook?
2. what are some of the use case(s) for the suggested hook?
3. are there any particular difficulties in implementing a suggested hook?
4. are there any arguments against implementing a suggested hook?

Goldziher:
So here are some suggestions:

1. An `after_response` hook that will fire once the response cycle finishes. The use case for this I can see is to be able to plug some instrumentation, e.g. prometheus, to gather matrics, emit logs etc. I don't see any difficulties implementing it. I also dont really see a reason not to do this, but maybe someone else does?
2. An `after_exception` hook that will fire after an exception response has been handled. The use case is basically identical to the one above, i.e. instrumentation. Again I dont see difficulties implementing it. One reason not to do this is to avoid bloat - also we might be able to combine this hook with the previous one suggested and then just have a single hook.

wassafshahzad:
@Goldziher We could take a look at Django signals, There are a few good examples which could be useful such as `pre_save` and `post_save`

Goldziher:
> @Goldziher We could take a look at Django signals, There are a few good examples which could be useful such as `pre_save` and `post_save`

But Django has a builtin ORM. We don't. Do you know Fastify? We can take a queue from what they got.

wassafshahzad:
> > @Goldziher We could take a look at Django signals, There are a few good examples which could be useful such as `pre_save` and `post_save`
> 
> But Django has a builtin ORM. We don't. Do you know Fastify? We can take a queue from what they got. 

I can look into it

wassafshahzad:
@Goldziher sorry for the late reply, I looked into fastify lifecyclehooks and `preSerialization` looks like a good hook to add in addition to `after_response` and `after_exception`

This PR will take some time since I am new to this repo and library in general

Goldziher:
> @Goldziher sorry for the late reply, I looked into fastify lifecyclehooks and `preSerialization` looks like a good hook to add in addition to `after_response` and `after_exception`
> 
> This PR will take some time since I am new to this repo and library in general

No problem. I might do this depending on time available.

PR #279, opened 2022-07-14
added after_response

Goldziher:
This PR adds an `after_response` life-cycle hook, as discussed in #250


starlite/handlers/http.py
```diff
@@ -225,6 +233,23 @@ def resolve_after_request(self) -> Optional[AfterRequestHandler]:
...
+        This method is memoized so the computation occurs only once
+        """
+        if self.resolved_after_response is BaseRouteHandler.empty:
+            self.resolved_after_response = None
+            for layer in self.ownership_layers:
+                if layer.after_response:
+                    self.resolved_after_response = layer.after_response
+            if self.resolved_after_response is not None and ismethod(self.resolved_after_response):
+                # python automatically binds class variables, which we do not want in this case.
+                self.resolved_after_response = self.resolved_after_response.__func__
```
peterschutt, line 250: Does using 
```python
if layer.after_response:
    self.resolved_after_response = staticmethod(layer.after_response)
```
remove the need to do this?

starlite/handlers/http.py
```diff
@@ -258,13 +283,16 @@ def validate_handler_function(self) -> None:
         if "data" in signature.parameters and "GET" in self.http_methods:
             raise ImproperlyConfiguredException("'data' kwarg is unsupported for 'GET' request handlers")
 
-    def _get_response_from_data(
+    def get_response_from_data(
```
peterschutt, line 286: Given that `HTTPRouteHandler` is part of our public api, I'd prefer to leave this as non-public implementation detail until there was a good reason to expose them.

Same would apply to `resolve_after_response()` and `_process_after_request_hook()`. 

Just my opinion, but is there a reason that a user should ever need to call these directly?

peterschutt (approved):
A couple of comments, nothing serious but I do question why we need to remove the demarcation of those methods on `HTTPRouteHandler` as private implementation detail.

I notice you like to do housekeeping changes in your PRs - i.e., ordering of parameters, method naming and other things that aren't directly related to the PR feature. All good, but they do add noise that makes reviewing a bit harder. Perhaps we should draw up a set of standards that you want to uphold for things like that, do a pass over the code base to clean them all up in one hit, and then make sure we hold new PRs to those standards so that we don't need to clean up after them in later PRs.

starlite/handlers/http.py
cofin: I didn't have a lot of time when reviewing, but I had several mypy errors when I quickly tested this.

starlite/handlers/http.py
cofin: It's probably better to keep the public API as simple as possible.  I can't think of a reason for a user to call these at the moment, but I'm good either way we land on this one.

cofin (approved):
The code looks great to me, and I was able to test it locally without issue.

starlite/handlers/http.py
peterschutt: I get 1:

```python
starlite/handlers/http.py:247: error: Incompatible types in assignment (expression has type "staticmethod[Optional[Awaitable[None]]]", variable has type "Union[Callable[[Request[Any, Any]], None], Callable[[Request[Any, Any]], Awaitable[None]], None, Type[empty]]")  [assignment]
```
Which is a result of the type that we declare for the attribute in `__init__()`. Just to be clear, I don't think this is a big deal, but just asking the question as `staticmethod()` is kinda the front-door to avoiding implicit binding of functions to class instances, from [the docs](https://docs.python.org/3/library/functions.html#staticmethod):

> This is needed in some cases where you need a reference to a function from a class body and you want to avoid the automatic transformation to instance method

docs/usage/13-lifecycle-hooks.md
```diff
@@ -0,0 +1,90 @@
+# Life Cycle Hooks
+
+Life cycle hooks allows a user to execute a function at a certain point, as indicated by the hook's name, during the
+request-response cycle.
+
+## Before Request
+
+The `before_request` hook runs immediately before calling the route handler function. It accepts either a sync or async
+function that receives the `starlite.connection.Request` instance as its sole parameter. While the handler function
+doeds not need to return a value, if it does return a value other than `None`, then the route handler will not be called
```
yocalebo, line 10: Just someone monitoring this project from a distance but seem to have a minor typo here doeds -> does

starlite/handlers/http.py
Goldziher: fixed

docs/usage/13-lifecycle-hooks.md
Goldziher: fixed

starlite/handlers/http.py
Goldziher: So static method causes test failures. You are welcome to look into it if and when you have the time.

peterschutt: yep that's cool - it would prob. be better to use everywhere or nowhere for consistencies sake anyway :+1:

Goldziher:
> A couple of comments, nothing serious but I do question why we need to remove the demarcation of those methods on `HTTPRouteHandler` as private implementation detail.
> 
> I notice you like to do housekeeping changes in your PRs - i.e., ordering of parameters, method naming and other things that aren't directly related to the PR feature. All good, but they do add noise that makes reviewing a bit harder. Perhaps we should draw up a set of standards that you want to uphold for things like that, do a pass over the code base to clean them all up in one hit, and then make sure we hold new PRs to those standards so that we don't need to clean up after them in later PRs.

Sorry if the house-keeping causes too much noise. My collegues also dont like my habits in this regard TBH. This isn't about enforcing a rule - when some code grows, as the number of arguments in the route handlers, it occasionally needs to be cleaned up and handled. Thats whats happening here in particular. And yes, I follow the scout's dictum of leaving the camp ground cleaner than when you found it.

peterschutt:
> > A couple of comments, nothing serious but I do question why we need to remove the demarcation of those methods on `HTTPRouteHandler` as private implementation detail.
> > I notice you like to do housekeeping changes in your PRs - i.e., ordering of parameters, method naming and other things that aren't directly related to the PR feature. All good, but they do add noise that makes reviewing a bit harder. Perhaps we should draw up a set of standards that you want to uphold for things like that, do a pass over the code base to clean them all up in one hit, and then make sure we hold new PRs to those standards so that we don't need to clean up after them in later PRs.
> 
> Sorry if the house-keeping causes too much noise. My collegues also dont like my habits in this regard TBH. This isn't about enforcing a rule - when some code grows, as the number of arguments in the route handlers, it occasionally needs to be cleaned up and handled. Thats whats happening here in particular. And yes, I follow the scout's dictum of leaving the camp ground cleaner than when you found it.

Not too much noise - some noise :wink:

Issue #188, opened 2022-06-28
Bug: exception representation swallows errors


Bobronium:
From https://github.com/starlite-api/starlite/pull/174#issuecomment-1166689792:

> Also a shame that pytest seems to be stripping out the error message pointing to the extra install requirement.

Turns out, It's not pytest, it's actually starlite :)

`StarLiteException` eats exception args, doesn't pass them to `Exception` and doesn't define custom `__str__` 
While `__repr__` is defined, that not what python uses to display errors.

My proposal: ditch custom `__repr__` altogether and rely on `Exception.args`.
And make `detail` a property of `args` joined together (which is actually just `Exception.__str__`)

Goldziher:
Go ahead

Goldziher:
Another thing we need to take care when doing this is to support headers being passed to exception, as is done in `starlette.exceptions.HTTPException`.

@Bobronium will you be taking over this issue?

Bobronium:
@peterschutt, yep, I'm working on it. Though it may take a couple of days or even weeks, since I'm a bit more busy with life :)

I'm fighting with backwards/cross compatibility with `starlette` at the moment.

Bobronium:
Ok, so I've been working on this for couple of days, but struggling to continue.

If anyone wants to pick up on this, please let me know. 

I'm still committed to do this, but don't want to stop anyone who's able to do it quicker than me. If no one will pick this up, there's a great chance I will do it some time later.

Goldziher:
Ok, thanks for your efforts @Bobronium

PR #281, opened 2022-07-16
Issue 188: resolve exceptions swallowing args

Goldziher:
This PR addresses #188 by calling `super().__init__(...)` in the exception constructors, and handling *args and **kwargs.

### DIFF
--- a/starlite/exceptions.py
+++ b/starlite/exceptions.py
@@ -14,9 +14,9 @@ from starlette.status import (
 
 
 class StarLiteException(Exception):
-    def __init__(self, detail: str = ""):
+    def __init__(self, *args: Any, detail: str = "", **kwargs: Dict[str, Any]):
         self.detail = detail
-        super().__init__()
+        super().__init__(*args, **kwargs)
 
     def __repr__(self) -> str:
         if self.detail:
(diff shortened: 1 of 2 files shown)


Goldziher:
@Bobronium perhaps you want to review?

starlite/exceptions.py
```diff
@@ -14,9 +14,9 @@
 
 
 class StarLiteException(Exception):
-    def __init__(self, detail: str = ""):
+    def __init__(self, *args: Any, detail: str = "", **kwargs: Dict[str, Any]):
         self.detail = detail
-        super().__init__()
+        super().__init__(*args, **kwargs)
 
     def __repr__(self) -> str:
```
Bobronium, line 21: Can we get rid of custom repr as well? `Exception.__repr__` does `StarLiteException('Message')` which is more conventional + a valid python syntax.

Goldziher: I followed here the StarletteException pattern - if we dont override it, we will be left with their override in place:


```python
class HTTPException(Exception):
    def __init__(
        self,
        status_code: int,
        detail: typing.Optional[str] = None,
        headers: typing.Optional[dict] = None,
    ) -> None:
        if detail is None:
            detail = http.HTTPStatus(status_code).phrase
        self.status_code = status_code
        self.detail = detail
        self.headers = headers

    def __repr__(self) -> str:
        class_name = self.__class__.__name__
        return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
```

This is their code

Goldziher: Waiting on your approval or further suggestions 😉

Bobronium: Which is fine? It'll only be in StarliteHTTPException. Why do we have to override it?

Goldziher: Because we inherit it

Bobronium: Sorry, I'm not following. I've updated original question, hope it does reads clearer now

Goldziher: Lemme suggest you submit a PR with some tests and if all is good, we merge it?

Bobronium: Don't have capacity for that at the moment

Goldziher: NP. So let's keep it as is until you do 😉

Bobronium: It is up to you really. Though, it's still unclear why we have to override, if we inherit it? And if we do, what you expect me to do in PR?

Goldziher: Remove repr, test it, and assert the result is satisfactory and useful

PR #288, opened 2022-07-18
Allow Partial to annotate fields of nested classes

### LINKED ISSUE #287: Partial doesn't work with inherited fields
```python
from starlite import Partial, get

class Parent(BaseModel):
    foo: int

class Child(Parent):
    bar: int

@get("/test")
def example(obj: Partial[Child]) -> None:
    print(obj)
```

In the above example, `Partial[Child]` only accepts the field `bar: Optional[int]` and ignores all fields from the superclass. I couldn't find this behaviour documented anywhere so I assume this isn't intended?

```python
Python 3.10.5 (main, Jun 23 2022, 17:14:57) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> from starlite import Partial
>>> from pydantic import BaseModel
>>> class Parent(BaseModel):
...     foo: int
...
>>> class Child(Parent):
...     bar: int
...
>>> PartialChild = Partial[Child]
>>> PartialChild.__annotations__
{'bar': typing.Optional[int]}
>>>
```

This behaviour can also be seen above

Harry-Lees:
Closes #287

### DIFF
--- a/docs/usage/11-data-transfer-objects.md
+++ b/docs/usage/11-data-transfer-objects.md
@@ -187,6 +187,57 @@ class MyClassDTO(BaseModel):
     third: str
 ```
 
+## Partial DTOs
+
+For [PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH) HTTP methods, you may only need to partially modify a resource. In these cases, DTOs can be wrapped with Partial. `Partial` can only be used on pydantic models.
+
+```python
+from pydantic import BaseModel
+from starlite.types import Partial
+
+
+class CompanyDTO(BaseModel):
+    id: int
+    name: str
+    worth: float
+
+
+PartialCompanyDTO = Partial[CompanyDTO]
+```
+
+The created `PartialCompanyDTO` is equivalent to the following declaration:
+
+```python
+from pydantic import BaseModel
+
+
+class PartialCompanyDTO(BaseModel):
+    id: Optional[int]
+    name: Optional[str]
+    worth: Optional[float]
+```
+
+`Partial` can also be used inline when creating routes.
+
+```python
+from pydantic import UUID4
+from starlite.controller import Controller
+from starlite.handlers import patch
+from starlite.types import Partial
+
+from my_app.orders.models import UserOrder
+
+
+class UserOrderController(Controller):
+    path = "/user"
+
+    @patch(path="/{order_id:uuid}")
+    async def update_user_order(
+        self, order_id: UUID4, data: Partial[UserOrder]
+    ) -> UserOrder:
+        ...
+```
+
 ## DTO Methods
 
 ### from_model_instance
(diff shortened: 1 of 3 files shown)


Harry-Lees:
Since this PR changes the behaviour of Partial to return `{}` on all classes that don't inherit from BaseModel, do you think it would be sensible to just raise an Exception from Partial if a class is passed in that doesn't inherit from BaseModel.

I think it may help prevent some headaches down the line for anyone who tries and can't figure out why Partial is just returning `{}` and sort of suppressing an error? If they're using a type checker (I assume most people using this library would be?!) then the checker will catch it for them since the generic type, T, that Partial inherits from is bound to BaseModel so I'm not sure what your opinions on this are.

Goldziher:
> Since this PR changes the behaviour of Partial to return `{}` on all classes that don't inherit from BaseModel, do you think it would be sensible to just raise an Exception from Partial if a class is passed in that doesn't inherit from BaseModel.
> 
> I think it may help prevent some headaches down the line for anyone who tries and can't figure out why Partial is just returning `{}` and sort of suppressing an error? If they're using a type checker (I assume most people using this library would be?!) then the checker will catch it for them since the generic type, T, that Partial inherits from is bound to BaseModel so I'm not sure what your opinions on this are.

Yes, an exception makes sense tbh. You can use the `ImproperlyConfiguredException` for this. Can you also update the docs regarding partial to make this clear?

Harry-Lees:
> > Since this PR changes the behaviour of Partial to return `{}` on all classes that don't inherit from BaseModel, do you think it would be sensible to just raise an Exception from Partial if a class is passed in that doesn't inherit from BaseModel.
> > I think it may help prevent some headaches down the line for anyone who tries and can't figure out why Partial is just returning `{}` and sort of suppressing an error? If they're using a type checker (I assume most people using this library would be?!) then the checker will catch it for them since the generic type, T, that Partial inherits from is bound to BaseModel so I'm not sure what your opinions on this are.
> 
> Yes, an exception makes sense tbh. You can use the `ImproperlyConfiguredException` for this. Can you also update the docs regarding partial to make this clear?

Where should I put the documentation for this? I looked through the docs and I couldn't find any API reference for Partial, just lots of examples :)

peterschutt:
> Where should I put the documentation for this? I looked through the docs and I couldn't find any API reference for Partial, just lots of examples :)

I can speak for this. For API docs do them in the class/func/method docstrings using google style. Just describe the ins and outs - and only really needed to do for parts of the public interface. For internal interfaces, just a good docstring description of what is going on is enough. I've got a task there to add the api docs to the pages site, so it would be good to have if you've added to the public interface anywhere, or want to pad existing stuff out.

As for the prose sections, if there isn't one already and it warrants it then feel free to add a section.. or maybe you can use a note section somewhere around the examples if it doesn't warrant a full section, such as:
https://github.com/starlite-api/starlite/blob/a2c8a86d42fb347133a5c0c278ae9bc950c75cb7/docs/usage/1-routers-and-controllers.md?plain=1#L46-L49
 Note the prettier ignore bit if you go that route.

Goldziher:
@all-contributors add @Harry-Lees code doc

PR #292, opened 2022-07-19
Split `openapi.create_paramter_for_handler` into several methods to reduce complexity

185504a9:
once again a simple refactor which was mentioned as needed by https://github.com/starlite-api/starlite/issues/203 and fixes [this](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&id=starlite-api_starlite&open=AYH3kLXqoMi-EPUvUENi)

the first filter, which extracts params that are `not in layered_parameters`, had its logic moved to a method called `_get_unlayered_params` but i'm not sure if it's the best naming decision

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ dist/
 node_modules/
 results/
 site/
+.scannerwork/
 
 # files
 *.iml
--- a/starlite/openapi/parameters.py
+++ b/starlite/openapi/parameters.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from pydantic.fields import ModelField
 
     from starlite.handlers import BaseRouteHandler
+    from starlite.provide import Provide
 
 
 def create_path_parameter_schema(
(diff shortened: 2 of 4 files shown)


peterschutt:
Thanks heaps for this! I'll have a look sometime in the next few hours. Cheers.

starlite/openapi/parameters.py
```diff
@@ -115,6 +116,64 @@ def create_parameter(
     )
 
 
+def _get_unlayered_params(
```
peterschutt, line 119: `get_recursive_handler_parameter()`?

We should use the full word `parameter` in these names to match the other function names in the module.

Also, don't worry about making these private functions as we wouldn't consider the rest of this module public interface either so no need to discriminate these.

starlite/openapi/parameters.py
```diff
@@ -164,35 +218,14 @@ def create_parameter_for_handler(
     for field_name, signature_model_filed in filter(
```
peterschutt, line 218: There's a typo here, `signature_model_filed`, would you mind fixing?

peterschutt (changes requested):
Thanks for the refactor :)

starlite/openapi/parameters.py
185504a9: >Also, don't worry about making these private functions as we wouldn't consider the rest of this module public interface either so no need to discriminate these.

it's more of a personal thing that i do without thinking since i usually name all utility functions that way in order to visually separate them. i'll make them public

starlite/openapi/parameters.py
```diff
@@ -115,6 +116,70 @@ def create_parameter(
...
+    field_name: str,
+    model_field: "ModelField",
+    dependencies: Dict[str, "Provide"],
+    route_handler: "BaseRouteHandler",
+    path_parameters: List[Dict[str, Any]],
+    generate_examples: bool,
+) -> List[Parameter]:
+    """Create and return parameters for a handler. If the provided field is not a dependency,
+    a normal parameter is created and returned as a list, otherwise a recursive call
+    may be triggered by the parent call function"""
```
peterschutt, line 129: > otherwise a recursive call may be triggered by the parent call function

I think this is a bit misleading as it is this function that triggers the recursive call. How about:
```
"""... otherwise `create_parameter_for_handler()` is called to generate parameters for the dependency."""
```
Something along those lines?

peterschutt (changes requested):
Just one comment about the doc strings and then this is good to go from my point of view.

starlite/openapi/parameters.py
185504a9: you're absolutely right

starlite/openapi/parameters.py
```diff
@@ -115,6 +116,71 @@ def create_parameter(
...
+    signature_field_info = signature_model_field.field_info
+
+    field_info = layer_field_info
+    # allow users to manually override Parameter definition using Parameter
+    if signature_field_info.extra.get(EXTRA_KEY_IS_PARAMETER):
+        field_info = signature_field_info
+
+    field_info.default = (
+        signature_field_info.default
+        if signature_field_info.default not in [Undefined, Ellipsis]
```
Goldziher, line 164: Please move `[Undefined, Ellipsis]` into a variable in global scope, and change from list to set to increase speed

Goldziher (approved):
Looks great. Please address the two comments in place. I will leave this PR to @peterschutt to merge when you're done.

.gitignore
```diff
@@ -11,6 +11,7 @@ dist/
 node_modules/
 results/
 site/
+.scannerwork/
```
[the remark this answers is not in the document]
peterschutt: It's from running sonar scanner locally I believe.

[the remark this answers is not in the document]
Goldziher: Oh. I just use the sonarlint plugin - it allows one to specify remote configs as well, so it can be directly linked to the starlite sonar org

starlite/openapi/parameters.py
185504a9: sure but i won't be able to resolve this until next month. if you guys want this sooner i won't mind to have it resolved by someone else

Issue #180, opened 2022-06-26
Enhancement: OpenTelemetry integration


Goldziher:
So, as the title says. We need to consider how to best support OpenTelemetry. We might need to add some hooks to allow for optimal instrumentation. We should also consider creating an OT instrumentationn library, see for example: https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/fastapi/fastapi.html

Goldziher:
@cofin do you want to take this one?

cofin:
Yep.  I'll take a look at it over the weekend.

cofin:
Un-assigning for now until we dial in the instrumentation enhancements we've been discussing.

Goldziher:
@cofin - are u doing this?

cofin:
I had planned to, but I don't have the bandwidth currently. I can try to pick it back up soon or hand off.

Goldziher:
It would be good if you could do this when you have bandwidth. We need this - and I too am somewhat busy 😉

cofin:
Understood. I need it soon as well, so I'll make some time for it. I'll ping you on discord when I'm getting started.

Goldziher:
@seladb you interested in picking this one perhaps?

seladb:
Sure @Goldziher I think I can take it! Might take me some time though because I'm busy with other stuff at the moment

seladb:
@Goldziher @cofin I looked up the OT integration with FastAPI and here is what I understand:
- The instrumentation class is implemented in the OT repo (and **NOT** in the FastAPI repo): 
  - https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py
- The way it's used is outside of the FastAPI app:
   ```python
    import fastapi
    from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor

    app = fastapi.FastAPI()

    @app.get("/foobar")
    async def foobar():
        return {"message": "hello world"}

    FastAPIInstrumentor.instrument_app(app)
    ```
- What the instrumentor does is add a middleware to the app: 
  - https://github.com/open-telemetry/opentelemetry-python-contrib/blob/744851b3cf3a58471e44c3a20f143275fa16cdce/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L184 
[...]

PR #293, opened 2022-07-20
Enhancement: OpenTelemetry integration #180

cofin:
This enhancement enabled Open Telemetry integration.  It is set up very similar to the Compression backend. 

To use, set `instrumentation_config=InstrumentationConfig( backend='opentelemetry', excluded_urls="/test_path"  ),` in the Starlite application config.  

To do:
- Additional test cases
- Pass pre-commit pylint checks.  Current there is an `import-error` getting thrown.
- Verify metrics can be collected from third party utility
- Documentation


starlite/middleware/instrumentation/opentelemetry.py
```diff
@@ -0,0 +1,74 @@
...
+            excluded_urls=excluded_urls,
+            server_request_hook=server_request_hook,
+            client_request_hook=client_request_hook,
+            client_response_hook=client_response_hook,
+            tracer_provider=tracer_provider,
+            default_span_details=_get_route_details,
+        )
+
+
+def _get_route_details(
```
peterschutt, line 49: Any thoughts on a work around? Could we use `scope["state"]`?

peterschutt:
What do you use for ingestion? I might have a crack at getting the template app feeding into something like https://signoz.io/

cofin:
> 

Signoz is the one that I see often.  I think it's specifically for APM.  I'd also like to try prometheus.

cofin:
I'm going to close this and re-work the implementation.  I'll open a new PR once it's ready.

seladb:
@cofin is there anything fundamentally wrong in the implementation that is in the closed PR? It looks quite ok to me... 

Just wondering if you had a completely different idea on how to implement this

PR #297, opened 2022-07-21
Adds `exceptions.utils.create_exception_response()`.

### LINKED ISSUE #294: Make default exception handler response creation logic available from outside ExceptionHandlingMiddleware`
As far as I can see, Starlite does not log the exceptions, the only thing I see is:

```
INFO:     127.0.0.1:38622 - "GET / HTTP/1.1" 500 Internal Server Error
```

I would like that Starlite prints out the exceptions behind and their trace.

Thanks.

peterschutt:
- makes `starlite.exceptions` a sub-package with same api as module it replaces.
- adds `starlite.exceptions.utils` module to house the response generation helper function.

Closes #294


peterschutt:
I chose to expose the utility under `starlite.exceptions` as that is already public api, whereas `starlite.utils` isn't - hence the conversion of `starlite.exceptions` to a package. LMK thoughts on that.

Also a slight behavioral change in that the status code is always included in the response body, previously it was omitted if the exception was not a HTTPException.

starlite/exceptions/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+from .exceptions import HTTPException
+
+__all__ = ["create_exception_response"]
+
+
+class ExceptionResponseContent(BaseModel):
+    detail: Optional[str]
+    extra: Optional[Union[Dict[str, Any], List[Any]]]
+    status_code: int
```
Goldziher, line 18: You can set the default here to 500

starlite/exceptions/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+from starlite.enums import MediaType
+from starlite.response import Response
+
+from .exceptions import HTTPException
+
+__all__ = ["create_exception_response"]
+
+
+class ExceptionResponseContent(BaseModel):
```
Goldziher, line 15: Nice!

Goldziher (approved):

starlite/exceptions/utils.py
peterschutt: Yes good idea, however I had to change the `to_dict()` call from `exclude_defaults=True` to `exclude_none=True`, which adds another slight change in behavior to the exception responses. If anyone is doing `if error_response['extra"] is None` for example, this would be breaking... but "extra" wasn't always in the exception response anyway so I doubt this would be common.

PR #299, opened 2022-07-22
`orjson` use in `WebSocket.{receive,send}_json()`

peterschutt:
Implementations are functionally the same as upstream except for orjson usage and we raise our own exceptions.


Goldziher (approved):
Merge when ready

Issue #300, opened 2022-07-22
Enhancement: Add OpenAPI 3.0.3 support


LonelyVikingMichael:
This would enable support for interactive Swagger documentation.

The downside is that this will become redundant once Swagger UI adopts OpenAPI 3.1.0.

timwedde:
As far as I have experienced, the only thing standing between us supporting Swagger in Starlite is that Starlite (correctly) marks the schema as 3.1. SwaggerUI explicitly blocks this because they want to fully support the JSON schema in 3.1 first before allowing this, but if you have a 3.1 YAML schema, it already works perfectly fine with the JS bundle you can get right now.

I've implemented a controller to emulate the FastAPI behavior like this:
```py
# -*- coding: utf-8 -*-
import orjson as json
from typing import Optional, Any
from starlite.datastructures import State
from starlite import Controller, Request, MediaType, get


class SwaggerController(Controller):
    """
    This is an internal controller that exists to provide the /docs route
    that is standard in FastAPI. Starlite only provides ReDoc by default, so
    we patch it in like this as a convenience feature.
    """

    path = "/docs"
    tags = ["docs"]

    def get_swagger_ui_html(
        self,
        openapi_url: str,
        title: str,
        swagger_js_url: str = "https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui-bundle.js",
        swagger_css_url: str = "https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui.css",
        swagger_favicon_url: str = "https://fastapi.tiangolo.com/img/favicon.png",
        oauth2_redirect_url: Optional[str] = None,
        init_oauth: Optional[dict[str, Any]] = None,
        swagger_ui_parameters: Optional[dict[str, Any]] = None,
    ) -> str:
        current_swagger_ui_parameters = {
            "dom_id": "'#swagger-ui'",
            "layout": "'BaseLayout'",
            "deepLinking": "true",
            "showExtensions": "true",
            "showCommonExtensions": "true",
        }
        if swagger_ui_parameters:
            current_swagger_ui_parameters.update(swagger_ui_parameters)

        html = f"""
        <!DOCTYPE html>
        <html>
        <head>
        <link type="text/css" rel="stylesheet" href="{swagger_css_url}">
        <link rel="shortcut icon" href="{swagger_favicon_url}">
        <title>{title}</title>
        </head>
        <body>
        <div id="swagger-ui"></div>
[...]

PR #302, opened 2022-07-22
Support for SwaggerUI

timwedde:
As per #300, this adds basic support for SwaggerUI by monkey-patch-downgrading the version to one that is not instantly rejected by the default SwaggerUI bundle. This is valid because 3.1 support for SwaggerUI is largely blocked by the newly-introduced JSON schema support, which we don't use in Starlite (it's all YAML). More info in the tracking issue [here](https://github.com/swagger-api/swagger-ui/issues/5891).

Other than that I tried to stick as close as possible to the ReDoc implementation, leaving out things like extra config values or OAuth support to be added at a later time.

One thing that may be slightly weird for developers using this functionality is the route where this is currently served at `/schema/swagger`. Since `/schema` is already taken by ReDoc (and I assume we can't namespace it since that'd break compatibility) I put it there, but it's slightly different from the ReDoc convention right now. One option would be to also introduce `/schema/redoc` to at least make the routes consistent, with the default ReDoc UI being kept on the root of the controller.

### DIFF
--- a/starlite/openapi/controller.py
+++ b/starlite/openapi/controller.py
@@ -1,3 +1,4 @@
+import copy
 from typing import TYPE_CHECKING
 
 from orjson import OPT_INDENT_2, dumps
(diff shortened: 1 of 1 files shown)


starlite/openapi/controller.py
```diff
@@ -37,6 +38,60 @@ def retrieve_schema_json(self, request: Request) -> "OpenAPI":
...
+    def swagger_ui(self, request: Request) -> str:
+        """Endpoint that serves SwaggerUI"""
+        schema = self.schema_from_request(request)
+        if self.dumped_schema == "":
+            # Note: Fix for Swagger rejection OpenAPI >=3.1
+            # We force the version to be lower to get the default JS bundle to accept it
+            # This works flawlessly as the main blocker for Swagger support for OpenAPI 3.1 is JSON schema support
+            # Since we use the YAML format this is not an issue for us and we can do this trick to get support right now
+            # We use deepcopy to avoid changing the actual schema on the request. Since this is a cached call the effect is minimal
+            schema = copy.deepcopy(schema)
```
Goldziher, line 51: Pydantic models have a copy method, so you can simply use it instead

Goldziher (commented):
Looks good. Thanks, we'll test this. Can you add a test ensuring the swagger endpoint is available? Equivalent to the tests we have for redoc.

starlite/openapi/controller.py
timwedde: Ah sweet, I didn't know that. I've switched it over.

timwedde:
Do you perchance have a pointer towards the existing ReDoc tests? I perused the tests folder but I didn't find any.

Goldziher:
> Do you perchance have a pointer towards the existing ReDoc tests? I perused the tests folder but I didn't find any.

Lol, seems we don't have any. In that case, can you add such sanity tests?

timwedde:
Yep, will do. I assume "available" in this context simply means "doesn't throw an exception and returns HTML and status 20X"?

Goldziher:
> Yep, will do. I assume "available" in this context simply means "doesn't throw an exception and returns HTML and status 20X"?

exactly

Goldziher:
@timwedde - i merged this into a temporary branch. Please add a tests PR as a followup. I will now test this.

PR #303, opened 2022-07-22
Support for SwaggerUI (#302)

Goldziher:
* Support for SwaggerUI

As per #300

* Use built-in copy method

I tested this and it works, great stuff @timwedde 

Please add tests asap.


timwedde:
Nice. Will push an extra PR with tests after dinner.

PR #306, opened 2022-07-23
OpenAPIConfig.use_handler_docstrings

peterschutt:
Adds ability to configure openapi schema generation to use the route handler docstring for operation description.

If `False` docstrings are ignored, this is default behavior for backward compatibility.

If `True`, `HTTPRouteHandler.description` takes precedence if set, otherwise the docstring is used.

### DIFF
--- a/starlite/config.py
+++ b/starlite/config.py
@@ -9,9 +9,11 @@ from typing import (
     Tuple,
     Type,
     Union,
+    cast,
 )
 from urllib.parse import urlencode
 
+from openapi_schema_pydantic.util import construct_open_api_with_schema_class
 from openapi_schema_pydantic.v3.v3_1_0.contact import Contact
 from openapi_schema_pydantic.v3.v3_1_0.external_documentation import (
     ExternalDocumentation,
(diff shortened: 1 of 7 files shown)


peterschutt:
Moved `Starlite.create_openapi_schema_model` to `OpenAPIConfig.create_openapi_schema_model()` as the coupling between `Starlite` and that method was quite low and I think it is a better separation of concerns. However, that isn't critical to the PR so I can easily reduce the scope of the change if that is preferred.

cofin (approved):
Looks good!

starlite/config.py
```diff
@@ -177,6 +183,32 @@ def to_openapi_schema(self) -> OpenAPI:
             ),
         )
 
+    def create_openapi_schema_model(self, router: "Starlite") -> OpenAPI:
```
Goldziher, line 186: Shouldn't this be typed as Router?

Goldziher (approved):

starlite/config.py
peterschutt: I named the parameter `router` with [this](https://github.com/starlite-api/starlite/issues/241#issuecomment-1191976760) in mind. However, the method accesses `router.plugins` which doesn't exist on `Router`, only `Starlite`.

So what do you think about https://github.com/starlite-api/starlite/issues/241#issuecomment-1191976760?

If that's a non-starter, then I'll just name the parameter to this method `app` and leave the type as is. If it is a starter, eventually this would be typed `Router`.

Goldziher: Then I'd rename it to app

peterschutt: Are you saying that idea is a non-starter?

PR #310, opened 2022-07-26
Fix issues with Rust route map bindings

nramos0:
Fixes some issues with the current state of the Rust bindings. Testing can now be run as described here: https://github.com/starlite-api/starlite/pull/270#issuecomment-1190713507, but with a small change:
What was previously (under my own system directories)
```
env PYTHONPATH="/Users/haxi/Documents/code/starlite/nramos-starlite/.venv/lib/python3.10/site-packages" cargo test --no-default-features
```
should be changed to 
```
env PYTHONPATH="/Users/haxi/Documents/code/starlite/nramos-starlite/.venv/lib/python3.10/site-packages" cargo test --no-default-features -F auto-initialize
```

PyO3's auto-initialize feature also needs to be behind a feature flag in our own crate that is off by default, otherwise building it for those that don't want to run tests (ex. normal Starlite users) won't be able to install correctly without Python shared libraries.

### DIFF
--- a/extensions/rust/Cargo.toml
+++ b/extensions/rust/Cargo.toml
@@ -8,8 +8,9 @@ name = "route_map_rs"
 crate-type = ["cdylib"]
 
 [dependencies]
-pyo3 = { version = "0.16.5", features = ["auto-initialize"] }
+pyo3 = { version = "0.16.5" }
 
 [features]
 extension-module = ["pyo3/extension-module"]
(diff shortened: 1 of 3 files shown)


nramos0:
It still remains to make tests more direct and easier to run, without all of the complications described in the original PR linked above, I will work on doing this within the next several days

Goldziher:
@all-contributors add @nramos0 code

PR #312, opened 2022-07-28
Misc rust fixes/changes

Dr-Emann:
* Use a `.pyi` file to expose the types of the rust extensions module directly
* Remove the unused duplicate code at `starlite/extensions/rust` (the real code is at `extensions/rust`
* Rename the rust extension module to just `starlite.route_map` rather than `starlite.route_map_rs`.
* Replace some macros with less magical options

Let me know if we want to back out any of these (e.g. I'm most unsure of the rename: are we planning to try to make the rust part optional?)


extensions/rust/src/route_map.rs
```diff
@@ -290,14 +292,12 @@ impl<'rm> ConfigureNodeView<'rm> {
...
-                let route_handler = route.getattr("route_handler")?;
-                let middleware_stack =
-                    build_route_middleware_stack(py, &ctx, route, route_handler)?;
-                asgi_handlers.insert($handler_type.to_string(), middleware_stack.to_object(py));
-            };
-        }
+        let mut generate_single_route_handler_stack = |handler_type: &str| -> PyResult<()> {
+            let route_handler = route.getattr("route_handler")?;
+            let middleware_stack = build_route_middleware_stack(py, &ctx, route, route_handler)?;
+            asgi_handlers.insert(handler_type.to_string(), middleware_stack.into());
```
nramos0, line 298: I've just discovered that the `.into()` here is not necessary. PyObject is just an alias for Py<PyAny>, which is returned by `build_route_middleware_stack`. This also means that the call to `middleware_stack.to_object(py)` that I wrote before is also unnecessary, that one can be changed too

extensions/rust/src/route_map.rs
```diff
@@ -311,9 +311,9 @@ impl<'rm> ConfigureNodeView<'rm> {
                 asgi_handlers.insert(method, middleware_stack.to_object(py));
```
nramos0, line 311: This call to `to_object()` can be removed as well

starlite/route_map.pyi
```diff
@@ -1,43 +1,36 @@
-# A proxy for a Rust extension class that implements the interface below
-# The Rust implementation is located at ./starlite/extensions/rust
```
nramos0, line 2: Maybe this comment should be added back (with the correct path of `./extensions/rust`) as an indicator to other people reading the code not aware that this file is type hinting for a Rust extension

nramos0 (changes requested):
Everything pretty much looks good to me, just a couple small things

PR #313, opened 2022-07-28
PoC: Cython 

Goldziher:
This PR is a PoC for Cython compilation. 

To test it locally clone, `poetry install` and then run `poetry build`. It will create a dist folder with the compiled wheel, and leave *.so files in the `starlite` folder. You can then run the test suite - which will pass even after deleting the *.py files that have been compiled. 

We need to test and see if there are significant performance gains from using Cython here. Note the following:

1. Cython cant handle classes with generics
2. Cython is not compatible with the pydantic `validate_arguments` decorator

Whats required to make this something that we can actually build upon?

1. update of CI to build wheels for all required targets and testing of these.
2. update of the release scripts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ site/
 .DS_Store
 .python-version
 coverage.*
+**/*.so
--- a/.gitignore
+++ b/.gitignore
@@ -3,20 +3,22 @@
 .idea/
 .mypy_cache/
 .pytest_cache/
+.scannerwork/
 .tox/
 .venv/
 .vscode/
 __pycache__/
+build/
 dist/
 node_modules/
 results/
 site/
-.scannerwork/
 
 # files
+**/*.c
+**/*.so
 *.iml
-.coverage
 .DS_Store
+.coverage
 .python-version
 coverage.*
-**/*.so
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
         exclude: "\\.idea/(.)*"
       - id: trailing-whitespace
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.37.2
+    rev: v2.37.3
     hooks:
       - id: pyupgrade
         args: ["--py37-plus"]
@@ -27,6 +27,7 @@ repos:
     rev: v2.1.0
     hooks:
       - id: codespell
+        args: [-L, crate]
   - repo: https://github.com/asottile/blacken-docs
     rev: v1.12.1
     hooks:
@@ -41,11 +42,11 @@ repos:
       - id: bandit
         exclude: "test_*"
         args: ["-iii", "-ll", "-s=B308,B703"]
-  - repo: https://github.com/hadialqattan/pycln
-    rev: v2.0.4 # Possible releases: https://github.com/hadialqattan/pycln/releases
-    hooks:
-      - id: pycln
-        args: [--config=pyproject.toml]
+  #  - repo: https://github.com/hadialqattan/pycln
+  #    rev: v2.0.4 # Possible releases: https://github.com/hadialqattan/pycln/releases
+  #    hooks:
+  #      - id: pycln
+  #        args: [--config=pyproject.toml]
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.9.2
     hooks:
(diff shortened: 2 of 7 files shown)


peterschutt:
Following the instructions above work for me (ubuntu 20.04).

Pointing a project at the cython branch for install doesn't seem to build the extensions as part of the install however. Should it?

What sort of speedup would you be looking for to deem this worthwhile?

Goldziher:
To install it you have to have cython installed as well, before it. We need to see what speed ups are there if any TBH.

peterschutt:
> To install it you have to have cython installed as well, before it.

Ah yep, that's right... I just had a look at one of my old projects that uses cython and I have this in the `pyproject.toml`:

```toml
[build-system]
requires = ["poetry-core>=1.0.0a9", "setuptools~=58.2", "numpy~=1.21", "cython~=0.29"]
build-backend = "poetry.core.masonry.api"
```

... been a long time since I put that together, but I just tested installing the lib into a fresh environment direct from github via poetry and it does build the extension without leaving cython around.

> We need to see what speed ups are there if any TBH.

Without targeted optimizations we can only expect them to be modest, I think. However, if we already have cython in the build chain we can go to town on hot spots in `.pyx` files, but you end up with stuff like this:

```python
@cython.boundscheck(False) # turn off bounds-checking for entire function
@cython.wraparound(False)  # turn off negative index wrapping for entire function
cpdef racesimulator(np.ndarray[int64_t, ndim=2, mode='c'] all_nums_np,
                    np.ndarray[double, ndim=2, mode='c'] all_cdfs_np,
                    uint sims=1000, uint placings=3):
    if all_nums_np.size != all_cdfs_np.size:
        raise ValueError(
            'Nums, CDFs and jockeys input arrays must be of equal size'
        )
    # static type vars
    cdef uint found, place_ix, place_ix2, runner_ix, race_ix, sim_ix = 0
    ...

    # initialise memory views
    cdef int64_t [:,:] all_nums = all_nums_np
    ...

    with nogil:
        for sim_ix in range(sims):
            for race_ix in range(num_races):
                ...
```
Once you get that stuff to the point where you can drop the gil, it's blazing fast, but you lose all the high-level niceties of python, and given that we are establishing a precedent for targeted optimizations with rust, I doubt we'd ever go to these lengths via cython anyway. 

So from my point of view, if we don't lose too much by introducing cython, 0 downstream impact and there is significant speed up, then I'm on board. However, if any of those things don't hold I'd be hesitant.

Goldziher:
> > To install it you have to have cython installed as well, before it.
> 
> Ah yep, that's right... I just had a look at one of my old projects that uses cython and I have this in the `pyproject.toml`:
> 
> ```toml
> [build-system]
> requires = ["poetry-core>=1.0.0a9", "setuptools~=58.2", "numpy~=1.21", "cython~=0.29"]
> build-backend = "poetry.core.masonry.api"
> ```
> 
> ... been a long time since I put that together, but I just tested installing the lib into a fresh environment direct from github via poetry and it does build the extension without leaving cython around.
> 
> > We need to see what speed ups are there if any TBH.
> 
> Without targeted optimizations we can only expect them to be modest, I think. However, if we already have cython in the build chain we can go to town on hot spots in `.pyx` files, but you end up with stuff like this:
> 
> ```python
> @cython.boundscheck(False) # turn off bounds-checking for entire function
> @cython.wraparound(False)  # turn off negative index wrapping for entire function
> cpdef racesimulator(np.ndarray[int64_t, ndim=2, mode='c'] all_nums_np,
>                     np.ndarray[double, ndim=2, mode='c'] all_cdfs_np,
>                     uint sims=1000, uint placings=3):
>     if all_nums_np.size != all_cdfs_np.size:
>         raise ValueError(
>             'Nums, CDFs and jockeys input arrays must be of equal size'
>         )
>     # static type vars
>     cdef uint found, place_ix, place_ix2, runner_ix, race_ix, sim_ix = 0
>     ...
> 
>     # initialise memory views
>     cdef int64_t [:,:] all_nums = all_nums_np
>     ...
> 
>     with nogil:
>         for sim_ix in range(sims):
>             for race_ix in range(num_races):
>                 ...
> ```
> 
> Once you get that stuff to the point where you can drop the gil, it's blazing fast, but you lose all the high-level niceties of python, and given that we are establishing a precedent for targeted optimizations with rust, I doubt we'd ever go to these lengths via cython anyway.
> 
> So from my point of view, if we don't lose too much by introducing cython, 0 downstream impact and there is significant speed up, then I'm on board. However, if any of those things don't hold I'd be hesitant.

i agree. My benchmarking so far showed its not really giving us a major boost. But i havent really delved into compiling a lot of the code etc., checkout build.py and see whats what.

PR #315, opened 2022-07-29
Enhancement: Rust bindings

Goldziher:


Goldziher:
@Dr-Emann so what do you say  - is there a benefit to using this, or should we close this and stick with pure python?

PR #317, opened 2022-07-29
Rust Bindings: CI

Goldziher:


Cargo.toml
```diff
@@ -9,7 +9,7 @@ crate-type = ["cdylib"]
 path = "extensions/route_map/lib.rs"
 
 [dependencies]
-pyo3 = { version = "0.16.5" }
+pyo3 = { version = "*" }
```
Dr-Emann, line 12: This tends to be pretty frowned upon in rust, it's usually better to specify a version of dependencies.

.github/workflows/ci.yaml
```diff
@@ -41,13 +51,23 @@ jobs:
...
       - name: Test
         run: poetry run pytest
         if: matrix.python-version != '3.10'
       - name: Test with Coverage
         run: poetry run pytest --cov=. --cov-report=xml
         if: matrix.python-version == '3.10'
+      - uses: actions-rs/toolchain@v1
+        name: Install latest nightly
+        with:
+          toolchain: stable
```
Dr-Emann, line 64: This name doesn't seem to match up with the toolchain we actually install

PR #318, opened 2022-07-30
Rust Bindings: Start of unit tests

Dr-Emann:
I got a first unit test up and working, I figured this would be a good starting point to build off.


extensions/route_map/route_map.rs
```diff
@@ -475,3 +479,96 @@ impl RouteMap {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Goldziher, line 484: Question, why do you place the tests in the same file?

I would regard this as a code smell. 

Another thing to consider is that this file is already quite large and it should probably be split regardless.

vrslev: https://doc.rust-lang.org/stable/rust-by-example/testing.html

Dr-Emann: It's not uncommon in rust to have the unit tests in the same file.

It's most important that it's in a submodule, though, so the fields, etc don't need to be made public.

I've moved the contents of the `route_map` module into the root of the crate (which is named `route_map`), removed all the public fields/methods (which ended up finding an unused method). Then I moved this test into the `test` module

PR #324, opened 2022-08-01
Fix static subpaths

Dr-Emann:
* Prevent adding routes which are below a configured static path
* Avoid replacing multiple strings when stripping the static prefix

See commit messages and added tests for more info


Goldziher (approved):
Great, merge when green.

PR #325, opened 2022-08-01
Remove all trailing slashes when normalizing paths

Dr-Emann:
`normalize_path` is supposed to make sure the path does not end in a
slash, but it only removes the last character if it is a slash.

Additionally, replace all strings of 2+ slashes with a single slash in
one pass, rather than replacing groups of two iteratively

### DIFF
--- a/tests/utils/test_url.py
+++ b/tests/utils/test_url.py
@@ -12,16 +12,28 @@ from starlite.utils.url import join_paths, normalize_path
         ("/path/", "sub/", "/path/sub"),
         ("path/", "sub/", "/path/sub"),
         ("path", "sub/", "/path/sub"),
+        ("/", "/root/sub", "/root/sub"),
     ],
 )
 def test_join_url_fragments(base: str, fragment: str, expected: str) -> None:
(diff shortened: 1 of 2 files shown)


tests/utils/test_url.py
```diff
@@ -18,6 +18,17 @@ def test_join_url_fragments(base: str, fragment: str, expected: str) -> None:
     assert join_paths([base, fragment]) == expected
 
 
-@pytest.mark.parametrize("base,expected", [("/path", "/path"), ("path/", "/path"), ("path", "/path")])
+@pytest.mark.parametrize(
+    "base,expected",
+    [
+        ("", ""),
```
Goldziher, line 24: Shouldn't this be `/`?

Dr-Emann: I would have thought so, but quite a few tests break if it is.

Dr-Emann: Fixed by changing the `join_paths` function to always return normalized paths

Goldziher (approved):

PR #327, opened 2022-08-02
Rust type safe refactor

Dr-Emann:
A pretty major refactor, but I wanted to try to make a try at making things safer (no components/children to keep in sync, less stringly typed, and avoiding e.g. `*` possibly being either a real segment or a placeholder.


extensions/route_map/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+                    asgi_handlers: rhs_handlers,
+                },
+            ) => {
+                if !lhs_params.as_ref(py).eq(rhs_params.as_ref(py))? {
+                    return Err(wrappers::ImproperlyConfiguredException::new_err(format!(
+                        "Should not use routes with conflicting path parameters at {path}"
+                    )));
+                }
+
+                for (ty, handler) in rhs_handlers {
```
nramos0, line 116: Consider renaming `ty` to `handler_type`?

extensions/route_map/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+            path_parameters,
+            is_static,
+            new_handler_group,
+        )?;
+
+        Ok(())
+    }
+
+    /// Given a scope, retrieves the correct ASGI App for the route
+    fn resolve_asgi_app(&self, py: Python<'_>, scope: wrappers::Scope) -> PyResult<Py<PyAny>> {
```
nramos0, line 303: the `<'_>` here can be elided as well

extensions/route_map/search.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub(crate) fn find_handler_group<'a>(&'a self, full_path: &'a str) -> PyResult<FindResult<'a>> {
+        let mut path = normalize_path(full_path);
+        let mut param_values = Vec::new();
+        let mut node = &self.root;
+
+        let handler_group: &HandlerGroup =
+            if let Some(handler_group) = self.plain_routes.get(path.as_ref()) {
+                handler_group
+            } else {
+                for component in get_base_components(&path) {
```
nramos0, line 61: We could consider first storing the result of `get_base_components` in a variable and using its size as a "maximum" length of `param_values`, and optimistically allocate the vector size instead by calling `param_values.reserve()`. Definitely not a certainty that we would want to do this, but something to consider. The "slow" case is when there are a lot of nested nodes, and if many of them contain path parameters, there will most likely be more than one allocation. We could prevent this by always allocating with a capacity of `get_base_component`'s return value size, but it would add a constant single allocation for cases where it was never going to allocate (when placeholder_child never exists during traversal)

extensions/route_map/test/init.rs
```diff
@@ -0,0 +1,106 @@
...
+fn init_one_deep_path() {
+    Python::with_gil(|py| {
+        let mut route_map = RouteMap::new(py, false).unwrap();
+        let mut path = vec!["a"; 50_000].join("/");
+        // Ensure path has a placeholder
+        path.insert_str(0, "/{x:str}/");
+        let route = make_route(py, &path, "get").unwrap();
+
+        route_map.add_route(py, route).unwrap();
+        drop(route_map);
```
nramos0, line 63: I would imagine we can remove the drop here?

extensions/route_map/test/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+use pyo3::prelude::*;
+use pyo3::types::PyList;
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+use std::hash::Hash;
+use std::ptr;
+
+mod init;
+
+fn node_empty(node: &Node) -> bool {
```
nramos0, line 11: Maybe move to a `Node` method

nramos0 (changes requested):
All the logic looks good to me, and I really like the type safety changes, what with the addition of the wrappers around the accesses of different attributes and creating enums for the string types we previously used. There are some changes I think we can consider making, some are just style and others clarity/performance. I could also help make the changes if that's fine with you

extensions/route_map/util.rs
```diff
@@ -1,141 +1,51 @@
...
-        for (a, b) in a.iter().zip(b.iter()) {
-            if !a.into_py_dict(py).eq(b.into_py_dict(py))? {
-                eq = false;
-                break;
-            }
-        }
-    }
-    Ok(eq)
+/// Iterator over components of a path (skipping empty components)
+pub(crate) fn get_base_components(path: &str) -> Vec<&str> {
```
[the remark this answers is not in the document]
Dr-Emann: I don't agree with the reasoning behind that lint, and it [conflicts](https://github.com/rust-lang/rust-clippy/issues/5369) with an official rust lint (`unreachable_pub`).

I personally much prefer to know when reading the function "this is definitely not exposed outside this crate".

extensions/route_map/lib.rs
Dr-Emann: I go back and forth on this. There is a [lint](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#elided-lifetimes-in-paths) for this as well. I can remove it though.

extensions/route_map/search.rs
```diff
@@ -0,0 +1,105 @@
...
+        match plain_routes.entry(String::from(path)) {
+            Entry::Occupied(entry) => entry.into_mut().merge(py, handler_group, path)?,
+            Entry::Vacant(entry) => entry.insert(handler_group),
+        }
+    };
+    Ok(handler_group)
+}
+
+impl RouteMap {
+    pub(crate) fn find_handler_group<'a>(&'a self, full_path: &'a str) -> PyResult<FindResult<'a>> {
```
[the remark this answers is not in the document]
Dr-Emann: Good catch, it used to have param_values borrowed from `full_path`, but that changed to Strings

extensions/route_map/test/mod.rs
Dr-Emann: Even though I literally did it elsewhere in this change, I always forget you can add impl blocks in sub modules. I was gonna say "but it's only used in the tests", but of course.. We can just define it as a method only in the test module.

pyproject.toml
```diff
@@ -26,7 +26,9 @@ classifiers = [
     "Topic :: Software Development",
     "Typing :: Typed",
 ]
-include = ["CHANGELOG.md"]
+include = [
+    { path = "extensions/**", format = "sdist" },
```
Dr-Emann, line 30: Note to self: we also need to add `Cargo.*` to the sdist, or we can't build the rust code.

extensions/route_map/lib.rs
```diff
@@ -1,16 +1,358 @@
...
+        let path_parameters: &PyAny = route.path_parameters()?;
+        let is_static = self.static_paths.contains(path);
+
+        let new_handler_group = if is_static {
+            if !route.is_asgi(&self.ctx)? {
+                return Err(wrappers::ImproperlyConfiguredException::new_err(format!(
+                    "Static route handlers must be asgi handlers at {path}"
+                )));
+            }
+            if path_parameters.is_true()? {
```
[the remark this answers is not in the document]
Dr-Emann: 1. I wish this method was `PyAny::is_truthy()`.
2. Yeah. I was trying to be more general (`is_empty` requires a sequence/mapping, truthyness doesn't), but we already require it to be a sequence because we extract a vec from it. Will use `is_empty`.

extensions/route_map/search.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::{get_base_components, util, wrappers, HandlerGroup, Node, RouteMap};
+
+pub(crate) fn find_insert_handler_group<'a>(
+    root: &'a mut Node,
+    plain_routes: &'a mut HashMap<String, HandlerGroup>,
+    path: &str,
+    path_parameters: &PyAny,
+    is_static: bool,
+    handler_group: HandlerGroup,
+) -> PyResult<&'a mut HandlerGroup> {
```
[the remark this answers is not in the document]
Dr-Emann: Yeah, originally the code inspected/added to the returned handler_group, but since I changed it to just provide a group to merge, the return value isn't needed anymore. Will remove.

extensions/route_map/util.rs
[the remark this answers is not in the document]
nramos0: Fair enough, I sometimes am also back and forth on this lint. I guess we are not exposing things from the crate in the normal sense anyway though, since we know only the things within #[pymethods] will be available to python and everything else is inaccessible. I'm not sure what we should do about visibility in general. I'm fine with leaving these pub(crate)s in the end

extensions/route_map/search.rs
Dr-Emann: I think the usual case will be to have a quite small number of placeholder components, so I think I'm okay with starting empty for now, we can benchmark if it will matter.

PR #339, opened 2022-08-07
Issue 338 update responses

### LINKED ISSUE #338: Proposal: allow for route handlers to return tuples to handle response headers and cookies.
So as I see it the current API is somewhat constricted, as also pointed out in #278. Namely, if the user wants to set response header, cookies or background tasks the only option really is to use an instance of `starlite.responses.Response` or `starlette.responses.Response`. For example, lets say a user returns the following json response:

```python
from starlite import get


@get("/greet/{name:str}")
def greet_handler(name: str) -> dict[str, str]:
    return {"hello": name}
```

If the user wants to set a specific header on this response he or she has three options, either return a response class, use the `after_request` life-cycle hook, or use a middleware. While all of these are viable, if the user wants to do some sort of computation in the route handler function and set the header based on it, only the response class route is viable:

```python
from starlite import Response, MediaType, get
from starlette.status import HTTP_200_OK


@get("/greet/{name:str}")
def greet_handler(name: str) -> Response:
    return Response(content={"hello": name}, media_type=MediaType.JSON, status_code=HTTP_200_OK, headers={"MY_HEADER": name})
```

The problem with the above is that OpenAPI docs are not going to be very good in this case, because we do not know the response content. 

One option would be to make the `Response` class generic and then use the typing args that go into it. This though is not very scalable and it will break existing code.

So the other option I can think of, and which I think is better, is to allow route handlers to return a NamedTuple of values if needed:

```python
from typing import NamedTuple, Any, Optional, Dict, Tuple, TYPE_CHECKING, Generic, TypeVar
from starlite import get

if TYPE_CHECKING:
    from starlette.background import BackgroundTask

T = TypeVar("T")
    
class ResponseValues(NamedTuple, Generic[T]):
    data: T
    headers: Optional[Dict[str, str]]
    cookies: Optional[Dict[str, str]]
    background: Optional["BackgroundTask"]


@get("/greet/{name:str}")
def greet_handler(name: str) -> ResponseValues[Dict[str, str]]:
    return ResponseValues(data={"hello": name}, headers={"MY_HEADER": name})
```

We will be able to access the generic value here for documentation, and this is going to be simpler then to modify the response class.

Goldziher:
This PR addresses and resolves #338


starlite/handlers/http.py
```diff
@@ -280,24 +281,43 @@ def _get_response_from_data(
...
-            if not app.template_engine:
-                raise ImproperlyConfiguredException("Template engine is not configured")
-            return TemplateResponse(
-                context=data.context,
-                template_name=data.name,
-                template_engine=app.template_engine,
-                status_code=self.status_code,
-                headers=headers,
-            )
+        if isinstance(data, (Redirect, File, Stream, Template)):
```
peterschutt, line 284: Could make `StarliteType` abstract with a `to_starlette_response()` method:

```python
# datastructures.py
class StarliteType(BaseModel, abc.ABC):
    ...
    @abc.abstractmethod
    def to_starlette_response(self, media_type: MediaType, headers: dict[str, str], status_code: int) -> StarletteResponse:
        """
        Construct a `starlette.responses.Response` appropriate for type.

        Args:
            media_type (MediaType): Media type for response.
            headers (dict[str, str]): Headers to be merged with `self.headers` and added to the response.
            status_code (int): Status code for response.

        Returns:
            StarletteResponse
        """
        ...


class File(StarliteType):
    ...
    def to_starlette_response(self, media_type: MediaType, headers: dict[str, str], status_code: int) -> FileResponse:
        return FileResponse(
            background=self.background,
            filename=self.filename,
            headers={**headers, **self.headers},
            media_type=media_type,
            path=self.path,
            stat_result=self.stat_result,
        )
```
This method would then collapse down to:

```python
    def _get_response_from_data(
        self,
        headers: dict,
        data: Union[StarletteResponse, StarliteType],
        media_type: Union[MediaType, str],
        app: "Starlite",
) -> StarletteResponse:
    if isinstance(data, StarliteType):
        return data.to_starlette_response(media_type, headers, self.status_code)
    return data
```
Could even do away with this method altogether in this case and do that instance check in `to_response()`.

docs/usage/5-responses/0-responses-intro.md
```diff
@@ -0,0 +1,25 @@
...
+    id: int
+    name: str
+
+
+@get("/resources")
+def retrieve_resource() -> Resource:
+    return Resource(id=1, name="my resource")
+```
+
+In the example above, the route handle function returns an instance of the `Resource` pydantic class. This value will
```
peterschutt, line 22: ~handle~ handler

docs/usage/5-responses/0-responses-intro.md
```diff
@@ -0,0 +1,25 @@
...
+then be used by Starlite to construct an instance of the `starlite.response.Response` class using defaults values: the
+response status code will be set to `200` and it's `Content-Type` header will be set to `application/json`.
+The `Resource` instance will be serialized into JSON and set the response body.
```
peterschutt, line 25: "... set as the response body."

docs/usage/5-responses/1-media-type.md
```diff
@@ -0,0 +1,90 @@
+# Media Type
+
+You do not have to specify the `media_type` kwarg in the route handler function if the response is should be JSON. But
```
peterschutt, line 3: "if the response ~is~ should be JSON."

docs/usage/5-responses/1-media-type.md
```diff
@@ -0,0 +1,90 @@
...
+```python
+from starlite import MediaType, get
+
+
+@get("/resources", media_type=MediaType.TEXT)
+def retrieve_resource() -> str:
+    return "The rumbling rabbit ran around the rock"
+```
+
+The value of the `media_type` kwarg affect both the serialization of response data and the generation of OpenAPI docs.
```
peterschutt, line 16: "... kwarg ~affect~ affects both..."

docs/usage/5-responses/2-status-codes.md
```diff
@@ -0,0 +1,41 @@
...
+    When using the `route` decorator with multiple http methods, the default status code is `200`.
+
+<!-- prettier-ignore -->
+!!! note
+    The default for `delete` is `204` because by default it is assumed that delete operations return no data.
+    This though might not be the case in your implementation - so take care of setting it as you see fit.
+
+<!-- prettier-ignore -->
+!!! tip
+    While you can write integers as the value for `status_code`, e.g. `200`, its best practice to use constants (also in
```
peterschutt, line 38: "...~its~ it's best practice..."

docs/usage/5-responses/3-returning-responses.md
```diff
@@ -0,0 +1,54 @@
...
+            id=1,
+            name="my resource",
+            headers={"MY-HEADER": "xyz"},
+            cookies=[Cookie("my-cookie", value="abc")],
+        )
+    )
+```
+
+As you can see above, the `starlite.response.Response` class accepts a generic argument - in this case the pydantic
+model `Resource`. This allows Starlite to infer from the Response type the correct typing for OpenAPI generation.
```
peterschutt, line 54: either "... ~Response~ `Response` type...`" or "... ~Response~ response type..."

docs/usage/5-responses/4-response-headers.md
```diff
@@ -0,0 +1,203 @@
...
+    route_handlers=[router],
+    response_headers={
+        "app-level-header": ResponseHeader(
+            value="app header", description="app level header"
+        )
+    },
+)
+```
+
+In the above example the response returned from `my_route_handler` will have a headers set from each layer of the
```
peterschutt, line 51: "... will have ~a~ headers..."

peterschutt (approved):
Hopefully I didn't focus on any docs that are still TODO, but happy to give them a once over when they are there.

All looks good! I have a suggestion for moving the responsibility of constructing the starlette response types to `StarliteType` and also a suggestion to consider a more specific name for the `StarliteType` base class.

Cheers!

starlite/handlers/http.py
Goldziher: maybe in a followup.

starlite/datastructures.py
```diff
@@ -19,15 +21,57 @@ def copy(self) -> "State":
...
+    """domain for which the cookie is valid."""
+    secure: Optional[bool] = None
+    """https is required for the cookie."""
+    httponly: Optional[bool] = None
+    """forbids javascript to access the cookie via 'Document.cookie'."""
+    samesite: Literal["lax", "strict", "none"] = "lax"
+    """controls whether or not a cookie is sent with cross-site requests. Defaults to 'lax'."""
+
+
 class StarliteType(BaseModel):
```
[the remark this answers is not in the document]
Goldziher: I eliminated it altogether.

docs/usage/5-responses/5-response-cookies.md
```diff
@@ -0,0 +1,271 @@
...
+While the above scheme works great for static cookie values, it doesn't allow for dynamic cookies. Because cookies are
+fundamentally a type of response header, we can utilize the same patterns we use for
+setting [dynamic headers](./4-response-headers.md#dynamic-headers) also here.
+
+### Setting Response Headers Using Annotated Responses
+
+We can simply return a response instance directly from the route handler and set the cookies list manually
+as you see fit, e.g.:
+
+```python
```
peterschutt, line 123:
```python
from random import randint

from starlette.status import HTTP_200_OK
from starlite import Response, get
from starlite.datastructures import Cookie
from starlite.enums import MediaType


class Resource(BaseModel):
    id: int
    name: str


@get(
    "/resources",
    response_cookies=[
        Cookie(
            key="Random-Header",
            description="a random number in the range 1 - 100",
            documentation_only=True,
        )
    ],
)
def retrieve_resource() -> Response[Resource]:
    return Response(
        Resource(
            id=1,
            name="my resource",
        ),
        cookies=[Cookie(key="Random-Header", value=str(randint(1, 100)))],
        media_type=MediaType.JSON,
        status_code=HTTP_200_OK,
    )
```

docs/usage/5-responses/5-response-cookies.md
```diff
@@ -0,0 +1,271 @@
...
+the [annotated response](3-returning-responses.md#annotated-responses) we return. To this end we do not set a `value`
+for it and we designate it as `documentation_only=True`.
+
+### Setting Response Headers Using the After Request Hook
+
+An alternative pattern would be to use an [after request handler](../13-lifecycle-hooks.md#after-request). We can define
+the handler on different layers of the application as explained in the pertinent docs. We should take care to document
+the cookies on the corresponding layer:
+
+```python
```
peterschutt, line 166:
```python
from random import randint

from starlite import Response, Router, get
from starlite.datastructures import Cookie


class Resource(BaseModel):
    id: int
    name: str


@get("/resources")
def retrieve_resource() -> Resource:
    return Resource(
        id=1,
        name="my resource",
    )


def after_request_handler(response: Response) -> Response:
    response.set_cookie(
        **Cookie(key="Random-Header", value=str(randint(1, 100))).dict(
            exclude_none=True, exclude={"documentation_only"}
        )
    )
    return response


router = Router(
    path="/router-path",
    route_handlers=[retrieve_resource],
    after_request=after_request_handler,
    response_cookies=[
        Cookie(
            key="Random-Header",
            description="a random number in the range 1 - 100",
            documentation_only=True,
        )
    ],
)
```

peterschutt: note the `exclude={"documentation_only"}`

docs/usage/5-responses/5-response-cookies.md
```diff
@@ -0,0 +1,271 @@
...
+```
+
+In the above we set the cookie using an `after_request_handler` function on the router level. Because the
+handler function is applied on the router, we also set the documentation for it on the router.
+
+We can use this pattern to fine-tune the OpenAPI documentation more granularly by overriding cookie specification as
+required. For example, lets say we have a router level cookie being set and a local cookie with the same key but a
+different value range:
+
+```python
```
peterschutt, line 217:
```python
from random import randint

from starlette.status import HTTP_200_OK
from starlite import Response, Router, get
from starlite.datastructures import Cookie
from starlite.enums import MediaType


class Resource(BaseModel):
    id: int
    name: str


@get(
    "/resources",
    response_cookies=[
        Cookie(
            key="Random-Cookie",
            description="a random number in the range 100 - 1000",
            documentation_only=True,
        )
    ],
)
def retrieve_resource() -> Response[Resource]:
    return Response(
        Resource(
            id=1,
            name="my resource",
        ),
        cookies=[Cookie(key="Random-Cookie", value=str(randint(100, 1000)))],
        media_type=MediaType.JSON,
        status_code=HTTP_200_OK,
    )


def after_request_handler(response: Response) -> Response:
    response.set_cookie(
        **Cookie(key="Random-Cookie", value=str(randint(1, 100))).dict(
            exclude_none=True, exclude={"documentation_only"}
        )
    )
    return response


router = Router(
    path="/router-path",
    route_handlers=[retrieve_resource],
    after_request=after_request_handler,
    response_cookies=[
        Cookie(
            key="Random-Cookie",
            description="a random number in the range 1 - 100",
            documentation_only=True,
        )
    ],
)
```
Changed the key of the cookies from `Random-Header` to `Random-Cookie`.

When I run this example, I get two `set-coookie` headers:

```text
HTTP/1.1 200 OK
date: Mon, 08 Aug 2022 23:12:05 GMT
server: uvicorn
content-length: 29
content-type: application/json
set-cookie: Random-Cookie=419; Path=/; SameSite=lax
set-cookie: Random-Cookie=20; Path=/; SameSite=lax
```
I suppose because the cookies are normalised before the after request handler is called.

peterschutt (changes requested):
From tesing the docs examples.

Issue #296, opened 2022-07-20
Enhancement: CSRF Middleware


cofin:
Implement a middleware to provide CSRF validation.  No additional libraries should be required since everything we'd need is built into Starlette.

seladb:
@cofin @Goldziher I'm trying to tackle this issue. We can use [Starlette's 3rd-party `CSRFMiddleware`](https://github.com/frankie567/starlette-csrf) and write a simple wrapper around it or write our own. What would you prefer?

Goldziher:
> @cofin @Goldziher I'm trying to tackle this issue. We can use [Starlette's 3rd-party `CSRFMiddleware`](https://github.com/frankie567/starlette-csrf) and write a simple wrapper around it or write our own. What would you prefer?

Hi, and thanks. Please no third party dependencies unless they are very strongly maintained, which isn't the case here.

seladb:
Sounds good! In that case we can write our own implementation which should be somewhat similar to [`CSRFMiddleware`](https://github.com/frankie567/starlette-csrf)

seladb:
@Goldziher you can assign this issue to me, it might take me a few days because I'm still new to starlite but I think I can get it done

Goldziher:
> @Goldziher you can assign this issue to me, it might take me a few days because I'm still new to starlite but I think I can get it done

Done. You're welcome to join our discord where we can chat 😉

cofin:
Thanks, @seladb.  Let us know how we can help!

Goldziher:
done

PR #340, opened 2022-08-08
Issue #296 - add CSRF middleware

seladb:


starlite/middleware/csrf.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            decoded_request_token = self._decode_csrf_token(request_csrf_token)
+            decoded_cookie_token = self._decode_csrf_token(cookie_csrf_token)
+            return secrets.compare_digest(decoded_request_token, decoded_cookie_token)
+        except BadSignature:
+            return False
+
+    @classmethod
+    def _get_error_response(cls) -> PlainTextResponse:
+        return PlainTextResponse(content="CSRF token verification failed", status_code=403)
```
seladb, line 103: Do we want to enable customizing the error?

seladb (commented):
I didn't yet update the docs, it'd be great to get your feedback on the implementation first. Once it's approved I can work on the docs as well

starlite/config.py
```diff
@@ -141,6 +142,20 @@ def dict(self, *args, **kwargs) -> Dict[str, Any]:  # type: ignore[no-untyped-de
...
+
+    secret: str
+    cookie_name: str = "csrftoken"
+    cookie_path: str = "/"
+    header_name: str = "x-csrftoken"
+    cookie_secure: bool = False
+    cookie_httponly: bool = False
+    cookie_samesite: str = "Lax"
+    cookie_domain: Optional[str] = None
+    safe_methods: Set[str] = {"GET", "HEAD", "OPTIONS", "TRACE"}
```
Goldziher, line 156: Better to type as: `Set[literal["GET","POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS", "TRACE"]]`. Also see `starlite.types.Method`, you might want to update that type and use it instead.

starlite/middleware/csrf.py
Goldziher: you could instead of return an error response like so raise a `starlite.exceptions.PermissionDeniedException`. It will be caught by the exception handler and transformed into a response. This is better because if the user uses JSON it will be JSON etc.

starlite/config.py
```diff
@@ -141,6 +142,20 @@ def dict(self, *args, **kwargs) -> Dict[str, Any]:  # type: ignore[no-untyped-de
...
+class CSRFConfig(BaseModel):
+    """Class containing CSRF configuration."""
+
+    secret: str
+    cookie_name: str = "csrftoken"
+    cookie_path: str = "/"
+    header_name: str = "x-csrftoken"
+    cookie_secure: bool = False
+    cookie_httponly: bool = False
+    cookie_samesite: str = "Lax"
```
Goldziher, line 154: i think its actually `lax` rather. Also, typing should be I think `Literal["lax", "strict", "none"]`:

```suggestion
    cookie_samesite: Literal["lax", "strict", "none"] = "lax"
```

Also make sure to import `Literal` from `typing_extensions` for backwards compat with py 3.7.

starlite/config.py
```diff
@@ -141,6 +142,20 @@ def dict(self, *args, **kwargs) -> Dict[str, Any]:  # type: ignore[no-untyped-de
         return super().dict(*args, **kwargs)
 
 
+class CSRFConfig(BaseModel):
```
Goldziher, line 145: note, you can rebase your PR on my own PR which is about to be merged. In that PR there is a type `starlite.datastructures.Cookie`, which you could use as well. In either case - take a look there-->  https://github.com/starlite-api/starlite/pull/339

starlite/middleware/csrf.py
```diff
@@ -0,0 +1,103 @@
...
+        super().__init__(app)
+        self.app = app
+        self.config = config
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        request: Request = Request(scope=scope)
```
Goldziher, line 38:
```suggestion
        request = Request(scope=scope)
```

starlite/middleware/csrf.py
```diff
@@ -0,0 +1,103 @@
...
+        csrf_cookie = request.cookies.get(self.config.cookie_name)
+        existing_csrf_token = request.headers.get(self.config.header_name)
+
+        send_f = send
```
Goldziher, line 42: please update the variable name to be more descriptive. E.g. `original_send` etc.

starlite/middleware/csrf.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if request.method not in self.config.safe_methods:
+            if not self._csrf_tokens_match(existing_csrf_token, csrf_cookie):
+                response = self._get_error_response()
```
Goldziher, line 46: see my comment about the option of raising an exception here.

starlite/middleware/csrf.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _csrf_tokens_match(self, request_csrf_token: Optional[str], cookie_csrf_token: Optional[str]) -> bool:
+        if not (request_csrf_token and cookie_csrf_token):
+            return False
+        try:
+            decoded_request_token = self._decode_csrf_token(request_csrf_token)
+            decoded_cookie_token = self._decode_csrf_token(cookie_csrf_token)
+            return secrets.compare_digest(decoded_request_token, decoded_cookie_token)
+        except BadSignature:
+            return False
```
Goldziher, line 99: So, please change thiis - you are using exceptions to do the if/else switching here. Rather, return None from the `_decode_csrf_token` method instead of raising an exception, and then return false if either `self._decode_csrf_token(request_csrf_token)` or `self._decode_csrf_token(cookie_csrf_token)` are `None`.

Goldziher (commented):
Really great stuff. I left some comments but really excellent work.

Goldziher (approved):
Approve. Awaiting docs.

starlite/middleware/csrf.py
seladb: Thanks for the advice, done!

starlite/config.py
seladb: Good point, done!

starlite/config.py
seladb: Done!

starlite/middleware/csrf.py
seladb: When I remove the type hint `mypy` complains 🤦‍♂️

starlite/middleware/csrf.py
seladb: Done!

starlite/middleware/csrf.py
seladb: Done!

starlite/middleware/csrf.py
seladb: Done!

starlite/config.py
seladb: @Goldziher do you want to do it in this PR or in a separate one?

seladb:
> Approve. Awaiting docs.

Thanks for the quick review @Goldziher ! Do you want me to add docs to this PR?

Goldziher:
> > Approve. Awaiting docs.
> 
> Thanks for the quick review @Goldziher ! Do you want me to add docs to this PR?

Yes please, and docstrings. If you have questions ask @peterschutt here or on discord.

seladb:
@Goldziher @peterschutt I have added documentation and docstrings, please let me know if anything else is missing

docs/usage/7-middleware.md
```diff
@@ -144,6 +144,58 @@ app = Starlite(
...
+csrf_config = CSRFConfig(secret="my-secret")
+
+app = Starlite(route_handlers=[...], csrf_config=csrf_config)
+```
+
+You can pass the following kwargs to `CSRFConfig`:
+
+- `secret` - this is the only mandatory parameter, it's a string that is used to create an HMAC to sign the CSRF token
+- `cookie_name` - the CSRF cookie name, set to `csrftoken` by default
+- `cookie_path` - the CSRF cookie path, set to `/` be default
```
peterschutt, line 187: ~be~ by

peterschutt (approved):
Nice work, just one small typo to fix.

`CSRFMiddleware` will need to be added to the reference docs but the rest of the middleware types haven't been included either yet, so I've added it to #341 so we'll do at the same time as the others.

Cheers!

docs/usage/7-middleware.md
Goldziher, line 187:
```suggestion
- `cookie_path` - the CSRF cookie path, set to `/` by default
```

Goldziher:
@all-contributors add @seladb code

PR #346, opened 2022-08-09
Update `to_response` and datastructures

Goldziher:
This is a followup PR that cleans the `to_response` method and implements @peterschutt's suggestions for datastructures.  I additionally add some docs and a docker compose setup to run the docs locally (didnt work due to plugins).

You'll note i removed the `validate_arguements` from the respone classes. This was causing a lot of problems with typing, and it additionally hurts performance I think.


peterschutt (approved):
Nice and thanks for doing all those docs!

Issue #354, opened 2022-08-11
Application doesn't fallback to index.html when the file is not specified


infohash:
```python
app = Starlite(route_handlers=[device_router, health_check],
               static_files_config=[
                   StaticFilesConfig(
                       directories=['templates'], path='/home', html_mode=True)
                   ],
               openapi_config=OpenAPIConfig(title='Starlite Service', version='0.0.1'))
```

## Structure

```
── src
│   ├── ...
│   ├── templates
│   │   ├── 404.html
│   │   ├── index.html
│   ├── ...
```

Path set in `StaticFilesConfig`: http://127.0.0.1:8000/home

## From Documentation

[Static Files](https://starlite-api.github.io/starlite/usage/0-the-starlite-app/3-static-files/)

> If html_mode is enabled and no specific file is requested, the application will fall back to serving index.html. If no file is found the application will look for a 404.html file in order to render a response, otherwise a 404 NotFoundException will be returned.

## Observed Behaviour

When I don't specify the file, it doesn't fallback to `index.html`. It is falling back to `404.html`.

## Expected Behaviour

Visiting http://127.0.0.1:8000/home should serve `index.html` instead of 404.html. If `index.html` is not available, then it should fallback to 404.html.

## What's Working

http://127.0.0.1:8000/home/index.html is working.

Goldziher:
ok, this is a starlette bug. For now append index.html to the path - either its solved in Starlette or we resolve it here. 

See: https://github.com/encode/starlette/discussions/1806

infohash:
Let's wait for the fix in upstream.

cofin:
These changes look good to me.

Goldziher:
v1.10.1 is being released now. Please test the fix @infohash

PR #355, opened 2022-08-11
#issue 354 html mode not working as expected

Goldziher:
This PR resovles issue #354 by fixing the handling of static paths, which was broken in one of the refactors apparently.


cofin (approved):
These changes look good.

Issue #367, opened 2022-08-14
Bug: 1.11 breaks hello world


peterschutt:
```python
"""
Minimal Starlite application.
"""
from typing import Any

from starlite import Starlite, get


@get("/")
def hello_world() -> dict[str, Any]:
    """Hi"""
    return {"hello": "world"}


app = Starlite(route_handlers=[hello_world])
```

```python-traceback
(starlite-hello-world-nPMVAxaQ-py3.10) peter@peter-Inspiron-15-5510:~/PycharmProjects/starlite-hello-world$ poetry run uvicorn main:app
Traceback (most recent call last):
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/bin/uvicorn", line 8, in <module>
    sys.exit(main())
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/uvicorn/main.py", line 407, in main
    run(
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/uvicorn/main.py", line 575, in run
    server.run()
  File "/home/peter/.cache/pypoetry/virtualenvs/starlite-hello-world-nPMVAxaQ-py3.10/lib/python3.10/site-packages/uvicorn/server.py", line 60, in run
    return asyncio.run(self.serve(sockets=sockets))
[...]

PR #368, opened 2022-08-14
Adds helper method to handle difference caused by builtin vs generic collection types

peterschutt:
For issue #367

```python
>>> import inspect
>>> inspect.isclass(dict[str, object])
True
>>> from typing import Dict
>>> inspect.isclass(Dict[str, object])
False
```
Change adds a method to handle the difference between the response from `inspect.isclass()` when the input is a generic collection type imported from `typing` or  a generic builtin available in 3.9+.


Goldziher:
So, instead of method I'd put it as util and then use it wherever we use issubclass. Also in the future

Goldziher (approved):
Great 😸

Issue #373, opened 2022-08-15
Bug: Certain routes break the route_map structure


waweber:
I found a few ways to specify routes that result in a `route_map` which doesn't work as expected. Here's an example script:

```python
from pprint import pprint
from starlite import Starlite, get, Controller

class Example(Controller):
    # Breaks because app.py:279 replaces all "user:str" with the empty string
    @get("/test1/user:str/{user:str}")
    def route1(self, user: str) -> str:
        return f"User: {user}"

    # Throws an exception because "_components" is a special value in the route map
    @get("/test2/_components/{user:str}")
    def route2(self, user: str) -> str:
        return f"User: {user}"

    # Breaks because "*" is a special value in the route map.
    # GET /test3/foo/bar returns "User: foo"
    @get("/test3/*/{user:str}")
    def route3(self, user: str) -> str:
        return f"User: {user}"

app = Starlite(route_handlers=[Example])
pprint(app.route_map)
```

The bug from example 1 starts at the regex used to parse route parameters (https://github.com/starlite-api/starlite/blob/main/starlite/routes/base.py#L21). `re.findall` will match "{user:str}" and return "user:str" without the brackets as the match because the pattern has exactly one capturing group. This is then `.replace`d with the empty string when building the route map (https://github.com/starlite-api/starlite/blob/3914b5090d6db579152a24e954dd31cc89684bac/starlite/app.py#L283) but since it doesn't include the brackets, it could include non-parameter url segments.

Example 2 throws an exception during route map construction simply because `_components` is a special name. There is currently no documentation about what path components are reserved names, but I think a proper fix would be to make the metadata not be an element inside the prefix tree.

Example 3 is similar in that "*" has special meaning in the prefix tree that is not documented. This results in an extra path param, and the `user` param will no longer be properly mapped.

Goldziher:
Hiya,

Well TBH Im aware that `_components` and `*` will break. I believe that these are not possible edge cases in user land, but rather require that the user intentionally tries to break stuff. In this case we do not need to cater to the user. This is why this is not guard against. We could though add validation specifically against using these in the `_validate_path_parameters` method on the `BaseRoute` class, which you already discovered. Saying that, I am open to any PR that has an elegant and clean solution to this.

As for the first issue `/test1/user:str/{user:str}` - here you indeed identify a bug IMO. We should validate the path string and ensure that all path params are well formed or through an exception. If you want to contribute this, it would be great.

PR #374, opened 2022-08-16
Fix bugs involving "*" and path params in routes

waweber:
- Instead of using "*" as the key in the route map for path params, use a sentinel object which will never be equal to any other value
- When creating routes, parse the route path into a sequence of strings and parameter definitions. Re-use this when building the route map instead of parsing the path again.
- Allow starlite.utils.url.join_paths to accept an iterable of strings.

This should fix some of the issues in #373. In another PR I will look at making a proper class for route map nodes, so that their config can use slots and not share the same dict key space as path components.

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type, Union, cast
+from typing import TYPE_CHECKING, Dict, List, Optional, Set, Type, Union, cast
 
 from pydantic import validate_arguments
 from pydantic.fields import FieldInfo
@@ -6,7 +6,7 @@ from starlette.middleware import Middleware as StarletteMiddleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.middleware.trustedhost import TrustedHostMiddleware
 
-from starlite.asgi import StarliteASGIRouter
+from starlite.asgi import RouteMapDict, StarliteASGIRouter, _path_param
 from starlite.config import (
     CacheConfig,
     CompressionConfig,
@@ -45,8 +45,10 @@ if TYPE_CHECKING:
     from pydantic_openapi_schema.v3_1_0.open_api import OpenAPI
     from starlette.types import ASGIApp, Receive, Scope, Send
 
+    from starlite.asgi import ComponentsSet, _PathParam
     from starlite.handlers.base import BaseRouteHandler
     from starlite.handlers.websocket import WebsocketRouteHandler
+    from starlite.routes.base import PathParameterDefinition
 
 DEFAULT_OPENAPI_CONFIG = OpenAPIConfig(title="Starlite API", version="1.0.0")
 """
@@ -164,7 +166,7 @@ class Starlite(Router):
         self.compression_config = compression_config
         self.plain_routes: Set[str] = set()
         self.plugins = plugins or []
-        self.route_map: Dict[str, Any] = {}
+        self.route_map: RouteMapDict = {}
         self.routes: List[BaseRoute] = []
         self.state = State()
 
@@ -267,7 +269,7 @@ class Starlite(Router):
 
         return ExceptionHandlerMiddleware(app=app, exception_handlers=exception_handlers, debug=self.debug)
 
-    def _add_node_to_route_map(self, route: BaseRoute) -> Dict[str, Any]:
+    def _add_node_to_route_map(self, route: BaseRoute) -> RouteMapDict:
         """Adds a new route path (e.g. '/foo/bar/{param:int}') into the
         route_map tree.
 
@@ -279,16 +281,19 @@ class Starlite(Router):
         current_node = self.route_map
         path = route.path
         if route.path_parameters or path in self._static_paths:
-            for param_definition in route.path_parameters:
-                path = path.replace(param_definition["full"], "")
-            path = path.replace("{}", "*")
-            components = ["/", *[component for component in path.split("/") if component]]
+            components = cast("List[Union[str, _PathParam, PathParameterDefinition]]", ["/", *route.path_components])
             for component in components:
-                components_set = cast("Set[str]", current_node["_components"])
+                components_set = cast("ComponentsSet", current_node["_components"])
+
+                if isinstance(component, dict):
+                    # Represent path parameters using a special value
+                    component = _path_param
+
                 components_set.add(component)
+
                 if component not in current_node:
                     current_node[component] = {"_components": set()}
-                current_node = cast("Dict[str, Any]", current_node[component])
+                current_node = cast("RouteMapDict", current_node[component])
                 if "_static_path" in current_node:
                     raise ImproperlyConfiguredException("Cannot have configured routes below a static path")
         else:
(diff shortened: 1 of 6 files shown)


tests/test_path_resolution.py
```diff
@@ -184,3 +184,28 @@ def handler_fn(a: int = 0, b: int = 0, c: int = 0, d: int = 0) -> None:
 
     with pytest.raises(ImproperlyConfiguredException):
         create_test_client(handler_fn)
+
+
+@pytest.mark.parametrize(
+    "handler_path, request_path, expected_status_code, expected_param",
+    [
+        ("/name:str/{name:str}", "/name:str/test", HTTP_200_OK, "test"),
```
Goldziher, line 192: I don't think this should be an `HTTP_200_OK`, rather `name:str` should raise an `ImproperlyConfiguredException` because this is a clear situation where the user forgot the place the value inside brackets - `{...}`.

tests/test_path_resolution.py
```diff
@@ -184,3 +184,28 @@ def handler_fn(a: int = 0, b: int = 0, c: int = 0, d: int = 0) -> None:
...
+
+    with create_test_client(handler_fn) as client:
+        response = client.get(request_path)
+        assert response.status_code == expected_status_code
+
+        if response.ok:
+            if expected_param is not None:
+                assert response.text == expected_param
+            else:
+                assert response.content == b""
```
Goldziher, line 211: what are you testing here and how does this relate to the test case🙃 ?

tests/route/test_route_map.py
```diff
@@ -16,10 +17,20 @@
...
+def is_path_in_route_map(route_map: RouteMapDict, path: str, path_params: Set[str]) -> bool:
     if not path_params:
         return path in route_map
-    components = ["/", *[component for component in param_pat.sub("*", path).split("/") if component]]
+    components = cast(
+        "List[Union[str, _PathParam]]",
+        [
+            "/",
+            *[
+                _path_param if param_pat.fullmatch(component) else component
```
Goldziher, line 28: Can you rename `param_pat` to `param_pattern`? This is somewhat confusing (I know its not new, but while at it..)

starlite/app.py
```diff
@@ -164,7 +166,7 @@ def __init__(
         self.compression_config = compression_config
         self.plain_routes: Set[str] = set()
         self.plugins = plugins or []
-        self.route_map: Dict[str, Any] = {}
+        self.route_map: RouteMapDict = {}
```
Goldziher, line 169: So note on the `RouteMapDict` type - I dont really see the point of using it TBH over using Dict[str, Any]. Its not such a big deal so its fine to include if you feel strongly about it, but I believe in being explicit wherever this doesn't hurt code cleanliness. 

I have a comment lower regarding naming which is relevant.

starlite/asgi.py
```diff
@@ -20,6 +29,18 @@
     from starlite.types import LifeCycleHandler
 
 
+# Sentinel object to represent a path param in the route map
+class _PathParam:
```
Goldziher, line 33: some points here-

1. no need for the leading underscore IMO
2. I would rename it to PathParamPlaceholder or something such like
3. comment should be a docstring
4. whats the point of having a repr method?

You can simply use the class, no need to use an instance for it. See usage for example of `EmptyType` and `Empty` from `starlite.types`.

starlite/asgi.py
```diff
@@ -20,6 +29,18 @@
...
+    def __repr__(self) -> str:  # pragma: no cover
+        return "*"
+
+
+_path_param = _PathParam()
+
+RouteMapDict = MutableMapping[Union[str, _PathParam], Any]
```
Goldziher, line 40: A better name IMO would be `RouteMapNode`. Also, any reason not to use `Dict[]` as a type?

starlite/routes/base.py
```diff
@@ -98,38 +102,51 @@ def create_handler_kwargs_model(self, route_handler: "BaseRouteHandler") -> Kwar
...
+            if param_match:
+                # Parse and validate components that are path params
+                param = param_match.group(1)
+                cls._validate_path_parameter(param)
+                param_name, param_type = (p.strip() for p in param.split(":"))
+                parsed_components.append(
+                    PathParameterDefinition(name=param_name, type=param_type_map[param_type], full=param)
+                )
+            else:
+                # Just append strings
```
Goldziher, line 144: IMO this comment is redundant - either clarify the rationale, or remove it.

starlite/routes/base.py
```diff
@@ -98,38 +102,51 @@ def create_handler_kwargs_model(self, route_handler: "BaseRouteHandler") -> Kwar
...
-        return path, path_format, path_parameters
+
+        parsed_components: List[Union[str, PathParameterDefinition]] = []
+
+        # Handle each component in the route path
+        components = [component for component in path.split("/") if component]
+        for component in components:
+            param_match = param_match_regex.fullmatch(component)
+            if param_match:
+                # Parse and validate components that are path params
```
Goldziher, line 136: see my next comment, applies here and above.

starlite/routes/base.py
```diff
@@ -98,38 +102,51 @@ def create_handler_kwargs_model(self, route_handler: "BaseRouteHandler") -> Kwar
...
+                parsed_components.append(
+                    PathParameterDefinition(name=param_name, type=param_type_map[param_type], full=param)
+                )
+            else:
+                # Just append strings
+                parsed_components.append(component)
+
+        # Build the url used for the OpenAPI schema
+        path_format = join_paths(
+            f"{{{component['name']}}}" if isinstance(component, dict) else component for component in parsed_components
```
Goldziher, line 149: Looks somewhat weird TBH. Might be simpler to parse mentally if split into two assignments.

Goldziher (commented):
Looks good, I added some comments.

tests/test_path_resolution.py
waweber: Not sure it is possible to know that in all cases, as "name:str" is well within the domain of valid strings that can appear in a URL. Previously, this route path got parsed as "/{name:str}" and this test case is to assure that doesn't happen any more.

tests/test_path_resolution.py
waweber: A check to make sure the parameter was correctly parsed from the request URL, which is returned by the test route handler. Previously, "/*/{name:str}" was parsed as two path parameters, meaning a request for "/foo/bar" would return "foo"

starlite/app.py
waweber: After including the sentinel, the type becomes `Dict[Union[str, PathParamPlaceholderType], Any]` which in some places (like in `cast()` calls) makes the annotation long/dense.

starlite/asgi.py
waweber: Using an instances allows the repr to show something other than `<class 'starlite.asgi.PathParamPlaceholder'>` when being inspected in a debugger, in this case a * which is what the original implementation used to denote a param placeholder. But just referencing the class directly works just as well.

Goldziher (approved):

Goldziher:
@all-contributors add @waweber code

PR #381, opened 2022-08-20
Adds `skip_validation` flag to `Dependency` function.

### LINKED ISSUE #193: Enhancement: allow users to skip pydantic validation for dependencies.
This comes out of an issue with generics in #138.

Generics are a tricky space for pydantic to operate within, and there will always be edge cases, or they will take an opinion about resolving generic types that will not be desirable to every case, such as the case in #138.

We are in a shared, but separate space to pydantic on that front. We use pydantic under the hood to resolve/validate these types, but that's not immediately obvious (https://starlite-api.github.io/starlite/usage/2-route-handlers/0_route_handlers_concept/#). When developers are injecting dependencies into Starlite handlers, it is not necessarily going to be front-of-mind that they are going to be receiving their dependency _after_ it has worked its way through the pydantic internals, and so it is possible that this type of issue is a consistent theme for us.

Ref: https://github.com/starlite-api/starlite/issues/138#issuecomment-1157066885_

peterschutt:
Closes #193

Still WIP: needs tests, but good to go for review on logic.


starlite/constants.py
```diff
@@ -3,3 +3,5 @@
 EXTRA_KEY_REQUIRED = "required"
 EXTRA_KEY_VALUE_TYPE = "value_type"
 EXTRA_KEY_IS_PARAMETER = "is_parameter"
+EXTRA_KEY_IS_DEPENDENCY = "_is_dependency"
```
peterschutt, line 6:
```suggestion
EXTRA_KEY_IS_DEPENDENCY = "is_dependency"
```

starlite/constants.py
```diff
@@ -3,3 +3,5 @@
...
+EXTRA_KEY_SKIP_VALIDATION = "_skip_validation"
```
peterschutt, line 7:
```suggestion
EXTRA_KEY_SKIP_VALIDATION = "skip_validation"
```

starlite/utils/dependency.py
```diff
@@ -2,17 +2,31 @@
...
+    Args:
+        val(Any): value to be tested
+
+    Returns:
+        `True` if `val` is `FieldInfo` created by [`Dependency()`][starlite.params.Dependency] function.
+    """
+    return isinstance(val, FieldInfo) and bool(val.extra.get(EXTRA_KEY_IS_DEPENDENCY))
+
+
+def skip_dependency_validation(val: Any) -> bool:
```
Goldziher, line 21: A better name imho:

```suggestion
def should_skip_dependency_validation(val: Any) -> bool:
```

starlite/utils/dependency.py
```diff
@@ -2,17 +2,31 @@
...
 
-    Returns
-    -------
-    bool
+    Returns:
+        `True` if `val` is `FieldInfo` created by [`Dependency()`][starlite.params.Dependency] function and
+        `skip_validation=True` is set.
     """
-    return isinstance(val, FieldInfo) and bool(val.extra.get("is_dependency"))
+    return is_dependency_field(val) and val.extra.get(EXTRA_KEY_SKIP_VALIDATION, False)
```
Goldziher, line 32:
```suggestion
    return is_dependency_field(val) and val.extra.get(EXTRA_KEY_SKIP_VALIDATION)
```

starlite/signature.py
```diff
@@ -322,6 +322,18 @@ def signature_parameters(self) -> Generator[SignatureParameter, None, None]:
                 continue
             yield SignatureParameter(self.fn_name, name, parameter)
 
+    def skip_parameter_validation(self, parameter: SignatureParameter) -> bool:
```
Goldziher, line 325: Predicates should start in is, should etc. 

```suggestion
    def should_skip_parameter_validation(self, parameter: SignatureParameter) -> bool:
```

starlite/params.py
```diff
@@ -115,7 +116,8 @@ def Body(
 
 
 @validate_arguments(config={"arbitrary_types_allowed": True})
-def Dependency(*, default: Any = Undefined) -> Any:
+def Dependency(*, default: Any = Undefined, skip_validation: bool = False) -> Any:
     """Creates a pydantic FieldInfo instance with an extra kwargs, used for
     both parameter parsing and OpenAPI schema generation."""
-    return Field(default, is_dependency=True)
+    extra: Dict[str, Any] = {EXTRA_KEY_IS_DEPENDENCY: True, EXTRA_KEY_SKIP_VALIDATION: skip_validation}
```
Goldziher, line 122: I think this declaration is redundant, you can golf it into the function call

Goldziher (approved):

starlite/params.py
peterschutt: Without it I get:

```
starlite/params.py:123: error: Argument 2 to "Field" has incompatible type "**Dict[str, bool]"; expected "Optional[Callable[[], Any]]"  [arg-type]
starlite/params.py:123: error: Argument 2 to "Field" has incompatible type "**Dict[str, bool]"; expected "str"  [arg-type]
```

Which I assume is why you did the same thing in `Path` and `Parameter`.

PR #384, opened 2022-08-20
Issue #289- eager eval of is async

### LINKED ISSUE #289: Enhancement: Eager eval of async vs. sync guards
In `LifecycleHook` we eagerly evaluate whether a callable is async or sync, so that we don't need to do that for every call to the handler around every request.

Guards would probably benefit from the same - currently we discriminate for every call:

https://github.com/starlite-api/starlite/blob/a2c8a86d42fb347133a5c0c278ae9bc950c75cb7/starlite/handlers/base.py#L220-L228

If we itroduce a `Guard` type like `LifecycleHook` then that could probably reduce down to:

```python
     async def authorize_connection(self, connection: "HTTPConnection") -> None: 
         """ 
         Ensures the connection is authorized by running all the route guards in scope 
         """ 
         for guard in self.resolve_guards(): 
                 await guard(connection, copy(self))
```

Goldziher:
This PR resolves #289 by adding an `AsyncCallable` wrapper. I then applied if everywhere we can pre-compute this. 

Note- i removed the `LifeCycleHook` wrapper because it was redundant with this addition.


cofin (approved):

starlite/utils/sync.py
```diff
@@ -0,0 +1,29 @@
...
+        Args:
+            fn: Callable to wrap - can be any sync or async callable.
+        """
+        if is_async_callable(fn):
+            self.fn = fn
+        else:
+            self.fn = partial(run_sync, fn)  # type: ignore
+
+    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
+        return await self.fn(*args, **kwargs)  # type: ignore
```
peterschutt, line 29: I remember looking at paramspec for `LifecycleHook` but I couldn't get it to behave the way I thought it should. 

We are trying to get this type of thing to type correctly right?

```python
from starlite.utils.sync import AsyncCallable


def fn(y: int, z: str) -> str:
    ...


async def test() -> None:
    a = AsyncCallable(fn=fn)
    await a(1, "two")
```
Mypy:
```
scratch_8.py:10: error: Argument 1 to "__call__" of "AsyncCallable" has incompatible type "int"; expected <nothing>  [arg-type]
scratch_8.py:10: error: Argument 2 to "__call__" of "AsyncCallable" has incompatible type "str"; expected <nothing>  [arg-type]
```

One thing that stands out is that without binding `P` to the class the reference to `P` in signature for both `__init__` and `__call__` probably doesn't correlate with eachother.

A couple of quick examples I've found using paramspec and `__call__()` both make the class generic on `P` before referencing in signatures, e.g., https://github.com/python/mypy/issues/12595 and https://github.com/python/mypy/issues/12169.

peterschutt, line 29:
```suggestion
from functools import partial
from typing import Callable, Generic, TypeVar

from anyio.to_thread import run_sync
from typing_extensions import ParamSpec

from starlite.utils.predicates import is_async_callable

P = ParamSpec("P")
T = TypeVar("T")


class AsyncCallable(Generic[P, T]):
    __slots__ = ("args", "kwargs", "fn")

    def __init__(self, fn: Callable[P, T]):
        """Utility class that wraps a callable and ensures it can be called as
        an async function.

        Args:
            fn: Callable to wrap - can be any sync or async callable.
        """
        if is_async_callable(fn):
            self.fn = fn
        else:
            self.fn = partial(run_sync, fn)  # type: ignore

    async def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
        return await self.fn(*args, **kwargs)  # type: ignore
```
This seems to make the example in my previous comment type correctly.

peterschutt: I've PRd some changes to this branch.

Issue #385, opened 2022-08-20
Bug: SimpleCacheBackend is not thread safe


Goldziher:
## Issue

The `SimpleCacheBackend` class is too naive and cannot work properly in a threaded environment. 

## Solutioin

Update the cache backend to handle threading.

peterschutt:
closed in #386

PR #386, opened 2022-08-20
Issue #385 cache is not thread safe

Goldziher:
This PR updates the handling of cache by doing the following:

1. Update `SimplyCacheBackend` to be async and use locking.
2. Adds a `Cache` class that is generated fromt the `CacheConfig` and ensures that cache access is consistently thread safe and async, regardless of the cache backend provided (e.g. a sync redis client for example).

Note: I encountered a pydantic bug and reported it --> https://github.com/pydantic/pydantic/issues/4408, for the time being I changed `backend` to optional in `CacheConfig` and set the default in the `to_cache` function call. 

P.S this PR is pointing the other pending PR (for now).


starlite/cache/base.py
```diff
@@ -1,51 +1,152 @@
...
         """
 
     @overload  # type: ignore[misc]
     def set(self, key: str, value: Any, expiration: int) -> Any:
         ...
 
-    async def set(self, key: str, value: Any, expiration: int) -> Awaitable[Any]:
-        """Set a value in cache for a given key with a given expiration in
-        seconds.
+    async def set(self, key: str, value: Any, expiration: int) -> Any:
```
cofin, line 33: Should this be `Awaitable[Any]`?

cofin (approved):
Approved with one comment re: the `Awaitable[Any]` change

starlite/cache/base.py
Goldziher: It creates other issues. This way mypyc infers it correctly. Pylint doesn't, but it's less of an issue.

PR #388, opened 2022-08-20
FIX: Coerce compression backend from str with type hints.

cofin:
This fix adds `str` to the type hint for `CompressionConfig.backend`.  It enables IDE type hints and allows for passing `gzip`, `brotli`, or the actual enums as valid values.


tests/middleware/test_compression_middleware.py
```diff
@@ -275,6 +275,6 @@ def test_brotli_middleware_invalid_mode() -> None:
 
 def test_invalid_compression_middleware() -> None:
     try:
-        create_test_client(route_handlers=[handler], compression_config=CompressionConfig(backend="super-zip"))  # type: ignore
+        create_test_client(route_handlers=[handler], compression_config=CompressionConfig(backend="super-zip"))
```
peterschutt, line 278: There are two tests for loading from a string that both actually load using the enum:

1. https://github.com/starlite-api/starlite/blob/a160eed488c2f7d30c7b57077c9d2086a14c9625/tests/middleware/test_compression_middleware.py#L97
2. https://github.com/starlite-api/starlite/blob/a160eed488c2f7d30c7b57077c9d2086a14c9625/tests/middleware/test_compression_middleware.py#L136

starlite/config/compression.py
```diff
@@ -23,7 +23,7 @@ class CompressionConfig(BaseModel):
     'compression_config' key.
     """
 
-    backend: Union[CompressionBackend]
+    backend: Union[CompressionBackend, str]
     """
         [CompressionBackend][starlite.enums.CompressionBackend] or dotted path for
         compression backend to import.
```
peterschutt, line 29: Is `"or dotted path for compression backend to import"` still relevant? Disregard if I'm just not looking in the right place but I can't see where that takes place - at least not local to the config type. If not, perhaps you could type this with a `Literal` instead of `str`.

peterschutt (approved):
Cheers :clinking_glasses:

starlite/config/compression.py
cofin: Actually, this is a good idea.  I think we can do a `PyObject` Union here to add this functionality, but will need some additional testing.  I'll make a followup PR once I test this out.  

This PR makes it so that you can use the string value from the Enum instead of the Enum itself.  (`gzip` and `brotli` in uppercase or lowercase are the only acceptable values)

tests/middleware/test_compression_middleware.py
cofin: Good catch.  I thought I had written these in the original release but I didn't examine the test cases in enough detail.  Updated

starlite/config/compression.py
peterschutt: > Actually, this is a good idea.

Do you mean allowing the dotted path? What is the use case that supports?

I wasn't very clear but I was just suggesting to remove that from the docs if we don't have the functionality and typing it `backend: Union[CompressionBackend, Literal["gzip", "brotli"]]`

cofin: I was thinking that it might be easier for a user to specify a completely custom backend this way, but perhaps not. 
Would we even need the enum of we swapped to a Literal?

starlite/config/compression.py
```diff
@@ -23,10 +23,9 @@ class CompressionConfig(BaseModel):
     'compression_config' key.
     """
 
-    backend: Union[CompressionBackend]
+    backend: Union[CompressionBackend, Literal["gzip", "brotli"]]
```
Goldziher, line 26: Why not drop the enum entirely?

cofin: I agree, I don't know that there is much value in keeping both.  Since using the Literal save an import, I think it makes sense to drop the enum

Goldziher: cool, lets do that.

starlite/config/compression.py
```diff
@@ -59,19 +58,41 @@ class CompressionConfig(BaseModel):
...
-    def brotli_mode_must_be_valid(cls, v: Union[BrotliMode, str]) -> BrotliMode:  # pylint: disable=no-self-argument
+    @validator("backend", pre=True, always=True)
+    def backend_must_be_valid(  # pylint: disable=no-self-argument
+        cls, v: Union[CompressionBackend, Literal["gzip", "brotli"]]
+    ) -> CompressionBackend:
         """Compression Backend Validation.
 
         Args:
-            v (CompressionBackend|str): Holds the selected compression backend
+            v (CompressionBackend|Literal["gzip", "brotli"]): Holds the selected compression backend
```
Goldziher, line 68: no need to type the args since there is typing in code.

starlite/config/compression.py
```diff
@@ -59,19 +58,41 @@ class CompressionConfig(BaseModel):
...
         """
+        if isinstance(v, str):
+            try:
+                v = CompressionBackend(v)
+            except KeyError as e:
+                raise ValueError(f"{v} is not a valid compression backend mode") from e
+        return v
+
+    @validator("brotli_mode", pre=True, always=True)
+    def brotli_mode_must_be_valid(cls, v: Union[BrotliMode, str]) -> BrotliMode:  # pylint: disable=no-self-argument
```
Goldziher, line 84: same comments

Goldziher (approved):

Goldziher (approved):

Issue #343, opened 2022-08-09
Documentation: Self-contained examples in docs


peterschutt:
Objective is to ensure that the examples in the docs do not become stale as the library evolves. 

Leveraging `pymdownx.snippets` mkdocs markdown extension allows us to move the examples into their own modules that can be tested alongside the rest of the test suite.

Plan:
- [x] create `examples/` directory
- [x] create `tests/examples/`
- [x] audit and create TODO list of each self contained example in docs
- [ ] transfer all examples to a self-contained module in `examples/` and include a test suite for each in `tests/examples`

For each example in the TODO list:
- [ ] create a module in `examples/` and transfer in the example code
- [ ] create a suite for the module in `tests/examples` that tests the facets of the example that it demonstrates
- [ ] replace the example in the docs with an external reference code block, e.g., 

````md
```py title="Tests a Thing"
--8<-- "examples/test_thing.py"
```
````

<a href="https://polar.sh/litestar-org/litestar/issues/343">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/343/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/343/pledge.svg">
</picture>
</a>

Jarmos-san:
I think this is complete(?) in #1073, @provinzkraut can you clarify. Regardless, I think this issue can be closed because Janek had already implemented the script to run each examples in the documentations during the build process.

provinzkraut:
> I think this issue can be closed because Janek had already implemented the script to run each examples in the documentations during the build process.

You’re mistaken there. Yes, we have a tool to execute run examples and record their results when building the docs but:

1. This is not a replacement for tests. Failures are not reported in the CI / testing process in general 
2. Most examples aren’t yet configured for this

PR #392, opened 2022-08-21
Adds `examples/` and `examples/tests`.

### LINKED ISSUE #393: Enhancement: add GH action to test examples/hello_world.py on minimal install
We need to have the full set of extras available to run the test suite, but this can hide bugs for apps running without a subset of extras, or the minimal set of dependencies.

The task is to add a step to our GH actions workflow that installs starlite with no extras and ensures that the example in `examples/hello_world.py` runs without error.

### LINKED ISSUE #428: Dependency: Brotli missing from test dependencies
All `full` extras dependencies are available in dev-deps, except for `brotli`.

To reproduce:

```shell
$ poetry env remove python
$ poetry install
$ poetry run pytest .
```

This should mean we don't need to require `poetry install -extras full` in `CONTRIBUTING.md` and `ci.yaml`

peterschutt:
This PR includes work for a number of testing based issues.

- includes a hello-world example and test
- moves the Startup and Shutdown example from docs into examples/ and adds tests
- adds a test to CI of a minimal starlite application with no extras installed
- adds `brotli` to dev-dependencies and removes requirement to install `full` extras in testing contexts.

Start for #343 
Closes #393 
Closes #428

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
           key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --no-interaction --no-root --extras="full"
+        run: poetry install --no-interaction --no-root
       - name: Set pythonpath
         run: echo "PYTHONPATH=$PWD" >> $GITHUB_ENV
       - name: Test
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,8 +60,38 @@ jobs:
           name: coverage-xml
           path: coverage.xml
         if: matrix.python-version == '3.10'
+  test_minimal:
+    # tests the hello-world example against a minimal install
+    name: Test No Extras Installed
+    needs: validate
+    runs-on: ubuntu-latest
+    env:
+      python_version: "3.10"
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.python_version }}
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+          installer-parallel: true
+      - name: Install App Dependencies
+        run: poetry install --no-interaction --no-root --no-dev
+      - name: Install Test Dependencies
+        run: poetry run python -m pip install pytest pytest-asyncio requests
+      - name: Set pythonpath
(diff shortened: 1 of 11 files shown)


peterschutt:
I'd like to be able to write the examples using modern typing annotations, e.g., `dict[...]` not `Dict[...]` - I think most users will be running a recent version, also reduces imports keeping the examples more concise. However, having the tests in with the normal test suite fail on test collection, so skipping the tests based on version isn't enough, we need to not even collect the tests on earlier versions. 

Maybe can move the tests into another test suite, `test_examples/` or `examples/tests` and only run them in CI on 3.10. Thoughts?

Goldziher:
> I'd like to be able to write the examples using modern typing annotations, e.g., `dict[...]` not `Dict[...]` - I think most users will be running a recent version, also reduces imports keeping the examples more concise. However, having the tests in with the normal test suite fail on test collection, so skipping the tests based on version isn't enough, we need to not even collect the tests on earlier versions.
> 
> Maybe can move the tests into another test suite, `test_examples/` or `examples/tests` and only run them in CI on 3.10. Thoughts?

I concur the example tests should be separated from the library from tests. 

As for code organization- how about to move example snippets into an `examples/src` and than have `examples/test` or even have example files such as `route_example.py` and then just next to it `test_route_example`. 

I wouldnt want to have too many root level directories.

cofin (approved):

Goldziher (approved):

peterschutt:
Feedback on the addition of the minimal install test in CI please

.github/workflows/ci.yaml
```diff
@@ -52,13 +52,50 @@ jobs:
...
         with:
           name: coverage-xml
           path: coverage.xml
         if: matrix.python-version == '3.10'
+  test_minimal:
+    # tests the hello-world example against a minimal install
+    needs: validate
+    runs-on: ubuntu-latest
+    env:
+      python_version: "3.10"
```
Goldziher, line 68: why do you need a different job for this? cant you testit in the `Test Examples` stage above?

Goldziher:
> Feedback on the addition of the minimal install test in CI please

done. I dont understand why its required.

.github/workflows/ci.yaml
peterschutt: It is because the Test Examples job runs on environment with full extras: `run: poetry install --no-interaction --no-root --extras="full"`.

This is to explicitly test the app on the minimal install, so different environment required.

Goldziher: The dev dependencies are equal to full extras.

peterschutt: Well all except Brotli, but I'll admit that once I saw brotli wasn't there I assumed that the other optionals weren't there either. I'll refactor this. So just oversight that brotli isn't there?

Goldziher: Yup

Issue #350, opened 2022-08-10
Enhancement: Instrumentation hooks


Goldziher:
The suggestion here is to add several new hooks, primarly meant for instrumentation and any use case that needs access to some internals.

Hooks:

* `after_startup`: fires after the ASGI lifecycle startup finishes and calls whatever handler (sync or async) is defined by passing to it the Starlite app instance. The application will block connections until the hook returns / ends. 
* `after_shutdown`: similar to the previous one, accept it fires when shutdown is finished. 
* `after_exception`: calls a handler when an exception occurs. The handler is given the exception and it must raise an exception - this allows for example to record exceptions or transform them, but is not meant to replace an exception handler. 

Some other options that crossed my mind but im not sure about:

* `before_startup`: called as part of the ASGI lifecycle startup, before any function set in the `on_startup` callable list. Receives the app instance. 
* `before_shutdown`: same as above, but for shutdown. Receives the app instance.
* `before_routing`: called when an ASGI call is received by the app and before resolving routing etc. Receives the ASGI `scope` object

Please give input and any suggestions you may have.

lewoudar:
Hello, honestly I'm not sure to see the importance of `after_startup`, `after_shutdown` and the `before` counterparts. It is over engineering for me since we already have callbacks for startup and shutdown. For the two others I have no opinion.

cofin:
I like the idea of all of these hooks.  

On another project, I could have really benefited from a before/after shutdown hook.   If we get to this level of granularity, I think we have quite a bit of control on how/what to implement with the OpenTelemetry integrations.

At the moment, I can't think of anything I'd add to the list above, but will update accordingly if something comes to mind.

peterschutt:
How about before/after request hooks that are guaranteed to be called, configured at the app level and separate from handler resolution. 

Use-case is this:

```py
async def session_after_request(response: Response) -> Response:
    """
    Passed to `Starlite.after_request`.

    Inspects `response` to determine if we should commit, or rollback the database
    transaction.

    Finally, calls `remove()` on the scoped session.

    Parameters
    ----------
    response : Response
        The outgoing response.

    Returns
    -------
    Response
        Passed through from input assuming that commit/roll back does not error.
    """
    if 200 <= response.status_code < 300:
        await AsyncScopedSession.commit()
    else:
        await AsyncScopedSession.rollback()
    await AsyncScopedSession.remove()
    return response
```
[...]

PR #399, opened 2022-08-23
Issue #398 extended plugin protocol

### LINKED ISSUE #398: Enhancement: Extend Plugin Protocol
The idea is to extend the plugin protocol to include support for hooks, middleware etc.

Goldziher:
This PR extends the plugin protocol by adding an `on_app_init` method. 

Resolves #398

### DIFF
--- a/starlite/plugins/base.py
+++ b/starlite/plugins/base.py
@@ -12,7 +12,7 @@ from typing import (
     Union,
 )
 
-from typing_extensions import Protocol, get_args, runtime_checkable
+from typing_extensions import Protocol, TypeGuard, get_args, runtime_checkable
 
 if TYPE_CHECKING:
     from pydantic import BaseModel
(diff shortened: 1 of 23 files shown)


starlite/plugins/base.py
```diff
@@ -12,45 +12,258 @@
...
+        ResponseCookies,
+        ResponseHeadersMap,
+        ResponseType,
+    )
+
 ModelT = TypeVar("ModelT")
 
 
 @runtime_checkable
 class PluginProtocol(Protocol[ModelT]):  # pragma: no cover
```
cofin, line 40: presumably, we'd add the additional hooks referenced in #350 here as well?

Goldziher: Yup. 

Another alternative is to pass the Starlite instance to the plugin before finishing init. Can make plugins simpler, but  not sure.

cofin (approved):

Issue #397, opened 2022-08-23
Enhancement: Integrate Authentication with Swagger Docs


cofin:
Enable OpenAPI generated Swagger docs should be aware of API authentication mechanism to allow for easy testing from the web browser.

Here is a sample from a FastAPI project that shows the desired functionality:
<img width="751" alt="Screen Shot 2022-08-22 at 9 16 00 PM" src="https://user-images.githubusercontent.com/204685/186054390-efc6f888-f3a3-4375-be0d-15e31a9b1f9a.png">
<img width="677" alt="Screen Shot 2022-08-22 at 9 15 30 PM" src="https://user-images.githubusercontent.com/204685/186054468-7dee5009-f905-4fbd-be95-410a445e522b.png">

After authentication, the user should be able to invoke services with the credentials. 

API Key & OAUTH2 support are included out of the box.  Could be extended to support OpenID Connect

toudi:
I stumbled upon this issue because I tried to generate openapi specs and realized that something wasn't right. Anyhow, I was able to move just one step further with this piece of code:

```python
from pydantic_openapi_schema.v3_1_0.security_scheme import SecurityScheme

TokenSecurity: SecurityScheme = SecurityScheme(
    type="apiKey", description="Token based authentication", schema="bearer"
)

class CustomOpenAPIConfig(OpenAPIConfig):
    def to_openapi_schema(self) -> OpenAPI:
        schema = super().to_openapi_schema()
        schema.components = Components(securitySchemes={"BearerAuth": TokenSecurity})

        return schema

app = Starlite(
    # ....
    openapi_config=CustomOpenAPIConfig( .... ),
)

```

however I still do not understand how am I supposed to decorate particular views with `security` attribute. the openAPI docs specify this:

(via https://swagger.io/docs/specification/authentication/ )

```yaml
paths:
  /billing_info:
    get:
      summary: Gets the account billing info
      security:
        - OAuth2: [admin]   # Use OAuth with a different scope
```
however I looked in Router, the `get` decorator and I couldn't find this parameter. Is it because this is simply not implemented yet ?

Goldziher:
> I stumbled upon this issue because I tried to generate openapi specs and realized that something wasn't right. Anyhow, I was able to move just one step further with this piece of code:
> 
> ```python
> from pydantic_openapi_schema.v3_1_0.security_scheme import SecurityScheme
> 
> TokenSecurity: SecurityScheme = SecurityScheme(
[...]

PR #400, opened 2022-08-23
implement security and components properties to support authorization…

toudi:
… in swagger UI

solves #397 

I would also like to add some documentation and unit tests, but I just don't know how to do it. could you guide me somehow?

here's a brief description on how to use it:

first, create a default authorization method (I'm using http bearer token)
```python
from pydantic_openapi_schema.v3_1_0.security_scheme import SecurityScheme

TokenSecurity: SecurityScheme = SecurityScheme(
    type="http",
    scheme="bearer",
)
```

next, create some routes:

```python
from urllib.request import Request
from starlite import Router, get

@get("/")
def hello_world(request: Request) -> str:
    return "hello, world!"


@get("/public", security=[])
def public(request: Request) -> str:
    return "public endpoint"

hello_router = Router(path="/hello", route_handlers=[hello_world, public])
```

Finally, construct the app object:
```python
app = Starlite(
 route_handlers=[hello_router],
 openapi_config=OpenAPIConfig(
    components=Components(securitySchemes={"BearerAuth": TokenSecurity}),
    security=[{"BearerAuth": []}],
 )
)
```


cofin:
Thanks for the contribution!  

One way you could handle this is to use reference the `openapi_schema` from your application object.   I'm sure there are others.  

Here is out I export the schema to a dict.
```python

@cli.command(
    name="export-openapi-schema",
)
def export_api_schema(
    export_format: SchemaExportFormat = SchemaExportFormat.JSON,
) -> None:
    """Export OpenAPI schema"""
    from opdba.core.asgi import application

    console.print("Exporting API Schema")
    schema = application.openapi_schema
    console.print_json(schema.json())


```

Your test would want to assert that the block you'd expect to have the OpenAPI security configuration to exist.  

@peterschutt , @Goldziher  may have a much easier approach thought.

Thanks!

toudi:
my idea was to hook into starlite/tests/openapi/test_schema.py but I just don't know how yet. as for the examples, I just couldn't find any - maybe they're added in the separate repository with the documentation or something :)

Goldziher:
Hi @toudi - and thanks. I will review this in several hours and offer some advice.

Goldziher:
So, checkout how tags are resolved recursively - we should support declaring security on controller and router as well.

starlite/handlers/http.py
```diff
@@ -663,6 +667,7 @@ def __init__(
             response_description: Text used for the route's response schema description section.
             summary: Text used for the route's schema summary section.
             tags: A list of string tags that will be appended to the OpenAPI schema.
+            security: A list of strings that describe authentication methods that override the default ones.
```
peterschutt, line 670: This doesn't match the type declared in signature.

starlite/handlers/http.py
```diff
@@ -783,6 +790,7 @@ def __init__(
...
```
peterschutt, line 793: Again, seems to be all the doc-strings are out of date so I won't flag the others.

peterschutt (changes requested):
Thanks @toudi!

As @Goldziher mentioned, the configs that are available on the handlers are typically available on each level of the application, and are resolved down to the handler so the closest definition to the handler applies. Jump into discord if you want further guidance on that.

Still needs tests and docs.

I put a minimal example together based on your comments, which I'll copy in below. Please feel free to update to what you think would be the appropriate minimal example for documentation purposes and I'll incorporate with work on #392.

peterschutt:
Here's that example:

```py
from pydantic_openapi_schema.v3_1_0 import Components, SecurityScheme
from starlite import Router, Starlite, get
from starlite.config import OpenAPIConfig

TokenSecurity: SecurityScheme = SecurityScheme(
    type="http",
    scheme="bearer",
)


@get("/")
def hello_world() -> str:
    """Protected route"""
    return "Hello, private!"


@get("/public", security=[])
def public() -> str:
    """Public route"""
    return "Hello, public!"


hello_router = Router(path="/hello", route_handlers=[hello_world, public])


app = Starlite(
    route_handlers=[hello_router],
    openapi_config=OpenAPIConfig(
        title="OpenAPI Authn Hello World",
        version="1.0.0",
        components=Components(securitySchemes={"BearerAuth": TokenSecurity}),
        security=[{"BearerAuth": []}],
        use_handler_docstrings=True,
    ),
)
```

peterschutt:
@cofin how does this compare to the amount of config required to get this to work in fast?

cofin:
This feels pretty comparable.  Here is the most basic example for FastAPI:

```python
from fastapi import Depends, FastAPI
from fastapi.security import OAuth2PasswordBearer

app = FastAPI()

oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")


@app.get("/items/")
async def read_items(token: str = Depends(oauth2_scheme)):
    return {"token": token}
 ```

FastAPI also implements an API Key Header/Cookie and simple HTTP auth

peterschutt:
> This feels pretty comparable. Here is the most basic example for FastAPI:
> 
> ```python
> from fastapi import Depends, FastAPI
> from fastapi.security import OAuth2PasswordBearer
> 
> app = FastAPI()
> 
> oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
> 
> 
> @app.get("/items/")
> async def read_items(token: str = Depends(oauth2_scheme)):
>     return {"token": token}
> ```
> 
> FastAPI also implements an API Key Header/Cookie and simple HTTP auth

I see, so depending on the auth method somewhere in the dependency chain both provides the authentication from the request and configures the OpenAPI. Whereas we will have separate flows for those if I understand all this correctly? I just wonder if we can't get the doc generation more integrated with the middleware somehow, so that specifying the middleware configures the docs.

Goldziher:
@cofin @toudi @peterschutt, i added an example to the `starlite-jwt` package, where i  implemented SecuritySchema already. See how it looks-> https://github.com/starlite-api/jwt-auth

Goldziher:
@toudi please install pre-commit and run it locally (instructions are in CONTRIBUTING.md). In particular, the failure now means you need to adjust the pylint configs in pyproject.toml in the folder's root.

toudi:
ok, will do, I just need to find some time - hopefully tonight :)

cofin:
> > This feels pretty comparable. Here is the most basic example for FastAPI:
> > ```python
> > from fastapi import Depends, FastAPI
> > from fastapi.security import OAuth2PasswordBearer
> > 
> > app = FastAPI()
> > 
> > oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
> > 
> > 
> > @app.get("/items/")
> > async def read_items(token: str = Depends(oauth2_scheme)):
> >     return {"token": token}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > FastAPI also implements an API Key Header/Cookie and simple HTTP auth
> 
> I see, so depending on the auth method somewhere in the dependency chain both provides the authentication from the request and configures the OpenAPI. Whereas we will have separate flows for those if I understand all this correctly? I just wonder if we can't get the doc generation more integrated with the middleware somehow, so that specifying the middleware configures the docs.

Correct - Somewhere in the dependency chain you'll have something like above.  I agree it would be better for us to specify it in the middleware.  I'll test the `starlite-jwt` updates today.

What updates are remaining on the PR at this point?

cofin:
No worries at all!  Thanks for the work on this.

toudi:
I've added unit tests and fixed the docstrings. Ideally, I was hoping to write some test that would instantiate a test HTTP client, then try to access a protected route and make sure that it would return 400 but then I realized that this would be implementation-dependent - means that there's a different way of checking an AuthToken from, say, a OAuth2 or jwt and this would have to be combined with middleware implementation which seems heavy.

Anyhow, I hope that this is going in the good direction as I'm definetely not an expert on OpenAPI :)

Goldziher:
> I've added unit tests and fixed the docstrings. Ideally, I was hoping to write some test that would instantiate a test HTTP client, then try to access a protected route and make sure that it would return 400 but then I realized that this would be implementation-dependent - means that there's a different way of checking an AuthToken from, say, a OAuth2 or jwt and this would have to be combined with middleware implementation which seems heavy.
> 
> Anyhow, I hope that this is going in the good direction as I'm definetely not an expert on OpenAPI :)

Thanks @toudi - I'll check this soon. Or maybe @peterschutt can?

peterschutt:
I'll be able to have a look in an hour or so :+1:

Goldziher (approved):
Ok, this implementation doesnt factor in layering - but i can implement this on main. So ill merge this and work on this further. Thanks for your contribution @toudi

Goldziher:
@all-contributors add @toudi code

PR #401, opened 2022-08-24
testing pydantic v1.10.0a2

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -23,8 +23,8 @@ sniffio = ">=1.1"
 typing-extensions = {version = "*", markers = "python_version < \"3.8\""}
 
 [package.extras]
-doc = ["packaging", "sphinx-rtd-theme", "sphinx-autodoc-typehints (>=1.2.0)"]
-test = ["coverage[toml] (>=4.5)", "hypothesis (>=4.0)", "pytest (>=7.0)", "pytest-mock (>=3.6.1)", "trustme", "contextlib2", "uvloop (<0.15)", "mock (>=4)", "uvloop (>=0.15)"]
+doc = ["packaging", "sphinx-autodoc-typehints (>=1.2.0)", "sphinx-rtd-theme"]
+test = ["contextlib2", "coverage[toml] (>=4.5)", "hypothesis (>=4.0)", "mock (>=4)", "pytest (>=7.0)", "pytest-mock (>=3.6.1)", "trustme", "uvloop (<0.15)", "uvloop (>=0.15)"]
 trio = ["trio (>=0.16)"]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Goldziher:
> Hi, please can you kick off tests so I can check if v1.10 will break anything.

sure thing.

Goldziher:
we actually have a pending issue for this --> https://github.com/starlite-api/starlite/issues/394

Goldziher:
are you certain it installed? because the lock file has not been modified.

samuelcolvin:
Oh, :facepalm: I was working through quite a few libraries.

Goldziher:
> hopefully that's working, can you kick off tests.
> 
> BTW, the project will be much more attractive to contribute to if you remove the "awaiting approval", see [here](https://twitter.com/samuel_colvin/status/1562374350287867904).

I changed the settings. Since you are not new user on github, you should be golden 😉 

I'd love some contributions from you ofc. So with that hurdle out of the way... lol

samuelcolvin:
Okay, this is actually failing. Could it be related to https://github.com/pydantic/pydantic/issues/3061?

Goldziher:
> Okay, this is actually failing. Could it be related to https://github.com/pydantic/pydantic/issues/3061?

I'll have to debug for ascertain what's the failure. Will do this shortly.

samuelcolvin:
I'm looking into it now, seems it's not https://github.com/pydantic/pydantic/issues/3061.

Anything you find would be helpful.

While investigating that issue, I've found another that only applies when pydantic is not compiled, trying to isolate it now.

Goldziher:
So, looking at this test:

```python
from starlette.status import HTTP_201_CREATED

from starlite import Body, RequestEncodingType, post
from starlite.testing import create_test_client
from tests.kwargs import Form


def test_request_body_url_encoded() -> None:
    body = Body(media_type=RequestEncodingType.URL_ENCODED)

    test_path = "/test"
    data = Form(name="Moishe Zuchmir", age=30, programmer=True).dict()

    @post(path=test_path)
    def test_method(data: Form = body) -> None:
        assert isinstance(data, Form)

    with create_test_client(test_method) as client:
        response = client.post(test_path, data=data)
```

The model called `Form` is defined as:

```python


class Form(BaseModel):
    name: str
    age: int
    programmer: bool
```

And we are generating a pydantic model from the route handler that has a `__field__` attribute like this:

Screenshot 2022-08-24 at 18 08 10

Yes, when the generated model is called using the `data` kwarg, a validation exception is raised:

Screenshot 2022-08-24 at 18 05 00

This is raised here:

```python
   @classmethod
    def parse_values_from_connection_kwargs(
        cls, connection: Union[Request, WebSocket], **kwargs: Any
    ) -> Dict[str, Any]:
        """Given a dictionary of values extracted from the connection, create
        an instance of the given SignatureModel subclass and return the parsed
        values.

        This is not equivalent to calling the '.dict'  method of the
        pydantic model, because it doesn't convert nested values into
        dictionary, just extracts the data from the signature model
        """
        try:
            signature = cls(**kwargs)
            return {key: signature.resolve_field_value(key) for key in cls.__fields__}
        except ValidationError as exc:
            raise cls.construct_exception(connection, exc) from exc
```

samuelcolvin:
Also, see https://github.com/pydantic/pydantic/pull/4432.

That issue is probably caused by https://github.com/pydantic/pydantic/issues/4253, I'll look into once I've fixed the other one.

samuelcolvin:
@Goldziher the remaining bug should be fixed in #402.

I think pydantic is now correct and starlite is incorrect, hence I'd prefer to keep the change in pydantic and fix compatibility in starlite.

Let me know what you think.

Goldziher:
> @Goldziher the remaining bug should be fixed in #402.
> 
> I think pydantic is now correct and starlite is incorrect, hence I'd prefer to keep the change in pydantic and fix compatibility in starlite.
> 
> Let me know what you think.

great, thanks. I merged your PR.

samuelcolvin:
All passing, thanks so much for your help.

PR #402, opened 2022-08-24
remove empty aliases

samuelcolvin:
As per https://github.com/pydantic/pydantic/issues/4253 (change in https://github.com/pydantic/pydantic/pull/4252) an empty alias `''` is now treated differently from `None` - e.g. no alias.

Therefore for starlite to be be compatible with pydantic v1.10, we need to remove these empty alias values.

This should be backwards compatible with earlier versions of pydantic since until v.10 an empty alias was treated as identical to `''`.

### DIFF
--- a/starlite/params.py
+++ b/starlite/params.py
@@ -90,7 +90,6 @@ def Parameter(
     extra.update(value_type=value_type)
     return Field(
         default,
-        alias="",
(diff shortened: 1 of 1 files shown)


Goldziher:
@all-contributors add @samuelcolvin code

PR #408, opened 2022-08-25
Browser refuses set cookie if the size of session is larger than max-size limit 4096 allowed by the browser

infohash:
# PR Checklist

- [✅ ] Have you followed the guidelines in `CONTRIBUTING.md`?
- [ ✅] Have you got 100% test coverage on new code?

# Minimal Reproducible Example
Run it as is. Visit `/` which redirects to `/health`. The assertion error will be raised if the cookies were refused by the browser.

```python
import secrets

import uvicorn
from starlette.middleware.sessions import SessionMiddleware
from starlette.status import HTTP_308_PERMANENT_REDIRECT
from starlite import DefineMiddleware, get, Request, Starlite
from starlite.datastructures import Redirect

# Set size of cookie value here, currently set to 1512 bytes to fail the
# assertion.
overflow = secrets.token_hex(1512)


@get(path='/', status_code=HTTP_308_PERMANENT_REDIRECT)
async def root(request: Request) -> Redirect:
    # Fill the session with large string to exceed the size of max-size limit
    # of the cookie in browser which is 4096 bytes.
    request.session['overflow'] = overflow
    return Redirect(path='/health')


@get(path='/health')
async def health_check(request: Request) -> str:
    # If the browser refused cookie in the response of the previous route
    # handler, this assertion will fail. This assertion will only pass if the
    # size of overflow is reduced to keep the session size less than 4096
    # bytes.
    assert tuple(request.session.items()) == (('overflow', overflow),)
    return 'health checked'

app = Starlite(route_handlers=[health_check, root],
               middleware=[DefineMiddleware(SessionMiddleware, secret_key=secrets.token_hex(16))],
               debug=True)

if __name__ == '__main__':
    uvicorn.run(app)
```

# What's Working?
Response is registered by the browser but it refuses to set cookie in the response if it exceeds the max 4096 size limit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ pyyaml = "*"
 requests = { version = "*", optional = true }
 starlette = "*"
 typing-extensions = "*"
-pre-commit = "^2.20.0"
+itsdangerous = "^2.1.2"
 
 [tool.poetry.dev-dependencies]
 freezegun = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ pyyaml = "*"
 requests = { version = "*", optional = true }
 starlette = "*"
 typing-extensions = "*"
+pre-commit = "^2.20.0"
 
 [tool.poetry.dev-dependencies]
 freezegun = "*"
--- a/tests/middleware/test_middleware_handling.py
+++ b/tests/middleware/test_middleware_handling.py
@@ -1,4 +1,5 @@
 import logging
+import secrets
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, List, cast
 
 import pytest
@@ -6,6 +7,7 @@ from pydantic import BaseModel
 from starlette.middleware import Middleware
 from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.middleware.cors import CORSMiddleware
+from starlette.middleware.sessions import SessionMiddleware
 from starlette.middleware.trustedhost import TrustedHostMiddleware
 
 from starlite import (
@@ -205,3 +207,23 @@ def test_middleware_call_order() -> None:
         client.get("/router/controller/handler")
 
         assert results == [0, 1, 2, 3, 4, 5, 6, 7]
+
+
+@pytest.mark.xfail(reason="Cannot set cookie of size greater than the allowed max-size.", strict=True)
+def test_session_middleware_should_fail_setcookie() -> None:
+    """Test if size of set cookie is greater than max-size limit of cookie
+    allowed by the browsers."""
+
+    @get("/session")
+    def session_handler(request: Request) -> None:
+        """Mock route handler to set session."""
+        # Fill the session with large string to exceed the max-size limit.
+        request.session["mock_cookie"] = secrets.token_hex(1514)
+
+    client = create_test_client(
+        route_handlers=[session_handler],
+        middleware=[DefineMiddleware(SessionMiddleware, secret_key=secrets.token_hex(16))],
+    )
+    response = client.get("/session")
+    assert response.status_code == 200
+    assert len(response.cookies["session"]) < 4096


peterschutt:
Seems that Flask [issues a warning](https://stackoverflow.com/a/66497589/6560549) for same.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ pyyaml = "*"
 requests = { version = "*", optional = true }
 starlette = "*"
 typing-extensions = "*"
+itsdangerous = "^2.1.2"
```
peterschutt, line 47: What is this for?

infohash: Session Middleware dependency.

infohash:
The size of the session is not even close to 4096 bytes. Even by setting a string of size 1512 bytes in session, when the session middleware sets it as a cookie in the response header, the set cookie field size grows to 4100+ bytes.

pyproject.toml
peterschutt: Oh I see, maybe we should just depend on `starlette[full]` for test environment.

infohash: Even for the user environment, `itsdangerous` is required to use `SessionMiddleware`.

peterschutt: I don't see any justification in this PR for adding itsdangerous as a prod dependency for everyone who uses starlite.

peterschutt:
> The size of the session is not even close to 4096 bytes. Even by setting a string of size 1512 bytes in session, when the session middleware sets it as a cookie in the response header, the set cookie field size grows to 4100+ bytes which is then rejected by the browser. Run the example, you will see the failure.

I see, have you done any debugging you can share?

peterschutt:
Are you sure it is not this? https://github.com/encode/starlette/issues/1259

Here's the response headers for the redirect that I get with your example:

```
HTTP/1.1 308 Permanent Redirect
date: Thu, 25 Aug 2022 21:18:32 GMT
server: uvicorn
content-length: 0
location: /health
set-cookie: session=eyJvdmVyZmxvdyI6ICIifQ==.YwfnKQ.N4BXrcIduaYMFgt69HmI2hVBZkE; path=/; Max-Age=1209600; httponly; samesite=lax
```

infohash:
Yes, the problem is coming from here:

https://github.com/encode/starlette/blob/31164e346b9bd1ce17d968e1301c3bb2c23bb418/starlette/middleware/sessions.py#L62-L63

Based on the example, these are the changes in its size.
```python
>>> len(json.dumps(scope["session"]))
3040
>>> len(b64encode(json.dumps(scope["session"]).encode("utf-8")))
4056
data = b64encode(json.dumps(scope["session"]).encode("utf-8"))
>>>len(self.signer.sign(data))
4091
>>> 4091 - 3040
1051  # Increased by
```

I reproduced the same situation in flask with the same key value pairs in `flask.session`.
https://github.com/pallets/flask/blob/36af821edf741562cdcb6c60d63f23fa9a1d8776/src/flask/sessions.py#L409
```python
>>> len(json.dumps(flask.session))
3040
>>> len(self.get_signing_serializer(app).dumps(dict(session)))
2362
```

It's `starlette.middlware.SessionMiddleware`'s bug, not at starlite.

--

The response headers will show the set cookie field but the browser will silently ignore it for exceeding the max-size limit. If the assertion is failing that means the browser has ignored the cookie.

peterschutt:
> Even by setting a string of size 1512 bytes in session

```python-console
>>> import secrets
>>> overflow = secrets.token_hex(1512)
>>> len(overflow.encode("utf-8"))
3024
>>> 
```
So the original value you are setting in the session is twice as large as you say it is in the example.

But I get the same error if I set `session["overflow"] = "a"` which I think is to do with the starlette error I linked above.

peterschutt:
So what should we be doing here in your opinion?

infohash:
As SessionMiddleware is not the extension of Starlite, I'll upstream the fix.

PR #410, opened 2022-08-26
Issue #400: Add support for OpenAPI Security Scheme

Goldziher:
This PR includes the code contributed in #400 by @toudi, I'm extending it to finish the implementation quickly because we have a bug in the `starlite-jwt` due to an oversight on my part...

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -132,6 +132,7 @@ class Starlite(Router):
         route_handlers: List[ControllerRouterHandler],
         static_files_config: Optional[Union[StaticFilesConfig, List[StaticFilesConfig]]] = None,
         template_config: Optional[TemplateConfig] = None,
+        security: Optional[List[Dict[str, List[str]]]] = None,
         tags: Optional[List[str]] = None,
     ):
         """The Starlite application.
@@ -177,6 +178,7 @@ class Starlite(Router):
                 function decorated by the route handler decorators.
             static_files_config: An instance or list of [StaticFilesConfig][starlite.config.StaticFilesConfig]
             template_config: An instance of [TemplateConfig][starlite.config.TemplateConfig]
+            security: A list of dictionaries that will be added to the schema of all route handlers in the application.
             tags: A list of string tags that will be appended to the schema of all route handlers under the application.
         """
 
@@ -213,6 +215,7 @@ class Starlite(Router):
             response_cookies=response_cookies,
             response_headers=response_headers,
             route_handlers=route_handlers,
+            security=security,
             tags=tags,
         )
         self._init = True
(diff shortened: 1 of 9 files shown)


starlite/openapi/path_item.py
```diff
@@ -58,23 +97,7 @@ def create_path_item(
                 ),
                 requestBody=request_body,
                 parameters=parameters,  # type: ignore[arg-type]
+                security=security or None,
```
peterschutt, line 100:
```suggestion
                security=security,
```
Same as tags.

starlite/app.py
```diff
@@ -132,6 +132,7 @@ def __init__(
         route_handlers: List[ControllerRouterHandler],
         static_files_config: Optional[Union[StaticFilesConfig, List[StaticFilesConfig]]] = None,
         template_config: Optional[TemplateConfig] = None,
+        security: Optional[List[Dict[str, List[str]]]] = None,
```
peterschutt, line 135:
```python
...
from pydantic_openapi_schema.v3_1_0.security_requirement import SecurityRequirement
...
```
```suggestion
        security: Optional[List[SecurityRequirement]] = None,
```
Should give us a cross-reference to the type in P.O.S (eek about the acronym!) where there is a good description of its semantics.

starlite/app.py
```diff
@@ -177,6 +178,7 @@ def __init__(
                 function decorated by the route handler decorators.
             static_files_config: An instance or list of [StaticFilesConfig][starlite.config.StaticFilesConfig]
             template_config: An instance of [TemplateConfig][starlite.config.TemplateConfig]
+            security: A list of dictionaries that will be added to the schema of all route handlers in the application.
```
peterschutt, line 181:
```suggestion
            security: A list of dictionaries that will be added to the schema of all route handlers in the application. See [SecurityRequirement][pydantic_openapi_schema.v3_1_0.security_requirement.SecurityRequirement] for details.
```

starlite/openapi/path_item.py
```diff
@@ -44,9 +81,11 @@ def create_path_item(
                 request_body = create_request_body(
                     field=handler_fields["data"], generate_examples=create_examples, plugins=plugins
                 )
+
+            tags, security = extract_layered_values(route_handler)
             operation = Operation(
                 operationId=route_handler.operation_id or handler_name,
-                tags=extract_tags_from_route_handler(route_handler),
+                tags=list(set(tags)) if tags else None,
```
peterschutt, line 88:
```suggestion
                tags=tags,
```
The ternary and deduplication is done in return from `extract_layered_values()` by the looks.

peterschutt (approved):
Made a suggestion to replace the `Dict[str, list[str]]` with `SecurityRequirement` and include a cross-ref in the docs for the `security` parameters. I only suggested the change once but think it should be done across all the references.

starlite/app.py
Goldziher: will do.

cofin (approved):

Issue #412, opened 2022-08-26
Enhancement: A Better Session Cookie Management 


infohash:
We need a more optimal approach on managing session cookies. The problem with `starlette.middleware.session.SessionMiddleware` is that it does not create multiple cookies if there is a large session data to store that cannot be fit into one `Set-Cookie` HTTP response header. The max-size limit per cookie allowed by the browsers is `4096 bytes` beyond which browsers silently ignores the `Set-Cookie` header.

Adding to the problem of storing entire session data into just one cookie, `base64` encoder further increases the size of `Set-Cookie` header. This causes the issue with storing Oauth2 and OIDC tokens and their metadata whose size is already too large to store in one `Set-Cookie` header. If the user is using multiple Identity Providers in the app, that will multiply the session data.

# What we need

1. Support for splitting cookies if the size of the session is greater than 4096 bytes. Browsers allows 150-180 cookies per domain and max-size of 4096 bytes per cookie.

[Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)
>  To send multiple cookies, multiple `Set-Cookie` headers should be sent in the same response. 

2. `base64` encoding of `Set-Cookie` header is recommended by [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648).

> To maximize compatibility with user agents, servers that wish to
   store arbitrary data in a cookie-value SHOULD encode that data, for
   example, using Base64.

But we can also use custom encoders that will not increase the size that much. Nice to have if we can compress the size of encoded string.
[What are allowed characters in cookies?](https://stackoverflow.com/a/1969339/6538535)
> In practice you cannot use non-ASCII characters in cookies at all. If you want to use Unicode, control codes or other arbitrary byte sequences, the cookie_spec demands you use an ad-hoc encoding scheme of your own choosing and suggest URL-encoding (as produced by JavaScript's encodeURIComponent) as a reasonable choice.

[Flask seems to be using a custom encoder](https://github.com/pallets/flask/blob/36af821edf741562cdcb6c60d63f23fa9a1d8776/src/flask/sessions.py#L353) which actually decreases the size after encoding and signing which is impressive but also complex.

---

# Notes

1. Splitting the session into multiple cookies solves the size problem so it is fine to use `base64`.

2. We can upstream the feature 1 to `SessionMiddleware` but if you have more ideas to improve session management further, we can offer a middleware out of box that manages the session.

# References For Cookie Specification

[...]

PR #423, opened 2022-08-28
Issue #412: add session middleware

Goldziher:
This is a draft PR - adding session middleware. It addresses #412 .


starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+    domain: Optional[str] = None
+    """Domain for which the cookie is valid."""
+    secure: Optional[bool] = None
+    """Https is required for the cookie."""
+    httponly: Optional[bool] = None
+    """Forbids javascript to access the cookie via 'Document.cookie'."""
+    samesite: Literal["lax", "strict", "none"] = "lax"
+    """Controls whether or not a cookie is sent with cross-site requests. Defaults to 'lax'."""
+
+    @validator(always=True)
```
infohash, line 72: Missing name of the validating field `secret`.

starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+        Args:
+            data: Datum to dump
+
+        Returns:
+            List of encoded bytes string of a maximum length equal to the 'CHUNK_SIZE' constant.
+
+        """
+        serialized = dumps(data)
+        nonce = urandom(NONCE_SIZE)
+        encrypted = self.aesgcm.encrypt(nonce, nonce + serialized)
```
infohash, line 119: Missing third argument associated data. See [cryptography.hazmat.primitives.ciphers.aead.AESGCM](https://cryptography.io/en/latest/hazmat/primitives/aead/#cryptography.hazmat.primitives.ciphers.aead.AESGCM)

We don't have associated data to authenticate so we can set its value to a arbitrary byte string b'session'.

```python
aad = b'session'
encrypted = self.aesgcm.encrypt(nonce, nonce + serialized, aad)
```

starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Args:
+            data: A list of strings derived from the request's session cookie(s).
+
+        Returns:
+            A deserialized session value.
+        """
+        decoded = b64decode("".join(data))
+        nonce = decoded[:NONCE_SIZE]
+        decrypted = self.aesgcm.decrypt(nonce, decoded)
```
infohash, line 142: Same, set `associated_data` to `None`.

decrypted = self.aesgcm.decrypt(nonce, decoded, associated_data=None)

starlite/middleware/session.py
Goldziher: fixed

starlite/middleware/session.py
Goldziher: its typed as optional, so this is ok. Ill update it though to be explicit.

starlite/middleware/session.py
Goldziher: same

starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+                headers = MutableHeaders(scope=message)
+                if should_vacate_session:
+                    cookie_params = self.config.dict(exclude_none=True, exclude={"secret", "max_age"})
+                    headers.append("Set-Cookie", Cookie(value="null", expires=0, **cookie_params).to_header())
+                else:
+                    data = self.dump_data(scope.get("session"))
+                    cookie_params = self.config.dict(exclude_none=True, exclude={"secret"})
+                    if len(data) == 1:
+                        cookie_params.pop("key")
+                        headers.append("Set-Cookie", Cookie(value=data[0].decode("utf-8"), **cookie_params).to_header())
```
infohash, line 180: Missing `key` argument for `Cookie` class.

```python
headers.append("Set-Cookie", Cookie(key=self.config.key, value=data[0].decode("utf-8"), **cookie_params).to_header())

File: starlite/middleware/session.py [hunk 5]
```diff
@@ -0,0 +1,217 @@
...
+        serialized = dumps(data)
+        nonce = urandom(NONCE_SIZE)
+        encrypted = self.aesgcm.encrypt(nonce, nonce + serialized)
+        encoded = b64encode(encrypted)
+
+        if len(encoded) <= CHUNK_SIZE:
+            return [encoded]
+
+        result: List[bytes] = []
+        for i in range(0, len(encrypted), CHUNK_SIZE):
```
infohash, line 126: The variable here should be `encoded`, not `encrypted`. Oh and we can also use list comprehension here.

starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+    from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+except ImportError as e:
+    raise MissingDependencyException("cryptography is not installed") from e
+
+if TYPE_CHECKING:
+    from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+ONE_DAY_IN_SECONDS = 60 * 60 * 24
+NONCE_SIZE = 12
+CHUNK_SIZE = 4096
```
infohash, line 34: This should be less than 4096 to accomodate for the word "session=", "session-1=", "session-2", ...

starlite/middleware/session.py
```diff
@@ -0,0 +1,217 @@
...
+                        cookie_params.pop("key")
+                        headers.append("Set-Cookie", Cookie(value=data[0].decode("utf-8"), **cookie_params).to_header())
+                    else:
+                        for i, datum in enumerate(data):
+                            cookie_params.pop("key")
+                            headers.append(
+                                "Set-Cookie",
+                                Cookie(
+                                    value=datum.decode("utf-8"), key=f"{self.config.key}-{i}", **cookie_params
+                                ).to_header(),
```
infohash, line 188: Same thing , `....to_header().strip('Set-Cookie: '))`

infohash (changes requested):
I have QA'ed and tested all conditions and all line of code. Here are the suggested fixes that when I added during QA, it made the middleware fully functional.

starlite/middleware/session.py
Goldziher: done

starlite/middleware/session.py
Goldziher: done

infohash:
## Requirements

1. from cryptography.hazmat.primitives.ciphers.aead import AESGCM raises `InvalidTag` exception when the MAC validation fails. Decryption block should handle this exception. Clear the session if the exception is raised.
2. Need one method to validate max-age of cookies. Clear the session if expired.

Goldziher:
@infohash thanks for the testing - im sending you an invite to the organization. It will give you the ability to push to the branch. You will note I started testing - one test is failing. If you feel like puhsing most tests, fixing this up etc. It would be nice. Also you other fixes

> ## Requirements
> 1. `cryptography.hazmat.primitives.ciphers.aead import AESGCM` raises `InvalidTag` exception when the MAC validation fails. Decryption block should handle this exception. Clear the session if the exception is raised.
> 2. Need one method to validate max-age of cookies. Clear the session if expired.

yup. Do you want to take care of this? Also tests are partial. I can continue working on this tomorrow - but feel free to push commits to this branch.

infohash:
Thank you so much for the invite. I'll take care of the rest of the test cases and requirement.

infohash:
Another thing, the `additional_data` will contain 2 things.

1. The timestamp of when the session was encrypted so that we can verify when the cookie was set.
2. Max-age of the cookie to verify its validity from the time of when it was set.

**A**dditional **D**ata in AE**AD** is only required to be integrity protected, not encrypted so we can append AD to the ciphertext.

Additionally, we can put cookie path in AD.

pyproject.toml
```diff
@@ -62,6 +62,7 @@ sqlalchemy = "*"
 tortoise-orm = "*"
 tox = "*"
 uvicorn = "*"
+cryptography = "*"
```
peterschutt, line 65: Should we treat this dependency the same way we do `brotli`, i.e., optional and include in full extras?

Goldziher: We could, but what s the advantage of doing this?

peterschutt: Size of venv with only starlite is 92M, size of venv with starlite and cryptography is 108MB.. if it is only needed for session middleware why should it be non-optional?

Goldziher: It is optional - these are dev dependencies

peterschutt: oh ha, right - it isn't named in the prod dependencies section though.

starlite/middleware/session.py
```diff
@@ -0,0 +1,216 @@
...
+
+            Args:
+                message: An ASGI message.
+
+            Returns:
+                None
+            """
+            if message["type"] == "http.response.start":
+                headers = MutableHeaders(scope=message)
+                if should_vacate_session:
```
infohash, line 171: @Goldziher We cannot set empty cookie at client side if the session was corrupted or expired. This is because if the session was expired or corrupted, we should pass an empty session to the user so that the user code can fill it with its own authentication parameters that will be set as a cookie in the response. If the user doesn't make any changes to the session, then in the response, the empty session will be set as a cookie.

I am making changes to it and will show you soon why the design has to be like this.

Goldziher: ok, I trust you

starlite/middleware/session.py
```diff
@@ -0,0 +1,250 @@
...
+    """Maximal age of the cookie before its invalidated."""
+    scopes: conlist(Literal["http", "websocket"], min_items=1, max_items=2) = ["http", "websocket"]  # type: ignore[valid-type]
+    """Scopes for the middleware - options are 'http' and 'websocket' with the default being both"""
+    path: str = "/"
+    """Path fragment that must exist in the request url for the cookie to be valid. Defaults to '/'."""
+    domain: Optional[str] = None
+    """Domain for which the cookie is valid."""
+    secure: Optional[bool] = None
+    """Https is required for the cookie."""
+    httponly: Optional[bool] = None
```
Goldziher, line 72: Should not be optional and default to `True`. Session cookies should not be accessible from JS by default.

starlite/middleware/session.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Args:
+            associated_data: A dictionary contains the timestamp of the AEAD Tag and max-age of the session.
+
+        Returns:
+            bool: True if the session is NOT expired else False.
+        """
+        timestamp = associated_data["timestamp"]
+        max_age = associated_data["max_age"]
+        if timestamp + max_age <= int(time.time()):
```
Goldziher, line 146: You could simply store the expiration value before hand, and check it. 

Also, no need for an if / else clause here. 

`return timestamp + max_age <= int(time.time())` is a boolean

infohash: Good point. This private method is unnecessary boilerplate.

starlite/middleware/session.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Returns:
+            None
+        """
+        if scope["type"] in self.config.scopes:
+            scope.setdefault("session", {})
+            connection = HTTPConnection(scope)
+            cookie_keys = sorted(key for key in connection.cookies if self.config.key in key)
+            if cookie_keys:
+                with contextlib.suppress(KeyError):
```
Goldziher, line 244: why do you supress the exceptions here and below? This is where I would expect to see try / except.

infohash: The session should remain empty if the exception occurs. When the user code will see an empty session, it must start authentication again to set new session.

starlite/middleware/session.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Notes:
+            - The returned list is composed of a chunks of a single base64 encoded
+                string that is encrypted using AES-CGM.
+
+        Returns:
+            List of encoded bytes string of a maximum length equal to the 'CHUNK_SIZE' constant.
+        """
+        serialized = dumps(data, default=Response.serializer, option=OPT_SERIALIZE_NUMPY)
+        aad = dumps({"time_to_live": round(time.time()) + self.config.max_age})
```
Goldziher, line 129: `time_to_live` --> `ttl`, this is an accepted abbreviation. 

```suggestion
        aad = dumps({"ttl": round(time.time()) + self.config.max_age})
```

This though is not the correct term - because the `ttl` is actually the `max_age`. The term should be `expiration` or `expiry`.

starlite/middleware/session.py
```diff
@@ -0,0 +1,233 @@
...
+            data: A list of strings derived from the request's session cookie(s).
+
+        Returns:
+            A deserialized session value.
+        """
+        decoded = b64decode(b"".join(data))
+        nonce = decoded[:NONCE_SIZE]
+
+        aad = None
+        aad_footer = decoded.find(AAD)
```
Goldziher, line 148: footer is not clear to me. Footer in what sense?

starlite/middleware/session.py
```diff
@@ -0,0 +1,233 @@
...
+        Args:
+            data: A list of strings derived from the request's session cookie(s).
+
+        Returns:
+            A deserialized session value.
+        """
+        decoded = b64decode(b"".join(data))
+        nonce = decoded[:NONCE_SIZE]
+
+        aad = None
```
Goldziher, line 147: I would rename this variable to `aad_string`

starlite/middleware/session.py
infohash: AAD is appended to the ciphertext which is the footer of the ciphertext.

starlite/middleware/session.py
```diff
@@ -0,0 +1,233 @@
...
+        aad_footer = decoded.find(AAD)
+        if aad_footer != -1:
+            aad = decoded[aad_footer:].replace(AAD, b"")
+
+        encrypted_session = decoded[NONCE_SIZE:aad_footer]
+        decrypted = self.aesgcm.decrypt(nonce, encrypted_session, associated_data=aad)
+
+        session_validation = loads(aad)["time_to_live"] > round(time.time())
```
Goldziher, line 155: better to split this into a variable assignment:

```python
aad_data = loads(aad)
return loads(decrypted if aad_data and aad_data["expiration"] >= round(time.time())
```

starlite/middleware/session.py
infohash: Better to expand it in its full form. `authenticated_associated_data` or just `associated_data` to keep it self descriptive and for snake case naming.

starlite/middleware/session.py
Goldziher: ok, i would eliminate this variable - instead use decoded.find(...) both in the if and in the assignment.

starlite/middleware/session.py
infohash: How about `expires_at`?

Goldziher: also works.

starlite/middleware/session.py
infohash:
```python 
if decoded.find(AAD) != -1:
aad = decoded[decoded.find(AAD) : ].replace(AAD, b"")
```

2 function calls to find a substring byte with O(n) complexity. This won't be optimal. I will change the variable name instead.

starlite/middleware/session.py
infohash: No changes for this. It's not possible for associated data to be None. The previous line will raise the Invalid Tag exception if AAD has been removed/tampered by the attacker.

```python
session_validation = loads(associated_data)["expires_at"] > round(time.time())
return loads(decrypted) if session_validation else {}
```

starlite/middleware/session.py
```diff
@@ -0,0 +1,233 @@
...
+    """Maximal age of the cookie before its invalidated."""
+    scopes: conlist(Literal["http", "websocket"], min_items=1, max_items=2) = ["http", "websocket"]  # type: ignore[valid-type]
+    """Scopes for the middleware - options are 'http' and 'websocket' with the default being both"""
+    path: str = "/"
+    """Path fragment that must exist in the request url for the cookie to be valid. Defaults to '/'."""
+    domain: Optional[str] = None
+    """Domain for which the cookie is valid."""
+    secure: Optional[bool] = None
+    """Https is required for the cookie."""
+    httponly: Optional[bool] = True
```
peterschutt, line 73: I don't think this needs to be `Optional` as explicitly setting this value to `None` has the same outcome as setting it to `False` right? Just keeps the api that little bit simpler IMO.

pyproject.toml
```diff
@@ -63,6 +63,7 @@ sqlalchemy = "*"
...
```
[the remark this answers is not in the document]
peterschutt: Would probably need to re-lock afterwards.

tests/middleware/test_session_middleware.py
```diff
@@ -0,0 +1,142 @@
...
+)
+def test_config_validation(secret: bytes, should_raise: bool) -> None:
+    if should_raise:
+        with pytest.raises(ValidationError):
+            SessionCookieConfig(secret=SecretBytes(secret))
+    else:
+        SessionCookieConfig(secret=SecretBytes(secret))
+
+
+@pytest.fixture(autouse=True)
```
peterschutt, line 48: Does this need to be `autouse`?

peterschutt (commented):
Hey @infohash - this is really nice work!

I've made a couple of comments here - no explicit approval yet as I've run out of time for now and would like to keep going through it in a bit more detail.

For docs:
- It would be great if you could create an example in `examples/` that we'd feature in the usage docs for the implementation. We pair that with a simple test to ensure that our doc examples don't go stale.
- Usage docs
- include the public api in the reference docs

For tests, I'm wondering if we should have some explicit tests around the tampering behavior? Thoughts? 

Also, you could get your test coverage up to 100% by adding a test that shows that the else branch in `SessionMiddleware.__call__()` gets called if the scope isn't in `self.config.scopes`.

I'll spend a bit more time on this after work today, good stuff!

starlite/params.py
```diff
@@ -202,4 +202,4 @@ def Dependency(*, default: Any = Undefined, skip_validation: bool = False) -> An
         skip_validation (bool, optional): If `True` provided dependency values are not validated by signature model.
     """
     extra: Dict[str, Any] = {EXTRA_KEY_IS_DEPENDENCY: True, EXTRA_KEY_SKIP_VALIDATION: skip_validation}
-    return Field(default, **extra)  # type: ignore[pydantic-field]
+    return Field(default, **extra)
```
Goldziher: this change is probably an error - run these commands:

```shell
pre-commit clean
pre-commit gc
pre-commit install
pre-commit run --all-files
```

infohash:
@peterschutt I'll start writing examples and update the documentation for it. The requested review is just for the code changes.

starlite/middleware/session.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    max_age: conint(ge=1) = ONE_DAY_IN_SECONDS * 14  # type: ignore[valid-type]
+    """Maximal age of the cookie before its invalidated."""
+    scopes: conlist(Literal["http", "websocket"], min_items=1, max_items=2) = ["http", "websocket"]  # type: ignore[valid-type]
+    """Scopes for the middleware - options are 'http' and 'websocket' with the default being both"""
+    path: str = "/"
+    """Path fragment that must exist in the request url for the cookie to be valid. Defaults to '/'."""
+    domain: Optional[str] = None
+    """Domain for which the cookie is valid."""
+    secure: Optional[bool] = None
```
Goldziher, line 71:
```suggestion
    secure: bool = False
```

Goldziher:
Ok, this PR is ready to ship basically. We need documentation but thats it.

infohash:
Okay, I will raise a separate PR for example and doc.

infohash (approved):
Example and documentation update will be in next PR.

Goldziher:
@all-contributors add @infohash code

PR #426, opened 2022-08-29
Enhancement: Authentication Middleware exclude URLs and error handling improvements

### LINKED ISSUE #424: Abstract Authentication Middleware should respect `error_handlers` defined at the app/router level?
Should the `AbstractAuthenticationMiddleware` use the `exception_handlers`?

Given the example below, the `exeption_formatter` is not called for `NotAuthorized` exception, and you have to override the `error_response` method to return the custom exception format.

```python
from typing import Union

from starlette.requests import HTTPConnection

from starlite import (
    AbstractAuthenticationMiddleware,
    AuthenticationResult,
    Controller,
    HTTPException,
    MediaType,
    NotAuthorizedException,
    Request,
    Response,
    Router,
    Starlite,
    get,
)
from starlite.exceptions.exceptions import PermissionDeniedException


class ExceptionResponse(BaseModel):
    status_code: int
    message: str


def exception_formatter(_: Request, exception: Exception) -> Response:
    return Response(
        content=ExceptionResponse(
            status_code=getattr(exception, "status_code", 500), message=getattr(exception, "detail", "")
        ),
        status_code=getattr(exception, "status_code", 500),
        media_type=MediaType.JSON,
    )


class AuthenticationMiddleware(AbstractAuthenticationMiddleware):
    async def authenticate_request(self, request: HTTPConnection) -> AuthenticationResult:

        if not (token := request.headers.get("Authorization")):
            raise NotAuthorizedException()

        return AuthenticationResult(user=None, auth=token)


    path = "/needs-auth"

    @get()
    async def needs_auth(self) -> str:
        return "needs auth token"


router = Router(
    path="/api",
    route_handlers=[MyController],
    middleware=[AuthenticationMiddleware],
)

app = Starlite(
    route_handlers=[router],
    exception_handlers={HTTPException: exception_formatter},
)
```

cofin:
`AbstractAuthenticationMiddleware` now propagates exceptions and includes an `exclude` attribute to allow URLs to bypass authentication.

closes #424


starlite/middleware/authentication.py
```diff
@@ -43,47 +42,34 @@ class AbstractAuthenticationMiddleware(ABC, MiddlewareProtocol):
     The status code to for websocket authentication errors.
```
peterschutt, line 42: Should we try and support these still or just remove them and make a breaking change?

```python
    error_response_media_type = MediaType.JSON
    """
    The 'Content-Type' to use for error responses.
    """
    websocket_error_status_code = 4000
    """
    The status code to for websocket authentication errors.
    """
```

peterschutt (approved):
Looks good.

I suggest including an example of `exclude` in docs and reviewing the included members in reference docs for `AbstractAuthenticationMiddleware` - we probably need to include `__init__` and remove `error_response_media_type` and `websocket_error_status_code` if we remove them from the middlware protocol.

starlite/middleware/authentication.py
Goldziher: So for websocket connections, also @cofin, we have to handle it here - websockets do not have a response instance and are not treated by error handlers. 

The `error_response_media_type` is ok to drop in my opinion.

starlite/middleware/authentication.py
```diff
@@ -43,47 +42,34 @@ class AbstractAuthenticationMiddleware(ABC, MiddlewareProtocol):
...
+            await self.app(scope, receive, send)
+        else:
             if scope["type"] in self.scopes:
                 auth_result = await self.authenticate_request(HTTPConnection(scope))
                 scope["user"] = auth_result.user
                 scope["auth"] = auth_result.auth
             await self.app(scope, receive, send)
-        except (NotAuthorizedException, PermissionDeniedException) as e:
-            if scope["type"] == ScopeType.WEBSOCKET:  # pragma: no cover
-                await send({"type": "websocket.close", "code": self.websocket_error_status_code, "reason": repr(e)})
```
Goldziher, line 66: this code path needs to be kept

peterschutt: Isn't the `else` clause of the exception middleware pretty much this?

```python
            if scope["type"] == ScopeType.HTTP:
                ...
            else:
                status_code = (
                    exc.status_code if isinstance(exc, StarletteHTTPException) else HTTP_500_INTERNAL_SERVER_ERROR
                )
                # The 4000+ code range for websockets is customizable, hence we simply add it to the http status code
                status_code += 4000
                reason = repr(exc)
                await send({"type": "websocket.close", "code": status_code, "reason": reason})
```

Goldziher: valid point. In this case - my points are mut. I will just remove the class vars, make a breaking change, and update the docs before finishing this.

Goldziher (approved):

starlite/middleware/authentication.py
Goldziher: Lets drop these.

docs/usage/8-authentication.md
```diff
@@ -145,11 +145,13 @@ Finally, we need to pass our middleware to the Starlite constructor:
 
 ```python title="my_app/main.py"
 from starlite import Starlite
+from starlite.middleware.base import DefineMiddleware
 
 from my_app.security.authentication_middleware import JWTAuthenticationMiddleware
 
+auth_mw = DefineMiddleware(JWTAuthenticationMiddleware, excluded="schema")
```
cofin, line 152: Is this the correct way to define middleware that take arguments?

Goldziher: Yup. 

Although there is a JWTAuth.middkeware that dies it for you

Issue #418, opened 2022-08-27
Enhancement: Allow disabling openapi documentation sites and schema download endpoints via config


Goldziher:
Currently the OpenAPIController always exposes `redoc`, `swagger-ui` and `stoplight elements` static sites, and a `yaml` + `json` schema download endpoints. This though might not be desirable. 

Tasks:

1. add a new method on the controller called `render_404_not_found` or something such like, which renders a simple 404 page.
2. add a configuration option to the `OpenAPIConfig` class that determines which endpoints are enabled. For example, a list or set of literal identifiers - which by default includes all values. 
3. add logic to return a 404 if a particular endpoint is not enabled in the config.

Note:

Access the app level config is possble using the request instance. See how this is currently implemented in the `OpenAPIController.root` method as an example.

ingjavierpinilla:
Writing a method that returns an html string would make the tests green even if the endpoint is not authorized because you would still have a HTTP 200 code.

Any idea how this could be solved? 
I thought about using the class ExceptionResponseContent but as far as I understood, it doesn't allow you to render html.

Goldziher:
You can return the following:

```python
from starlite import Get, MediaType, Response
from starlette.status import HTTP_404_NOT_FOUND

@get(...)
def handler() -> Response:
   return Response(content=<html string>, status_code=HTTP_404_NOT_FOUND, media_type=MediaType.HTML)
```

ingjavierpinilla:
Thanks for your answer.

Not completely, the to_response function treats the response as flat data and tries to construct a new response using a response object as the content parameter.

Or am I missing something? 
Is there any way to avoid this behavior and return the response directly?

Maybe it is an issue we want to address before this (418) one?

Goldziher:
are you sure you are on an updated version of `starlite`? Look here `tests/handlers/http/test_to_response.py` - this is working as expected.

RSPersonal:
Hi can i pick this up?

ingjavierpinilla:
Sure

ingjavierpinilla:
> Hi can i pick this up?

Sure @RSPersonal

RSPersonal:
So bear with me, its my first issue on a open source project. I also seem to run into the same issue as @ingjavierpinilla. 

I wrote this method in the controller. 

` @staticmethod
    def render_404_not_found(self):
        html_response = f"""
            <!DOCTYPE html>
            <html lang="en">
            <head>
                <title>404</title>
                {self.favicon}
[...]

PR #430, opened 2022-08-29
Issue #418: Allow disabling openapi documentation sites and schema download endpoints via config

ingjavierpinilla:
It solves [418](https://github.com/starlite-api/starlite/issues/418) and the subsequent suggestions made during the discussion

TLDR:
mainly but not limited to

- Add a list of enabled endpoints for openapi documentation.
- Change openapi controller return type from string to return objects.

# PR Checklist


cofin (changes requested):
Thanks for the change here.  It's something I was thinking we needed as well.  Can you please update the documentation for configuration instructions before we merge this?

starlite/openapi/controller.py
```diff
@@ -65,6 +67,57 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
             raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
         return request.app.openapi_schema
 
+    @staticmethod
+    def check_exposed_endpoints(request: Request) -> Optional[Response]:
```
peterschutt, line 71: Sorry if it got a bit confusing in the issue discussion before, but we came to the conclusion that this needed to be a method that was called from within the route handlers to return the response, rather than via the `before_request` hook, so that the method could be easily overridden if desired.

This was also the reason that I suggested we change the response type of the handler methods so that they return a `Response` object in all cases, otherwise it would have been fine to leave those as `str` if we were going to use the `before_request` approach.

Pseudo code, but something along these lines:

```python
    @get(path="/elements", include_in_schema=False)
    def stoplight_elements(self, request: Request) -> Response:
        if should_serve_elements():
            # return the 200 response
        else:
            # return the 404 response
```
It'll work around those validation errors in CI too

ingjavierpinilla: Alright, thanks for the clarification.
I'll change it o7

starlite/config/openapi.py
```diff
@@ -107,6 +107,10 @@ class OpenAPIConfig(BaseModel):
     """
         The static schema generator to use for the "root" path of `/schema/`.
     """
+    allowed_endpoints: Set[str] = {"redoc", "swagger", "elements", "openapi.json", "openapi.yaml"}
```
Goldziher, line 110: I suggest to rename this to `enabled_endpoints` - it's a bit clearer

starlite/openapi/controller.py
```diff
@@ -65,6 +66,31 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
             raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
         return request.app.openapi_schema
 
+    @staticmethod
+    def should_serve_endpoint(request: Request) -> bool:
+        """This method verifies that the requested path is within the allowed
```
Goldziher, line 71: You are missing documentation for the args.

starlite/openapi/controller.py
```diff
@@ -65,6 +66,31 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
+            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If the application
+                `openapi_config` attribute is `None`.
+        """
+        if not request.app.openapi_config:  # pragma: no cover
+            raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
+
+        path = set(filter(None, request.url.path.split("/")))
+
+        config = request.app.openapi_config
+        if path == {"schema"} and config.root_schema_site in config.allowed_endpoints:
```
Goldziher, line 87: This is not going to work. A user can subclass the controller and change the root path, e.g. from `/schema` to `/api-docs`. Additionally the user can override the methods to change their paths.

ingjavierpinilla: You're right, I will rethink how to validate when accessing the root, in case root_schema_site is not in enabled_endpoints (which should not happen, but I still think it is important to check it).

starlite/openapi/controller.py
```diff
@@ -65,6 +66,35 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
             raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
         return request.app.openapi_schema
 
+    def should_serve_endpoint(self, request: Request) -> bool:
+        """This method verifies that the requested path is within the allowed
+        endpoints in the openapi_config.
+
+        Args:
+            request: A [Starlite][starlite.connection.Request] instance.
```
peterschutt, line 74:
```suggestion
            request: To be tested if endpoint activated.
```
Referencing the type of parameters in the description is unnecessary as the ref docs framework already renders that information with the arg description.

image

starlite/openapi/controller.py
```diff
@@ -65,6 +66,35 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
+
+        Returns:
+            A boolean.
+
+        Raises:
+            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If the application
+                `openapi_config` attribute is `None`.
```
peterschutt, line 81:
```suggestion
            ImproperlyConfiguredException: If the application `openapi_config` attribute is `None`.
```

No need to cross-ref the type, it will get resolved as a cross ref without it, e.g:
Before:

image

after

image

starlite/openapi/controller.py
```diff
@@ -95,12 +129,25 @@ def retrieve_schema_yaml(self, request: Request) -> "OpenAPI":
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered YAML object..
+            A [Response][starlite.response.Response] instance with a rendered YAML object
```
peterschutt, line 132: Again, redundant to reference the type in the description:

image

peterschutt (changes requested):
Quick review from a documentation perspective that highlights the workings of reference docs framework. I know some of these points are not consistent with existing docs on the controller, but that doesn't mean we shouldn't be moving forward with good practices! Good stuff, cheers:)

starlite/openapi/controller.py
```diff
@@ -58,13 +59,40 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
             An [OpenAPI][pydantic_openapi_schema.v3_1_0.open_api.OpenAPI] instance.
 
         Raises:
-            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If the application
-                `openapi_schema` attribute is `None`.
+            ImproperlyConfiguredException: If the application `openapi_config` attribute is `None`.
```
Goldziher, line 62: Why this change? The reference docs wouldnt be generated like this. Unless there is a good reason for this, please change back.

starlite/openapi/controller.py
```diff
@@ -58,13 +59,40 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
         if not request.app.openapi_schema:  # pragma: no cover
             raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
         return request.app.openapi_schema
 
+    def should_serve_endpoint(self, request: Request) -> bool:
+        """This method verifies that the requested path is within the enabled
+        endpoints in the openapi_config.
+
+        Args:
+            request: To be tested if endpoint enabled.
```
Goldziher, line 73: So, since we are generating the reference documentation from the doc strings, we need to follow the `mkdocstrings` notation here. This means that this needs to be updated to include the linkage, e.g.

```suggestion
            request: The [Request][starlite.connection.Request] instance.
```

starlite/openapi/controller.py
```diff
@@ -58,13 +59,40 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
+
+        Returns:
+            A boolean.
+
+        Raises:
+            ImproperlyConfiguredException: If the application `openapi_config` attribute is `None`.
```
Goldziher, line 79: Same as previously, for the linkage:

```suggestion
            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If the application `openapi_config` attribute is `None`.
```

starlite/openapi/controller.py
```diff
@@ -58,13 +59,40 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
+        """
+        if not request.app.openapi_config:  # pragma: no cover
+            raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
```
Goldziher, line 82:
```suggestion
            raise ImproperlyConfiguredException("Starlite has not been instantiated with an OpenAPIConfig")
```

starlite/openapi/controller.py
```diff
@@ -58,13 +59,40 @@ def get_schema_from_request(request: Request) -> "OpenAPI":
...
+
+        request_path = set(filter(None, request.url.path.split("/")))
+        root_path = set(filter(None, self.path.split("/")))
+
+        config = request.app.openapi_config
+        if request_path == root_path and config.root_schema_site in config.enabled_endpoints:
+            return True
+
+        if request_path & config.enabled_endpoints:
```
Goldziher, line 91: I suggest to `or` these two conditions instead of having two if clauses. Alternatively, save them into variables with clear names and then check `if <var1> or <var2>:`

starlite/openapi/controller.py
```diff
@@ -95,12 +127,25 @@ def retrieve_schema_yaml(self, request: Request) -> "OpenAPI":
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered YAML object..
+            resoponse: With a rendered YAML object
```
Goldziher, line 130: `resoponse`.

Also, here too we need linkage, thus:

```suggestion
            A [Response][starlite.response.Response] instance with the YAML object rendered into a string.
```

starlite/openapi/controller.py
```diff
@@ -109,12 +154,25 @@ def retrieve_schema_json(self, request: Request) -> "OpenAPI":
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered JSON object..
+            resoponse: With a rendered YAML object
```
Goldziher, line 157: Multiple issues here - both a typo and `YAML` instead of `JSON`.

```suggestion
            A [Response][starlite.response.Response] instance with the JSON object rendered into a string.
```

starlite/openapi/controller.py
```diff
@@ -95,12 +127,25 @@ def retrieve_schema_yaml(self, request: Request) -> "OpenAPI":
...
+        if not request.app.openapi_config:  # pragma: no cover
+            raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
+
+        if self.should_serve_endpoint(request):
+            return Response(
+                content=self.get_schema_from_request(request),
+                status_code=HTTP_200_OK,
+                media_type=OpenAPIMediaType.OPENAPI_YAML,
+            )
+        return Response(
```
Goldziher, line 141: ok, here the response should not be of type HTML. It should be of type `OpenAPIMediaType.OPENAPI_YAML` with content being an empty dictionary.

starlite/openapi/controller.py
```diff
@@ -109,12 +154,25 @@ def retrieve_schema_json(self, request: Request) -> "OpenAPI":
...
+        if not request.app.openapi_config:  # pragma: no cover
+            raise ImproperlyConfiguredException("Starlite has not been instantiated with OpenAPIConfig")
+
+        if self.should_serve_endpoint(request):
+            return Response(
+                content=self.get_schema_from_request(request),
+                status_code=HTTP_200_OK,
+                media_type=OpenAPIMediaType.OPENAPI_JSON,
+            )
+        return Response(
```
Goldziher, line 168: same comment basically as above regarding this response not being of the correct type. This should be a `OpenAPIMediaType.OPENAPI_JSON` response with an empty dictionary.

starlite/openapi/controller.py
```diff
@@ -125,56 +183,83 @@ def root(self, request: Request) -> str:
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered html string.
+            resoponse: With the rendered site defined in root_schema_site.
```
Goldziher, line 186: `resoponse` 😉 , you should be more careful with copy/paste.

Please link the response references here as above.

starlite/openapi/controller.py
```diff
@@ -125,56 +183,83 @@ def root(self, request: Request) -> str:
...
 
         Raises:
-            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If the application
-                `openapi_config` attribute is `None`.
+            ImproperlyConfiguredException: If the application `openapi_config` attribute is `None`.
```
Goldziher, line 189: missing linkage, please revert this change.

starlite/openapi/controller.py
```diff
@@ -125,56 +183,83 @@ def root(self, request: Request) -> str:
...
+    def swagger_ui(self, request: Request) -> Response:
         """Route handler responsible for rendering Swagger-UI.
 
         Args:
             request:
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered html string.
+            response: With a rendered swagger documentation site
```
Goldziher, line 213: No typo! but missing linkage, please update.

starlite/openapi/controller.py
```diff
@@ -125,56 +183,83 @@ def root(self, request: Request) -> str:
...
+    def stoplight_elements(self, request: Request) -> Response:
         """Route handler responsible for rendering StopLight Elements.
 
         Args:
             request:
                 A [Request][starlite.connection.Request] instance.
 
         Returns:
-            A rendered html string.
+            response: With a rendered stoplight elements documentation site
```
Goldziher, line 235: same as prev.

Goldziher (changes requested):
Thanks for the PR!

Some issues - primarily with the documentation. Please address these. You can then run

```shell
docker compose up
```

To launch the doc site locally and check all is well. Once the doc issues are addressed, I think this can be merged.

starlite/openapi/controller.py
peterschutt: @Goldziher I think you should check out my previous review

Goldziher: ok, as the reference tzar decides 😉

starlite/openapi/controller.py
peterschutt: These too, I suggested we don't need to cross-ref the type of the arg in the description as the docs already work it out:

image

starlite/openapi/controller.py
ingjavierpinilla: I'm not a big fan of this idea, I prefer separated and simple if statements. It's easier to read and understand imho

Goldziher: as you see fit

Goldziher:
@all-contributors add @ingjavierpinilla code

PR #431, opened 2022-08-29
Issue 371 add path type query parameter

### LINKED ISSUE #371: Enhancement: Add 'path' type to path parameters
Starlette's route path parameters include a `path` type that matches any further path components. Example: `Route('/uploaded/{rest_of_path:path}', uploaded)` would match `/uploaded/some-dir/some-file.txt` with `rest_of_path` set to `some-dir/some-file.txt`.

A couple cases where this might be useful:

- A file upload service where uploaded files are exposed as a virtual directory tree
- A blog where posts have extra path components for readability and SEO which are otherwise ignored, like `/posts/1234/2022/8/14/how-to-use-path-parameters/` with only `/posts/1234` being necessary

It doesn't look like this is supported yet in Starlite. Is there a recommended approach in the meantime?

Goldziher:
# PR Checklist

- [wip] Have you updated the prose documentation?
- [wip] Have you updated the reference documentation?

This PR resolves #371. It updates the supported types of path parameters, to include the following:

```python
param_type_map = {
    "str": str,
    "int": int,
    "float": float,
    "uuid": UUID,
    "decimal": Decimal,
    "date": date,
    "datetime": datetime,
    "time": time,
    "timedelta": timedelta,
    "path": Path,
}
```

All of them except `Path` have been previously supported, but not in a verbose way, so its just a bit of UX improvement there.

The type `Path` though was more complex. What it does is regard the path from the point at which the parameter is declared to be a value. Example:

```python
def test_path_param_type_resolution() -> None:
    @get("/some/test/path/{my_param:path}")
    def handler(my_param: Path) -> None: 
        assert my_param == "/1/2/3/4/some-file.txt"

        response = client.get("/some/test/path/1/2/3/4/file.txt")
```

Missing:

- [] updated docs

### DIFF
--- a/starlite/asgi.py
+++ b/starlite/asgi.py
@@ -1,6 +1,20 @@
 from datetime import date, datetime, time, timedelta
+from decimal import Decimal
 from inspect import getfullargspec, isawaitable, ismethod
-from typing import TYPE_CHECKING, Any, Dict, List, Set, Tuple, Type, Union, cast
+from pathlib import Path
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Set,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
+from uuid import UUID
 
 from pydantic.datetime_parse import (
     parse_date,
@@ -31,7 +45,7 @@ class PathParamNode:
 
 
 class PathParameterTypePathDesignator:
-    """Sentinel object to represent a path param in the route map."""
+    """Sentinel object to a path parameter of type 'path'."""
 
 
 PathParamPlaceholderType = Type[PathParamNode]
@@ -57,7 +71,15 @@ class StarliteASGIRouter(StarletteRouter):
         """Traverses the application route mapping and retrieves the correct
         node for the request url.
 
-        Raises NotFoundException if no correlating node is found
+        Args:
+            path: The request's path.
+            scope: The ASGI connection scope.
+
+        Raises:
+             NotFoundException: if no correlating node is found.
+
+        Returns:
+            A tuple containing the target RouteMapNode and a list containing all path parameter values.
         """
         path_params: List[str] = []
         current_node = self.app.route_map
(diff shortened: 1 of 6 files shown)


cofin (approved):

starlite/asgi.py
```diff
@@ -49,17 +62,20 @@ def _traverse_route_map(self, path: str, scope: "Scope") -> Tuple[RouteMapNode,
...
                 current_node = cast("RouteMapNode", current_node[component])
                 if "_static_path" in current_node:
                     self._handle_static_path(scope=scope, node=current_node)
                     break
                 continue
-            if PathParamPlaceholder in components_set:
+            if PathParamNode in components_set:
+                current_node = cast("RouteMapNode", current_node[PathParamNode])
+                if PathParameterTypePathDesignator in components_set:
+                    path_params.append(normalize_path(path.removeprefix("/".join(path.split("/")[:idx]))))
```
peterschutt, line 76: testing with 

```py
@get("/{path:path}")
def hello_world(path: Path) -> dict[str, Any]:
    """Hi."""
    return {"hello": str(path)}
```
Requesting `http://localhost:8000/a/b/c/d/e` returns `{"hello":"/a/b/c/d/e"}` but perhaps the leading `/` should not be included in the path, as it is outside the parameter.

I found a test from both flask and starlette that shows they both exclude the leading `/`:

```python
def test_static_folder_with_ending_slash():
    app = flask.Flask(__name__, static_folder="static/")

    @app.route("/<path:path>")
    def catch_all(path):
        return path

    rv = app.test_client().get("/catch/all")
    assert rv.data == b"catch/all"
```
and

```python
    # Test path conversion
    response = client.get("/path/some/example")
    assert response.json() == {"path": "some/example"}
    assert (
        app.url_path_for("path-convertor", param="some/example") == "/path/some/example"
    )
```

peterschutt (approved):
Looks good! just a query about whether we should include the leading `/` in the path.

starlite/asgi.py
Goldziher: ok, i modified it. please test and see if this works for you now.

peterschutt: Sorry dude, have shutdown and gone to bed so can't test rn but looks like it would've done the trick👍

PR #439, opened 2022-08-31
Fixes annotation of `**kwargs` across the library

peterschutt:
Changes `**kwargs: Dict[str, Any]` to `**kwargs: Any` in `MiddlewareProtocol.__init__()` signature.

---

# PR Checklist

### DIFF
--- a/starlite/middleware/base.py
+++ b/starlite/middleware/base.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Callable, Dict
+from typing import TYPE_CHECKING, Any, Callable
 
 from typing_extensions import Protocol, runtime_checkable
 
@@ -8,20 +8,24 @@ if TYPE_CHECKING:
 
 @runtime_checkable
 class MiddlewareProtocol(Protocol):  # pragma: no cover
-    def __init__(self, app: "ASGIApp", **kwargs: Dict[str, Any]):
-        """The MiddlewareProtocol is a PEP 544 Protocol that species the
-        requirements for an ASGI middleware.
+    def __init__(self, app: "ASGIApp", **kwargs: Any):
+        """Defines the requirements of a compliant ASGI middleware.
+
+        Middleware should assign the received `app` to `self` in order to reference it in
+        [`__call__()`][starlite.middleware.base.MiddlewareProtocol.__call__].
 
         Args:
-            app: An ASGIApp, this value is the next ASGI handler to call in the middleware stack.
+            app: The next ASGI handler to call in the middleware stack.
             **kwargs: Any key value pairs that are required for the middleware.
         """
 
     async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> None:
-        """The call method is the ASGIApp entry point. Once the middleware
-        finishes doing whatever it is meant to be doing,
+        """Executes the ASGI middleware.
 
-        it should call the next ASGI handler and await it - or await a response created in its closure.
+        Called by the previous middleware in the stack if a response is not awaited prior.
+
+        Upon completion, middleware should call the next ASGI handler and await it - or await a response created in its
+        closure.
 
         Args:
             scope: The ASGI connection scope.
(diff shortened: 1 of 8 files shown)


starlite/middleware/base.py
```diff
@@ -49,7 +53,7 @@ def __init__(self, middleware: Callable[..., "ASGIApp"], *args: Any, **kwargs: A
             *args: Positional arguments to pass to the callable.
             **kwargs: Key word arguments to pass to the callable.
 
-        Notes:
+        !!! Note
```
peterschutt, line 56: This change is to make the admonition default open - see: https://github.com/mkdocstrings/python/issues/22

starlite/middleware/base.py
```diff
@@ -8,20 +8,24 @@
 
 @runtime_checkable
 class MiddlewareProtocol(Protocol):  # pragma: no cover
-    def __init__(self, app: "ASGIApp", **kwargs: Dict[str, Any]):
-        """The MiddlewareProtocol is a PEP 544 Protocol that species the
-        requirements for an ASGI middleware.
+    def __init__(self, app: "ASGIApp", **kwargs: Any):
```
Goldziher, line 11: what does this fix?

peterschutt: `**kwargs` aren't a `dict`, e.g., I can't call that method with `MyMiddleware(app, {"k": "v"})`. The key value of `**kwargs` is always a `str` so that is inferred and only the value type needs to be specified.

Ref: https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

Goldziher: it is a dict - but destructured, you can call `MyMiddleware(app, **{"k": "v"})`. Furthermore, you can call all the dict methods on kwargs.

starlite/middleware/base.py
peterschutt: Do you mean when you type it you get a tab completion hint for `Notes:` but not one for `!!! Note`?

Goldziher: This is the way it is with standard google docstrings:

Screenshot 2022-08-31 at 13 54 44

And this is how this looks with the admonition

Screenshot 2022-08-31 at 13 54 59

peterschutt: I see, well I suppose it depends what you want to optimize for - neither is perfect.

Goldziher: and if instead we do this:

```markdown
        Notes:
            - The callable will be passed a kwarg `app`, which is the next ASGI app to call in the middleware stack.
            It therefore must define such a kwarg.
```

It looks like this:

Screenshot 2022-08-31 at 13 56 45

Which I think is the best one

starlite/middleware/base.py
peterschutt: It isn't redundant, it is correct.

```python
from typing import Any


def fn_correct(**kwargs: Any) -> None:
    reveal_type(kwargs)  # main.py:5: note: Revealed type is "builtins.dict[builtins.str, Any]"
 


def fn_incorrect(**kwargs: dict[str, Any]) -> None:
    reveal_type(kwargs)  # main.py:9: note: Revealed type is "builtins.dict[builtins.str, builtins.dict[builtins.str, Any]]"

```

Goldziher: ok, then can u replace all other occurances while at it?

starlite/middleware/base.py
peterschutt: This is from live docs:

image

This is my local:

image

peterschutt: We can just leave it and try and get it sorted upstream?

Goldziher: ok

Goldziher (approved):

starlite/middleware/base.py
peterschutt:
```suggestion
        Notes:
```

PR #447, opened 2022-09-01
Enhancement: use `picologging.dictConfig`

### LINKED ISSUE #414: Enhancement: Support dictConfig in picologging
`picologging` will implement a `dictConfig` with the following PR: https://github.com/microsoft/picologging/issues/53

We should enhance our integration to call this method once it's officially released.

cofin:
closes #414


starlite/logging/__init__.py
```diff
@@ -6,6 +6,11 @@
 
 from starlite.logging.standard import QueueListenerHandler
 
+try:
```
cofin, line 9: Is there a better way to do this?  I try to import it and set to `None` if it fails.  In the `configure`, we try to run `picologging.dictConfig` if `picologging` is found in the logging class name.

starlite/logging/__init__.py
```diff
@@ -49,5 +54,24 @@ class LoggingConfig(BaseModel):
...
+        import and set the dictConfig
+        """
+        for logging_class in _find_keys(self.handlers, "class"):
+            if "picologging" in logging_class and picologging_config:
+                picologging_config.dictConfig(self.dict(exclude_none=True))
+                break
         config.dictConfig(self.dict(exclude_none=True))
+
+
+def _find_keys(node: Union[List, Dict], key: str) -> Any:
```
Goldziher, line 69: Missing docstring. Also typing should be Generator[Iterable, None, None] if I understand the code correctly

starlite/logging/__init__.py
Goldziher: This is fine

Goldziher (approved):

PR #456, opened 2022-09-04
Fix: 'url_for' raises KeyError

infohash:


Goldziher:
ok, technically this is not a bug - the `url_for` is a Starlette method we never implemented. But I will take this oppurtinuity to fix it.

Goldziher (approved):

Issue #411, opened 2022-08-26
Enhancement: Logging Middleware


Goldziher:
We've decided to include an optional logging middleware. The middleware should log incoming requests and out-going responses. It needs to be configurable:

1. The user should be able to dictate the values that are being logged and the logging ourput structure. 
2. The middleware should support obfuscating data and excluding logging for specifc paths.
3. The middleware must use the logging config as a basis - i.e. register the logger as part of the config, and allow the user to decide on dictate what logger is actually used (e.g. picologger, stdlib, some structlogger etc.)

Swannbm:
I am starting something.

Swannbm:
Here some thinking about how to do it:

>The middleware should log incoming requests and out-going responses

I will make a first PR with this capabilities (so very simple). I'll add two debug log, one on incoming request and the other on outgoing response. I am thinking using INFO level.

I will name the logger classicaly using __name__.

> 1. The user should be able to dictate the values that are being logged

Are you speaking of choosing if he logs both (incoming / outgoing) or only one of them or even none of them.

Another idea could be to choose what is logged:
- url requested
- content-type expected
- browser used and other client information
- response status
...

> and the logging output structure.

I will add a format option that will override logging configuration for this logger.

> 2. The middleware should support obfuscating data and excluding logging for specifc paths.

I will add tow options:
- first, a callable list to parse the str that will be logged (that's for obfuscating).
- second a path exclusion list but I am unsure of the purpose of that. Shouldn't we log something minimal for that ?

> 3. The middleware must use the logging config as a basis - i.e. register the logger as part of the config, and allow the user to decide on dictate what logger is actually used (e.g. picologger, stdlib, some structlogger etc.)

I will add a logger_class in the configuration to load the correct one.

I am not an expert here, please feel free to complete your wish.

Issue #415, opened 2022-08-26
Enhancement: Rich and StructLog integration


cofin:
Create a documented example and/or integration to allow for `rich` traceback handling and logging handler.

Here's an example using the LoggingHandler

https://rich.readthedocs.io/en/latest/logging.html

john-ingles:
Like this?

`import logging
from starlite.logging import LoggingConfig

config = LoggingConfig(
    handlers={
        "console": {
            "class": "logging.StreamHandler",
            "level": "DEBUG",
            "formatter": "standard",
        },
        "rich": {
            "class": "rich.logging.RichHandler",
            "level": "DEBUG",
            "formatter": "standard",
        },
        "queue_listener": {
            "class": "starlite.QueueListenerHandler",
            "handlers": ["cfg://handlers.console"],
        },
    },
    loggers={"app": {"level": "INFO", "handlers": ["rich"]}},
)

config.configure()

logger = logging.getLogger("app")

logger.info("This is a test")`

john-ingles:
Maybe add
```"rich_tracebacks": True```
to the rich handler in handlers would be enough?

**Comment by john-ingles** [comment]
I got tracebacks to work by setting rich_tracebacks to True and the level to Error.

No the question is, should this be an integration or should we just use the code above in a documented example?
Any thoughts @Goldziher?

One reason why I think it won't work as an integration is that I don't think Rich works well with the queuehandler and queuelistener (at least I got thread errors when I tried adding the RichHandler to the queue_listener handlers list.

**Comment by Goldziher** [comment]
@cofin can you review this pls?

**Comment by cofin** [comment]
I think a simple documented example would be great.

FWIW, here is something I've used with FastAPI and recently adapter for Starlite.  It's been working for me in my limited testing, but i haven't done any testing with multiprocessing (or threads).  

Maybe we could add some notes on when this configuration might be a problem?

```python

class RichHandler(_RichHandler):
    """Rich Handler Config"""

    def __init__(self, *args, **kwargs) -> None:  # type: ignore
        super().__init__(
            rich_tracebacks=settings.app.LOG_LEVEL.lower() == "debug",
            console=Console(markup=True),
            markup=True,
            show_path=False,
            omit_repeated_times=False,
        )
```

And I used this for `LoggingConfig.handlers`:
```python
handlers={
        "console": {
            "class": "app.config.logging.RichHandler",
            "level": "DEBUG",
            "formatter": "standard",
        },
        "queue_listener": {
[...]

PR #462, opened 2022-09-06
Add logging middleware (411) draft

Swannbm:
# PR Checklist

Hello,

This is a draft PR for #411 because I am unsure of what is required to satisfy the issue. It's actuall my my first PR on public repo and I've got a lot of questions.

- I've chosen to configure logging at startup => is it a good idea ?
- Logging configuration is in starlite.logging not in starlite.config => should we move it ?
- I've chose to put Logging middleware in starlite.middlewares but I could have put it in starlite.logging => what would be the best ?
- My test, although it cover my code at 100% is very small (but the also is the middleware) => do you have any idea of edge case that I should add ?
- I log both incoming and outgoing event. Most web framework log only outgoing. What do you think should be the default ? Should I add specific configuration in LoggingConfig for activate incoming, outgoing, both ?
- I log those informations : '{hostname}:{port} - {method} {path} {scheme}/{http_version} {result}' (outgoing example: testserver:8000 - GET /my/path HTTPS/1.1 200), do you want more or less informations ?
- I've got the port from request.base_url.port, is it the good one (there also are ports in scope with server and client) ?

I need help on one failing test: test_request_body_logging_middleware in tests/middleware/test_middleware_handling.py. Initialize logging breaks capslog fixture. Anyone encouter this situation and knows how to make capslog working ?

Improvements:
- In order to ease making custom LoggingMiddleware, I am thinking adding a get_logging_middleware() to LoggingConfig. Is it a good idea ? Should I do that in that PR or make another one ?
- 

Thank you all for your help and recommandations.

Swann


starlite/middleware/logging.py
```diff
@@ -0,0 +1,79 @@
...
+from starlite.middleware.base import MiddlewareProtocol
+
+if TYPE_CHECKING:
+    from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+    from starlite.logging import LoggingConfig
+
+
+logger = logging.getLogger(__name__)
+
```
john-ingles, line 15: The only thing I think my proposed code from #415 would change is the need to add a line here.
```python
logger.handlers[0].start_listener()
```

cofin:
I'll try to take a look at this tonight if this is still an issue. 

We probably also need to add tests using the `picologging` logger as well.

Goldziher:
@Swannbm please convert this PR to a draft (it should be on the right side pane)l.

cofin:
Converted this PR to a draft

Swannbm:
Here the minutes of my (quick) discussion with Goldziher:
1. Copy SessionMiddleware to fetch the middleware from Config instance
2. Do not init logging up front, let the user do the job
3. Extend LoggingConfig with properties only for the middleware
4. keep current file organization to avoid breaking change (LoggingConfig will stay in logging.__init__.py)

All this points are opened to discussion.

Goldziher:
> Here the minutes of my (quick) discussion with Goldziher:
> 1. Copy SessionMiddleware to fetch the middleware from Config instance
> 2. Do not init logging up front, let the user do the job
> 3. Extend LoggingConfig with properties only for the middleware
> 4. keep current file organization to avoid breaking change (LoggingConfig will stay in logging.__init__.py)
> 
> All this points are opened to discussion.

 Rebase on development 😉, the `LoggingConfig` was moved.

Swannbm:
I have made the change (and necessary revert). I havn't check test coverage yet, I will do it later, my free time is comming to an end for today. @Goldziher and @cofin (even @john-ingles ) could you have a code review please ?

starlite/config/logging.py
```diff
@@ -61,6 +215,35 @@ def configure(self) -> None:
                 break
         config.dictConfig(self.dict(exclude_none=True))
```
cofin, line 216: Should this exclude the additional middleware classes in addition to None?

starlite/config/logging.py
```diff
@@ -9,6 +19,137 @@
...
+        Args:
+            app: The 'next' ASGI app to call.
+            config: The LoggingConfig instance.
+        """
+        super().__init__(app)
+        self.app = app
+        self.config = config
+        self.logger = self.get_logger()
+
+    def get_logger(self) -> Logger:
```
cofin, line 95: Why don't we use the same method in configure to return the logger for picologging if it's in the class config?

starlite/config/logging.py
cofin, line 216: I think we need to exclude the additional columns added that don't apply to the logger `dictConfig` (`middleware_*`)

docs/usage/7-middleware/3-builtin-middlewares/6-logging-middleware.md
```diff
@@ -0,0 +1,111 @@
...
+from starlite import Starlite, Request, get
+from starlite.config.logging import LoggingConfig, LoggingMiddleware
+
+
+@get("/")
+def my_handler(request: Request) -> None:
+    return json.dumps({"hello": "world"})
+
+
+class PicologgingMiddleware(LoggingMiddleware):
```
cofin, line 101: This won't be needed if we add the recommended change below.

cofin (changes requested):
Take a look at the short list of updates I recommended.  Let me know what you think.

starlite/config/logging.py
Swannbm: See documentation, I recommand to subclass for adding picologging support. We can provide a PicologgingMiddleware if you'd like. Moreover, we could move picologging configuration very strange piece of code in here.

starlite/config/logging.py
Swannbm: Should be better know.

docs/usage/7-middleware/3-builtin-middlewares/6-logging-middleware.md
Swannbm: I would prefere add PicologgingMiddleware and PicologgingConfig in starlite standard if you want to support it. I really dislike the way the configure method works... It is hard to understand and hard to maintain. Subclassing is more straight forward and obvious.

starlite/config/logging.py
cofin: My only hesitation is that it now requires additional middleware instead of a logging config change in order to get `picologging` working.  Given that there are a few outstanding bugs right now with full `picologging` integration, I'm inclined to ignore it for now.  

However, in general, I think that we should focus on making it simple to configure.  In my mind, that means the logging middleware should automatically use `picologging` if it's configured in the LoggingConfig, but I could be persuaded otherwise.

@Goldziher  @peterschutt  you guys have strong feelings either way?

docs/usage/7-middleware/3-builtin-middlewares/6-logging-middleware.md
cofin: I don't disagree that the way it works has a bit of a code smell to it.  How do you propose the logging config get changed to match this?

starlite/config/logging.py
Goldziher: I think the Middleware should be agnostic to the logging config. The question is how to achieve this. 

Optimally the Middleware would be able to call a function and receive the configured logger - this could be stdlib, picologger or anything else. 

Tell me, can we use `logging.getLogger` with picologging? If so I would just use this and decouple the Middleware from the logging config. They don't need to be tightly coupled.

cofin: Yeah, I think we can.  If i remember correctly in the tests, we show that get logger retuned from `logging.getLogger` is the picologger when it is configured.

Swannbm: Even if API are identical between both logging and picologging, we need to import the correct one and call the selected one when initializing it and when we try to fetch a logger in a standard part of starlite or in a third party.

I see tow possibilities:
1/ add a config option (something like logging_class) and check that option when calling logging.config.dictConfig()
2/ add a subclass of LoggingConfig called PicologgingConfig (which is almost the same as before)

I would like to avoid any configuration parsing to find a keyword. It looks like a needle in a haystack, it's hard to explain and seems to me very prone to unwanted behavior.

To untight Middleware from config, we  need to provide a straightforward way to get a logger (also usefull for third party that would like to add debuging info). It might be part of another issue, but exploring this could help us make a choice (maybe).

Here is a case:

```python
# /third/part/starlite-websocket/entrypoint.py

from startlite.logging import get_logger

logger = get_logger(__name__)

logger.debug("initilializing listening server")
...
```

Under the hood, we need to ensure that get_logger return a logger according to user choice... which seems very tricky atm... I think we need to make a lazy call or something...

I also might have completly missed yhour point @Goldziher :D

cofin: > I would like to avoid any configuration parsing to find a keyword. It looks like a needle in a haystack, it's hard to explain and seems to me very prone to unwanted behavior.

I completely agree here. 


> To untight Middleware from config, we need to provide a straightforward way to get a logger (also usefull for third party that would like to add debuging info). It might be part of another issue, but exploring this could help us make a choice (maybe).

I also think this is a good approach.

Goldziher:
any updates?

starlite/config/logging.py
Swannbm: I have made some tests and using a LoggingConfig parameter to choose which logging module user wants end up with complex and prone to bug code.

Here is a sample to illustrate (not tested yet):

```python
import sys

class LoggingConfig(BaseModel):
    """a wrapper for configuring any logging module that does not respect standard logging API."""

    logging_module: str = "logging"
        """Explicitly define the module name you want to use for logging."""

    def _get_logging_module(self) -> Callable:
            """Return instanciated module from logging_module configuration"""
            try:
                return sys.modules.get(
                    self.config.logging_module,
                    __import__(self.config.logging_module),  # if not already loaded
                )
            except ImportError as exc:
                raise ImportError(f"You need to manually install {self.module_name}") from exc

    def getLogger(self) -> Any:
        return self._get_logging_module().getLogger(self.config.middleware_logger_name)

    def configure(self) -> None:
        dict_config = {
            k: v for k, v in self.dict(exclude_none=True).items()
            if not k.startswith("middleware_") and k != "logging_module"
        }
        # calls config from logging_module
        config = getattr(self._get_logging_module(), "config")
        config.dictConfig(dict_config)
```

In another hand, making the same with inheritance is really more simple, using import statements top file (where it belongs) and explicitly saying which module to use seems to me way simpler:


```python
import logging
import logging.config
import picologging


class LoggingConfig(BaseModel):
    """a wrapper for configuring any logging module that does not respect standard logging API."""

    def getLogger(self) -> Any:
        return logging.getLogger(self.config.middleware_logger_name)

    def configure(self) -> None:
        dict_config = {
            k: v for k, v in self.dict(exclude_none=True).items()
            if not k.startswith("middleware_") and k != "logging_module"
        }
        # calls config from logging_module
        logging.config.dictConfig(dict_config)


class PiccologgingConfig(LoggingConfig):
    def getLogger(self) -> Any:
        return picologging.getLogger(self.config.middleware_logger_name)

    def configure(self) -> None:
        dict_config = {  # TODO move into dedicated method
            k: v for k, v in self.dict(exclude_none=True).items()
            if not k.startswith("middleware_") and k != "logging_module"
        }
        # calls config from logging_module
        picologging.config.dictConfig(dict_config)
```

Whatever our choice above, in the middleware (or whatever third party) we just have to call `self.config.getLogger()`.

In another issue, we should think on how to make this available through `from starlite import getLogger`.
We should also untight the "Convenience `pydantic` model for configuring logging" from app logging configuration.

I definitely vote for inheritance. 

@cofin do you agree ? Can I proceed ?

Goldziher: this is not in scope for a middleware PR. Please see the new issue i created. Youre welcome to add a PR there.

PR #468, opened 2022-09-07
Move Logging config to starlite.config

john-ingles:
Decouple logging config from logging module and move to starlite.config.logging

# PR Checklist

It makes more sense to me to move the LoggingConfig to the starlite.config folder.  I made a new folder there for logging and readjusted imports across the repo to maintain the public api (so no need to modify docs).  The difference now is that the user doesn't have to look in the logging module to see the default logging config model and could even add their own/modify the existing one if they wish.


starlite/config/__init__.py
```diff
@@ -1,7 +1,10 @@
+from starlite.logging import QueueListenerHandler
```
cofin, line 1: Should this also be exported here?  it looks like it's still in the original `starlite.logging.__init__`

starlite/config/__init__.py
```diff
@@ -14,4 +17,6 @@
     "StaticFilesConfig",
     "TemplateConfig",
     "CompressionConfig",
+    "LoggingConfig",
+    "QueueListenerHandler",
```
cofin, line 21: Same as above

cofin (changes requested):
Looks good other than the question about the `QueueListenerHandler`.  I'll merge once we've resolved.

starlite/config/__init__.py
john-ingles: I didn't move it to starlite.config because it's a function of the logging itself and unrelated to configuration

john-ingles: Wait, I think I see what you mean.  It doesn't need to be exported out of config.\_\_init\_\_.  I'll make a change

cofin:
Looks good to me.  Nice work!

cofin (approved):

cofin:
@all-contributors add @john-ingles code

Issue #460, opened 2022-09-04
Enhancement: Update Test Client


Goldziher:
We need to enhance the `TestClient` class with methods to easily set / clear authentication headers - i.e., a session cookie or a JWT token header with a specified key.

ingjavierpinilla:
Hi.
I'm a bit confused about this requirement. Do you want a method to set/remove the Middleware from the TestClient?

Goldziher:
> Hi. I'm a bit confused about this requirement. Do you want a method to set/remove the Middleware from the TestClient?

For JWT its simple and we dont need to do anything, basically you set a header. But the `SessionMiddleware` is something we have as part of Starlite. The thing is - in order to create a user session, you basically need to go through the middleware now. 

I was thinking of the way django allows you to "login" a user in the context of tests. Obviously we cannot do the same because we are not an ORM and user management system as well, but what we can do is create a session and send it as part of the request. This is very annoying to do right now because you basically need to serialize the cookie on your own. So I was thinking we might find a way to do this - although I am not sure how feasible it is TBH. @infohash - do you have input on this?

infohash:
A fixture will work.

```python
def mock_asgi_app(scope: "Scope", receive: "Receive", send: "Send") -> None:
    pass

@pytest.fixture()
def session_middleware() -> SessionMiddleware:
    return SessionMiddleware(app=mock_asgi_app, config=SessionCookieConfig(secret=TEST_SECRET))


def create_session(size: int = 16) -> Dict[str, str]:
    return {"key": secrets.token_hex(size)}


@pytest.fixture()
def create_test_cookies(session_middleware) -> str:
    cookies = ';'.join(f"session-{i}={serialize.decode('utf-8')}" for i, serialize in enumerate([b'abc', b'xyz']))
    return cookies

request= create_test_request(cookie=create_test_cookies)

>>> request.cookies
{'session-0': 'abc', 'session-1': 'xyz'}
```

infohash:
I have written the fixtures for it. Will raise the PR soon.

infohash:
@Goldziher Is it fine to keep random data in the fixture to create session cookie or do we need an option for tester to specify what session cookie should contain?

Goldziher:
[...]

PR #471, opened 2022-09-10
Added Fixture To Create Session Cookies

infohash:
Addresses #460

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,17 @@
 import os
 import pathlib
-from typing import AsyncGenerator, Callable
+import secrets
+from typing import TYPE_CHECKING, AsyncGenerator, Callable
 
 import pytest
 from piccolo.conf.apps import Finder
 from piccolo.table import create_db_tables, drop_db_tables
+from pydantic import SecretBytes
+
+from starlite.middleware.session import SessionCookieConfig, SessionMiddleware
+
+if TYPE_CHECKING:
+    from starlette.types import Receive, Scope, Send
 
 
 def pytest_generate_tests(metafunc: Callable) -> None:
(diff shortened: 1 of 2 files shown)


tests/middleware/test_session_middleware.py
```diff
@@ -20,25 +20,16 @@
...
 @pytest.mark.parametrize(
     "secret, should_raise",
     [
-        [urandom(16), False],
-        [urandom(24), False],
-        [urandom(32), False],
-        [urandom(17), True],
-        [urandom(4), True],
-        [urandom(100), True],
+        [os.urandom(16), False],
```
Goldziher, line 27: Whats the reasoning for using `os.urandom` rather than `urandom`?

tests/conftest.py
```diff
@@ -35,3 +42,26 @@ async def scaffold_piccolo() -> AsyncGenerator:
...
+
+#############################
+# Session Middleware Fixtures
+#############################
+async def mock_asgi_app(scope: "Scope", receive: "Receive", send: "Send") -> None:
+    pass
+
+
+@pytest.fixture()
+def session_middleware() -> SessionMiddleware:
```
Goldziher, line 55: My thinking was not only to expose these as internal test fixtures, but also expose to the user a way to easily create sessions in tests.

tests/middleware/test_session_middleware.py
infohash: Function imports from the standard library, keep their namespace implicit to their base module name.

tests/conftest.py
infohash: Something like whenever the session middleware is used in tests, it must implicitly starts with a default session?

Goldziher: dunno, if you have a solution go for it.

infohash: The best way is to have test `scope`, test `receive` and test `send` context managers inside `starlite.testing`. They will accept user arguments and default arguments to initialize them with user data. We will set a default value for `scope['session']` as well.

Goldziher: Sounds good @infohash

infohash: Merge this PR, I will raise a separate PR for this.

Goldziher (approved):

PR #475, opened 2022-09-11
Add log_config and rich_logging attribs to app

john-ingles:
Adds two new instance attributes to starlite.app.Starlite class.
1. log_config - Takes in a dict of config options to pass to logging.dictConfig. Allows custom log configuration by the user.
2. rich_logging - Default False.  If True, swaps out the default console log handler with a RichHandler.  If rich isn't present when rich_logging = True, raises an ImproperlyConfiguredException

# PR Checklist

I'll probably need some help writing tests and docs for this assuming there are no changes needed.

### DIFF
--- a/starlite/config/logging.py
+++ b/starlite/config/logging.py
@@ -1,8 +1,11 @@
-from logging import config
-from typing import Any, Dict, Generator, Iterable, List, Optional, Union
+from logging import config, getLogger
+from typing import TYPE_CHECKING, Any, Dict, Generator, Iterable, List, Optional, Union
 
 from pydantic import BaseModel
 from typing_extensions import Literal
+
+if TYPE_CHECKING:
+    from starlite.datastructures import State
 
 try:
     from picologging import config as picologging_config
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
     """This will be the configuration for the root logger. Processing of the configuration will be as for any logger,
     except that the propagate setting will not be applicable."""
 
-    def configure(self) -> None:
+    def _enable_rich_logging(self, log_config_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """Overwrite the default console log configuration with rich log
+        configuration.
+
+        Args:
+            log_config_dict (Dict[str,Any]): dict representation of LoggingConfig model
+        Returns:
+            Dict[str,Any]: dict representation of LoggingConfig model with rich console logging in place of default console logging
+        """
+        log_config_dict["formatters"]["rich"] = {"format": "%(name)s - %(message)s"}
+
+        log_config_dict["handlers"].pop("console", None)  # remove console handler if present
+
+        log_config_dict["handlers"]["p_rich"] = {
+            "class": "rich.logging.RichHandler",
+            "level": "DEBUG",
+            "rich_tracebacks": True,
+            "formatter": "rich",
+        }
+        if "cfg://handlers.console" in log_config_dict["handlers"]["queue_listener"]["handlers"]:
+            log_config_dict["handlers"]["queue_listener"]["handlers"].remove("cfg://handlers.console")
+
+        log_config_dict["handlers"]["queue_listener"]["handlers"].append("cfg://handlers.p_rich")
+
+        return log_config_dict
+
+    def configure(self, state: "State") -> None:
         """Configured logger with the given configuration.
 
         If the logger class contains the word `picologging`, we try to
         import and set the dictConfig
         """
+        log_config_dict = self.dict(exclude_none=True)
+
+        if getattr(state, "rich_logging", False):
+            log_config_dict = self._enable_rich_logging(log_config_dict)
+
         for logging_class in find_keys(self.handlers, "class"):
             if "picologging" in logging_class and picologging_config:
-                picologging_config.dictConfig(self.dict(exclude_none=True))
+                picologging_config.dictConfig(log_config_dict)
                 break
-        config.dictConfig(self.dict(exclude_none=True))
+        else:  # no break
+            config.dictConfig(log_config_dict)
+
+        # disable uvicorn errors from showing on the console
+        uvicorn_error = getLogger("uvicorn.error")
+        uvicorn_error.disabled = True
+        uvicorn_access = getLogger("uvicorn.access")
+        uvicorn_access.disabled = True
 
 
 def find_keys(node: Union[List, Dict], key: str) -> Generator[Iterable, None, None]:
--- a/starlite/config/logging.py
+++ b/starlite/config/logging.py
@@ -4,8 +4,6 @@ from typing import TYPE_CHECKING, Any, Dict, Generator, Iterable, List, Optional
 from pydantic import BaseModel
 from typing_extensions import Literal
 
-if TYPE_CHECKING:
-    from starlite.datastructures import State
 
 try:
     from picologging import config as picologging_config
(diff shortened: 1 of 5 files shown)


starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
     """This will be the configuration for the root logger. Processing of the configuration will be as for any logger,
     except that the propagate setting will not be applicable."""
 
-    def configure(self) -> None:
+    def _enable_rich_logging(self, log_config_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """Overwrite the default console log configuration with rich log
+        configuration.
+
+        Args:
+            log_config_dict (Dict[str,Any]): dict representation of LoggingConfig model
```
Goldziher, line 60: no need to type the parameters here, since there is typing in place in the funciton. 
```suggestion
            log_config_dict: dict representation of LoggingConfig model
```

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
+        Returns:
+            Dict[str,Any]: dict representation of LoggingConfig model with rich console logging in place of default console logging
```
Goldziher, line 62: also here, no need for typing

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
             if "picologging" in logging_class and picologging_config:
-                picologging_config.dictConfig(self.dict(exclude_none=True))
+                picologging_config.dictConfig(log_config_dict)
                 break
-        config.dictConfig(self.dict(exclude_none=True))
+        else:  # no break
+            config.dictConfig(log_config_dict)
+
+        # disable uvicorn errors from showing on the console
+        uvicorn_error = getLogger("uvicorn.error")
```
Goldziher, line 100: this is problematic - uvicorn might not be used. Users can use any ASGI server, such as `daphne` or `hypercorn`, or even something we are not aware of.

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
+        configuration.
+
+        Args:
+            log_config_dict (Dict[str,Any]): dict representation of LoggingConfig model
+        Returns:
+            Dict[str,Any]: dict representation of LoggingConfig model with rich console logging in place of default console logging
+        """
+        log_config_dict["formatters"]["rich"] = {"format": "%(name)s - %(message)s"}
+
+        log_config_dict["handlers"].pop("console", None)  # remove console handler if present
```
Goldziher, line 66:
```suggestion
        log_config_dict["handlers"].pop("console", None)  
```

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
+    def _enable_rich_logging(self, log_config_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """Overwrite the default console log configuration with rich log
+        configuration.
+
+        Args:
+            log_config_dict (Dict[str,Any]): dict representation of LoggingConfig model
+        Returns:
+            Dict[str,Any]: dict representation of LoggingConfig model with rich console logging in place of default console logging
+        """
+        log_config_dict["formatters"]["rich"] = {"format": "%(name)s - %(message)s"}
```
Goldziher, line 64: Probably best to think of how the user can configure this.

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
+            "formatter": "rich",
+        }
+        if "cfg://handlers.console" in log_config_dict["handlers"]["queue_listener"]["handlers"]:
+            log_config_dict["handlers"]["queue_listener"]["handlers"].remove("cfg://handlers.console")
+
+        log_config_dict["handlers"]["queue_listener"]["handlers"].append("cfg://handlers.p_rich")
+
+        return log_config_dict
+
+    def configure(self, state: "State") -> None:
```
Goldziher, line 81: this will work fine for `on_startup`, but users can also call configure indepdently. So this should either be optional or have a default value.

starlite/config/logging.py
```diff
@@ -49,17 +52,55 @@ class LoggingConfig(BaseModel):
...
         """Configured logger with the given configuration.
 
         If the logger class contains the word `picologging`, we try to
         import and set the dictConfig
         """
+        log_config_dict = self.dict(exclude_none=True)
+
+        if getattr(state, "rich_logging", False):
```
Goldziher, line 89: Perphaps its simply possible to enhance the pydantic model itself with some values - such as a boolean flag for enabling rich logging. Then there will not be a need for saving values in state.

What you would then do is exclude these properties when calling self.dict.

john-ingles:
@Goldziher  Thanks for the feedback.  I'll work on those.  In the mean time, any advice on getting through the pre-commit checks?  Looks like we might need to increase the number of allowed parameters to a function as well tell it not to worry about rich not being resolvable.  Is there a better way of checking to see if rich is installed than importing it?  Importlib?

Goldziher:
> @Goldziher Thanks for the feedback. I'll work on those. In the mean time, any advice on getting through the pre-commit checks? Looks like we might need to increase the number of allowed parameters to a function as well tell it not to worry about rich not being resolvable. Is there a better way of checking to see if rich is installed than importing it? Importlib?

For pylint - you can add `too-many-statements` to the disabled rules section for pylint in `pyproject.toml`:

```toml
[tool.pylint.MESSAGE_CONTROL]
disable = [
...
]
```

As for checking if something is installed - its standard practice to try to import something and catch `ImportError`. E.g.:

```python
try:
   from rich import x
except ImportError:
   x = None
```

You can then test `if x:` etc.

starlite/config/logging.py
john-ingles: I'll remove these lines.

I'm not sure how to suppress non-intended log messages from making their way to the console.  Perhaps it should be left up to the user?

starlite/config/logging.py
john-ingles: I agree that having the default format strings buried in the LoggingConfig model isn't great.  Would it work to put them in starlite.constants and import them?  Or is that too far removed from the logging config?

Goldziher: constants should be placed at the top of the file probably, but this is not what im talking about.

The user should be able to define the formatter string. Here you only have name and message, but there are other variables exposed by the logger. 

I would try to add this as a class attribute.

john-ingles: An attribute of LoggingConfig or Starlite?

starlite/config/logging.py
john-ingles: new commit addresses this

starlite/config/logging.py
john-ingles: new commit addresses this

starlite/config/logging.py
Goldziher: Of `LoggingConfig`

Goldziher:
whats the status of this?

john-ingles:
> whats the status of this?

I've been waiting to see what will happen with the logging middleware #411 .  It looked like it was going to be merged soon and might have implications for this that might require big changes.  In the mean time, do you have any advice for the failing tests?

Goldziher:
Let's proceed with this one. Regardless of the other PR.

I'll look at the tests.

Goldziher:
@john-ingles - you can rebase on main. It should fix the issue in CI

john-ingles:
I think I'll close this draft because of #511's submission.

Goldziher:
As you wish. I think you can add a simple PR for rich logging based on it. I am about to update it shortly.

Issue #316, opened 2022-07-29
Enhancement: Extended documentation for multiple responses


lewoudar:
Hi folks,
I just discovered this framework today, it seems really promising. But I already have one question after reading the docs.
It seems like it is not possible to document alternative responses for a route path (or may be I read too fast), i.e. something like [this](https://fastapi.tiangolo.com/advanced/additional-responses/) in FastAPI.

Goldziher:
You can just type the return value as a union type and it will infer the responses for you.

lewoudar:
Ok I see, but the issue here is that you can't customize the message, description, etc...

Goldziher:
> Ok I see, but the issue here is that you can't customize the message, description, etc...

Per response content body content? Not at present. 

To be clear - we are talking here bout customizing the description of elements inside whats called `oneOf` In OpenAPI, correct?

lewoudar:
> To be clear - we are talking here bout customizing the description of elements inside whats called oneOf In OpenAPI, correct?

Yes!

Goldziher:
> > To be clear - we are talking here bout customizing the description of elements inside whats called oneOf In OpenAPI, correct?
> 
> Yes!

Ok. I'll Label this as an enhancement request then. You're welcome to contribute a PR to handle this, or someone else will get to it 😀

lewoudar:
Unfortunetaly I don't have too much time now to dig in the code 😢
You probably want to affect it to another person

Goldziher:
No worries

Goldziher:
@seladb any progress?

seladb:
Unfortunately I didn't have time to get to it yet, but it seems like a good issue to work on 🙂
If someone else wants to take it I can un-assign myself, otherwise would it be ok to give me a few more weeks?

Goldziher:
> Unfortunately I didn't have time to get to it yet, but it seems like a good issue to work on 🙂
> If someone else wants to take it I can un-assign myself, otherwise would it be ok to give me a few more weeks?

Sure, take your time.

seladb:
@Goldziher just to make sure: do we want to support additional responses with different status codes or just customizing the description of existing return values?

**Option 1:** support additional responses can be quite similar to what FastAPI has, for example:
```python
@get(..., responses={404, {"model": Message}})
``` 

**Option 2:** allow customizing the description of existing return values - not yet sure how to achieve this because there is no description for each return type 🤔 
[...]

PR #478, opened 2022-09-14
Issue #316 - Extended documentation for multiple responses

seladb:
# PR Checklist

# Description

Corresponding issue: https://github.com/starlite-api/starlite/issues/316

This PR only handles additional responses with `model` which is similar to [this](https://fastapi.tiangolo.com/advanced/additional-responses/#additional-response-with-model) in FastAPI.

If we want to be more compatible with FastAPI we can also add [Additional media types for the main response](https://fastapi.tiangolo.com/advanced/additional-responses/#additional-media-types-for-the-main-response), but this can be done in a separate PR

### DIFF
--- a/starlite/datastructures.py
+++ b/starlite/datastructures.py
@@ -371,7 +371,7 @@ class FormMultiDict(ImmutableMultiDict[str, Any]):
 
 
 class AdditionalResponse(BaseModel):
-    """Container type of additional responses"""
+    """Container type of additional responses."""
 
     model: Type[BaseModel]
     """A model that describes the content of the response"""
--- a/starlite/datastructures.py
+++ b/starlite/datastructures.py
@@ -368,3 +368,14 @@ class FormMultiDict(ImmutableMultiDict[str, Any]):
         for _, value in self.multi_items():
             if isinstance(value, UploadFile):
                 await value.close()
+
+
+class AdditionalResponse(BaseModel):
+    """Container type of additional responses"""
+
+    model: Type[BaseModel]
+    """A model that describes the content of the response"""
+    generate_examples: bool = True
+    """Generate examples for the response content"""
+    description: str = "Additional response"
+    """A description of the response"""
--- a/starlite/datastructures.py
+++ b/starlite/datastructures.py
@@ -368,14 +368,3 @@ class FormMultiDict(ImmutableMultiDict[str, Any]):
         for _, value in self.multi_items():
             if isinstance(value, UploadFile):
                 await value.close()
-
-
-class AdditionalResponse(BaseModel):
-    """Container type of additional responses."""
-
-    model: Type[BaseModel]
-    """A model that describes the content of the response"""
-    generate_examples: bool = True
-    """Generate examples for the response content"""
-    description: str = "Additional response"
-    """A description of the response"""
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -188,10 +188,11 @@ def _create_data_handler(
     async def handler(data: Any, plugins: List["PluginProtocol"], **kwargs: Any) -> StarletteResponse:
         data = await _normalize_response_data(data=data, plugins=plugins)
         normalized_cookies = _normalize_cookies(cookies, [])
+        normalized_headers = _normalize_headers(headers)
         response = response_class(
             background=background,
             content=data,
-            headers=headers,
+            headers=normalized_headers,
             media_type=media_type,
             status_code=status_code,
         )
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -26,7 +26,6 @@ from starlette.status import (
 
 from starlite.constants import REDIRECT_STATUS_CODES
 from starlite.datastructures import (
-    AdditionalResponse,
     BackgroundTask,
     BackgroundTasks,
     File,
@@ -40,6 +39,7 @@ from starlite.exceptions import (
     ValidationException,
 )
 from starlite.handlers.base import BaseRouteHandler
+from starlite.openapi.datastructures import ResponseSpec
 from starlite.plugins import get_plugin_for_value
 from starlite.provide import Provide
 from starlite.response import Response
@@ -271,7 +271,7 @@ class HTTPRouteHandler(BaseRouteHandler["HTTPRouteHandler"]):
         operation_id: Optional[str] = None,
         raises: Optional[List[Type[HTTPException]]] = None,
         response_description: Optional[str] = None,
-        responses: Optional[Dict[int, AdditionalResponse]] = None,
+        responses: Optional[Dict[int, ResponseSpec]] = None,
         security: Optional[List[SecurityRequirement]] = None,
         summary: Optional[str] = None,
         tags: Optional[List[str]] = None,
@@ -623,7 +623,7 @@ class get(HTTPRouteHandler):
         operation_id: Optional[str] = None,
         raises: Optional[List[Type[HTTPException]]] = None,
         response_description: Optional[str] = None,
-        responses: Optional[Dict[int, AdditionalResponse]] = None,
+        responses: Optional[Dict[int, ResponseSpec]] = None,
         security: Optional[List[SecurityRequirement]] = None,
         summary: Optional[str] = None,
         tags: Optional[List[str]] = None,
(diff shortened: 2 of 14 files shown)


starlite/datastructures.py
```diff
@@ -368,3 +368,14 @@ async def close(self) -> None:
...
+
+class AdditionalResponse(BaseModel):
+    """Container type of additional responses."""
+
+    model: Type[BaseModel]
+    """A model that describes the content of the response"""
+    generate_examples: bool = True
+    """Generate examples for the response content"""
+    description: str = "Additional response"
+    """A description of the response"""
```
seladb, line 381: I'm not sure this is the right place for this class, please advise if you think we should move it

seladb (commented):
@Goldziher @peterschutt I assume we need to add documentation for this feature. I'll definately do that, but can you do a quick review to see if I'm my implementation makes sense?

starlite/handlers/http.py
```diff
@@ -268,6 +270,7 @@ def __init__(
         operation_id: Optional[str] = None,
         raises: Optional[List[Type[HTTPException]]] = None,
         response_description: Optional[str] = None,
+        responses: Optional[Dict[int, AdditionalResponse]] = None,
```
peterschutt, line 273: could this be called `additional_responses`?

starlite/handlers/http.py
```diff
@@ -309,6 +312,8 @@ def __init__(
             response_cookies: A list of [Cookie](starlite.datastructures.Cookie] instances.
             response_headers: A string keyed dictionary mapping [ResponseHeader][starlite.datastructures.ResponseHeader]
                 instances.
+            responses: A dictionary of additional status codes and a description of their expected content (which is
+                assumed to be 'application/json'). This information will be included in the OpenAPI schema
```
peterschutt, line 316: Do we need to assume json? You could have `AdditionalResponse.media_type` that defaults to `MediaType.JSON` but would allow customization if required.

starlite/datastructures.py
peterschutt: Seems reasonable enough to me.

peterschutt (commented):
Work looks good to me, just some minor comments in the review.

Taking a step back, there seems to be quite a bit of crossover between `responses` and `raises` here right?

E.g., with this handler:

```python
class ServerError(BaseModel):
    key: str

@get(
    "/",
    responses={"500": AdditionalResponse(model=ServerError, description="Server Error")},
)
def hello_world() -> ResponseType:
    """Hi."""
    return {"hello": "world"}
```
I get 

image

and with this:

```python
@get("/", raises=[InternalServerException])
def hello_world() -> ResponseType:
    """Hi."""
    return {"hello": "world"}
```
I get the actual exception response body mapped out:

image

Taking nothing away from the work you've done here, it's nice, but are there some example use-cases that aren't simply documenting exceptions that we can use to document the feature?

starlite/handlers/http.py
Goldziher: in the OA 3.1 specs this is called responses afaik. please verify this @seladb by looking at the `pydantic-openapi-schema` lib.

peterschutt: And if so, very happy to leave it as it is.

starlite/handlers/http.py
Goldziher: +1, media type should be specified per response type:

https://spec.openapis.org/oas/latest.html#responses-object

The thing is this, we need to be able to account for a use case that looks like this:


```python
@get("/some-path/{num: int}")
def handler(num: int) -> Response:
    if num == 1:
       return Response(media_type=MediaType.JSON, status_code=HTTP_200_OK, content={"hello": "world"})
    elif num == 2:
        return Response(media_type=MediaType.TEXT, status_code=HTTP_202_ACCEPTED, content="hello world")
    return Response(media_type=MediaType.HTML, status_code=HTTP_201_CREATED, content="<html><body><div>hello world</div></body></html>")
```

That is, the user can dynamically decide to return different media types, different status codes etc.

starlite/datastructures.py
Goldziher: I would suggest to rename this from `AdditionalResponse` to `ResponseSpec` or something along these lines. I would also move it out of `starlite/datastructures` and place it under `starlite/openapi/datastructures` - unless there is extra logic to add there, in which case its fine to keep in this loc.

starlite/handlers/http.py
Goldziher: it is: https://spec.openapis.org/oas/latest.html#responses-object

And the key in `operation` is called `responses`: https://spec.openapis.org/oas/latest.html#operation-object

starlite/openapi/responses.py
```diff
@@ -182,11 +200,16 @@ def create_responses(
...
+        ),
     }
     exceptions = route_handler.raises or []
     if raises_validation_error and ValidationException not in exceptions:
         exceptions.append(ValidationException)
     for status_code, response in create_error_responses(exceptions=exceptions):
         responses[status_code] = response
+
+    for status_code, response in create_additional_responses(route_handler, plugins):
+        if status_code not in responses:
```
Goldziher, line 212: how about raising an `ImproperlyConfiguredException` if it is already there? This happens on app bootstrap, so it should be ok and let the user know if there is a misconfiguration in place.

tests/openapi/test_responses.py
```diff
@@ -1,5 +1,7 @@
 from http import HTTPStatus
 
+import pytest
+from pydantic.main import BaseModel
```
Goldziher, line 4:
```suggestion
from pydantic import BaseModel
```

Goldziher (commented):
looks good

starlite/handlers/http.py
seladb: So we should leave it as it, right?
BTW, this is also how it's called in FastAPI...

peterschutt: Yep, leave it.

tests/openapi/test_responses.py
seladb: Addressed in https://github.com/starlite-api/starlite/pull/478/commits/dc2b02e52214f4ef2be388ce4d5f114615bea6bb

starlite/openapi/responses.py
seladb: Addressed in https://github.com/starlite-api/starlite/pull/478/commits/dc2b02e52214f4ef2be388ce4d5f114615bea6bb

starlite/handlers/http.py
seladb: Addressed in https://github.com/starlite-api/starlite/pull/478/commits/dc2b02e52214f4ef2be388ce4d5f114615bea6bb

starlite/datastructures.py
seladb: Addressed in https://github.com/starlite-api/starlite/pull/478/commits/dc2b02e52214f4ef2be388ce4d5f114615bea6bb

starlite/handlers/http.py
seladb: Also fixed `responses` documentation in https://github.com/starlite-api/starlite/pull/478/commits/8450baea0643b40be4c6332a0e894a32d01855d8

seladb:
I added documentation in https://github.com/starlite-api/starlite/pull/478/commits/fc8060cf81a9cde1fc2262d9038c1359ae5cc035 and https://github.com/starlite-api/starlite/pull/478/commits/1a28bfd7fb76e76bee5c40f66e42114faf0dcc37

Goldziher (approved):
Great!

Please update the reference docs before merging.

seladb:
> Great!
> 
> Please update the reference docs before merging.

@Goldziher I'm not familiar with `mkdocstrings`, how can I generate `.md` files from code?

seladb:
> > Great!
> > Please update the reference docs before merging.
> 
> @Goldziher I'm not familiar with `mkdocstrings`, how can I generate `.md` files from code?

Nevermind, I figured it out 😄 : https://github.com/starlite-api/starlite/pull/478/commits/a18a977dcc7a5b4d27398e38c20d711645fa880a

seladb:
@Goldziher I didn't notice the default merge option for PRs isn't "Squash and merge" 🤦‍♂️ (we should probably make it the default option).
Can you fix the commit tree in `main`?

Issue #455, opened 2022-09-04
Enhancement: Path Resolver For Route Handler Methods On A Controller


infohash:
**What's the feature you'd like to ask for.**
With `Starlite.route_handler_method_view`, you can resolve paths of route handler instance. So for example, if a decorator receives a route handler instance, it can resolve its path with this dictionary. But it raises `KeyError` when the route handler methods are desfined on a controller.

This is because `Starlite.Controllers.get_route_handlers` creates a shallow copy of the instance when creating a list of the route handlers defined on the controller. It means the instance registered in the `Starlite.route_handler_method_view` is different from what decorator received and when the decorator will try to resolve its path, it will raise `KeyError`.

```python
...
for f_name in route_handler_fields:
    source_route_handler = cast("BaseRouteHandler", getattr(self, f_name))
    route_handler = copy(source_route_handler)
    route_handler.owner = self
    route_handlers.append(route_handler)

return route_handlers
```

**Additional context**

```python
from starlite import Controller, Starlite, get


def route_decorator(route_handler_instance):

    def wrapper(*args, **kwargs):
            # do something
            ...
            try:
                app.route_handler_method_view[route_handler_instance])
            except KeyError:
                raise
            # do something
            ...
    return wrapper


class Star(Controller):

    path = "/test"

    @route_decorator
    @get()
    def test_handler(self) -> None:
        pass


@route_decorator
@get(path="/root")
def root_handler() -> None:
    pass


app = Starlite(route_handlers=[Star, root_handler])

**Comment by Goldziher** [comment]
Ok, @infohash - I dont see any down side to adding a PR for this so feel free to do so.

**Comment by infohash** [comment]
I tested the code without creating copy. It works fine but there must be a design reason behind this at the time of writing the code. Do you remember anything?

**Comment by Goldziher** [comment]
yes, it was meant to ensure there is a degree of encapsulation, i.e. not use the same memory addresses. 

I think we can drop this.

**Comment by infohash** [comment]
It turns out, it has more deep use case than just encapsulation. By creating a copy of the route handler of a controller, you preserve the route ownership of the route handler.

If we don't create copy, all 3 asserts will be true if they are checked like this.
```
first_router = Router(path="/first", route_handlers=[MyController])
assert first_router.routes[0].route_handlers[0].owner.owner is first_router

second_router = Router(path="/second", route_handlers=[MyController])
assert second_router.routes[0].route_handlers[0].owner.owner is second_router

[...]

PR #480, opened 2022-09-14
Path Resolver For Route Handler Methods On A Controller

infohash:
Implements #455

### DIFF
--- a/starlite/handlers/base.py
+++ b/starlite/handlers/base.py
@@ -60,7 +60,6 @@ class BaseRouteHandler(Generic[T]):
         "owner",
         "paths",
         "signature_model",
-        "_source_route_handler",
     )
 
     @validate_arguments(config={"arbitrary_types_allowed": True})
@@ -86,7 +85,6 @@ class BaseRouteHandler(Generic[T]):
         self.name = name
         self.opt: Dict[str, Any] = opt or {}
         self.owner: Optional[Union["Controller", "Router"]] = None
-        self._source_route_handler: Optional[Any] = None
         self.signature_model: Optional[Type["SignatureModel"]] = None
         self.paths = (
             {normalize_path(p) for p in path}
--- a/starlite/handlers/base.py
+++ b/starlite/handlers/base.py
@@ -60,6 +60,7 @@ class BaseRouteHandler(Generic[T]):
         "owner",
         "paths",
         "signature_model",
+        "_source_route_handler",
     )
 
     @validate_arguments(config={"arbitrary_types_allowed": True})
@@ -85,8 +86,13 @@ class BaseRouteHandler(Generic[T]):
         self.name = name
         self.opt: Dict[str, Any] = opt or {}
         self.owner: Optional[Union["Controller", "Router"]] = None
+        self._source_route_handler: Optional[Any] = None
         self.signature_model: Optional[Type["SignatureModel"]] = None
-        self.paths = {normalize_path(p) for p in path} if path and isinstance(path, list) else {normalize_path(path or "/")}  # type: ignore
+        self.paths = (
+            {normalize_path(p) for p in path}
+            if path and isinstance(path, list)
+            else {normalize_path(path or "/")}  # type: ignore
+        )
 
     @property
     def dependency_name_set(self) -> Set[str]:
--- a/starlite/router.py
+++ b/starlite/router.py
@@ -1,3 +1,4 @@
+import collections
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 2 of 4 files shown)


starlite/handlers/base.py
```diff
@@ -85,8 +86,13 @@ def __init__(
         self.name = name
         self.opt: Dict[str, Any] = opt or {}
         self.owner: Optional[Union["Controller", "Router"]] = None
+        self._source_route_handler: Optional[Any] = None
```
infohash, line 89: @Goldziher What should be the exact type hint here? The method that assigns value to this attribute can assign either of type `HTTPRouteHandler` or `WebSocketRouteHandler`.

Goldziher: It can be `BaseRouteHandler`, but I dont understand why these changes are required.

infohash: See `starlite.router.Router.route_handler_method_view`. Unlike `route_handler_method_map` which maps a path to alias route handlers, `route_handler_method_view` maps true route handler to its path so that decorators can resolve a path when given a route handler.

Goldziher:
So im trying to understand the use of this. From my PoV the use is hacky:

```python
  assert app.route_handler_method_view[handler] == ["/first/root", "/second/root"]
  assert app.route_handler_method_view[MyController.get_method] == ["/first/test", "/second/test"]
  assert app.route_handler_method_view[MyController.post_method] == ["/first/test", "/second/test"]
```

This requires a user facing API, not having users access a method that is basically meant for internal processing such as route_handler_method_view. 

See the app level method `Starlite.get_handler_index_by_name` as an example.

As for the implementation - you're currently saving the references, which is very inefficent in terms of performance - basically each route handler being copied has a reference to the original just for this mapping. 

Whats the actual use case we want to achieve with this change? users pushing a function as an argument to receive its path?

```python

    router = Router(path="/", route_handlers=[MyController])

    @get(path="/root")
    def handler() -> None:
        paths = router.get_paths_from_handler(MyController.get_method)
```

To me this API doesn't make much sense TBH. But even if it is, we don't need to save the entire route handler to achieve this, rather we can consider several different options to do this. The one I would go for looks like this:

Store the original route_handler.fn qualname as a kind of unique identifier on the router handler:

```starlite/handlers/base.py
    def _set_identifier(self):
        if not self.identifier:
            self.identifier = f"{self.__qualname__}::{self.fn.__qualname__}"
```

I would then use this value as a basis. No need for memory references etc, because it should be identical in the original and copy.

infohash:
@Goldziher Fair point. I'm thinking we can just set a private attribute that stores the path of the route handler. Then we can scrape `starlite.router.Router.route_handler_method_view` as it's not very useful.

starlite/router.py
```diff
@@ -208,18 +209,18 @@ def route_handler_method_map(
...
         Returns:
              A dictionary mapping route handlers to paths
         """
-        route_map: Dict[Union[ASGIRouteHandler, "HttpMethod", HTTPRouteHandler, WebsocketRouteHandler], str] = {}
+        route_map: Dict[str, List[str]] = collections.defaultdict(list)
         for route in self.routes:
             if isinstance(route, HTTPRoute):
                 for route_handler in route.route_handlers:
-                    route_map[route_handler] = route.path
+                    route_map[route_handler.fn.__qualname__].append(route.path)  # type: ignore
```
infohash, line 221: @Goldziher Can `route_handler.fn` ever be `None` here?

Goldziher: no

infohash: Okay, then `#  type: ignore` is fine.

infohash:
@Goldziher How do you like this now? I have changed `route_handler_method_view` to contain `__qualname__` maps to its all registered routes. Now, the decorators will be able to resolve the decorated route handler to its path.

Goldziher (approved):

tests/test_router_registration.py
```diff
@@ -125,9 +126,20 @@ def test_register_router_on_itself() -> None:
...
         ...
 
-    router = Router(path="/", route_handlers=[MyController, handler])
-    assert router.route_handler_method_view[handler] == "/test"
+    first_router = Router(path="/first", route_handlers=[MyController, handler])
+    second_router = Router(path="/second", route_handlers=[MyController, handler])
+
+    # Test routers.
+    assert first_router.route_handler_method_view[handler.fn.__qualname__] == ["/first/root"]  # type: ignore
+    assert second_router.route_handler_method_view[MyController.get_method.fn.__qualname__] == ["/second/test"]  # type: ignore
```
Goldziher, line 138: So using `__qualname__` is 99% sufficent but there are cases where it will not work as intended. Consider the following case:

```python
from starlite import put, post

def _handler(data: dict) -> dict:
    ...

put_handler = put("/")(handler)
post_handler = post("/")(handler)
```

Obviously this is a contrived example. The point though is that a truely unique identifier should consider a combination of the handler and fn qualnames. Hence my original example. 

Im leaving it to you to decide on the final implementation.

infohash: 2 route handler instances are decorating the same `fn` so their routes are adding up in the same route list, see the updates test case. I think this is fine. The point of this is to make it easy for the decorator to resolve all routes that goes through that `fn` so that the developer can add custom functionality based on these routes.

Goldziher: ok, as i said - i leave it to you to decide on the final implementation. please merge when ready.

PR #483, opened 2022-09-15
Issue #482 set media type on response container

### LINKED ISSUE #482: File response should accept byte stream and `media_type` as an alternative to file path.

As of v1.18.1: Starlite's [File](https://starlite-api.github.io/starlite/usage/5-responses/7-file-responses/) response only supports returning existing files.

``` python
return File(
        path=Path(Path(__file__).resolve().parent, "report").with_suffix(".pdf"),
        filename="repost.pdf",
    )

```

However, it should be able to support returning files directly from byte streams such as BytesIO. This would enable apps that can't use filesystem (eg. Google App Engine) to return files without much workaround.

In addition to that, File response should _optionally_ accept `media_type` as it is quite useful to set it dynamically depending on the file content.

Therefore, we should be able to return file by passing data and media type to File class like:
``` python
...
data = io.BytesIO()
... do something data..
return File(
        content=data,
        filename=filename,
        media_type='application/vnd.ms-excel',
        headers={f'Content-Disposition': f'attachment; filename={filename}'},
    )
```

❤️  Thank you very much :)

peterschutt:
Support setting `media_type` on `ResponseContainer`.

A specific use-case is returning files where the media type is unknown at time of route handler definition.

Closes #482 

# PR Checklist

### DIFF
--- a/starlite/datastructures.py
+++ b/starlite/datastructures.py
@@ -36,6 +36,7 @@ from starlette.responses import StreamingResponse
 from starlite_multipart.datastructures import UploadFile as MultipartUploadFile
 from typing_extensions import Literal, ParamSpec
 
+from starlite.enums import MediaType
 from starlite.openapi.enums import OpenAPIType
 
 P = ParamSpec("P")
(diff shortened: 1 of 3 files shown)


peterschutt:
A pretty small change in LOC terms. I'll back fill tests and docs once you guys are happy to go ahead with it. See #482 for reference

Goldziher (approved):

cofin (approved):

PR #494, opened 2022-09-18
Replace Starlette HTTPConnection, Request and WebSocket with Starlite classes

Goldziher:
# PR Checklist

This PR removes the dependency on Starlette for the `Request` and `WebSocket`, by introducing a new `ASGIConnection` class.


starlite/config/compression.py
```diff
@@ -87,11 +83,14 @@ def to_middleware(self, app: "ASGIApp") -> "MiddlewareProtocol":
...
             )
-        return handler(app=app, **self.dict())
+
+            return cast("ASGIApp", GZipMiddleware(app=app, **self.dict()))
+
+        from starlite.middleware.compression.brotli import (  # pylint: disable=import-outside-toplevel
+            BrotliMiddleware,
+        )
+
+        return BrotliMiddleware(app=app, **self.dict())
```
cofin, line 96: Does this need the same cast from above?

Goldziher: It doesn't. Because we use strong typing, the Starlette stuff needs to be casted.

cofin (approved):

PR #495, opened 2022-09-19
cli/cleo

Alex-CodeLab:
proposal to add Cleo CLI .

The idea is to have a somewhat standardized way to add new CLI commands.  
Simply add new commands in the `commands` folder.

example: 

            app = Starlite(route_handlers=[hello_world])
            
            if __name__ == '__main__':
                from  commands.cli import Cli
                cli = Cli(app)
                cli.run()  

.

          $ python main.py routes
          /some-path               hello_world              {'GET', 'HEAD'}
          /some-path/{some_id:int} hello_world              {'GET', 'HEAD'}
          /socket                  my_websocket_handler     set()
          /schema/redoc            redoc                    {'GET', 'HEAD'}
          /schema/openapi.json     retrieve_schema_json     {'GET', 'HEAD'}
          /schema/openapi.yaml     retrieve_schema_yaml     {'GET', 'HEAD'}
          /schema                  root                     {'GET', 'HEAD'}
          /schema/elements         stoplight_elements       {'GET', 'HEAD'}
          /schema/swagger          swagger_ui               {'GET', 'HEAD'}


commands/cli.py
```diff
@@ -0,0 +1,35 @@
...
+    usage example:
+        # ...
+        app = Starlite(route_handlers=[hello_world])
+        if __name__ == '__main__':
+            from commands.cli import Cli
+            cli = Cli(app)
+            cli.run()
+    """
+
+    def __init__(self, starletapp: Starlite):
```
Goldziher, line 23:
```suggestion
    def __init__(self, app: "Starlite"):
```

commands/cli.py
```diff
@@ -0,0 +1,35 @@
...
+        self.st_app = starletapp
+        self.cli = Application()
+        for command in commands.__all__:
+            cli_command = getattr(commands, command)
+            try:
+                self.cli.add(cli_command(self.st_app))
+            except Exception:
```
Goldziher, line 30: whats happening here?

commands/cli.py
```diff
@@ -0,0 +1,35 @@
...
+                self.cli.add(cli_command())  # pylint: disable = no-value-for-parameter
+
+    def run(self) -> int | Any:
```
Goldziher, line 33: int | Any is a weak typing. What does cli.run return?

commands/list_routes.py
```diff
@@ -0,0 +1,24 @@
+from cleo.commands.command import Command
+from cleo.helpers import argument, option
+
+from starlite import Starlite
+
+
+class ListRoutes(Command):  # type: ignore
```
Goldziher, line 7: why the type ignore?

commands/cli.py
Alex-CodeLab: It dynamically adds the commands, like   `cleo.add(my_command())`  ,  
however, some commands want access to the app,  like `cleo.add(ListRoutes(app))`  .

This tries both options, so it does not force anyone to have the `app` in their commands.  
I don't see another way to make this optional (but perhaps there is?).

commands/cli.py
Alex-CodeLab: it should be an `int` (afaik), but mypy does not like that. `error: Returning Any from function declared to return "int"  [no-any-return]`  (??)

However, it does accept this:

    def run(self) -> int:
        errorcode: int = self.cli_.run()
        return errorcode 

I think that should not be necessary. However, in this case it does make it a bit more readable.

commands/cli.py
Goldziher: Yes, you can use inspection. Look at how we check if a lifespan handler requires state.

commands/cli.py
Goldziher: So your issue is `cleo` has not been added to the `additional_dependencies` of MyPy and the other tools in `.precommit-config.yaml`. This You don't get proper type checks

commands/cli.py
```diff
@@ -0,0 +1,34 @@
+from cleo.application import Application
+
+import commands
+from starlite import Starlite
+
+
+class Cli:
```
cofin, line 7: Should this by `CLI`?

Goldziher: It should, you can add a proposal to change it: https://haacked.com/archive/2019/06/03/suggested-changes/

Goldziher:
any updates?

commands/cli.py
```diff
@@ -0,0 +1,35 @@
+from typing import Any
+
+from cleo.application import Application
+
+import commands
+from starlite import Starlite
+
+
+class Cli:
```
cofin, line 7:
```suggestion
class CLI:
```

commands/cli.py
cofin: TIL.  Thanks!

PR #497, opened 2022-09-20
Create `RequestFactory` to eventually replace `create_test_request`

seladb:
Addresses issue https://github.com/starlite-api/starlite/issues/459

To make it easier to review, this PR only adds `RequestFactory` and doesn't remove `create_test_request`. Once this is merged I'll open another PR to replace and remove `create_test_request`

The `RequestFactory` is heavily inspired by [Django's `RequestFactory`](https://docs.djangoproject.com/en/4.1/topics/testing/advanced/#the-request-factory)

# PR Checklist


starlite/testing.py
```diff
@@ -280,6 +281,188 @@ def test_my_handler() -> None:
...
+            type="http",
+            method=http_method,
+            server=(self.server, self.port),
+            root_path=self.root_path.rstrip("/"),
+            path=path,
+            headers=[],
+            app=self.app,
+            user=user,
+            auth=auth,
+        )
```
infohash, line 314: Also add `session` here so that the user can set pre-loaded session for testing. The `session` here will be a dictionary that contains cookie key as keys and cookie values as its values. We need to check if user has passed cookies. If passed, make them into a dictionary and assign that value to `session` when returning scope here. See the comment below.

starlite/testing.py
```diff
@@ -280,6 +281,188 @@ def test_my_handler() -> None:
...
+            ((key.lower()).encode("latin-1", errors="ignore"), value.encode("latin-1", errors="ignore"))
+            for key, value in headers.items()
+        ]
+
+    def _create_request_with_data(
+        self,
+        http_method: HttpMethod,
+        path: str,
+        headers: Optional[Dict[str, str]] = None,
+        cookies: Optional[Union[List["Cookie"], str]] = None,
```
infohash, line 346: This is where we can create a dictionary from cookies and pass it to `session`. See the comment above.

starlite/testing.py
```diff
@@ -280,6 +281,188 @@ def test_my_handler() -> None:
...
+        server: str = "test.org",
+        port: int = 3000,
+        root_path: str = "",
+    ):
+        self.app = app
+        self.server = server
+        self.port = port
+        self.root_path = root_path
+
+    def _create_scope(
```
infohash, line 297: Likewise, can we have `create_send` & `create_receive` also for the user to test his own middlewares. Better to keep these 3 public. We do need APIs for creation of `scope`, `send` & `receive`.

infohash:
@Goldziher Is it better to create `session` from `cookies` or should `session` take argument from the user directly?

Goldziher:
This looks good!

Please add doc strings to all methods, even private ones, and add a doc string to the `__init__` method which gives an explanation on the class etc. See for example the doc string in `starlite.datastructures.State.__init__`.

seladb:
> This looks good!
> 
> Please add doc strings to all methods, even private ones, and add a doc string to the `__init__` method which gives an explanation on the class etc. See for example the doc string in `starlite.datastructures.State.__init__`.

Added here: https://github.com/starlite-api/starlite/pull/497/commits/f68b46a01b34157045d4e6a2c8f7c0c743617ac3

starlite/testing.py
seladb: @infohash as per my understanding `Recieve` and `Send` are basically callables that take/return a `Message` instance. I'm not sure it's related to the request factory 🤔 

I probably missed something, can you please guide me on what you'd expect the methods to do?

infohash: When a middleware is called, it takes `scope`, `send`, & `receive` as arguments. We need these 3 for the user to test his own middlewares.

starlite/testing.py
```diff
@@ -280,6 +281,356 @@ def test_my_handler() -> None:
...
+
+        self._create_cookie_header(headers, cookies)
+        scope["headers"] = self._build_headers(headers)
+        return Request(scope=scope)  # type: ignore[arg-type]
+
+    def get(
+        self,
+        path: str,
+        headers: Optional[Dict[str, str]] = None,
+        cookies: Optional[Union[List["Cookie"], str]] = None,
```
infohash, line 470: Also add `session` as an argument here and in rest of the methods. It will take dictionary of type `Dict[str, Any]` as value from the user. Then assign this value to `scope['session']`.

seladb: `session` added here: **https://github.com/starlite-api/starlite/pull/497/commits/d871e917dc768b2bd00ffe2c346e00cb281d647f**

starlite/testing.py
seladb: Yeah I know, but I'm not sure how it's related to the `RequestFactory` and how you expect to use those methods/properties with it. 
Can you give an example of how you'd test a hypothetical middleware with a `RequestFactory` and those methods/properties?

infohash: I think it's out of scope of `RequestFactory`. I will raise a separate PR for this. Then `_create_scope` method here will be moved.

infohash (approved):

starlite/testing.py
seladb: ok sure. Please tag me on the PR, I'm curious what you have in mind 😃

seladb:
@Goldziher is it safe to merge?

peterschutt (approved):
Should add docs: https://starlite-api.github.io/starlite/usage/14-testing/#creating-a-test-request and also add the class and public members to reference docs: https://github.com/starlite-api/starlite/blob/main/docs/reference/testing.md

seladb:
> Should add docs: https://starlite-api.github.io/starlite/usage/14-testing/#creating-a-test-request and also add the class and public members to reference docs: https://github.com/starlite-api/starlite/blob/main/docs/reference/testing.md

@peterschutt I think I'll do it as part of the next PR which removes `create_test_request` otherwise it might be confusing to have 2 ways to create test request, what do you think? 🤔

Goldziher (approved):

Goldziher:
> @Goldziher is it safe to merge?

merge when ready

seladb:
> > Should add docs: https://starlite-api.github.io/starlite/usage/14-testing/#creating-a-test-request and also add the class and public members to reference docs: https://github.com/starlite-api/starlite/blob/main/docs/reference/testing.md
> 
> @peterschutt I think I'll do it as part of the next PR which removes `create_test_request` otherwise it might be confusing to have 2 ways to create test request, what do you think? 🤔

When thinking about it again, since `create_test_request` is exposed to users, can we actually remove it? If we do and someone uses it, it'd be a breaking change for them.
What do you think we should do?
- Keep both `create_test_request` and the new `RequestFactory` and mark `create_test_request` as deprecated? (if so, what's the best way to mark something as deprecated?)
- Or just remove `create_test_request` once this PR is merged?

This decision also affects the documentation we want to add about the new `RequestFactory`

peterschutt:
> When thinking about it again, since `create_test_request` is exposed to users, can we actually remove it? If we do and someone uses it, it'd be a breaking change for them.
> What do you think we should do?

It should be deprecated if we are to remove it. Can you re-implement it so that keeps the same interface but uses `RequestFactory` underneath?

starlite/testing.py
```diff
@@ -280,6 +281,374 @@ def test_my_handler() -> None:
...
+            http_method: The request's HTTP method.
+            session: A dictionary of session data.
+            user: A value for `request.scope["user"]`
+            auth: A value for `request.scope["auth"]`
+        Returns:
+            A dictionary that can be passed as a scope to the [Request][starlite.connection.Request] c'tor.
+        """
+
+        return dict(
+            type="http",
```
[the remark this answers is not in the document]
seladb: Fixed in https://github.com/starlite-api/starlite/pull/497/commits/97f26df61252a9471e5f767abddb452b3160bbfd

seladb:
> > When thinking about it again, since `create_test_request` is exposed to users, can we actually remove it? If we do and someone uses it, it'd be a breaking change for them.
> > What do you think we should do?
> 
> It should be deprecated if we are to remove it. Can you re-implement it so that keeps the same interface but uses `RequestFactory` underneath?

Yes sure, that's a good idea. So I'll update the documentation in this PR and mention `create_test_request` as deprecated. In a following PR I'll remove all of our internal usages of `create_test_request` and re-implement this method to use `RequestFactory`. @peterschutt does it work?

peterschutt:
> Yes sure, that's a good idea. So I'll update the documentation in this PR and mention `create_test_request` as deprecated. In a following PR I'll remove all of our internal usages of `create_test_request` and re-implement this method to use `RequestFactory`. @peterschutt does that work?

Sounds reasonable to me. If it isn't hard to implement it using `RequestFactory` then I wouldn't even worry about the deprecation - saves people having to rewrite tests if we can keep the interface.

seladb:
> > Yes sure, that's a good idea. So I'll update the documentation in this PR and mention `create_test_request` as deprecated. In a following PR I'll remove all of our internal usages of `create_test_request` and re-implement this method to use `RequestFactory`. @peterschutt does that work?
> 
> Sounds reasonable to me. If it isn't hard to implement it using `RequestFactory` then I wouldn't even worry about the deprecation - saves people having to rewrite tests if we can keep the interface.

ok sure, I'll do it in a separate PR

PR #499, opened 2022-09-20
Feature: Implements `on_app_init` hook

peterschutt:
# `on_app_init` callbacks

## Issue with the current `PluginProtocol.on_app_init()` system.

- The `Plugin.on_app_init()` callback occurs before route handler registration, so plugins do not 
have the opportunity to receive the handlers passed to the application before they are applied to 
the app.
- Certain application attributes have pre-processing applied to them on assignment to self. 
Therefore, if a plugin wants to manage or modify these attributes, the plugins need to perform this
step of pre-processing, or at least understand it, before they can operate on these configuraion 
attributes. E.g., `as_async_callable_list()`.
- Difficult to be confident in longevity of implementations. Simple things like order of 
configuration/attribute changes, or changes to wrapper objects in later releases may impact the way 
plugins need to work to operate on certain configuration values.
- We run the risk of a plugin that abuses internal apis in `PluginProtocol.on_app_init()` becoming popular and making a rod for our backs in terms of internal refactoring.

## Proposed solution

- We construct a pydantic model of the arguments that the application object receives, called 
`AppConfig`. We already use the `validate_arguments()` decorator, that constructs a signature 
model to validate arguments passed to the app constructor, so this would just be replacing that 
validation with an explicit model.
- The arguments to the starlite constructor are used to instantiate this `AppConfig` object.
- After the configuration object has been instantiated with the parameters passed to `Starlite`, 
any handlers registered for `on_app_config` callback receive the `AppConfig` instance and must 
return it.
- Config callbacks are passed the `AppConfig` object in the order that they are passed to 
the `Starlite` constructor and must define their own behavior and recommendations for where the 
plugin should be inserted into the callback list.
- Benefit of this is that plugins can be built around the certainty of semantic versioning guarantees of 
the `Starlite` interface.

## Proposed implementation

The implementation is totally backward compatible.

Registered `on_app_config` hook handlers must receive and return an instance of `AppConfig`.

The `AppConfig` object is first instantiated from parameters passed to `Starlite.__init__()`, and then passed to each of the registered hook handlers.

The resultant `AppConfig` instance is used to instantiate the application.

Removes the use of `pydantic.validate_arguments()` on `Starlite.__init__()` as the instantiation of the `AppConfig` object does the same thing.

There is a test to ensure that the parameters of the `Starlite` constructor are matched by properties on `AppConfig`.

There is a test that ensures all properties on the `AppConfig` object are accessed within the `Starlite` constructor.

# PR Checklist

### DIFF
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -257,7 +257,6 @@ class Starlite(Router):
 
         # creates app config object from parameters
         config = AppConfig(
-            route_handlers=route_handlers,
             after_exception=after_exception or [],
             after_request=after_request,
             after_response=after_response,
@@ -285,6 +284,7 @@ class Starlite(Router):
             response_class=response_class,
             response_cookies=response_cookies or [],
             response_headers=response_headers or {},
+            route_handlers=route_handlers,
             security=security or [],
             static_files_config=static_files_config or [],
             tags=tags or [],
@@ -333,7 +333,7 @@ class Starlite(Router):
         for plugin in self.plugins:
             plugin.on_app_init(app=self)
 
-        for route_handler in route_handlers:
+        for route_handler in config.route_handlers:
             self.register(route_handler)
 
         if self.openapi_config:
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -174,7 +174,7 @@ class Starlite(Router):
         static_files_config: Optional[Union["StaticFilesConfig", List["StaticFilesConfig"]]] = None,
         tags: Optional[List[str]] = None,
         template_config: Optional["TemplateConfig"] = None,
-    ):
+    ) -> None:
         """The Starlite application.
 
         `Starlite` is the root level of the app - it has the base path of "/" and all root level
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -1,8 +1,6 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Dict, List, Optional, Set, Union, cast
 
-from pydantic import validate_arguments
-from pydantic_openapi_schema.v3_1_0 import SecurityRequirement
 from starlette.middleware import Middleware as StarletteMiddleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.middleware.trustedhost import TrustedHostMiddleware
@@ -14,15 +12,7 @@ from starlite.asgi import (
     RouteMapNode,
     StarliteASGIRouter,
 )
-from starlite.config import (
-    CacheConfig,
-    CompressionConfig,
-    CORSConfig,
-    CSRFConfig,
-    OpenAPIConfig,
-    StaticFilesConfig,
-    TemplateConfig,
-)
+from starlite.config import AppConfig, CacheConfig, OpenAPIConfig
 from starlite.datastructures import State
 from starlite.exceptions import ImproperlyConfiguredException
 from starlite.handlers.asgi import asgi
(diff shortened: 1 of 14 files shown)


cofin (approved):

starlite/app.py
```diff
@@ -141,42 +139,42 @@ class Starlite(Router):
...
-        exception_handlers: Optional[ExceptionHandlersMap] = None,
-        guards: Optional[List[Guard]] = None,
-        middleware: Optional[List[Middleware]] = None,
-        on_shutdown: Optional[List[LifeSpanHandler]] = None,
-        on_startup: Optional[List[LifeSpanHandler]] = None,
+        dependencies: Optional[Dict[str, "Provide"]] = None,
+        exception_handlers: Optional["ExceptionHandlersMap"] = None,
+        guards: Optional[List["Guard"]] = None,
+        middleware: Optional[List["Middleware"]] = None,
+        on_app_config: Optional[List["OnAppConfigHandler"]] = None,
```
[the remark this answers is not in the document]
Goldziher: I'd say so 😉

starlite/app.py
```diff
@@ -243,50 +240,89 @@ def __init__(
...
-        self.plugins = plugins or []
         self.route_map: RouteMapNode = {}
         self.routes: List[BaseRoute] = []
         self.state = State()
-        self.static_files_config = static_files_config
-        self.template_engine = create_template_engine(template_config)
-        super().__init__(
+
+        # creates app config object from parameters
+        app_conf = AppConfig(
```
Goldziher, line 255: I'd call this `config` rather than `app_conf` for brevity. 

Reason: I'm personally allergic to abbreviations like this. So I'd go for either  either `app_config` or `config`.

Goldziher (approved):
LGTM.

Please update the docs before merging, and also add attribute comments in the AppConfig class.

starlite/app.py
peterschutt: `config` it is - have to repeat it a lot so was trying to keep the var size down.

starlite/app.py
```diff
@@ -243,68 +244,107 @@ def __init__(
...
+            tags=config.tags,
         )
         self._init = True
 
         for plugin in self.plugins:
             plugin.on_app_init(app=self)
 
-        for route_handler in route_handlers:
+        for route_handler in config.route_handlers:
             self.register(route_handler)
```
peterschutt, line 337: @Goldziher `Router.register()` gets called twice for these route handlers. Once when passed to `Router.__init__()` in `super().__init__()` call above, and again now, when `Starlite.register()` calls `super().register()` - is this intentional, or can we leave the routes out of `Router.__init__()` call?

starlite/app.py
```diff
@@ -243,68 +244,107 @@ def __init__(
...
+            dependencies=config.dependencies,
+            exception_handlers=config.exception_handlers,
+            guards=config.guards,
+            middleware=config.middleware,
+            parameters=config.parameters,
+            path="",
+            response_class=config.response_class,
+            response_cookies=config.response_cookies,
+            response_headers=config.response_headers,
+            route_handlers=config.route_handlers,
```
peterschutt, line 327:
```suggestion
            route_handlers=[],
```
See https://github.com/starlite-api/starlite/pull/499#discussion_r978227790

peterschutt: tests pass as we go on to register them after the `Router.__init__()` call.

starlite/app.py
Goldziher: Aight

Goldziher (approved):

PR #501, opened 2022-09-21
Reduce cognitive complexity

jonadaly:
# PR Checklist

This PR reduces the cognitive complexity as flagged by SonarCloud (see #203) in the following modules:
- typing.py
- router.py
- parsers.py
- kwargs.py

### DIFF
--- a/starlite/kwargs.py
+++ b/starlite/kwargs.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Dict,
     List,
+    Mapping,
     NamedTuple,
     Optional,
     Set,
@@ -24,6 +25,7 @@ from pydantic.fields import (
     ModelField,
     Undefined,
 )
+from starlette.datastructures import URL
 
 from starlite.constants import (
     EXTRA_KEY_IS_PARAMETER,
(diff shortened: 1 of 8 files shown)


starlite/kwargs.py
```diff
@@ -332,51 +334,48 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
+        if "request" in self.expected_reserved_kwargs:
+            reserved_kwargs["request"] = connection
+        if "socket" in self.expected_reserved_kwargs:
+            reserved_kwargs["socket"] = connection
+        if "data" in self.expected_reserved_kwargs:
+            reserved_kwargs["data"] = self._get_request_data(request=cast("Request", connection))
+        return {**reserved_kwargs, **path_params, **query_params, **header_params, **cookie_params}
+
+    @staticmethod
+    def _collect_params(params: Mapping[str, Any], expected: Set[ParameterDefinition], url: URL) -> Dict[str, Any]:
```
jonadaly, line 373: This is marginally slower than what was happening before (because there are two iterations through the required params) but has the advantage of picking up all missing params, rather than just the first.

starlite/parsers.py
```diff
@@ -1,6 +1,6 @@
 from contextlib import suppress
 from functools import reduce
-from typing import TYPE_CHECKING, Any, Dict, List, Tuple, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Union, cast
```
jonadaly, line 3: The lgtm tool is complaining about this unused import, but it is required for the cast on line 46 (otherwise flake8 complains)

starlite/parsers.py
```diff
@@ -67,11 +67,11 @@ def parse_form_data(media_type: "RequestEncodingType", form_data: "FormMultiDict
         if not isinstance(value, MultipartUploadFile):
```
jonadaly, line 67: The LGTM tool is complaining about an unused import on line 14, but removing the import (or marking it with `# noqa` is causing the flake8 pre-commit check to fail (because the import is used on line 48).

If anyone has suggestions for how to resolve this I would be grateful!

starlite/kwargs.py
Goldziher: speed is not a concern here because the logic in this file is running once - on application startup, and pre-computing the results.

starlite/router.py
```diff
@@ -223,32 +221,49 @@ def route_handler_method_view(
                 route_map[cast("WebSocketRoute", route).route_handler.fn.__qualname__].append(route.path)  # type: ignore
         return route_map
 
-    @staticmethod
+    RouteHandlerMap = ItemsView[
```
Goldziher, line 224: you can move it to `starlite.types.composite`, although its fine to leave in place as well.

starlite/router.py
```diff
@@ -223,32 +221,49 @@ def route_handler_method_view(
...
+        if isinstance(value, BaseRouteHandler):
+            return cls._map_route_handlers_for_base_route_handler(value)
+        # we are dealing with a controller
+        return cls._map_route_handlers_for_controller(value)
+
+    @staticmethod
+    def _map_route_handlers_for_base_route_handler(value: BaseRouteHandler) -> RouteHandlerMap:
+        """Maps route handlers to HTTP methods for an input
+        BaseRouteHandler."""
+        handlers_map: Dict[str, Any] = {}
```
Goldziher, line 247: the correct type here and in other places is:

```python
Dict[str, Union[WebsocketRouteHandler, ASGIRouteHandler, Dict["HttpMethod", HTTPRouteHandler]]
```

Consider creating a type alias for `Union[WebsocketRouteHandler, ASGIRouteHandler, Dict["HttpMethod", HTTPRouteHandler]` and using it where pertinent

starlite/router.py
```diff
@@ -223,32 +221,49 @@ def route_handler_method_view(
...
+            if isinstance(value, HTTPRouteHandler):
+                handlers_map[path] = {http_method: value for http_method in value.http_methods}
+            elif isinstance(value, (WebsocketRouteHandler, ASGIRouteHandler)):
+                handlers_map[path] = value
+        return handlers_map.items()
+
+    @staticmethod
+    def _map_route_handlers_for_controller(value: Controller) -> RouteHandlerMap:
+        """Maps route handlers to HTTP methods for an input Controller."""
+        handlers_map: Dict[str, Any] = collections.defaultdict(dict)
```
Goldziher, line 258: why a defaultdict?

tests/test_parsers.py
```diff
@@ -11,11 +15,38 @@ def test_parse_query_params() -> None:
         "polluting": False,
     }
     request = create_test_request(query=query)  # type: ignore
-    result = parse_query_params(connection=request)
+    result = parsers.parse_query_params(connection=request)
```
Goldziher, line 18: i usually prefer to not use a namespace, i.e. `parse_query_params` rather than `parsers.parse_query_params`

Goldziher (approved):
Looks very good!

Goldziher:
Are you on our discord? If so - shoot me a line and let's have a chat.

starlite/router.py
jonadaly: Because it avoids the check to see whether the key already exists in the dict - it removes the need for the first two lines here:

```
if not isinstance(handlers_map.get(path), dict):
    handlers_map[path] = {}
for http_method in route_handler.http_methods:
    handlers_map[path][http_method] = route_handler
```

tests/test_parsers.py
jonadaly: Happy to go with the flow

Goldziher:
invite to the org was sent to you. you can merge once youre done.

starlite/router.py
jonadaly: Looking a bit deeper, I can't really see why the type here shouldn't be:

`Dict[str, Dict["HttpMethod", Union[WebsocketRouteHandler, ASGIRouteHandler, HTTPRouteHandler]]`

Unless there's something I'm missing about the WebsocketRouteHandler and ASGIRouteHandler types that makes them dict-like?

tests/test_parsers.py
peterschutt: When using namespaces, I like when things are named to read nicer/avoid repetition, e.g.,  `parse.query_params` rather than `parsers.parse_query_params`. 

All good though, just throwing in my 2c, use whatever naming you are happy with/get rid of the namespace altogether as @Goldziher suggests.

starlite/parsers.py
peterschutt: Pycharm has this problem too:

image

Their inspections mustn't be wired to destringify the first argument to cast.

I think the flake8 is working properly and problem with lgtm.

There is an lgtm ignore `# lgtm[py/unused-import]`([ref](https://lgtm.com/rules/6770079/), [ref](https://lgtm.com/help/lgtm/alert-suppression#python))

peterschutt: See their note though:

> If you've [enabled](https://lgtm.com/help/lgtm/managing-automated-code-review#enabling-pr-integration) automated code review of pull requests for your projects, it's important to note that alert suppression changes made in a pull request are only applied when that pull request is merged. So, if you suppress an alert in a pull request, LGTM will continue to report the related alert for as long as the pull request is open. For more information, see [Automated code review FAQs](https://lgtm.com/help/lgtm/automated-code-review-faqs#why-automated-code-review-reports-misses-alert).

jonadaly: I guess I'll leave it alone then!

PR #506, opened 2022-09-22
Update __init__ return typing to be consistent

Harshal6927:
# PR Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,6 @@ jobs:
             v1-venv-${{ runner.os }}-${{ matrix.python-version }}
             v1-venv-${{ runner.os }}
       - name: Install dependencies
-        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
         run: poetry install --no-interaction --no-root
       - name: Set pythonpath
         run: echo "PYTHONPATH=$PWD" >> $GITHUB_ENV
--- a/docs/usage/7-middleware/1-using-middleware/2-middleware-call-order.md
+++ b/docs/usage/7-middleware/1-using-middleware/2-middleware-call-order.md
@@ -25,7 +25,7 @@ def test_middleware_call_order() -> None:
 
     def create_test_middleware(middleware_id: int) -> type[MiddlewareProtocol]:
         class TestMiddleware(MiddlewareProtocol):
-            def __init__(self, app: ASGIApp):
+            def __init__(self, app: ASGIApp) -> None:
                 self.app = app
 
             async def __call__(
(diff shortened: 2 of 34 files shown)


Goldziher:
Ok, if we want to go this path we should actually use the `NoReturn` type in all init methods.

peterschutt:
> Ok, if we want to go this path we should actually use the `NoReturn` type in all init methods.

Mypy doesn't let you do that:

```
src/worker/parse.py:14: error: The return type of "__init__" must be None  [misc]
Found 1 error in 1 file (checked 92 source files)
```

Seems [`NoReturn`](https://docs.python.org/3/library/typing.html#typing.NoReturn) is for functions that can never return, e.g., always raise:

> Special type indicating that a function never returns.

Some refs:
- https://github.com/python/mypy/issues/10342#issuecomment-822960700
- https://github.com/microsoft/pyright/issues/3188#issuecomment-1066015281

Goldziher:
Aight

Harshal6927:
I added return type because initializer of MakoTemplateEngine class have None as an return type. It is for the constancy.

peterschutt:
@Harshal6927 why don't we broaden this PR out and annotate the return for all `__init__()` methods that need it, at least that way we are consistent everywhere:

image

Harshal6927:
> @Harshal6927 why don't we broaden this PR out and annotate the return for all `__init__()` methods that need it, at least that way we are consistent everywhere:
> 
> image

Should I do it?

peterschutt:
> Should I do it?

I think so, best to be consistent.

And [pep recommended](https://peps.python.org/pep-0484/#the-meaning-of-annotations):

> (Note that the return type of `__init__` ought to be annotated with -> None. The reason for this is subtle. If `__init__` assumed a return annotation of -> None, would that mean that an argument-less, un-annotated `__init__` method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that `__init__` ought to have a return annotation; the default behavior is thus the same as for other methods.)

Harshal6927:
@peterschutt Ok I'll update my PR.

jonadaly (approved):

.github/workflows/ci.yaml
```diff
@@ -45,7 +45,6 @@ jobs:
             v1-venv-${{ runner.os }}-${{ matrix.python-version }}
             v1-venv-${{ runner.os }}
       - name: Install dependencies
-        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
Goldziher, line 48: so this change should be reverted, once done this can be merged.

Goldziher:
@all-contributors add @Harshal6927 code

PR #520, opened 2022-09-27
Initialize Default Value Of Session As An Empty Dictionary For Test RequestFactory

infohash:


starlite/testing.py
```diff
@@ -352,6 +352,8 @@ def _create_scope(
         Returns:
             A dictionary that can be passed as a scope to the [Request][starlite.connection.Request] c'tor.
         """
+        if session is None:
+            session = {}
```
seladb, line 356: I wonder if we should take care of this in the `ASGIConnection` level, or at least make `scope` a pydantic model instead of a simple `dict`. That way issues like this won't happen... what do you think?

infohash: Agreed. Scopes will be used very often to test middlewares.

seladb: So do you think we should make `scope` a pydantic model?
If yes I can work on that...
cc @Goldziher

Goldziher: Sure. It's already a typed duct though

infohash: Yes.

seladb: Yes that's correct, I'll check if the type checker complains if I put the wrong type (it should I guess?)

Issue #524, opened 2022-09-28
Update RequestFactory


Goldziher:
This is a followup issue.

We should update the RequestFactory in the following:

1. extend the scope generation to support most things we define in the scope type (see `starlite.types.asgi_types`)
2. add documentation in the usage/testing docs regarding how to use this.

seladb:
@Goldziher I looked into everything we have in `scope` and is being used in starlite's `Request` and `ASGIConnection`. Here are properties I can add:
- `state: Dict[str, Any]`
- `client: Tuple[str, int]`
- `route_handler: "RouteHandlerType"`

The following keys are in `scope` but are not being used anywhere in starlite's request object, hence I don't think we should add them:
- `asgi: ASGIVersion`
- `http_version: str` (I checked in Starlette and it is always set to a const value of `1.1`)
- `raw_path: bytes`
- `extensions: Dict[str, Dict[object, object]]`

There is also `path_params: Dict[str, str]` which is created from the request path, but it's not easy to parse and I don't think it adds much value, so I don't think we should add it.

Please let me know what you think. I'll work on this in the coming days.

Goldziher:
So, the scope object actually originates from the ASGI server (uvicorn, daphne etc.). If the user uses an ASGI server that supports HTTP2, such as `hypercorn`, then `http_version` will be `2.0`. In the future there might be `3.0` versions etc. So its good to have it. 

`raw_path` should equal path - only in bytes. it shouldnt be passed in. 

`extensions` are specified by the ASGI specification, this is a value passed by the ASGI server - again. The only place we use it is for `push promise` - which is undocumented currently. Still, users might have use cases for this - although its true this is really marginal and we dont need to cater to this. 

`path_params` - yes, its not easy to set at all - but maybe allow users to set the value, otherwise default it to an empty dict.

PR #532, opened 2022-09-30
Update `RequestFactory`

seladb:
- Addresses #524 
- Convert the `scope` used in the `RequestFactory` from `dict` to `HTTPScope` (typed-dict)
- Add the missing parameters to the `scope`
- Removed documentation for `create_test_request` which was removed from the codebase

# PR Checklist


starlite/testing.py
```diff
@@ -368,8 +384,14 @@ def _create_scope(
...
             auth=auth,
             query_string=urlencode(query_params, doseq=True).encode() if query_params else b"",
-            path_params=[],
+            path_params=path_params,
             client=(self.server, self.port),
+            state=state,
+            asgi=ASGIVersion(spec_version="3.0", version="3.0"),
+            http_version=http_version or "1.1",
+            raw_path=path.encode("ascii"),
+            route_handler=None,  # type: ignore[typeddict-item]
```
seladb, line 393: I'm not sure what is the right way to let users choose the `route_handler`. As per my understanding, there is no easy way to extract the handler from the `app` object, and letting the users choose just any handler is probably not useful. Any idea how to implement that?

starlite/testing.py
```diff
@@ -368,8 +384,14 @@ def _create_scope(
             user=user,
             auth=auth,
             query_string=urlencode(query_params, doseq=True).encode() if query_params else b"",
-            path_params=[],
+            path_params=path_params,
             client=(self.server, self.port),
+            state=state,
+            asgi=ASGIVersion(spec_version="3.0", version="3.0"),
```
seladb, line 390: I'm not sure this is the right way to build `ASGIVersion`. Please let me know if there's a better way

starlite/testing.py
infohash: Is route handler required to construct scope?

seladb: It's a non-optional parameter in [`BaseScope`](https://github.com/starlite-api/starlite/blob/d6313981c9129da2f88efb09a9d14424733c781b/starlite/types/asgi_types.py#L78) but not required to create the `Request` object

infohash: Can you use `app.route_handler_method_map().keys()`. This will give you all the registered route handler instances.

seladb: But what if the app starts with `route_handlers=[]` or if it starts with more than one `route_handlers`? Which one should we choose?

Goldziher: i wouldnt auto pick a value - i would suggest the following:

1. create a default placeholder, and use it if the user doesnt pass anything.
2. allow the user to pass a handler in manually. Consider this:

```python
@get()
def handler() -> None:
   ...

request = RequestFactory().get(path="/", route_handler=handler)
```

infohash: Using raw_path and method, you can resolve its associated route handler by using `app.route_handler_method_map`.

```python
app.route_handler_method_map['/test'][method]
```

seladb: Should the handler be of type `RouteHandlerType`, as the type in [`BaseScope`](https://github.com/starlite-api/starlite/blob/d6313981c9129da2f88efb09a9d14424733c781b/starlite/types/asgi_types.py#L78)?

Goldziher: yup

Goldziher (approved):

tests/utils/test_sequence.py
```diff
@@ -16,6 +16,6 @@ def y() -> None:
         pass
 
     unique_functions = unique([x, x, y, y])
-    assert unique_functions == [x, y]
+    assert unique_functions == [x, y] or [y, x]
```
seladb, line 19: Fixing a flaky test (not related to this PR)

PR #533, opened 2022-09-30
Feature: Preloaded Session For Test Client

infohash:
To mock raw session cookies as if they are coming from the browser, this feature will allow the user to create raw session cookies for the test requests. If the session middleware is enabled by the user in tests, the raw session cookies will be loaded by the session middleware and the route handler will see preloaded session. The user will have to pass `SessionCookieConfig` instance to the class because it contains key.

# How To Use It In Tests?
```python
import os

import pytest
from pydantic import SecretBytes
from starlite.middleware.session import SessionCookieConfig


@pytest.fixture(scope="class")
def session_config(self) -> SessionCookieConfig:
    return SessionCookieConfig(secret=SecretBytes(os.urandom(16)))


@pytest.fixture()
def app(self, session_config: SessionCookieConfig) -> Starlite:
    @get(path="/test")
    def my_handler() -> None:
        pass

    # Set up session middleware.
    return Starlite(route_handlers=[my_handler], middleware=[session_config.middleware])


def test_something(app: Starlite, session_config: SessionCookieConfig) -> None:
    with TestClient(app=app, session_config=session_config) as client:
        cookies = client.create_session_cookies(session_data={"user": "test_user"})
        # Pass raw cookies to the request.
        client.get(url="/test", cookies=cookies)


# Now your route handler will have preloaded session.
```

From now `RequestFactory` methods will have default `path` set to `"/"`. These methods are called frequently during tests to mock request so unless the user really cares about specific request path, keep it default.


Goldziher (approved):
Looks good! Before merging please:

1. add this method to the docs/reference/testing section.
2. add this method to the docs/usage/testing secion.

PR #539, opened 2022-10-01
updated plugin

Goldziher:
# PR Checklist

### DIFF
--- a/starlite/plugins/sql_alchemy/config.py
+++ b/starlite/plugins/sql_alchemy/config.py
@@ -1,0 +1,166 @@
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, Union
+
+from orjson import OPT_SERIALIZE_NUMPY, dumps, loads
+from pydantic import BaseConfig, BaseModel
+from sqlalchemy.orm import Query, Session
+from typing_extensions import Literal
+
+from starlite.exceptions import MissingDependencyException
+from starlite.middleware.base import DefineMiddleware
+from starlite.utils import default_serializer
+
+from ... import BaseLoggingConfig, LoggingConfig
+
+try:
+    from sqlalchemy import create_engine
+    from sqlalchemy.engine import Engine
+    from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, create_async_engine
+    from sqlalchemy.future import Engine as FutureEngine
+    from sqlalchemy.pool import Pool
+
+    from .middleware import SQLAlchemySessionMiddleware
+except ImportError as e:
+    raise MissingDependencyException("sqlalchemy is not installed") from e
+
+if TYPE_CHECKING:
+    from starlite.datastructures import State
+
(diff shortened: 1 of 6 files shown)


starlite/plugins/sql_alchemy/middleware.py
```diff
@@ -0,0 +1,41 @@
...
+            session_class: SQLAlchemy Session class to use.
+            **kwargs: Any configuration to pass to the session class.
+        """
+        self.app = app
+        self.session_scope_key = session_scope_key
+        self.engine_app_state_key = engine_app_state_key
+        self.session_class = session_class
+        self.kwargs = kwargs
+
+    async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> None:
```
cofin, line 38: Should the middleware handle the transaction commit/rollback or no?

starlite/plugins/sql_alchemy/config.py
```diff
@@ -0,0 +1,218 @@
...
+    class Config(BaseConfig):
+        arbitrary_types_allowed = True
+
+    connection_string: str
+    middleware_class: Type[SQLAlchemySessionMiddleware] = SQLAlchemySessionMiddleware
+    create_engine_callable: Callable[[str], Union[Engine, FutureEngine]] = create_engine
+    create_async_engine_callable: Callable[[str], AsyncEngine] = create_async_engine
+    create_async_engine: bool = True
+    session_class: Optional[Type[T]] = None
+    session_scope_key: str = "db_session"
```
cofin, line 124: I think we need a `session_maker_callable` and `async_session_maker_callable`

starlite/plugins/sql_alchemy/config.py
```diff
@@ -0,0 +1,218 @@
...
+        arbitrary_types_allowed = True
+
+    connect_args: Optional[Dict[str, Any]] = None
+    echo: Optional[bool] = None
+    echo_pool: Optional[bool] = None
+    enable_from_linting: Optional[bool] = None
+    future: bool = True
+    hide_parameters: Optional[bool] = None
+    isolation_level: Optional[IsolationLevel] = None
+    json_deserializer: Callable[[str], Any] = loads
```
cofin, line 90: `json_deserializer` is not implemented on all engines.  If the engine doesn't have a native JSON type in SQL, then this parameter will fail.  It should be excluded for those engines.  The `json_serializer` function should exist, so it is likely fine.

Goldziher: i added an `engine_supports_json` boolean. The reason is that sqlite does support it, from certain versions, as do other DBs, so this puts it at the hand of the user.

starlite/plugins/sql_alchemy/config.py
Goldziher: it just one callable - its the sessionmaker, which gets a different session types.

Goldziher: I addded:

```python
    session_maker: Type[sessionmaker] = sessionmaker
```

cofin (approved):
Nice work on this.  I think this is going to be really useful.

Issue #540, opened 2022-10-02
Proper indexed route reverse


jtraub:
**What's the feature you'd like to ask for.**
Sometimes I need to redirect client app to another endpoint. To do that I need to generate an url for this endpoint.

Starlite provides `get_handler_index_by_name` which returns a dict with original path but it can not be used as an url path in case it contains path parameters. So in order to create an url to a handler like this

```python
@get(path="/complex/{param:int}/path/{param2:str}/", name="complex-path")
def handler(param: int, param2: str) -> str:
    ...
```

user would have to parse path and replace parameters with actual values.

I think it would be very nice to have a function like Django's [`reverse`](https://docs.djangoproject.com/en/4.1/ref/urlresolvers/#reverse) which would accept route index name and path parameters as kwargs.

jtraub:
My attempt to implement such function can be seen here

https://github.com/jtraub/starlite/compare/main...jtraub:starlite:route_reverse

I think included tests give an idea about function behaviour

Goldziher:
That's pretty cool. Wanna add a PR?

Also, check this bit if code out:
https://github.com/starlite-api/starlite/blob/main/starlite/router.py#L198

This was added by @infohash , and it's a full mapping of all route handlers, via their `handler.fn.__qualname__` to paths. It's a better source than the name index, because it doesn't rely on names as such. But it's not exposed via a method that retrieves the path, it's just a mapping for now. 

We need a solution that normalizes these two tbh.

jtraub:
> That's pretty cool. Wanna add a PR?

Sure. I will add docs tomorrow and then submit a PR.

>  full mapping of all route handlers, via their handler.fn.__qualname__ to paths.

I saw it when I went through sources.
I still think that names are better than FQN in such cases because names can be short, concise (compare `group_info` vs `app.core.api.group.info`) and do not depend on codebase organization. Another benefit of named resources is that this concept is familiar to developers coming from other frameworks (aiohttp, fastapi, django, flask and many others). 

However, merging this piece of code and route names together would allow to resolve a few problems.
* named routes can not be registered multiple times under different routers
* named routes do not work with optional path parameters

Goldziher:
Ok, go ahead and merge these. We dont need this on the router level IMO. Please correct me if im wrong @infohash - but we can have the logic only on the app instance, and have a single method for it.

infohash:
[...]

PR #541, opened 2022-10-03
Add route_reverse function

jtraub:
# PR Checklist

This PR addresses feature requested in #540

I should note that `route_reverse` can accept strings instead of `datetime`, `date`, `time`, `timedelta`, `float`, `Path`.
This is the deliberate decision.

Starlite parses path parameters using pydantic functions like `parse_datetime`, `parse_time` and others. These functions parse dates and times in various formats and return python objects but straight `str` invocation over these objects doesn't always return initial string. See 

```
>>> from pydantic.datetime_parse import parse_time
>>> t = parse_time('14:30')
>>> str(t)
'14:30:00'
>>> 
```

Initially I considered adding format strings and performing strict checking types of passed kwargs but the function interface became too complex. Also I am aiming to wrap `route_reverse` into a template function and expose it in Jinja2/Make which is another reason for keeping its interface simple.


Goldziher:
Let me know when you want this reviewed

jtraub:
@Goldziher could you please take a look at the code?

starlite/app.py
```diff
@@ -512,26 +600,52 @@ def _add_node_to_route_map(self, route: BaseRoute) -> RouteMapNode:
...
+        if isinstance(route, (WebSocketRoute, ASGIRoute)):
+            route_handlers.append(route.route_handler)
+        else:
+            route_handlers.extend(cast("HTTPRoute", route).route_handlers)
+
+        return route_handlers
+
+    def _get_handler_qualname(self, handler: "RouteHandlerType") -> str:
+        """Retrieves __qualname__ for passed route handler."""
+        if handler.fn and hasattr(handler.fn, "__qualname__"):
```
Goldziher, line 615: this is a place to use `cast` because there is no way to get a handler without `.fn` - it will raise an exception before reaching this point. 

```python
   return cast("AnyCallable", handler.fn).__qualname__
```

Or something along these lines.

starlite/app.py
```diff
@@ -512,26 +600,52 @@ def _add_node_to_route_map(self, route: BaseRoute) -> RouteMapNode:
...
+            return handler.fn.__qualname__
+
+        return type(handler.fn).__qualname__
+
+    def _store_handler_to_route_mapping(self, route: BaseRoute) -> None:
+        """Stores mappings:
+
+            * between route handlers and routes;
+            * between route handlers and names.
```
Goldziher, line 624: Please remove the bullet points- its incorrect style wise. 

```
Stores the mapping of route handlers to routes and to route handler names. 
```

Alternatively, add a `Notes:` section.

starlite/app.py
```diff
@@ -512,26 +600,52 @@ def _add_node_to_route_map(self, route: BaseRoute) -> RouteMapNode:
...
 
-        for route_handler in filter(lambda x: x.name, route_handlers):
-            name = cast("str", route_handler.name)
-            if name in self._route_handler_index:
+        for handler in route_handlers:
+            qualname = self._get_handler_qualname(handler)
+            self._route_mapping[qualname].append(route)
+
+            name = handler.name
+            if not name:
```
Goldziher, line 639: so, i think we should revisit this. I would suggest we dont force users to define a handler.name to get this functionality, instead we can do one of the following:

```python
name = handler.name or cast("AnyCallable", handler.fn).__name__
```

This might be quite intuitive to use - the problem is of course some people will have functions with identical names. 

So we can also instead do this:

```python
name = handler.name or cast("AnyCallable", handler.fn).__qualname__
```

Certainly not so intuitive, but will work.

starlite/app.py
```diff
@@ -438,15 +442,99 @@ def handler() -> None:
...
+            else:
+                output.append(component)
+
+        return join_paths(output)
+
+    @property
+    def route_handler_method_view(self) -> Dict[str, List[str]]:
+        """
+        Returns:
+             A dictionary mapping route handlers to paths
```
Goldziher, line 531:
```suggestion
             A dictionary mapping route handlers to paths.
```

starlite/app.py
```diff
@@ -438,15 +442,99 @@ def handler() -> None:
...
+            ValidationException: If path parameters are missing in **path_parameters or have wrong type.
+
+        Returns:
+            A fully formatted url path or None.
+        """
+        handler_index = self.get_handler_index_by_name(name)
+        if handler_index is None:
+            return None
+
+        allow_str_instead = [datetime, date, time, timedelta, float, Path]
```
Goldziher, line 498: not sure it will work, but better to use a set for this because it way faster.

Goldziher (approved):
Very nice work

starlite/app.py
jtraub: @Goldziher so your idea is to use `__qualname__` as default name for handler indexes. Did I get this right?

starlite/app.py
jtraub: @Goldziher I wasn't sure if `handler.fn` is not None at all cases during route tree building but since you say so we can drop `not handler.fn` check.

However, we can not get rid of `hasattr(handler.fn, "__qualname__")` since `.fn` is a callable and an instance of a class with `__call__` method is a callable too but it has no `__qualname__` attr.

I think If I remove this check the method would fail for `StaticFiles` which is a Starlette class with `__call__` method. I haven't checked it manually though but I explored the code in order to see how I can create a template function which generates links to static files.

starlite/app.py
jtraub: It works with set just fine

jtraub:
@Goldziher I've added default name (__qualname__) for routes and tests for this as well as fixed typos and formatting issues in the docs and docstrings

Goldziher:
> @Goldziher I've added default name (__qualname__) for routes and tests for this as well as fixed typos and formatting issues in the docs and docstrings

Should I merge this or is it waiting for more commits?

Goldziher:
@all-contributors add @jtraub code

PR #554, opened 2022-10-06
Add support for using custom `Request` and `WebSocket` classes.

Goldziher:
# PR Checklist

This PR adds support for using custom subclasses of `Request` and `WebSocket` in the application by passing these to the Starlite constructor. It forms a solid basis for extension further via plugins. For example, consider an `SQLAlchemyRequest` class that has properties to access the session etc.

### DIFF
--- a/docs/reference/config/0-app-config.md
+++ b/docs/reference/config/0-app-config.md
@@ -27,6 +27,7 @@
             - openapi_config
             - parameters
             - plugins
+            - request_class
             - response_class
             - response_cookies
             - response_headers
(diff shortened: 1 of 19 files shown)


cofin (approved):

odiseo0 (commented):
Should we fix the warning before we merge or it doesn't matter?

cofin:
We should fix it

cofin:
Is this an actual error?  It sure looks like `Any` is still being used to me: 
```python
request: "Request[Any, Any]" = scope["app"].request_class(scope=scope)
```

cofin:
I think this is an erroneous error.  Pre-commit also completed successfully and it would have raised an error if it was actually unused.

PR #549, opened 2022-10-06
Updates SQLAlchemy plugin type detection.

peterschutt:
Makes detection of declarative model types and instances compatible with 1.x and 2.0.

Closes #548

# PR Checklist

- [x] Have you followed the guidelines in `CONTRIBUTING.md`?
- [x] Have you got 100% test coverage on new code?
- [x] Have you updated the prose documentation?
- [x] Have you updated the reference documentation?

PR #556, opened 2022-10-07
Makes SQLAlchemyConfig.connection_string optional.

### LINKED ISSUE #555: Testing: DeprecationWarning: Setting per-request cookies
```
tests/test_testing.py::test_test_client[True-session_data0]
tests/test_testing.py::test_test_client[False-session_data1]
  /home/peter/.cache/pypoetry/virtualenvs/starlite-CEGnKvDD-py3.10/lib/python3.10/site-packages/httpx/_client.py:800: DeprecationWarning: Setting per-request cookies=<...> is being deprecated, because the expected behaviour on cookie persistence is ambiguous. Set cookies directly on the client instance instead.
    warnings.warn(message, DeprecationWarning)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

HTTPX are deprecating setting cookies per request.

```python
@pytest.mark.parametrize("enable_session, session_data", [(True, {"user": "test-user"}), (False, {})])
def test_test_client(enable_session: bool, session_data: Dict[str, str]) -> None:
    ...
    with TestClient(app=app, session_config=session_config if enable_session else None) as client:
        cookies = client.create_session_cookies(session_data=session_data)
        client.get("/test", cookies=cookies)
        assert app.state.value == 1
```

Easy fix to silence the warning:

```python
@pytest.mark.parametrize("enable_session, session_data", [(True, {"user": "test-user"}), (False, {})])
def test_test_client(enable_session: bool, session_data: Dict[str, str]) -> None:
    ...
    with TestClient(app=app, session_config=session_config if enable_session else None) as client:
        client.cookies = client.create_session_cookies(session_data=session_data)
        client.get("/test")
        assert app.state.value == 1
```

Perhaps we should consider adding `TestClient.set_session_cookies()` method as well.

peterschutt:
One of either `connection_string` or `engine_instance` must be set.

Adds some tests for #549.

Closes #555

# PR Checklist

### DIFF
--- a/starlite/plugins/sql_alchemy/config.py
+++ b/starlite/plugins/sql_alchemy/config.py
@@ -12,7 +12,7 @@ from typing import (
 )
 
 from orjson import OPT_SERIALIZE_NUMPY, dumps, loads
-from pydantic import BaseConfig, BaseModel, validator
+from pydantic import BaseConfig, BaseModel, root_validator, validator
 from typing_extensions import Literal
 
 from starlite.config.logging import BaseLoggingConfig, LoggingConfig
@@ -147,7 +147,7 @@ class SQLAlchemyConfig(BaseModel):
     class Config(BaseConfig):
         arbitrary_types_allowed = True
 
-    connection_string: str
+    connection_string: Optional[str] = None
     """Database connection string in one of the formats supported by SQLAlchemy.
 
     Notes:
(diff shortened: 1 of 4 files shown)


starlite/plugins/sql_alchemy/config.py
```diff
@@ -236,6 +236,29 @@ def validate_before_send_handler(  # pylint: disable=no-self-argument
...
+        Args:
+            values: Field values, after validation.
+
+        Returns:
+            Field values.
+        """
+        connection_string = values.get("connection_string")
+        engine_instance = values.get("engine_instance")
+        if connection_string is None and engine_instance is None:
+            message = "One of 'connection_string' or 'engine_instance' must be provided."
```
Goldziher, line 255: I would simply raise here. Then change elif to if. Same goes after

Goldziher (approved):

starlite/plugins/sql_alchemy/config.py
peterschutt: yeh agree

PR #560, opened 2022-10-07
Adds `TestClient.update_session_cookies()`

### LINKED ISSUE #557: Feature: Add TestClient.add_session_cookies() method
See #555 for context.

httpx are deprecating setting cookies per request, e.g,. `client.get(..., cookies=...)`.

We now have some code in our test suite like this:

```python
    with TestClient(app=app, session_config=session_config if enable_session else None) as client:
        client.cookies = client.create_session_cookies(session_data=session_data)
        client.get("/test")
        assert app.state.value == 1
```

We could add a method `add_session_cookies()`, so the pattern would become:

```python
    with TestClient(app=app, session_config=session_config if enable_session else None) as client:
        client.add_session_cookies(session_data=session_data)
        client.get("/test")
        assert app.state.value == 1
```

peterschutt:
Due to httpx deprecation of passing cookies to `Client.request` it makes sense to have a method that updates the cookies instead of retuning them.

Closes #557

# PR Checklist

### DIFF
--- a/starlite/testing/test_client.py
+++ b/starlite/testing/test_client.py
@@ -144,6 +144,54 @@ class TestClient(StarletteTestClient):
         encoded_data = self.session.dump_data(data=session_data)
         return {f"{self.session.config.key}-{i}": chunk.decode("utf-8") for i, chunk in enumerate(encoded_data)}
 
+    def update_session_cookies(self, session_data: Dict[str, Any]) -> None:
+        """Creates raw session cookies that are loaded into session by the
+        Session Middleware. It simulates cookies the same way as if they are
+        coming from the browser. Your tests must set up session middleware to
+        load raw session cookies into the session.
+
+        Cookies are added to `TestClient.cookies`.
+
+        Examples:
+
+            ```python
+            import os
+
+            import pytest
+            from pydantic import SecretBytes
+            from starlite.middleware.session import SessionCookieConfig
+            from starlite.testing import TestClient
+
+
+            @pytest.fixture(scope="class")
+            def session_config(self) -> SessionCookieConfig:
+                return SessionCookieConfig(secret=SecretBytes(os.urandom(16)))
+
+
+            @pytest.fixture()
+            def app(self, session_config: SessionCookieConfig) -> Starlite:
+                @get(path="/test")
+                def my_handler() -> None:
+                    pass
+
+                # Set up session middleware.
+                return Starlite(route_handlers=[my_handler], middleware=[session_config.middleware])
+
+
+            def test_something(app: Starlite, session_config: SessionCookieConfig) -> None:
+                with TestClient(app=app, session_config=session_config) as client:
+                    client.update_session_cookies(session_data={"user": "test_user"})
+                    # Pass raw cookies to the request.
+                    client.get(url="/test")
+            ```
+        """
+        if self.session is None:
+            return
+        encoded_data = self.session.dump_data(data=session_data)
+        self.cookies.update(
+            {f"{self.session.config.key}-{i}": chunk.decode("utf-8") for i, chunk in enumerate(encoded_data)}
+        )
+
 
 def create_test_client(
     route_handlers: Union["ControllerRouterHandler", List["ControllerRouterHandler"]],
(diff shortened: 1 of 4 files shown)


starlite/testing/test_client.py
```diff
@@ -144,6 +144,54 @@ def test_something(app: Starlite, session_config: SessionCookieConfig) -> None:
...
+                    client.update_session_cookies(session_data={"user": "test_user"})
+                    # Pass raw cookies to the request.
+                    client.get(url="/test")
+            ```
+        """
+        if self.session is None:
+            return
+        encoded_data = self.session.dump_data(data=session_data)
+        self.cookies.update(
+            {f"{self.session.config.key}-{i}": chunk.decode("utf-8") for i, chunk in enumerate(encoded_data)}
```
infohash, line 192: This will overwrite the existing session keys because `enumerate` will initialize the value of `i` from 0 which will lead to the creation of `session-0`, `session-1`, ... . If these keys already exist in `self.cookies`, they will be overwritten.

peterschutt: yes, had a choice of using 'update' semantics or 'set' semantics.. in this context, the client will be starting with an empty cookie jar 99% of the time right - so shouldn't matter, or does it? I don't care which way it goes TBH. Change it?

peterschutt: nvm I see what you're saying - I'll change, good pickup. Wonder why that test is passing??

infohash:
It returns whatever cookies that session middleware creates for the response. The `cookies` argument in the request methods of httpx used to overwrite the cookies that were set on the `httpx.Client` instance. Although, this has been deprecated, if the user wants to overwrite the cookies in his new test request, he can do this:

```python
@pytest.fixture()
def test_client(self, app: Starlite, session_config: SessionCookieConfig) -> TestClient:
    with TestClient(app=app, base_url=self.BASE_URL, session_config=session_config) as client:
        yield client


def test_something(test_client: TestClient) -> None:
    response  = client.get(url="/test", cookies=cookies)
    ...
```
The cookies in the `response.cookies` are implicitly carry forwarded to the next request so the cookies are already updating the same way we see in browsers. But let's say if the user explicitly wants to overwrite cookies, `httpx` wants us to overwrite it from the instance like this:

```python
test_client.cookies = test_client.create_session_cookies(session_data={'new': 'cookie'})
```

Is there anything that I am missing?

peterschutt:
> Is there anything that I am missing?

No, I was just fixing the deprecation warning in the tests and thought that using the client to generate cookies that are set on the client may as well be a distinct method.

```python
test_client.cookies = test_client.create_session_cookies(session_data={'new': 'cookie'})
# vs
test_client.set_session_cookies(session_data={"new": "cookie"})
```
Correct I was wrong to use 'update', but do you think not even worth creating the `set._session_cookies()` method? If so I'll just close this off.

tests/test_testing.py
```diff
@@ -196,12 +196,13 @@ def start_up_handler(state: State) -> None:
     @get(path="/test")
     def test_handler(state: State, request: Request) -> None:
         assert state.value == 1
-        assert request.session == session_data
+        assert request.session == {**session_data, "more": "data"}
```
peterschutt, line 199: @wallrod this should have failed right?

infohash:
```python
test_client.set_session_cookies(session_data={"new": "cookie"})
```
This can be done to consolidate the two line of code that I am currently doing:
```python
cookies = test_client.create_session_cookies(session_data={'new': 'cookie'})
test_client.cookies = cookies
```
Okay, make `create_session_cookies` method as private and from `set_session_cookies`, call `_create_session_cookies` which will assign the value to `self.cookies`.

tests/test_testing.py
infohash: It should. Let me check.

infohash:
However, the utility of consolidating the code won't be that much. Here's the example that I'm doing in my tests:

```python
cookies = test_client.create_session_cookies(session_data=request_factory.session)
test_client.cookies = cookies
end_session_redirect_response = test_client.get(url=self.POST_LOGOUT_REDIRECT_PATH, follow_redirects=False)
```

You must be wondering why am I storing raw session cookies in a separate variable. That's because in some case, I repeat the mock request to the handler and assert its behaviour. For that I reuse the raw session cookies that I stored in the variable.

peterschutt:
> Okay, make `create_session_cookies` method

It's in the ref docs, I don't think we should just make it private.. I'm starting to think we just close this off.

PR #563, opened 2022-10-07
Inject request into template context

jtraub:
# PR Checklist

This PR introduces implicit request injection into template context. Injected request will override existing `request` value that user might have passed into context.

### DIFF
--- a/docs/usage/15-templating.md
+++ b/docs/usage/15-templating.md
@@ -50,6 +50,9 @@ The `name` kwarg passed to the `Template` class is the filename for the given te
 directories specifies for this file until it finds it or an exception will be raised. The `context` kwarg is a
 dictionary specifying context data that is passed to the engine.
 
+Starlite automatically injects current request into context under `request` name. Note that if you try pass any value
+under `request` key it will be overridden with the current request.
+
 ## Defining a Custom Template Engine
 
 If you wish to use another templating engine, you can easily do so by
--- a/starlite/datastructures/response_containers.py
+++ b/starlite/datastructures/response_containers.py
@@ -34,6 +34,7 @@ from starlite.enums import MediaType
 if TYPE_CHECKING:
 
     from starlite.app import Starlite
+    from starlite.connection import Request
     from starlite.response import TemplateResponse
 
 
@@ -59,7 +60,12 @@ class ResponseContainer(GenericModel, ABC, Generic[R]):
 
     @abstractmethod
     def to_response(
-        self, headers: Dict[str, Any], media_type: Union["MediaType", str], status_code: int, app: "Starlite"
+        self,
+        headers: Dict[str, Any],
+        media_type: Union["MediaType", str],
+        status_code: int,
+        app: "Starlite",
+        request: "Request",
     ) -> R:  # pragma: no cover
         """Abstract method that should be implemented by subclasses. Returns a
         Starlette compatible Response instance.
@@ -69,6 +75,7 @@ class ResponseContainer(GenericModel, ABC, Generic[R]):
             media_type: A string or member of the [MediaType][starlite.enums.MediaType] enum.
             status_code: A response status code.
             app: The [Starlite][starlite.app.Starlite] application instance.
+            request: A [Request][starlite.connection.request.Request] instance.
 
         Returns:
             A Response Object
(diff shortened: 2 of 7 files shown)


starlite/datastructures/response_containers.py
```diff
@@ -59,7 +60,12 @@ class Config(BaseConfig):
 
     @abstractmethod
```
Goldziher, line 61: why do you need to update the abstract method with the `request` kwarg? you can define it only for `Template` IMO.

starlite/datastructures/response_containers.py
```diff
@@ -213,11 +240,12 @@ def to_response(
         from starlite.exceptions import ImproperlyConfiguredException
         from starlite.response import TemplateResponse
 
+        context = self.context or {}
         if not app.template_engine:
             raise ImproperlyConfiguredException("Template engine is not configured")
         return TemplateResponse(
             background=self.background,
-            context=self.context,
+            context={**context, "request": request},
```
Goldziher, line 248: you dont need to hardcode it here in a way that will overide the user. I would suffest doing this:

```python
context = self.context or {}
context.setdefault(request=request)
```

starlite/response.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         template_name: str,
         template_engine: "TemplateEngineProtocol",
         status_code: int,
-        context: Optional[Dict[str, Any]] = None,
+        context: Dict[str, Any],
```
Goldziher, line 111: why is this required?

starlite/datastructures/response_containers.py
Goldziher: ok, i see why you need this value.

docs/usage/15-templating.md
```diff
@@ -50,6 +50,9 @@ The `name` kwarg passed to the `Template` class is the filename for the given te
 directories specifies for this file until it finds it or an exception will be raised. The `context` kwarg is a
 dictionary specifying context data that is passed to the engine.
 
+Starlite automatically injects current request into context under `request` name. Note that if you try pass any value
```
Goldziher, line 53: I dont think this is the way to go, instead:

```text
Starlite injects the current Request instance into the template context under the `request` key by default. 
```

Goldziher (commented):
Nice work! Some small adjustments and we are good to go.

starlite/datastructures/response_containers.py
jtraub: Uhm, as a user I would prefer to be sure that some template globals won't be overridden accidentally allowing me to simplify jinja2/make function code and some of my tests.

starlite/response.py
jtraub: At the point context is guaranteed to be a dict with `request` so removing `Optional` is just making typing a bit more strict.

tests/handlers/http/test_to_response.py
```diff
@@ -120,7 +120,7 @@ def test_function() -> Response:
     with create_test_client(test_function) as client:
         http_route: HTTPRoute = client.app.routes[0]  # type: ignore
         route_handler = http_route.route_handlers[0]
-        response = await route_handler.to_response(data=route_handler.fn(), plugins=[], app=None)  # type: ignore
+        response = await route_handler.to_response(data=route_handler.fn(), plugins=[], app=None, request=None)  # type: ignore
```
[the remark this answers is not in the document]
jtraub: @Goldziher I wanted to keep test_to_response.py simple - hence passing `None` instead of `request` as we need to test that `handler.to_response` returns proper response and request doesn't really matter here.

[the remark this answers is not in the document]
Goldziher: sure but it will save you on a lot of type ignores. just make one request instance nad use it - will actually save you code.

starlite/datastructures/response_containers.py
Goldziher: ok.

tests/handlers/http/test_to_response.py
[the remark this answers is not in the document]
jtraub: It is 4AM for me so I will update the code tomorrow in the morning.

I should note though that `# type: ignore` comments won't go away because of `app=None` and `data=route_handler.fn()` arguments

starlite/datastructures/response_containers.py
jtraub: @Goldziher sorry, I am a bit tired and didn't understand your last comment.

Do you want me to leave it as is (forbidding the user to override `request` global in templates) or remove overriding? If you want me to change the behavior I will do it tomorrow.

Goldziher: Ok. Let's go with your preference

PR #569, opened 2022-10-08
Fix route handler name indexing

### LINKED ISSUE #567: Default name in handler index is not unique
**Describe the bug**
Default name for route handlers might be not unique.

[Here](https://github.com/starlite-api/starlite/pull/541#discussion_r987565608) we decided to use `handler.fn.__qualname__` as default name in handler index but today I learned the hard way that `__qualname__` is not unique if your route handlers are functions with the same name located in different files. Sorry, I didn't check python docs carefully to make sure that picked named would be unique.

Easy repro
```
/tmp/test 
> cat a.py 
def create():
    pass

/tmp/test 
> cat b.py 
def create():
    pass

/tmp/test 
> python
Python 3.10.7 (main, Sep  6 2022, 21:22:27) [GCC 12.2.0] on linux
>>> from a import create as create_a
>>> from b import create as create_b
>>> create_a.__qualname__
'create'
>>> create_b.__qualname__
'create'
>>> 
```

Django uses `func.__module__ + "." + func.__qualname__' for similar purposes as can be see [here](https://github.com/django/django/blob/974942a75039ba43e618f6a5ff95e08b5d5176fd/django/contrib/admindocs/utils.py#L22)

I tried it https://github.com/starlite-api/starlite/compare/main...jtraub:starlite:fix_default_name_in_index and it works fine for my case.
Maybe "internal" method `app._get_handler_qualname` should be moved to utils so the user can import and use it - at the moment I ha to duplicate the code in my app

Goldziher:
# PR Checklist

This PR addresses and resolves #567

### DIFF
--- a/starlite/handlers/base.py
+++ b/starlite/handlers/base.py
@@ -211,6 +211,18 @@ class BaseRouteHandler(Generic[T]):
         if not self.fn:
             raise ImproperlyConfiguredException("Cannot call _validate_handler_function without first setting self.fn")
 
+    @property
+    def identifier(self) -> str:
+        """
+
+        Returns:
+            A path like identifier for the route handler.
+        """
+        target = cast("Any", self.fn)
+        if not hasattr(target, "__qualname__"):
+            target = type(target)
+        return f"{type(self).__name__}::{target.__module__}.{target.__qualname__}"
+
     def __str__(self) -> str:
         """
         Returns:
@@ -218,7 +230,4 @@ class BaseRouteHandler(Generic[T]):
         """
         if self.name:
             return self.name
-        target = cast("Any", self.fn)
-        if not hasattr(target, "__qualname__"):
-            target = type(target)
-        return f"{target.__module__}.{target.__qualname__}"
+        return self.identifier
--- a/starlite/handlers/base.py
+++ b/starlite/handlers/base.py
@@ -211,23 +211,12 @@ class BaseRouteHandler(Generic[T]):
         if not self.fn:
             raise ImproperlyConfiguredException("Cannot call _validate_handler_function without first setting self.fn")
 
-    @property
-    def identifier(self) -> str:
-        """
-
-        Returns:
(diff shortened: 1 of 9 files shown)


starlite/handlers/base.py
```diff
@@ -210,3 +210,24 @@ def _validate_handler_function(self) -> None:
...
+    def identifier(self) -> str:
+        """
+
+        Returns:
+            A path like identifier for the route handler.
+        """
+        target = cast("Any", self.fn)
+        if not hasattr(target, "__qualname__"):
+            target = type(target)
+        return f"{type(self).__name__}::{target.__module__}.{target.__qualname__}"
```
jtraub, line 224: I am a bit unsure about using `type(self).__name__` here.

The user code will be depending on "internal" class names of Starlite (correct me if I am wrong but I haven't imported and  used HTTPRouteHandler in my code yet) and also Starlite internal code will have troubles accessing handlers which use custom user classes.

For example, right now I am experimenting with serving static files and related template functions. In short, I added a class derived from `starlette.static_files.StaticFiles` which introduces a bit of custom logic so I can host some of my static files on S3 and also I added a template function `asset('css', 'myfile.css')` which generates url to myfile.css taking into account "path" parameter in `StaticFilesConfig`.

Internally `asset` uses handler index to find all handlers for static files and inspects them.

Obviously with this change I won't be able to share this function with my other projects, not propose to include into Starlite.

Goldziher: This should be the name of the current class. If you think it's a problem I will remove this.

jtraub: Another reason why I don't like it is that simple `{target.__module__}.{target.__qualname__}` works better because I can use [importlib](https://docs.python.org/3/library/importlib.html) and import class/function for inspection/instantiation without having to remove the prefix first.

Goldziher: ok, done. I simplified it as well. Please review 😉

PR #574, opened 2022-10-10
Fix mypy and tests failure

jtraub:
# PR Checklist

I am experimenting a lot now and failing mypy checks/tests are so irritating so I decided to fix them at least locally. If you've already fixed it in your code then close the PR.

Due to recent change signature change of `ModelFactory.is_constrained_field` (went from `bool` to `TypeGuard`) in pydantic-factories mypy was able to pickup that `is_constrained_field` actually has been returning true for `ConstrainedFrozenSet` for almost a year :-) while union in the appropriate type annotation in Starlite didn't include it.

Also, test is fixed which is now failing due to changes in starlite-multipart. I went and fixed the test itself since according to RFCS (https://www.rfc-editor.org/rfc/rfc7230#section-3.2 and https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2 ) header names are case insensitive.

P.S. Actually it seems that Starlite and pydantic-factories miss handling for `ConstrainedDate` which is present in pydantic.

### DIFF
--- a/tests/kwargs/test_multipart_data.py
+++ b/tests/kwargs/test_multipart_data.py
@@ -72,7 +72,7 @@ async def form_with_headers_handler(request: Request) -> Dict[str, Any]:
                 "filename": value.filename,
                 "content": content.decode(),
                 "content_type": value.content_type,
-                "headers": list(value.headers.items()),
+                "headers": [[name.lower(), value] for name, value in value.headers.items()],
             }
         else:
             output[key] = value
@@ -188,9 +188,9 @@ def test_multipart_request_multiple_files_with_headers(tmpdir: Any) -> None:
                 "content": "<file2 content>",
                 "content_type": "text/plain",
                 "headers": [
-                    ["Content-Disposition", 'form-data; name="test2"; filename="test2.txt"'],
+                    ["content-disposition", 'form-data; name="test2"; filename="test2.txt"'],
                     ["x-custom", "f2"],
-                    ["Content-Type", "text/plain"],
+                    ["content-type", "text/plain"],
                 ],
             },
         }
--- a/starlite/openapi/schema.py
+++ b/starlite/openapi/schema.py
@@ -8,6 +8,7 @@ from pydantic import (
     ConstrainedBytes,
     ConstrainedDecimal,
     ConstrainedFloat,
+    ConstrainedFrozenSet,
     ConstrainedInt,
     ConstrainedList,
     ConstrainedSet,
@@ -93,7 +94,7 @@ def create_string_constrained_field_schema(field_type: Union[Type[ConstrainedStr
 
 
 def create_collection_constrained_field_schema(
-    field_type: Union[Type[ConstrainedList], Type[ConstrainedSet]],
+    field_type: Union[Type[ConstrainedList], Type[ConstrainedSet], Type[ConstrainedFrozenSet]],
     sub_fields: Optional[List[ModelField]],
     plugins: List["PluginProtocol"],
 ) -> Schema:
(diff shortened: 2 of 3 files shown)


tests/kwargs/test_multipart_data.py
```diff
@@ -188,9 +188,9 @@ def test_multipart_request_multiple_files_with_headers(tmpdir: Any) -> None:
                 "content": "<file2 content>",
                 "content_type": "text/plain",
                 "headers": [
-                    ["Content-Disposition", 'form-data; name="test2"; filename="test2.txt"'],
+                    ["Content-Disposition".lower(), 'form-data; name="test2"; filename="test2.txt"'],
```
Goldziher, line 191:
```suggestion
                    ["content-disposition", 'form-data; name="test2"; filename="test2.txt"'],
```

tests/kwargs/test_multipart_data.py
```diff
@@ -188,9 +188,9 @@ def test_multipart_request_multiple_files_with_headers(tmpdir: Any) -> None:
...
                     ["x-custom", "f2"],
-                    ["Content-Type", "text/plain"],
+                    ["Content-Type".lower(), "text/plain"],
```
Goldziher, line 193:
```suggestion
                    ["content-type", "text/plain"],
```

Goldziher (approved):

Goldziher:
according to the ASGI specification:

```text
headers (Iterable[[byte string, byte string]]) – An iterable of [name, value] two-item iterables, where name is the header name, and value is the header value. Order of header values must be preserved from the original HTTP request; order of header names is not important. Duplicates are possible and must be preserved in the message as received. Header names should be lowercased, but it is not required; servers should preserve header case on a best-effort basis. Pseudo headers (present in HTTP/2 and HTTP/3) must be removed; if :authority is present its value must be added to the start of the iterable with host as the header name or replace any existing host header already present.
```

see: https://asgi.readthedocs.io/en/latest/specs/www.html

jtraub:
> according to the ASGI specification

@Goldziher I think that headers test outside of ASGI related stuff should not rely on case because we are using `httpx` and some 3rd party packages which know nothing about ASGI spec.

However, it would be great to narrow down current failure and fix it in starlite-multipart so it preserves header case

tests/kwargs/test_multipart_data.py
jtraub: Fixed both in a force pushed change to keep commit message clean

Goldziher:
> > according to the ASGI specification
> 
> @Goldziher I think that headers test outside of ASGI related stuff should not rely on case because we are using `httpx` and some 3rd party packages which know nothing about ASGI spec.
> 
> However, it would be great to narrow down current failure and fix it in starlite-multipart so it preserves header case

I dont follow. What do you want to do in starlite-multipart?

jtraub:
> I dont follow. What do you want to do in starlite-multipart?

Here starlite-multipart https://github.com/starlite-api/starlite-multipart/blob/main/starlite_multipart/parser.py#L27 lowers headers.

I think it would be great to preserve case if possible (unless there is too much work needed for that)

jtraub:
Uhm, for some reason I am not getting mypy failures regarding type: ignore comments.

Will try to reinstall pre-commit hook and fix it too.

Goldziher:
> > I dont follow. What do you want to do in starlite-multipart?
> 
> Here starlite-multipart https://github.com/starlite-api/starlite-multipart/blob/main/starlite_multipart/parser.py#L27 lowers headers.
> 
> I think it would be great to preserve case if possible (unless there is too much work needed for that)

not too much work at all - i added this yesterday. We can have it preserve case and then lower it on our end, or inside the parser component. Youre welcome to implement this.

jtraub:
@Goldziher sorry, I have no experience with Github actions so I don't really understand why ci/test fails during package installation.

Goldziher:
I'll fix it on main

PR #576, opened 2022-10-11
Add memcached cache backend

seladb:
# PR Checklist

### DIFF
--- a/starlite/cache/__init__.py
+++ b/starlite/cache/__init__.py
@@ -1,14 +1,8 @@
 from .base import Cache, CacheBackendProtocol
-from .memcached_cache_backend import MemcachedCacheBackend, MemcachedCacheBackendConfig
-from .redis_cache_backend import RedisCacheBackend, RedisCacheBackendConfig
 from .simple_cache_backend import SimpleCacheBackend
 
 __all__ = (
     "CacheBackendProtocol",
     "SimpleCacheBackend",
     "Cache",
-    "MemcachedCacheBackend",
-    "MemcachedCacheBackendConfig",
-    "RedisCacheBackend",
-    "RedisCacheBackendConfig",
 )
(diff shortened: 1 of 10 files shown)


starlite/cache/__init__.py
```diff
@@ -1,5 +1,8 @@
 from .base import Cache, CacheBackendProtocol
-from .redis_cache_backend import RedisCacheBackend, RedisCacheBackendConfig
 from .simple_cache_backend import SimpleCacheBackend
 
-__all__ = ("CacheBackendProtocol", "SimpleCacheBackend", "RedisCacheBackend", "RedisCacheBackendConfig", "Cache")
+__all__ = (
+    "CacheBackendProtocol",
+    "SimpleCacheBackend",
+    "Cache",
+)
```
seladb, line 8: The reason I removed redis backend is that both redis and memcached are optional dependencies and if they're not installed - importing anything from `starlite.cache.*` will fail

starlite/cache/memcached_cache_backend.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Notes:
+            - expiration is in seconds.
+            - return value is not used by Starlite internally.
+
+        Returns:
+            None
+        """
+
+        await self._memcached_client.set(key.encode(), pickle.dumps(value), exptime=expiration)
```
seladb, line 79: `pickle` is needed here because `aiomcache.Client.set()` requires both the key and value to be of type `bytes`. Maybe we should consider exposing the serialization method as part of `MemcachedCacheBackendConfig` and default it to `pickle`?

Goldziher: sure

Goldziher (approved):
looks great. 

you sure you dont want to become a maintainer @seladb ? 😉

starlite/cache/memcached_cache_backend.py
seladb: Done in https://github.com/starlite-api/starlite/pull/576/commits/2de4c81122f95a6a6e6d32a41bf91231684da008

PR #577, opened 2022-10-12
Add url_for function in templates

jtraub:
# PR Checklist

This PR does three things:
* Fixes reference docs for exceptions so they are actually exposed in the docs.
* Makes available `url_for` function in Jinja2 and Mako templates. This function is just a plain wrapper over `app.route_reverse`.
* Changes behaviour of `app.route_reverse` function so it raises `NoRouteMatchFoundException` exception in case route handler with given name does no exist or path parameters do not match.

My colleagues and I were playing with Starlite a lot recently and we noticed that returning None when there is no route handler with given name and raising ValidationErrorException if path parameters do not match makes user code a bit more messy having to check for None and handle exceptions at the same time. So it seems natural to raise exception in both cases as it is an exceptional situation anyway (we can not, say, generate HATEOS links to insert into response).

Other frameworks do this too. Check out [Starlette](https://github.com/encode/starlette/blob/master/starlette/routing.py#L24) and [Mansonite](https://github.com/MasoniteFramework/masonite/blob/539e3ad2ad6f46849bdee1972eb154e2cc1932ab/src/masonite/routes/Router.py#L62) for example. So I think this is a good change.


starlite/connection/base.py
```diff
@@ -285,21 +285,20 @@ def clear_session(self) -> None:
...
             name: The 'name' of the request route handler.
             **path_parameters: Values for path parameters in the route
 
         Raises:
-            ValidationException: If path parameters are missing in **path_parameters or have wrong type.
+            NoRouteMatchFoundException: If route with 'name' does not exist, path parameters are missing in **path_parameters or have wrong type.
 
         Returns:
-            If a route handler with the given name is found, it returns a string representing the absolute url of the route handler.
+            String representing the absolute url of the route handler.
```
Goldziher, line 299:
```suggestion
            A string representing the absolute url of the route handler.
```

starlite/exceptions/http_exceptions.py
```diff
@@ -116,6 +116,10 @@ class ServiceUnavailableException(HTTPException):
     status_code = HTTP_503_SERVICE_UNAVAILABLE
 
 
+class NoRouteMatchFoundException(InternalServerException):
+    """Route with given name and passed parameters can not be found."""
```
Goldziher, line 120:
```suggestion
    """A route with the given name could not be found."""
```

jtraub:
@Goldziher I have fixed typos and wording according to your review.

starlite/template/mako.py
```diff
@@ -11,11 +12,23 @@
...
-class MakoTemplateEngine(TemplateEngineProtocol["MakoTemplate"]):
+class MakoTemplate(TemplateProtocol):
+    template: "_MakoTemplate"
+
+    def __init__(self, template: "_MakoTemplate"):
+        super().__init__()
+        self.template = template
+
+    def render(self, *args: Any, **kwargs: Any) -> str:
+        kwargs["url_for"] = partial(url_for, kwargs)
```
Goldziher, line 27: shouldn't this be `partial(url_for, **kwargs)`?

Goldziher (approved):

PR #588, opened 2022-10-13
Fix `NoReturn` as allowed return typing for `delete` decorators

### LINKED ISSUE #585: Fix: remove  `NoReturn` as a supported return value for `delete`.
### Discussed in https://github.com/starlite-api/starlite/discussions/583

<sup>Originally posted by **Goldziher** October 13, 2022</sup>
We currenty support `NoReturn` as a valid return value for the `delete` method, which cannot support returning a value with uvicorn (which freaks out if it does return a value, while hypercorn and other servers do not). 

We should decide how we want to proceed given that in the official PEP for this type it is determined that it should be used only for special functions that never return a value, like endless loops or functions that raise exceptions: https://peps.python.org/pep-0484/#the-noreturn-type. </div>

Goldziher:
# PR Checklist

This PR resolves #585

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ While supporting function based route handlers, Starlite also supports and promo
 controllers:
 
 ```python title="my_app/controllers/user.py"
-from typing import List, Optional, NoReturn
+from typing import List, Optional
 
 from pydantic import UUID4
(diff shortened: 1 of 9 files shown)


cofin (approved):
Looks good to me.

PR #590, opened 2022-10-14
Add url_for_static_asset to app and templates

jtraub:
# PR Checklist

This PR adds:
* `name` kwarg to `StaticFilesConfig` which has the same behavior as `name` kwarg in route handlers;
* `app.asset_url_path` to generate url paths to files under given `StaticFiles`.
* `request.asset_url` to generate absolute urls to files.
* `asset_url_path` function for templates

Sample piece of code

```python
from starlite import Starlite, StaticFilesConfig

app = Starlite(
    route_handlers=[...],
    static_files_config=[
        StaticFilesConfig(
            directories=["static"], path="/complex/static/path", name="static"
        ),
    ],
)

url_path = app.asset_url_path("static", "file.pdf")
# /complex/static/path/file.pdf
```

First I was thinking about extending `app.route_reverse` function since its internal behavior is similar but decided to create a separate function because `route_reverse` interface would have to be extended making it harder to use in templates. Another reason is that I think generating static URLs later should be delegated to `StaticFiles` and its descendants - this way we can easily work with cloud storage like S3 to store large files and make them accessible only to selected users via API (quite cumbersome to implement with nginx).

At the moment implementation is very basic. It does limited checks and simply concatenates static path and passed file path. More complicated checks e.g. file existence can be implemented in `StaticFiles` descendants later; right now we can just establish interfaces in functions facing the user code.

I am not sure about the name. Maybe `asset_url_path` can be changed to something more short and concise?

### DIFF
--- a/docs/usage/0-the-starlite-app/3-static-files.md
+++ b/docs/usage/0-the-starlite-app/3-static-files.md
@@ -1,12 +1,12 @@
 # Static Files
 
 Static files are served by the app from predefined locations. To configure static file serving, either pass an
-instance of `starlite.config.StaticFilesConfig` or a list thereof to the Starlite constructor using
-the `static_files_config` kwarg.
+instance of [`starlite.config.StaticFilesConfig`][starlite.config.static_files.StaticFilesConfig] or a list
+thereof to the Starlite constructor using the `static_files_config` kwarg.
 
-For example, lets say our Starlite app is going to serve **regular files** from the "my_app/static" folder and **html
-documents** from the "my_app/html" folder, and we would like to serve the **static files** on the "/files" path,
-and the **html files** on the "/html" path:
+For example, lets say our Starlite app is going to serve **regular files** from the `my_app/static` folder and **html
+documents** from the `my_app/html` folder, and we would like to serve the **static files** on the `/files` path,
+and the **html files** on the `/html` path:
 
 ```python
 from starlite import Starlite, StaticFilesConfig
@@ -27,3 +27,23 @@ be sent over, otherwise a **404 response** will be sent.
 If `html_mode` is enabled and no specific file is requested, the application will fall back to serving `index.html`. If
 no file is found the application will look for a `404.html` file in order to render a response, otherwise a 404
 `NotFoundException` will be returned.
+
+You can provide `name` parameter to `StaticFilesConfig` to identify given config and generate links to files in folders
+belonging to that config. `name` should be a unique string across all static configs and
+[route handlers](2-route-handlers/4-route-handler-indexing.md)
+
+```python
+from starlite import Starlite, StaticFilesConfig
+
+app = Starlite(
+    route_handlers=[...],
+    static_files_config=[
+        StaticFilesConfig(
+            directories=["static"], path="/some_folder/static/path", name="static"
+        ),
+    ],
+)
+
+url_path = app.url_for_static_asset("static", "file.pdf")
+# /some_folder/static/path/file.pdf
+```
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type, Union, c
 from starlette.middleware import Middleware as StarletteMiddleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.middleware.trustedhost import TrustedHostMiddleware
+from starlette.staticfiles import StaticFiles
 from typing_extensions import TypedDict
 
 from starlite.asgi import (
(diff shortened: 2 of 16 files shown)


docs/usage/0-the-starlite-app/3-static-files.md
```diff
@@ -1,12 +1,12 @@
 # Static Files
 
 Static files are served by the app from predefined locations. To configure static file serving, either pass an
-instance of `starlite.config.StaticFilesConfig` or a list thereof to the Starlite constructor using
-the `static_files_config` kwarg.
+instance of [`starlite.config.StaticFilesConfig`][starlite.config.static_files.StaticFilesConfig] or a list
```
Goldziher, line 4:
```suggestion
instance of [`starlite.config.StaticFilesConfig`][starlite.config.static_files.StaticFilesConfig], or a list
```

docs/usage/0-the-starlite-app/3-static-files.md
```diff
@@ -27,3 +27,23 @@ be sent over, otherwise a **404 response** will be sent.
 If `html_mode` is enabled and no specific file is requested, the application will fall back to serving `index.html`. If
 no file is found the application will look for a `404.html` file in order to render a response, otherwise a 404
 `NotFoundException` will be returned.
+
+You can provide `name` parameter to `StaticFilesConfig` to identify given config and generate links to files in folders
+belonging to that config. `name` should be an unique string across all static configs and
```
Goldziher, line 32:
```suggestion
belonging to that config. `name` should be a unique string across all static configs and
```

docs/usage/0-the-starlite-app/3-static-files.md
```diff
@@ -27,3 +27,23 @@ be sent over, otherwise a **404 response** will be sent.
...
+[route handlers](2-route-handlers/4-route-handler-indexing.md)
+
+```python
+from starlite import Starlite, StaticFilesConfig
+
+app = Starlite(
+    route_handlers=[...],
+    static_files_config=[
+        StaticFilesConfig(
+            directories=["static"], path="/complex/static/path", name="static"
```
Goldziher, line 42: I'd suggest to change from `/complex/static/path` to:

```suggestion
            directories=["static"], path="/some_folder/static/path", name="static"
```

docs/usage/15-templating.md
```diff
@@ -57,6 +57,9 @@ URLs for indexed route handlers in templates can be generated with the `url_for`
 matches [`route_reverse`][starlite.app.Starlite.route_reverse] behaviour. More details about route handler indexing
 can be found [here](2-route-handlers/4-route-handler-indexing.md)
 
+URLs for static files can be created with the `asset_url_path` function. Signature and behaviour is identical to
```
Goldziher, line 60:
```suggestion
URLs for static files can be created using the `asset_url_path` function. It's signature and behaviour are identical to
```

Also consider the name - better to be clear about the use, so I would go with `create_url_for_static_asset`

starlite/app.py
```diff
@@ -6,6 +6,7 @@
 from starlette.middleware import Middleware as StarletteMiddleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.middleware.trustedhost import TrustedHostMiddleware
+from starlette.staticfiles import StaticFiles
```
Goldziher, line 9: what is this import used for?

starlite/app.py
```diff
@@ -540,6 +541,42 @@ def get_membership_details(group_id: int, user_id: int) -> None:
 
         return join_paths(output)
 
+    def asset_url_path(self, name: str, file_path: str) -> str:
```
Goldziher, line 544: as suggested above- `create_url_for_static_asset` or simply `url_for_static_asset` if you want a somewhat shorter version

starlite/template/base.py
```diff
@@ -25,6 +25,24 @@ def url_for(context: Mapping[str, Any], route_name: str, **path_parameters: Any)
     return request.app.route_reverse(route_name, **path_parameters)
 
 
+def asset_url_path(context: Mapping[str, Any], name: str, file_path: str) -> str:
```
Goldziher, line 28: same as in other places

Goldziher (approved):
cool stuff

starlite/app.py
jtraub: It is used in the function body for the following check
 
 ```python
 if not isinstance(handler_fn, StaticFiles):
            raise NoRouteMatchFoundException(f"Handler with name {name} is not a static files handler")
 ```
 
The check has been introduced because I want to move actual URL generation into StaticFiles subclass to cover various cases (I am actually already doing that in my app) so since we use shared handlers index some sort of check is really needed.

Goldziher: ok

peterschutt (approved):
Nice work!

PR #598, opened 2022-10-14
Add `csrf_token` function to templates

Goldziher:
# PR Checklist


starlite/template/base.py
```diff
@@ -21,8 +28,25 @@ def url_for(context: Mapping[str, Any], route_name: str, **path_parameters: Any)
...
+
+    Args:
+        context: The template context.
+
+
+    Returns:
+        A CSRF token if the app level `csrf_config` is set, otherwise an empty string.
+    """
+    csrf_config = context["request"].app.csrf_config
+    return generate_csrf_token(csrf_config.secret) if csrf_config else ""
```
jtraub, line 49: I might be wrong here but it looks like this helper should read token from cookies.

Because otherwise token in cookies and token in templates (hidden field) will be different

jtraub:
I just realized that the middleware doesn't extract token from POST values. Or I might be missing it in the code..

Goldziher:
> I just realized that the middleware doesn't extract token from POST values. Or I might be missing it in the code..

not by default. This is configurable:

```python
      if request.method in self.config.safe_methods:
            token = scope["_csrf_token"] = generate_csrf_token(secret=self.config.secret)  # type: ignore
            await self.app(scope, receive, self.create_send_wrapper(send=send, csrf_cookie=csrf_cookie, token=token))
```

And the default methods are:

```python
 {"GET", "HEAD"}
```

@seladb engineered it like this, following the OWASP guidelines: https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

jtraub:
> not by default. This is configurable

I am talking about checking CSRF token stored in a hidden input on form submission like Django does.

Goldziher:
> > not by default. This is configurable
> 
> I am talking about checking CSRF token stored in a hidden input on form submission like Django does.

This is not handled yet. I'm send the input now using the `csrf_input` context value.

I will see about the next step. Can you link to the Django checks you mention?

jtraub:
Sure

https://github.com/django/django/blob/main/django/middleware/csrf.py#L369

And here some random CSRF middleware which does the same thing

https://github.com/gnat/csrf-starlette-fastapi/blob/main/csrf_middleware.py#L27

jtraub (approved):

peterschutt:
[Flask-WTF](https://flask-wtf.readthedocs.io/en/1.0.x/csrf/) does this by default too, just need to render the field in the form.

starlite/middleware/csrf.py
```diff
@@ -50,18 +52,28 @@ async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> No
...
+        if not existing_csrf_token and content_type in {
+            RequestEncodingType.URL_ENCODED,
+            RequestEncodingType.MULTI_PART,
+        }:
+            form = await request.form()
+            existing_csrf_token = form.get("_csrf_token", None)
+
         if request.method in self.config.safe_methods:
-            await self.app(scope, receive, self.create_send_wrapper(send=send, csrf_cookie=csrf_cookie))
+            token = scope["_csrf_token"] = generate_csrf_token(secret=self.config.secret)  # type: ignore
```
jtraub, line 68: I just got up and understood what has been bothering me with this code.

`csrf_token` will contain new token on each safe request.

As far as I understand TestClient persists cookies between requests so an easy check would be modifying `test_csrf_form_parsing` test to perform two get requests in a row before parsing response with BS

```python
_ = client.get("/")
response = client.get("/")
```

Goldziher: true. What would be the solution in your opinion?

jtraub: Do something like this

```python
csrf_cookie_exists = csrf_cookie is None
if not(csrf_cookie_exists):
    token = scope["_csrf_token"] = csrf_cookie
else:
    token = scope["_csrf_token"] = generate_csrf_token(secret=self.config.secret)  # type: ignore
await self.app(scope, receive, self.create_send_wrapper(send=send, set_cookie=csrf_cookie_exists, token=token))
.....

def create_send_wrapper(self, send: "Send", token: str, set_cookie: bool) -> "Send":
....

    async def send_wrapper(message: "Message") -> None:
    ....
        if set_cookie and message["type"] == "http.response.start":
            ...
    return send_wrapper

```

jtraub: Sorry, I did this quickly out of my head so the code contains too many `not`. Should rename the variable to `csrf_cookie_doesnot_exists` and negate it so we can omit `not`

PR #601, opened 2022-10-15
Support cache control headers

seladb:
# PR Checklist


starlite/handlers/http.py
```diff
@@ -424,10 +424,10 @@ def resolve_response_headers(self) -> "ResponseHeadersMap":
         resolved_response_headers = {}
         for layer in self.ownership_layers:
             resolved_response_headers.update(layer.response_headers or {})
-        if self.cache_control:
-            resolved_response_headers.update(
-                {CacheControlHeader.CACHE_CONTROL_HEADER: ResponseHeader(value=self.cache_control.to_header())}
-            )
+            if cache_control := getattr(layer, "cache_control", None):
```
provinzkraut, line 427: Probably should be 

```python
cache_control = getattr(layer, "cache_control", None)
if cache_control:
   ...
```

To support Python < 3.8

seladb: Thanks, fixed!

starlite/datastructures/headers.py
```diff
@@ -0,0 +1,118 @@
...
+
+from starlite.exceptions import ImproperlyConfiguredException
+
+if TYPE_CHECKING:
+    from typing import Any, Dict
+
+
+class Header(BaseModel, ABC):
+    """An abstract type for HTTP headers."""
+
```
provinzkraut, line 14: The `documentation_only` field is missing. You mention it in the documentation, so it should be added here as well.

tests/response/test_response_headers.py
```diff
@@ -72,3 +73,27 @@ def my_handler(data: Dict[str, str]) -> Dict[str, str]:
         response = client.post("/test", json={"hello": "world"})
         assert response.status_code == HTTP_201_CREATED
         assert response.headers.get("test-header") == "test value"
+
+
```
provinzkraut, line 77: This should probably include tests for overwrite behaviour between `response_headers=` and `cache_control=`. It's mentioned in the docs, so accidentally changing the overwrite cascade in some way could break things if people rely on the specific order.

tests/datastructures/test_headers.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_cache_control_to_header() -> None:
+    header = CacheControlHeader(max_age=10, private=True)
+    expected_header_values = ["max-age=10, private", "private, max-age=10"]
+    assert header.to_header() in expected_header_values
+    assert header.to_header(include_header_name=True) in [f"cache-control: {v}" for v in expected_header_values]
+
+
+def test_cache_control_from_header() -> None:
```
provinzkraut, line 14: Maybe test all possible fields here to make sure they're included / of the right type after parsing

provinzkraut (changes requested):
Looks mostly good, I just left a few minor notes regarding documentation and tests.

The `documentation_only` field is missing though, which should definitely be added.

Goldziher (approved):
Great stuff. I have no comments on top of what @provinzkraut already wrote. Merge when ready.

docs/usage/5-responses/4-response-headers.md
```diff
@@ -228,3 +228,56 @@ router = Router(
...
+
+@get("/population")
+def get_population_count() -> int:
+    return 100000
+
+
+router = Router(
+    route_handlers=[MyController, get_population_count],
+    cache_control=CacheControlHeader(max_age=2_628_288, public=True),
+)
```
peterschutt, line 273: As this looks like a complete example, please have a look at #622 for how we prefer to include complete examples in the docs.

You can write the example as a script, include a test for it, and then cross reference the script in the docs.

The example tests are run in CI so this ensures that our documentation examples don't go stale.

starlite/datastructures/headers.py
seladb: I don't think `documentation_only` is needed in these "extra" headers. Do you see a use case where someone needs to add them only for documentation (OpenAPI) purposes?

Where do I mention it in the documentation?

peterschutt (approved):
Nice! Please see my review about how to include complete examples in the documentation.

starlite/datastructures/headers.py
Goldziher: This is meant for the cases where users set headers dynamically. For example, when returning a response instance they create.

seladb: I'm not sure I understand... 

When users want to use `CacheControlHeader` they'll likely do something like this:

```python
@get(path="/handler", cache_control=CacheControlHeader(max_age=1000))
def my_handler(self) -> None:
    pass
```

Behind the scenes, [we create a `ResponseHeader` instance](https://github.com/seladb/starlite/blob/da0ef9943b2521564772aa7bc4d41ca8e8b90051/starlite/handlers/http.py#L433) with a default value of `documentation_only=False` and add it to the list of `resolved_response_headers`.

In this scenario, what's the use case for creating a `CacheControlHeader(..., documentation_only=True)`?

In other words, why would users want to create a `CacheControlHeader` header if they don't want to return it in the response?

starlite/controller.py
```diff
@@ -64,6 +65,11 @@ class Controller:
         It receives the [Request][starlite.connection.Request] instance and any
         non-`None` return value is used for the response, bypassing the route handler.
     """
+    cache_control: Optional["CacheControlHeader"]
+    """
+         A [CacheControlHeader][starlite.datastructures.CacheControlHeader] header to add to route handlers
+         of this controller.
+    """
```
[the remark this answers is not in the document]
seladb: Fixed here: https://github.com/starlite-api/starlite/pull/601/commits/7432876afd1d81e445a423ade5f7c9e326750491

starlite/config/app.py
```diff
@@ -103,6 +103,11 @@ class Config(BaseConfig):
     """
     Configures caching behavior of the application.
     """
+    cache_control: Optional[CacheControlHeader]
+    """
+    A `cache-control` header of type [CacheControlHeader][starlite.datastructures.CacheControlHeader] to add to
+    route handlers of this app.
+    """
```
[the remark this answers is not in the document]
seladb: Fixed here: https://github.com/starlite-api/starlite/pull/601/commits/7432876afd1d81e445a423ade5f7c9e326750491

starlite/datastructures/headers.py
provinzkraut: They might want to return it in the response, but the view could be dependent on the response itself, so they couldn’t be set at the time when they declare the handler.

In this case, having `documentation_only=True` means they can set that when declaring the handler, but set the "real header" dynamically within the function body.

> Where do I mention it in the documentation?

Disregard that please. You didn’t actually mention it in the docs, I just mixed up some lines there while reading.

seladb: Sure, added here: https://github.com/starlite-api/starlite/pull/601/commits/2089e9002542eee8d171517424d80bd77999a1fe

tests/response/test_response_headers.py
seladb: Sure, the test was added here: https://github.com/starlite-api/starlite/pull/601/commits/56a8814ae9e7b1a30bc34b00d4236cbc82734fad

tests/datastructures/test_headers.py
seladb: I'm not sure there's a reason to test all possible fields because there's no per-field logic. The logic is per field type - either "boolean" fields (like `public`, `no-store`, etc.) or "integer" fields (like `max-age`). This test tests both field types

provinzkraut: It would serve more to codify the field names, so changing their names / types would fail tests if not done in a backwards compatible way.

provinzkraut (commented):
Looks good! Nothing more to add except the remaining open comment, but we can go either way about this.

docs/usage/5-responses/4-response-headers.md
seladb: Thanks! Fixed here: https://github.com/starlite-api/starlite/pull/601/commits/7dc701b993ca6f097558cf2df6f32584ae6128da

provinzkraut (approved):

tests/datastructures/test_headers.py
seladb: Sure, fixed here: https://github.com/starlite-api/starlite/pull/601/commits/0f7dd0639e3401d984f94fc942036d191ce9aae7

PR #604, opened 2022-10-16
Native support for `TypedDict`.

### LINKED ISSUE #591: Feature: Support TypedDict for request body
Currently starlite supports pydantic BaseModel and dataclass for body input. I would also like support for TypedDict.

**Additional context**
I was trying out the [python prisma client](https://github.com/RobertCraigie/prisma-client-py). It generates lots of useful classes, for example, TaskCreateInput, TaskUpdateInput, etc. as typed dicts. I tried using these to type the response body.
```python
class TaskController(Controller):
    path: str = "/api/task"

    @post()
    async def create(self, data: TaskCreateInput) -> Task:
        task = await db.task.create(data=data)
        return task

```

but it is not shown in the docs
image

P.S. It works in FastAPI.

peterschutt:
Allow using instances of `TypedDict` as a type for `data` kwarg.

Support openapi docs, and `Partial`.

Closes #591 

# PR Checklist


docs/usage/11-data-transfer-objects.md
```diff
@@ -2,7 +2,7 @@
 <!-- markdownlint-disable MD052 -->
 
 Starlite includes a [`DTOFactory`][starlite.dto.DTOFactory] class that allows you to create DTOs from pydantic models,
-dataclasses and any other class supported via plugins.
+dataclasses, typeddict, and any other class supported via plugins.
```
peterschutt, line 5:
```suggestion
dataclasses, `TypedDict`, and any other class supported via plugins.
```

starlite/utils/model.py
```diff
@@ -20,6 +21,7 @@ def create_parsed_model_field(value: Type[Any]) -> "ModelField":
 
 
 _dataclass_model_map: Dict[Any, Type[BaseModel]] = {}
+_typeddict_model_map: Dict[Any, Type[BaseModel]] = {}
```
peterschutt, line 24: We don't really need type specific maps here, do we?

Goldziher: Not really

peterschutt:
I don't think `DTO.from_model_instance()` can be supported for `TypedDict` - correct me if I'm wrong - there is no typing info retained when a `dict` is constructed from a `TypedDict`.

It doesn't look like we actually use that method internally for anything though:

image

starlite/utils/predicates.py
```diff
@@ -66,4 +77,40 @@ def is_optional_union(annotation: Any) -> bool:
...
+        value: tested to determine if instance or type of `dataclass`.
+
+    Returns:
+        `True` if instance or type of `dataclass`.
+    """
+    return is_dataclass(value)
+
+
+def is_typeddict_typeguard(value: Any) -> "TypeGuard[TypedDictType]":
+    """Wrapper for `is_dataclass()` that narrows type.
```
peterschutt, line 108:
```suggestion
    """Wrapper for `is_typeddict()` that narrows type.
```

docs/index.md
```diff
@@ -81,7 +81,7 @@ class User(BaseModel):
     id: UUID4
 ```
 
-Alternatively, you can **use a dataclass** – either from dataclasses or from pydantic:
+Alternatively, you can **use a dataclass** – either from dataclasses or from pydantic, or a `TypedDict`:
```
peterschutt, line 84:
```suggestion
Alternatively, you can **use a dataclass** – either from dataclasses or from pydantic, or a [`TypedDict`][typing.TypedDict]:
```
From memory this gets cross refs to the stdlib docs - will do for the rest too.

starlite/types/builtin_types.py
```diff
@@ -0,0 +1,27 @@
...
+]
+
+DataclassType: "TypeAlias" = "Type[DataclassProtocol]"
+
+DataclassTypeOrInstance: "TypeAlias" = "Union[DataclassType, DataclassProtocol]"
+
+NoneType = type(None)
+
+# mypy issue: https://github.com/python/mypy/issues/11030
+TypedDictType: "TypeAlias" = "Type[TypedDict]"  # type:ignore[valid-type]
```
peterschutt, line 27: I don't think we really want to make these public types, so I haven't exported them into the top level `starlite.types`, nor added to docs.

The typing of `TypedDictType` is the same as what pydantic does.

Added the `NoneType` alias as there were quite a few `type(None)` calls throughout lib and tests.

The dataclass types allowed for better typing of `utils.models` and `DTO` related stuff.

Also, using `TypeAlias` here is probably the right way:tm: to structure modules of aliases like this without the runtime imports rather than the:

```python
if TYPE_CHECKING:
    from somewhere import something
else:
    something = Any
```
...pattern that we have elsewhere in `starlite.types`.

starlite/utils/predicates.py
```diff
@@ -66,4 +77,40 @@ def is_optional_union(annotation: Any) -> bool:
...
+def is_typeddict_typeguard(value: Any) -> "TypeGuard[TypedDictType]":
+    """Wrapper for `is_typeddict()` that narrows type.
+
+    Args:
+        value: tested to determine if instance or type of `dataclass`.
+
+    Returns:
+        `True` if instance or type of `dataclass`.
+    """
+    return is_typeddict(value)
```
peterschutt, line 116: Thoughts on these predicates and naming?

Goldziher (approved):

starlite/types/builtin_types.py
Goldziher: Indeed. 

We can inject globals there, but it's messy

starlite/types/partial.py
```diff
@@ -1,102 +1,140 @@
...
+            item: A `TypedDict` class.
+        """
+        type_hints: Dict[str, Any] = {}
+        for key_name, value_type in get_type_hints(item).items():
+            if NoneType in get_args(value_type):
+                type_hints[key_name] = value_type
+                continue
+            type_hints[key_name] = Optional[value_type]
+        type_name = cls._create_partial_type_name(item)
+        cls._models[item] = TypedDict(type_name, type_hints, total=False)  # type:ignore
```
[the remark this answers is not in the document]
Goldziher: Np

starlite/utils/predicates.py
```diff
@@ -66,4 +77,40 @@ def is_optional_union(annotation: Any) -> bool:
...
+    Args:
+        value: tested to determine if type of `dataclass`.
+
+    Returns:
+        `True` if `value` is a `dataclass` type.
+    """
+    return is_dataclass(value) and isinstance(value, type)
+
+
+def is_dataclass_type_or_instance_typeguard(value: Any) -> "TypeGuard[DataclassTypeOrInstance]":
```
Goldziher, line 95: We could instead have `is_dataclass_instance_typeguard` and `is_dataclass_class_typeguard`

Goldziher (approved):
Really nice

starlite/utils/predicates.py
peterschutt: yep went with "class" instead of "type"

PR #626, opened 2022-10-19
Enhancement: Replace Starlette Responses

Goldziher:
# PR Checklist

This PR removes the usage of Starlette based `Response` objects by introducing our own version - with a compatible API. 

Relates to #612


.pre-commit-config.yaml
```diff
@@ -70,12 +70,10 @@ repos:
             "flake8-print",
             "flake8-simplify",
             "flake8-type-checking",
-            "flake8-pytest-style",
             "flake8-implicit-str-concat",
             "flake8-noqa",
             "flake8-return",
             "flake8-secure-coding-standard",
-            "flake8-eradicate",
```
Goldziher, line 78: this plugin was giving false positives on comments

.pre-commit-config.yaml
```diff
@@ -70,12 +70,10 @@ repos:
             "flake8-print",
             "flake8-simplify",
             "flake8-type-checking",
-            "flake8-pytest-style",
```
Goldziher, line 73: this plugin was just noise IMO

Goldziher:
PR is still missing docs and some polishing touches, but it is substantially done and the tests are in place - so its a good time to start reviewing. Please do!

peterschutt (commented):
Haven't made all the way through yet!

starlite/datastructures/response_containers.py
```diff
@@ -237,6 +236,9 @@ def to_response(
         Returns:
             A TemplateResponse instance
         """
+        from starlite.response import (  # pylint: disable=import-outside-toplevel
+            TemplateResponse,
+        )
```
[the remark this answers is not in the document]
Goldziher: apparently its resolved!

infohash:
So with this PR, instead of using starlette `RedirectResponse` in the middleware, I will be able to use `RedirectResponse` from starlite?

Goldziher:
> So with this PR, instead of using starlette `RedirectResponse` in the middleware, I will be able to use `RedirectResponse` from starlite?

that is correct

starlite/response/base.py
```diff
@@ -0,0 +1,327 @@
...
+        Raises:
+            TypeError: if value is not supported
+        """
+        return default_serializer(value)
+
+    def render(self, content: Any) -> bytes:
+        """
+        Handles the rendering of content T into a bytes string.
+        Args:
+            content: An arbitrary value of type T
```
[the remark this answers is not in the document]
Goldziher: fixed

provinzkraut:
[In docs/usage/5-responses/2-status-codes.md](https://github.com/starlite-api/starlite/blob/replace-response/docs/usage/5-responses/2-status-codes.md?plain=1#L40) this line should be:

```diff
- tests). Starlette includes easy to use statuses that are exported from `starlette.status`, e.g. `HTTP_200_OK`
+ tests). Starleite includes easy to use statuses that are exported from `starlite.status`, e.g. `HTTP_200_OK`
```

Goldziher:
> [In docs/usage/5-responses/2-status-codes.md](https://github.com/starlite-api/starlite/blob/replace-response/docs/usage/5-responses/2-status-codes.md?plain=1#L40) this line should be:
> 
> ```diff
> - tests). Starlette includes easy to use statuses that are exported from `starlette.status`, e.g. `HTTP_200_OK`
> + tests). Starleite includes easy to use statuses that are exported from `starlite.status_codes`, e.g. `HTTP_200_OK`
> ```

fixed

provinzkraut (approved):

starlite/datastructures/response_containers.py
```diff
@@ -24,21 +24,24 @@
...
+    StreamingResponse,
+    TemplateResponse,
+)
 
 if TYPE_CHECKING:
     from starlite.app import Starlite
     from starlite.connection import Request
 
-R = TypeVar("R", bound=StarletteResponse)
+R = TypeVar("R")
```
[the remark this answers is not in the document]
Goldziher: I agree

starlite/datastructures/response_containers.py
```diff
@@ -57,6 +59,8 @@ class Config(BaseConfig):
     """A list of Cookie instances to be set under the response 'Set-Cookie' header. Defaults to None."""
     media_type: Optional[Union[MediaType, str]] = None
     """If defined, overrides the media type configured in the route decorator"""
+    encoding: str = "utf-8"
+    """The encoding to used for the response headers."""
```
jtraub, line 63: Probably this should be `The encoding to be used for the response headers`

starlite/datastructures/response_containers.py
```diff
@@ -88,22 +92,36 @@ class File(ResponseContainer[FileResponse]):
...
+    filename: Optional[str] = None
     """The filename"""
     stat_result: Optional[os.stat_result] = None
     """File statistics"""
+    chunk_size: int = ONE_MEGA_BYTE
+    """The size of chunks to use when streaming the file"""
+    content_disposition_type: "Literal['attachment', 'inline']" = "attachment"
+    """The type of the 'Content-Disposition'. Either 'inline' or 'attachment'."""
+    etag: Optional[str] = None
+    """An optional etag for the file. If not provided, and etag will be generated automatically."""
```
jtraub, line 104: Another typo? `If not provided, an etag`

starlite/datastructures/response_containers.py
Goldziher: fixed

starlite/datastructures/response_containers.py
Goldziher: fixed

starlite/datastructures/response_containers.py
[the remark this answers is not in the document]
Goldziher: I bound it to `Response` - before i had some cyclical import issues, but thats fixed now.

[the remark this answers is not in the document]
Goldziher: Ok, I remembered why I did this. It shouldn't be a protocol - it should be `Response`, but mypy is going ape shit for some reason. I have no clue why to be honest, i tried fixing it but its some issue with GenericModel + Generic and mypy in this setup. Anyhow, for now I will keep it without binding, because we are using ResponseContainer generics internally.

starlite/datastructures/cookie.py
```diff
@@ -51,3 +51,18 @@ def to_header(self, **kwargs: Any) -> str:
             if value is not None:
                 simple_cookie[self.key][key] = value
         return simple_cookie.output(**kwargs).strip()
+
+    def __eq__(self, other: Any) -> bool:
```
provinzkraut, line 55: I'm actually not sure if this is a good approach. It doesn't really do a strict equality check, since it does not compare for value and other attributes. Maybe this shouldn't be a dunder method, but a separate `is_equal` method with a `strict` flag defaulting to `True`. We could still implement `__eq__` as a callback to  `self.is_equal(strict=True)` then.

I'd also do the canonical "return `NotImplemented`" if the types don't match

starlite/response/base.py
```diff
@@ -0,0 +1,328 @@
...
+        *,
+        status_code: int = HTTP_200_OK,
+        media_type: Union[MediaType, "OpenAPIMediaType", str] = MediaType.JSON,
+        background: Optional[Union["BackgroundTask", "BackgroundTasks"]] = None,
+        headers: Optional[Dict[str, Any]] = None,
+        cookies: Optional["ResponseCookies"] = None,
+        encoding: str = "utf-8",
+        is_head_response: bool = False,
+    ) -> None:
+        """This is the base Starlite HTTP response class, its is used as a
```
provinzkraut, line 70: Typo: "its is used" > "it is used"
Typo: "Used as a basis" > "Used as the basis"

tests/response/test_base_response.py
```diff
@@ -0,0 +1,178 @@
...
+    @get("/delete")
+    def delete_cookie_handler() -> Response:
+        response = Response(content=None)
+        response.delete_cookie(
+            "test",
+            "abc",
+        )
+        assert len(response.cookies) == 1
+        return response
+
```
provinzkraut, line 82: Since values don't matter for cookie deletion, maybe add a test for that case. I.e. that a cookie is deleted if its key matches but its value does not.

tests/response/test_base_response.py
```diff
@@ -0,0 +1,178 @@
...
+        return Response(content=body, status_code=status)
+
+    with create_test_client(handler) as client:
+        response = client.get("/")
+        if should_raise:
+            assert response.status_code == HTTP_500_INTERNAL_SERVER_ERROR
+        else:
+            assert response.status_code == status
+            assert "content-length" not in response.headers
+
```
provinzkraut, line 139: Maybe add a test for the case that a handler **should** return `None`, but doesn't (and the other way around) when no explicit status code is set (i.e a handler is annotated as `- > None`  without a return code)

tests/response/test_file_response.py
```diff
@@ -0,0 +1,107 @@
...
+    assert filled_by_bg_task == ""
+    client = TestClient(app)
+    response = client.get("/")
+    expected_disposition = 'attachment; filename="example.png"'
+    assert response.status_code == HTTP_200_OK
+    assert response.content == content
+    assert response.headers["content-type"] == "image/png"
+    assert response.headers["content-disposition"] == expected_disposition
+    assert "content-length" in response.headers
+    assert "last-modified" in response.headers
```
provinzkraut, line 56: Probably best to test the correctness of these values

tests/response/test_streaming_response.py
```diff
@@ -0,0 +1,142 @@
...
+            yield b"chunk "
+
+    response = StreamingResponse(content=stream_indefinitely())
+
+    with anyio.move_on_after(1) as cancel_scope:
+        await response({}, receive_disconnect, send)  # type: ignore
+    assert not cancel_scope.cancel_called, "Content streaming should stop itself."
+
+
+def test_streaming_response() -> None:
```
provinzkraut, line 62: This should be tested with a synchronous generator as well 
(also applies to the following tests)

provinzkraut (changes requested):

starlite/response/base.py
```diff
@@ -0,0 +1,327 @@
...
+        Args:
+            key: Header key.
+            value: Header value.
+
+        Returns:
+            None.
+        """
+        self.headers[key] = value
+
+    def set_etag(self, etag: str) -> None:
```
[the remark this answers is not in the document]
Goldziher: ok

starlite/response/base.py
```diff
@@ -0,0 +1,327 @@
...
+        return b""
+
+    @property
+    def content_length(self) -> Optional[int]:
+        """
+
+        Returns:
+            Returns the value for the 'Content-Length' header, if applies.
+        """
+        if self.status_allows_body and isinstance(self.body, bytes):
```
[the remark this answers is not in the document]
Goldziher: fair point

starlite/response/base.py
```diff
@@ -0,0 +1,327 @@
...
+            return len(self.body)
+        return None
+
+    @property
+    def encoded_headers(self) -> List[Tuple[bytes, bytes]]:
+        """
```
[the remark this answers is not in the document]
Goldziher: done

starlite/response/file.py
```diff
@@ -0,0 +1,167 @@
...
+        """Creates an etag.
+
+        Notes:
+            - Function is derived from flask.
+
+        Returns:
+            An etag.
+        """
+        check = adler32(str(path).encode("utf-8")) & 0xFFFFFFFF
+        return f"{self.stat_result.st_mtime}-{self.stat_result.st_size}-{check}"
```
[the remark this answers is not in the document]
Goldziher: done

starlite/response/redirect.py
```diff
@@ -0,0 +1,53 @@
...
+            status_code: An HTTP status code. The status code should be one of 301, 302, 303, 307 or 308,
+                otherwise an exception will be raised. .
+            headers: A string keyed dictionary of response headers. Header keys are insensitive.
+            cookies: A list of [Cookie][starlite.datastructures.Cookie] instances to be set under the response 'Set-Cookie' header.
+            encoding: The encoding to used for the response headers.
+        Raises:
+            [ImproperlyConfiguredException][starlite.exceptions.ImproperlyConfiguredException]: If status code is not a redirect status code.
+        """
+        if status_code not in REDIRECT_STATUS_CODES:
+            raise ImproperlyConfiguredException(
```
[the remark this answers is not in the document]
Goldziher: Done

starlite/response/base.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Args:
+            key: Key of the cookie.
+            path: Path of the cookie.
+            domain: Domain of the cookie.
+
+        Returns:
+            None.
+        """
+        self.cookies = [
```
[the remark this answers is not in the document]
Goldziher: nice

starlite/datastructures/cookie.py
Goldziher: cookie equality is based on these parameters in browsers, thats the reason. Why wouldnt this be a good idea in your opinion?

starlite/response/base.py
Goldziher: fixed

tests/response/test_streaming_response.py
Goldziher: there are tests with sync generators elsewhere that use this, but i will add

tests/response/test_base_response.py
Goldziher: can do. The point really is what makes the browser delete the cookie, which is not the value but rather the key+path+domain combination.

tests/response/test_base_response.py
Goldziher: aight

tests/response/test_file_response.py
Goldziher: ok

Goldziher: done

starlite/datastructures/cookie.py
provinzkraut: It makes sense in the HTTP world, but less so in the python world. In the python world it might be expected that two cookies are only equal if their public attributes all equal.

You have a point though, for this specific implementation it makes sense to check equality like this. Probably should be mentioned in the docs somewhere then.

tests/response/test_base_response.py
provinzkraut: Correct. I just think we should have a test that checks that we are doing it on our side as well. What I mean is:

Test that `response.delete_cookie("test, "abc")` should deletes any cookie that matches these parameters, but ignore things like its value.

tests/response/test_streaming_response.py
provinzkraut: Must have missed them! Maybe move them here then because that's where you'd expect them (at least I did)

provinzkraut (approved):

PR #629, opened 2022-10-20
Fix #628

### LINKED ISSUE #628: Bug: Cookies with the strings "session" in their name interfere with the session middleware
**Describe the bug**
If a request contains the a cookie with the name containing the string `session`, it will result in invalid session data 

**To Reproduce**

```python
from secrets import token_bytes
from typing import Any

from starlite import Request, Starlite, get, post
from starlite.middleware.session import SessionCookieConfig

session_config = SessionCookieConfig(secret=token_bytes())

session_data = {"foo": "bar"}


@get("/")
def get_something(request: Request) -> dict[str, Any]:
    return request.session


@post("/")
def set_session_data(request: Request) -> None:
    request.set_session(session_data)


app = Starlite(
    route_handlers=[get_something, set_session_data],
    middleware=[session_config.middleware],
)

client = TestClient(app)

assert client.post("/session").json() == session_data
assert client.get("/session", cookies={"thisisnotasessioncookie": "foo"}).json() == session_data
```

**Additional context**

This is happening because of [this line](this line: https://github.com/starlite-api/starlite/blob/main/starlite/middleware/session.py#L259).

```python
cookie_keys = sorted(key for key in connection.cookies if self.config.key in key)
```
It is meant to check for multiple session cookies, in case data has to be stored in multiple cookies due to size constraints. In these cases, session cookies are created following the schema `session-<i>`, but this isnt' checked for. The line should be

```python
cookie_keys = sorted(key for key in connection.cookies re.fullmatch(fr"{self.config.key}(?:-\d+)?", "session-"))
```
to only check for cookies following the exact pattern potentially set by the middleware (regex should be compiled on the middleware instance beforehand).

provinzkraut:
Fixes #628 by only attempting to read data from cookies if their name matches the pattern used by the middleware for cookie names (`<key>` or `<key>-<n>` if multiple cookies are being used).

# PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+[1.29.1]
+
+- fix `SessionMiddleware` handling non-session cookies with `session` anywhere in their name
+
 [1.29.0]
 
 - Adds native support for `TypedDict` as data type.
--- a/starlite/middleware/session.py
+++ b/starlite/middleware/session.py
@@ -1,5 +1,6 @@
 import binascii
 import contextlib
+import re
 import time
 from base64 import b64decode, b64encode
 from os import urandom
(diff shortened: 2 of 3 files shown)


peterschutt (approved):
Good catch!

Goldziher:
@all-contributors add @provinzkraut code

PR #633, opened 2022-10-20
Allow to exclude routes in CSRFMiddleware

jtraub:
# PR Checklist

### DIFF
--- a/docs/usage/7-middleware/3-builtin-middlewares/3-csrf-middleware.md
+++ b/docs/usage/7-middleware/3-builtin-middlewares/3-csrf-middleware.md
@@ -19,11 +19,11 @@ amount=1000usd&to=attacker@evil.com
 This middleware prevents CSRF attacks by doing the following:
 
 1. On the first "safe" request (e.g `GET`) - set a cookie with a special token created by the server
-2. On each subsequent "unsafe" request (e.g `POST`) - make sure the request contains an additional header that has
-   this token
+2. On each subsequent "unsafe" request (e.g `POST`) - make sure the request contains either a form field or
+   an additional header that has this token
 
-To enable CSRF protection in a Starlite application simply pass an instance of `starlite.config.CSRFConfig`
-to the Starlite constructor:
+To enable CSRF protection in a Starlite application simply pass an instance of
+[`starlite.config.CSRFConfig`][starlite.config.CSRFConfig] to the Starlite constructor:
 
 ```python
 from starlite import Starlite, CSRFConfig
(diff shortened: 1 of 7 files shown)


starlite/app.py
```diff
@@ -603,8 +603,6 @@ def _create_asgi_handler(self) -> "ASGIApp":
             asgi_handler = TrustedHostMiddleware(app=asgi_handler, allowed_hosts=self.allowed_hosts)  # type: ignore
         if self.cors_config:
             asgi_handler = CORSMiddleware(app=asgi_handler, **self.cors_config.dict())  # type: ignore
-        if self.csrf_config:
-            asgi_handler = CSRFMiddleware(app=asgi_handler, config=self.csrf_config)
```
jtraub, line 607: I had to move CSRFMiddleware application to the route handler level (` _build_route_middleware_stack` function) because at this point of execution route is not resolved yet.

starlite/middleware/csrf.py
```diff
@@ -37,6 +38,12 @@ def __init__(
         """
         self.app = app
         self.config = config
+        self.exclude: Optional[Pattern[str]] = None
```
jtraub, line 41: This piece of code has been duplicated from `AbstractAuthenticationMiddleware`.

It feels like we can think and generalize skipping in middleware processing but I still need to play with websocket/generic middlewares first before I can come up with good protocol change.

jtraub:
Tests fail on 3.11 because of changes in enum formatting

Changed [IntEnum](https://docs.python.org/3.11/library/enum.html#enum.IntEnum), [IntFlag](https://docs.python.org/3.11/library/enum.html#enum.IntFlag) and [StrEnum](https://docs.python.org/3.11/library/enum.html#enum.StrEnum) to now inherit from ReprEnum, so their [str()](https://docs.python.org/3.11/library/stdtypes.html#str) output now matches [format()](https://docs.python.org/3.11/library/functions.html#format) (both str(AnIntEnum.ONE) and format(AnIntEnum.ONE) return '1', whereas before str(AnIntEnum.ONE) returned 'AnIntEnum.ONE'.

https://docs.python.org/3.11/whatsnew/3.11.html#enum

image 
image

starlite/middleware/csrf.py
Goldziher: It doesn't work with protocols I'm afraid. We can introduce an AbstractBaseMiddleware

Goldziher (approved):
Looks very good.

docs/usage/7-middleware/3-builtin-middlewares/3-csrf-middleware.md
```diff
@@ -33,4 +33,19 @@ csrf_config = CSRFConfig(secret="my-secret")
 app = Starlite(route_handlers=[...], csrf_config=csrf_config)
 ```
 
+Some routes can be marked as being exempt from the protection by this middleware via
```
peterschutt, line 36:
```suggestion
Some routes can be marked as being exempt from the protection offered by this middleware via
```

peterschutt (approved):
Looks good!

Issue #632, opened 2022-10-20
Enhancement: Etags


Goldziher:
We should implement and `Etag` class following the discussion in #573 .

Goldziher:
@seladb @provinzkraut - one of you wants to take this one up?

provinzkraut:
I have the basic ETag implementation already in a branch, just need to adapt it to the changes introduced with the cache-control support.

So yeah, I'll pick this up!

PR #661, opened 2022-10-23
Add ETag headers (#573)

provinzkraut:
Add ETag header support as per #632 

# PR Checklist

### DIFF
--- a/examples/datastructures/headers/etag.py
+++ b/examples/datastructures/headers/etag.py
@@ -1,0 +1,56 @@
+import random
+import time
+
+from starlite import Controller, Starlite, get
+from starlite.datastructures import ETag
+from starlite.enums import MediaType
+from starlite.response import Response
+
+
+class MyController(Controller):
+    etag = ETag(value="foo")
+
+    @get("/chance_of_rain")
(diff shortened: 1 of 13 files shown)


examples/datastructures/headers/etag.py
```diff
@@ -0,0 +1,56 @@
...
+def get_population_count_dynamic() -> Response[str]:
+    """The etag defined in this route handler will not be returned, and does
+    not need a value.
+
+    It will only be used for OpenAPI generation.
+    """
+    population_count = random.randint(0, 1000)
+    return Response(
+        content=str(population_count),
+        headers={"etag": ETag(value=str(population_count))},
```
Goldziher, line 50: Pydantic should be able to handle stringification

Goldziher (approved):

Goldziher:
Nicely done

examples/datastructures/headers/etag.py
provinzkraut: It can, but mypy and pyright complain about it. If I make the conversion implicit, I'll have to add a `# type: ignore` just to make the pre-commit checks pass, which is a bit noisy for an example IMO.

Issue #421, opened 2022-08-28
Documentation: Improve Documentation


Goldziher:
This is a "standing" refactor ticket. Contributors are welcome to add PRs addressing it, and it is as such a good place to start contributing to the project. PRs addressing any of the following are welcome:

1. Edit and improve the documentation language.
2. Add / improve examples.
3. Write and improve references (using docstrings in code and linking to the references section)
4. Improve the mkdocs site by any of the following: improving CSS, updating setup and plugins, improving UI / UX, updating icons and graphics.
5. Create better logos, repository social cards etc. for Starlite.

Additionally, we welcome any write-ups and blogs posts about the framework - and will gladly link to them in our docs as well as add attributions using `all-contributors`

provinzkraut:
The current plan of attack would look a little something like this:

1. **Examples**  (#343)
    - [ ] Extract all inline-examples from the docs that are more than snippets
    - [ ] Ensure all examples are self-contained and runnable
    - [ ] Ensure all examples are tested
    - [ ] Ensure all examples work on the minimal supported Python version (3.7)
    - [ ] Add alternative examples for higher python versions (can be automated in the build step I think)
2. **Tutorials**
    - [ ] Full step-by-step application tutorial, covering all basic topics, requiring no prior knowledge
    - [ ] Tutorials covering more specific topics (caching, advanced integration, etc.)
3. **Usage**
    - [ ]  Remove tutorial-like sections from "Usage" docs, make them more terse on basics and more in-depth on the technical side
    - [ ] Improve legibility / style of text passages so they have a somewhat coherent writing style (we should come up with some general guidelines for this)
4. **Reference**
    - [ ]  Show inherited members (#427)
    - [ ] Make all docstrings of public interfaces adhere to the same style (#776)
    - [ ] Enforce the docstring style with a pre-commit hook (#776)

PR #679, opened 2022-10-24
Docs rebrush

dkress59:
This PR focuses on the readability of the docs.

The **Tahoma** font has been **replaced** with the system fallback for Helvetica – Tahoma was initially chosen because of its large x-height (height from the baseline to the bow of the letter "h"), plus it's a slab-serif and was thought to be a nice match with the logo's font.
Turns out, a **larger x-height subjectively decreases readability** on screen (opposed to in print), and Tahoma really does not have the most beautiful slabs.

**Y-margins** and line-heights also generally have been **increased** for a better flow from paragraph to paragraph, chapter to chapter. In future there will be even more sophisticated rules for margins between certain types of content.

**Admonition colours** have been adjusted to match the overall theme, with colours also used in the `<code />`/`<pre />` blocks.

**Search Bar and Search Results** have been improved for the light theme.

There will be another PR with the aforementioned spacing/sizing improvements, more highlighting/boldening on keywords and I am also planning on improving the social cards and other points from #421.


Goldziher:
can you run `pre-commit run --all-files`?

dkress59:
sorry about that, lemme quickly install python etc 😆

Issue #672, opened 2022-10-24
Enhancment: Replace starlette's `ServerErrorMiddleware` with our own implementation


provinzkraut:
This is part of #612. Some details have already been laid out there.

jtraub:
I will take of this tonight

PR #696, opened 2022-10-28
Replace ServerErrorMiddleware with own exception printer

jtraub:
# PR Checklist

- [] Have you updated the prose documentation?
- [] Have you updated the reference documentation?

This PR replaces Starlette's ServerErrorMiddleware with custom exception printer. See #672 

I need to document this (docstrings + update the code) + make it handle exception causes before submitting to review.


Goldziher:
If you want assistance making the template look good, we can ask @dkress59 to give you a hand.

jtraub:
Damn, I have to run some errands. I'll be back in 2.5 hours and fix pre-commit complaints I accidentally introduced with the latest force push

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+    """Return html representation of exception frames."""
+    frames = inspect.getinnerframes(exc.__traceback__, frame_limit) if exc.__traceback__ else []
+    result = []
+    for idx, frame in enumerate(reversed(frames)):
+        result.append(create_frame_html(frame=frame, collapsed=(idx > 0)))
+
+    return "".join(result)
+
+
+def create_html_response_contet(exc: Exception, request: "Request", frame_limit: int = 15) -> str:
```
peterschutt, line 80:
```suggestion
def create_html_response_content(exc: Exception, request: "Request", frame_limit: int = 15) -> str:
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+
+tpl_dir = Path(__file__).parent / "templates"
+
+
+def get_symbol_name(frame: inspect.FrameInfo) -> str:
+    """Return full name of the function that is being executed by the given
+    frame."""
+    classname = ""
+    locals_dict = frame.frame.f_locals
+    # probably we are inside of an instance method
```
peterschutt, line 22:
```suggestion
    # probably we are inside an instance method
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+    return body_tpl.format(
+        scripts=scripts,
+        styles=styles,
+        error=f"<span>{escape(exc.__class__.__name__)}</span> on {request.method} {escape(request.url.path)}",
+        error_description=escape(str(exc)),
+        exception_data="".join(exception_data),
+    )
+
+
+def create_plain_text_response_contet(exc: Exception) -> str:
```
peterschutt, line 106:
```suggestion
def create_plain_text_response_content(exc: Exception) -> str:
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+    return f"{classname}{frame.function}"
+
+
+def create_line_html(
+    line: str,
+    line_no: int,
+    frame_index: int,
+    idx: int,
+) -> str:
+    """Return html representation of a line including real line number in the
```
peterschutt, line 40: Docstrings aren't consistent with the rest of the project.

peterschutt (approved):
Nice! Just a few minor things.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+        exception_data="".join(exception_data),
+    )
+
+
+def create_plain_text_response_contet(exc: Exception) -> str:
+    """Return exception traceback in plain text."""
+    return "".join(format_exception(type(exc), value=exc, tb=exc.__traceback__))
+
+
+def create_debug_response(request: "Request", exc: Exception) -> Response:
```
peterschutt, line 111: Are we happy having this function as public api, and adding to the reference docs?

I have some exception translation handlers that look like this:

```python
def service_exception_to_http_response(
    request: Request[Any, Any], exc: ServiceException
) -> Response[ExceptionResponseContent]:
    http_exc: type[HTTPException]
    if isinstance(exc, UnauthorizedException):
        http_exc = ForbiddenException
    else:
        http_exc = InternalServerException
    if http_exc is InternalServerException and request.app.debug:
        return _create_error_response_from_starlette_middleware(request, exc)
    return create_exception_response(http_exc())
```
And ideally I'd swap out `_create_error_response_from_starlette_middleware()` for this.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+from starlite.response import Response
+from starlite.status_codes import HTTP_500_INTERNAL_SERVER_ERROR
+
+if TYPE_CHECKING:
+    from starlite.connection import Request
+
+tpl_dir = Path(__file__).parent / "templates"
+
+
+def get_symbol_name(frame: inspect.FrameInfo) -> str:
```
Goldziher, line 17: Please destructure rather than import the namespace. Thus `FrameInfo` instead of `inspect. FrameInfo`. Then you can actually do:

`"FrameInfo"` and move this to type checking import only.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,121 @@
...
+    """Return full name of the function that is being executed by the given
+    frame."""
+    classname = ""
+    locals_dict = frame.frame.f_locals
+    # probably we are inside of an instance method
+    # this piece assumes that the code uses standard names "self" and "cls"
+    # in instance and class methods
+    if locals_dict.get("self"):
```
Goldziher, line 25: how about:


```python
cls_or_instance = locals_dict.get("self") or locals_dict.get("cls")

classname = get_name(cls_or_instance) if cls_or_instance else "unknown"
```

note `get_name` is a util we have in place already

starlite/middleware/exceptions/debug_response.py
Goldziher: by which @peterschutt means we should use google style doc strings and make sure all params are documented properly.

jtraub: @Goldziher @peterschutt can we clarify requirements for documenting functions and methods because I am confusedl

CONTRIBUING.md states

> All public functions and methods should be documented with a doc string. The project uses the [Google style docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Private methods should have a doc string explaining what they do, but do not require an elaborate doc string.

I guess we should update the doc if we want to follow Google style at all times. 

Also, I saw a lot of code that uses simple docstrings. Like [this](https://github.com/starlite-api/starlite/blob/main/starlite/plugins/tortoise_orm.py#L102). Am I right assuming that it is due to it being an old piece?

Goldziher: > @Goldziher @peterschutt can we clarify requirements for documenting functions and methods, please, because I am confused
> 
> CONTRIBUING.md states
> 
> > All public functions and methods should be documented with a doc string. The project uses the [Google style docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Private methods should have a doc string explaining what they do, but do not require an elaborate doc string.
> 
> I guess we should update the doc if we want to follow Google style at all times.
> 
> Also, I saw a lot of code that uses simple docstrings. Like [this](https://github.com/starlite-api/starlite/blob/main/starlite/plugins/tortoise_orm.py#L102). Am I right assuming that it is due to it being an old piece?

Yes, its an ongoing effort - but all new code that has public methods should be fully documented.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,123 @@
...
+    """Return html representation of exception frames."""
+    frames = getinnerframes(exc.__traceback__, frame_limit) if exc.__traceback__ else []
+    result = []
+    for idx, frame in enumerate(reversed(frames)):
+        result.append(create_frame_html(frame=frame, collapsed=(idx > 0)))
+
+    return "".join(result)
+
+
+def create_html_response_content(exc: Exception, request: "Request", frame_limit: int = 15) -> str:
```
Goldziher, line 82: i wonder whether we should also handle `ExceptionGroups`, which would require we use a backport for this.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,123 @@
...
+        cause_error_description = f"<h3><span>{escape(str(cause))}</span></h3>"
+        cause_error = f"<h4><span>{escape(cause.__class__.__name__)}</span></h4>"
+        exception_data.append(
+            f'<div class="cause-wrapper">{cause_header}{cause_error}{cause_error_description}{cause_data}</div>'
+        )
+        cause = cause.__cause__
+
+    scripts = (tpl_dir / "scripts.js").read_text()
+    styles = (tpl_dir / "styles.css").read_text()
+    body_tpl = (tpl_dir / "body.html").read_text()
```
Goldziher, line 98: I think you can do these outside the function and store as variables, rather than calculate for each exception.

starlite/middleware/exceptions/debug_response.py
jtraub: I was thinking about this but decided not to handle groups in the Starlite since the framework  itself and its dependencies, nor other major libs,  won't be throwing `ExceptionGroup`s in the   near future.

The only source of them would be the user code and I don't think it is good idea to add another package to the dependencies list only to show more details about a a group.

I am working on a rich exception renderer which will be a separate package. I suggest to handle exception groups in this package and direct users willing to see nice output for their exception groups to install the package.

starlite/middleware/exceptions/debug_response.py
jtraub: I put them inside the function to save on memory in production environments.

I am planning to add syntax highlight via [Highlight.js](https://highlightjs.org/) and something else which would increase file size up to ~70kb. I concluded that it is better to have a small performance impact in the dev envrionment where performance doesn't matter in most cases than take even small pieces of memory with useless data in production.

starlite/middleware/exceptions/debug_response.py
jtraub: Understood. Updated doc strings.

starlite/middleware/exceptions/debug_response.py
jtraub: @peterschutt can we wait a bit (probably a week) with this? I am working on a separate project that provides more elaborate exception renderer. Most likely function interface will remain the same but I prefer to be 100% sure before declaring something a public API.

starlite/middleware/exceptions/debug_response.py
Goldziher: Sure

starlite/middleware/exceptions/debug_response.py
peterschutt: Yes, of course.

It would be nice if there was a way to push the logic of serving the debug response down into `create_exception_response()` so that users can get that for free when creating their own exception handlers such as above. Issue is that `create_exception_response()` doesn't have a way to check `app.debug`.

starlite/middleware/exceptions/debug_response.py
jtraub: @Goldziher @peterschutt could you please check docstrings again. English is not my first language so I would appreciate your help here.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,186 @@
...
+    source code.
+
+    Args:
+        line: A string representing the current line.
+        line_no: The line number associated with the executed line.
+        frame_index: Index of the executed line in the code context.
+        idx: Index of the current line in the code context.
+
+    Returns:
+        A string containing HTML representation of the given line
```
peterschutt, line 61:
```suggestion
        A string containing HTML representation of the given line.
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,186 @@
...
+        # so adding index (aka relative number) of a line in the code context we can calculate its actual number in the source file,
+        "line_no": line_no - frame_index + idx,
+        "line": escape(line).replace(" ", "&nbsp"),
+        "line_class": "executed-line" if idx == frame_index else "",
+    }
+    return template.format(**data)
+
+
+def create_frame_html(frame: "FrameInfo", collapsed: bool) -> str:
+    """Produce html representation of the given frame object including filename
```
peterschutt, line 75:
```suggestion
    """Produce HTML representation of the given frame object including filename
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,186 @@
...
+        "code": "".join(code_lines),
+        "frame_class": "collapsed" if collapsed else "",
+    }
+    return frame_tpl.format(**data)
+
+
+def create_exception_html(exc: BaseException, line_limit: int) -> str:
+    """Produce html representation of the exception.
+
+    [frames](https://docs.python.org/3/reference/datamodel.html#frame-objects).
```
peterschutt, line 104:
```suggestion
    """Produce HTML representation of the exception
    [frames](https://docs.python.org/3/reference/datamodel.html#frame-objects).
```

peterschutt: I think it is docformatter pre-commit lint that breaks the doc strings into multiple lines like this.

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Args:
+        exc: An Exception instance to generate .
```
peterschutt, line 107:
```suggestion
        exc: An Exception instance to generate.
```

starlite/middleware/exceptions/debug_response.py
```diff
@@ -0,0 +1,186 @@
...
+        exc: An Exception instance to render debug response from.
+
+    Returns:
+        A string containing exception traceback.
+    """
+    return "".join(format_exception(type(exc), value=exc, tb=exc.__traceback__))
+
+
+def create_debug_response(request: "Request", exc: Exception) -> Response:
+    """Create debug response either in plain text or html depending on client
```
peterschutt, line 169:
```suggestion
    """Create debug response either in plain text or HTML depending on client
```

peterschutt (approved):

starlite/middleware/exceptions/debug_response.py
provinzkraut: Yes. You have to preemptively break the lines yourself. We should maybe see if we can fix that somehow.

provinzkraut (approved):

PR #630, opened 2022-10-20
Server-side sessions and unified session backends

provinzkraut:
This is a bit WIP. Aims to implement #617 

# PR Checklist

- [x] Have you followed the guidelines in `CONTRIBUTING.md`?
- [x] Have you got 100% test coverage on new code?
- [x] Have you updated the prose documentation?
- [x] Have you updated the reference documentation?

PR #697, opened 2022-10-28
Support setting sessions explicit to `Empty`.

provinzkraut:
This was supported in the previous `SessionMiddleware` and was erroneously removed in #630. Test have been added.

# PR Checklist

### DIFF
--- a/starlite/connection/base.py
+++ b/starlite/connection/base.py
@@ -5,6 +5,7 @@ from typing import (
     Generic,
     List,
     Optional,
+    Type,
     TypeVar,
     Union,
     cast,
@@ -260,7 +261,7 @@ class ASGIConnection(Generic[Handler, User, Auth]):
         """
         return self.app.cache
 
-    def set_session(self, value: Union[Dict[str, Any], "BaseModel"]) -> None:
+    def set_session(self, value: Union[Dict[str, Any], "BaseModel", Type["Empty"]]) -> None:
         """Helper method to set the session in scope.
 
         If the [Starlite SessionMiddleware][starlite.middleware.session.SessionMiddleware] is
--- a/starlite/connection/base.py
+++ b/starlite/connection/base.py
@@ -5,7 +5,6 @@ from typing import (
     Generic,
     List,
     Optional,
-    Type,
     TypeVar,
     Union,
     cast,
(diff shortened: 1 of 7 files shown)


starlite/middleware/session/base.py
```diff
@@ -276,17 +277,18 @@ async def store_in_message(
...
-        serialised_data = self.serlialize_data(scope_session, scope)
-        await self.set(session_id=session_id, data=serialised_data)
-
         cookie_params = self.config.dict(
             exclude_none=True,
             exclude={"secret", "key"} | set(self.config.__fields__) - set(BaseBackendConfig.__fields__),
         )
 
-        if scope_session:
+        if scope_session is not Empty:
```
infohash, line 285: Yes I removed `Empty`. If the session is cleared it becomes an empty dictionary because the user should be able to refill the session after clearing it and for an empty session (as in empty dict) this block should not run.

provinzkraut: Hm. I guess you have a point there, however, removing the `Empty` check breaks starlite-sessions. That’s why I re-added it.

Goldziher: why would the user refill the same dictionary? Also, the `scope_session` is a value stored in the scope, its not support to be exposed directly to used as such. So `Empty` is correct here IMO.

Goldziher (approved):

starlite/middleware/session/base.py
provinzkraut: It also makes it easier for the server-side backends to know when to explicity delete a session instead of replacing it with an empty one.

starlite/connection/base.py
```diff
@@ -260,7 +261,7 @@ def cache(self) -> "Cache":
         """
         return self.app.cache
 
-    def set_session(self, value: Union[Dict[str, Any], "BaseModel"]) -> None:
+    def set_session(self, value: Union[Dict[str, Any], "BaseModel", Type["Empty"]]) -> None:
```
odiseo0, line 264: Why `Type["Empty"]`? Shouldn't be `EmptyType` instead?

starlite/middleware/session/base.py
infohash: @Goldziher What I mean is if the user does `request.session.clear`, the session becomes an empty dictionary for which this if condition should not become True.

provinzkraut: @infohash You're right about that, this is actually a mistake. Wonder why it's not caught by the tests.. I implemented this correctly for the cookie backend. Will update accordingly!

starlite/connection/base.py
provinzkraut: It should be, good catch!

odiseo0 (approved):
LGTM

PR #705, opened 2022-10-29
Test client cleanup

provinzkraut:
- Removes `cryptography` as a dependency of the `testing` package-extra
- Moves `TestClient` async session-backend methods to run in `TestClient.portal` instead of `anyio.run` directly
- Ensures `TestClient` tests in `tests/testing` run with both `asyncio`and `trio` backend

# PR Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ aiomcache = {version = "*", optional = true}
 anyio = ">=3"
 brotli = { version = "*", optional = true }
 cryptography = { version = "*", optional = true }
-httpx = { version = ">=0.22", optional = true }
+httpx = { version = ">=0.22" }
 orjson = "*"
 picologging = { version = "*", optional = true }
 pydantic = "*"
@@ -84,10 +84,9 @@ trio = "*"
 fakeredis = "^1.9.4"
 
 [tool.poetry.extras]
-testing = ["httpx"]
 brotli = ["brotli"]
 cryptography = ["cryptography"]
-full = ["httpx", "brotli", "cryptography", "picologging", "structlog"]
+full = ["brotli", "cryptography", "picologging", "structlog"]
 picologging = ["picologging"]
 structlog = ["structlog"]
 redis = ["redis"]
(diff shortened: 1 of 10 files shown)


tests/testing/test_testing.py
```diff
@@ -255,13 +270,17 @@ def get_session_data(request: Request) -> Dict[str, Any]:
...
-    with TestClient(app=app, session_config=session_config) as client:
+    with TestClient(app=app, session_config=session_config, backend=test_client_backend) as client:
         client.set_session_data(session_data)
         assert session_data == client.get("/test").json()
 
 
 @pytest.mark.parametrize("with_domain", [False, True])
-def test_test_client_get_session_data(with_domain: bool, session_config: "BaseBackendConfig") -> None:
+def test_test_client_get_session_data(
+    with_domain: bool,
```
infohash, line 280: Was this test using this argument earlier?

provinzkraut: As a matter of fact it wasn't. It should have, but it didn't. I just tried adding it but it fails because of the issue described in #706. Will add an `xfail` for now.

pyproject.toml
```diff
@@ -84,7 +84,7 @@ trio = "*"
 fakeredis = "^1.9.4"
 
 [tool.poetry.extras]
-testing = ["httpx", "cryptography"]
+testing = ["httpx"]
```
Goldziher, line 87: i think we can eliminate the testing extra - and have httpx installed as part of the core package, hence have the test client as a builtin and also simplify the export / import paths (we can then export from the root `__init__.py` file)

Goldziher (approved):

pyproject.toml
provinzkraut: Could you elaborate on simplifying the export / import paths? You mean just re-export `TestClient` and `create_test_client` from `starlite/__init__.py`?

tests/testing/test_testing.py
provinzkraut: Addressed

pyproject.toml
Goldziher: Yup. 

This is possible if httpx is not an extra

provinzkraut: Done

infohash (approved):

PR #711, opened 2022-10-30
Fix transient cryptography dependency in testing.test_client.client

provinzkraut:
This changes the implementations of the session-related methods in `TestClient` to use the actual middleware by constructing fake "Scope" and "ASGIConnection" objects

# PR Checklist

### DIFF
--- a/starlite/testing/test_client/client.py
+++ b/starlite/testing/test_client/client.py
@@ -1,5 +1,6 @@
 import warnings
 from contextlib import ExitStack, contextmanager
+from http.cookiejar import CookieJar
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -15,23 +16,19 @@ from typing import (
 from urllib.parse import urljoin
 
 from anyio.from_thread import BlockingPortal, start_blocking_portal
+from starlette.datastructures import MutableHeaders
 
-from starlite import HttpMethod, ImproperlyConfiguredException
+from starlite import ASGIConnection, HttpMethod, ImproperlyConfiguredException
 from starlite.exceptions import MissingDependencyException
-from starlite.middleware.session.base import ServerSideBackend, ServerSideSessionConfig
-from starlite.middleware.session.cookie_backend import (
-    CookieBackend,
-    CookieBackendConfig,
-)
 from starlite.testing.test_client.life_span_handler import LifeSpanHandler
 from starlite.testing.test_client.transport import (
     ConnectionUpgradeException,
     TestClientTransport,
 )
-from starlite.types import AnyIOBackend, ASGIApp
+from starlite.types import AnyIOBackend, ASGIApp, HTTPResponseStartEvent
 
 try:
-    from httpx import USE_CLIENT_DEFAULT, Client, Response
+    from httpx import USE_CLIENT_DEFAULT, Client, Cookies, Request, Response
 except ImportError as e:
     raise MissingDependencyException(
         "To use starlite.testing, install starlite with 'testing' extra, e.g. `pip install starlite[testing]`"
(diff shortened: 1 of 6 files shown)


starlite/testing/test_client/client.py
```diff
@@ -53,16 +50,40 @@
...
 
-def raise_for_unsupported_session_backend(backend: "BaseSessionBackend") -> None:
-    raise ImproperlyConfiguredException(f"Backend of type {type(backend)!r} is currently not supported")
+def fake_http_send_message(headers: MutableHeaders) -> HTTPResponseStartEvent:
+    headers.setdefault("content-type", "application/text")
+    return HTTPResponseStartEvent(type="http.response.start", status=200, headers=headers.raw)
+
+
+def fake_asgi_connection(app: ASGIApp, cookies: Dict[str, str]) -> ASGIConnection[Any, Any, Any]:
+    scope = {
```
Goldziher, line 66: Since we have multiple places in testing where we create a mock scopes, it might be useful to split this into a utility that offers type casting inside it.

starlite/testing/test_client/client.py
```diff
@@ -124,11 +145,13 @@ def session(self) -> "CookieBackend":
             "To access the session backend directly, use the session_backend attribute",
             PendingDeprecationWarning,
```
Goldziher, line 146: I know this is before you, hut why are we deprecating this?

starlite/testing/test_client/client.py
```diff
@@ -653,42 +676,35 @@ def test_something(self, test_client: TestClient) -> None:
...
-            session_id = self.cookies.setdefault(
-                self.session_backend.config.key, self.session_backend.generate_session_id()
-            )
-            await self.session_backend.set(session_id, serialized_data)
-        elif isinstance(self.session_backend, CookieBackend):
-            for key, value in self._create_session_cookies(self.session_backend, data).items():
-                self.cookies.set(key, value)
-        else:
-            raise_for_unsupported_session_backend(self.session_backend)
+        mutable_headers = MutableHeaders({})
```
Goldziher, line 685: You can simply use a dict here i think

provinzkraut:
The fake scope still needs to be enriched dynamically

Goldziher (approved):

starlite/testing/test_client/client.py
provinzkraut: Hm, idk. You suggested changing the name from `session` to `session_backend` at some point, so I made `session` only return a `CookieBackend` and deprecate the method. We could remove the deprecation as well, but I think in the long run, having only the `session_backend` does make more sense, since it's "correcter". It *does* return a session backend, and not a session.

starlite/testing/test_client/client.py
provinzkraut: Do they share a lot of the functionality? Maybe we can just create a `create_fake_base_scope` function, that accepts kwargs to update the scope?

starlite/testing/test_client/client.py
Goldziher: Sure, I didn't remember. Let's set a version for dropping this? Say 1.40?

starlite/testing/test_client/client.py
provinzkraut: I could, but then I'd have to convert the dict to `MutableHeaders` (or `Headers`) in `fake_http_send_message` (for converting them to a bytes list), and IMO it's a bit clearer like this, since the same object gets used throughout.

starlite/testing/test_client/client.py
provinzkraut: Shouldn't this only happen in a major version since we're following semver?

Goldziher: we can do this - but we are not going for a major version until pydantic 2.0 is out i think.

provinzkraut: Well I don't mind not following semver. It's not the end-all be-all.

Issue #640, opened 2022-10-22
Replace starlette.datastructures.Headers and other MultiDict-like structures with aio-libs/multidict


jtraub:
As a part of migration from Starlette we should replace `starlette.datastructures.Headers`, `starlette.datastructures.MutableHeaders` and `starlette.datastructures.ImmutableMultiDict` with another multidict implementation.

Proposed library is  [multidict](https://github.com/aio-libs/multidict)

Convenient usage list for mentioned classes can be found in #634 

Related issue in starlite-multipart tracker - https://github.com/starlite-api/starlite-multipart/issues/5

Goldziher:
It occurs to me we can create our own version using https://docs.python.org/3/library/collections.html#collections.ChainMap

This has the advantage if being both fast and a builtin. It's not a multidict, but we can use it as a basis. 

Thoughts?

provinzkraut:
We could, but I'm not sure if we should. I don't think it's that fast, certainly not faster than the proposed multidict. And we'd have to add most of the implementation, not sure if there's really any benefit to using `ChainMap` here really (or maybe you have some genius idea on how to do that, that I just can't see? :upside_down_face: If so, please do share!).

Goldziher:
> We could, but I'm not sure if we should. I don't think it's that fast, certainly not faster than the proposed multidict. And we'd have to add most of the implementation, not sure if there's really any benefit to using `ChainMap` here really (or maybe you have some genius idea on how to do that, that I just can't see? :upside_down_face: If so, please do share!).

Nothing genius I'm afraid. Let's stick with multidicting This

provinzkraut:
I had this on my plate already, unless @jtraub wants to take this on I'd take care of it in the following days?

provinzkraut:
Implemented in #732

PR #732, opened 2022-11-01
Replace starlettes `Headers` and `MutableHeaders`

provinzkraut:
This is part of #640 and replaces starlette's `Headers` with our own `multidict`-based version, and `MutableHeaders` with an adapted version from starlette

# PR Checklist

### DIFF
--- a/docs/reference/datastructures/3-headers.md
+++ b/docs/reference/datastructures/3-headers.md
@@ -1,5 +1,38 @@
 # Headers
 
+::: starlite.datastructures.Headers
+    options:
+        members:
+            - __init__
+            - from_scope
+            - from_message
+            - raw
+            - keys
+            - values
+            - items
+            - getlist
+            - mutablecopy
+
+
+::: starlite.datastructures.MutableHeaders
+    options:
+        members:
+            - __init__
+            - from_scope
+            - from_message
+            - raw
+            - keys
+            - values
+            - items
+            - getall
+            - getlist
+            - mutablecopy
+            - setdefault
+            - update
+            - append
+            - add_vary_header
+
+
 ::: starlite.datastructures.ResponseHeader
     options:
         members:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ starlette = ">=0.21"
 starlite-multipart = ">=1.2.0"
 structlog = { version = "*", optional = true }
 typing-extensions = "*"
+multidict = "^6.0.2"
 
 [tool.poetry.group.dev.dependencies]
 aiomcache = "*"
--- a/starlite/datastructures/headers.py
+++ b/starlite/datastructures/headers.py
@@ -1,13 +1,58 @@
 import re
 from abc import ABC, abstractmethod
-from typing import Any, ClassVar, Dict, Optional
+from typing import Any, ClassVar, Dict, Iterator, List, Mapping, Optional, Tuple
+from multidict import MultiDict, CIMultiDictProxy, CIMultiDict
 
 from pydantic import BaseModel, Extra, Field, ValidationError, validator
+from starlette.datastructures import MutableHeaders
 from typing_extensions import Annotated
 
 from starlite.exceptions import ImproperlyConfiguredException
 
 ETAG_RE = re.compile(r'([Ww]/)?"(.+)"')
+
+
+class Headers(CIMultiDictProxy[str, str]):
+    """
+    An immutable, case-insensitive multidict.
+    """
+
+    def __init__(
+        self,
+        headers: Optional[Mapping[str, str]] = None,
+        raw: Optional[List[Tuple[bytes, bytes]]] = None,
+        scope: Optional[Mapping[str, Any]] = None,
+    ) -> None:
+        _list: List[Tuple[str, str]] = []
+        if headers is not None:
+            assert raw is None, 'Cannot set both "headers" and "raw".'
+            assert scope is None, 'Cannot set both "headers" and "scope".'
+            _list = [(key.lower(), value) for key, value in headers.items()]
+        elif raw is not None:
+            assert scope is None, 'Cannot set both "raw" and "scope".'
+            _list = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in raw]
+        elif scope is not None:
+            _list = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in scope["headers"]]
+        super().__init__(CIMultiDict(_list))
+
+    @property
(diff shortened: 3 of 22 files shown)


docs/reference/datastructures/3-headers.md
```diff
@@ -1,5 +1,38 @@
...
+            - keys
+            - values
+            - items
+            - getall
+            - getlist
+            - mutablecopy
+            - setdefault
+            - update
+            - append
+            - add_vary_header
```
Goldziher, line 33: i usually sort keys alphabetically here, not by the order of the code. 

tip: https://plugins.jetbrains.com/plugin/2162-string-manipulation

pyproject.toml
```diff
@@ -56,6 +56,7 @@ starlette = ">=0.21"
 starlite-multipart = ">=1.2.0"
 structlog = { version = "*", optional = true }
 typing-extensions = "*"
+multidict = "^6.0.2"
```
Goldziher, line 59:
```suggestion
multidict = ">=6.0.2"
```

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,396 @@
+"""The code for `MutableHeaders` and parts of `Headers` was adopted from https:
```
Goldziher, line 1: why do we need to adopt these? why not implement our own version using the new mutldict package?

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,396 @@
...
+    def mutablecopy(self) -> "MutableHeaders":
+        """Create a mutable copy.
+
+        Returns:
+            A [MutableHeaders][starlite.datastructures.headers.MutableHeaders] instance
+        """
+        return MutableHeaders(self.raw[:])
+
+
+class MutableHeaders(Mapping):
```
jtraub, line 166: Why do we use only `CIMultiDictProxy`? The lib has mutable implementation as well.

I guess there is a good reason not to use it but I can not grasp it.

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,396 @@
...
+        Args:
+            scope: An ASGI Scope
+
+        Returns:
+            Headers
+        """
+        return cls(list(scope["headers"]))
+
+    @classmethod
+    def from_message(cls, message: "Message") -> "Headers":
```
Goldziher, line 94: `Message` is too wide here, it should be restrictred only to the ASGI event types that support headers IMO.

starlite/datastructures/headers.py
Goldziher, line 166: so, i don't quite understand why this is vendored. Why not implement our own version? We could use either `CIMultiDict` or `MultiDict` as base classes

starlite/utils/deprecation.py
```diff
@@ -0,0 +1,95 @@
+import inspect
```
Goldziher, line 1: this is pretty cool!

starlite/utils/deprecation.py
```diff
@@ -0,0 +1,95 @@
...
+P = ParamSpec("P")
+DeprecatedKind = Literal["function", "method", "attribute", "property", "class", "parameter"]
+
+
+def warn_deprecation(
+    version: str,
+    deprecated_name: str,
+    kind: DeprecatedKind,
+    *,
+    removal_in: Optional[str] = None,
```
Goldziher, line 18:
```suggestion
    removal_in: str ="next major version",
```

?

Goldziher (commented):
Awaiting some discussion in slack.

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,396 @@
...
+
+
+class Headers(CIMultiDictProxy[str]):
+    """An immutable, case-insensitive [multidict](https://multidict.aio-
+    libs.org/en/stable/multidict.html#cimultidictproxy) for HTTP headers."""
+
+    def __init__(self, headers: Optional[Union[Mapping[str, str], "RawHeadersList"]] = None) -> None:
+        headers_: Union[Mapping[str, str], List[Tuple[str, str]]] = {}
+        if isinstance(headers, list):
+            headers_ = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in headers]
```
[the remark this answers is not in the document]
provinzkraut: No. `_encode_headers` turns string headers to bytes headers. Here we turn bytes headers into string headers.

starlite/datastructures/headers.py
provinzkraut: Because we don't want a mutable multidict here. `Headers` should be immutable

jtraub: It doesn't make any sense. Word `Mutable` is literally in the class name and yet we want Headers to be immutable

provinzkraut: Oh, I get it now. I was confused before, sorry. This was done because the starlette version of this exposes internal datastrucutres that allow the headers to be mutated via side effects, which I tried to keep in an effort to be backwards compatible. The actual `CIMultiDictProxy` of course does not support this.

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,182 @@
...
+
+        Returns:
+            Headers
+
+        Raises:
+            ValueError: If the message does not have a `headers` key
+        """
+        return cls(scope["headers"])
+
+    def to_header_list(self) -> "RawHeadersList":
```
jtraub, line 65: Do we test this method?

Goldziher (approved):

starlite/datastructures/headers.py
```diff
@@ -1,15 +1,182 @@
...
+        indices = self._find_indices(name)
+        for i in indices[::-1]:
+            del self.headers[i]
+
+    def __len__(self) -> int:
+        return len(self.headers)
+
+    def __iter__(self) -> Iterator[str]:
+        """iter over header names including duplicates."""
+        return iter(h[0].decode("utf-8") for h in self.headers)
```
jtraub, line 177: I am puzzled to see here `decode("utf-8")` because header names are encoded to `latin-1` everywhere

jtraub (approved):

starlite/datastructures/headers.py
provinzkraut: Oh, good catch. That must have been an autocomplete I overlooked (from my brain or IDE :D)

starlite/datastructures/headers.py
provinzkraut: Now we do :upside_down_face:

PR #743, opened 2022-11-03
Added per request caching of dependencies

ottermata:
This avoids executing a dependency multiple times, if it's used more than once in the dependency tree.

Still unsure about the following:
* using `request.state` for the cache
  * the key `"_dependency_cache"` in general, without the leading underscore perhaps?
  Basically: Should this be part of the public interface?
* is there still a use case for `use_cache`?

# PR Checklist

### DIFF
--- a/starlite/datastructures/provide.py
+++ b/starlite/datastructures/provide.py
@@ -1,4 +1,7 @@
+from inspect import Parameter, Signature
 from typing import TYPE_CHECKING, Any, Optional, cast
+
+from anyio import Lock
 
 from starlite.types import Empty
 from starlite.utils.sync import AsyncCallable, is_async_callable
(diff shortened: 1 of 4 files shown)


starlite/datastructures/provide.py
```diff
@@ -1,27 +1,79 @@
...
 
 if TYPE_CHECKING:
     from typing import Type
 
+    from starlite.connection import Request
     from starlite.signature import SignatureModel
     from starlite.types import AnyCallable
 
 
+def _wrap_cache_for_request(fn: "AnyCallable", override_name: Optional[str]) -> "AnyCallable":
```
Goldziher, line 17: this is a good place to use a `Generic` parameter T, and later to use `ParamSpec` P. See for an example: `starlite/utils/sync.py`.

starlite/datastructures/provide.py
```diff
@@ -1,27 +1,79 @@
...
+    """Wrap the function to cache the result per request."""
+    if override_name:
+        key = override_name
+    else:
+        key = fn.__name__
+    lock = Lock()
+    wrapped_sig = sig = Signature.from_callable(fn)
+    include_request = True
+    if "request" not in sig.parameters:
```
Goldziher, line 26:
```suggestion

    if "request" not in sig.parameters:
```

starlite/datastructures/provide.py
```diff
@@ -1,27 +1,79 @@
...
+        include_request = False
+        params = list(sig.parameters.values())
+        params.append(Parameter("request", Parameter.KEYWORD_ONLY, annotation="Request"))
```
Goldziher, line 29: clever. 

I would though not do this here - instead of manipulating the signature model to add this parameter, I would add resolution logic to handle `cache_per_request` the kwargs logic: https://github.com/starlite-api/starlite/blob/main/starlite/kwargs.py

starlite/datastructures/provide.py
```diff
@@ -1,27 +1,79 @@
...
 
         Args:
             dependency: Callable to inject, can be a function, method or class.
             use_cache: Cache the dependency return value. Defaults to False.
+            cache_per_request: Cache the dependency return value per request. Defaults to False.
+            cache_key: Override the key for per request caching. Defaults to the function name.
             sync_to_thread: Run sync code in an async thread. Defaults to False.
         """
         self.dependency = cast("AnyCallable", AsyncCallable(dependency) if sync_to_thread else dependency)
+        if cache_per_request:
```
Goldziher, line 75:
```suggestion

        if cache_per_request:
```

starlite/datastructures/provide.py
```diff
@@ -1,27 +1,79 @@
...
+            self.dependency = _wrap_cache_for_request(self.dependency, cache_key)
         self.use_cache = use_cache
```
Goldziher, line 77:
```suggestion
        self.use_cache = use_cache
        
```

Goldziher (commented):
Very nice PR, highly appreciated! I left some comments.

starlite/datastructures/provide.py
ottermata: I moved the logic over to the KwargsModel, is this what you meant?

Goldziher (approved):
very nice stuff

Goldziher:
@all-contributors add @ottermata code

Goldziher:
Let me know when ready, this can be merged as fat as im concerned.

ottermata:
Feel free to merge it, I'm new to this 😅

Goldziher:
> Feel free to merge it, I'm new to this 😅

very nice work

Issue #667, opened 2022-10-24
Enhancment: Replace starlette's `URL`, `URLPath` and `Address` with our own implementation


provinzkraut:
This is part of #612 / #634.

We should replace `URL`, `URLPath` and `Address` from `starlette.datastructures` with our own implementation.

PR #759, opened 2022-11-06
add `QueryMultiDict`

Goldziher:
# PR Checklist

This PR adds a multi dict for query params. 

It might be that this implementation will degrade performance a bit, so we need to test and see how it stacks vis-a-vis the previous one. I will run some benchmarks tomorrow to see.

### DIFF
--- a/starlite/datastructures/multi_dicts.py
+++ b/starlite/datastructures/multi_dicts.py
@@ -1,4 +1,5 @@
-from typing import Any, Iterable, List, Mapping, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple, Union
+from urllib.parse import parse_qsl
 
 from multidict import MultiDict, MultiDictProxy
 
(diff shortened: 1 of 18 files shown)


peterschutt (approved):

provinzkraut:
@Goldziher Let me know what the outcome of your tests is. If it looks good, I'll incorporate it into #667

starlite/datastructures/multi_dicts.py
```diff
@@ -41,3 +42,41 @@ async def close(self) -> None:
...
+
+        Returns:
+            A QueryMultiDict instance
+        """
+        _bools = {"true": True, "false": False, "True": True, "False": False}
+        return cls(
+            (k, v) if v not in _bools else (k, _bools[v]) for k, v in parse_qsl(query_string, keep_blank_values=True)
+        )
+
+    def dict(self) -> Dict[str, List[Any]]:
```
provinzkraut, line 68: Why not

```suggestion
    def dict(self) -> Dict[str, List[Any]]:
        return {self.getall(key) for key in set(self.keys)}`
```

starlite/datastructures/multi_dicts.py
```diff
@@ -41,3 +42,41 @@ async def close(self) -> None:
         for _, value in self.multi_items():
             if isinstance(value, UploadFile):
                 await value.close()
+
+
+class QueryMultiDict(MultiDict[Any]):
```
provinzkraut, line 47: The way we parse it, the items can only ever be a `str` or `bool`, so why not annotate it as such?

starlite/datastructures/multi_dicts.py
Goldziher: i would assume this is slower, ill test

PR #761, opened 2022-11-07
Add `*args` override to `Logger` protocol

cofin:
This PR extends the `Logger` protocol to include overrides for `*args`.  This prevents `logger.info("%s finished",my_var)` from being reported as a typing error.

### DIFF
--- a/starlite/types/protocols.py
+++ b/starlite/types/protocols.py
@@ -20,7 +20,7 @@ class Logger(Protocol):  # pragma: no cover
 
         Args:
              event: Log message.
-             **kwargs: Any kwargs.
+             *args: Any args.
         """
         ...
 
@@ -29,6 +29,7 @@ class Logger(Protocol):  # pragma: no cover
 
         Args:
              event: Log message.
+             *args: Any args.
              **kwargs: Any kwargs.
         """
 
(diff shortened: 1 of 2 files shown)


tests/logging_config/test_logging_config.py
```diff
@@ -77,6 +77,8 @@ def test_standard_queue_listener_logger(logger: "Logger", caplog: "LogCaptureFix
     with caplog.at_level("INFO"):
         logger.info("Testing now!")
         assert "Testing now!" in caplog.text
+        var = "test_var"
+        logger.info("%s", var)
```
cofin, line 81: I missed the actual assert check here.  Will add shortly.

starlite/types/protocols.py
```diff
@@ -1,57 +1,178 @@
...
         """
+        ...
 
+    @overload
+    def debug(self, event: str, *args: Any) -> Any:
+        """Outputs a log message at 'DEBUG' level.
+
+        Args:
+             event: Log message.
+             **kwargs: Any kwargs.
```
jtraub, line 23: Shouldn't it be `*args`?

cofin: Good catch.  I've pushed updates to all of the docstrings.

PR #784, opened 2022-11-10
Fix #783 by setting content-disposition to inline in html_mode

### LINKED ISSUE #783: Bug: StaticFiles sends files as `content-disposition: 'attachment'` in html-mode
**Describe the bug**
When using `StaticFiles` in html-mode, files are being sent with `content-disposition: 'attachment'`

**To Reproduce**
Create an `html/index.html` file. Run:

```python
from starlite import Starlite, StaticFilesConfig, TestClient

app = Starlite(
    static_files_config=[StaticFilesConfig(path="/", directories=["html"], html_mode=True)], route_handlers=[]
)

    res = client.get("/index.html")
    assert not res.headers["content-disposition"].startswith("attachment")
```

provinzkraut:
This fixes #783, by correctly setting the `Content-Disposition` header to `inline` instead of `attachment`.

# PR Checklist

### DIFF
--- a/starlite/static_files/base.py
+++ b/starlite/static_files/base.py
@@ -1,5 +1,5 @@
 from os.path import commonpath, join
-from typing import TYPE_CHECKING, List, Literal, Tuple, Union
+from typing import TYPE_CHECKING, List, Tuple, Union
 
 from starlite.enums import ScopeType
 from starlite.exceptions import MethodNotAllowedException, NotFoundException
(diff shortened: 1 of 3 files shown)


jtraub:
@provinzkraut wouldn't normal mode be affected too? I mean what what would happen if I remove `html_mode=True` and try to open localhost:8000/index.html

provinzkraut:
> @provinzkraut wouldn't normal mode be affected too? I mean what would happen if I remove `html_mode=True` and try to open localhost:8000/index.html

Then it would have the content-disposition as `attachment`, as intended.

Issue #669, opened 2022-10-24
Enhancment: Replace starlette's `CORSMiddleware` with our own implementation


provinzkraut:
This is part of #612. Some details have already been laid out there.

seladb:
I think I can take it

Goldziher:
@seladb wassup? any progress?

seladb:
Hi @Goldziher! I'm currently busy creating a new release for [PcapPlusPlus](https://github.com/seladb/PcapPlusPlus) (my other open-source project) so I don't have much time to contribute to Starlite. Then I'm planning to take some time off. If someone wants to take this issue please feel free to do so. Otherwise, I hope I'll be able to get back to it by the end of November

Goldziher:
hi, ok thanks. No I would like to get this done within the next week or so. So I will take this over.

Issue #634, opened 2022-10-21
Meta: Tracking issue for dropping Starlette


provinzkraut:
# Tracking issue for the current progress of dropping starlette dependencies

## How to use?

The following list can serve as an entry-point when tackling the remove of a specific starlette dependencies. It lists the names imported from starlette by their source module and all files that import it. 

It is autogenerated with the `ast` module, and will be updated accordingly.

## Overview
- [x] **`starlette.status`** (#626)

  <details><summary>Names</summary>

  - [x] `HTTP_400_BAD_REQUEST`

    <details><summary>References</summary>

    - [x] [examples/application_hooks/after_exception_hook.py](https://github.com/starlite-api/starlite/blob/main/examples/application_hooks/after_exception_hook.py)
    - [x] [starlite/exceptions/http_exceptions.py](https://github.com/starlite-api/starlite/blob/main/starlite/exceptions/http_exceptions.py)
    - [x] [tests/app/test_error_handling.py](https://github.com/starlite-api/starlite/blob/main/tests/app/test_error_handling.py)
    - [x] [tests/dependency_injection/test_http_handler_dependency_injection.py](https://github.com/starlite-api/starlite/blob/main/tests/dependency_injection/test_http_handler_dependency_injection.py)
    - [x] [tests/kwargs/test_cookie_params.py](https://github.com/starlite-api/starlite/blob/main/tests/kwargs/test_cookie_params.py)
    - [x] [tests/kwargs/test_header_params.py](https://github.com/starlite-api/starlite/blob/main/tests/kwargs/test_header_params.py)
    - [x] [tests/kwargs/test_layered_params.py](https://github.com/starlite-api/starlite/blob/main/tests/kwargs/test_layered_params.py)
    - [x] [tests/kwargs/test_path_params.py](https://github.com/starlite-api/starlite/blob/main/tests/kwargs/test_path_params.py)
    - [x] [tests/kwargs/test_query_params.py](https://github.com/starlite-api/starlite/blob/main/tests/kwargs/test_query_params.py)
    - [x] [tests/openapi/test_responses.py](https://github.com/starlite-api/starlite/blob/main/tests/openapi/test_responses.py)
    - [x] [tests/routing/test_path_resolution.py](https://github.com/starlite-api/starlite/blob/main/tests/routing/test_path_resolution.py)
    - [x] [tests/test_exception_handlers.py](https://github.com/starlite-api/starlite/blob/main/tests/test_exception_handlers.py)
[...]

PR #792, opened 2022-11-11
Add `CORS` middleware and support for `OPTIONS` requests

### LINKED ISSUE #754: Enhancement: Add Support for `OPTIONS` method
### Discussed in https://github.com/starlite-api/starlite/discussions/602

<sup>Originally posted by **Goldziher** October 15, 2022</sup>
Currently Starlite does not support `OPTIONS`.

For `OPTIONS` specifically we need to factor in the behaviour of the CORS middleware, which can in priniciple accept `OPTIONS` preflight requests. The Starlette CORSMiddleware implements this in such a way that preflight responses are returned for `OPTIONS` only when the header `Access-Control-Request-Method` is set. I am not certain this is the correct behaviour and we need to clarify this, we might need to either subclass the Starlette middleware or create our own version if it isn't the desired behaviour. 

Otherwise for `OPTIONS` I think we should probably resolve this with a special method that populates the `Allow` header. The question really is what values need to be returned from this method, which will determine where and how implementation should look.

Please add you thoughts. </div>

Goldziher:
# PR Checklist

This PR closes #754 and #669, it additionally closes the Starlette Meta tracking (#634) issue as complete.


starlite/config/cors.py
```diff
@@ -12,33 +16,155 @@ class CORSConfig(BaseModel):
...
+        """Ensure that the trusted hosts have correct domain wildcards.
+        Args:
+            value: A list of headers.
+
+        Returns:
+            A list of lower-cased headers.
+        """
+        return [v.lower() for v in value]
+
+    @property
```
jtraub, line 64: Since we are dropping 3.7 support can we use [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) instead?

starlite/config/cors.py
```diff
@@ -12,33 +16,155 @@ class CORSConfig(BaseModel):
...
+                        if not self.is_allow_all_methods
+                        else {"DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"}
+                    )
+                )
+            setattr(self, "_preflight_headers", headers)
+
+        return cast("Dict[str, str]", self._preflight_headers)
+
+    @property
+    def simply_headers(self) -> Dict[str, str]:
```
jtraub, line 145:
```suggestion
    def simple_headers(self) -> Dict[str, str]:
```

starlite/middleware/cors.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Returns:
+            An ASGI send function.
+        """
+
+        async def wrapped_send(message: "Message") -> None:
+            if message["type"] == "http.response.start":
+                message.setdefault("headers", [])
+                headers = MutableScopeHeaders.from_message(message=message)
+                headers.update(self.config.simply_headers)
```
jtraub, line 59:
```suggestion
                headers.update(self.config.simple_headers)
```

starlite/config/cors.py
jtraub: Typo, I guess, because you store computed value of this in `self._simple_headers`

starlite/config/cors.py
```diff
@@ -12,33 +16,155 @@ class CORSConfig(BaseModel):
...
+    _compiled_regex: Union[EmptyType, Pattern] = PrivateAttr(Empty)
+    _is_allow_all_origins: Union[EmptyType, bool] = PrivateAttr(Empty)
+    _is_allow_all_methods: Union[EmptyType, bool] = PrivateAttr(Empty)
+    _is_allow_all_headers: Union[EmptyType, bool] = PrivateAttr(Empty)
+    _preflight_headers: Union[EmptyType, Dict[str, str]] = PrivateAttr(Empty)
+    _simple_headers: Union[EmptyType, Dict[str, str]] = PrivateAttr(Empty)
+
+    @validator("allow_headers", always=True)
+    def validate_allow_headers(cls, value: List[str]) -> List[str]:  # pylint: disable=no-self-argument
+        """Ensure that the trusted hosts have correct domain wildcards.
```
jtraub, line 55: Either I am too tired after a stressful week or this docstring doesn't make sense because it mentions trusted hosts.

starlite/config/cors.py
```diff
@@ -12,33 +16,155 @@ class CORSConfig(BaseModel):
...
     expose_headers: List[str] = []
     """
-        List of headers that are exposed via the 'Access-Control-Expose-Headers' header.
+    List of headers that are exposed via the 'Access-Control-Expose-Headers' header.
     """
     max_age: int = 600
     """
-        Response aching TTL in seconds, defaults to 600.
-        Sets the 'Access-Control-Max-Age' header.
+    Response aching TTL in seconds, defaults to 600. Sets the 'Access-Control-Max-Age' header.
```
peterschutt, line 47:
```suggestion
    Response caching TTL in seconds, defaults to 600. Sets the 'Access-Control-Max-Age' header.
```

starlite/config/cors.py
peterschutt: There are potentially a lot of places we can use cached_property throughout the code base now.

starlite/middleware/cors.py
```diff
@@ -0,0 +1,70 @@
...
+
+        origin = headers.get("origin")
+
+        if not origin:
+            await self.app(scope, receive, send)
+        else:
+            await self.app(scope, receive, self.send_wrapper(send=send, origin=origin, has_cookie="cookie" in headers))
+
+    def send_wrapper(self, send: "Send", origin: str, has_cookie: bool) -> "Send":
+        """Wrap 'send' to ensure that state is not disconnected.
```
peterschutt, line 46:
```suggestion
        """Wrap `send` to ensure that state is not disconnected.
```

starlite/routes/http.py
```diff
@@ -232,3 +241,82 @@ async def _set_cached_response(
...
+                    if cors_config.is_allow_all_headers:
+                        response_headers["Access-Control-Allow-Headers"] = ", ".join(
+                            sorted(set(pre_flight_requested_headers) | DEFAULT_ALLOWED_CORS_HEADERS)
+                        )
+                    elif any(
+                        header.lower() not in cors_config.allow_headers for header in pre_flight_requested_headers
+                    ):
+                        failures.append("headers")
+
+                return (
```
peterschutt, line 300: I think this would read nicer without the ternary, e.g.,

```python
If failures:
    return Response(...)
return Response(...)
```

tests/middleware/test_cors_middleware.py
```diff
@@ -0,0 +1,75 @@
...
+    assert cors_config.expose_headers == []
+
+    with create_test_client(route_handlers=[], cors_config=cors_config) as client:
+        unpacked_middleware = []
+        cur = client.app.asgi_handler
+        while hasattr(cur, "app"):
+            unpacked_middleware.append(cur)
+            cur = cast("Any", cur.app)
+        else:
+            unpacked_middleware.append(cur)
```
peterschutt, line 28: Maybe a utility for unpacking middleware as this is the 2nd use of the pattern.

jtraub (approved):

peterschutt (approved):

PR #798, opened 2022-11-12
Cleanup docs examples

provinzkraut:
This is part of the documentation overhaul project, aimed at starting the process to solve #343 in particular. It does a few things

- Add a small CLI utility to find and extract code examples in the docs, with the added option to scaffold tests for them
- Extract examples from and add tests for:
  - Path params
  - Query params
  - Header and cookie params
  - Layered params
  - Returning responses
  - Response headers
  - Response cookies
  - Streaming responses
- Rewrite the "Query params" section (because it was hard to make succinct, self-contained examples for it)

# PR Checklist


provinzkraut:
Opening this for review now, since I don't want it to get too big. I'm gonna follow up with a few more PRs, so this is going to be a sequential process.

examples/responses/response_cookies_4.py
```diff
@@ -0,0 +1,40 @@
...
+def after_request_handler(response: Response) -> Response:
+    response.set_cookie(**Cookie(key="Random-Cookie", value=str(randint(1, 100))).dict)
+    return response
+
+
+router = Router(
+    path="/router-path",
+    route_handlers=[retrieve_resource],
+    after_request=after_request_handler,
+    response_cookies=[
```
jtraub, line 31: @provinzkraut could you please check. It looks like a real bug.

Adding `response_cookies` param to the router makes Starlite to emit `set-cookie` header twice. In this case second header has empty value.

tools/doc_examples.py
```diff
@@ -0,0 +1,172 @@
...
+
+    parser_check.set_defaults(func=check_command)
+    parser_extract.set_defaults(func=extract_command)
+
+    args = parser.parse_args()
+    args.func(args)
+
+
+if __name__ == "__main__":
+    cli()
```
Goldziher, line 172: can you document this cli ?

Goldziher (approved):

tools/doc_examples.py
provinzkraut: WDYM? Or rather, where do you think it should be documented?

examples/responses/response_cookies_4.py
provinzkraut: I `xfaile`ed this test for now.

PR #805, opened 2022-11-14
Improve performance

Goldziher:
# PR Checklist


starlite/kwargs.py
```diff
@@ -432,16 +510,14 @@ async def resolve_dependency(
             The resolved dependency value
         """
         if dependency.provide.cache_per_request:
-            async with dependency.provide.lock:
```
ottermata, line 435: Without the lock the `cache_per_request` config won't work as expected (at least the way I imagined it).
Since I ended up not needing this, we could just remove that functionality completely.

Coming from FastAPI, the current default may be confusing though. Their behavior is to cache dependencies per request, but resolve them sequentially. This way they don't need a lock, but if someone has multiple async dependencies they won't run concurrently.

Maybe a better way would be to collect all required dependencies first, batch them into groups based on their sub-dependencies and then sequentially resolve the batches?
An example:
Dependencies A and B have no dependencies, C and D both require A, E requires B and C.
Resolve A and B in the first batch concurrently, then resolve C and D in the second batch (also  concurrently) and finally resolve E.
B could also be resolved in the second batch, which would require one less TaskGroup.

Goldziher: so the issue why I removed this - locks are apparently extremely memory intensive to create and save. We need to come up with a better strategy to handle this that does not impact resources like this. Whatever we gain from async resolution using a task group, we are losing form the overhead of this.

jtraub: @Goldziher can we do concurrent resolution ONLY if user explicitly requests it?

like xpresso does - https://xpresso-api.dev/0.42.3/advanced/dependencies/performance/#concurrent-execution

ottermata: @Goldziher I reworked the way dependencies are resolved as per my suggestion above: https://github.com/ottermata/starlite/commit/dd229a0d06d4a3289319631c141899ab74afaea4

The main difference with this approach is that all dependencies only ever get called once per request. I can't really imagine a reason why someone would want to have them executed multiple times, so I think this is fine.

Can you run you benchmarks with that change included? I can open a PR if needed.

Goldziher: Sure, open a pr

provinzkraut: @ottermata Can you rebase against the `performance_updates` branch in a separate branch of yours? Then I can benchmark the three.

ottermata: @provinzkraut Here's the branch: https://github.com/ottermata/starlite/tree/performance_updates/rework_dependencies

starlite/asgi/routing_trie/mapping.py
```diff
@@ -16,11 +17,51 @@
...
+        root_node: The root trie node.
+        route: The route that is being added.
+
+    Returns:
+        A trie node.
+    """
+    if route.path_parameters:
+        raise ImproperlyConfiguredException("Path parameters cannot be configured for a static path.")
+
+    # we need to ensure that we can traverse the map both view the full path key, e.g. "/my-route/sub-path" and
```
Goldziher, line 40:
```suggestion
    # we need to ensure that we can traverse the map both through the full path key, e.g. "/my-route/sub-path" and
```

starlite/parsers.py
```diff
@@ -59,3 +61,20 @@ def parse_cookie_string(cookie_string: str) -> Dict[str, str]:
...
+
+@lru_cache
+def parse_query_string(query_string: bytes) -> Tuple[Tuple[str, Any], ...]:
+    """Parse a query string into a list of tuples.
+
+    Args:
+        query_string: A query string.
+
+    Returns:
+        A MultiDict instance
```
Goldziher, line 74:
```suggestion
        A tuple of key value pairs. 
```

starlite/parsers.py
```diff
@@ -59,3 +61,20 @@ def parse_cookie_string(cookie_string: str) -> Dict[str, str]:
     for k, v in filter(lambda x: x[0] or x[1], ((k.strip(), v.strip()) for k, v in cookies)):
         output[k] = unquote(unquote_cookie(v))
     return output
+
+
+@lru_cache
+def parse_query_string(query_string: bytes) -> Tuple[Tuple[str, Any], ...]:
+    """Parse a query string into a list of tuples.
```
Goldziher, line 68:
```suggestion
    """Parse a query string into a tuple of key value pairs.
```

starlite/response/base.py
```diff
@@ -240,9 +235,8 @@ def content_length(self) -> Optional[int]:
             return len(self.body)
         return None
 
-    @property
-    def encoded_headers(self) -> List[Tuple[bytes, bytes]]:
-        """Response headers as a list of tuples of bytes.
+    def encode_headers(self) -> List[Tuple[bytes, bytes]]:
+        """Encode the response headers as a list of tuples of bytes.
```
Goldziher, line 239:
```suggestion
        """Encode the response headers as a list of byte tuples.
```

starlite/kwargs.py
ottermata: @provinzkraut I rebased against the current version

starlite/datastructures/provide.py
```diff
@@ -15,43 +14,47 @@
 class Provide:
     """A wrapper class for dependency injection."""
 
-    __slots__ = ("dependency", "use_cache", "cache_per_request", "cache_key", "lock", "value", "signature_model")
+    __slots__ = (
+        "dependency",
+        "use_cache",
+        "lock",
```
ottermata, line 20: Provide.lock does not exists anymore.
```suggestion
```

ottermata (approved):

PR #806, opened 2022-11-14
Resolve forward refs on signature models.

### LINKED ISSUE #802: Enhancement: resolve forward refs when producing signature models
We need to support where users declare `from __future__ import annotations` in modules where handlers and dependencies are declared.

peterschutt:
Passes name of module where the handler/dependency is defined to `create_model()`

Closes #802 

# PR Checklist

### DIFF
--- a/starlite/signature.py
+++ b/starlite/signature.py
@@ -1,4 +1,4 @@
-from inspect import Parameter, Signature
+from inspect import Parameter, Signature, getmodule
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
 
 UNDEFINED_SENTINELS = {Undefined, Signature.empty}
 SKIP_NAMES = {"self", "cls"}
-SKIP_VALIDATION_NAMES = {"request", "socket", "state", "scope"}
+SKIP_VALIDATION_NAMES = {"request", "socket", "state", "scope", "receive", "send"}
 
 
 class SignatureModel(BaseModel):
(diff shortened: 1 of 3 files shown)


starlite/signature.py
```diff
@@ -41,7 +41,7 @@
 
 UNDEFINED_SENTINELS = {Undefined, Signature.empty}
 SKIP_NAMES = {"self", "cls"}
-SKIP_VALIDATION_NAMES = {"request", "socket", "state", "scope"}
+SKIP_VALIDATION_NAMES = {"request", "socket", "state", "scope", "receive", "send"}
```
peterschutt, line 44: This isn't necessary for the PR, but curious on thoughts on whether we should be validating these?

starlite/signature.py
```diff
@@ -194,7 +195,8 @@ def __init__(self, fn: "AnyCallable", plugins: List["PluginProtocol"], dependenc
         if fn is None:
             raise ImproperlyConfiguredException("Parameter `fn` to `SignatureModelFactory` cannot be `None`.")
         self.signature = Signature.from_callable(fn)
-        self.fn_name = fn.__name__ if hasattr(fn, "__name__") else "anonymous"
+        self.fn_name = getattr(fn, "__name__", "anonymous")
+        self.fn_module_name = getattr(fn, "__module__", "pydantic.main")
```
peterschutt, line 199: Using `"pydantic.main"` as the default here preserves the current behaviour if there is no `__module__` attribute. We could similarly pass `__name__` so that `starlite.signature` is the module, but it's really just for the resolution to get access to global names so not sure that it matters a great deal.

starlite/signature.py
Goldziher: how would we?

Goldziher (approved):

PR #815, opened 2022-11-17
Rework dependencies

ottermata:
Dependencies currently have two problems:
1. They are resolved sequentially
2. They may be resolved multiple times within a single request

In order to solve the first without introducing locks, the resolution algorithm was changed to:
1. Recursively getting a list of all dependencies, with their required sub-dependencies
2. Creating a sequence of dependency batches. The dependencies per batch can run concurrently, but depend on values from the previous batches. Each batch is created by:
  a. Getting all unaccounted for dependencies without remaining sub-dependencies
  b. Removing those dependencies from the list of overall dependencies
  c. Removing them from the remaining sub-dependencies of the other dependencies

This way the following dependency graph will be split into these batches: `(A, B, C), (D, E), (F)`
```
┌─F─┐
│   │
▼   │
D   │   E
│   │   │
▼   ▼   ▼
A   B   C
```
This is done once when the KwargsModel is created. When actually resolving the dependencies, we only need to iterate over the batches and resolve the the dependencies without checking for sub-dependencies. In order to avoid unnecessary overhead we only create the task group when there is more than one dependency per batch.

This also solves the second problem, all dependencies are only called once per request. If in the above graph F were to also require A or C, nothing would change. If E were to require D, it would only move to F's batch.

# PR Checklist

### DIFF
--- a/starlite/datastructures/provide.py
+++ b/starlite/datastructures/provide.py
@@ -17,7 +17,6 @@ class Provide:
     __slots__ = (
         "dependency",
         "use_cache",
-        "lock",
         "value",
         "signature_model",
         "sync_to_thread",
--- a/starlite/kwargs.py
+++ b/starlite/kwargs.py
@@ -16,6 +16,7 @@ from typing import (
     cast,
 )
 
+from anyio import create_task_group
 from pydantic.fields import (
     SHAPE_DEQUE,
     SHAPE_FROZENSET,
@@ -74,6 +75,13 @@ class Dependency:
         self.key = key
         self.provide = provide
         self.dependencies = dependencies
+
+    def __eq__(self, other: Any) -> bool:
+        # check if memory address is identical, otherwise compare attributes
+        return other is self or (isinstance(other, self.__class__) and other.key == self.key)
+
+    def __hash__(self) -> int:
+        return hash(self.key)
 
 
 def merge_parameter_sets(first: Set[ParameterDefinition], second: Set[ParameterDefinition]) -> Set[ParameterDefinition]:
@@ -153,6 +161,67 @@ def create_query_default_dict(
     return output
 
 
+async def resolve_dependency(
+    dependency: "Dependency", connection: Union["WebSocket", "Request"], kwargs: Dict[str, Any]
+) -> None:
+    """Resolve a given instance of [Dependency][starlite.kwargs.Dependency]. All required sub dependencies must already
+    be resolved into the kwargs. The result of the dependency will be stored in the kwargs.
+
+    Args:
+        dependency: An instance of [Dependency][starlite.kwargs.Dependency]
+        connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
+        kwargs: Any kwargs to pass to the dependency, the result will be stored here as well.
+    """
+    signature_model = get_signature_model(dependency.provide)
+    dependency_kwargs = signature_model.parse_values_from_connection_kwargs(connection=connection, **kwargs)
+    kwargs[dependency.key] = await dependency.provide(**dependency_kwargs)
+
+
+def create_dependency_batches(expected_dependencies: Set[Dependency]) -> List[Set[Dependency]]:
+    """Calculate batches for all dependencies, recursively.
+
+    Args:
+        expected_dependencies: A set of all direct [Dependencies][starlite.kwargs.Dependency].
+
+    Returns:
+        A list of batches.
+    """
+    dependencies_to: Dict[Dependency, Set[Dependency]] = {}
+    for dependency in expected_dependencies:
+        if dependency not in dependencies_to:
+            map_dependencies_recursively(dependency, dependencies_to)
+
+    batches = []
+    while dependencies_to:
+        current_batch = {
+            dependency
+            for dependency, remaining_sub_dependencies in dependencies_to.items()
+            if not remaining_sub_dependencies
+        }
+
+        for dependency in current_batch:
+            del dependencies_to[dependency]
+            for others_dependencies in dependencies_to.values():
+                others_dependencies.discard(dependency)
+
+        batches.append(current_batch)
+
+    return batches
+
+
+def map_dependencies_recursively(dependency: Dependency, dependencies_to: Dict[Dependency, Set[Dependency]]) -> None:
+    """Recursively map dependencies to their sub dependencies.
+
+    Args:
+        dependency: The current dependency to map.
+        dependencies_to: A map of dependency to its sub dependencies.
+    """
+    dependencies_to[dependency] = set(dependency.dependencies)
+    for sub in dependency.dependencies:
+        if sub not in dependencies_to:
+            map_dependencies_recursively(sub, dependencies_to)
+
+
 class KwargsModel:
     """Model required kwargs for a given RouteHandler and its dependencies.
 
@@ -160,8 +229,8 @@ class KwargsModel:
     """
 
     __slots__ = (
+        "dependency_batches",
         "expected_cookie_params",
-        "expected_dependencies",
         "expected_form_data",
         "expected_header_params",
         "expected_path_params",
@@ -190,17 +259,16 @@ class KwargsModel:
 
         Args:
             expected_cookie_params: Any expected cookie parameter kwargs
-            expected_dependencies:  Any expected dependency kwargs
+            expected_dependencies: Any expected dependency kwargs
             expected_form_data: Any expected form data kwargs
             expected_header_params: Any expected header parameter kwargs
             expected_path_params: Any expected path parameter kwargs
-            expected_query_params:  Any expected query parameter kwargs
+            expected_query_params: Any expected query parameter kwargs
             expected_reserved_kwargs: Any expected reserved kwargs, e.g. 'state'
             sequence_query_parameter_names: Any query parameters that are sequences
             is_data_optional: Treat data as optional
         """
         self.expected_cookie_params = expected_cookie_params
-        self.expected_dependencies = expected_dependencies
         self.expected_form_data = expected_form_data
         self.expected_header_params = expected_header_params
         self.expected_path_params = expected_path_params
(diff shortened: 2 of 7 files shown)


starlite/datastructures/provide.py
```diff
@@ -15,30 +13,24 @@
...
+    __slots__ = ("dependency", "use_cache", "value", "signature_model")
 
     def __init__(
         self,
         dependency: "AnyCallable",
         use_cache: bool = False,
-        cache_per_request: bool = False,
-        cache_key: Optional[str] = None,
         sync_to_thread: bool = False,
+        **_kwargs: Any,
```
Goldziher, line 23: what are kwargs used for?

starlite/kwargs.py
```diff
@@ -409,39 +418,94 @@ def _collect_params(params: Mapping[str, Any], expected: Set[ParameterDefinition
...
+        kwargs: Dict[str, Any],
+    ) -> None:
+        """Resolve the dependencies concurrently. The passed in kwargs will be modified with the results.
+
+        Args:
+            dependencies: An iterable for instances of [Dependency][starlite.kwargs.Dependency].
+            connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
+            kwargs: Kwargs to pass to dependencies
+        """
+        if len(dependencies) == 0:
```
Goldziher, line 444:
```suggestion
        if not dependencies:
```

starlite/kwargs.py
```diff
@@ -409,39 +418,94 @@ def _collect_params(params: Mapping[str, Any], expected: Set[ParameterDefinition
...
+        Args:
+            connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
+            kwargs: Kwargs to pass to dependencies
+        """
+        for batch in self.dependency_batches:
+            await self._resolve_dependencies_concurrently(batch, connection, kwargs)
+
+    async def _resolve_dependencies_concurrently(
+        self,
+        dependencies: Collection["Dependency"],
```
Goldziher, line 433: collection is too generic. This is not a collection, its a list or an iteratabe.

starlite/kwargs.py
```diff
@@ -409,39 +418,94 @@ def _collect_params(params: Mapping[str, Any], expected: Set[ParameterDefinition
...
+        Args:
+            dependency: An instance of [Dependency][starlite.kwargs.Dependency]
+            connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
+            kwargs: Any kwargs to pass recursively.
+        """
         signature_model = get_signature_model(dependency.provide)
-        await resolve_dependencies_concurrently(self, dependency.dependencies, connection, kwargs)
         dependency_kwargs = signature_model.parse_values_from_connection_kwargs(connection=connection, **kwargs)
-        return await dependency.provide(**dependency_kwargs)
+        kwargs[dependency.key] = await dependency.provide(**dependency_kwargs)
```
Goldziher, line 467: dependency.provide is not always async. Its important to note this - this is configured by the user.

ottermata: Not sure I understand, Provide.__call__ is async?

starlite/datastructures/provide.py
ottermata: Thought this might be better than just removing the `cache_per_request` and `cache_key` options. I can remove this if that's not a concern.

peterschutt:
It would be nice if you could provide some context to the PR. E.g., a high level description of the changes you are making to the algorithm and why (i.e., what problem is it solving?). It would make this a lot easier to review.

ottermata:
@peterschutt I updated the description

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
-    ) -> Any:
-        """Given an instance of [Dependency][starlite.kwargs.Dependency], recursively resolves its dependency graph.
+    async def resolve_dependencies(self, connection: Union["WebSocket", "Request"], kwargs: Dict[str, Any]) -> None:
+        """Resolve all dependencies into the kwargs, recursively.
 
         Args:
-            dependency: An instance of [Dependency][starlite.kwargs.Dependency]
             connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
-            **kwargs: Any kwargs to pass recursively.
+            kwargs: Kwargs to pass to dependencies
```
Goldziher, line 516:
```suggestion
            **kwargs: Kwargs to pass to dependencies
```

This is the correct format. You can test for doc issues by running `docker compose up --build`, which will bring the docs up locally.

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
+        """
+        for batch in self.dependency_batches:
+            await self._resolve_dependencies_concurrently(batch, connection, kwargs)
 
-        Returns:
-            The resolved dependency value
+    async def _resolve_dependencies_concurrently(
```
Goldziher, line 521: i think most of these methods can be moved into functions from the class. There is no point in using self here.

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
+    ) -> None:
+        """Resolve the dependencies concurrently. The passed in kwargs will be modified with the results.
+
+        Args:
+            dependencies: An iterable for instances of [Dependency][starlite.kwargs.Dependency].
+            connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
+            kwargs: Kwargs to pass to dependencies
+        """
+        if not dependencies:
+            pass
```
Goldziher, line 535: this should never be the case - this function should not be called at all if dependencies is empty.

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
+        elif len(dependencies) == 1:
+            await self._resolve_dependency(next(iter(dependencies)), connection, kwargs)
```
Goldziher, line 537: I dont think this should be inside thi function - it should be outside it. This function should only include whats now under `else`, and be called only when there are multiple depenendencies to resolve concurrently.

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
+        else:
+            async with create_task_group() as task_group:
+                for dependency in dependencies:
+                    task_group.start_soon(self._resolve_dependency, dependency, connection, kwargs)
+
+    @staticmethod
+    async def _resolve_dependency(
```
Goldziher, line 544: this shouldnt be a private static method. Instead it should be the contents of `resolve_dependency` itself, or something else (see my ending comment.

starlite/kwargs.py
```diff
@@ -499,26 +508,94 @@ def to_kwargs(self, connection: Union["WebSocket", "Request"]) -> Dict[str, Any]
...
-        for sub_dependency in dependency.dependencies:
-            kwargs[sub_dependency.key] = await self.resolve_dependency(
-                dependency=sub_dependency, connection=connection, **kwargs
-            )
         dependency_kwargs = signature_model.parse_values_from_connection_kwargs(connection=connection, **kwargs)
-        return await dependency.provide(**dependency_kwargs)
+        kwargs[dependency.key] = await dependency.provide(**dependency_kwargs)
+
+    @classmethod
+    def _get_dependency_batches(cls, expected_dependencies: Set[Dependency]) -> List[Set[Dependency]]:
```
Goldziher, line 560: same as previous, these two class methods should be moved out of the class. 

Regarding names:

instead of `_get_dependency_batches` --> `batch_dependencies`, `split_dependencies_into_batches`, `create_dependency_batches etc.` 

`_get_dependency_batches_helper ` --> this name is meaningless. Try ` map_dependencies_recurisvely` or something like this.

tests/dependency_injection/test_request_local_caching.py
```diff
@@ -0,0 +1,34 @@
+from starlite import Provide, create_test_client, get
+from starlite.status_codes import HTTP_200_OK
+
+
+def test_caching_per_request() -> None:
```
Goldziher, line 5: I thought dependenc caching is gone 😉 , in fact i know its gone. 

I think you should probably rebase and remove this.

What I am not seeing-- tests that ensure the functionality you are adding works as expected. And I would like to run benchmarks to ensure that this improves performance and doesn't cause memory issues.

Goldziher (commented):
Interesting contribution! I would like to be a bit more cautious with this. i am not sure what the performance gain of using a task group here gives us - it adds complexity and we need to ascertain this. I am sure @provinzkraut can help us with testing DI in some way. 

The other comment I have is more general - if you look at the logic in `kwargs.py` on main, you will see that i added `extractors` there. There is a function called `create_connection_value_extractor`, which is used to creat a precomputed extractor for data for different types. I think we should opt in to precompute a function that handles dependencies for a given kwargs model in advance, not repeat the calculation on each request.

starlite/kwargs.py
ottermata: I don't think the ** is correct? The kwargs are passed in as a dictionary, so that they can be mutated.

tests/dependency_injection/test_request_local_caching.py
ottermata: Dependency "caching" gets reintroduced (every dependency is only resolved once), so I reused the test. Maybe change the name to be more clear?

I added a test for the batching.

ottermata:
> Interesting contribution! I would like to be a bit more cautious with this. i am not sure what the performance gain of using a task group here gives us - it adds complexity and we need to ascertain this. I am sure @provinzkraut can help us with testing DI in some way.

Feel free to test the performance of this, I rebased against main.

> The other comment I have is more general - if you look at the logic in `kwargs.py` on main, you will see that i added `extractors` there. There is a function called `create_connection_value_extractor`, which is used to creat a precomputed extractor for data for different types. I think we should opt in to precompute a function that handles dependencies for a given kwargs model in advance, not repeat the calculation on each request.

The dependency batches are stored on the KwargsModel, just like the extractors. So I assume they should be cached just the same? I refactored the code to create the batches within the constructor, makes it a lot simpler as well.

PR #820, opened 2022-11-18
Support iterable headers in the scope/message

### LINKED ISSUE #813: Wrapping NewRelic around Starlite app breaks handlers with path parameters
Ask your question...
I tried to integrate starlite with newrelic 

```python
import os
from starlite import Starlite
import newrelic.agent

from routes import home

newrelic.agent.initialize('newrelic.ini')

app = Starlite(route_handlers=[home])
newrelic.agent.ASGIApplicationWrapper(app)
```

but getting error
`CIMultiDict keys should be either str or subclasses of str`

could you guys help me why this happen?

jtraub:
# PR Checklist

Our code assumed that `headers` in `scope` and `message` will be `List[Tuple[byte, byte]]` while the [spec says](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) it should be an `iterable`. asgiref typing follows the spec - https://github.com/django/asgiref/blob/main/asgiref/typing.py#L62 so we should do the same.
In this PR I change typing and make sure we can process not only lists but iterables.

Tests use a tuple as alternative because judging by the source of dozens of ASGI apps people either throw a list or a tuple.

This PR closes #813

### DIFF
--- a/starlite/datastructures/headers.py
+++ b/starlite/datastructures/headers.py
@@ -183,7 +183,7 @@ class MutableScopeHeaders(MutableMapping):
         """
         existing = self.get(key)
         if existing is not None:
-            value = ", ".join([*existing.split(","), value])
+            value = ",".join([*existing.split(","), value])
         self[key] = value
 
     def __getitem__(self, key: str) -> str:
--- a/starlite/datastructures/headers.py
+++ b/starlite/datastructures/headers.py
@@ -24,7 +24,12 @@ from starlite.datastructures.multi_dicts import MultiMixin
 from starlite.exceptions import ImproperlyConfiguredException
 
 if TYPE_CHECKING:
-    from starlite.types.asgi_types import HeaderScope, Message, RawHeadersList
+    from starlite.types.asgi_types import (
+        HeaderScope,
+        Message,
+        RawHeaders,
+        RawHeadersList,
+    )
 
 ETAG_RE = re.compile(r'([Ww]/)?"(.+)"')
 
@@ -41,7 +46,7 @@ class Headers(CIMultiDictProxy[str], MultiMixin[str]):
             libs.org/en/stable/multidict.html#cimultidictproxy).
     """
 
-    def __init__(self, headers: Optional[Union[Mapping[str, str], "RawHeadersList", MultiMapping]] = None) -> None:
+    def __init__(self, headers: Optional[Union[Mapping[str, str], "RawHeaders", MultiMapping]] = None) -> None:
         """Initialize `Headers`.
 
         Args:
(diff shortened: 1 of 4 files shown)


peterschutt (approved):

starlite/datastructures/headers.py
```diff
@@ -41,18 +46,20 @@ class Headers(CIMultiDictProxy[str], MultiMixin[str]):
...
-            if isinstance(headers, list):
-                headers_ = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in headers]
-            elif headers:
-                headers_ = headers
+            if headers:
+                if isinstance(headers, Mapping):
+                    headers_ = headers  # pyright: ignore
+                else:
+                    headers_ = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in headers]
+
```
provinzkraut, line 62: I think this whole section should be simplified to

```python
        if headers and not isinstance(headers, MultiMapping):
            headers_: Union[Mapping[str, str], List[Tuple[str, str]]] = {}
            if isinstance(headers, Mapping):
                headers_ = headers  # pyright: ignore
            else:
                headers_ = [(key.decode("latin-1"), value.decode("latin-1")) for key, value in headers]

            super().__init__(CIMultiDict(headers_))
        else:
            super().__init__(headers)
```

provinzkraut (approved):
Nice!

starlite/datastructures/headers.py
jtraub: @provinzkraut uhm, not sure that we can do this replacement. If `headers` is None in your code we will invoke `super().__init__(None)` while the original code will perform `super().__init__(CIMultiDict({}))` in this case.

I might be wrong though as I am writing from my phone

provinzkraut: `headers and not isinstance(headers, MultiMapping)` is `False` when `headers` is `None`, which means it will go directly to `super().__init__(headers)` (which can handle `None`).

jtraub: No, it can not handle `None`. Try it yourself

```
> python
...
>>> from starlite.datastructures.headers import Headers
>>> Headers(None)
  File "/home/jtraub/dev/starlite/starlite/datastructures/headers.py", line 65, in __init__
    super().__init__(headers)
TypeError: ctor requires CIMultiDict or CIMultiDictProxy instance, not <class 'NoneType'>
```

provinzkraut: Oh. Yea, I guess that makes sense.. :upside_down_face:

jtraub:
I don't have time this week to adapt the PR

Issue #827, opened 2022-11-18
Issue With Conflicting Path Parameters


peterschutt:
### Discussed in https://github.com/starlite-api/starlite/discussions/825

<div type='discussions-op-text'>

<sup>Originally posted by **stonedprogrammer** November 19, 2022</sup>
Hello,

I didn't want to submit an issue as I wasn't sure if this was an issue on my end and I couldn't find anything related in discussions nor issues. I keep getting the following error:
```
starlite.exceptions.http_exceptions.ImproperlyConfiguredException: 500: Should not use routes with conflicting path parameters
```

with the following controller:
```
class VmcController(Controller):
    path = "/vmc"

    @delete("/{name:str}")
    async def delete_vm(self, name: str) -> None:
        return

    @post("/{id:int}")
    async def restart_vm(self, id: int) -> str:
        return "string"
```

Based on https://starlite-api.github.io/starlite/usage/1-routing/3-controllers/?h=contr, this should be fine as the path + http method combination is unique for both of them, yet each time I get that same error. When I keep the delete route by itself, it works, and vice versa for the post route. Am I doing something incorrectly here?

I have been able to fix it by adding another path level (e.g. changing the delete path to `/delete/{name:str}`, but this doesn't really mesh with the rest of the API routing.</div>

peterschutt:
TL;DR - this can work if the parameters on any path level are the same, i.e. if they were both `{id:int}` or `{name:str}`.

Whether this turns out to be a documentation issue, or something we actually try to resolve via changes to the routing would probably need to come from @Goldziher. However, I look at that example and think we should probably strive to make it work, but I'd need to spend more time trying to understand the feasibility of that.

Path parameters are represented by a single node in the routing trie at any level, e.g., the routing trie for this set of handlers ends up looking like this, prior to the point that the validation raises the error:

```
.
└── /
    └── vmc/
        └── {path parameter}
            ─ DELETE
            ─ OPTIONS
            ─ POST
```
The POST and OPTIONS routes of the 2nd handler _are_ added to the node's "asgi handlers" mapping, but the OPTIONS route of the 2nd handler overwrites that of the first, and nonetheless, immediately thereafter the validation error is raised because the path parameter of the 2nd route is not equal to the path parameter that seeded that trie node (where equality is defined as name ("id", "name"), full name ("name:str", "id:int") and type (`str`, `int`) all being equal).

So if you could define these routes such that they both accepted the same parameter it will work.

[...]

PR #830, opened 2022-11-20
Define routing trie path parameter nodes by type.

peterschutt:
For discussion around #825 and #827.

Main differences are:
- uses the type of the path parameter as the child node key
- tracks type of path parameter child nodes on current node
- uses ability to cast path component to param type to route

# PR Checklist


starlite/asgi/routing_trie/mapping.py
```diff
@@ -54,8 +54,10 @@ def add_map_route_to_trie(
                 if component.type is Path:
                     current_node["is_path_type"] = True
                     break
-
-                next_node_key: Union[Type[PathParameterSentinel], str] = PathParameterSentinel
+                # if component.type in current_node["children"]:
+                #     raise ImproperlyConfiguredException("Cannot have adjacent path parameter nodes of the same type.")
```
peterschutt, line 58: I think this would be correct behavior, but if enabled there are a lot of test errors.

starlite/asgi/routing_trie/traversal.py
```diff
@@ -106,16 +103,21 @@ def traverse_route_map(
...
+        else:
+            path_params.append(component)  # type: ignore[arg-type]
+
+            return traverse_route_map(
+                current_node=current_node["children"][type_],
+                path=path,
+                path_components=path_components,
+                path_params=path_params,
+                scope=scope,
+            )
```
peterschutt, line 120: I'm thinking performance will be a thing here, but just wanted to get a POC going.

tests/app/test_error_handling.py
```diff
@@ -17,6 +19,7 @@ def my_route_handler(param: int, data: Person) -> None:
         assert len(extra) == 3
 
 
+@pytest.mark.xfail()
```
peterschutt, line 22: In this test, we have route `/{param:int}` and request `/abc`, then test for 400. I think that this should be 404.

tests/kwargs/test_path_params.py
```diff
@@ -127,9 +127,9 @@ def test_method() -> None:
         ["float", float, 1.01],
         ["uuid", UUID, uuid4()],
         ["decimal", Decimal, Decimal("1.00001")],
-        ["date", date, date.today().isoformat()],
-        ["datetime", datetime, datetime.now().isoformat()],
-        ["timedelta", timedelta, timedelta(days=1).total_seconds()],
+        # ["date", date, date.today().isoformat()],
+        # ["datetime", datetime, datetime.now().isoformat()],
+        # ["timedelta", timedelta, timedelta(days=1).total_seconds()],
         ["path", Path, "/1/2/3/4/some-file.txt"],
```
peterschutt, line 133: I think we should discuss whether we need `date`, `datetime` and `timedelta` as path parameter types.

As we say in the docs:
> ...as long as parameter inside the function declaration is typed with a "higher" type to which the lower type can be coerced, this is fine

Whenever these are used the type is declared in the handler/dependency signature and pydantic does the parsing anyway.

We could avoid making the change breaking by aliasing them to `str`.

tests/routing/test_path_resolution.py
```diff
@@ -127,25 +126,25 @@ def handler_fn(some_id: int = 1) -> str:
...
-def test_path_validation(handler_path: str, request_path: str, expected_status_code: int) -> None:
+def test_path_validation(handler_path: str, request_path: str) -> None:
     @get(handler_path)
     def handler_fn(**kwargs: Any) -> None:
         ...
 
     with create_test_client(handler_fn) as client:
         response = client.get(request_path)
-        assert response.status_code == expected_status_code
+        assert response.status_code == HTTP_404_NOT_FOUND
```
peterschutt, line 147: I think all these should really be 404, thoughts?

peterschutt:
Hey @starlite-api/maintainers, only draft, but I'm more interested in discussing the concept rather than code review at this point.

I've put some comments in where I think discussion is relevant, but feel free to start your own..

Goldziher:
Can you recreate this based on main?

peterschutt:
> Can you recreate this based on main?

yes will do

peterschutt:
Closed in favor of #839

PR #839, opened 2022-11-22
Path param type discrimination

peterschutt:
# PR Checklist

For #825 and #827


starlite/asgi/routing_trie/types.py
```diff
@@ -69,25 +72,49 @@ class RouteTrieNode:
...
+    Args:
+        parent_node: The parent of the new node, if not root.
+        component: The path component that the node is to represent, or `None` if root.
+
     Returns:
         A route map node instance.
     """
+    parent_path = ("" if parent_node is None else parent_node.path).rstrip("/")
+    node_path = f"{{{component.full}}}" if isinstance(component, PathParameterDefinition) else component or ""
+    path = "/".join((parent_path, node_path))
```
peterschutt, line 104: I added the path of the current node so that I could use it to generate an error message but it might be useful elsewhere also.

starlite/routes/asgi.py
```diff
@@ -27,12 +27,13 @@ def __init__(
...
             route_handler: An instance of [ASGIRouteHandler][starlite.handlers.asgi.ASGIRouteHandler].
         """
-        self.route_handler = route_handler
         super().__init__(
             path=path,
             scope_type=ScopeType.ASGI,
             handler_names=[unwrap_partial(route_handler.handler_name)],
         )
+        self.route_handler = route_handler
+        self.is_mount = route_handler.is_mount
```
peterschutt, line 36: The alternative to defining `is_mount` here and on `BaseRoute` was this line in `add_route_to_trie`:

```python
is_mount = hasattr(route, "route_handler") and getattr(route.route_handler, "is_mount", False) # pyright: ignore
```
Just thought it cleaned it up a bit.

starlite/asgi/routing_trie/mapping.py
```diff
@@ -56,12 +52,40 @@ def add_mount_route(
...
+            identical to `param_def`.
+    """
+    child_param_with_same_type = current_node.child_path_parameter_type_map.get(param_def.type)
+    if child_param_with_same_type:
+        if param_def != child_param_with_same_type:
+            existing_node = current_node.children[child_param_with_same_type]
+            raise ImproperlyConfiguredException(
+                f"Path parameter '{{{param_def.full}}}' for path '{route.path}' conflicts with parameter "
+                f"path '{existing_node.path}'."
+            )
```
peterschutt, line 76: We could probably go a bit further with this validation, e.g., if there has been a broader type already registered, e.g., a `float` param registered before an `int` param, or a `str` param registered before a `uuid` param, we could raise an error.

starlite/asgi/routing_trie/traversal.py
```diff
@@ -30,37 +28,43 @@ def traverse_route_map(
...
-
-            path_params.append((param_definition.name, param_definition.parser(component)))
-            current_node = current_node.children[PathParameterSentinel]
-
-        continue
+        if current_node.path_type_path_param_definition:
+            path_params.append(
+                (
+                    current_node.path_type_path_param_definition.name,
+                    normalize_path(path.removeprefix(current_node.path)),
```
peterschutt, line 40: Another use for `RouteTrieNode.path` - this is slightly more efficient with `removeprefix()`:
```
$ poetry run python -m timeit -s 'from starlite.utils import normalize_path' -s 'path="/a/b/c/d/e"' -s 'pre="/a/b"' 'path.removeprefix(pre)'
10000000 loops, best of 5: 29.2 nsec per loop
$ poetry run python -m timeit -s 'from starlite.utils import normalize_path' -s 'components=["c","d","e"]' '"/".join(components)'
5000000 loops, best of 5: 42.9 nsec per loop
```
Although, `normalize_path()` dwarfs the difference in the two.

starlite/utils/path.py
```diff
@@ -13,7 +15,7 @@ def normalize_path(path: str) -> str:
     """
     path = path.strip("/")
     path = "/" + path
-    return re.sub("//+", "/", path)
+    return multi_slash_pattern.sub("/", path)
```
peterschutt, line 18: pre-compiling this makes sense

starlite/asgi/routing_trie/traversal.py
```diff
@@ -30,37 +28,43 @@ def traverse_route_map(
...
+            return current_node, path_params, path
+
+        for param_definition in current_node.child_path_parameter_type_map.values():
+            try:
+                value = param_definition.parser(component)
+            except ValueError:
+                continue
+            path_params.append((param_definition.name, value))
+            current_node = current_node.children[param_definition]
+            break
```
peterschutt, line 52: If there is going to be a performance impact in routing requests this is where it will come from, and will be relative to the number of adjacent path parameters that are children to the node.

starlite/asgi/routing_trie/traversal.py
```diff
@@ -30,37 +51,43 @@ def traverse_route_map(
...
-
-            path_params.append((param_definition.name, param_definition.parser(component)))
-            current_node = current_node.children[PathParameterSentinel]
-
-        continue
+        if current_node.path_type_path_param_definition:
+            path_params.append(
+                (
+                    current_node.path_type_path_param_definition.name,
+                    normalize_path(removeprefix(path, current_node.path)),
```
peterschutt, line 63: Another use for `RouteTrieNode.path` - this is slightly more efficient with `removeprefix()`:
```
$ poetry run python -m timeit -s 'from starlite.utils import normalize_path' -s 'path="/a/b/c/d/e"' -s 'pre="/a/b"' 'path.removeprefix(pre)'
10000000 loops, best of 5: 29.2 nsec per loop
$ poetry run python -m timeit -s 'from starlite.utils import normalize_path' -s 'components=["c","d","e"]' '"/".join(components)'
5000000 loops, best of 5: 42.9 nsec per loop
```
Although, `normalize_path()` dwarfs the difference in the two.

peterschutt:
The case in #827 would (almost) be solved now.

The only thing that remains an issue there is that I don't think we get guaranteed route registration order from routes that are registered via a controller class.

I still need to add tests, and update docs, but right to go for a review.

Goldziher:
So, I don't understand what this PR aims to accomplish- can you explain?

Also, some of the changes to traversal might have negative performance impact - i fine tuned this with a lot (A LOT) of benchmarks. So please do some benchmarks so there is a clear picture of what is what.

peterschutt:
It does this:

```
.
└── Root Node/
    └── vmc/
        ├── {path parameter: int}
        │   ─ OPTIONS
        │   ─ POST
        └── {path parameter: str}
            ─ OPTIONS
            ─ DELETE
```

I.e., allows for adjacent path parameter nodes as long as they aren't of the same type.

From the sounds of your comment in the relevant issue, spending any more time on this is not going to be productive.

PR #845, opened 2022-11-22
Update custom state injection and add `ImmutableState` class

Goldziher:
This PR is starting as an RFC. I would like to discuss the implementation. We discussed the need of making state immutable in some circumstances. The question is by what mechanism etc.


starlite/datastructures/state.py
```diff
@@ -12,11 +25,135 @@ class State(MutableMapping[str, Any]):
...
+        """Initialize an `ImmutableState` instance.
+
+        Args:
+             state: An object to initialize the state from. Can be a dict, an instance of 'ImmutableState', or a tuple of key value paris.
+        """
+
+        if isinstance(state, ImmutableState):
+            state = state._state
+
+        if not isinstance(state, dict) and isinstance(state, Iterable):
```
provinzkraut, line 38: A deep copy should always be created here, regardless unless `state` is an `ImmutableState` itself. Otherwise you can end up with weird side effects and the `ImmutableState` isn't really immutable.

starlite/datastructures/state.py
```diff
@@ -12,11 +25,135 @@ class State(MutableMapping[str, Any]):
...
+    def mutable_copy(self) -> "State":
+        """Return a mutable copy of the state object.
+
+        Returns:
+            A `State`
+        """
+        return State(self._state)
+
+    def dict(self) -> Dict[str, Any]:
+        """Create a shallow copy of the wrapped dict.
```
provinzkraut, line 110:
```suggestion
        """Return a shallow copy of the wrapped dict.
```

starlite/datastructures/state.py
```diff
@@ -157,30 +251,25 @@ def __delattr__(self, key: str) -> None:
...
-        """Return a shallow copy of the wrapped dict.
+    def frozen_copy(self) -> "ImmutableState":
+        """Return a shallow copy of the state object, setting it to be frozen.
 
         Returns:
-            A dict
+            A `State`
         """
-        return copy(self._state)
+        return ImmutableState(self)
```
provinzkraut, line 275: See comment above: The state returned here is not frozen. It's more akin to a "view" as it will reflect changes made to the mutable state it's wrapping. Would be resolved by addressing the above comment and copying mutable objects in `ImmutableState.__init__`.

docs/usage/0-the-starlite-app/2-using-application-state.md
```diff
@@ -7,17 +7,68 @@ The advantage of using application `state`, is that it can be accessed during mu
...
+
+
+@get("/")
+def handler(state: MyState) -> dict:
+    state.increment()
+    return state.dict()
+```
+
+While this is very powerful, it might encourage users to follow anti-patterns: its important to emphasize that using
+state can lead to hard to reason about code as well difficult to understand bugs due to changes in different ASGI
```
provinzkraut, line 62:
```suggestion
state can lead to code that's hard to reason about and bugs that are difficult to understand, due to changes in different ASGI
```

tests/datastructures/test_state.py
```diff
@@ -1,21 +1,44 @@
...
+    assert not state
+    state["first"] = "first"
+    state["second"] = "second"
+    assert state.first == "first"
+    assert state["second"] == "second"
+    del state["first"]
+    del state.second
+    assert "first" not in state
+    assert "second" not in state
+    assert isinstance(state.frozen_copy(), ImmutableState)
```
provinzkraut, line 42: Add tests to check that changes made to the initial object passed to `ImmutableState` don't propagate?

provinzkraut:
A general question: Would it not be better to have `State` be immutable by default, and make it so you specifically have to request a `MutableState` if you want to make mutations? That way we would discourage the pattern of mutating state in handlers directly. I feel like mentioning in the docs that you should use `ImmutableState` in most cases but then don't make it the default is a bit contradictory. (I know this would break backwards compatibility, but we could make that change at some point at least?)

Goldziher:
> A general question: Would it not be better to have `State` be immutable by default, and make it so you specifically have to request a `MutableState` if you want to make mutations? That way we would discourage the pattern of mutating state in handlers directly. I feel like mentioning in the docs that you should use `ImmutableState` in most cases but then don't make it the default is a bit contradictory. (I know this would break backwards compatibility, but we could make that change at some point at least?)

We should slate this for V2. Can you create a milestone to track V2? I think we can start planning this properly.

starlite/datastructures/state.py
```diff
@@ -12,11 +25,135 @@ class State(MutableMapping[str, Any]):
...
+        """
+        return self.__class__(copy(self._state))
+
+    def mutable_copy(self) -> "State":
+        """Return a mutable copy of the state object.
+
+        Returns:
+            A `State`
+        """
+        return State(self._state)
```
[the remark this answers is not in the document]
Goldziher: I'm aware. I thought of actually making this a view rather than a frozen object. It's still immutable. Either way there are cases where users are going to be unhappy. 

What do you think?

starlite/datastructures/state.py
Goldziher: Why deep?

tests/datastructures/test_state.py
Goldziher: Yup, once we come to an agreement above

starlite/datastructures/state.py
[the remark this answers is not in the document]
provinzkraut: Hmm. I'm not sure. The current solution is neither nor, which I think is the worst of both worlds :D

So either it should be truly frozen, which means creating a deep copy of the mutable objects it's based on, or it should be a proper view and declared as such (i.e. being named `StateView`).
I have a slight preference for the former (since then it's truly immutable and free from side effects) but maybe we should gather more opinions on this?

starlite/datastructures/state.py
provinzkraut: Because that would properly follow the semantics. An object that is "frozen" should **never** change. Making a shallow copy would allow situations where it could change.

provinzkraut:
> We should slate this for V2. Can you create a milestone to track V2? I think we can start planning this properly.

Agree. Created the milestone.

starlite/datastructures/state.py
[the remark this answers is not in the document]
Goldziher: nah, lets go with the immutable version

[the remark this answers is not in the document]
provinzkraut: Okay so you wanna update this to complete immutability?

provinzkraut (approved):

PR #853, opened 2022-11-23
Support different types of path parameters for different leaf nodes

Goldziher:
# PR Checklist

This PR resolves #825 by modifying the parsing and resolution of params to occur AFTER routing. Impact on performance is positive(!) and this is a better design. This does not address the requirement to have better validaiton - which we should spinoff into a different PR.  

Here is a benchmark:

params_rps_mean


provinzkraut (approved):
Nicely done! Cleaner code, increased performance, more features.. That's the kind of PR I like :heart:

PR #856, opened 2022-11-24
Dependencies with `yield`

provinzkraut:
This implements dependencies with `yield` for sync and async generators in HTTP and WebSocket handlers.

# PR Checklist


starlite/datastructures/provide.py
```diff
@@ -95,21 +88,30 @@ def add(self, generator: Union[Generator[Any, None, None], AsyncGenerator[Any, N
...
+        return AsyncCallable(wrapped)
 
-    def to_background_tasks(self) -> List[BackgroundTask]:
-        return [BackgroundTask(self._wrap_next(gen)) for gen in self._generators]
+    def wrap_asgi(self, app: "ASGIApp") -> "ASGIApp":
+        async def wrapped(scope: "Scope", receive: "Receive", send: "Send") -> None:
+            if len(self._generators) == 1:
+                await self._wrap_next(self._generators[0])()
+            elif self._generators:
+                async with anyio.create_task_group() as tg:
```
Goldziher, line 109:
```suggestion
                async with create_task_group() as task_group:
```

I know im annoying, sorry.

Goldziher (approved):

starlite/datastructures/provide.py
provinzkraut: No worries, I like me a thorough review :heart:

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
             and other.use_cache == self.use_cache
             and other.value == self.value
         )
+
+
+class DependencyCleanupGroup:
+    """Wrapper for generator based dependencies.
+
+    Simplify cleanup by wrapping `next`/`anext` calls in `BackgroundTasks` and providing facilities to `throw` /
```
Goldziher, line 91: BackgroundTasks

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+
+        Returns:
+            None
+        """
+        if self._closed:
+            raise RuntimeError("Cannot call cleanup on a closed DependencyCleanupGroup")
+        self._generators.append(generator)
+
+    def _wrap_next(self, generator: "AnyGenerator") -> Callable[[], Coroutine[None, None, None]]:
+        if isasyncgen(generator):
```
Goldziher, line 122: better to pre-calculate this in the init function and ensure that everything is an async callable.

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+    async def cleanup(self) -> None:
+        """Execute cleanup by calling `next` / `anext` on all generators.
+
+        If there are multiple generators to be called, they will be executed in a `TaskGroup`.
+
+        Returns:
+            None
+        """
+        if self._closed:
+            raise RuntimeError("Cannot call cleanup on a closed DependencyCleanupGroup")
```
Goldziher, line 143:
```suggestion
            raise RuntimeError("Cannot call cleanup on a closed DependencyCleanupGroup")
            
```

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+        self._closed = True
+        if len(self._generators) == 1:
+            await self._wrap_next(self._generators[0])()
```
Goldziher, line 146:
```suggestion
            await self._wrap_next(self._generators[0])()
            return
```

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+        elif self._generators:
```
Goldziher, line 147:
```suggestion
```

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+        """Execute cleanup by calling `next` / `anext` on all generators.
+
+        If there are multiple generators to be called, they will be executed in a `TaskGroup`.
+
+        Returns:
+            None
+        """
+        if self._closed:
+            raise RuntimeError("Cannot call cleanup on a closed DependencyCleanupGroup")
+        self._closed = True
```
Goldziher, line 144:
```suggestion
        if not self. _generators:
             return
             
        self._closed = True
```

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+        if len(self._generators) == 1:
+            await self._wrap_next(self._generators[0])()
+        elif self._generators:
+            async with create_task_group() as task_group:
+                for generator in self._generators:
+                    task_group.start_soon(self._wrap_next(generator))
```
Goldziher, line 150: what about cancellation? shouldn't we cancel the task_group on exception?

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+            An ASGI callable
+        """
+
+        async def wrapped(scope: "Scope", receive: "Receive", send: "Send") -> None:
+            await self.cleanup()
+            await app(scope, receive, send)
+
+        return wrapped
+
+    async def __aenter__(self) -> None:
```
Goldziher, line 168: ?

starlite/datastructures/provide.py
provinzkraut: That won't work. The generators are not known at initialization time of the class, since they're being dynamically added with the `add` method.

starlite/datastructures/provide.py
provinzkraut: Well it has to have an `__aenter__` method if it has an `__aexit__` method? The use as a context manager is just easier than manually catching exceptions, throwing them into the generators and re-raising.

starlite/kwargs.py
```diff
@@ -171,10 +176,18 @@ async def resolve_dependency(
         dependency: An instance of [Dependency][starlite.kwargs.Dependency]
         connection: An instance of [Request][starlite.connection.Request] or [WebSocket][starlite.connection.WebSocket].
         kwargs: Any kwargs to pass to the dependency, the result will be stored here as well.
+        cleanup_group: DependencyCleanupGroup to which generators returned by `dependency` will be added
     """
     signature_model = get_signature_model(dependency.provide)
     dependency_kwargs = signature_model.parse_values_from_connection_kwargs(connection=connection, **kwargs)
-    kwargs[dependency.key] = await dependency.provide(**dependency_kwargs)
+    value = await dependency.provide(**dependency_kwargs)
+    if isgenerator(value):
```
Goldziher, line 184:
```suggestion

    if isgenerator(value):
```

starlite/kwargs.py
```diff
@@ -171,10 +176,18 @@ async def resolve_dependency(
...
+        cleanup_group.add(value)
+        value = next(value)
+    elif isasyncgen(value):
+        cleanup_group.add(value)
+        value = await async_next(value)
```
Goldziher, line 189:
```suggestion
        value = await async_next(value)
        
```

starlite/datastructures/provide.py
```diff
@@ -68,3 +83,114 @@ def __eq__(self, other: Any) -> bool:
...
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional["Traceback"],
+    ) -> None:
+        """If an exception was raised within the contextmanager block, throw it into all generators."""
+        if exc_val:
+            await self.throw(exc_val)
+
+    async def throw(self, exc: BaseException) -> None:
```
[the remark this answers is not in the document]
provinzkraut: So, if an exception occurs within the handler function, we want to throw it into the generators, so it's being raised there, in order to make it possible to react differently when an exception occurred.

Something like this:
```python
def db_session():
    session = sessionmaker()
    try:
        yield session
    except:
        session.rollback()
    finally:
        session.commit()
```

starlite/routes/websocket.py
```diff
@@ -56,10 +57,15 @@ async def handle(self, scope: "WebSocketScope", receive: "Receive", send: "Send"
...
             await self.route_handler.authorize_connection(connection=websocket)
 
-        kwargs = await self._resolve_kwargs(websocket=websocket)
-        await self.route_handler.fn.value(**kwargs)
+        kwargs, cleanup_group = await self._resolve_kwargs(websocket=websocket)
 
-    async def _resolve_kwargs(self, websocket: "WebSocket[Any, Any]") -> Dict[str, Any]:
+        async with cleanup_group:
+            await self.route_handler.fn.value(**kwargs)
+        await cleanup_group.cleanup()
```
Goldziher, line 64: if youre using this as a context manager - you can call cleanup on the exit from the context.

starlite/routes/http.py
```diff
@@ -142,45 +143,57 @@ async def _call_handler_function(
...
 
-            parsed_kwargs = route_handler.signature_model.parse_values_from_connection_kwargs(connection=request, **kwargs)  # type: ignore
+            parsed_kwargs = route_handler.signature_model.parse_values_from_connection_kwargs(
+                connection=request, **kwargs
+            )
 
-        if route_handler.has_sync_callable:
-            return route_handler.fn.value(**parsed_kwargs)
-        return await route_handler.fn.value(**parsed_kwargs)
+        async with (cleanup_group or DependencyCleanupGroup()):
```
Goldziher, line 191: no cleanup called here?

starlite/datastructures/provide.py
provinzkraut: Actually, I think we should not. If we do, it means errors raised during the cleanup of independent dependencies would affect other dependencies. This sounds like a bad case of side effects to me. E.g. if I have a dependency that commits things to db on cleanup, it shouldn't fail because an unrelated file failed to close.

starlite/routes/http.py
provinzkraut: No. Cleanup is being called after the handler function is called, before the request is sent. Errors are thrown in the generators in between that. The two stages are separated, partly to allow for the same execution flow in websockets.

starlite/routes/websocket.py
provinzkraut: See comment below. The points at which exceptions are thrown and cleanup is called have to be separated for the http routes, so cleanup cannot automatically happen during exit.

starlite/kwargs.py
provinzkraut: Is there something wrong on my end? Those lines are the same for me.

starlite/datastructures/provide.py
provinzkraut: It should be close regardless though.

starlite/kwargs.py
jtraub: @provinzkraut I think it is about visual highlight for the `if` with an empty line before it.

provinzkraut: Oh. Yes. Thanks. I'm blind :upside_down_face:

Goldziher (approved):

PR #859, opened 2022-11-24
Ruff Integration

cofin:
Integrates `ruff` into the pre-commit workflow.

To Do: 
- [x] Implement `ruff` configuration based on existing `.flake8` 
- [x] Enable additional features (`pyupgrade`,`isort`, etc.)
- [ ] Remove components that `ruff` replaces


cofin:
So we do have a few plugins that haven't been incorporated into ruff yet.   Here is the summary of my testing:

What can be replaced with ruff:

- `pycodestyle`
- `pyupgrade`
- `flake8`
- `flake8-bugbear`
- `flake8-comprehensions`
- `flake8-mutable`
- `flake8-print`
- `isort` *I don't see a way to support `multi_line_output = 3`, but we may not need it.  I will open an issue with the `ruff` project to see if there is a way to enable or if it's on the roadmap.



The following specific checks aren't currently supported.
`SCS108` from 'flake8-secure-coding-standard'
`E225`, `W503`, `E203` from `pycodestyle`

The following utilities are not handled by ruff:

- `flake8-noqa`
- `flake8-return` 
- `flake8-simplify`
- `flake8-type-checking`
- `flake8-implicit-str-concat`
- `flake8-secure-coding-standard`
- `flake8-encodings`
- `flake8-use-fstring`
- `flake8-use-pathlib`
- `flake8-mutable`


I've currently left all of the steps in the pre-commit, and I've also added the ruff check.  

What do you guys think the best way to deal with this is?

cofin:
I've also not found a good way to ignore or disable subdirectories with a wild card.  For instance, we want to allow the use of `assert` in our the `tests/` and `examples/` folder, but not in the main code.  

The top 2 entries here work, but the bottom wildcard ones do not.  I've also tried several wildcard variations with no luck:
```toml

[tool.ruff.per-file-ignores]
"__init__.py" = ['F401']
"starlite/types/builtin_types.py" = ['F401']
"tests/**/*.py" = ['S101','D101','D103']
"examples/**/*.py" = ['S101','D101','D103']
```

PR #864, opened 2022-11-27
Add security module and JWT auth contrib

Goldziher:
# PR Checklist

Please see the code as an RFC, I will be updating it of course, but I want everyone to have visibility to what im doing.


starlite/contrib/jwt/jwt_token.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        A datetime instance
+    """
+    if value.tzinfo is not None:
+        value.astimezone(timezone.utc)
+    return value.replace(microsecond=0)
+
+
+class Token(BaseModel):
```
provinzkraut, line 33: Does this really need to be a pydantic model?
I feel like since we plan to drop pydantic as a hard dependency, we shouldn't unnecessarily be adding new pydantic-dependant functionality, since we'll need to remove it at some point not too far in the future anyway.

(This applies to other pydantic usage here as well)

Goldziher: i will refactor, now that tests are there.

Goldziher: So after spending some time on this - yes, we should use pydantic here. Its way simpler and cleaner. We can also not drop pydantic as a dependency because we use it for OpenAPI schema generation, so there is no point in fighting this.

provinzkraut: So this is off the table then?

> One change I want to discuss is to switch from Pydantic to dataclasses as the default fir Starlite. I don't mean remove Pydantic at all, but rather make Pydantic an optional dependency, without which Starlite will default to using the stdlib dataclasses.

The conclusion you drew here is quite the setback then in regard to supporting other serialization/validation libraries. If we still need to use pydantic under the hood to generate OpenAPI schemas, it severely limits the scope of where we can actually drop pydantic.

Also, it may be simpler and cleaner with pydantic, but I think it will harm performance substantially. If the token is a pydantic model which is being validated on every request sent, it could practically halve the performance gain in some situations, if someone were to pick something else for serialization.

I feel like *if* we make it possible to opt for the serialization/validation library of your choice, we shouldn't be sneaking pydantic in elsewhere. Using it for something like configs and stuff I can get behind, as that's a cost only paid at startup time. But what you're suggesting is that everything that has to be added to the OpenAPI schema **needs** to be pydantic, which means there are barely any places we can truly switch from pydantic.

I'd rather we invest the time and effort to somehow bridge these things, so they are not required to be pydantic models themselves. Because otherwise, we might as well not do the migration away from pydantic to begin with.

Goldziher: Not really. It's a very large amount of work to remove Pydantic from our OpenAPI logic, and I don't see why we should do this. 

We can support different serialization / validation backends, this doesn't require we remove Pydantic. The effort required to remove this dependency is excessive and the benefit is too low. I mean, if you feel like putting in the 100+ hours we'd need for this change just so we have one less dependecy, go ahead, but I think it's not worth the effort. 

Finally - for config classes that are evaluated at startup, Pydantic is actually perfect. It's performance is not an issue in this case and it offers a vastly better alternative to dataclasses.

provinzkraut: > We can support different serialization / validation backends, this doesn't require we remove Pydantic. 
No, but we talked about removing it from the internal datastructures where it's not strictly necessary. 

> I mean, if you feel like putting in the 100+ hours we'd need for this change just so we have one less dependecy, go ahead

Not really, no :upside_down_face: 

> Finally - for config classes that are evaluated at startup, Pydantic is actually perfect. It's performance is not an issue in this case and it offers a vastly better alternative to dataclasses.

Yes, I agree with you there!

> > Using it for something like configs and stuff I can get behind

cofin:
What do you guys think about providing a base User model similar to Django as part of the built in security?  The user would then extend the implementation with whatever persistence model they're using and any additional columns/relationships.

provinzkraut:
> What do you guys think about providing a base User model similar to Django as part of the built in security? The user would then extend the implementation with whatever persistence model they're using and any additional columns/relationships.

I had this sort of functionality in my initial outline I think. I filed it under "later", for after we're done with the basic implementation. So yes, I think something like this would be nice to have.

Goldziher:
> What do you guys think about providing a base User model similar to Django as part of the built in security? The user would then extend the implementation with whatever persistence model they're using and any additional columns/relationships.

i dont quite understand whats the usecase tbh. what do you guys think?

sonar-project.properties
```diff
@@ -1,7 +1,7 @@
 sonar.projectKey=starlite-api_starlite
 sonar.organization=starlite-api
 sonar.coverage.exclusions=**/__init__.py, **/*.js
-sonar.cpd.exclusions=starlite/params.py, starlite/handlers/*, starlite/connection.py
+sonar.cpd.exclusions=starlite/params.py, starlite/handlers/*, starlite/connection.py, starlite/contrib/jwt/jwt_auth.py
```
Goldziher, line 4: sonar misidentifies duplications there

docs/index.md
```diff
@@ -27,9 +27,9 @@
 Starlite is a light, opinionated and flexible ASGI API framework built on top of **[pydantic](https://github.com/samuelcolvin/pydantic)**.
 
 The Starlite framework supports **[plugins](usage/10-plugins/0-plugins-intro.md)**, ships
-with **[dependency injection](usage/6-dependency-injection/0-dependency-injection-intro.md)**, **[authentication](usage/8-authentication/0-intro.md)**
+with **[dependency injection](usage/6-dependency-injection/0-dependency-injection-intro.md)**, **[security](usage/8-security/0-intro.md)**
```
provinzkraut, line 30:
```suggestion
with **[dependency injection](usage/6-dependency-injection/0-dependency-injection-intro.md)**, **[security primitives](usage/8-security/0-intro.md)**
```

"Ships with security" doesn't really make sense IMO.

docs/usage/9-testing/4-pydantic-factories.md
```diff
@@ -0,0 +1,105 @@
+# Using pydantic-factories
+
+Starlite bundles the library [pydantic-factories](https://github.com/Goldziher/pydantic-factories), which offers an easy
```
provinzkraut, line 3:
```suggestion
[Pydantic-factories](https://github.com/starlite-api/pydantic-factories) offers an easy
```

Add note on how to install via pip

examples/security/using_session_auth.py
```diff
@@ -0,0 +1,138 @@
...
+# The SessionAuth class requires a handler callable
+# that takes the session dictionary, and returns the
+# 'User' instance correlating to it.
+#
+# The session dictionary itself is a value the user decides
+# upon. So for example, it might be a simple dictionary
+# that holds a user id, for example: { "id": "abcd123" }
+#
+# Note: The callable can be either sync or async - both will work.
+async def retrieve_user_handler(session: dict[str, Any], connection: ASGIConnection) -> Optional[User]:
```
provinzkraut, line 53:
```suggestion
async def retrieve_user_handler(session: Dict[str, Any], connection: ASGIConnection) -> Optional[User]:
```
Examples should work on Python 3.8

examples/security/using_session_auth.py
```diff
@@ -0,0 +1,138 @@
+from typing import Any, Literal, Optional
```
provinzkraut, line 1:
```suggestion
from typing import Any, Dict, Literal, Optional
```

sonar-project.properties
provinzkraut, line 4: Why? (Just curious)

starlite/contrib/jwt/jwt_auth.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+class JWTAuth(Generic[UserType], AbstractSecurityConfig[UserType, Token]):
+    """JWT Authentication Configuration.
+
+    This class is the main entry point to the library, and it includes methods to create the middleware, provide login
+    functionality, and create OpenAPI documentation.
+    """
+
+    algorithm: str = "HS256"
```
provinzkraut, line 30: I think this could be a `Literal`, just for convenience.

starlite/contrib/jwt/jwt_auth.py
```diff
@@ -0,0 +1,359 @@
...
+            The encoded token formatted for the HTTP headers
+        """
+        if self.openapi_components.securitySchemes:
+            security = self.openapi_components.securitySchemes.get(self.openapi_security_scheme_name, None)
+            if isinstance(security, SecurityScheme):
+                return f"{security.scheme} {encoded_token}"
+        return encoded_token
+
+
+class JWTCookieAuth(Generic[UserType], JWTAuth[UserType]):
```
provinzkraut, line 194: This looks awfully similar to the session backend cookie config. Maybe we can/should abstract this?

starlite/contrib/jwt/jwt_token.py
```diff
@@ -0,0 +1,126 @@
...
+    constr,
+    validator,
+)
+
+from starlite import ImproperlyConfiguredException
+from starlite.exceptions import NotAuthorizedException
+
+
+def _normalize_datetime(value: datetime) -> datetime:
+    """Convert the given value into UTC and strips microseconds.
```
provinzkraut, line 20:
```suggestion
    """Convert the given value into UTC and strip microseconds.
```

starlite/contrib/jwt/middleware.py
```diff
@@ -0,0 +1,164 @@
...
+        app: "ASGIApp",
+        algorithm: str,
+        auth_header: str,
+        exclude: Optional[Union[str, List[str]]],
+        exclude_opt_key: str,
+        retrieve_user_handler: "AsyncCallable[[Token, ASGIConnection[Any, Any, Any]], Awaitable[Any]]",
+        scopes: "Scopes",
+        token_secret: str,
+    ):
+        """Check incoming requests for an encoded token in the auth header specified, and if present retrieves the user
```
provinzkraut, line 35:
```suggestion
        """Check incoming requests for an encoded token in the auth header specified, and if present retrieve the user
```

This should go into the class docstring.

starlite/contrib/jwt/middleware.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns:
+            AuthenticationResult
+
+        Raises:
+            [NotAuthorizedException][starlite.exceptions.NotAuthorizedException]: If token is invalid or user is not found.
+        """
+        auth_header = connection.headers.get(self.auth_header)
+        if not auth_header:
+            raise NotAuthorizedException("No JWT token found in request header")
+        _, _, encoded_token = auth_header.partition(" ")
```
provinzkraut, line 72:
```suggestion
        encoded_token = auth_header.partition(" ")[-1]
```

If you want to assert a certain amount of parts, it should be done explicitly. Else grabbing the last part is simpler and more explicit like this.

starlite/contrib/jwt/middleware.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns:
+            AuthenticationResult
+
+        Raises:
+            [NotAuthorizedException][starlite.exceptions.NotAuthorizedException]: If token is invalid or user is not found.
+        """
+        auth_header = connection.headers.get(self.auth_header) or connection.cookies.get(self.auth_cookie_key)
+        if not auth_header:
+            raise NotAuthorizedException("No JWT token found in request header or cookies")
+        _, _, encoded_token = auth_header.partition(" ")
```
provinzkraut, line 163:
```suggestion
       encoded_token = auth_header.partition(" ")[-1]
```
See above comment

starlite/security/base.py
```diff
@@ -0,0 +1,166 @@
...
+
+    from starlite.config import AppConfig
+    from starlite.middleware.base import DefineMiddleware
+
+
+UserType = TypeVar("UserType")
+AuthType = TypeVar("AuthType")
+
+
+class AbstractSecurityConfig(ABC, Generic[UserType, AuthType], GenericModel):
```
provinzkraut, line 36: I'm thinking if this shouldn't be the config for a plugin. Then you wouldn't need to manually set up the middleware and `on_init` hooks.

pyproject.toml
```diff
@@ -80,6 +80,7 @@ pre-commit = "*"
 pytest = "*"
 pytest-asyncio = "*"
 pytest-cov = "*"
+python-jose = "*"
```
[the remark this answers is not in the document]
Goldziher: good catch

sonar-project.properties
Goldziher: because the login method is duplicated for parameters but has a different body

starlite/contrib/jwt/jwt_auth.py
Goldziher: no, there are too many options.

starlite/security/base.py
Goldziher: we would need to set a plugin instead, its exactly the same amount of work. These are not plugins though - and they wont implement most of the methods specified by the plugin protocol, so there is no point in that.

starlite/contrib/jwt/jwt_auth.py
provinzkraut: Too bad

starlite/security/base.py
provinzkraut: I don't agree that it's the same amount of work. Passing in the middleware and `on_init` handler vs simply passing in a plugin. The latter feels more "works out of the box". But I'm fine with not having it a plugin as well.

starlite/contrib/jwt/jwt_auth.py
provinzkraut: I was talking about the cookie fields. But I guess this level of abstraction might also be just fine.

provinzkraut (approved):

starlite/security/base.py
Goldziher: but you dont pass the middleware, you just pass the `on_init` handler. See all the examples and tests

PR #866, opened 2022-11-27
Docs update

provinzkraut:
Minor updates to our docs, fixes some outdated information and references and adds benchmarks.

# PR Checklist

- [x] Have you followed the guidelines in `CONTRIBUTING.md`?
- [ ] Have you got 100% test coverage on new code?
- [x] Have you updated the prose documentation?
- [ ] Have you updated the reference documentation?

Issue #420, opened 2022-08-27
Bug: Documentation Site doesn't work properly in mobile.


Goldziher:
The Doc site isn't working well in mobile. You need to press the back arrow to find the menu to see usage and reference.

omertalmi5:
I would like to solve that issue.
Can you explain the expected behavior by you?
It looks well for me, the menu keeps its state.
Attaching reproducing from my mobile.

https://user-images.githubusercontent.com/43521606/187301451-3bd62049-ab5d-4b5e-b8bf-47a286ad73ce.mp4

Goldziher:
> @Goldziher 
> I would like to solve that issue.
> Can you explain the expected behavior by you?
> It looks well for me, the menu keeps its state.
> Attaching reproducing from my mobile.
> 
> https://user-images.githubusercontent.com/43521606/187301451-3bd62049-ab5d-4b5e-b8bf-47a286ad73ce.mp4
> 
> 

Broker tov Omer,

To be honest I resolved this issue a few hours before you write this, and forgot to close the issue.

alphanoobie:
Hello! I was trying to work on this issue but was not able to find a fix for it.
It would be a great help if you could tell me a gist of how you solved it @Goldziher

PR #869, opened 2022-11-28
refixing the mobile navbar

StitiFatah:
The issue #420 mentioned that needing to press the back arrow to access top top level menu on mobile wasn't the wanted behavior. It's been fixed by #433 but reintroduced by #866 

Not sure if you want to keep Starlite or go back to Overview  since it forces to pulls out the index (Intro in the navbar)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -39,8 +39,8 @@ extra_css:
   - css/typo.css
   - css/color.css
 nav:
+  - index.md
   - Starlite:
-    - Intro: index.md
     - benchmarks.md
     - governance.md
     - contributing.md


provinzkraut:
I cannot reproduce this at the moment?

StitiFatah:
Aren't you getting this when pushing th hamburger button on mobile ? Maybe it's cached on your side ?
navbar_starlite

StitiFatah:
Do we agree that we're talking about what we get directly after pushing the button and not after pressing the "previous" button ? 

If it's the case I'm pretty sure the screenshot I posted is what's happening on the actually deployed version of the docs as I tried with multiple browsers on mobile/desktop in incognito mode.

Here's a video on Google Chrome which I never use and thus don't have anything cached but I cleared the cache anyway : https://watch.screencastify.com/v/rFaQ1ulXJE8cToup8Ve6

provinzkraut:
> Do we agree that we're talking about what we get directly after pushing the button and not after pressing the "previous" button?

Yes. I can only reproduce this in Chromium on my desktop, it works fine in Firefox, and mobile Safari + Firefox. Not sure why?

provinzkraut:
Aight, since this happened before I'm just gonna assume it's an issue on my end. I'm not really happy with the fix though to be honest, since it kinda messes with the structure and makes things a bit convoluted. 
I'll take a look tomorrow and see if I can come up with a better solution, else I'm just gonna merge this.

StitiFatah:
I feel it's messy too but it was just like that before #866 with `Overview` instead of `Starlite`. I could have done something if it was from scratch but I'm not familiar enough with mkdocs to know if it's doable.

In my opinion it's a theme problem, it's not obvious that "Home" is the selected menu in the screenshot below (local version after fix) nor that we can expend it, but it just lists  the different points of index.md so people aren't missing anything and have all the menus clearly accessible, while the back button issue may be annoying, I've actually seen someone complaining about it on reddit, thinking he couldn't access the docs at all on mobile, that's why I checked.



Screenshot_2022-11-28_18-26-39

provinzkraut:
@StitiFatah I created an alternative solution in #873. Can you confirm that this works for you?

StitiFatah:
Yes it works and is way cleaner like that 
Screenshot_2022-11-29_11-00-19

provinzkraut:
Nice! Thanks for bringing this up!

PR #889, opened 2022-11-30
fix: cookie caching issues

Goldziher:
# PR Checklist

### DIFF
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -1,5 +1,4 @@
 from enum import Enum
-from functools import lru_cache
 from inspect import Signature, isawaitable
 from itertools import chain
 from operator import attrgetter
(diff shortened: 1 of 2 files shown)


jtraub (approved):

starlite/handlers/http.py
```diff
@@ -82,16 +81,12 @@
 MSG_SEMANTIC_ROUTE_HANDLER_WITH_HTTP = "semantic route handlers cannot define http_method"
 
 
-@lru_cache(1024)
```
provinzkraut, line 85: Wouldn't it be better to include `value` and such fields in the `Cookie.__hash__`, so the cache works properly instead of simply removing the cache?

Goldziher: its a quid pro quo, this function uses set functionality to create a unique list of cookies. if we include value there, we have to do some complex iterations to ensure that the cookies are unique, this is very bad performance wise. So thats really the question - do we have a suboptimal function but cached, or a pretty optimal function uncached. I went with the letter because its more elegant code wise. 

Note: We cannot benchmark this at present because none of our benchmarks uses responses being returned from handlers.

jtraub: @provinzkraut it feels like we will be loosing to much on determining whether the cache should be bypassed or not. Notice that we pass sets of cookies here.

provinzkraut: > So thats really the question - do we have a suboptimal function but cached, or a pretty optimal function uncached. I went with the letter because its more elegant code wise.

That's fair. 

> We cannot benchmark this at present because none of our benchmarks uses responses being returned from handlers.

I'll add it to the "missing benchmarks" issue.

PR #891, opened 2022-11-30
`MessagePack` support and serialization with `msgspec`

provinzkraut:
Replaces `orjson` with [msgspec](https://github.com/jcrist/msgspec) as the default JSON ser/de library and adds support for `MessagePack` as a serialization target.

Note that this cannot be released currently, as a feature we need for compatibility hasn't yet been released on `msgspec`s side.


docs/usage/4-request-data.md
```diff
@@ -0,0 +1,195 @@
...
+async def handle_file_upload(
+    data: list[UploadFile] = Body(media_type=RequestEncodingType.MULTI_PART),
+) -> None:
+    ...
+```
+
+
+## MessagePack data
+
+To receive `MessagePack` data, you can either specify the appropriate `Content-Type`
```
Goldziher, line 190: We use the MediaType enum, we should explain this is a synonym

Goldziher (approved):
Looks gr8

docs/usage/4-request-data.md
provinzkraut: This is in line with how the sections for multipart/form data describe it. Use the enum for things inside of Starlite, and literal header values when describing headers from incoming requests.

starlite/utils/serialization.py
```diff
@@ -20,4 +23,87 @@ def default_serializer(value: Any) -> Any:
...
+    """
+    if issubclass(type_, BaseModel):
+        return type_(**value)
+    raise TypeError(f"Unsupported type: {type(value)!r}")
+
+
+_msgspec_json_encoder = msgspec.json.Encoder(enc_hook=default_serializer)
+_msgspec_json_decoder = msgspec.json.Decoder(dec_hook=dec_hook)
+_msgspec_msgpack_encoder = msgspec.msgpack.Encoder(enc_hook=default_serializer)
+_msgspec_msgpack_decoder = msgspec.msgpack.Decoder(dec_hook=dec_hook)
```
jcrist, line 51: A `dec_hook` is only called for [typed decoding](https://jcristharif.com/msgspec/usage.html#typed-decoding), where an output `type` is specified. Since these decoders never specify a type, the `dec_hook` here is doing nothing and can be removed (perhaps I should add a warning for this in `msgspec`).

```suggestion
_msgspec_json_encoder = msgspec.json.Encoder(enc_hook=default_serializer)
_msgspec_json_decoder = msgspec.json.Decoder()
_msgspec_msgpack_encoder = msgspec.msgpack.Encoder(enc_hook=default_serializer)
_msgspec_msgpack_decoder = msgspec.msgpack.Decoder()
```

starlite/utils/serialization.py
```diff
@@ -20,4 +23,87 @@ def default_serializer(value: Any) -> Any:
...
+    """Decode a JSON string/bytes into an object.
+
+    Args:
+        raw: Value to decode
+
+    Returns:
+        An object
+    """
+    if isinstance(raw, str):
+        raw = raw.encode("utf-8")
```
jcrist, line 79: Are strings ever passed to `decode_json` in a hot path? Or is this just for compatibility? We could support this natively in `msgspec` (which would be slightly faster than doing this), but decoding from a strings should be avoided for performance reasons when possible.

starlite/parsers.py
```diff
@@ -57,9 +55,8 @@ def parse_url_encoded_form_data(encoded_data: bytes, encoding: str) -> Dict[str,
     """
     decoded_dict: DefaultDict[str, List[Any]] = defaultdict(list)
     for k, v in parse_query_string(query_string=encoded_data, encoding=encoding):
-        if isinstance(v, str):
-            with suppress(JSONDecodeError):
-                v = loads(v)
+        with suppress(JSONDecodeError):
+            v = loads(v) if isinstance(v, str) else v
```
Goldziher, line 59: @jcrist - this is the one unsupported use we have, which we need to use the stdlib json for. This parses `urlencoded` request bodies.

Values can be stringified lists: `"[1,2,3,4]"` and stringified numbers / floats, which are converted into their respective data types.

jcrist: Thanks for the ping. Wouldn't the `starlite.utils.decode_json` method work here? The overhead of decoding from a string here should be dwarfed by the performance of upstream processing code like `parse_qsl`. These strings are also likely to be short, so the unnecessary copy when decoding from a utf-8 string shouldn't matter as much.

Goldziher: No, this raises a `TypeError` from msgspec. 

Of course this is not the main performance issue, but we have no choice but to do both in this case, that or start using regex, which can be a problem.

jcrist: Looks like it's working to me?

```python
In [8]: decode_json("[1, 2, 3, 4, 5]")  # starlite.utils.serialization.decode_json
Out[8]: [1, 2, 3, 4, 5]

In [9]: decode_json("1")
Out[9]: 1
```

starlite/utils/serialization.py
provinzkraut: There is actually a reason to have those in there. We want to support using `Structs` and typed decoding in the future, so I already added those.

jcrist: :+1: If this is code you don't expect to be useful now, but do expect to be useful in the future then that's fine. Just wanted to clarify that as written the `dec_hook` will never be called since no `type=` kwarg was passed in.

starlite/parsers.py
Goldziher: yup, i wonder why it did work for me before.

.pre-commit-config.yaml
```diff
@@ -124,7 +124,7 @@ repos:
             fsspec,
             multidict,
             opentelemetry-instrumentation-asgi,
-            orjson,
+            msgspec@git+https://github.com/jcrist/msgspec.git,
```
jcrist, line 127: msgspec 0.10.0 has now been released, should be able to pin to the latest release now.

Goldziher:
I think this can be finished no? Msgspec released an update

provinzkraut:
> I think this can be finished no? Msgspec released an update

Yes, working on it.

provinzkraut:
There's a bug in `msgspec`s dataclasses implementation. I reported it here https://github.com/jcrist/msgspec/issues/233. We should probably wait for a fix until we release this.

PR #902, opened 2022-12-03
Fix #899 - Generator based dependencies with cached responses

### LINKED ISSUE #899: Bug: Caching response throws AttributeError on GET: Can't pickle local object 'DependencyCleanupGroup.wrap_asgi.<locals>.wrapped'
**Describe the bug**
When caching the response of a get request for a `Template`, I get the following exception which throws a server 500 error:

`AttributeError on GET:  Can't pickle local object 'DependencyCleanupGroup.wrap_asgi.<locals>.wrapped'`

**To Reproduce**
Here is a reproducible example:  

```
import uvicorn

from starlite import Starlite, get, Request, State, Template, Controller, TemplateConfig
from starlite.template.jinja import JinjaTemplateEngine


class FrontEnd(Controller):
    @get("/", cache=60)
    async def serveHomepage(self, request: Request, state: State) -> Template:
        return Template(
            name="homepage.html.j2",
            context={"request": request},
        )

    @get("/hello-world", cache=60)
    async def hello_world(self) -> dict[str, str]:
        return {"hello": "world"}


app = Starlite(
    route_handlers=[FrontEnd],
    template_config=TemplateConfig(
        directory="src/templates", engine=JinjaTemplateEngine
    ),
    debug=True,
)

if __name__ == "__main__":
    uvicorn.run(app, host="127.0.0.1", port=8000)


```

You can use this for the Jinja template:

```
<html>
<head>
</head>
<body>
    <h1>Homepage Header</h1>
</body>
</html>
```

**Additional context**
If you remove the cache value from the `Template` route (or set it to zero), the `get` requests are served as expected.  

In the example code, I also included a route for a non-template request, `/hello-world`, which does not throw an error.  So the problem seems to lie with `Template` responses, or some subset of responses.

provinzkraut:
Fix #899 by not wrapping the response but instead calling the cleanup after all response processing is done. This maintains the current order of execution.

# PR Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -173,7 +173,7 @@ repos:
             types-redis,
           ]
   - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.281
+    rev: v1.1.282
     hooks:
       - id: pyright
         additional_dependencies:
--- a/starlite/datastructures/provide.py
+++ b/starlite/datastructures/provide.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
     from inspect import Traceback
 
     from starlite.signature import SignatureModel
-    from starlite.types import AnyCallable, AnyGenerator, ASGIApp, Receive, Scope, Send
+    from starlite.types import AnyCallable, AnyGenerator
 
 
 class Provide:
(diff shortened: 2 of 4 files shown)


jtraub (approved):
This solution is even better than what we discussed in the beginning.

It works for me with the provided example

PR #905, opened 2022-12-04
Fix issue with OpenAPI Controller and ASGI root_path

smithk86:
# PR Checklist

I was testing a _Starlite_ app with _uvicorn_'s _--root-path_ option, and found (what appears to be) a bug.

`OpenAPIController` compares `request.url.path` to the path given by the client to determine if the request is asking for something openapi-related but `request.url.path` can have an ASGI _root_path_ prepended to it if it is given. This was causing all openapi routes to throw a 404.

I think I found an elegant solution to address this issue.

Cheers!
Kyle


Goldziher (approved):

Goldziher:
@all-contributors add @smithk86 code

PR #908, opened 2022-12-04
add support for pagination

Goldziher:
# PR Checklist


provinzkraut:
I think this is missing some important features:

- The pagination objects should (optionally) include URLs for the previous/current/next pages
- There should be a class that does actual pagination, with pages
- It should be possible to use pagination in sync handlers

I'm also not crazy about the `TestCursorPagination(data=...).get_paginated_data(cursor=..., results_per_page=...)` pattern. Creating an instance of a class and then only calling one method on it seems like an antipattern to me and like the class should just have been a callable.

The class-based setup also facilitates some bad practices when working with databases, i.e. storing a database connection that's controlled externally on a class instance. This is required because `get_items` and `get_total` need database access, which they can't get through a param.

starlite/datastructures/pagination.py
```diff
@@ -0,0 +1,149 @@
...
+
+T = TypeVar("T")
+C = TypeVar("C", int, str, UUID)
+
+if TYPE_CHECKING:
+    from starlite.types import SyncOrAsyncUnion
+
+
+@dataclass
+class LimitOffsetPagination(Generic[T]):
```
provinzkraut, line 15: Why not just `OffsetPagination`? A bit shorter and the "Limit" is implied

Goldziher:
> * The pagination objects should (optionally) include URLs for the previous/current/next pages

Thats very complex to implement and get right, and its completely redundant. This is not the design in there. So for this one - no.

> * There should be a class that does actual pagination, with pages

I will add this, although not required IMO. 

> * It should be possible to use pagination in sync handlers

Ok, i'll work this in. 

> I'm also not crazy about the `TestCursorPagination(data=...).get_paginated_data(cursor=..., results_per_page=...)` pattern. Creating an instance of a class and then only calling one method on it seems like an antipattern to me and like the class should just have been a callable.

Its not. I will work on this again, but it requires more implementation than a simply callable. 

> The class-based setup also facilitates some bad practices when working with databases, i.e. storing a database connection that's controlled externally on a class instance. This is required because `get_items` and `get_total` need database access, which they can't get through a param.

Thats a non-sequitor. The entire point of providing a class is to provide a standardized interface to doing persistence handling. Its not a bad practice - IT IS THE POINT.

This was not a particularly helpful review btw.

PR #920, opened 2022-12-10
Handle mounted apps properly when checking for exclusion in middlewares

### LINKED ISSUE #781: Bug: Rate Limiter Middleware: exclude path is ignored, static files throw 429 error
**Describe the bug**
When using the RateLimiter middleware, the `exclude` path does not seem to be honored.  In the code example below, the app serves an HTML response to the root URL path.  In the HTML response, there is a reference to a static file which serves CSS.  With a rate limit of 1 request per second, the homepage will get served, but the CSS file throws a 429 error code.

**To Reproduce**

APP:
```python
import uvicorn
from starlite import (
    Request,
    Starlite,
    State,
    StaticFilesConfig,
    Template,
    TemplateConfig,
    get,
)
from starlite.middleware import RateLimitConfig
from starlite.template.jinja import JinjaTemplateEngine


@get("/")
async def serveHomepage(request: Request, state: State) -> Template:

    return Template(
        name="index.html.j2",
        context={"request": request},
    )


rate_limit_config = RateLimitConfig(
    rate_limit=("second", 1),
    exclude=[r"^/src.*$"],
)

app = Starlite(
    route_handlers=[serveHomepage],
    template_config=TemplateConfig(
        directory="src/templates", engine=JinjaTemplateEngine
    ),
    static_files_config=[
        StaticFilesConfig(directories=["src/static"], path="/src/static"),
        StaticFilesConfig(
            directories=["src/templates"], path="/src/templates", html_mode=True
        ),
    ],
    middleware=[rate_limit_config.middleware],
)

if __name__ == "__main__":
    uvicorn.run(app, host="127.0.0.1", port=8000)

```

HTML TEMPLATE:
```html
<html>

<head>
  <link href="/src/static/css/styles.css" rel="stylesheet" type="text/css" />
</head>

<body>
  <h1>Header</h1>
</body>

</html>
```

STYLES.CSS... you can just make an empty file  😁

**Additional context**
On my local, I'm running Starlite 1.38.0 on Python 3.11.  Maybe there's an issue with how I'm declaring the path, but I've tried numerous string formats for the excludes option, none seem to work:

`exclude=[r"^/src.*$"]`
`exclude=[r"^src.*$"]`
`exclude=["/src/"]`
`exclude=["src/"]`
`exclude=["/src/static/css/styles.css"]`

jtraub:
# PR Checklist

This PR closes #781


starlite/middleware/utils.py
```diff
@@ -51,6 +51,11 @@ def should_bypass_middleware(
         return True
     if exclude_opt_key and scope["route_handler"].opt.get(exclude_opt_key):
         return True
-    if exclude_path_pattern and exclude_path_pattern.findall(scope["path"]):
+    path = (
+        scope["path"]
+        if scope.get("mount_path") is None
```
jtraub, line 56: I made this check instead of `scope["mount_path"]` to make sure that a middleware would work even if it is used outside of Starlite with its `scope`.

starlite/asgi/routing_trie/traversal.py
```diff
@@ -82,6 +85,9 @@ def parse_path_params(
         parameter_definitions: The parameter definitions tuple from the route.
         path_param_values: The string values extracted from the url
 
+    Raises:
+        ValueError: If any of path parameters can not be parsed into a value.
```
jtraub, line 89: While `parse_path_params` doesn't explicitly raises `ValueError` it is raised indirectly so I decided to update the docsstring here b/c it took me some time to understand sources of `KeyError` and `ValueError`  exceptions catched in `parse_path_to_route`

Goldziher (commented):
I'm not sure this can proceed. I need to benchmark this. It looks to me this is going to degrade performance. I'll keep you posted.

PR #931, opened 2022-12-14
feat: make SignatureModel inherit error_msg_templates

asomethings:
Make `SignatureModel` inherit `error_msg_templates` from parameter annotation.

Since pydantic does not support nested error messages, starlite users can not leverage error message templates because every parameter becomes a field of `SignatureModel`. 

This PR makes pydantic's `error_msg_template` work. But if there is duplicate `error_msg_templates` key then last parameter's `error_msg_template` will be used.

# PR Checklist

### DIFF
--- a/starlite/signature.py
+++ b/starlite/signature.py
@@ -27,6 +27,7 @@ from starlite.exceptions import (
 )
 from starlite.plugins.base import PluginMapping, PluginProtocol, get_plugin_for_value
 from starlite.utils import (
+    is_class_and_subclass,
     is_dependency_field,
     is_optional_union,
     should_skip_dependency_validation,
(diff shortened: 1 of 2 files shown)


Goldziher (commented):
Hi,

First off - thanks for contributing! The. Is in general good. BUT - this PR can't bs merged. For two reasons- 

1. SignatureModel is intended to be an internal. As such it's not a user facing part of the framework and we are not going to publicly document or expose it. As my such this is not the right place to do this kind of changes. 

2. We are actually rebuilding this part of the app to move away from Pydantic, because this is too slow.

asomethings:
@Goldziher That sounds great! Thank for the review!

PR #932, opened 2022-12-14
Add documentation for TestClient.portal()

smithk86:
# PR Checklist

I have found the the blocking portal helper on `TestClient` to be extremely helpful with testing and thought it could use some documentation. Any feedback on the additions is welcome.

Cheers!
Kyle

### DIFF
--- a/docs/reference/testing/0-test-client.md
+++ b/docs/reference/testing/0-test-client.md
@@ -15,6 +15,7 @@
             - head
             - set_session_data
             - get_session_data
+            - portal
 
 ::: starlite.testing.test_client.WebSocketTestSession
     options:
--- a/docs/usage/9-testing/1-test-client.md
+++ b/docs/usage/9-testing/1-test-client.md
@@ -85,3 +85,18 @@ two methods:
 ```py title="Getting session data"
 --8<-- "examples/testing/get_session_data.py"
 ```
+
+## Using a blocking portal
+
+The `TestClient` uses a feature of [anyio](https://anyio.readthedocs.io/en/stable/) called a **Blocking Portal**.
+
+The [anyio.abc.BlockingPortal](https://anyio.readthedocs.io/en/stable/api.html#anyio.abc.BlockingPortal) allows `TestClient`
+to execute asynchronous functions using a synchronous call. `TestClient` creates a blocking portal to manage
+`Starlite`'s async logic, and it allows `TestClient`'s API to remain fully synchronous.
+
+Any tests that are using an instance of `TestClient` can also make use of the blocking portal to execute asynchronous functions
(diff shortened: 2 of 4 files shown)


Goldziher (commented):
Nice! Please add tests for the examples. 

Any suggestions on how? @starlite-api/maintainers

examples/testing/portal.py
```diff
@@ -0,0 +1,38 @@
+import asyncio
```
provinzkraut, line 1: Use `anyio` instead of asyncio here. This better highlights `anyio support.

examples/tests/testing/test_portal.py
```diff
@@ -0,0 +1,3 @@
+from examples.testing.portal import test_portal_example_event, test_portal_example_sleep
+
+__all__ = ["test_portal_example_sleep", "test_portal_example_event"]
```
provinzkraut, line 3: What is the purpose of this?

provinzkraut (approved):

provinzkraut:
> Nice! Please add tests for the examples.
Any suggestions on how? @starlite-api/maintainers

It should get collected by pytest automatically since files and functions have the proper naming for that, so no need to add explicit tests.

examples/tests/testing/test_portal.py
smithk86: @Goldziher Requested the examples get run during testing. Since the examples are valid tests, they can be imported and _pytest_ will run them as-is. I added the `__all__` because _pycln_ will nuke them as unused imports without it.

provinzkraut: How about renaming `examples/testing/portal.py` to `examples/testing/test_with_portal.py`? Then it should get automatically get collected by pytest.

smithk86: I like the idea about renaming. I agree _examples/tests/testing/test_portal.py_ was fairly hacky.

smithk86: This is done

examples/testing/portal.py
smithk86: I'm struggling a bit with this one. One of my example includes how to spawn a background task during startup and check on it after the app has started. According to [anyio issue #457](https://github.com/agronholm/anyio/issues/457), this is not and will not be supported since there is no inherent monitor for the task; any exceptions are thrown into the void if nothing explicitly checks on it.

It sounds like a task scheduler would a better option but, for the sake of simplicity, that is not an option here. I could change the example but I was also trying to show to await something which is running on `Starlite`'s event loop.

Goldziher (approved):

provinzkraut:
@all-contributors add @smithk86 docs

Issue #935, opened 2022-12-15
Enhancement: Manually clear request cache


provinzkraut:
Currently there's no way to manually clear the request cache. This would be nice to have.

Ideally we'd prefix the cache key with a Starlite-namespace key for redis and memcached to make this easier.

Goldziher:
I think this is now handled by the CLI, no? otherwise users can define a cache key builder. Is there anything further requried here?

provinzkraut:
> I think this is now handled by the CLI, no? otherwise users can define a cache key builder. Is there anything further requried here?

It's not handled by the CLI. The CLI can handle sessions, but not the cache. A custom cause key builder isn't sufficient enough as well. As I outlined in the initial comment, we need to add the keys to a namespace we can then access later, similar to how the session backends handle it.

IMO we should just merge the two storage backends / create a generic storage interface. They basically have the same functionality, just wrapped slightly different.

PR #936, opened 2022-12-15
Add CLI

provinzkraut:
# PR Checklist


cofin (commented):
This looks great to me and close to the approach I've used in the past.  What do you think its remaining to complete on this?

provinzkraut:
So two things I want to add before merging this:

1. Dynamic registration of sub-commands (this has to be done in such a way that thing's don't raise exceptions when the CLI dependencies are not installed). 
2. #935 (maybe, not really a blocker, I'd just like to release the CLI with it)

The project template thing can wait for now. Ideally I'd like you or @peterschutt to get on this since you've done good work on similar things already. If not, I'll take this one some time in the mid-near future.

starlite/cli.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class _StarliteCLIException(ClickException):
+    def __init__(self, message: str) -> None:
+        super().__init__(style(message, fg="red"))
+
+
+def _autodiscover_app() -> tuple[str, Starlite]:
+    if app_path := getenv("STARLITE_APP"):
+        console.print(f"Using starlite app from env: [bright_blue]{app_path!r}")
```
JacobCoffee: Perhaps we should cap the framework name here to match the rest of the commits?

 `console.print(f"Using Starlite app from env: [bright_blue]{app_path!r}")`

starlite/utils/cli.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
```
provinzkraut, line 1: Yes, this whole module is a bit hacky but doing this enables us to provide a way to dynamically register commands, that doesn't break if the extra dependencies for the CLI are not installed. This could be particularly useful for external libraries that want to hook into the CLI (think `starlite users create ...`)

pyproject.toml
```diff
@@ -88,6 +92,11 @@ structlog = "*"
 tortoise-orm = "*"
 tox = "*"
 trio = "*"
+click = "^8.1.3"
```
Goldziher, line 95: Switch to `*` for all deps

starlite/cli.py
```diff
@@ -0,0 +1,326 @@
+from __future__ import annotations
```
Goldziher, line 1: Please keep consistency with the rest of the code base.

Goldziher (approved):

starlite/cli.py
provinzkraut: Didn't we say we would start to adopt this a while ago? I might be mistaken but I think I remember something along the lines of wanting to use it for new things and upgrading other stuff on an as-needed basis.

Goldziher: we discussed migrating everything in one go. For now we should keep everything consistent I think. 

Also - can you increase coverage?

PR #946, opened 2022-12-16
fix: support utf-8 encoded strings in form data

NiclasHaderer:
# PR Checklist

Currently UTF-8 form data cannot be parsed and results in an internal server error.  
I added a test which test that and included the fix.

### DIFF
--- a/starlite/parsers.py
+++ b/starlite/parsers.py
@@ -38,10 +38,10 @@ def parse_query_string(query_string: bytes, encoding: str = "utf-8") -> Tuple[Tu
     Returns:
         A tuple of key value pairs.
     """
-    _bools = {b"true": True, b"false": False, b"True": True, b"False": False}
+    _bools = {"true": True, "false": False, "True": True, "False": False}
     return tuple(
-        (k.decode(encoding), v.decode(encoding) if v not in _bools else _bools[v])
-        for k, v in parse_qsl(query_string, keep_blank_values=True)
+        (k, v if v not in _bools else _bools[v])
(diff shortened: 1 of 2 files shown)


Goldziher (commented):
Hi, thanks. 

The problem with the PR is that you are applying the fix on the query string parser. This is used for regular query strings as well, and your fix will hurt performance. 

Instead, move the fix to `parse_url_encoded_form_data`, this will be better performance wise.

NiclasHaderer:
So should I just create a second function which is only used for body parsing or what do you have in mind?

Goldziher:
> So should I just create a second function which is only used for body parsing or what do you have in mind?

Sure

starlite/parsers.py
```diff
@@ -35,19 +35,37 @@ def parse_query_string(query_string: bytes, encoding: str = "utf-8") -> Tuple[Tu
...
+        encoding: The encoding to use.
+
     Returns:
         A tuple of key value pairs.
     """
     _bools = {"true": True, "false": False, "True": True, "False": False}
     return tuple(
         (k, v if v not in _bools else _bools[v])
-        for k, v in parse_qsl(str(query_string, encoding=encoding), keep_blank_values=True, encoding=encoding)
+        for k, v in parse_qsl(str(form_query_string, encoding=encoding), keep_blank_values=True, encoding=encoding)
```
Goldziher, line 62: why do you need to pass encoding to `str`, and then again to `parse_qsl`?

NiclasHaderer: Good question. That's a mistake on my side.
I'll remove the encoding from `parse_qsl`

Goldziher:
@all-contributors add @NiclasHaderer code

PR #949, opened 2022-12-17
Move `JinjaTemplateEngine` and `MakoTemplateEngine` to contrib

Goldziher:
# PR Checklist


starlite/template/__init__.py
```diff
@@ -1,3 +1,38 @@
...
+
+        export: Any = JinjaTemplateEngine
+        module = "jinja"
+    elif name == "MakoTemplateEngine":
+        from starlite.contrib.mako import MakoTemplateEngine
+
+        export = MakoTemplateEngine
+        module = "mako"
+    else:
+        from starlite.contrib.mako import MakoTemplate
```
provinzkraut, line 26: I don't understand this part. If the imported name is anything besides `JinjaTemplateEngine` and `MakoTemplateEngine` it will default to `MakoTemplate`? Why?

starlite/template/__init__.py
```diff
@@ -1,3 +1,38 @@
...
+
+        export = MakoTemplate
+        module = "mako"
+
+    warn(
```
provinzkraut, line 31: We should probably add this to the deprecation utils, in case we want to customise / use them for some automation at some point.

starlite/template/__init__.py
Goldziher: It raises an exception a few lines above if the name is not in these three

starlite/template/__init__.py
Goldziher: I'd rather do that only when we actually need it.

starlite/template/__init__.py
provinzkraut: Yeah.. You're right. I guess I'm not fully awake yet. Nevermind then

starlite/template/__init__.py
provinzkraut: Hm. We could do that as well. I just think it would be nice to have consistent formatting of our deprecation messages.

Goldziher: Sure, I'll add

provinzkraut (approved):

PR #970, opened 2022-12-20
Fix 964: Jinja Template response ignores MediaType settings

provinzkraut:
Add support for setting a `media_type` in template responses.
If no media type is passed explicitly, try to infer the correct one using `mimetypes.guess_type` or fall back to `text/plain`. 

# PR Checklist

### DIFF
--- a/starlite/response/template.py
+++ b/starlite/response/template.py
@@ -1,3 +1,4 @@
+from pathlib import PurePath
 from typing import TYPE_CHECKING, Any, Dict, Optional, Union
 
 from starlite.enums import MediaType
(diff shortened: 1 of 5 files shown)


starlite/response/template.py
```diff
@@ -36,16 +39,28 @@ def __init__(
...
-            cookies: A list of [Cookie][starlite.datastructures.Cookie] instances to be set under the response 'Set-Cookie' header.
+            cookies: A list of [Cookie][starlite.datastructures.Cookie] instances to be set under the response
+                'Set-Cookie' header.
             encoding: Content encoding
+            media_type: A string or member of the [MediaType][starlite.enums.MediaType] enum. If not set, try to infer
+                the media type based on the template name. If this fails, fall back to `text/plain`.
         """
+        if media_type == MediaType.JSON:  # we assume this is the default
+            suffixes = PurePath(template_name).suffixes
+            for suffix in suffixes:
```
Goldziher, line 50: This code block can be polished i think.

Goldziher (approved):
Great

starlite/response/template.py
provinzkraut: I tried a few variations but I think this is the cleanest. Do you have any suggestions?

As I said in the issue, we have to assume `MediaType.JSON` is the default, since it will always be set - there's no way to receive `None` here, unless we change how / when we set default media types completely.

As for the suffixes part, `os.splitext` only returns the last suffix, which would break for common naming conventions such as `template.html.jinja2`. I could re-implement `PurePath.suffixes` here, but what would be the point of that?

Issue #958, opened 2022-12-17
Enhancement: Log tracebacks to console in debug mode


provinzkraut:
Currently exceptions are silenced when in debug mode and only show up in the responses.

The exception middleware should be changed such that when in debug mode, errors and their tracebacks are logged to the console.

JacobCoffee:
Assign this to me please, I'll attempt to fix this weekend

Goldziher:
done

PR #976, opened 2022-12-22
(958) Add exception logging in debug mode

JacobCoffee:
Re: Issue #958 

# PR Checklist


tests/middleware/test_exception_handler_middleware.py
```diff
@@ -118,3 +118,30 @@ async def after_exception_hook_handler(exc: Exception, scope: "Scope", state: "S
...
+
+
+def test_exception_handler_middleware_debug_logging() -> None:
+    from unittest.mock import patch
+
+    @get("/test")
+    def handler() -> None:
+        raise RuntimeError()
+
+    with create_test_client(handler) as client, patch("logging.Logger.debug") as mock_debug:
```
Goldziher, line 130: to enable logging you should pass a `LoggingConfig` to the app - you can do this with the `create_test_client` helper as well, simply pass `logging_config=LoggingConfig()`.

Goldziher:
so, wassup here? Why does this take so long?

JacobCoffee:
not 100% sure on the testing here, can someone verify it is checking correctly?

starlite/middleware/exceptions/middleware.py
```diff
@@ -47,6 +47,8 @@ async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> No
             await self.app(scope, receive, send)
         except Exception as e:  # pylint: disable=broad-except
             starlite_app = scope["app"]
+            if self.debug and (logger := starlite_app.logger):
+                logger.debug("Exception in ASGI application", exc_info=True)
```
Goldziher, line 51: I suggest to change this message: 

```suggestion
                logger.debug("exception raised for request to route %s", scope["path"], exc_info=True)
```

Maybe add more details from the scope context as well. Also - lets use lower case messages, which we use in the exception messages we raise as well.

Goldziher: note: https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

tests/middleware/test_exception_handler_middleware.py
```diff
@@ -118,3 +120,36 @@ async def after_exception_hook_handler(exc: Exception, scope: "Scope", state: "S
...
+        assert "Internal Server Error" in caplog.text
+        assert "ValueError" in response.text
+        assert "Test debug exception" in response.text
+
+    # Debug Off -> Exception is not logged
+    with caplog.at_level("INFO"), create_test_client(handler, logging_config=LoggingConfig()) as client:
+        client.app.debug = False
+        response = client.get("/test")
+        assert response.status_code == HTTP_500_INTERNAL_SERVER_ERROR
+        assert "Exception in ASGI application" not in caplog.text
```
Goldziher, line 144: since this was changed, this condition will alway be true. so you should update this line.

provinzkraut:
@JacobCoffee What's the status of this?

JacobCoffee:
I am working on a fix for the pytest issue via this
https://github.com/pytest-dev/pytest/issues/3697#issuecomment-792129636

peterschutt:
@JacobCoffee have you seen this fixture in the logging middleware tests?

https://github.com/starlite-api/starlite/blob/901596e566cd1b0ab902e67867300b7d8cf4e348/tests/middleware/test_logging_middleware.py#L29-L38

I'd say that's there to solve the same problem that you are hitting here, although there might be other reasons for it also. However, if you are looking for a low friction way to get this through, then you could move that fixture out of `tests/middleware/test_logging_middleware.py` and move it into `tests/middleware/conftest.py`, change the log level that is set in the fixture to debug, and set the logger returned from that callable on to your application logger for the test.

I've done this and have tested that the full suite passes, so I'll PR it into your fork and you can take a look.

If you do still want to go the route of implementing that capturing context manager pattern from the linked issue, you should perhaps look at whether we can apply that to the logging middleware tests where the current fixture-based workaround is being applied as well.

provinzkraut:
@peterschutt 

> I've done this and have tested that the full suite passes

This fixture was my first attempt as well, but it didn't seem to work as intended.


```python
@pytest.fixture
def get_logger():
    return LoggingConfig(
        handlers=default_handlers,
        loggers={
            "starlite": {"level": "DEBUG", "handlers": ["queue_listener"], "propagate": True},
        },
    ).configure()


@pytest.mark.parametrize(
    "debug,logging_config",
    [
        (True, LoggingConfig()),
        (False, LoggingConfig()),
        (False, None),
    ],
)
def test_exception_handler_middleware_debug_logging(
    get_logger, caplog: "LogCaptureFixture", debug: bool, logging_config: Optional[LoggingConfig]
) -> None:
    @get("/test")
    def handler() -> None:
        raise ValueError("Test debug exception")

    app = Starlite([handler], logging_config=logging_config, debug=debug)
    app.get_logger = get_logger

    with TestClient(app=app) as client, caplog.at_level("DEBUG"):

        response = client.get("/test")
        assert response.status_code == HTTP_500_INTERNAL_SERVER_ERROR
        assert "Test debug exception" in response.text

        if debug and logging_config:
            assert len(caplog.records) == 1
            # other assertions for the content of this record here
        else:
            assert not caplog.records
```

Still fails for me. How did you get it working?

Anyway, I tested the workaround @JacobCoffee linked, and it seems to me that it's a more stable solution. I think it's worth generally adopting that for our tests. Wdyt?

peterschutt:
I hit that too, but the code accesses `app.logger`, it doesn't call `app.get_logger()`. So, instead  of `app.get_logger = ...` I used `app.logger = get_logger("starlite")`.

peterschutt:
> I think it's worth generally adopting that for our tests. Wdyt?

I agree, but whether that is something that @JacobCoffee wants to do as part of this task, or we split it out to another issue, I'm not fussed.

starlite/middleware/exceptions/middleware.py
```diff
@@ -47,6 +47,8 @@ async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> No
             await self.app(scope, receive, send)
         except Exception as e:  # pylint: disable=broad-except
             starlite_app = scope["app"]
+            if self.debug and (logger := starlite_app.logger):
```
Goldziher, line 50:
```suggestion
            if self.debug and (logger := starlite_app.get_logge()):
```

no?

Goldziher (approved):

Goldziher:
@all-contributors add @JacobCoffee code

starlite/middleware/exceptions/middleware.py
JacobCoffee: Perhaps, but it broke my test 🙃 looking into why tonight hopefully

provinzkraut: Maybe because the function isn’t called `get_logge` but `get_logger`? 😉

JacobCoffee: i fixed that as soon as i committed the suggestion and realized 😛

peterschutt: If this change breaks the test, it most likely is because we are patching `app.logger` with response of calling the `get_logger` fixture in the test, whereas for this to work, we'd have to patch the `app.get_logger` attribute with the get_logger fixture itself.

In the starlite constructor, we both assign the get_logger callable and a logger instance to the application instance:

```python
        if self.logging_config:
            self.get_logger = self.logging_config.configure()
            self.logger = self.get_logger("starlite")
```
...and it wasn't obvious to me which one we should be using in this case either.

provinzkraut:
@Goldziher Why was this merged? Tests are not passing.

PR #982, opened 2022-12-24
Add support for auto generation of DTOs from response data

Goldziher:
# PR Checklist

### DIFF
--- a/starlite/dto.py
+++ b/starlite/dto.py
@@ -61,6 +61,7 @@ class DTO(GenericModel, Generic[T]):
 
     class Config(BaseConfig):
         arbitrary_types_allowed = True
+        orm_mode = True
 
     dto_source_model: ClassVar[Any]
     dto_field_mapping: ClassVar[Dict[str, str]]
@@ -127,13 +128,16 @@ class DTO(GenericModel, Generic[T]):
             Instance of source model type.
         """
         values = self.dict()
+
         for dto_key, original_key in self.dto_field_mapping.items():
             value = values.pop(dto_key)
             values[original_key] = value
+
         if self.dto_source_plugin is not None and self.dto_source_plugin.is_plugin_supported_type(
             self.dto_source_model
         ):
             return cast("T", self.dto_source_plugin.from_dict(model_class=self.dto_source_model, **values))
+
         # we are dealing with a pydantic model or dataclass
         return cast("T", self.dto_source_model(**values))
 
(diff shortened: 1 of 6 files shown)


dialvarezs:
It doesn't seem to be working at least with this MWE:
```python
from sqlalchemy import Column, Integer, String, select
from sqlalchemy.orm import Mapped, declarative_base

from starlite import DTOFactory, HTTPException, Starlite, get, post
from starlite.plugins.sql_alchemy import SQLAlchemyConfig, SQLAlchemyPlugin
from starlite.status_codes import HTTP_404_NOT_FOUND

Base = declarative_base()

sqlalchemy_config = SQLAlchemyConfig(
    connection_string="sqlite+aiosqlite:///test.sqlite", dependency_key="async_session"
)
sqlalchemy_plugin = SQLAlchemyPlugin(config=sqlalchemy_config)
dto_factory = DTOFactory(plugins=[sqlalchemy_plugin])


class User(Base):
    __tablename__ = "user"
    id: Mapped[int] = Column(Integer, primary_key=True)
    name: Mapped[str] = Column(String)
    secret: Mapped[str] = Column(String)

    class Config:
        orm_mode = True


CreateUserDTO = dto_factory("CreateUserDTO", User, exclude=["id"])
ReadUserDTO = dto_factory(name="ReadUserDTO", source=User, exclude=["secret"])


async def on_startup() -> None:
    """Initialize the database."""
    async with sqlalchemy_config.engine.begin() as conn:
        await conn.run_sync(Base.metadata.create_all)


@post(path="/users")
async def create_user(
    data: CreateUserDTO,
    async_session: AsyncSession,
) -> ReadUserDTO:
    """Create a new user and return it."""
    user: User = data.to_model_instance()
    async_session.add(user)
    await async_session.commit()
    return user


@get(path="/users/{user_id:int}")
async def get_user(user_id: str, async_session: AsyncSession) -> ReadUserDTO:
    """Get a user by its ID and return it.

    If a user with that ID does not exist, return a 404 response
    """
    result = await async_session.scalars(select(User).where(User.id == user_id))
    user: User | None = result.one_or_none()
    if not user:
        raise HTTPException(
            detail=f"User with ID {user_id} not found",
            status_code=HTTP_404_NOT_FOUND,
        )
    return user


@get(path="/users/")
async def get_users(async_session: AsyncSession) -> list[ReadUserDTO]:
    """Get all users."""
    result = await async_session.scalars(select(User))
    users: list[User] = result.all()
    return users


app = Starlite(
    route_handlers=[create_user, get_user, get_users],
    on_startup=[on_startup],
    plugins=[sqlalchemy_plugin],
)
```

Specifically, the DTO subclass checking is failing, `issubclass(return_annotation, DTO[Any])` is `False` everytime.

Goldziher:
@dialvarezs - thanks for this. I incorporated this in a test - and it works fine. Are you using SQLA 2.0b? Perhaps there is some difference there, but with v1.4.5 its working as expected. I created an issue to extend our support to v2.0 officially, which will allow us to execute the tests against v2.0 as well.

dialvarezs:
Strange, I was using 1.4.45 as well...
But with the last changes you made is working for me as well.

So, it's ok for me then!

dialvarezs (approved):

PR #995, opened 2022-12-27
Support layered `type_encoders`

provinzkraut:
# PR Checklist

### DIFF
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -68,6 +68,7 @@ from starlite.types import (
     ResponseCookies,
     ResponseHeadersMap,
     ResponseType,
+    TypeEncodersMap,
 )
 from starlite.utils import Ref, annotation_is_iterable_of_type, is_async_callable
 from starlite.utils.predicates import is_class_and_subclass
@@ -190,6 +191,7 @@ def _create_data_handler(
     response_class: "ResponseType",
     return_annotation: Any,
     status_code: int,
+    type_encoders: Optional[TypeEncodersMap],
 ) -> "AsyncAnyCallable":
     """Create a handler function for arbitrary data."""
     normalized_headers = [
@@ -206,6 +208,7 @@ def _create_data_handler(
             content=data,
             media_type=media_type,
             status_code=status_code,
+            type_encoders=type_encoders,
         )
         response.raw_headers = raw_headers
 
@@ -293,6 +296,7 @@ class HTTPRouteHandler(BaseRouteHandler["HTTPRouteHandler"]):
         "deprecated",
         "description",
         "etag",
+        "has_sync_callable",
         "http_methods",
         "include_in_schema",
         "media_type",
@@ -309,7 +313,7 @@ class HTTPRouteHandler(BaseRouteHandler["HTTPRouteHandler"]):
         "sync_to_thread",
         "tags",
         "template_name",
-        "has_sync_callable",
+        "type_encoders",
     )
 
     has_sync_callable: bool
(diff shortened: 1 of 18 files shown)


starlite/handlers/http.py
```diff
@@ -563,6 +570,20 @@ def resolve_after_response(self) -> Optional["AfterResponseHookHandler"]:
 
         return cast("Optional[AfterResponseHookHandler]", self._resolved_after_response)
 
+    def resolve_type_encoders(self) -> Optional[TypeEncodersMap]:
+        """Resolve `type_encoders` by merging existing `type_encoders` from all layers.
+
+        Returns:
+            A `TypeEncodersMap` to use for this response or `None`
+        """
+        type_encoders: Optional[TypeEncodersMap] = None
```
Goldziher, line 579: you can default this to be {} and do an OR in the return statement.

Goldziher (approved):

starlite/types/composite_types.py
```diff
@@ -59,3 +59,4 @@
     Callable[..., ASGIApp], DefineMiddleware, Iterator[Tuple[ASGIApp, Dict[str, Any]]], Type[MiddlewareProtocol]
 ]
 MaybePartial = Union[T, partial]
+TypeEncodersMap = Dict[Any, Callable[[Any], Any]]
```
Goldziher, line 62: Should the callables return any or bytes?

Goldziher (approved):

starlite/types/composite_types.py
provinzkraut: `Any` is correct here. They only transform values into types that `msgspec` can understand, they don't serialize anything themselves.

starlite/response/base.py
```diff
@@ -119,6 +125,40 @@ def __init__(
...
+        "1.48.0",
+        pending=True,
+        info="Set `type_encoders` on a higher layer "
+        "instead or pass `type_encoders` to the `Response` instance directly",
+    )
+    def serializer(cls, value: Any) -> Any:
+        """Transform non-natively supported types into supported types.
+
+        Should raise `TypeError` if a type cannot be transformed into a supported type
+        """
```
peterschutt, line 139: Hey @provinzkraut - as this is meant to be a deprecation, shouldn't this still return a serialized value?

provinzkraut: Nope. This method isn’t actually being used anymore by the serializer, unless you override it in a subclass. Look at the `get_serializer` method.

This is done to provide an easier upgrade path for us, avoid passing around the default serializer, and a slight performance benefit.

peterschutt: But `Response.serializer()` is a documented method, and this changes its behavior.

E.g., this worked fine in the previous version:

```python
def default_serializer(value: Any) -> Any:
    """Serialize `value`.
    Args:
        value: To be serialized.
    Returns:
        Serialized representation of `value`.
    """
    if isinstance(value, pgproto.UUID):
        return str(value)
    return Response[Any].serializer(value)
```
Not a big issue, as now that I've pinpointed what is wrong I'm updating to the layered type encoders (which is a much better pattern), just wanted to check that this isn't an over site.

Smth similar going on in `starlite-pg-redis-docker` app, which I'll update now in due course.

provinzkraut: Ah, you’re right. I actually hadn’t thought of this case. Guess we can just revert to calling `default_serializer` in `serializer` to re-enable this. 
I’ll fix this tomorrow!

provinzkraut: @peterschutt Should be fixed by #1018

peterschutt: Thanks!

provinzkraut: Just needs a review so I can merge it :wink:

PR #1005, opened 2022-12-29
Sort tags in open api schema

### LINKED ISSUE #1004: Bug: openapi render for multiple tags isn't consistent
**Describe the bug**
When the openapi renders tags from both a controller and a route it is not deterministic. This may not be a bug? But it surprised me so thought I'd raise it.

I'm unsure if I'm doing something crazy but for a project, we check in the generated json openapi schema so we can browse the API live in gitlab. I've recently added a tag to both a controller and a route in it. But because the order of the tags isn't consistent they are going to keep flip flopping as we have a pre-commit that generates the json to make sure it's up to date. I hope that ramble makes sense...

**To Reproduce**
```python
from typing import Dict                                           
                                                                  
from starlite import Starlite, Controller, get                    
                                                                  
class TestController(Controller):                                 
    tags = ["a"]                                                  
                                                                  
    @get("/", tags=["b"])                                         
    def hello_world(self) -> Dict[str, str]:                      
        return {"hello": "world"}                                 
                                                                  
                                                                  
app = Starlite(route_handlers=[TestController])                   
print(app.openapi_schema.paths["/"].get.tags)                     
```
If you run that multiple times, you will see you get either:
```python
['a', 'b']
```
or
```python
['b', 'a']
```

**Additional context**
I believe the problem is [here](https://github.com/starlite-api/starlite/blob/835749112e8364c1516f45973c924774aca22ca9/starlite/openapi/path_item.py#L59) as it forces construction of a new set. Sorting them before returning would be viable as there shouldn't be _too many_ tags and it's a one time thing I believe?

But as I said, it may not be a problem you care about as I could be doing something silly.

ste-pool:
# PR Checklist

Fixes #1004

### DIFF
--- a/starlite/openapi/path_item.py
+++ b/starlite/openapi/path_item.py
@@ -56,7 +56,7 @@ def extract_layered_values(
             tags.extend(layer.tags)
         if layer.security:
             security.extend(layer.security)
-    return list(set(tags)) if tags else None, security or None
+    return sorted(set(tags)) if tags else None, security or None
 
 
 def create_path_item(
--- a/tests/openapi/test_tags.py
+++ b/tests/openapi/test_tags.py
@@ -23,7 +23,7 @@ def controller() -> Type[Controller]:
         path = "/controller"
         tags = ["controller"]
 
-        @get(tags=["handler"])
+        @get(tags=["handler", "a"])
         def _handler(self) -> Any:
             ...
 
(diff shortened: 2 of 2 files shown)


Goldziher (approved):

Goldziher:
@all-contributors add @ste-pool code

PR #1007, opened 2022-12-30
Feature/open api unique operation

maggias:
# PR Checklist

According to OpenAPI specification, operationId must be unique among all operations described in the API, https://spec.openapis.org/oas/v3.1.0#fixed-fields-7. Currently this breaks when array of http_method is supplied in HTTPRouteHandler decorator. This can also break if same method names are used in different routes.

This for instance causes an error when using openapi-generator to generate client stubs from schema.

This proposal adds the method name when multiple http_methods are used and also adds string paths from path_component as a prefix to make the operationId unique. It also formats operationId by using _ instead of " " to follow common programming naming conventions as suggested in the OpenAPI specification.

### DIFF
--- a/docs/usage/12-openapi/2-route-handler-configuration.md
+++ b/docs/usage/12-openapi/2-route-handler-configuration.md
@@ -23,7 +23,7 @@ You can also modify the generated schema for the route handler using the followi
 - `description`: Text used for the route's schema _description_ section.
 - `response_description`: Text used for the route's response schema _description_ section.
 - `operation_id`: An identifier used for the route's schema _operationId_. Defaults to the `__name__` attribute of the
-  wrapped function.
+  wrapped function. Will be prefixed with the method name when function is decorated with `HTTPRouteHandler` and multiple `http_method`. Will also be prefixed with path strings used in `Routers` and `Controllers` to make sure id is unique.
 - `deprecated`: A boolean dictating whether this route should be marked as deprecated in the OpenAPI schema. Defaults
   to `False`.
 - `raises`: A list of exception classes extending from `starlite.HttpException`. This list should describe all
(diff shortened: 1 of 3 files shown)


docs/usage/12-openapi/2-route-handler-configuration.md
```diff
@@ -23,7 +23,7 @@ You can also modify the generated schema for the route handler using the followi
 - `description`: Text used for the route's schema _description_ section.
 - `response_description`: Text used for the route's response schema _description_ section.
 - `operation_id`: An identifier used for the route's schema _operationId_. Defaults to the `__name__` attribute of the
-  wrapped function.
+  wrapped function. Will be prefixed with the method name when function is decorated with `HTTPRouteHandler` and multiple `http_method`. Will also be prefixed with path strings used in `Routers` and `Controllers` to make sure id is unique.
```
Goldziher, line 26: this is not the place for this - this is a defintion of what this parameter is. The explanation should appear using a note with admonition (search for `!!! note` to see examples).

starlite/openapi/path_item.py
```diff
@@ -87,7 +111,7 @@ def create_path_item(
 
             tags, security = extract_layered_values(route_handler)
             operation = Operation(
-                operationId=route_handler.operation_id or handler_name,
+                operationId=operation_id,
```
Goldziher, line 114: the above logic is ok - but, the standard in OA is to use camelCase rather than snake case. Can you switch to using this or do you think differently?

docs/usage/12-openapi/2-route-handler-configuration.md
maggias: Added as a note after the list of parameters

starlite/openapi/path_item.py
maggias: Sounds good, updated

Goldziher (approved):
Cool stuff!

Goldziher:
@all-contributors add @maggias code

PR #1011, opened 2022-12-30
Extend serializable types

### LINKED ISSUE #1003: Bug: Custom data types with validators cannot be serialized since 1.45
**Describe the bug**
Since version 1.45 custom data types with pydantic validators cannot be serialized anymore. Deserialization works fine.

**To Reproduce**
Example app that works in 1.44, but doesn't in 1.45+:
Note that the print gets executed.
```python
from starlite import Starlite, post, Request, Response, LoggingConfig
from starlite.utils import create_exception_response


class CustomType(str):
    @classmethod
    def __get_validators__(cls):
        yield cls.get_validators_must_yield_something_so_here_we_are

    @classmethod
    def get_validators_must_yield_something_so_here_we_are(cls, value):
        return cls(value)


class Model(BaseModel):
    foo: CustomType


@post('/foo')
def foo(data: Model) -> Model:
    print(data)
    return data


def logging_exception_handler(request: Request, exc: Exception) -> Response:
    request.logger.error("Application Exception", exc_info = exc)
    return create_exception_response(exc)


app = Starlite(
    [foo],
    logging_config = LoggingConfig(),
    exception_handlers={Exception: logging_exception_handler},
)
```

**Additional context**
This is the stacktrace that I got with the example above:
```
  File ".../python3.11/site-packages/starlite/response/base.py", line 216, in render
    return encode_json(content, self.serializer)
  File ".../python3.11/site-packages/starlite/utils/serialization.py", line 59, in encode_json
    return _msgspec_json_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.11/site-packages/starlite/utils/serialization.py", line 24, in default_serializer
TypeError: Unsupported type: <class 'app2.CustomType'>

  File ".../python3.11/site-packages/starlite/middleware/exceptions/middleware.py", line 47, in __call__
  File ".../python3.11/site-packages/starlite/routes/http.py", line 75, in handle
  File ".../python3.11/site-packages/starlite/routes/http.py", line 127, in _get_response_for_request
    response = await self._call_handler_function(
  File ".../python3.11/site-packages/starlite/routes/http.py", line 163, in _call_handler_function
    else await route_handler.to_response(
  File ".../python3.11/site-packages/starlite/handlers/http.py", line 606, in to_response
    return await response_handler(app=app, data=data, plugins=plugins, request=request)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.11/site-packages/starlite/handlers/http.py", line 203, in handler
    response = response_class(
               ^^^^^^^^^^^^^^^
  File ".../python3.11/site-packages/starlite/response/base.py", line 109, in __init__
    self.body = content if isinstance(content, bytes) else self.render(content)
                                                           ^^^^^^^^^^^^^^^^^^^^
  File ".../python3.11/site-packages/starlite/response/base.py", line 218, in render
    raise ImproperlyConfiguredException("Unable to serialize response content") from e
starlite.exceptions.http_exceptions.ImproperlyConfiguredException: 500: Unable to serialize response content
```

provinzkraut:
Fixes #1003, by adding support for subclassed stdlib types. Also simplifies some of the pydantic-specific type conversion logic. 

# PR Checklist

### DIFF
--- a/starlite/utils/serialization.py
+++ b/starlite/utils/serialization.py
@@ -1,27 +1,30 @@
+from collections import deque
+from decimal import Decimal
+from ipaddress import (
+    IPv4Address,
+    IPv4Interface,
+    IPv4Network,
+    IPv6Address,
+    IPv6Interface,
+    IPv6Network,
+)
 from pathlib import PurePosixPath
+from re import Pattern
 from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
 import msgspec
 from pydantic import (
-    AnyUrl,
     BaseModel,
     ByteSize,
     ConstrainedBytes,
     ConstrainedDate,
     ConstrainedDecimal,
-    ConstrainedFloat,
-    ConstrainedFrozenSet,
-    ConstrainedInt,
-    ConstrainedList,
-    ConstrainedSet,
-    ConstrainedStr,
-    EmailStr,
     NameEmail,
-    PaymentCardNumber,
     SecretField,
     StrictBool,
 )
 from pydantic.color import Color
+from pydantic.json import decimal_encoder
 
 if TYPE_CHECKING:
     from starlite.types import TypeEncodersMap
@@ -31,22 +34,34 @@ DEFAULT_TYPE_ENCODERS: "TypeEncodersMap" = {
     # pydantic specific types
     BaseModel: lambda m: m.dict(),
     ByteSize: lambda b: b.real,
-    EmailStr: str,
     NameEmail: str,
     Color: str,
-    AnyUrl: str,
     SecretField: str,
-    ConstrainedInt: int,
-    ConstrainedFloat: float,
-    ConstrainedStr: str,
     ConstrainedBytes: lambda b: b.decode("utf-8"),
-    ConstrainedList: list,
-    ConstrainedSet: set,
-    ConstrainedFrozenSet: frozenset,
     ConstrainedDecimal: float,
     ConstrainedDate: lambda d: d.isoformat(),
-    PaymentCardNumber: str,
-    StrictBool: int,  # pydantic compatibility
+    IPv4Address: str,
+    IPv4Interface: str,
+    IPv4Network: str,
+    IPv6Address: str,
+    IPv6Interface: str,
+    IPv6Network: str,
+    # pydantic compatibility
+    deque: list,
+    Decimal: decimal_encoder,
+    StrictBool: int,
+    Pattern: lambda o: o.pattern,
+    # support subclasses of stdlib types, e.g. pydantic's constrained types.
+    # see https://github.com/jcrist/msgspec/issues/248
+    # and https://github.com/starlite-api/starlite/issues/1003
+    str: str,
+    int: int,
+    float: float,
+    list: list,
+    tuple: tuple,
+    set: set,
+    frozenset: frozenset,
+    dict: dict,
 }
 
 
--- a/starlite/utils/serialization.py
+++ b/starlite/utils/serialization.py
@@ -51,7 +51,9 @@ DEFAULT_TYPE_ENCODERS: "TypeEncodersMap" = {
     Decimal: decimal_encoder,
     StrictBool: int,
     Pattern: lambda o: o.pattern,
-    # support subclasses of stdlib types, e.g. pydantic's constrained types.
+    # support subclasses of stdlib types, e.g. pydantic's constrained types. If no
+    # previous type matched, these will be the last type in the mro, so we use this to
+    # (attempt to) convert a subclass into its base class.
     # see https://github.com/jcrist/msgspec/issues/248
     # and https://github.com/starlite-api/starlite/issues/1003
     str: str,
--- a/tests/utils/test_serialization.py
+++ b/tests/utils/test_serialization.py
@@ -35,6 +35,34 @@ from tests import PersonFactory
 person = PersonFactory.build()
 
 
+class CustomStr(str):
+    pass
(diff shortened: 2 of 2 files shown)


starlite/utils/serialization.py
```diff
@@ -31,22 +34,36 @@
     # pydantic specific types
     BaseModel: lambda m: m.dict(),
     ByteSize: lambda b: b.real,
-    EmailStr: str,
     NameEmail: str,
     Color: str,
-    AnyUrl: str,
     SecretField: str,
-    ConstrainedInt: int,
```
provinzkraut, line 39: We don't need all those anymore. Since they all inherit from primitive type, they will be converted into it by the compat types at the end of this mapping.

starlite/utils/serialization.py
```diff
@@ -31,22 +34,36 @@
...
-    ConstrainedStr: str,
     ConstrainedBytes: lambda b: b.decode("utf-8"),
-    ConstrainedList: list,
-    ConstrainedSet: set,
-    ConstrainedFrozenSet: frozenset,
     ConstrainedDecimal: float,
     ConstrainedDate: lambda d: d.isoformat(),
-    PaymentCardNumber: str,
-    StrictBool: int,  # pydantic compatibility
+    IPv4Address: str,
```
provinzkraut, line 43: These were missing before

cofin (approved):
LGTM

starlite/utils/serialization.py
```diff
@@ -31,22 +34,36 @@
...
+    # see https://github.com/jcrist/msgspec/issues/248
+    # and https://github.com/starlite-api/starlite/issues/1003
+    str: str,
+    int: int,
+    float: float,
+    list: list,
+    tuple: tuple,
+    set: set,
+    frozenset: frozenset,
+    dict: dict,
```
jcrist, line 66: Since subclasses of `dict`, `list`, and `tuple` are handled natively, those keys will never be hit here and can be removed. I'm also happy to add native support for subclasses of `set` and `frozenset` for the same reasons I gave before (but for now we won't support `str`/`int`/`float` subclasses).

starlite/utils/serialization.py
```diff
@@ -31,22 +34,36 @@
...
     Color: str,
-    AnyUrl: str,
     SecretField: str,
-    ConstrainedInt: int,
-    ConstrainedFloat: float,
-    ConstrainedStr: str,
     ConstrainedBytes: lambda b: b.decode("utf-8"),
-    ConstrainedList: list,
-    ConstrainedSet: set,
-    ConstrainedFrozenSet: frozenset,
```
jcrist, line 45: When would a user ever end up with an instance of one of these pydantic types? AFAICT these types are for type annotations only, during decode I don't ever see pydantic returning concrete instances of them:

```python
In [1]: import pydantic

In [2]: class Test(pydantic.BaseModel):
   ...:     x: pydantic.constr(min_length=3)
   ...: 

In [3]: type(Test(x="12345").x)  # x is a str, not a ConstrainedStr
Out[3]: str

In [4]: type(Test.parse_raw(b'{"x": "12345"}').x)
Out[4]: str
```

starlite/utils/serialization.py
```diff
@@ -31,22 +34,36 @@
...
+    IPv4Address: str,
+    IPv4Interface: str,
+    IPv4Network: str,
+    IPv6Address: str,
+    IPv6Interface: str,
+    IPv6Network: str,
+    # pydantic compatibility
+    deque: list,
+    Decimal: decimal_encoder,
+    StrictBool: int,
```
jcrist, line 52: `StrictBool` is serialized as an int, not a boolean value?

provinzkraut: A `StrictBool` instance is serialised as an `int`, a field annotated as `StrictBool` itself is not. This is mostly just for the case when `StrictBool` is used directly as a value.

starlite/utils/serialization.py
provinzkraut: This is just to support the edge case of using instances of constrained classes. See my comment below.

Native support for subclassed `set` and `frozenset` would be very much welcome though!

starlite/utils/serialization.py
provinzkraut: > When would a user ever end up with an instance of one of these pydantic types?

This can only happen if the constrained value is being instantiated directly. I admit it's an edge case, and one could argue that it shouldn't be done, but I've seen it a few times in the wild.

starlite/utils/serialization.py
jcrist: I understand that, I'm just saying that because `dict`, `list`, and `tuple` (and in the next release `set`/`frozenset`) are natively supported, those entries in this mapping aren't necessary because a subclass of these will never result in `enc_hook` being called.

starlite/utils/serialization.py
jcrist: Ah, `StrictBool` is a subclass of `int` not `bool`. My mistake, thanks for the clarification.

starlite/utils/serialization.py
provinzkraut: You know what, you're right. I previously thought that there's some edge case where this was needed, but I can't reproduce it anymore, so I guess it was a fluke on my side. I'll remove this. Thanks for the feedback here!

PR #1034, opened 2023-01-05
Add examples about background tasks usage with responses

pomma89:
# PR Checklist

The reason why I opened this PR is that I did not understand how to pass route handler arguments to background tasks, but discussion #765 was helpful, and I thought that adding a few examples, as suggested within the discussion, might help other users.

This PR contains the following items:

- Moved existing example into a separate file.
- Added tests for new and previous examples.
- Name of kwarg was incorrectly documented as `background_task`, it should be `background`, if I am not mistaken.
- Given more visibility to new examples, because I thought that the use case in which a background task has to receive route handler arguments might be more common. I also emphasized the fact that background tasks passed into route decorators cannot use handler arguments.

Of course, I am open to any feedback, in particular about the fourth point.

Thank you very much,
Goodbye,
Alessio

### DIFF
--- a/docs/usage/5-responses/11-background-tasks.md
+++ b/docs/usage/5-responses/11-background-tasks.md
@@ -11,21 +11,8 @@ that will be called after the response finishes sending the data.
 
 Thus, in the following example the passed in background task will be executed after the response sends:
 
-```python
-import logging
-
-from starlite import BackgroundTask, get
-
-logger = logging.getLogger(__name__)
-
-
-async def logging_task(identifier: str, message: str) -> None:
-    logger.info(f"{identifier}: {message}")
-
-
-@get("/", background=BackgroundTask(logging_task, "greeter", message="was called"))
-def greeter() -> dict[str, str]:
-    return {"hello": "world"}
+```py title="Background Task Set in Decorator"
+--8<-- "examples/responses/background_tasks_2.py"
 ```
 
 When the `greeter` handler is called, the logging task will be called with any `*args` and `**kwargs` passed into the
--- a/docs/usage/5-responses/11-background-tasks.md
+++ b/docs/usage/5-responses/11-background-tasks.md
@@ -1,9 +1,8 @@
 # Background Tasks
 
-All Starlite responses and response containers (e.g. `File`, `Template` etc.) allow passing in a `background`
+All Starlite responses and response containers (e.g. `File`, `Template`, etc.) allow passing in a `background`
 kwarg. This kwarg accepts either an instance of [`BackgroundTask`][starlite.datastructures.background_tasks.BackgroundTask]
-or
-an instance of [`BackgroundTasks`][starlite.datastructures.background_tasks.BackgroundTasks], which wraps an iterable
+or an instance of [`BackgroundTasks`][starlite.datastructures.background_tasks.BackgroundTasks], which wraps an iterable
 of [`BackgroundTask`][starlite.datastructures.background_tasks.BackgroundTask] instances.
 
 A background task is a sync or async callable (function, method or class that implements the `__call__` dunder method)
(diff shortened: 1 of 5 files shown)


examples/responses/background_tasks_1.py
```diff
@@ -0,0 +1,20 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def logging_task(identifier: str, message: str) -> None:
+    logger.info("%s: %s", identifier, message)
+
+
+@get("/")
+def greeter(name: str) -> Response[dict[str, str]]:
```
provinzkraut, line 13: Use Python 3.8 compatible type annotations:

```suggestion
def greeter(name: str) -> Response[Dict[str, str]]:
```

examples/responses/background_tasks_1.py
```diff
@@ -0,0 +1,20 @@
+import logging
```
provinzkraut, line 1:
```suggestion
from typing import Dict

import logging
```

examples/responses/background_tasks_2.py
```diff
@@ -0,0 +1,17 @@
+import logging
```
provinzkraut, line 1:
```suggestion
from typing import Dict

import logging
```

examples/responses/background_tasks_2.py
```diff
@@ -0,0 +1,17 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def logging_task(identifier: str, message: str) -> None:
+    logger.info("%s: %s", identifier, message)
+
+
+@get("/", background=BackgroundTask(logging_task, "greeter", message="was called"))
+def greeter() -> dict[str, str]:
```
provinzkraut, line 13:
```suggestion
def greeter() -> Dict[str, str]:
```

examples/responses/background_tasks_3.py
```diff
@@ -0,0 +1,30 @@
+import logging
```
provinzkraut, line 1:
```suggestion
from typing import Dict

import logging
```

examples/responses/background_tasks_3.py
```diff
@@ -0,0 +1,30 @@
...
+async def logging_task(name: str) -> None:
+    logger.info("%s was greeted", name)
+
+
+async def saving_task(name: str) -> None:
+    greeted.add(name)
+
+
+@get("/")
+def greeter(name: str) -> Response[dict[str, str]]:
```
provinzkraut, line 18:
```suggestion
def greeter(name: str) -> Response[Dict[str, str]]:
```

examples/tests/responses/test_background_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+
+from examples.responses.background_tasks_1 import app as app_1
+from examples.responses.background_tasks_2 import app as app_2
+from examples.responses.background_tasks_3 import app as app_3
+from examples.responses.background_tasks_3 import greeted as greeted_3
+from starlite import TestClient
+
+
+def test_background_tasks_1() -> None:
+    with TestClient(app=app_1) as client, patch("examples.responses.background_tasks_1.logger.info") as mock_info:
```
provinzkraut, line 11: Use the [`caplog`](https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture) fixture here instead of mocking. Applies to the rest of the tests as well.

docs/usage/5-responses/11-background-tasks.md
```diff
@@ -1,48 +1,51 @@
...
+`BackgroundTask`.
 
-async def logging_task(identifier: str, message: str) -> None:
-    logger.info(f"{identifier}: {message}")
+!!! note
+    In the above example `"greeter"` is an arg and `message=f"was called with name {name}"` is a kwarg.
+    The function signature of `logging_task` allows for this, so this should pose no problem.
+    Starlite uses [`ParamSpec`][typing.ParamSpec] to ensure that a
+    [`BackgroundTask`][starlite.datastructures.background_tasks.BackgroundTask] is properly typed,
+    so it will get type checking for any passed in args and kwargs.
```
provinzkraut, line 25:
```suggestion
    [`BackgroundTask`][starlite.datastructures.background_tasks.BackgroundTask] is typed with [`ParamSpec`][typing.ParamSpec], enabling correct type checking for arguments and keyword arguments passed to it.
```

docs/usage/5-responses/11-background-tasks.md
```diff
@@ -1,48 +1,51 @@
...
+```
 
-You can also use the [`BackgroundTasks`][starlite.datastructures.background_tasks.BackgroundTasks] class instead, and pass
-to it an iterable (list, tuple etc.) of [`BackgroundTask`][starlite.datastructures.background_tasks.BackgroundTask]
-instances. This class accepts one optional kwargs aside from the tasks - `run_in_task_group`, which is a boolean flag
-that defaults to `False`. If you set this value to `True` than the tasks will run concurrently, using
+[`BackgroundTasks`][starlite.datastructures.background_tasks.BackgroundTasks] class
+accepts one optional kwargs aside from the tasks, `run_in_task_group`, which is a boolean flag
+that defaults to `False`. If you set this value to `True`, then the tasks will run concurrently, using
 an [`anyio.task_group`](https://anyio.readthedocs.io/en/stable/tasks.html).
```
provinzkraut, line 49:
```suggestion
accepts an optional keyword argument `run_in_task_group` with a default value of `False`. Setting this to `True` allows background tasks to run concurrently, using an [`anyio.task_group`](https://anyio.readthedocs.io/en/stable/tasks.html). 
```

provinzkraut (changes requested):
Looks good overall, just some minor issues.

examples/tests/responses/test_background_tasks.py
pomma89: Done in latest commit. Thank you for this suggestion, I did not know about `caplog` 👍

provinzkraut (approved):

provinzkraut:
@all-contributors add @pomma89 doc

PR #1039, opened 2023-01-05
Improve documentation SEO by updating `mike deploy` config

ThinksFast:
Sorry for all the commits on this, kinda messy, it took some work to get docs building on my fork.  

The code changes I'm pushing here are minimal:

- Added  `--no-redirect --title "$STARLITE_VERSION (latest)"` to the mike deploy statement.  This will produce static URLs for the `/latest/*` path, and it will append `(latest)` to the newest version of the docs.  i.e. `1.48 (latest)` will show in the docs website.  Small UI tweak for our users. 
- Removed `alias_type: symlink` from `mkdocs.yaml`.  That param doesn't work on our version of `mike`, and throws a warning during build. 

image

# PR Checklist


provinzkraut:
Just a question: Does the `--no-redirect` option have the same effect `alias_type: copy` would have had? Are we getting the same SEO benefits from that?

provinzkraut (approved):

provinzkraut:
@all-contributors add @ThinksFast doc

ThinksFast:
> Just a question: Does the `--no-redirect` option have the same effect `alias_type: copy` would have had? Are we getting the same SEO benefits from that?

Yep!  Turns out `alias_type` is only available on the code in their `main` branch, it wasn't part of the latest release.  But I confirmed with the maintainer of `mike`:  `alias_type: copy` does the same thing as specifying `latest` as an alias and adding `--no-redirect` to the deploy statement.  So this config achieves our immediate goals of having static URLs on the `/latest/*` path.  

One last comment: the maintainer of `mike` questioned whether we need `--force` in the deploy statement.  It was already there, so I left it, but worth discussing. 

> I'd strongly recommend against using --force unless you're 100% sure you need it, for the same reason you generally wouldn't use git push --force. It's usually a sign that you're doing something risky, if not outright wrong.

PR #1040, opened 2023-01-05
Fix issue when response is an empty iterable of DTOs

abdulhaq-e:
# PR Checklist

The implementation of the automatic `DTO` response converter has a small bug. If the response type is an iterable DTO, a check is made weather the first element in the iterable data is a `DTO`. This check will result in an `Indexerror` as the iterable data might have no elements at all.

### DIFF
--- a/starlite/handlers/http.py
+++ b/starlite/handlers/http.py
@@ -224,7 +224,7 @@ def _create_data_handler(
         if is_dto_annotation and not isinstance(data, DTO):
             data = return_annotation(**data) if isinstance(data, dict) else return_annotation.from_model_instance(data)
 
-        elif is_dto_iterable_annotation and not isinstance(data[0], DTO):  # pyright: ignore
+        elif is_dto_iterable_annotation and len(data) and not isinstance(data[0], DTO):  # pyright: ignore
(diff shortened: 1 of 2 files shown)


starlite/handlers/http.py
```diff
@@ -224,7 +224,7 @@ async def handler(data: Any, plugins: List["PluginProtocol"], **kwargs: Any) ->
         if is_dto_annotation and not isinstance(data, DTO):
             data = return_annotation(**data) if isinstance(data, dict) else return_annotation.from_model_instance(data)
 
-        elif is_dto_iterable_annotation and not isinstance(data[0], DTO):  # pyright: ignore
+        elif is_dto_iterable_annotation and len(data) and not isinstance(data[0], DTO):  # pyright: ignore
```
provinzkraut, line 227:
```suggestion
        elif is_dto_iterable_annotation and data and not isinstance(data[0], DTO):  # pyright: ignore
```

provinzkraut (changes requested):
Looks good, just one small change request!

starlite/handlers/http.py
abdulhaq-e: Done. Bad habits from other languages ^_^

PR #1046, opened 2023-01-06
Split starlite/cli.py into starlite/cli + submodules

provinzkraut:
# PR Checklist

### DIFF
--- a/starlite/__main__.py
+++ b/starlite/__main__.py
@@ -1,3 +1,3 @@
-from starlite.cli.main import starlite_group
+from starlite.cli.main import starlite_group  # pragma: no cover
 
-starlite_group()  # pylint: disable=E1120
+starlite_group()  # pylint: disable=E1120  # pragma: no cover
--- a/starlite/__main__.py
+++ b/starlite/__main__.py
@@ -1,0 +1,3 @@
+from starlite.cli.main import starlite_group
+
+starlite_group()  # pylint: disable=E1120
--- a/starlite/cli/main.py
+++ b/starlite/cli/main.py
@@ -1,0 +1,145 @@
+import inspect
+from typing import Optional
+
+from click import Context, argument, group, option, pass_context
+from rich.prompt import Confirm
+from rich.tree import Tree
+
+from starlite import HTTPRoute, Starlite, WebSocketRoute
+from starlite.utils.helpers import unwrap_partial
+
+from .utils import (
+    StarliteCLIException,
+    StarliteEnv,
+    StarliteExtensionGroup,
+    console,
+    get_session_backend,
+    show_app_info,
+)
+
+
+@group(cls=StarliteExtensionGroup)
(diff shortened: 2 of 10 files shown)


starlite/cli/main.py
```diff
@@ -0,0 +1,145 @@
...
+                route_type = "WS"
+            else:
+                route_type = "ASGI"
+            branch = tree.add(f"[green]{route.path}[/green] ({route_type})")
+            branch.add(f"[blue]{route.route_handler.name or route.route_handler.handler_name}[/blue]")
+
+    console.print(tree)
+
+
+@cli.group()
```
Goldziher, line 118: Why not split the commands themselves into a commands module? This file is pretty large

Goldziher (approved):

starlite/cli/main.py
provinzkraut: That's why this is though, isn't it? There's nothing but commands in this module. You could put everything aside from the root `cli` group into another module, but that's not really much of a difference IMO, since that's 10 out of ~200 lines.

Also, 200 lines isn't really that much, is it?

Goldziher: But if you use a module you can split the commands into different files. E.g. "session", "route", "openapi" etc. 

Btw, do we have a command to statically generate a schema file?

provinzkraut: > But if you use a module you can split the commands into different files. E.g. "session", "route", "openapi" etc.

I mean, we could. I just thought it's better to have one 200 lines file than to have 5 with only a couple lines on each, some with only like 10.

If you have a particularly unwieldy command, you can still put it into its own submodule? Just `from .main import cli` and add your stuff there. 

> Btw, do we have a command to statically generate a schema file?

Nope. But sounds like a good subcommand for `openapi` command group!

provinzkraut: Okay, I have an idea for a more broken down structure. I'll add it in a bit.

provinzkraut: @Goldziher I updated things. There's now a `commands` submodule, and groups/commands are simply imported in `cli.main` and added to the root `starlite_group` there. This should suffice for all our needs of more complex CLI stuff at the moment.

starlite/__main__.py
```diff
@@ -0,0 +1,3 @@
+from starlite.cli.main import starlite_group  # pragma: no cover
```
Goldziher, line 1: can you move this file into the cli module? does it need to be in the starlite root level?

Goldziher (approved):
much nicer

starlite/__main__.py
provinzkraut: Doesn't need to be, but this makes `python -m starlite` possible, which could be advantageous in some situations. Putting this into `__main__.py` is the canonical way to invoke a package with a CLI so I thought it's good if we adhere to this standard. I don't see any downsides to this. Wdyt?

PR #1048, opened 2023-01-07
Ensure http requests with invalid payloads return 400.

### LINKED ISSUE #1047: Bug: Malformed payload returns 500, should be 400
Given this application:

```python
from typing import Any

from starlite import Starlite, post


@post("/")
def hello_world(data: dict[str, Any]) -> dict[str, Any]:
    """Route Handler that outputs hello world."""
    return data


app = Starlite(route_handlers=[hello_world])
```

And:

```
$ curl --header "Content-Type: application/json" --request POST --data '{"a": "b",}' http://localhost:8000
```

Response is:

```
{"status_code":500,"detail":"DecodeError('JSON is malformed: trailing comma in object (byte 10)')"}
```

peterschutt:
At the point of parsing the `data` kwarg, this PR looks for a deserialization error and raises a client error http response.

Adds a new exception, `SerializationException` which wraps the `msgspec` operations. I think this better encapsulates the 3rd-party code.

Adds a new HTTP exception `ClientException` and makes other `4xx` errors a subclass.

Makes `ServiceUnavailableException` a subclass of `InternalServerException`.

Closes #1047

# PR Checklist

### DIFF
--- a/tests/utils/test_serialization.py
+++ b/tests/utils/test_serialization.py
@@ -29,7 +29,14 @@ from pydantic import (
 )
 from pydantic.color import Color
 
-from starlite.utils.serialization import default_serializer, encode_json
+from starlite.exceptions import SerializationException
+from starlite.utils.serialization import (
+    decode_json,
+    decode_msgpack,
+    default_serializer,
+    encode_json,
+    encode_msgpack,
+)
 from tests import PersonFactory
 
 person = PersonFactory.build()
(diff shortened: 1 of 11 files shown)


tests/utils/test_serialization.py
```diff
@@ -143,6 +150,13 @@ def test_pydantic_json_compatibility() -> None:
     assert json.loads(model.json()) == json.loads(encode_json(model))
 
 
-def test_unsupported_type_raises() -> None:
-    with pytest.raises(TypeError):
-        encode_json(lambda: None)
```
peterschutt, line 148: Is altering this an issue? FWICT encoders mostly raise `TypeError` but they have a test in msgspec that raises `EncodeError` if enum member type is invalid: https://github.com/jcrist/msgspec/blob/c8e316fa27b264260e28c57963c453cdf5afad85/tests/test_common.py#L315.

After this PR, the encoders will always raise `SerializationException` if either `TypeError` or `EncodeError` is raised, and there are no other test failures as a result, but is the `TypeError` relied upon somewhere given this test exists?

Goldziher (approved):
LGTM

tests/utils/test_serialization.py
provinzkraut: This shouldn’t be an issue as such, but it might be a breaking change if users rely on a catching a `TypeError` from this.

peterschutt: OK, I'll merge and if there is any blowback I'll revisit.

PR #1055, opened 2023-01-09
Migrate the "Usage Guidelines" section from MkDocs to Sphinx

Jarmos-san:
This is a draft PR which introduces the starting phase of the documentations overhaul. It includes migrating the "[Usage Guidelines](https://starlite-api.github.io/starlite/1.48/usage/0-the-starlite-app)" of the existing documentations to its Sphinx version.

# PR Checklist


docs/usage/app-state.rst
```diff
@@ -0,0 +1,157 @@
...
+into the ASGI ``scope`` mapping for each connection (i.e request or websocket connection)
+as ``scope[app].state``. This makes the application accessible whenever the scope mapping
+is available, e.g in middleware, on |Request|_ & |Websocket|_ instances (accessible as
+``request.app`` & ``socket.app`` respectively) as well as many other places.
+
+Therefore, `state` offers an easy way to share contextual data between disparate parts of
+the application as described below:
+
+.. code-block:: python
+
```
provinzkraut, line 25: Please use `literalinclude`. We don't want inline code, since all examples should be tested, which is easier to do if they're just `.py` files.

Jarmos-san: Yeah makes sense. I'll move all the example code snippets to a separate folder named `examples` then.

provinzkraut: It already exists. https://github.com/starlite-api/starlite/tree/main/examples. What's not in there can be extracted. Take a look at [this section](https://github.com/starlite-api/starlite/blob/main/CONTRIBUTING.md#cleaning-up-examples) in `CONTRIBUTING.md`.

docs/usage/app-hooks.rst
```diff
@@ -1,3 +1,82 @@
 =================
```
provinzkraut, line 1: So, I recommend we don't use the double under/over thing to designate different levels of sections, but instead the [standard]https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections) recommended by Sphinx. This is also how the automated tooling I've been using (pandoc/m2r) handle this. 

I'll see if we can add a linter of some sort to enforce this.

Jarmos-san: Is there an example `.rst` file I can refer to for a better understanding of what you've in mind?

Jarmos-san: So, is this what you've have in mind?

```restructuredtext
Chapter Headline
****************

Some content....

Chapter Subsection
------------------

Some content about the subsection....
```

provinzkraut: Yup. Since the characters used can be any punctuation, it's really just a convention, and we could make up our own, but I'd say it's best we simply follow an established one.

PR #1078, opened 2023-01-21
Response doesn't match OpenAPI schema when using aliases

WardPearce:
**Tests currently failing**

### Issue
I've created a few extra checks I believe should pass in regards to schema aliases. I'm not 100% certain how you'd guys would like to go about solving it. 

Somehow  `by_alias` on [this line](https://github.com/WardPearce/starlite/blob/1270892b81b375fae3f5ba087619ccc2985d6ecd/starlite/utils/serialization.py#L50) needs to be passed depending on `Starlite.openapi_config.by_alias`.

Related to Issue https://github.com/starlite-api/starlite/issues/952

### Context
I've been porting a older project what uses `"_id"` in its responses, so being able to continue using underscores is important to avoid breaking other clients interactions.

### Inconsistencies
- `by_alias` is true, request will use alias but response won't actually use the alias outside the schema.
- `by_alias` is false, request will use alias(?) but schema won't.
  - The following [schema](https://paaster.io/3rQtyWm2ZPOgt8vyTaRJW#puUdi83hBBhTn65AyTPOk58rGHCGzN-AshcNWcBnkYc), is generated with  `by_alias` is false. But in order for the request to succeed. `"_id"` must be used instead of `"id"`, but a response with the same object will use `"id"`
  - This results in type errors etc.

### DIFF
--- a/starlite/utils/serialization.py
+++ b/starlite/utils/serialization.py
@@ -47,7 +47,7 @@ DEFAULT_TYPE_ENCODERS: "TypeEncodersMap" = {
     Path: str,
     PurePath: str,
     # pydantic specific types
-    BaseModel: lambda m: m.dict(),
+    BaseModel: lambda m: m.dict(by_alias=False),
     ByteSize: lambda b: b.real,
     NameEmail: str,
     Color: str,
--- a/tests/openapi/test_config.py
+++ b/tests/openapi/test_config.py
@@ -1,7 +1,10 @@
+from re import A
+
 from pydantic import BaseModel, Field
 from pydantic_openapi_schema.v3_1_0 import Components, Example, Header
 
 from starlite import OpenAPIConfig, Starlite, get
+from starlite.testing.client.sync_client import TestClient
 
 
 def test_merged_components_correct() -> None:
@@ -38,10 +41,11 @@ def test_merged_components_correct() -> None:
 def test_by_alias() -> None:
     class ModelWithAlias(BaseModel):
         first: str = Field(alias="second")
+        underscore: str = Field(alias="_underscore")
 
     @get("/")
     def handler() -> ModelWithAlias:
-        return ModelWithAlias(second="abc")
+        return ModelWithAlias(second="abc", _underscore="test")
 
     app = Starlite(
         route_handlers=[handler], openapi_config=OpenAPIConfig(title="my title", version="1.0.0", by_alias=True)
@@ -49,11 +53,17 @@ def test_by_alias() -> None:
 
     assert app.openapi_schema
     assert app.openapi_schema.dict(exclude_none=True)["components"]["schemas"]["ModelWithAlias"] == {
-        "properties": {"second": {"type": "string", "title": "Second"}},
+        "properties": {
+            "second": {"type": "string", "title": "Second"},
+            "_underscore": {"type": "string", "title": " Underscore"},
+        },
         "type": "object",
-        "required": ["second"],
+        "required": ["second", "_underscore"],
         "title": "ModelWithAlias",
     }
+
+    with TestClient(app=app) as client:
+        assert client.get("/").json() == {"second": "abc", "_underscore": "test"}
 
     app = Starlite(
         route_handlers=[handler], openapi_config=OpenAPIConfig(title="my title", version="1.0.0", by_alias=False)
@@ -61,8 +71,14 @@ def test_by_alias() -> None:
 
     assert app.openapi_schema
     assert app.openapi_schema.dict(exclude_none=True)["components"]["schemas"]["ModelWithAlias"] == {
-        "properties": {"first": {"type": "string", "title": "Second"}},
+        "properties": {
+            "first": {"type": "string", "title": "Second"},
+            "underscore": {"type": "string", "title": " Underscore"},
+        },
         "type": "object",
-        "required": ["first"],
+        "required": ["first", "underscore"],
         "title": "ModelWithAlias",
     }
+
+    with TestClient(app=app) as client:
+        assert client.get("/").json() == {"first": "abc", "underscore": "test"}
--- a/tests/openapi/test_config.py
+++ b/tests/openapi/test_config.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel, Field
 from pydantic_openapi_schema.v3_1_0 import Components, Example, Header
 
 from starlite import OpenAPIConfig, Starlite, get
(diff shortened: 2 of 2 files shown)


Goldziher:
Hi, can you give some code from your responses for context?

WardPearce:
> Hi, can you give some code from your responses for context?

```py
class UserToSignModel(BaseModel):
    to_sign: str = Field(..., description="to be signed with ed25519 private key")
    id: ObjectIdStr = Field(..., alias="_id")



@get(
    path="/to-sign/{username:str}",
    description="Used to generate unique password to be signed by client's private key.",
    tags=["user"],
    raises=[UserNotFound],
    middleware=[RateLimitConfig(rate_limit=("minute", 3)).middleware],
)
async def to_sign(username: str) -> UserToSignModel:
    if await Sessions.mongo.user.count_documents({"username": username}) == 0:
        raise UserNotFound()

    to_sign = secrets.token_urlsafe(128)
    result = await Sessions.mongo.proof.insert_one(
        {
            "to_sign": to_sign,
            "username": username,
            "expires": datetime.now(tz=timezone.utc) + timedelta(seconds=60),
        }
    )
    return UserToSignModel(to_sign=to_sign, _id=result.inserted_id)
```

provinzkraut:
@WardPearce We have discussed this before, and the consensus at that time was that the OpenAPI configuration shouldn't change how things get serialised. Your implementation changes the `by_alias` setting unconditionally, which could break a lot of existing code. 

The recommended solution currently would be to override how `BaseModel` gets serialised:

```python
from starlite import Starlite

app = Starlite(..., type_encoders={lambda m: m.dict(by_alias=False)})
```

WardPearce:
> @WardPearce We have discussed this before, and the consensus at that time was that the OpenAPI configuration shouldn't change how things get serialised. Your implementation changes the `by_alias` setting unconditionally, which could break a lot of existing code.
> 
> The recommended solution currently would be to override how `BaseModel` gets serialised:
> 
> ```python
> from starlite import Starlite
> from pydantic import BaseModel
> 
> app = Starlite(..., type_encoders={lambda m: m.dict(by_alias=False)})
> ```

Yeah I haven't really implemented any changes to the serialization, just pointing out where it fails already. A custom encoder works sure. It just seems a bit clunky having to configure my schema to use alias then right a custom encoder to do the same.

Does seem a bit awkward to implement what I'm describing tho, and really its a very small issue.

provinzkraut:
> just pointing out where it fails already.

You cannot make the change you want without it failing, since changing how we treat `by_alias` when serialising Pydantic models is inherently a breaking change.

> Does seem a bit awkward to implement what I'm describing tho, and really its a very small issue.

I agree, except for the very small issue part. As I said before, having OpenAPI configuration affect how things get serialised would be unexpected and not the right way to go about this. It would also break with the paradigm that OpenAPI configuration is, well, just OpenAPI configuration. 

The "correct" way to handle this would be to add a layered parameter (e.g. `pydantic_by_alias`), which would then affect both the OpenAPI schema and how Pydantic models are serialised. Once we add this though, there's no good argument to not add more Pydantic specific configuration, like `exclude`, `exclude_unset` and such. In my opinion however, these are all best addressed via `type_encoders`, instead of having some configuration option somewhere; If you're doing that, you'd always end up leaving *some* option on the floor that someone wants, and then you have the case where you can configure *some* options in one place, and *some* in a completely different place and the boundary between them is completely arbitrary and therefore unintuitive. 

> It just seems a bit clunky having to configure my schema to use alias then right a custom encoder to do the same.

If it's clunky or not depends on the perspective I think. If you see it purely from the point of "I want this thing in both the OpenAPI schema and Pydantic models" then yes, it's clunky. But if you see them as two completely separate things (which they are), it makes more sense IMO. The OpenAPI configuration is where you configure your OpenAPI schema. The `type_encoders` is where you configure how data is normalised. `type_encoders` are by design very simplistic, as to allow users to easily extend functionality, without much configuration.

WardPearce:
> > just pointing out where it fails already.
> 
> You cannot make the change you want without it failing, since changing how we treat `by_alias` when serialising Pydantic models is inherently a breaking change.
> 
> > Does seem a bit awkward to implement what I'm describing tho, and really its a very small issue.
> 
> I agree, except for the very small issue part. As I said before, having OpenAPI configuration affect how things get serialised would be unexpected and not the right way to go about this. It would also break with the paradigm that OpenAPI configuration is, well, just OpenAPI configuration.
> 
> The "correct" way to handle this would be to add a layered parameter (e.g. `pydantic_by_alias`), which would then affect both the OpenAPI schema and how Pydantic models are serialised. Once we add this though, there's no good argument to not add more Pydantic specific configuration, like `exclude`, `exclude_unset` and such. In my opinion however, these are all best addressed via `type_encoders`, instead of having some configuration option somewhere; If you're doing that, you'd always end up leaving _some_ option on the floor that someone wants, and then you have the case where you can configure _some_ options in one place, and _some_ in a completely different place and the boundary between them is completely arbitrary and therefore unintuitive.
> 
> > It just seems a bit clunky having to configure my schema to use alias then right a custom encoder to do the same.
> 
> If it's clunky or not depends on the perspective I think. If you see it purely from the point of "I want this thing in both the OpenAPI schema and Pydantic models" then yes, it's clunky. But if you see them as two completely separate things (which they are), it makes more sense IMO. The OpenAPI configuration is where you configure your OpenAPI schema. The `type_encoders` is where you configure how data is normalised. `type_encoders` are by design very simplistic, as to allow users to easily extend functionality, without much configuration.
> 
> I hope that clears things up a bit about the design choices made here!

Yea that makes more sense now from a design prospective & probably better off to limit the amount of magic.
Using your recommendations it runs correctly.  
```py
    class ModelWithAlias(BaseModel):
        first: str = Field(alias="second")
        underscore: str = Field(alias="_underscore")

    @get("/")
    def handler() -> ModelWithAlias:
        return ModelWithAlias(second="abc", _underscore="test")

    @post("/")
    def handler_post(data: ModelWithAlias) -> None:
        return

    app = Starlite(
        route_handlers=[handler, handler_post],
        openapi_config=OpenAPIConfig(title="my title", version="1.0.0", by_alias=True),
        type_encoders={BaseModel: lambda m: m.dict(by_alias=True)},
    )

        assert client.get("/").json() == {"second": "abc", "_underscore": "test"}

        assert client.post("/", json={"second": "abc", "_underscore": "test"}).status_code == 201
```

Thank you very much for your help!

PR #1079, opened 2023-01-21
Sphinx migration versioning

provinzkraut:
Add basic versioning

# PR Checklist


docs/_static/js/version_select.js
```diff
@@ -0,0 +1,67 @@
+const baseURL = "https://starlite-api.github.io/starlite/"
```
provinzkraut, line 1: My current best idea is to put this into a template, so we can configure it via `docs/conf.py`, currently can't think of a more elegant way.

PR #1101, opened 2023-01-25
Class As Dependency

infohash:
Currently, the `Provide` class takes a callable and accordingly calls it based on if it is a sync callable or an async callable. But when the class with an async `__call__` method is given to Provide, it awaits its `__init__` method. Class is callable but not awaitable so I have added a condition to always call the class as a sync object.

The offset pagination with sqlalchemy depends on this fix as I am passing the paginator class directly to Provide to create its dependency.

### DIFF
--- a/docs/usage/responses.rst
+++ b/docs/usage/responses.rst
@@ -780,6 +780,20 @@ If you require async logic, you can implement
 the :class:`AbstractAsyncOffsetPaginator <starlite.datastructures.pagination.AbstractAsyncOffsetPaginator>` instead of
 the :class:`AbstractSyncOffsetPaginator <starlite.datastructures.pagination.AbstractSyncOffsetPaginator>`.
 
+Offset Pagination With SQLAlchemy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When retrieving paginated data from the database by using sqlalchemy, the paginator instance requires sqlalchemy session
+instance to make queries. This can be achieved with :ref:`dependency injection <usage/dependency-injection:dependency
+kwargs>`
+
+.. literalinclude:: /examples/datastructures/pagination/using_offset_pagination_with_sqlalchemy.py
+    :caption: Offset Pagination With SQLAlchemy
+    :language: python
+
+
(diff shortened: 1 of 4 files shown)


docs/usage/responses.rst
```diff
@@ -780,6 +780,20 @@ If you require async logic, you can implement
 the :class:`AbstractAsyncOffsetPaginator <starlite.datastructures.pagination.AbstractAsyncOffsetPaginator>` instead of
 the :class:`AbstractSyncOffsetPaginator <starlite.datastructures.pagination.AbstractSyncOffsetPaginator>`.
 
+Offset Pagination With SQLAlchemy
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When retrieving paginated data from the database by using sqlalchemy, the paginator instance requires sqlalchemy session
```
Goldziher, line 786:
```suggestion
When retrieving paginated data from the database using SQLAlchemy, the Paginator instance requires an SQLAlchemy session
```

starlite/datastructures/provide.py
```diff
@@ -1,4 +1,4 @@
-from inspect import isasyncgen
+import inspect
```
Goldziher, line 1: please import named members rather than the entire module, e.g. `from inspect import x,y,z`

Goldziher (approved):

starlite/datastructures/provide.py
infohash: Shouldn't we follow [Google style guide for import statements](https://google.github.io/styleguide/pyguide.html#22-imports):

> Use import statements for packages and modules only, not for individual classes or functions.

For better readability, I tend to enforce this rule on import statements of standard libraries to keep it explicit about which module in the standard library implements the function.

Guido Van Rossum:
> There are way more classes and functions than there are modules, so recalling where a particular thing comes from is much easier if it is prefixed with a module name. Often multiple modules happen to define things with the same name -- so a reader of the code doesn't have to go back to the top of the file to see from which module a given name is imported.

provinzkraut: On a personal level I agree with this. But this is how we're doing imports in the rest of the library, and we should keep it consistent.

Goldziher: It's a matter if style, and this is the style of the codebase.

PR #1103, opened 2023-01-26
Improve starlite run

kareemmahlees:
# PR Checklist


kareemmahlees:
@provinzkraut Hey, Could you give me some feedback about the latest 2 to 3 commits?

provinzkraut:
1. Please rebase from `main`. 
2. This is not the way to go. The CLI should automatically detect the installed servers. If you need to specify the server manually. If I have to do `starlite run --server=uvicorn` I'd rather just use `uvicorn app:app`

kareemmahlees:
@provinzkraut thank you for the guidance, but what would the CLI do if there are multiple servers installed?

provinzkraut:
> @provinzkraut thank you for the guidance, but what would the CLI do if there are multiple servers installed?

It should work the same as app autodiscovery. This has a list of locations / functions it looks for, and returns the first match. So just use the first supported server that's installed. I.e. check if uvicorn is available, if it is, use that, if not check if hypercorn is available etc.

provinzkraut:
Closing this as we won't add this functionality. Details can be found here: https://github.com/starlite-api/starlite/issues/1017#issuecomment-1408392418

PR #1105, opened 2023-01-26
Add event emitter and listeners support

Goldziher:
# PR Checklist


starlite/events/listener.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class EventListener:
+    """Decorator for event listeners"""
+
+    __slots__ = ("event_ids", "fn", "listener_id")
+
+    fn: "AsyncCallable[Any, Any]"
+
+    def __init__(self, event_id: Union[str, List[str]]):
```
peterschutt, line 17:
```suggestion
    def __init__(self, event_id: Union[str, Iterable[str]]):
```
`List` is unnecessarily restrictive, all we do is iterate over the values to construct the mapping in the emitter, so it could just as well be an iterator, or sequence of strings if order is important.

provinzkraut, line 17: Why not just 

```suggestion
    def __init__(self, *event_ids: str):
```

?

starlite/events/listener.py
```diff
@@ -0,0 +1,38 @@
...
+        self.event_ids: List[str] = [event_id] if isinstance(event_id, str) else event_id
+
+    def __call__(self, fn: "AnyCallable") -> "EventListener":
+        """Decorate a callable by wrapping it inside an instance of EventListener.
+
+        :param fn: Callable to decorate.
+        :return: An instance of EventListener
+        """
+        if not callable(fn):
+            raise ImproperlyConfiguredException("EventListener instance should be called as a decorator")
```
provinzkraut, line 31: I think this error message can be confusing. If you were to do

```python
@listener("foo")
class ListenForFoo:
    pass
```

it tells you that you should call `EventListener` as a decorator, which you did. And it doesn't tell you what the actual issue is.

provinzkraut (approved):

peterschutt (changes requested):

starlite/events/listener.py
peterschutt: @provinzkraut's arg packing suggestion is even better.

starlite/events/listener.py
Goldziher:
```suggestion
            raise ImproperlyConfiguredException("EventListener instance should be called as a decorator on a callable")
```

starlite/events/listener.py
```diff
@@ -0,0 +1,38 @@
...
+    __slots__ = ("event_ids", "fn", "listener_id")
+
+    fn: "AsyncCallable[Any, Any]"
+
+    def __init__(self, event_id: Union[str, List[str]]):
+        """Create a decorator for event handlers.
+
+        :param event_id: The id of the event to listen to or a list of event ids to listen to.
+        """
+        self.event_ids: List[str] = [event_id] if isinstance(event_id, str) else event_id
```
Goldziher, line 22:
```suggestion
        self.event_ids: List[str] = list(event_ids)
```

PR #1108, opened 2023-01-27
1106 optional sequence query param

peterschutt:
# PR Checklist


peterschutt:
seems I need to change the impl for lower python versions, out of time for now, will address asap.

peterschutt:
Previously, we were using the pydantic shape to determine if the type was one of the sequence types, and it would ignore the optional union, e.g,.:

```python
class MyModel(BaseModel):
    a: list[int] | None
    
MyModel.__fields__
# {'a': ModelField(name='a', type=Optional[List[int]], required=False, default=None)}
MyModel.__fields__["a"].shape
# 2
MyModel.__fields__["a"].annotation
# list[int] | None
```

Where `2` is `SHAPE_LIST`.

Now we are detecting the sequence without pydantic, but to be behaviorally consistent, we should exclude the `Optional` from that calculation as well.

starlite/signature/models.py
```diff
@@ -76,12 +76,18 @@ def is_mapping(self) -> bool:
     @property
     def is_non_string_iterable(self) -> bool:
         """Check if the field type is an Iterable."""
-        return is_non_string_iterable(self.field_type)
+        field_type = self.field_type
+        if self.is_optional:
+            field_type = make_non_optional_union(field_type)
+        return is_non_string_iterable(field_type)
```
provinzkraut, line 82: Would it not be better to have this in `utils.predicates.is_non_string_iterable`?

starlite/signature/models.py
```diff
@@ -76,12 +76,18 @@ def is_mapping(self) -> bool:
...
 
     @property
     def is_non_string_sequence(self) -> bool:
         """Check if the field type is a non-string Sequence."""
-        return is_non_string_sequence(self.field_type)
+        field_type = self.field_type
+        if self.is_optional:
+            field_type = make_non_optional_union(field_type)
+        return is_non_string_sequence(field_type)
```
provinzkraut, line 90: See comment above: Move to `utils.predicates.is_non_string_sequence`?

starlite/utils/types.py
```diff
@@ -35,3 +50,16 @@ def annotation_is_iterable_of_type(
     ):
         return args[0] is type_value or isinstance(args[0], type_value) or is_class_and_subclass(args[0], type_value)
     return False
+
+
+def make_non_optional_union(annotation: Optional[UnionT]) -> UnionT:
+    """Make a `Union` type that excludes `NoneType`.
```
provinzkraut, line 56:
```suggestion
    """Make a :data:`Union <typing.Union>` type that excludes ``NoneType``.
```

starlite/utils/types.py
```diff
@@ -35,3 +50,16 @@ def annotation_is_iterable_of_type(
...
+
+    Args:
+        annotation: A type annotation.
+
+    Returns:
+        The union with all original members, except `NoneType`.
```
provinzkraut, line 62:
```suggestion
        The union with all original members, except ``NoneType``.
```

provinzkraut (changes requested):
Looks good, just a few comments on some details

starlite/signature/models.py
peterschutt: You mean move the optional stripping logic into `is_non_string_iterable()`?

starlite/utils/types.py
peterschutt: thanks for these comments, I'll have to brush up on my ReST!

starlite/signature/models.py
provinzkraut: Yes. Seems to make the most sense to me, since they'd also benefit from this additional check.

peterschutt: I suppose it boils down to whether (for example) `list[Any] | None` should be considered as a non-string sequence/iterable in the general case, or not.

Imagine a new contributor comes across `utils.predicates.is_non_string_iterable()` while doing some unrelated work, and they have no context of this issue, or these changes. Could it not be surprising that it also de-optionalizes the provided type, before testing whether it is an iterable?

The same argument could also be made about `SignatureField.is_non_string_iterable`, but at least the scope of use is somewhat narrowed to signature modelling here.

provinzkraut: > I suppose it boils down to whether (for example) list[Any] | None should be considered as a non-string sequence/iterable in the general case, or not.

It shouldn't, which means it's fine if this stays where it is. 

> The same argument could also be made about SignatureField.is_non_string_iterable, but at least the scope of use is somewhat narrowed to signature modelling here.

I was going to bring this up, which makes me think, maybe this could be solved by just giving it a different name?
Since effectively it does *not* just tell you if it's a non-string iterable. It's more like `is_non_string_iterable_or_non_string_iterable_none_union`, which admittedly is a mouthful, but I think *something* to indicate that this will unwrap `None` unions may be good.

provinzkraut (approved):

starlite/signature/models.py
peterschutt: well I made note of the behavior in the docstring, and referenced the issue as to why we do it.. It looks like we don't document these so if a good name for the property comes to me in my sleep, it should be safe enough to change.

Issue #983, opened 2022-12-24
Enhancement: Update SQLAlchemy Plugin to use SQLA 2.0


Goldziher:
We should update the SQLAlchemy for SQLA 2.0 compatibility. The plugin should be compatible with both v1.4.* and v2.0.0.

peterschutt:
Here's a `tox.ini` that will run the test suite against both 1.4 and 2.0:

```ini
[tox]
env_list = sa{14,20}
isolated_build = true

[base]
deps =
    aiomcache
    aiosqlite
    beanie
    beautifulsoup4
    brotli
    click
    cryptography
    fakeredis
    freezegun
    fsspec
    greenlet
    hypothesis
    jinja2
    jsbeautifier
    mako
    mongomock-motor
    opentelemetry-instrumentation-asgi
    opentelemetry-sdk
    piccolo
    picologging
    pre-commit
    pytest
    pytest-asyncio
    pytest-cov
    pytest-mock
    python-dotenv
    python-jose
    redis
    rich
    starlette
    structlog
    tortoise-orm
    trio
    uvicorn

[testenv]
setenv =
    PATH = {toxinidir}/tests{:}{env:PATH}
extras = full
commands = pytest

[testenv:sa14]
deps =
    sqlalchemy~=1.4
    {[base]deps}

[testenv:sa20]
deps =
    sqlalchemy~=2.0
    {[base]deps}
```

PR #1109, opened 2023-01-27
Sqla 2

peterschutt:
# PR Checklist

Ref: #983


peterschutt:
It didn't take much to get all the tests passing with the test models defined still in the 1.4 style.

The type checking is another story - if we are going to run sqlalchemy 2.0 in pre-commit environments, then we don't have the stubs library or the mypy plugin going, so there are _a lot_ of issues.

peterschutt:
Here's a `tox.ini` that runs the tests against both 2.0 and 1.4:

```ini
[tox]
env_list = sa{14,20}
isolated_build = true

[base]
deps =
    aiomcache
    aiosqlite
    beanie
    beautifulsoup4
    brotli
    click
    cryptography
    fakeredis
    freezegun
    fsspec
    greenlet
    hypothesis
    jinja2
    jsbeautifier
    mako
    mongomock-motor
    opentelemetry-instrumentation-asgi
    opentelemetry-sdk
    piccolo
    picologging
    pre-commit
    pytest
    pytest-asyncio
    pytest-cov
    pytest-mock
    python-dotenv
    python-jose
    redis
    rich
    starlette
    structlog
    tortoise-orm
    trio
    uvicorn

[testenv]
setenv =
    PATH = {toxinidir}/tests{:}{env:PATH}
extras = full
commands = pytest

[testenv:sa14]
deps =
    sqlalchemy~=1.4
    {[base]deps}

[testenv:sa20]
deps =
    sqlalchemy~=2.0
    {[base]deps}
```

Goldziher:
> It didn't take much to get all the tests passing with the test models defined still in the 1.4 style.
> 
> The type checking is another story - if we are going to run sqlalchemy 2.0 in pre-commit environments, then we don't have the stubs library or the mypy plugin going, so there are _a lot_ of issues.

You can add the stubs to the addition_dependencies arrays for mypy and pyright

peterschutt:
> You can add the stubs to the addition_dependencies arrays for mypy and pyright

It just depends which version of sqlalchemy we are going to run the linting against. E.g., if we are to have 2.0 installed in the pre-commit env, then we don't need/want the stubs and cannot run the plugin, but then we'll have typing errors for all of the 1.4 code.

If we are going to have 1.4 in the pre-commit environment, then we do want the stubs and plugin, but all of the 2.0 code won't type check.

I just pushed a commit that adds a 2nd pre-commit mypy environment aliased `mypy-sqla-1x` which runs the checks with the type stubs and plugin installed. The existing `mypy` hook will run the checks against 2.0, and I'd imagine that for either of them, I'm going to have to exclude checks for certain modules/sub-packages that contain 1.x or 2.x specific code.

I can probably reduce the scope of the modules/sub-packages checked by the 1x version to speed it up a bit.

As it stands, (locally at least) the 1x checks are passing and the 2x checks has many errors. So, next step will be to make them play nice together.

Goldziher:
Can't we simply deprecate the v1sqla plugin and add a V2 only version?

peterschutt:
> Can't we simply deprecate the v1sqla plugin and add a V2 only version?

Based on [this](https://github.com/starlite-api/starlite/issues/983#issue-1510170212j), I've been thinking about how to support both and I'm not totally convinced it is infeasible, yet.

I spent a bit of time going down a path that I didn't end up liking, which was to try to unwind a lot of the usage of the SQLAlchemy plugin from within our test suite to test other things, e.g., DTO factory is tested using the sqlalchemy plugin, but another way to go about that would be to flip that dependency and test that the sqlalchemy plugin works with the DTO factory, and move those tests into a walled off area of the test suite that is only collected based on the version of sqlalchemy that is available in the testing environment.

However I felt that was too disruptive overall. My next plan of attack was going to be to modify any existing internal model definitions (e.g., sqlalchemy-backed session and test models) to ones that will be supported by both 1.4 and 2.0, probably using [`__allow_unmapped__`](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-six-add-allow-unmapped-to-explicitly-typed-orm-models), and then just get on with abstracting the 1.4 v 2.0 differences within the plugin itself.

Even if we go down the path of having two totally separate plugins, there will still probably be a bit of common functionality that would be shared between the two. So we could either extract that stuff out to some sort of mini-lib, or use a compatibility layer like I've stubbed out here.

I'll spend a bit more time on it and see where that leaves my opinion on it.

provinzkraut:
> Based on https://github.com/starlite-api/starlite/issues/983#issue-1510170212j, I've been thinking about how to support both and I'm not totally convinced it is infeasible, yet.

If it's possible that would be great of course, but I wouldn't mind having two separate plugins as well. If it can be made to work but only with substantial efforts and workarounds, I'd say it's not worth it and we should just have two separate plugins.

peterschutt:
> > Based on https://github.com/starlite-api/starlite/issues/983#issue-1510170212j, I've been thinking about how to support both and I'm not totally convinced it is infeasible, yet.
> 
> 
> 
> If it's possible that would be great of course, but I wouldn't mind having two separate plugins as well. If it can be made to work but only with substantial efforts and workarounds, I'd say it's not worth it and we should just have two separate plugins.

It's going to be difficult-ish to support both from a testing standpoint, as we can only have one of either 1.4 or 2.0 in any environment. I can still make that work though with selective test collection and having multiple mypy pre-commits like I've setup here, but it will add a bit of complexity.

So, do we make starlite 1.x support sqlalchemy 1.x and starlite 2.x support sqlalchemy 2.x? All the environmental issues go away in that scenario.

Goldziher:
> > > Based on https://github.com/starlite-api/starlite/issues/983#issue-1510170212j, I've been thinking about how to support both and I'm not totally convinced it is infeasible, yet.
> > 
> > 
> > 
> > If it's possible that would be great of course, but I wouldn't mind having two separate plugins as well. If it can be made to work but only with substantial efforts and workarounds, I'd say it's not worth it and we should just have two separate plugins.
> 
> It's going to be difficult-ish to support both from a testing standpoint, as we can only have one of either 1.4 or 2.0 in any environment. I can still make that work though with selective test collection and having multiple mypy pre-commits like I've setup here, but it will add a bit of complexity.
> 
> So, do we make starlite 1.x support sqlalchemy 1.x and starlite 2.x support sqlalchemy 2.x? All the environmental issues go away in that scenario.

I think that will be best. Simply target V2 SQLA/Starlite

provinzkraut:
> I think that will be best. Simply target V2 SQLA/Starlite

I don't think we should do this. It would severely hinder the adoption of Starlite 2.0 when used in conjunction with the SQLAlchemy plugin. A lot of applications still run SQLAlchemy 1.4 and can't / won't upgrade in the near future. This includes on of the larger pieces I'm working on as well. Migration is planned, but won't happen for another ~6 months or so.

I see two ways to resolve this: Either use the convoluted test setup @peterschutt described or create a separate package for the SQLAlchemyPlugin targeting SQLAlchemy 1.x, and include it as an extra

If we go the second route, we could pin `sqlalchemy >= 2`, and add that as an `sqlalchemy` extra. Then we create a package with the SQLA 1.x plugin which has `sqlalchemy < 2` pinned, and add that as an `sqlalchemy1` extra.

peterschutt:
> > I think that will be best. Simply target V2 SQLA/Starlite
> 
> 
> 
> I don't think we should do this. It would severely hinder the adoption of Starlite 2.0 when used in conjunction with the SQLAlchemy plugin. A lot of applications still run SQLAlchemy 1.4 and can't / won't upgrade in the near future. This includes on of the larger pieces I'm working on as well. Migration is planned, but won't happen for another ~6 months or so.
> 
> 
> 
> I see two ways to resolve this: Either use the convoluted test setup @peterschutt described or create a separate package for the SQLAlchemyPlugin targeting SQLAlchemy 1.x, and include it as an extra
> 
> 
> 
> If we go the second route, we could pin `sqlalchemy >= 2`, and add that as an `sqlalchemy` extra. Then we create a package with the SQLA 1.x plugin which has `sqlalchemy < 2` pinned, and add that as an `sqlalchemy1` extra. 

Yes! I had the thought of making the 1.x plug-in an external lib as well.. that should give us the best of both worlds. I'll head in this direction.

PR #1128, opened 2023-02-01
Added tutorial for how to test route handlers that use SQLAlchemy Plugin

infohash:
Let me know if filenames are appropriate and docs is at the right place.

### DIFF
--- a/docs/examples/plugins/sqlalchemy_plugin/testing_with_sqlalchemy/sqla_basic_app.py
+++ b/docs/examples/plugins/sqlalchemy_plugin/testing_with_sqlalchemy/sqla_basic_app.py
@@ -1,0 +1,71 @@
+from typing import Optional, List, TYPE_CHECKING
+
+from sqlalchemy import Column, Integer, String, select
+from sqlalchemy.orm import Mapped, declarative_base
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from starlite import Controller, DTOFactory, HTTPException, Starlite, get
+from starlite.dto import DTO
+from starlite.plugins.sql_alchemy import SQLAlchemyConfig, SQLAlchemyPlugin
+from starlite.status_codes import HTTP_404_NOT_FOUND
+
+if TYPE_CHECKING:
+    from sqlalchemy.engine.result import ScalarResult
+
+Base = declarative_base()
+
+
+class Company(Base):  # pyright: ignore
+    """SQlAlchemy model of a table 'company'."""
+
+    __tablename__ = "company"
+    id: Mapped[int] = Column(Integer, primary_key=True)
+    name: Mapped[str] = Column(String)
+
+
+class CompanyController(Controller):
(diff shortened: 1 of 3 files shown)


provinzkraut (changes requested):
I don't think this is a pattern we should promote. Patching `.scalars` effectively circumnavigates almost all of the ORM, which means you're not actually testing the behaviour of your endpoint anymore, but mostly that your mock does what you expect it to do.

This pattern would *maybe* be fine for circumstances where you're testing *only* how the surrounding parts act, but usually when you're testing endpoints of an application that's not what you're interested in. You want to prove that the endpoint does what its supposed to do, and this cannot accomplish that.

infohash:
ORM classes are not part of user code, aren't they? So I am patching that class of ORM with the return values that would be there if the code was actually running and the endpoint will process that patched data.

provinzkraut:
> ORM classes are not part of the user code, aren't they? 

Yes they are. How you set up your models and their resulting interaction with the ORM is very much something you want to test. The way you test doesn't even guarantee that you're running the correct query.

You can change this:

```python
result: "ScalarResult" = await async_session.scalars(select(Company).where(Company.id == company_id))
```

to 
```python
result: "ScalarResult" = await async_session.scalars(select(Company).where(1 == 2))
```

and your tests still pass.

peterschutt:
I agree with @provinzkraut.

This is certainly one way to abstract the ORM away from unit tests, but I tend to think that if you need to go to these lengths to unit test your business logic, then you could probably stand to refactor the application to separate concerns better. So based on that, I don't think we should promote this as a canonical way to unit test an app that requires sqlalchemy.

As for what @provinzkraut mentioned about not _actually_ testing your application, that is correct, and integration tests are a better path if you want to be more thorough with testing before getting the app into some sort of dev or staging environment. I messed around with a pattern that uses `pytest-docker` to patch the application services with running redis and postgres instances that are spun up specifically for testing [which you can take a look at](https://github.com/starlite-api/starlite-pg-redis-docker/tree/main/tests/integration). I would definitely still call that code an "experiment", but it is one that I have found useful, and I'll likely iterate over it given time.

Having said all this, your write-up is well written, comprehensive and with good examples, and I am sure that there are people that this would help. How about making it a blog post?

infohash:
Fair points. Help me to understand one thing with writing unit tests. If a route handler depends on the response from the database to process the response before it can be returned to the user, how will you mock that response without patching return value of the internals of SQLA? If it was possible, I would mock the binary I/O response but it varies with the driver.

```python
result: "ScalarResult" = await async_session.scalars(select(Company).where(Company.id == company_id))

to

result: "ScalarResult" = await async_session.scalars(select(Company).where(1 == 2))

and the tests still pass.
```
Right, that's because I didn't assert the arguments of scalars. Like this:

```python
assert str(self.async_session_mock.scalars.call_args.args[0]) == str(select(Company).where(Company.id == company_id))
```

Issue #1137, opened 2023-02-03
Bug: LoggingMiddleware stores all responses permanently in memory (memory leak)


chris-telemetry:
I've been testing out the LoggingMiddleware and I believe it's holding the response data in memory. This is causing very high memory usage for some of our endpoints that return very large (~10-20MB) responses. This happens even when I set `response_log_fields=('status_code')`.

I'm using using starlite = {extras = ["full"], version = "1.50.2"} so the logger should be using picologging.

Here's a snippet for you, the json file I'm using to test with is about 20MB.
```python
import json
from starlite import Starlite, post, LoggingConfig
from starlite.middleware import LoggingMiddlewareConfig
import uvicorn
from typing import  Any
@post('/test')
async def test_endpt(data: Any) -> Any:
    with open("/Users/christophermoyer/code/scratch/test.json", "r") as f:
        data = json.load(f)
    return data

app = Starlite(
    route_handlers=[test_endpt],
    logging_config=LoggingConfig(),
    middleware=[LoggingMiddlewareConfig(response_log_fields=('status_code')).middleware]
    
)
if __name__ == '__main__':
    uvicorn.run("main:app", host="0.0.0.0", port=8001, reload=True) 
```

Tracemalloc showing memory is being held on to (LoggingMiddleware enabled):
```
[ Top 10 ]
starlite-fiddle-LywfGICr-py3.11/lib/python3.11/site-packages/starlite/utils/serialization.py:161: size=1730 MiB, count=61, average=28.4 MiB
starlite-fiddle-LywfGICr-py3.11/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py:208: size=64.0 KiB, count=1212, average=54 B
.pyenv/versions/3.11.0/lib/python3.11/json/decoder.py:353: size=61.8 KiB, count=675, average=94 B
.pyenv/versions/3.11.0/lib/python3.11/asyncio/locks.py:168: size=46.0 KiB, count=124, average=380 B
starlite-fiddle-LywfGICr-py3.11/lib/python3.11/site-packages/starlite/connection/request.py:89: size=45.8 KiB, count=61, average=768 B
starlite-fiddle-LywfGICr-py3.11/lib/python3.11/site-packages/starlite/utils/serialization.py:192: size=44.7 KiB, count=614, average=75 B
[...]

PR #1138, opened 2023-02-03
Fix memory leak in logging middleware

Goldziher:
# PR Checklist

This PR address the issue reported in #1137 by cleaning up values from the state object.

### DIFF
--- a/starlite/utils/scope.py
+++ b/starlite/utils/scope.py
@@ -1,7 +1,6 @@
 from typing import TYPE_CHECKING, Any, Optional
 
 from starlite.constants import SCOPE_STATE_NAMESPACE
-from starlite.types import Empty
 
 if TYPE_CHECKING:
     from starlite.types import Scope, Serializer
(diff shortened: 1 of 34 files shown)


starlite/utils/scope.py
```diff
@@ -52,9 +52,7 @@ def get_starlite_scope_state(scope: "Scope", key: str, default: Any = Empty) ->
         Value mapped to ``key`` in internal connection scope namespace.
     """
     namespace = scope["state"].setdefault(SCOPE_STATE_NAMESPACE, {})
-    if default is Empty:
-        return namespace.get(key)
-    return namespace.setdefault(key, default)
+    return namespace.pop(key, default if default is not Empty else None)
```
provinzkraut, line 55: I don't think that's right. Accessing a value should not delete it. If you call `get_starlite_scope_state(scope, "foo")` twice in a row, it should not return a different result both times.

Maybe just delete the value inside the middleware?

Either that or move this function into the `middleware/logigng.py`, since that's the only place its being used.

provinzkraut: I'm also not sure this fixes the issue, since overriding the key with an empty string doesn't seem to fix it either, which also results in the value not being referenced anymore in the state.

Goldziher: i updated the implementation

provinzkraut: I'd still check with @chris-telemetry before merging this

Issue #1149, opened 2023-02-04
Bug: middleware not excluding static path


ravi-b-m:
I have a auth middleware and I want to exclude static routes defined by static file config and they are not excluded

```python
auth_mw = DefineMiddleware(JWTAuthenticationMiddleware, exclude=["health", "docs"])

app = Starlite(route_handlers=[health_check, InterestsController],
               static_files_config=[
                   StaticFilesConfig(directories=["docs"], path="/docs", name="docs")
               ],
               logging_config=log_config,
               middleware=[auth_mw])
```

How do I exclude static files from middleware ?

Goldziher:
can you try:

```python
auth_mw = DefineMiddleware(JWTAuthenticationMiddleware, exclude=["health", "docs.*"])

app = Starlite(route_handlers=[health_check, InterestsController],
     static_files_config=[
         StaticFilesConfig(directories=["docs"], path="/docs", name="docs")
     ],
     logging_config=log_config,
     middleware=[auth_mw])
```

the value is a regex.

ravi-b-m:
unfortunately that didn;t work
```python
auth_mw = DefineMiddleware(JWTAuthenticationMiddleware, exclude=["health", "docs.*"])

app = Starlite(route_handlers=[health_check, InterestsController],
               static_files_config=[
                   StaticFilesConfig(directories=["docs"], path="/docs", name="docs")
               ],
               logging_config=log_config,
               middleware=[auth_mw])

if __name__ == "__main__":
    # port = environ.get("CDSW_READONLY_PORT")
    port = 8000
    uvicorn.run("main:app", host="127.0.0.1", port=port, log_level="debug", reload=True)
```
I still 401 ..

Goldziher:
> unfortunately that didn;t work
> ```python
> auth_mw = DefineMiddleware(JWTAuthenticationMiddleware, exclude=["health", "docs.*"])
> 
> app = Starlite(route_handlers=[health_check, InterestsController],
>                static_files_config=[
>                    StaticFilesConfig(directories=["docs"], path="/docs", name="docs")
>                ],
>                logging_config=log_config,
>                middleware=[auth_mw])
> 
> if __name__ == "__main__":
>     # port = environ.get("CDSW_READONLY_PORT")
>     port = 8000
>     uvicorn.run("main:app", host="127.0.0.1", port=port, log_level="debug", reload=True)
> ```
> I still 401 ..

Ok, I'll look into it

[...]

PR #1155, opened 2023-02-05
add tests

Goldziher:
This PR includes passing tests, which probably indicate that #1149 is not reproducible.

### DIFF
--- a/tests/contrib/jwt/test_auth.py
+++ b/tests/contrib/jwt/test_auth.py
@@ -7,12 +7,15 @@ from hypothesis import given, settings
 from hypothesis.strategies import integers, none, one_of, sampled_from, text, timedeltas
 
 from starlite import ASGIConnection, Request, Response, Starlite, get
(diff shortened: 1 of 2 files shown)


tests/contrib/jwt/test_auth.py
```diff
@@ -277,6 +280,37 @@ def west_handler() -> None:
         assert response.status_code == HTTP_401_UNAUTHORIZED
 
 
+def test_authentication_middleware_excludes_static_pahts(tmpdir: "Path") -> None:
```
JacobCoffee, line 283: `def test_authentication_middleware_excludes_static_paths(tmpdir: "Path") -> None:`

pahts -> paths

JacobCoffee, line 283: `def test_authentication_middleware_excludes_static_paths(tmpdir: "Path") -> None:`

pahts -> paths

JacobCoffee (changes requested):

PR #1160, opened 2023-02-06
Dto refactor

peterschutt:
# PR Checklist


Goldziher (approved):
Looks great. This solves the typing issues?

starlite/dto.py
```diff
@@ -56,265 +117,115 @@ def get_field_type(model_field: ModelField) -> Any:
...
+    class Config(BaseConfig):
+        orm_mode = True
 
-            class MyClass(BaseModel):
-                first: int
-                second: int
+    _model_type: ClassVar[Any]
+    _config: DTOConfig = DTOConfig()
+    _reverse_field_mappings: Dict[str, str]
+    plugin: ClassVar[PluginProtocol]
```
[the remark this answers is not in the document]
peterschutt: Under this design, the plugin is provided via subclassing `dto.Factory()`, e.g.:

```py
class DataclassDTOFactory(Factory[DataclassT]):
    plugin: ClassVar[DataclassPlugin] = DataclassPlugin()
```

I haven't locked the design in yet as I don't like that pydantic forces the end user to re-declare the `ClassVar` annotation on the subclass. Will revisit once the MVP is in place.

peterschutt:
> This solves the typing issues?

Type checkers  will be aware that the dto types are pydantic subclasses, and that they are subclasses of `dto.Factory`. They still won't have any idea about the model attributes (e.g., `dto_instance.a`) but they will know about methods like `dto_instance.to_model_instance()` and `DTOType.parse_obj()` etc etc. So all of the interface required to transform the dto instances into model instances and back to dto's will be fully typed, which is all that should matter for a DTO I think.

docs/examples/tests/data_transfer_objects/test_dto_model_interaction.py
```diff
@@ -11,13 +13,15 @@
 )
 
 
+@pytest.mark.xfail(reason="sqlalchemy plugin model map returns wrong instance for test")
```
peterschutt, line 16: These are failing as the new pattern keeps an instance of the plugin for the life of the factory type.. the plugin caches based on model name, and as we have multiple tests with the same `Company` model, the plugin is returning the pydantic model that is cached from the first test that has that model in it.

Goldziher: How about resetting the model cache?

Goldziher:
> > This solves the typing issues?
> 
> Type checkers  will be aware that the dto types are pydantic subclasses, and that they are subclasses of `dto.Factory`. They still won't have any idea about the model attributes (e.g., `dto_instance.a`) but they will know about methods like `dto_instance.to_model_instance()` and `DTOType.parse_obj()` etc etc. So all of the interface required to transform the dto instances into model instances and back to dto's will be fully typed, which is all that should matter for a DTO I think.

Hmm, how is this different from what was there already?

JacobCoffee (approved):

JacobCoffee:
Hmm it doesn't show the ones I reviewed... is there a way to selectively review? I click the checkboxes of all except a couple that just melted my brain thinking it would approve those only 😬

provinzkraut:
> is there a way to selectively review?

Nope. You can only review full PRs. The "viewed files" is just for your own personal bookkeeping.

starlite/dto/factory.py
```diff
@@ -0,0 +1,138 @@
...
+from starlite.utils import is_awaitable, is_not_awaitable
+
+from .config import Config as DTOConfig
+from .utils import create_field_definitions
+
+T = TypeVar("T")
+FactoryT = TypeVar("FactoryT", bound="Factory")
+
+
+class Factory(BaseModel, ABC, Generic[T]):
```
peterschutt, line 20:
```suggestion
class Factory(BaseModel, Generic[T]):
```

starlite/dto/factory.py
```diff
@@ -0,0 +1,138 @@
...
+    _reverse_field_mappings: ClassVar[Dict[str, str]]
+
+    def __class_getitem__(cls, item: TypeVar | type[T]) -> type[Factory[T]]:
+        if isinstance(item, TypeVar):
+            return cls
+
+        if get_origin(item) is Annotated:
+            item, config = get_args(item)
+            item = cast("type[T]", item)
+            config = cast("DTOConfig", config)
```
peterschutt, line 43: Maybe an actual runtime type check for the config object?

starlite/plugins/pydantic.py
```diff
@@ -0,0 +1,77 @@
...
+        """Produce a pydantic model from ``model_class``.
+
+        Args:
+            model_class: the pydantic model type.
+            **kwargs: not used in this implementation
+
+        Returns:
+            A pydantic model to represent the pydantic model.
+        """
+        return model_class
```
peterschutt, line 40: This should probably return a copy of the model.

starlite/utils/predicates.py
```diff
@@ -236,3 +261,19 @@ def is_typed_dict(value: Any) -> "TypeGuard[TypedDictClass]":
...
+        value: value to be tested.
+        type_: type for test.
+
+    Returns:
+        Bool indicating whether ``value`` is either a type of ``type_`` or an instance of ``type_``.
+    """
+    if is_class_and_subclass(value, t_type=type_):
+        return True
+
+    return isinstance(value, type_)
```
peterschutt, line 279: Does this accept a class as the value param?

tests/response/test_dto_response.py
```diff
@@ -1,148 +1,149 @@
-from sys import version_info
```
peterschutt, line 1: Fix this

.pre-commit-config.yaml
```diff
@@ -17,6 +17,7 @@ repos:
     hooks:
       - id: pyupgrade
         args: ["--py38-plus"]
+        exclude: "tests/dto/__init__|starlite/dto/factory"
```
peterschutt, line 20: Comment why excluded

peterschutt:
> > > This solves the typing issues?
> > 
> > Type checkers  will be aware that the dto types are pydantic subclasses, and that they are subclasses of `dto.Factory`. They still won't have any idea about the model attributes (e.g., `dto_instance.a`) but they will know about methods like `dto_instance.to_model_instance()` and `DTOType.parse_obj()` etc etc. So all of the interface required to transform the dto instances into model instances and back to dto's will be fully typed, which is all that should matter for a DTO I think.
> 
> Hmm, how is this different from what was there already?


Obviously a very different interface to create the DTO models, which is to achieve a nicer typing story.

After that, hopefully not functionally different at all.

.pre-commit-config.yaml
peterschutt: went one better, removed the exclusions

docs/examples/tests/data_transfer_objects/test_dto_model_interaction.py
peterschutt: the issue is that the classes are generated at load/compile time, so by the time we get to them in the tests, its already done. 

For example, I could just change the pattern from:

```py
class DataclassFactory(Factory[DataclassT]):
    """DTO Factory for dataclass models."""

    plugin: ClassVar[DataclassPlugin] = DataclassPlugin()
```

to

```py
class DataclassFactory(Factory[DataclassT]):
    """DTO Factory for dataclass models."""

    plugin_type: ClassVar[DataclassPlugin] = DataclassPlugin
```
And give every dto class generated by the factory its own plugin instance. I think that would solve it.. and it would let users have models with the same name across different modules without hitting the same issue also, so prob makes sense.

CHANGELOG.rst
```diff
@@ -530,7 +530,7 @@ Changelog
 ------
 
 -  refactored brotli middleware typing.
--  update Extended ``PluginProtocol`` with an ``on_app_init`` method.
+-  update Extended ``SerializationPluginProtocol`` with an ``on_app_init`` method.
```
peterschutt, line 533:
```suggestion
-  update Extended ``PluginProtocol`` with an ``on_app_init`` method.
```

docs/examples/data_transfer_objects/dto_auto_conversion.py
```diff
@@ -1,17 +1,14 @@
-from typing import List
-
 from sqlalchemy import Column, Float, Integer, String
```
peterschutt, line 1: `from __future__ import annotations`

docs/examples/tests/data_transfer_objects/test_dto_auto_conversion.py
```diff
@@ -1,10 +1,15 @@
+import pytest
+
 from examples.data_transfer_objects.dto_auto_conversion import app
 from starlite.testing import TestClient
 
 
+# TODO
+@pytest.mark.xfail(reason="need to get return type of signature model to resolve forward ref")
```
peterschutt, line 8: todo

docs/examples/tests/data_transfer_objects/test_dto_auto_conversion.py
```diff
@@ -1,10 +1,15 @@
...
 def test_app() -> None:
     with TestClient(app=app) as client:
         get_res = client.get("/1")
+        print(get_res.text)
```
peterschutt, line 12:
```suggestion
```

docs/examples/tests/data_transfer_objects/test_dto_model_interaction.py
```diff
@@ -11,16 +13,18 @@
...
 def test_dto_from_model_instance() -> None:
     assert isinstance(dto_instance, FromModelCompanyDTO)
 
     for field in FromModelCompanyDTO.__fields__:
         assert getattr(dto_instance, field) == getattr(company_instance, field)
 
 
+@pytest.mark.xfail(reason="sqlalchemy plugin model map returns wrong instance for test")
```
peterschutt, line 24: This test might still be failing as I haven't seen any xpassed but if it is, it isn't for this reason anymore.

docs/examples/tests/plugins/test_sqlalchemy_plugin.py
```diff
@@ -22,6 +22,7 @@ def clear_sqlite_dbs() -> Generator[None, None, None]:
         db_file.unlink()
 
 
+@pytest.mark.xfail(reason="sqlalchemy plugin model map returns wrong instance for test")
```
peterschutt, line 25: as above

docs/examples/tests/plugins/test_sqlalchemy_plugin.py
```diff
@@ -47,7 +48,8 @@ def test_relationships() -> None:
             "id": 1,
             "name": "Peter",
             "company_id": 1,
-            "company": {"id": 1, "name": "Peter Co.", "worth": 0.0},
+            # TODO: add relationships back in
```
peterschutt, line 51: todo

starlite/dto/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+
+    from starlite import dto
+    from starlite.utils.model import convert_dataclass_to_model
+    from starlite.utils.predicates import is_dataclass_class_or_instance
+
+    if TYPE_CHECKING:
+        from starlite.plugins import PluginProtocol
+
+
+    class DataclassPlugin(PluginProtocol[dataclass]):
```
peterschutt, line 20: change example to import the dataclass plugin from starlite.plugins

starlite/dto/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+            return model_class(**pydantic_model_instance.dict())
+
+        def to_dict(self, model_instance: dataclass) -> Dict[str, Any]:
+            return asdict(model_instance)
+
+        def from_dict(self, model_class: Type[dataclass], **kwargs: Any) -> dataclass:
+            return model_class(**kwargs)
+
+
+    DataclassT = TypeVar("DataclassT", bound=dataclass)
```
peterschutt, line 40: wrong

starlite/dto/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @dataclass()
+    class Model2:
+        c: float
+        one: Model1
+
+    ModelDTO = DataclassDTO[Model2]
+
+    # to add fields onto a DTO
+    class ModelDTO(DataclassDTO[Model]):
```
peterschutt, line 61: this pattern no longer works

starlite/dto/factory.py
```diff
@@ -0,0 +1,158 @@
...
+    from typing import Any
+
+ModelT = TypeVar("ModelT")
+FactoryT = TypeVar("FactoryT", bound="Factory")
+
+
+class Factory(Generic[ModelT]):
+    """Create :class:`DTO` type.
+
+    Subclass Factory and define `plugin_type`.
```
peterschutt, line 27: sphinxify

starlite/dto/factory.py
```diff
@@ -0,0 +1,158 @@
...
+from .config import Config as DTOConfig
+
+if TYPE_CHECKING:
+    from typing import Any
+
+ModelT = TypeVar("ModelT")
+FactoryT = TypeVar("FactoryT", bound="Factory")
+
+
+class Factory(Generic[ModelT]):
```
peterschutt, line 24: is `Factory` the right name?

starlite/dto/factory.py
```diff
@@ -0,0 +1,158 @@
...
+        cls_kwargs["_transfer_type"] = plugin_instance.to_data_container_class(
+            item,
+            exclude=config.exclude,
+            field_mappings=config.field_mapping,
+            fields=config.fields,
+            localns={**vars(typing), **vars(typing_extensions), **vars(item_module)},
+        )
+        cls_kwargs["_model_type"] = item
+
+        return type(f"Factory[{item.__name__}, {cls.plugin_type.__name__}]", (cls,), cls_kwargs)
```
peterschutt, line 97: if the class name changes, so does this

starlite/dto/factory.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def to_model(self) -> ModelT:
+        """Convert self into instance of the model type.
```
peterschutt, line 100: should reword

starlite/dto/factory.py
```diff
@@ -0,0 +1,158 @@
...
+    def from_model(cls: type[FactoryT], model_instance: ModelT) -> FactoryT:
+        """Create an instance of ``dto.Factory`` from an instance of ``ModelT``.
+
+        Args:
+            model_instance: instance of the model
+
+        Returns:
+            Instance of ``dto.Factory``
+        """
+        return cls.from_buffer(encode_json(cls._plugin_instance.to_dict(model_instance)), RequestEncodingType.JSON)
```
peterschutt, line 128: not sure this is the right way to go.. it keeps it simple though

starlite/dto/sqlalchemy.py
```diff
@@ -0,0 +1,14 @@
...
+from starlite.plugins.sql_alchemy import SQLAlchemyPlugin
+
+from .factory import Factory
+from .types import SQLAlchemyT
+
+__all__ = ("SQLAlchemyFactory",)
+
+
+class SQLAlchemyFactory(Factory[SQLAlchemyT]):
+    """A DTO Factory type for dataclasses models."""
```
peterschutt, line 12: wrong

starlite/handlers/http_handlers.py
```diff
@@ -192,8 +191,8 @@ def _create_data_handler(
     ]
     cookie_headers = [cookie.to_encoded_header() for cookie in cookies if not cookie.documentation_only]
     raw_headers = [*normalized_headers, *cookie_headers]
-    is_dto_annotation = is_class_and_subclass(return_annotation, DTO)
-    is_dto_iterable_annotation = annotation_is_iterable_of_type(return_annotation, DTO)
+    is_dto_annotation = is_class_and_subclass(return_annotation, dto.Factory)
+    is_dto_iterable_annotation = annotation_is_iterable_of_type(return_annotation, dto.Factory)
```
peterschutt, line 195: these are failing in the examples when the return annotation is a forward ref due to `from __future__ import annotations` - Signature.from_callable() isn't resolving the runtime type, so these are always false under those conditions.

starlite/handlers/http_handlers.py
```diff
@@ -214,16 +213,13 @@ async def handler(data: Any, plugins: List["SerializationPluginProtocol"], **kwa
         if isawaitable(data):
             data = await data
 
-        if is_dto_annotation and not isinstance(data, DTO):
-            data = return_annotation(**data) if isinstance(data, dict) else return_annotation.from_model_instance(data)
+        if is_dto_annotation:
+            data = data.transfer_instance
```
peterschutt, line 217: this removes the functionality where we can type a handler with a dto as return type, and then return something different from the handler.

It is a strange pattern to support IMO - so i pulled it for simplicity.

E.g. we were supporting a pattern that will always result in type errors..

```py
def handler() -> dto.Factory[MyModel]:
   ...
   return MyModel()  # type error
```

peterschutt: @starlite-api/maintainers ping

peterschutt: previously, when DTOs were pydantic models, and we had `# type:ignore`'s everywhere anyway, it may not have been as much as an issue.

starlite/plugins/base.py
```diff
@@ -1,32 +1,19 @@
...
+    from pydantic import BaseModel
     from pydantic_openapi_schema.v3_1_0 import Schema
 
     from starlite.app import Starlite
+    from starlite.enums import RequestEncodingType
+    from starlite.types.protocols import DataclassProtocol
 
 ModelT = TypeVar("ModelT")
-DataContainerT = TypeVar("DataContainerT", bound=Union[BaseModel, DataclassProtocol, TypedDict])  # type: ignore[valid-type]
+DataContainerT = TypeVar("DataContainerT", bound="BaseModel | DataclassProtocol | TypedDict")  # type: ignore[valid-type]
```
peterschutt, line 16: We aren't using dataclasses or typeddicts as container types though, are we?

starlite/plugins/base.py
```diff
@@ -102,7 +137,7 @@ def is_plugin_supported_type(value: Any) -> TypeGuard[ModelT]:
         """
         raise NotImplementedError()
 
-    def to_dict(self, model_instance: ModelT) -> Union[Dict[str, Any], Awaitable[Dict[str, Any]]]:
+    def to_dict(self, model_instance: ModelT) -> "dict[str, Any]":
```
peterschutt, line 140: maybe roll this change back given the headaches with testing using the blocking portal has caused in tortoise plugin.

starlite/plugins/dataclasses.py
```diff
@@ -0,0 +1,92 @@
...
+        :param model_class: The model class that serves as a basis.
+        :param exclude: do not include fields of these names in the model.
+        :param field_mappings: to rename, and re-type fields.
+        :param fields: additional fields to add to the container model.
+        :param localns: additional namespace for forward ref resolution.
+        :param kwargs: Any kwargs.
+        :return: The generated data container class.
+        """
+
+        pydantic_model = pydantic.dataclasses.dataclass(model_class).__pydantic_model__
```
peterschutt, line 53: todo: exclude/field_mapping/fields support

starlite/plugins/piccolo_orm.py
```diff
@@ -16,17 +13,31 @@
...
+    _models_map: "dict[type[Table], type[BaseModel]]" = {}
+    _data_models_map: "dict[type[Table], type[BaseModel]]" = {}
+
+    def to_data_container_class(
+        self,
+        model_class: "type[PiccoloModelT]",
+        exclude: "set[str] | None" = None,
+        field_mappings: "Mapping[str, str | tuple[str, Any]] | None" = None,
+        fields: "dict[str, tuple[Any, Any]] | None" = None,
+        localns: "dict[str, Any] | None" = None,
```
peterschutt, line 38: todo: support exclude/field_mapping/fields/localns

starlite/plugins/pydantic.py
```diff
@@ -0,0 +1,165 @@
...
+from .utils import get_field_type, remap_field
+
+if TYPE_CHECKING:
+    from typing import Any, Mapping, TypeGuard
+
+    from starlite.enums import RequestEncodingType
+
+
+class FromPydantic(ABC, SerializationPluginProtocol[ModelT, BaseModel], Generic[ModelT]):
+    """Plugins that use pydantic as transfer type."""
```
peterschutt, line 21:
```suggestion
    """Plugins that use pydantic as container type."""
```

starlite/plugins/pydantic.py
```diff
@@ -0,0 +1,165 @@
...
+from .base import ModelT, SerializationPluginProtocol
+from .utils import get_field_type, remap_field
+
+if TYPE_CHECKING:
+    from typing import Any, Mapping, TypeGuard
+
+    from starlite.enums import RequestEncodingType
+
+
+class FromPydantic(ABC, SerializationPluginProtocol[ModelT, BaseModel], Generic[ModelT]):
```
peterschutt, line 20: consider name

starlite/plugins/pydantic.py
```diff
@@ -0,0 +1,165 @@
...
+        fields = fields or {}
+        for field_name, model_field in model_class.__fields__.items():
+            if field_name in exclude:
+                continue
+            field_type = get_field_type(model_field=model_field)
+            if field_name in field_mappings:
+                field_name, field_type = remap_field(field_mappings, field_name, field_type)
+            fields[field_name] = field_type, model_field.field_info
+        return create_model(
+            model_class.__name__,
```
peterschutt, line 121: different name?

starlite/plugins/sql_alchemy/plugin.py
```diff
@@ -351,59 +346,63 @@ def parse_model(model_class: Type[DeclarativeMeta]) -> Mapper:
...
+                if field_name in exclude:
+                    continue
                 if column.default and type(column.default.arg) in ModelFactory.get_provider_map():
-                    field_definitions[name] = (self.get_pydantic_type(column.type), column.default.arg)
+                    field_type = (self.get_pydantic_type(column.type), column.default.arg)
                 elif not column.nullable:
-                    field_definitions[name] = (self.get_pydantic_type(column.type), ...)
+                    field_type = (self.get_pydantic_type(column.type), ...)
                 else:
-                    field_definitions[name] = (self.get_pydantic_type(column.type), None)
```
peterschutt, line 375: i nuked the relationship parsing for now to simplify.

todo: de-nuke

starlite/plugins/tortoise_orm.py
```diff
@@ -22,18 +23,71 @@
...
     from typing_extensions import TypeGuard
 
+    from starlite.enums import RequestEncodingType
+
+
+TortoiseModelT = TypeVar("TortoiseModelT", bound="Model")
+
 
-class TortoiseORMPlugin(SerializationPluginProtocol[Model, BaseModel], OpenAPISchemaPluginProtocol[Model]):
+class TortoiseORMPlugin(SerializationPluginProtocol[Model, PydanticModel], OpenAPISchemaPluginProtocol[Model]):
```
peterschutt, line 37: can prob just extract out a mixin class to dedup the code between here and `FromPydantic`

starlite/plugins/tortoise_orm.py
```diff
@@ -58,23 +112,40 @@ def _create_pydantic_model(model_class: Type[Model], **kwargs: Any) -> "Type[Pyd
...
 
         This function uses memoization to ensure we don't recompute unnecessarily.
+
+        :param model_class: The model class that serves as a basis.
+        :param exclude: do not include fields of these names in the model.
+        :param field_mappings: to rename, and re-type fields.
+        :param fields: additional fields to add to the container model.
+        :param localns: additional namespace for forward ref resolution.
+        :param kwargs: Any kwargs.
+        :return: The generated data container class.
```
peterschutt, line 138: todo: field mappings/fields/localns

starlite/plugins/tortoise_orm.py
```diff
@@ -86,29 +157,32 @@ def to_data_container_class(self, model_class: Type[Model], **kwargs: Any) -> Ty
...
         This class is passed in as the ``model_class`` kwarg.
         """
         return model_class().update_from_dict(data_container_instance.dict())
 
-    async def to_dict(self, model_instance: Model) -> Dict[str, Any]:  # pylint: disable=invalid-overridden-method
+    def to_dict(self, model_instance: TortoiseModelT) -> "dict[str, Any]":
         """Given an instance of a model supported by the plugin, return a dictionary of serializable values."""
         pydantic_model_class = self.to_data_container_class(type(model_instance))
-        data = await pydantic_model_class.from_tortoise_orm(model_instance)
+        with anyio.start_blocking_portal() as portal:
```
peterschutt, line 177: using a blocking portal here instead of making the protocol support both dict and awaitable return values for this one case.. but the tortoise tests end up hanging.. i've `pytest.mark.skip`ped them for now.

@starlite-api/maintainers anyone know how to handle this better (preferably modify the tests, assuming this is fine at runtime).

starlite/plugins/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        field_name, field_type = mapping
+    else:
+        field_name = mapping
+    return field_name, field_type
+
+
+def create_field_definitions(
+    exclude: set[str], field_mapping: Mapping[str, str | tuple[str, Any]], fields: dict[str, ModelField]
+) -> dict[str, tuple[Any, Any]]:
+    """Populate ``field_definitions``, ignoring fields in ``exclude``, and remapping fields in ``field_mapping``."""
```
peterschutt, line 45: where is this used??

starlite/signature/models.py
```diff
@@ -139,6 +141,17 @@ def is_required(self) -> bool:
 
         return not (self.is_optional or self.is_any) and (self.is_empty or self.default_value is None)
 
+    @property
+    def is_dto_field(self) -> bool:
```
peterschutt, line 145: todo: test this with forward ref annotations

tests/plugins/sql_alchemy_plugin/test_sql_alchemy_relationships.py
```diff
@@ -1,12 +1,14 @@
...
 
+import pytest
 from pydantic import BaseModel
 from typing_extensions import get_args
 
 from starlite.plugins.sql_alchemy import SQLAlchemyPlugin
 from tests.plugins.sql_alchemy_plugin import Company, Pet, User
 
 
+@pytest.mark.xfail(reason="TODO: add relationships back into sqlalchemy plugin")
```
peterschutt, line 11: todo

tests/plugins/tortoise_orm/test_tortoise_orm_plugin_integration.py
```diff
@@ -11,6 +13,8 @@
 )
 
 
+# TODO
+@pytest.mark.skip(reason="I've broken something by using the blocking portal")
```
peterschutt, line 17: todo: fix these tortoise tests, or rollback the blocking portal thing

starlite/plugins/tortoise_orm.py
Goldziher: You dont need the portal, you can do this:

```python
anyio.run(pydantic_model_class.from_tortoise_orm, model_instance)
```

starlite/plugins/utils.py
Goldziher: dunno, need to check. itsa utility for creating pydantic models it seems.

starlite/handlers/http_handlers.py
Goldziher: well, the point was that we were automatically generating the DTO there i think, no?

starlite/plugins/tortoise_orm.py
peterschutt: I started there - it was failing along the lines of there already being a running event loop.

Goldziher: And

```python
from_thread.run_sync(...)
```

?

starlite/handlers/http_handlers.py
peterschutt: Yes that is the point. My issue is that the code the pattern supports is a type error.

I will add the handling back if the consensus is against me. @provinzkraut any thoughts?

starlite/plugins/tortoise_orm.py
peterschutt: I'll start again from the beginning and see if I was missing something obvious yesterday..

starlite/plugins/utils.py
peterschutt: It was in the old `dto.py` module, then I moved it to `plugins/utils.py` but I'm not sure that I'm actually using it.. just a note to remind myself to check.

starlite/handlers/http_handlers.py
provinzkraut: I agree with @peterschutt that we shouldn’t support this as it is currently. If we want to do an implicit conversion, we’ll have to put the dto type elsewhere. Maybe a parameter on the handler or something like that?

Goldziher: We need to automatically generate DTOs to have feature parity with FastAPI. There you can define a `response model` and use it to return a response and generate docs. This is our equivalent. 

It can be done in a different pattern, as long as we keep parity.

provinzkraut: > It can be done in a different pattern, as long as we keep parity.

That's what I was thinking of. Maybe a `response_dto` kwarg on the route handler? That wouldn't mess with the typing and be more explicit.

starlite/plugins/base.py
```diff
@@ -91,32 +78,78 @@ class SerializationPluginProtocol(Protocol[ModelT, DataContainerT]):
...
+        Args
+            container_instance: the container instance
+
+        Returns
+            dict representation of container instance
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    def is_plugin_supported_type(value: Any) -> bool:
```
provinzkraut, line 129: I think this would better be called `supports_type`, as it makes more sense semantically.

`plugin.supports_type(foo)` vs `plugin.is_plugin_supported_type(foo)`. The latter feels a bit redundant.

starlite/handlers/http_handlers.py
peterschutt: > We need to automatically generate DTOs to have feature parity with FastAPI. There you can define a `response model` and use it to return a response and generate docs. This is our equivalent.
> 
> It can be done in a different pattern, as long as we keep parity.

For clarity, I still expect that we'll support this pattern:

```python
@post()
def handler(data: DomainType) -> DomainType:
   return data
```

Where, if `DomainType` is supported by a plugin, we will generate a data transfer model internally on startup, and that transfer type will represent _all_ attributes of the domain model - so it is probably only useful in narrow circumstances.

And @provinzkraut, you are suggesting something like this:

```python
@post(data_dto=DataDTO, response_dto=ResponseDTO)
def handler(data: DomainType) -> DomainType:
    return data
```
Where in this case, we use `DataDTO.from_bytes(...).to_model()` to instantiate the `DomainType` instance that is injected into the handler, and `ResponseDTO.from_model(data).to_bytes()` with the response from the handler to generate the actual response that is sent over the wire.

This would be functionally equivalent to:

```python
@post()
def handler(data: DataDTO) -> ResponseDTO:
    domain_instance = data.to_model()
    return ResponseDTO.from_model(domain_instance)
```

Which saves the explicit conversion to/from the dto type in user-land.

I think that is a nice pattern, and is effectively doing the same as:

```python
def handler() -> dto.Factory[MyModel]:
   ...
   return MyModel()  # type error
```
... but without the type errors.

peterschutt:
Current state of play:
- `dto.Factory` is a factory type, used to create type-specific dto types via type narrowing
- A `dto.Factory` has to be subclassed to provide it with a `SerializationPluginProtocol` type for it to be usable.
- This plugin holds all of the conversion utilities for:
  - bytes to container type instance/array of instances
  - container type instance to dict
  - container type instance from dict
  - model type instance to dict
  - k/v pairs to model instance
  - from model type to container type
  - from model instance to container instance
- e.g., `DataclassFactory` is responsible for conversion of raw data to/from domain models using python dataclasses as the domain modelling "type" for the application, and uses the `DataclassPlugin` to do a lot of the conversion heavy lifting.
- to create a dto type specific for a type, one uses type-narrowing, e.g., `MyDTO = DataclassFactory[MyDC]`, internally, this creates a "transfer model" which is a type of some ser/de type, be it pydantic, msgspec etc. The transfer model can be configured using `Annotated` e.g., `MyDTO = DataclassFactory[Annotated[MyDC, dto.Config()]`. The type of transfer model produced is dictated by the serialization plugin type.
- standard use of a dto would be something like:
```python
# transfering from bytes to a model instance
the_thing = MyDTO.from_bytes(buffer, media_type).to_model()
# do stuff with the model instance
the_thing.a += 1
# serialize the model instance
MyDTO.from_model(the_thing).to_bytes(media_type)
```

Things still to do:
- code needs a few passes and more tests
- i've ripped out a bit of functionality as I've gone, e.g., made `SerializationPluginProtocol.to_dict()` only sync now, and removed the generated container type caching. The issues left behind by removing these, and lost functionality need to be looked at through the prism of the new implementation (e.g., a type-narrowed subclass of `dto.Factory` does keep a cached version of the  model it has generated to represent that model/use-case).
- the number of methods and especially the number of static methods on the `SerializationPluginProtocol` tells me that should probably be simplified. I'm thinking of moving all of the methods that interact with the model type only from the serialization plugin to the dto factory type. So it would be that:
  -  a dto subclass owns a domain model type (e.g., `DataclassFactory`, `SQLAlchemyFactory` etc) and they are aware of a serialization plugin protocol implementation. 
  - Plugin protocols own the serialization library types (`PydanticPlugin`, `MsgspecPlugin` etc) and are totally unaware of the dtos and the domain modelling types. 
  - Of course, msgspec and pydantic models could be used for both domain model representation and ser/de so we'll have both a dto and ser plugin implementation for them.
- support for giving a `type_encoders` mapping to the dtos (probably both via subclassing and via `dto.Config()`.
- resolve [this](https://github.com/starlite-api/starlite/pull/1160#discussion_r1102687177)
- anything else???

ping @starlite-api/maintainers

starlite/handlers/http_handlers.py
Goldziher: That's great then, no issues from my end. We should support also i think specifying a dto_type as well to allow users to exclude fields etc. Or we can allow passing drop factory options via route handler kwargs or extra

provinzkraut: > And @provinzkraut, you are suggesting something like this:

Yes, that's exactly what I was thinking of!

> We should support also i think specifying a dto_type as well to allow users to exclude fields etc. Or we can allow passing drop factory options via route handler kwargs or extra

Would this not already be covered by e.g. `response_dto`, since the user could easily create a DTO with field exclusion from a factory?

```python
@post(response_dto=dto_factory("DomainTypeWithoutID", DomainType, exlude=["id"]))
def handler(data: DomainType) -> DomainType:
    return data
```

Although I'd argue the proper pattern to encourage here would be to define the DTO outside the handler so it's a static class that's available in the namespace.

provinzkraut:
@peterschutt 

I don't have any comments about the content as such, sounds all very good to me.

But it also sounds like there's still a lot of work to do. Let us know if you need a hand with any of this?

provinzkraut:
@peterschutt 

I don't have any comments about the content as such, sounds all very good to me.

But it also sounds like there's still a lot of work to do. Let us know if you need a hand with any of this?

peterschutt:
Thanks @provinzkraut - I've had a pretty full on couple of days with work, but a couple of days off coming up so I'm going to try to churn through this over the next couple and then assess. There's still a couple of other tasks I'd like to get through for 2.0 aside from this as well (sqlalchemy 2.0 plug-in and partial) so if you'd like to take on any of them you are welcome. Although, they both really depend on this task.

.pre-commit-config.yaml
```diff
@@ -173,6 +173,7 @@ repos:
             beanie,
             beautifulsoup4,
             brotli,
+            bs4,
```
peterschutt, line 176:
```suggestion
```

Goldziher (approved):

docs/examples/lifecycle_hooks/after_response.py
```diff
@@ -15,7 +15,7 @@ async def hello() -> Dict[str, int]:
     return COUNTER
 
 
-app = Starlite(route_handlers=[hello], after_response=after_response)
+app = Starlite(route_handlers=[hello], after_response=after_response, debug=True)
```
peterschutt, line 18:
```suggestion
app = Starlite(route_handlers=[hello], after_response=after_response)
```

peterschutt, line 18:
```suggestion
app = Starlite(route_handlers=[hello], after_response=after_response)
```

Issue #1136, opened 2023-02-03
Bug: JWT response body not using custom type encoders


WardPearce:
**Describe the bug**
JWT login parameter `response_body`, doesn't use custom type encoders. Resulting in responses not having consistent schema.

**To Reproduce**
```py
from typing import Any

from pydantic import BaseModel, Field
from starlite import ASGIConnection, Response, Starlite, TestClient, get
from starlite.contrib.jwt import JWTCookieAuth, Token


class User(BaseModel):
    id: str = Field(..., alias="_id")


async def retrieve_user_handler(
    token: Token, connection: ASGIConnection[Any, Any, Any]
) -> User:
    return User(_id=token.sub)


jwt_cookie_auth = JWTCookieAuth[User](
    retrieve_user_handler=retrieve_user_handler,
    token_secret="abc1234",
    exclude=["/"],
)


@get()
def handler() -> Response[User]:
    data = User(_id="1")
    return jwt_cookie_auth.login(identifier=str(data.id), response_body=data)


app = Starlite([handler], type_encoders={BaseModel: lambda m: m.dict(by_alias=True)})


with TestClient(app=app) as client:
    res = client.get("/")
    # This should pass, but doesn't because alias not being used.
    assert res.json() == {"_id": "1"}
```

cofin:
I took a quick look at this.  As you'd expect, if I modify the response in `JWTCookieAuth` to look like:

```python
        return Response(
            content=response_body,
            headers={self.auth_header: self.format_auth_header(encoded_token)},
            cookies=[cookie],
            media_type=response_media_type,
            status_code=response_status_code,
            type_encoders={BaseModel: lambda m: m.dict(by_alias=True)},
        )
```
The test case above passes.  

I'll keep investigating, but this seems to confirm that the app type_encoders aren't getting passed into the response when calling `Response` class directly.

Goldziher:
PR add @WardPearce -- please check it out and give your input

peterschutt:
@all-contributors add @WardPearce for bug

PR #1167, opened 2023-02-07
add support to type encoders for JWT config

Goldziher:
# PR Checklist

### DIFF
--- a/tests/contrib/jwt/test_auth.py
+++ b/tests/contrib/jwt/test_auth.py
@@ -5,10 +5,11 @@ from uuid import uuid4
 
 from hypothesis import given, settings
 from hypothesis.strategies import integers, none, one_of, sampled_from, text, timedeltas
+from pydantic import BaseModel, Field
 
 from starlite import ASGIConnection, Request, Response, Starlite, get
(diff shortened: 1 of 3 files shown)


WardPearce:
Adding `type_encoders` to `JWTAuth` seems like a easy fix, It does add a couple different points a developer must configure in order to have a consistent response schema but doesn't seem like a big deal.

tests/contrib/jwt/test_auth.py
```diff
@@ -377,3 +378,29 @@ def login_custom_handler() -> Response["User"]:
...
+        type_encoders={BaseModel: lambda m: m.dict(by_alias=True)},
+    )
+
+    @get()
+    def handler() -> Response[User]:
+        data = User(_id="1")
+        return jwt_cookie_auth.login(identifier=str(data.id), response_body=data)
+
+    with create_test_client([handler]) as client:
+        response = client.get("/")
```
WardPearce, line 405: Maybe assert JSON body matches as expected?

WardPearce (approved):

peterschutt:
Wouldn't this be better defined on `AbstractSecurityConfig`? Also I think we can pull in the custom type encoders that are defined on the application via the on-app-init hook, so the OP example in #1136 should work.

I'll PR into this branch.

PR #1169, opened 2023-02-07
Moves `type_encoders` config onto `AbstractSecurityConfig`.

peterschutt:
If the on-app-init hook handler is passed to the application, and the `type_encoders` is not explicitly set on the config, we grab the custom encoders from the application config.

# PR Checklist


peterschutt:
Given this PR, if #1136 OP changed:

```py
app = Starlite([handler], type_encoders={BaseModel: lambda m: m.dict(by_alias=True)})
```
to 
```py
app = Starlite([handler], type_encoders={BaseModel: lambda m: m.dict(by_alias=True)}, on-app-init=[jwt_cookie_auth.on_app_init])
```
The custom type decoders of the app would automatically be referenced by the jwt config.

PR #1173, opened 2023-02-08
Added configuration to starlite constructor

### LINKED ISSUE #1120: Enhancement: Allow configuration of an application via an `AppConfig` instance
We already have `AppConfig` in place, so it would be nice to be able to do something like 

```python
from starlite import Starlite, AppConfig

config = AppConfig(...)

app = Starlite.from_config(config)
```

wassafshahzad:
# PR Checklist

### Which ticket is this PR associated with ?
Closes #1120 

### What does this PR do
This PR adds the following features
- Adds a configuration object in Starlite constructor
- Adds sane defaults in AppConfig
- Updates pre-commits to fix isort issue

### DIFF
--- a/docs/examples/tests/openapi/test_customize_pydantic_model_name.py
+++ b/docs/examples/tests/openapi/test_customize_pydantic_model_name.py
@@ -11,6 +11,7 @@ def test_schema_generation() -> None:
             "paths": {
                 "/id": {
                     "get": {
+                        "summary": "RetrieveIdHandler",
                         "operationId": "IdRetrieveIdHandler",
                         "responses": {
                             "200": {
@@ -24,7 +25,7 @@ def test_schema_generation() -> None:
                             }
                         },
                         "deprecated": False,
-                    },
+                    }
                 }
             },
             "components": {
--- a/poetry.lock
+++ b/poetry.lock
@@ -1324,14 +1324,14 @@ socks = ["socksio (>=1.0.0,<2.0.0)"]
 
 [[package]]
 name = "hypothesis"
-version = "6.67.1"
+version = "6.68.0"
 description = "A library for property-based testing"
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "hypothesis-6.67.1-py3-none-any.whl", hash = "sha256:947839d8b5fd24d9b4d1e41ee10f7285d023f49f3b318519bf110fa9e5f3154e"},
-    {file = "hypothesis-6.67.1.tar.gz", hash = "sha256:7abb4361d796ae6432d1c0b22a514210a2a19cb0d85efca8463e619a80f3132c"},
+    {file = "hypothesis-6.68.0-py3-none-any.whl", hash = "sha256:064a5bf7a98cbdfa3589d8934a2c548524c6d108bfa368ec5ed8b09caa526108"},
+    {file = "hypothesis-6.68.0.tar.gz", hash = "sha256:8cfa1de15271debb2750545a460e58e3e84527d94157f38ab9309826953572f3"},
 ]
 
 [package.dependencies]


starlite/app.py
```diff
@@ -158,7 +158,7 @@ class Starlite(Router):
 
     def __init__(
```
provinzkraut, line 159: As previously discusses in #1120, this method should be [overloaded](https://docs.python.org/3/library/typing.html#typing.overload) so type checkers will assert that it's only ever called with either an `AppConfig` instance, **or** keyword args, not both. This way we can also ensure that `route_handlers` is not `None` when using the kwarg style, which should be checked explicitly aside from that as well IMO.

starlite/app.py
```diff
@@ -200,6 +200,7 @@ def __init__(
         template_config: Optional["TemplateConfig"] = None,
         type_encoders: Optional["TypeEncodersMap"] = None,
         websocket_class: Optional[Type["WebSocket"]] = None,
+        configuration: Optional[AppConfig] = None,
```
provinzkraut, line 203:
```suggestion
        config: Optional[AppConfig] = None,
```

`config` is the canonical way we refer to configuration objects (`AppConfig`, not `AppConfiguration`)

starlite/app.py
```diff
@@ -277,6 +278,8 @@ def __init__(
             type_encoders: A mapping of types to callables that transform them into types supported for serialization.
             websocket_class: An optional subclass of :class:`WebSocket <starlite.connection.websocket.WebSocket>` to use for
                 websocket connections.
+            configuration: An optional instance of :class:`AppConfig` <startlite.config.AppConfig> to replace group up initial
+                configurations. If passed in addition to kwargs related to configurations, those kwargs will be ignored.
```
provinzkraut, line 282:
```suggestion
            config: An instance of :class:`AppConfig` <startlite.config.AppConfig>. If passed, configure the application with values from this object instead of kwargs.
```

provinzkraut (changes requested):

starlite/app.py
peterschutt: As `on-app-init` plugins can also provide routes, and routes can be registered after app creation, it is hard to definitively say that `route_handlers` would _always_ be required in the case that a config object isn't directly passed to the constructor.

I don't think that will be a very common case, and the workaround in that instance would be to pass an empty list like we have to in 1.x when instantiating an app without handlers. But it is worth considering if we need to be so strict on that `route_handlers` requirement.

starlite/app.py
```diff
@@ -158,7 +158,7 @@ class Starlite(Router):
 
     def __init__(
         self,
-        route_handlers: List["ControllerRouterHandler"],
+        route_handlers: Optional[List["ControllerRouterHandler"]] = None,
```
peterschutt, line 161: I think you need to update your fork @wassafshahzad - the current signature for this is:

```py
route_handlers: "OptionalSequence[ControllerRouterHandler]" = None,
```

starlite/config/app.py
```diff
@@ -47,7 +47,7 @@ class AppConfig(BaseModel):
     class Config(BaseConfig):
         arbitrary_types_allowed = True
 
-    after_exception: SingleOrList[AfterExceptionHookHandler]
+    after_exception: SingleOrList[AfterExceptionHookHandler] = []
     """An application level :class:`exception hook handler <starlite.types.AfterExceptionHookHandler>` or list thereof.
```
peterschutt, line 51: these have also changed since you forked, e.g., `after_exception: List[AfterExceptionHookHandler]` is current typing for this.

starlite/app.py
wassafshahzad: @provinzkraut @peterschutt I am currently trying to understand the overload decorator but unfortunately I was unable to make a signature which would make sure that __init__ is called with either AppConfig or route_handlers not both 

Best I could come up with is, I would greatly appreciate some guidance
```
    @overload
    def __init__(self, config: AppConfig) -> None:
        ...
    
    @overload
    def __init__(self, route_handlers: List["ControllerRouterHandler"]) -> None:
        ...

```

Goldziher: I don't think overloads are the correct way to handle this, @starlite-api/maintainers . Since we are dealing here with Kwargs its going to be hard and a workaround will be pretty ugly. Instead I suggest the following, @wassafshahzad:

Create a static "factory" method called `from_config` on the Starlite class:

```python
class Starlite:
    def init(...):
       ...
     
    @static
    def from_config(config: "AppConfig") -> Starlite:
          return Starlite(**config.dict())
```

provinzkraut: Why not a `classmethod?

Goldziher: sure, a class method is also fine

wassafshahzad: Thanks for all the help guys

provinzkraut:
@wassafshahzad What's up with this? Are you no longer working on it?

wassafshahzad:
> @wassafshahzad What's up with this? Are you no longer working on it?

No I am I reverted my commit and rebased from main 
Then pushed the changes causing git to think it had 0 commits
I reverted my changes since I didn't want to pollute the commit history

provinzkraut:
Well I was a bit confused because the PR is closed. Care to reopen it?

wassafshahzad:
@provinzkraut I have the PR ready but running pre-commit produces errors on lines I did not even change for example

starlite\app.py:404: error: Item "None" of "Optional[Sequence[Union[Type[Controller], Union[HTTPRouteHandler, WebsocketRouteHandler, ASGIRouteHandler], Router, Callable[..., Any]]]]" has no attribute "__iter__" (not iterable)  [union-attr]

provinzkraut:
Please reopen the PR and push your latest changes so this can be checked.

peterschutt:
> @provinzkraut I have the PR ready but running pre-commit produces errors on lines I did not even change for example
> 
> 
> 
> starlite\app.py:404: error: Item "None" of "Optional[Sequence[Union[Type[Controller], Union[HTTPRouteHandler, WebsocketRouteHandler, ASGIRouteHandler], Router, Callable[..., Any]]]]" has no attribute "__iter__" (not iterable)  [union-attr]

This is because you'd changed the types of some of the collections on the config object to be optional.

PR #1176, opened 2023-02-08
refactor plugin protocol

Goldziher:
# PR Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -530,7 +530,7 @@ Changelog
 ------
 
 -  refactored brotli middleware typing.
--  update Extended ``PluginProtocol`` with an ``on_app_init`` method.
+-  update Extended ``SerializationPluginProtocol`` with an ``on_app_init`` method.
 
 
 
@@ -968,7 +968,7 @@ Changelog
 -  add support for multiple paths per route handler.
 -  add support for static files.
 -  update ``DTOFactory``.
--  update ``PluginProtocol`` - add ``from_dict`` methods.
+-  update ``SerializationPluginProtocol`` - add ``from_dict`` methods.
 -  update ``SQLAlchemyPlugin``.
 -  update dependency injection to allow for dependency injection into dependencies.
 -  update lifecycle support to allow for application state injection.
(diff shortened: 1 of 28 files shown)


CHANGELOG.rst
```diff
@@ -530,7 +530,7 @@ Changelog
 ------
 
 -  refactored brotli middleware typing.
--  update Extended ``PluginProtocol`` with an ``on_app_init`` method.
+-  update Extended ``SerializationPluginProtocol`` with an ``on_app_init`` method.
```
Goldziher, line 533:
```suggestion
-  update Extended ``PluginProtocol `` with an ``on_app_init`` method.
```

CHANGELOG.rst
```diff
@@ -968,7 +968,7 @@ Changelog
 -  add support for multiple paths per route handler.
 -  add support for static files.
 -  update ``DTOFactory``.
--  update ``PluginProtocol`` - add ``from_dict`` methods.
+-  update ``SerializationPluginProtocol`` - add ``from_dict`` methods.
```
Goldziher, line 971:
```suggestion
-  update ``PluginProtocol `` - add ``from_dict`` methods.
```

peterschutt (approved):
OK, plugins are now generic on both the source data type and the transform type.

I assume this is to allow us to make plugins that transform into msgspec structs? 

I know that I'd prefer my DTOs to be msgspec structs and parsed directly from the request buffer, but this will allow users to have a choice.

This plays into a redesign I was already considering for the DTO refactor. At the moment, DTO instances _are_ pydantic models, but they probably need to be their own type, and hold a reference to a thing that does the ser/de and validation, be that a pydantic model, or msgspec struct or whatever, and what that thing is will depend on the type of plugin the dto factory is given.

E.g, if a DTO factory type is given a plugin that is narrowed to pydantic base model, then it will produce a ser/de object that is a pydantic model, if it is given a plugin narrowed to msgspec structs, then it will produce a struct for its ser/de object.

Before this, I was not confident about meeting the 20th deadline (not just for the DTO refactor, but for everything I want to have in 2.0), but I think will be a positive redesign.

PR #1187, opened 2023-02-11
Fix critical bug in parsing headers

knowsuchagency:
# Description

This bug addresses a bug when parsing HTTP headers from `scope["headers"]`.

Currently, the `parse_headers` function is decorated with `lru_cache` which can't hash lists. 

This is problematic because lists (not tuples) are what the ASGI protocol allows for since they're serializable. 

The ASGI server will pass lists of lists to `parse_headers` leading to `TypeError`.

We are currently running into this bug with starlite `1.51.0`

> Because these messages could be sent over a network, they need to be serializable, and so they are only allowed to contain the following types:

    Byte strings
    Unicode strings
    Integers (within the signed 64-bit range)
    Floating point numbers (within the IEEE 754 double precision range; no Nan or infinities)
    Lists (tuples should be encoded as lists)
    Dicts (keys must be Unicode strings)
    Booleans
    None

This PR maintains the current internal API for `parse_headers` by casting its input to a tuple of tuples and passing it to the original, decorated function.

# PR Checklist

### DIFF
--- a/starlite/parsers.py
+++ b/starlite/parsers.py
@@ -1,7 +1,7 @@
 import os
 from functools import lru_cache
 from http.cookies import _unquote as unquote_cookie
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Tuple, Iterable
 from urllib.parse import unquote
 
 from fast_query_parsers import parse_query_string as fast_parse_query_string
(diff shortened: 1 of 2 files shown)


knowsuchagency:
@all-contributors please add @knowsuchagency  for bug

peterschutt:
Thanks @knowsuchagency someone will review asap.

peterschutt (approved):
Thanks again!

knowsuchagency:
Thank you @peterschutt !

PR #1209, opened 2023-02-17
Simplify response headers / cookies

provinzkraut:
# Content

- Change the type of  `response_headers` to `Sequence[ResponseHeader]  | Mapping[str, str]`
- Change the type of  `response_cookies` to `Sequence[Cookie]  | Mapping[str, str]`
- Change the signature of `Response` for:
    - `Response(headers=...)`:  `Sequence[ResponseHeader] | Mapping[str, str] | None`
    - `Response(cookies=...)`: `Sequence[Cookie] | Mapping[str, str] | None` 
- Add stricter typing to `ResponseHeader`, analogous to `Cookie`
- Minor refactoring of cookie/header normalization in `http_handlers.py`

Before:
```python
@get(
    response_headers={"foo": ResponseHeader(value="bar")},
    response_cookies=[Cookie(key="foo", value="bar")],
)
async def handler() -> str:
    ...
```

After:
```python
@get(
    response_headers=[ResponseHeader(name="foo", value="bar")],
    response_cookies=[Cookie(key="foo", value="bar")],
)
async def handler() -> str:
    ...
``` 

Overriding behaviour is kept the same, internal types are updated such that `resolve_response_headers` returns a `frozenset` like `resolve_response_cookies` does.

# Further ideas

In addition to the proposed change, I was thinking if we maybe want to update `response_headers` and `response_cookies` to accept plain `Mapping[str, str]`. @starlite-api/maintainers please comment.

# PR Checklist


odiseo0:
Makes sense to me, why it wasn't like that before?

starlite/handlers/http_handlers.py
```diff
@@ -86,7 +85,7 @@ def _normalize_cookies(local_cookies: FrozenSet["Cookie"], layered_cookies: Froz
     return [cookie for cookie in sorted_set if not cookie.documentation_only]
 
 
-def _normalize_headers(headers: "ResponseHeadersMap") -> Dict[str, Any]:
+def _normalize_headers(headers: FrozenSet["ResponseHeader"]) -> Dict[str, Any]:
```
Goldziher, line 88: Let's `lru_cache` this

Goldziher (approved):

peterschutt:
I like.

> In addition to the proposed change, I was thinking if we maybe want to update response_headers and response_cookies to accept plain Mapping[str, str]

so `response_headers={"foo": "bar", "baz": "qux"}` is functionally equivalent to `response_headers=[ResponseHeader(name="foo", value="bar"), ResponseHeader(name="baz", value="qux")]`?

Alternatively, `response_headers` could be typed `Sequence[tuple[str, str] | ResponseHeader]`, e.g., `response_headers=[("foo", "bar"), ResponseHeader(name="baz", value="quz", documentation_only=True)`.

provinzkraut:
> Alternatively, response_headers could be typed `Iterable[tuple[str, str] | ResponseHeader]`

The thing with that is, that it's a bit counterintuitive when not using a mix of strings and `ResponseHeader`s:
`response_headers=[("foo", "bar"), ("baz", "quz")]` just doesn't seem right compared to `response_headers={"foo": "bar", "baz": "quz"}`. The latter also gives you immediate feedback for duplicate keys in your IDE / type checker.

I was contemplating what the best solution would be here, and I settled for `Mapping[str, str] | Sequence[ResponseHeader]` because it's the easiest, and it does actually support all the use cases. 

The "most complete" would probably be something like `Mapping[str, str] | Iterable[ResponseHeader | tuple[str, str]]`,
so you can do `response_headers={"foo": "bar"}`, `response_headers=[ResponseHeader(name="foo", value="bar", documentation_only=True)]` or `response_headers=[("foo", "bar"), ResponseHeader(name="baz", value="quz", documentation_only=True]` if you really want the mix.

I'm not certain if we *need* to provide the flexibility to mix both styles, since `Sequence[ResponseHeader]` already allows to express everything you could, and `Mapping[str, str]` is basically just a convenience shortcut. I'd also argue that the mixed style leads to poor readability, and we might simply not want to promote such a pattern. 

wdyt?

provinzkraut:
@peterschutt I added a proposal for those changes in #1213

JacobCoffee (approved):

starlite/handlers/http_handlers.py
```diff
@@ -497,36 +504,53 @@ def resolve_response_class(self) -> Type["Response"]:
                 return layer.response_class
         return Response
 
-    def resolve_response_headers(self) -> "ResponseHeadersMap":
+    def resolve_response_headers(self) -> FrozenSet["ResponseHeader"]:
```
provinzkraut, line 507: Should those be cached on the instance as well?

starlite/handlers/http_handlers.py
```diff
@@ -497,36 +504,53 @@ def resolve_response_class(self) -> Type["Response"]:
...
 
         Returns:
             A dictionary mapping keys to :class:`ResponseHeader <starlite.datastructures.ResponseHeader>` instances.
         """
         resolved_response_headers: dict[str, "ResponseHeader"] = {}
+
         for layer in self.ownership_layers:
-            resolved_response_headers.update(layer.response_headers or {})
+            if layer_response_headers := layer.response_headers:
+                if isinstance(layer_response_headers, Mapping):
```
provinzkraut, line 517: I'm unsure if this is the right approach. This can't happen unless you modify a handler instance, which isn't a case we necessarily need to cover IMO, but from a type checking perspective, this is correct, since `Controller` doesn't differentiate between the type given in `__init__`  and the one set on the instance.  Maybe we should just raise an exception in that case, or `# type: ignore` it?

peterschutt:
> wdyt?

Sounds like you've put a lot more thought into this than I have, so I'm on board with what you think is best.

PR #1216, opened 2023-02-19
Future annotation improvements

Goldziher:
# PR Checklist

### DIFF
--- a/starlite/utils/compat.py
+++ b/starlite/utils/compat.py
@@ -32,7 +32,7 @@ def validate_arguments(fn: C) -> C:
     :param fn: A callable to decorate.
     :return: Either the decorated callable, if python version is 3.8, or the callable itself.
     """
-    if version_info < (3, 9):
+    if version_info < (3, 9):  # pragma: no cover
         return fn
 
     from pydantic import validate_arguments as pydantic_validate_arguments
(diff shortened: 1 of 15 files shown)


starlite/utils/compat.py
```diff
@@ -19,3 +20,23 @@ async def async_next(gen: AsyncGenerator[T, Any], default: D | EmptyType = Empty
...
+
+
+def validate_arguments(fn: C) -> C:
+    """Proxy pydantic `validate_arguments` decorator to ignore python 3.8, which doesn't support future
+    annotations.
+
+    :param fn: A callable to decorate.
+    :return: Either the decorated callable, if python version is 3.8, or the callable itself.
+    """
+    if version_info < (3, 9):  # pragma: no cover
```
provinzkraut, line 35: So we just don't validate the arguments for `3.8`? That should definitely be mentioned in the docs somewhere then, since it could cause bugs if you rely on the validation

tests/test_typing.py
```diff
@@ -1,4 +1,5 @@
 import dataclasses
+from types import NoneType
```
provinzkraut, line 2: `NoneType` is 3.10+ only

provinzkraut (changes requested):

PR #1223, opened 2023-02-20
fix: JWT Auth correctly handles `domain` configuration value.

cofin:
The JWT Cookie Auth configuration was missing the parameter for `domain` when setting the response cookie.  This PR adds that parameter.

### DIFF
--- a/tests/contrib/jwt/test_auth.py
+++ b/tests/contrib/jwt/test_auth.py
@@ -208,6 +208,7 @@ async def test_jwt_cookie_auth(
             "/my-endpoint",
         )
         assert response.status_code == HTTP_200_OK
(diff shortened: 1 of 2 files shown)


peterschutt (approved):

tests/contrib/jwt/test_auth.py
```diff
@@ -208,6 +208,7 @@ def login_handler() -> Response["User"]:
             "/my-endpoint",
         )
         assert response.status_code == HTTP_200_OK
+        assert response.cookies.get(jwt_auth.key) == f"key={jwt_auth.key}; Path=/; SameSite=lax"
```
cofin, line 211: This needs to be updated actually.

PR #1229, opened 2023-02-21
Fix #1228: create a new dictionary of values in obfuscate()

### LINKED ISSUE #1228: Bug: LoggingMiddleware is sending obfuscated session id to client
**Describe the bug**
When using the logging middleware and session middleware together, the logging middleware's cookie obfuscation is overwriting the session name with "*****" and that name is being pushed down to the client.

The initial set-cookie has the correct session id but subsequent requests do not.

**To Reproduce**
I created a test function in tests/middleware/test_logging_middleware.py which I believe confirms the bug:

```python
def test_logging_with_session_middleware() -> None:
    @post("/")
    async def set_session(request: Request) -> None:
        request.set_session({"hello": "world"})

    @get("/")
    async def get_session() -> None:
        pass

    logging_middleware_config = LoggingMiddlewareConfig()
    session_config = MemoryBackendConfig()

    with create_test_client(
        [set_session, get_session],
        logging_config=LoggingConfig(),
        middleware=[logging_middleware_config.middleware, session_config.middleware],
    ) as client:
        response = client.post("/")
        assert len(client.cookies.get("session", "")) == 64

        response = client.get("/")
        assert len(client.cookies.get("session", "")) == 64
```

The test results in the following exception:
```
>           assert len(client.cookies.get("session", "")) == 64
E           AssertionError: assert 5 == 64
E            +  where 5 = len('*****')
E            +    where '*****' = <bound method Cookies.get of <Cookies[<Cookie session=***** for testserver.local />]>>('session', '')
E            +      where <bound method Cookies.get of <Cookies[<Cookie session=***** for testserver.local />]>> = <Cookies[<Cookie session=***** for testserver.local />]>.get
E            +        where <Cookies[<Cookie session=***** for testserver.local />]> = <starlite.testing.client.sync_client.TestClient object at 0x7f4cbf7bea40>.cookies
```

**Additional Context**
Starlite version: 1.51.4

smithk86:
# PR Checklist

- [X] Have you followed the guidelines in `CONTRIBUTING.rst`?

It looks like this issue can be resolved by simply passing a copy of `connection.cookies` to the `obfuscate()` function in [starlite/utils/extractors.py](https://github.com/starlite-api/starlite/blob/e21052705875d0657cbe3aeeeadc1a074e7b354e/starlite/utils/extractors.py#L218). The addition of `copy()` allows the test I created to pass, and it fixes the issue in the application in which the bug was original observed.

### DIFF
--- a/tests/middleware/test_logging_middleware.py
+++ b/tests/middleware/test_logging_middleware.py
@@ -6,8 +6,10 @@ from structlog.testing import capture_logs
 
 from starlite import Cookie, LoggingConfig, Response, StructLoggingConfig, get, post
 from starlite.config.compression import CompressionConfig
+from starlite.connection import Request
 from starlite.middleware import LoggingMiddlewareConfig
-from starlite.status_codes import HTTP_200_OK
+from starlite.middleware.session.memory_backend import MemoryBackendConfig
+from starlite.status_codes import HTTP_200_OK, HTTP_201_CREATED
 from starlite.testing import create_test_client
 
 if TYPE_CHECKING:
@@ -190,3 +192,34 @@ def test_logging_messages_are_not_doubled(
         response = client.get("/")
         assert response.status_code == HTTP_200_OK
         assert len(caplog.messages) == 2
+
+
+def test_logging_middleware_with_session_middleware() -> None:
+    # https://github.com/starlite-api/starlite/issues/1228
+
+    @post("/")
+    async def set_session(request: Request) -> None:
+        request.set_session({"hello": "world"})
+
+    @get("/")
+    async def get_session() -> None:
+        pass
+
+    logging_middleware_config = LoggingMiddlewareConfig()
+    session_config = MemoryBackendConfig()
+
+    with create_test_client(
+        [set_session, get_session],
+        logging_config=LoggingConfig(),
+        middleware=[logging_middleware_config.middleware, session_config.middleware],
+    ) as client:
+        response = client.post("/")
+        assert response.status_code == HTTP_201_CREATED
+        assert "session" in client.cookies
+        assert client.cookies["session"] != "*****"
+        session_id = client.cookies["session"]
+
+        response = client.get("/")
+        assert response.status_code == HTTP_200_OK
+        assert "session" in client.cookies
+        assert client.cookies["session"] == session_id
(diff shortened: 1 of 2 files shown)


provinzkraut (changes requested):
Thanks for the report & fix!

I think that the proper way to tackle this though would be to change `obfuscate` in such a way that it does not mutate the dictionary in-place, but rather returns a new dictionary with the obfuscated values. This is also what the signature suggest it does.


```python
def obfuscate(values: Dict[str, Any], fields_to_obfuscate: Set[str]) -> Dict[str, Any]:
    return {
        key: "*****" if key.lower() in fields_to_obfuscate else value
        for key, value in values.items()
    }
```

smithk86:
@provinzkraut Your suggestion has been implemented. Thanks!

tests/middleware/test_logging_middleware.py
```diff
@@ -190,3 +192,33 @@ async def hello_world_handler() -> Dict[str, str]:
...
+
+    with create_test_client(
+        [set_session, get_session],
+        logging_config=LoggingConfig(),
+        middleware=[logging_middleware_config.middleware, session_config.middleware],
+    ) as client:
+        response = client.post("/")
+        assert response.status_code == HTTP_201_CREATED
+        assert "session" in client.cookies
+        assert len(client.cookies["session"]) == 64
```
provinzkraut, line 219: I think the better solution here would be to check against a constant, known value. The length based checking only works incidentally.

provinzkraut (changes requested):
Sorry that the review is in two batches, I was distracted and submitted the first one too early

tests/middleware/test_logging_middleware.py
smithk86: I originally was just going to check for `client.cookies["session"] != "*****"` but I figured that had some room for error as well.

What if I did `client.cookies["session"] != "*****"` but also confirmed the session id did not change between the two requests:

```python
response = client.post("/")
assert "session" in client.cookies
assert client.cookies["session"] != "*****"
session_id = client.cookies["session"]

response = client.get("/")
assert "session" in client.cookies
assert client.cookies["session"] == session_id
```

provinzkraut: That's a reasonable approach

smithk86: change has been pushed

smithk86:
@provinzkraut It looks like this bug is in the `main` codebase as well. Would you like a pull request for that as well?

provinzkraut:
> It looks like this bug is in the main codebase as well. Would you like a pull request for that as well?

Would be great!

provinzkraut:
@all-contributors add @smithk86 code

PR #1230, opened 2023-02-21
Infra: Run limited tests on Windows and MacOS

provinzkraut:
Change our CI so that in addition to running tests on Ubuntu, it will also run Python 3.11 tests on Windows and MacOS.

# PR Checklist


.github/workflows/ci.yaml
```diff
@@ -24,11 +24,22 @@ jobs:
...
         run: pre-commit run --show-diff-on-failure --color=always --all-files
   test:
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.11"]
+        include:
```
provinzkraut, line 32: I don't know if this is the most elegant way to do this?

Goldziher: I'd actually keep the stuff as I'd, and add another new workflow that runs only on push to main - running tests on these platforms.

provinzkraut: I'd like to keep it all in one workflow though because it's a bit easier to keep track of. How about including the matrix configuration for the OSes conditionally (so only on push to `main`)?

provinzkraut: > I would keep the matrix as is, and add another job just for windows and iOS.

Do we **really** need to test on iOS? Seems like an edge case :wink: 

<hr>

Separate jobs sound good. Maybe just make the testing part a reusable workflow and the call that from the base + windows/macos jobs? That should be a fairly clean solution and could get rid of some of the specific conditionals within the config.

Goldziher: No need for iOS prob. I develop on it

provinzkraut: So, after putting the tests in a reusable workflow, the workflows and jobs themselves are indeed easier, but it looks horrible in the UI, because it adds one unnecessary layer there. ("Tests And Linting / test (3.10) / test" instead of "Tests And Linting / test (3.10)"). 

Not sure if any way is better. Only way to avoid this while still having all tests contained within one workflow and only run the MacOS + Windows tests on a push to main would be to simply copy+paste the testing step and add a bunch of `if`s and conditions.

provinzkraut: @Goldziher What's the reasoning for not running the tests for MacOS and Windows on PRs? That would be the easiest / cleanest solution. It would also avoid having to add fixes post merge because something broke on those platforms that our CI didn't catch because it did not run.

PR #1232, opened 2023-02-21
Docs: Update versioning and release documentation

provinzkraut:
# PR Checklist

### DIFF
--- a/docs/about/starlite-releases.rst
+++ b/docs/about/starlite-releases.rst
@@ -74,20 +74,12 @@ Current version
 
 Maintenance versions
     When a new major version is released, the last *minor* version before it enters
-    maintenance mode. It will receive bugfixes and other critical patches at least until
-    the next major version is released.
-
-LTS versions
-    The *last minor* release of every other *major version* will be an LTS release,
-    guaranteed to be supported for the duration of the next 2 major versions
+    maintenance mode. It will receive bugfixes and other critical patches during the
+    next two release cycles
 
 
 In practice this means that, at any given time, there may be up to 3 currently supported
-releases:
-
-- The current major version
-- The major version preceding the current one
-- The last LTS release
+releases: The current version and the two major versions preceding it.
 
 
 Deprecation policy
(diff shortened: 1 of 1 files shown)


docs/about/starlite-releases.rst
```diff
@@ -38,12 +64,30 @@ won't be sweeping changes that require a complete rewrite of an application or m
...
+    updates in minor releases (see `Version numbering`_)
+
+Maintenance versions
+    When a new major version is released, the last *minor* version before it enters
+    maintenance mode. It will receive bugfixes and other critical patches at least until
+    the next major version is released.
+
+LTS versions
+    The *last minor* release of every other *major version* will be an LTS release,
+    guaranteed to be supported for the duration of the next 2 major versions
```
Goldziher, line 82: I'm not sure this is wise for us for now. I think maintenance as you called it is fine. 

Otherwise, we can say we will continue to support the last minor version for two major versions. 

So, `1.51.*` will be maintained until the release of `4.0.0`.

Goldziher (approved):
Looks good

docs/about/starlite-releases.rst
peterschutt: Agree, Current and Maintenance sufficient for now.
 
We should guarantee a healthy minimum period for version maintenance, so that the number of versions we maintain will be a function of cadence of major releases we make.

So lets say we go from 2.0 -> 3.0 fairly quickly, but then it takes us 2 years to go from 3.0 -> 4.0, I don't think we should still have to maintain 1.51 in 2025 due to a rigid rule about # of major versions.

JacobCoffee (approved):
Aside from what Peter and Goldziher mentioned, it is easy to read and understand

docs/about/starlite-releases.rst
provinzkraut: > I don't think we should still have to maintain 1.51 in 2025 due to a rigid rule about # of major versions.

This would be addressed by the (semi) fixed release schedule for major releases.

peterschutt (approved):

PR #1243, opened 2023-02-26
Signature refactors

peterschutt:
# PR Checklist

### DIFF
--- a/starlite/signature/models.py
+++ b/starlite/signature/models.py
@@ -179,43 +179,6 @@ class SignatureField:
             extra=extra or {},
         )
 
-    @classmethod
-    def from_model_field(cls, model_field: ModelField) -> SignatureField:
-        """Create a SignatureField instance from a pydantic ModelField.
-
-        Args:
-            model_field: A pydantic ModelField instance.
-
-        Returns:
-            A SignatureField
-        """
-        children = (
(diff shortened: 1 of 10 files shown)


starlite/signature/models.py
```diff
@@ -331,11 +285,48 @@ def to_dict(self) -> dict[str, Any]:
             return {key: self._resolve_field_value(key) for key in self.__fields__}
         return {key: self.__getattribute__(key) for key in self.__fields__}  # pylint: disable=unnecessary-dunder-call
 
+    @classmethod
+    def signature_field_from_model_field(cls, model_field: ModelField) -> SignatureField:
```
Goldziher, line 289: this change is a blocker - `ModelField` is a pydantic specific value, and the point of this method was to encapsulate the pydantic specific code on the PydanticSignatureModel, and thus isolate it. 

Please revert this change. 

Other changes are good.

peterschutt: I don't understand. This logic was on `SignatureParameter` before this change. This is moving the logic on to `PydanticSignatureModel` in order to isolate it.

Goldziher: than im an idiot.

Goldziher (approved):

PR #1252, opened 2023-03-01
Move sqlalchemy 1 into contrib

peterschutt:
This PR is first step toward refactoring our sqlalchemy interop, and interop with other 3rd party tools for starlite 2.0.

It creates a contrib package for supported 3rd party tools (empty at this point), but an important pattern to agree on before we start building out functionality for each. Each package will include support for using that 3rd party tool with starlite, for example  "on-app-init" plugins for app configuration, framework specific DTO factory, cli plugins, repository implementations, and so on.

This PR also moves the current sqlalchemy plugin into `starlite.contrib.sqlalchemy_1`, and moves tests for that into `tests.contrib.sqlalchemy_1`. Those tests will only be collected by pytest if we have sqlalchemy 1.x in the environment when running tests. If sqlalchemy 2.x is installed, they will not be collected. This would allow us to include a run of tests in CI that has sqlalchemy 1.x installed.

For now, I have removed any other tests that relied upon sqlalchemy 1.x models from our main test suite.

# PR Checklist


pyproject.toml
```diff
@@ -204,7 +207,7 @@ no-docstring-rgx = "(__.*__|main|test.*|.*test|.*Test|^_.*)$"
 logging-modules = ["logging", "picologging", "structlog"]
 
 [tool.coverage.run]
-omit = ["*/tests/*"]
+omit = ["*/tests/*", "starlite/contrib/sqlalchemy_1/*"]
```
Goldziher, line 210: well, if we support it we should test it - and have coverage, no?

tests/contrib/sqlalchemy_1/conftest.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+try:
+    import sqlalchemy
+except ImportError:
+    collect_ignore_glob = ["*"]
+else:
+    if sqlalchemy.__version__.startswith("2."):
```
Goldziher, line 8: you can put this in the try block

Goldziher (approved):

pyproject.toml
peterschutt: Sure we can test this stuff, but we can't test both sqlalchemy 2.x and sqlalchemy 1.x at the same time. So, we either use a tool like tox to run our tests with multiple environment configurations, or we add another job to CI that specifically tests against sqlalchemy 1.x. We'd then use `coverage combine` to combine the coverage reports from multiple runs before sending through to sonar. We should also probably run coverage against 3.8 as well and combine it also.

I'll create a separate issue for this, so we can address after the pattern has settled down.

PR #1254, opened 2023-03-01
Repository contrib

peterschutt:
# PR Checklist


peterschutt:
@cofin - this one moves the repository pattern into contrib with minimal changes from starlite-saqlalchemy.

Prob need to remove all of the postgres specific stuff, but feel free to pile your updates on top of this PR.

starlite/contrib/repository/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Abstraction over the data storage for the application."""
+from __future__ import annotations
+
+from . import abc, filters, types
+
+__all__ = [
```
Goldziher, line 6: Use tuple

starlite/contrib/repository/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Abstraction over the data storage for the application."""
+from __future__ import annotations
+
+from . import abc, filters, types
```
Goldziher, line 4: Why not export explicit classes or vars? E.g. `AbstractRepository`

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,167 @@
...
+
+from abc import ABCMeta, abstractmethod
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
+
+from .exceptions import NotFoundError
+
+if TYPE_CHECKING:
+    from .types import FilterTypes
+
+__all__ = ["AbstractRepository"]
```
Goldziher, line 12: Let's stick with tuples for this

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Args:
+            data: Instance to be added to the collection.
+
+        Returns:
+            The added instance.
+        """
+
+    @abstractmethod
+    async def delete(self, id_: Any) -> T:
```
Goldziher, line 42: Instead of `id_` why not `identifier` or `item_id`?

I think forward underscore looks bad and documents bad

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,167 @@
...
+        Args:
+            item: Anything that should have an attribute named as `self.id_attribute` value.
+
+        Returns:
+            The value of attribute on `item` named as `self.id_attribute`.
+        """
+        return getattr(item, cls.id_attribute)
+
+    @classmethod
+    def set_id_attribute_value(cls, id_: Any, item: T) -> Any:
```
Goldziher, line 156: Shoukdnt T be returned?

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,192 @@
...
+    """A repository implementation for tests.
+
+    Uses a `dict` for storage.
+    """
+
+    collection: MutableMapping[Hashable, ModelT]
+    model_type: type[ModelT]
+
+    def __init__(self, id_factory: Callable[[], Any] = uuid4, **_: Any) -> None:
+        super().__init__()
```
Goldziher, line 35: Might as well propagate kwargs i think

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,270 @@
...
+    except IntegrityError as exc:
+        raise ConflictError from exc
+    except SQLAlchemyError as exc:
+        raise RepositoryError(f"An exception occurred: {exc}") from exc
+
+
+class SQLAlchemyRepository(AbstractRepository[ModelT], Generic[ModelT]):
+    """SQLAlchemy based implementation of the repository interface."""
+
+    def __init__(self, *, session: AsyncSession, select_: Select[tuple[ModelT]] | None = None, **kwargs: Any) -> None:
```
Goldziher, line 63: Same comments everywhere regarding the forward underscore

Goldziher, line 63: Shouldn't this be `tuple[ModelT, ...]?

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,270 @@
...
+        """Get a list of instances, optionally filtered.
+
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The list of instances, after filtering applied.
+        """
+        for filter_ in filters:
```
Goldziher, line 138: You really love that forward underscore 😅

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,270 @@
...
+    async def check_health(cls, session: AsyncSession) -> bool:
+        """Perform a health check on the database.
+
+        Args:
+            session: through which we runa check statement
+
+        Returns:
+            `True` if healthy.
+        """
+        return (  # type:ignore[no-any-return]  # pragma: no cover
```
Goldziher, line 224: Why is this complaining in type checking?

Goldziher (approved):
Looks good. I hate those forward underscores

starlite/contrib/repository/__init__.py
peterschutt: Yeh, I like namespaces. Not that I mind if we move some names up to the top level, I just don't tend to by default.

starlite/contrib/repository/__init__.py
peterschutt: Cleaned up the rest of them also.

starlite/contrib/repository/abc.py
peterschutt: absolutely should

starlite/contrib/repository/testing/generic_mock_repository.py
peterschutt: Well the idea here is to swallow the kwargs. 

If an instance of this mock repository gets monkey-patched over the top of a SQLAlchemy repository (to test business logic in a service object, for example), the code that calls it will pass a session to the constructor. If that were passed up to `AbstractRepository.__init__()`, which calls `super().__init__(**kwargs)` itself, it will throw a `TypeError`.

I think it is better for the mock repo to swallow unexpected kwargs than to move that logic into the base class, where unexpected kwargs would then also get swallowed at runtime.

Happy to discuss any of this ofc.

starlite/contrib/sqlalchemy/repository.py
peterschutt: Yes, nice catch: https://github.com/sqlalchemy/sqlalchemy/blob/9d9e47fe9398b6bcda917f63441d2b5ad4b345c4/lib/sqlalchemy/sql/_typing.py#L155

starlite/contrib/sqlalchemy/repository.py
peterschutt: You might need to write a flake8 plugin :stuck_out_tongue_winking_eye:

starlite/contrib/sqlalchemy/repository.py
peterschutt: > Returning Any from function declared to return "bool"

I think with the textual select, `scalar_one()` is typed to return `Any`.

starlite/contrib/sqlalchemy/repository.py
peterschutt: Actually, no, it is `tuple[ModelT]`, that type is the type of row returned from the select query. When making an orm select, the query returns rows like `(Model,)`, and if a query returns more than one row the response is `[(Model,), (Model,)]`.
This is what the `scalar()`, `scalar_one()`, `scalar_first()` accessor methods are for, to return the first element of the row.

Goldziher: Ok... Weird

Goldziher (approved):

cofin:
I'm still not able to figure out why these totally unrelated test cases are suddenly failing here.  However, I think the repo code is ready to be reviewed.

JacobCoffee (approved):

ruff.toml
```diff
@@ -0,0 +1,118 @@
+
```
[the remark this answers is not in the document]
cofin: Yeah, I had it in the `pyproject.toml`.  I thought I'd try to prevent myself from accidentally committing it with this PR, so I moved it to a separate file. ...And then I proceeded to commit that new file to the PR.  :)

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+        Returns:
+            The retrieved instance.
+
+        Raises:
+            RepositoryNotFoundException: If no instance found identified by `item_id`.
+        """
+
+    @abstractmethod
+    async def get_one(self, **kwargs: Any) -> T:
+        """Get an instance if it exists or None.
```
peterschutt, line 98: This doc string is not correct.

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @abstractmethod
+    async def get_one_or_none(self, **kwargs: Any) -> T | None:
+        """Get an instance if it exists or None.
+
+        Args:
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The list of instances, after filtering applied.
```
peterschutt, line 115: again

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+    @abstractmethod
+    async def list_and_count(self, *filters: FilterTypes, **kwargs: Any) -> tuple[Sequence[T], int]:
+        """List records with total count.
+
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            Count of records returned by query, ignoring pagination.
```
peterschutt, line 139: and this

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Args:
+            data: Instance to be added to the collection.
+
+        Returns:
+            The added instance.
+        """
+
+    @abstractmethod
+    async def add_many(self, data: Sequence[T]) -> Sequence[T]:
```
peterschutt, line 46: does this actually need to be a sequence, or could it be any iterable, e.g., a set?

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+
+    @abstractmethod
+    async def get_one(self, **kwargs: Any) -> T:
+        """Get an instance if it exists or None.
+
+        Args:
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The list of instances, after filtering applied.
```
peterschutt, line 104: and this

starlite/contrib/repository/abc.py
```diff
@@ -0,0 +1,244 @@
...
+        """
```
peterschutt, line 105: this should include a "Raises" section

starlite/contrib/repository/abc.py
peterschutt: or the return type is not correct.. actually, this method's place is unclear to me.. is this meant to be more like `first()` or smth?

starlite/contrib/sqlalchemy/base.py
```diff
@@ -0,0 +1,112 @@
...
+    "uq": "uq_%(table_name)s_%(column_0_name)s",
+    "ck": "ck_%(table_name)s_%(constraint_name)s",
+    "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
+    "pk": "pk_%(table_name)s",
+}
+"""Templates for automated constraint name generation."""
+
+
+@listens_for(Session, "before_flush")
+def touch_updated_timestamp(session: Session, *_: Any) -> None:
```
peterschutt, line 34: I think we should find a way to only attach this listener only if the audit base is used, or make it a thing that is configurable by the end user.

peterschutt: also, on some backends this can be done using server side default, e.g., mysql/mariadb.

starlite/contrib/sqlalchemy/base.py
```diff
@@ -0,0 +1,112 @@
...
+
+    __abstract__ = True
+    __name__: str
+    __table__: Any
+
+    # noinspection PyMethodParameters
+    @declared_attr.directive
+    def __tablename__(cls) -> str:  # pylint: disable=no-self-argument
+        """Infer table name from class name."""
+        return cls.__name__.lower()
```
peterschutt, line 84: should we do CamelCase to snake_case for these do you think?

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,441 @@
...
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            Count of records returned by query, ignoring pagination.
+        """
+        select = self.select.add_columns(
+            over(
+                sql_func.count(  # pylint: disable=not-callable
```
peterschutt, line 251: is this type of thing well supported across different backends?

peterschutt (commented):
Looks good @cofin!

A few issues with docstrings, and I've raised a few discussion points.

We should focus on getting a test structure going for this where we can do integration tests across different servers. What do you think of that `pytest-docker` pattern from `starlite-saqlalchemy`?

starlite/contrib/repository/abc.py
provinzkraut: I think `Iterable` should be fine IMO

starlite/contrib/sqlalchemy/base.py
provinzkraut: Would certainly be useful, as that's the common naming scheme promoted by SQLA as well.

starlite/contrib/sqlalchemy/repository.py
provinzkraut: Supported by SQLite, MySQL/Maria, Oracle, MS and Postgres. Other I don't know, but that's good enough I think.

starlite/contrib/sqlalchemy/base.py
```diff
@@ -0,0 +1,112 @@
...
+}
+"""Templates for automated constraint name generation."""
+
+
+@listens_for(Session, "before_flush")
+def touch_updated_timestamp(session: Session, *_: Any) -> None:
+    """Set timestamp on update.
+
+    Called from SQLAlchemy's
+    [`before_flush`][sqlalchemy.orm.SessionEvents.before_flush] event to bump the `updated`
```
provinzkraut, line 38:
```suggestion
    :meth:`before_flush <sqlalchemy.orm.SessionEvents.before_flush>` event to bump the ``updated``
```

starlite/contrib/sqlalchemy/base.py
```diff
@@ -0,0 +1,112 @@
...
+    timestamp on modified instances.
+
+    Args:
+        session: The sync [`Session`][sqlalchemy.orm.Session] instance that underlies the async
```
provinzkraut, line 42:
```suggestion
        session: The sync :class:`Session <sqlalchemy.orm.Session>` instance that underlies the async
```

provinzkraut (commented):
Some comments on the docstrings

starlite/contrib/sqlalchemy/repository.py
cofin: Yeah, analytic window functions are standard now for most engines.  This is actually the reason why I left it like this instead of breaking it into 2 queries.  However, if we run into issues, we can add the fallback to use 2 queries for this.

starlite/contrib/sqlalchemy/base.py
cofin: Yes, I agree.  Snake case is the way to go for table names. I'll update this accoridngly.

starlite/contrib/sqlalchemy/base.py
cofin: I didn't do this because it's not as easy to detect if the session is dirty.  A `server_onupdate` fires any time an update happens, even if there is no actual modification.  However, i do like the idea of server generated defaults for timestamps.

starlite/contrib/repository/abc.py
cofin: can you guys think of a way we can preserve the typing and combine add_many and add into a single endpoint?  There's no reason why we can't just use `add_many` for all inserts.

starlite/contrib/repository/abc.py
cofin: `get` requires an `id` field lookup.  `get_one` allows you to pass in an arbitrary set of  kwargs to use for the where clause.  It return one or an exception.

`get_one` and `get` could be combined.

starlite/contrib/repository/abc.py
peterschutt: I like the explicitness of separate methods.. you could use `@overload` though..

provinzkraut: Agree with @peterschutt on this. Two separate methods are better than one that does both.

cofin: Fine with me.  Thanks for the quick feedback

cofin:
> We should focus on getting a test structure going for this where we can do integration tests across different servers. What do you think of that `pytest-docker` pattern from `starlite-saqlalchemy`?

I think this is probably the best path we have right now.  Should we bring this piece in with the Service layer or should it happen now?

starlite/contrib/repository/abc.py
cofin: So, to answer this.  The docstring is incorrect here.  It should return the instance found with the kwarg filter or raise an exception.

starlite/contrib/sqlalchemy/base.py
peterschutt: Oh cool, well that is a plus for doing it client side!

peterschutt:
> > We should focus on getting a test structure going for this where we can do integration tests across different servers. What do you think of that `pytest-docker` pattern from `starlite-saqlalchemy`?
> 
> I think this is probably the best path we have right now. Should we bring this piece in with the Service layer or should it happen now?

With this I think. The service layer can be easily tested against the mock repository.

We should probably set them up separate from the main test suite, and either run them selectively, or in parallel so the overall testing time doesn't take a hit.

starlite/contrib/repository/abc.py
peterschutt: ok got ya

starlite/contrib/repository/__init__.py
Goldziher: @cofin and @peterschutt - please change this - there is no point in importing the namespaces and reexporting them from the init file. If the user wants to import the namespace, he can easily do this `from starlite.contrib.repository import abc` without actually needing anything in the init file for this to work.

Either import content that should be exported from here, like `AbstractRepository` etc. or keep this file empty.

Goldziher (commented):
this is all very cool. can you create a followup ticket to document this? or add docs in this scope.

starlite/contrib/repository/abc.py
cofin: I could not make this work with an `Iterable`.  There are other places where we need an indexable iterator.  I've moved it back to a `Sequence` for now.

cofin: I actually updated it to a `list` since that's what is returned.  Any issues with that type?

cofin:
I think i've addressed or made comments on all of the feedback at this point.  I've left the ones open where I could use some feedback on.

I think the one last piece of this would be the docker based pytest from @peterschutt comment above.

starlite/contrib/repository/abc.py
peterschutt: > There are other places where we need an indexable iterator.

Such as?

We should always prefer abstract function inputs IMO, it allows for much nicer DX. E.g., `Iterable[ModelT]` would allow for things like:

```python
repo.add_many(Model(id=i) for i in ids)
```
or

```python
models = {i: Model(id=i) for id in ids}
repo.add_many(models.values())
```

cofin: For instance, when I swap from list to `Iterable`, one of the method that begins complaining is `count`
```python
    async def count(self, *filters: FilterTypes, **kwargs: Any) -> int:
        return len(await self.list(*filters, **kwargs))
```
```Argument 1 to "len" has incompatible type "Iterable[ModelT]"; expected "Sized"  [arg-type]mypy(error)```

and 

```python
    async def get_one_or_none(self, **kwargs: Any) -> ModelT | None:
        data = await self.list(**kwargs)
        return data[0] if len(data) > 0 else None

```
gives
```Value of type "Iterable[ModelT]" is not indexable  [index]mypy(error)```

Comment (cofin) [reply starlite/contrib/repository/abc.py]: This is in `GenericMockRepository`

**Comment by cofin** [comment]
> this is all very cool. can you create a followup ticket to document this? or add docs in this scope.

I've created a follow-up ticket for the documentation to this.

**Review by Goldziher — APPROVED** [review]

File: starlite/contrib/sqlalchemy/repository.py [hunk 22]
```diff
@@ -103,17 +111,29 @@ async def add_many(self, data: list[ModelT]) -> list[ModelT]:
...
+            for datum in data_to_insert:
+                if datum.get("id", None) is None:
+                    datum.update({"id": uuid4()})
+            await self.session.execute(
+                insert(self.model_type),
+                data_to_insert,
+            )
+            # select the records we just updated.
+            return await self.list(
+                CollectionFilter(field_name="id", values=[datum.get("id") for datum in data_to_insert])
```
provinzkraut, line 135: I think it's better to store the IDs within the loop above, so you don't have to loop again here.

tests/contrib/sqlalchemy/repository/test_sqlalchemy.py
```diff
@@ -397,3 +407,151 @@ def test_filter_collection_by_kwargs_raises_repository_exception_for_attribute_e
...
+        f"sqlite+aiosqlite:///{db_path}/test.db",
+        echo=True,
+        poolclass=NullPool,
+    )
+    try:
+        yield engine
+    finally:
+        await engine.dispose()
+        Path(f"{db_path}/test.db").unlink()
+        Path(db_path).rmdir()
```
provinzkraut, line 433:
```suggestion
async def get_sqlite_engine(tmp_path: Path) -> AsyncGenerator[AsyncEngine, None]:
    """Postgresql instance for end-to-end testing.

    Args:
        docker_ip: IP address for TCP connection to Docker containers.

    Returns:
        Async SQLAlchemy engine instance.
    """
    engine = create_async_engine(
        f"sqlite+aiosqlite:///{db_path}/test.db",
        echo=True,
        poolclass=NullPool,
    )
    try:
        yield engine
    finally:
        await engine.dispose()
```

cofin:
> > > We should focus on getting a test structure going for this where we can do integration tests across different servers. What do you think of that `pytest-docker` pattern from `starlite-saqlalchemy`?
> > 
> > 
> > I think this is probably the best path we have right now. Should we bring this piece in with the Service layer or should it happen now?
> 
> With this I think. The service layer can be easily tested against the mock repository.
> 
> We should probably set them up separate from the main test suite, and either run them selectively, or in parallel so the overall testing time doesn't take a hit.

I think this may be the only thing remaining on this PR.  I've added in sqlite testing so that it uses a temp folder to set up and dispose of the DB automatically.

tests/contrib/sqlalchemy/models.py
```diff
@@ -15,7 +15,7 @@ class Author(AuditBase):
     """The Author domain object."""
 
     name: Mapped[str]
-    dob: Mapped[date | None]
+    dob: Mapped[date] = mapped_column(nullable=True)
```
peterschutt, line 18: What's the issue with this one? The annotated type no longer matches the column type, right?

cofin: That's correct.  I couldn't get it to accept `Mapped[date|None]` in all Python version.  It also complained about `Mapped[Optional[date]]`, so I opted to use the override in `mapped_column`

cofin:
With this last edit, I've got everything I can think of included for this release. I think this is ready for one final review!

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+                        insert(self.model_type).returning(self.model_type),
+                        data_to_insert,  # pyright: reportGeneralTypeIssues=false
+                    )
+                )
+                for instance in instances:
+                    self.session.expunge(instance)
+                return instances
+            # when bulk insert with returning isn't supported, we ensure there is a unique ID on each record so that we can refresh the records after insert.
+            # ensure we have a UUID ID on each record so that we can refresh the data before returning
+            new_primary_keys: list[UUID] = []
```
peterschutt, line 126: `pragma: no cover` this section if we cannot hit it in unittests with sqlite - we'll have to make sure we get it with integration tests.

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+from starlite.contrib.repository.exceptions import ConflictError, RepositoryError
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Hashable, Iterable, MutableMapping
+    from typing import Any
+
+    from starlite.contrib.repository import FilterTypes
+    from starlite.contrib.sqlalchemy import base
+
+ModelT = TypeVar("ModelT", bound="base.Base | base.AuditBase")
```
peterschutt, line 21: I don't think this should be bound to sqlalchemy models.

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+        if hasattr(data, "updated") and hasattr(data, "created"):
+            # maybe the @declarative_mixin decorator doesn't play nice with pyright?
+            data.updated = data.created = now  # pyright: ignore
+        if allow_id is False:
+            id_ = self._id_factory()
+            self.set_id_attribute_value(id_, data)
+        self.collection[data.id] = data
+        return data
+
+    async def add_many(self, data: list[ModelT], allow_id: bool = False) -> list[ModelT]:
```
peterschutt, line 74:
```suggestion
    async def add_many(self, data: Iterable[ModelT], allow_id: bool = False) -> list[ModelT]:
```

peterschutt, line 74: I know this is not your design decision, but `allow_id` shouldn't be in the signature here I don't think as this should be called by production code where the param cannot be set. Maybe we should make it an arg to `__init__()`

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+
+        Returns:
+            The added instance.
+        """
+        now = datetime.now()  # noqa: DTZ005
+        for data_row in data:
+            if allow_id is False and self.get_id_attribute_value(data_row) is not None:
+                raise ConflictError("`add()` received identified item.")
+
+            if hasattr(data, "updated") and hasattr(data, "created"):
```
peterschutt, line 89:
```suggestion
            if hasattr(data_row, "updated") and hasattr(data_row, "created"):
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+                # maybe the @declarative_mixin decorator doesn't play nice with pyright?
+                data.updated = data.created = now  # pyright: ignore
```
peterschutt, line 91:
```suggestion
                data_row.updated = data_row.created = now  # pyright: ignore
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+            **kwargs: Identifier of the instance to be retrieved.
+
+        Returns:
+            a tuple that includes the instance and whether or not it needed to be created.
+
+        """
+        existing = await self.get_one_or_none(**kwargs)
+        if existing:
+            return (existing, False)
+        return (await self.add(self.model_type(**kwargs)), True)  # type: ignore[arg-type]
```
peterschutt, line 174:
```suggestion
            return existing, False
        return await self.add(self.model_type(**kwargs)), True  # type: ignore[arg-type]
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The retrieved instance or None
+
+        Raises:
+            RepositoryNotFoundException: If no instance found identified by ``kwargs``.
+        """
+        data = await self.list(**kwargs)
+        return self.check_not_found(data[0] if len(data) > 0 else None)
```
peterschutt, line 189:
```suggestion
        return self.check_not_found(data[0] if data else None)
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+        """Get instance identified by query filters or None if not found.
+
+        Args:
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The retrieved instance or None
+        """
+        data = await self.list(**kwargs)
+        return data[0] if len(data) > 0 else None
```
peterschutt, line 201:
```suggestion
        return data[0] if data else None
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+                collection.
+
+        Returns:
+            The updated instance.
+
+        Raises:
+            RepositoryNotFoundException: If no instance found with same identifier as ``data``.
+        """
+        item = self._find_or_raise_not_found(self.get_id_attribute_value(data))
+        # should never be modifiable
```
peterschutt, line 229:
```suggestion
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+                collection.
+
+        Returns:
+            The updated instances.
+
+        Raises:
+            RepositoryNotFoundException: If no instance found with same identifier as ``data``.
+        """
+        items = [self._find_or_raise_not_found(self.get_id_attribute_value(row)) for row in data]
+        # should never be modifiable
```
peterschutt, line 253:
```suggestion
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+        **kwargs: Any,
+    ) -> tuple[list[ModelT], int]:
+        """Get a list of instances, optionally filtered with a total row count.
+
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            Count of records returned by query, ignoring pagination.
```
peterschutt, line 298:
```suggestion
            List of instances, and count of records returned by query, ignoring pagination.
```

starlite/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -0,0 +1,346 @@
...
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            The list of instances, after filtering applied.
+        """
+        return list(self.filter_collection_by_kwargs(self.collection, **kwargs).values())
+
+    def filter_collection_by_kwargs(  # type:ignore[override]
```
peterschutt, line 314: wonder why the override ignore is necessary here, the signature looks right to me compared to the abc... :shrug:

peterschutt (commented):
Just focused on the mock repo here.. I'll push up some fixes

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+            data: list of Instances to be added to the collection.
+
+        This function has an optimized bulk insert based on the configured SQL dialect:
+        - For backends supporting `RETURNING` with `executemany`, a single bulk insert with returning clause is executed.
+        - For other backends, it does a bulk insert and then selects the inserted records
+
+        Returns:
+            The added instances.
+        """
+        data_to_insert: list[dict[str, Any]] = [v.to_dict() if isinstance(v, self.model_type) else v for v in data]  # type: ignore
```
peterschutt, line 112: can we not just pass the list of declarative models straight through to the bulk insert? I.e., do we need this dict conversion?

Also, why the `isinstance` check? The method signature says that the data is a list of `ModelT`, so anything else would be a type error, right?

Also, we can probably relax that data type back to `Collection` or `Iterable` or smth.

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+
+        Returns:
+            The retrieved instance or None
+        """
+        with wrap_sqlalchemy_exception():
+            statement = self._filter_select_by_kwargs(statement=self.statement, **kwargs)
+            instance = (await self._execute(statement)).scalar_one_or_none()
+            if instance:
+                self.session.expunge(instance)
+            return instance or None
```
peterschutt, line 256: Isn't instance already `ModelT | None` as the return value from `scalar_one_or_none()`, making the `or None` redundant?

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+        Args:
+            **kwargs: Identifier of the instance to be retrieved.
+
+        Returns:
+            a tuple that includes the instance and whether or not it needed to be created.
+        """
+        existing = await self.get_one_or_none(**kwargs)
+        if existing:
+            return (existing, False)
+        return (await self.add(self.model_type(**kwargs)), True)  # type: ignore[arg-type]
```
peterschutt, line 270:
```suggestion
            return existing, False
        return await self.add(self.model_type(**kwargs)), True  # type: ignore[arg-type]
```

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+
+        Args:
+            *filters: Types for specific filtering operations.
+            **kwargs: Instance attribute value filters.
+
+        Returns:
+            Count of records returned by query, ignoring pagination.
+        """
+        statement = self.statement.with_only_columns(
+            sql_func.count(  # pylint: disable=not-callable
```
peterschutt, line 283: do the pylint directives work with ruff? or can this be removed?

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+            data: A list of instances to update.  Each should have a value for `self.id_attribute` that exists in the
+                collection.
+
+        Returns:
+            The updated instances.
+
+        Raises:
+            RepositoryNotFoundException: If no instance found with same identifier as `data`.
+        """
+        data_to_update: list[dict[str, Any]] = [v.to_dict() if isinstance(v, self.model_type) else v for v in data]  # type: ignore
```
peterschutt, line 334: same comments here as for `add_many`

starlite/contrib/sqlalchemy/repository.py
provinzkraut: I removed most of the pylint things in the ruff PR, as lots of false positives don’t occur there. If you still need to suppress something, you can just use the `#noqa: ` notation with the pylint error code. The pylint specific notation isn’t needed anymore.

peterschutt: fwict, moving to ruff has been a massive positive - and to top it all off, my isort issues are gone :heart:

provinzkraut: It's a blessing, really

starlite/contrib/sqlalchemy/repository.py
```diff
@@ -0,0 +1,540 @@
...
+        statement = self._filter_select_by_kwargs(statement, **kwargs)
+        with wrap_sqlalchemy_exception():
+            result = await self._execute(statement)
+            count: int = 0
+            instances: list[ModelT] = []
+            for i, (instance, count_value) in enumerate(result):
+                self.session.expunge(instance)
+                instances.append(instance)
+                if i == 0:
+                    count = count_value
```
peterschutt, line 390:
```suggestion
            first, count = next(result)
            instances = [first]
            for instance, _ in result:
                self.session.expunge(instance)
                instances.append(instance)
```
I think this should work and looks a bit cleaner, haven't tested though  :worried:

starlite/contrib/repository/testing/generic_mock_repository.py
peterschutt: this is weird to me, the super class has an unconstrained type var as the collection arg type and returns same. IMO, this is exactly what we are implementing here.

starlite/contrib/sqlalchemy/repository.py
provinzkraut: This is the way it's done in the SQLAlchemy docs: https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-bulk-insert-statements. 

This though raises the question of the signature of this function. 

To me it would make most sense if either this function operates on raw data and not instances when using bulk inserts like this, or or it operates on instances and just uses the session methods for this (e.g. `add_all`). 

The current way mixes both, which is a bit confusing. If those are newly created instances, they'd also not belong to a session, so I'm not sure if `expunge` is even necessary? 

It also results in:

- Construction of transient instances
- Conversion of those instances into dicts
- Creation of new, non-transient but otherwise identical, instances from `session.scalars`  

which seems like an unnecessary round trip.

peterschutt: > This is the way it's done in the SQLAlchemy docs

Indeed, and having just tested, it definitely doesn't work with instances.

> To me it would make most sense if either this function operates on raw data and not instances when using bulk inserts like this, or or it operates on instances and just uses the session methods for this (e.g. add_all).

The general repository pattern is to deal in ORM objects, and it would be nice to remain consistent with that. However, parsing out of bytes, into some ser/de model, into an orm model, to finally convert to dicts, and back again into orm models is indeed a very round-about means to the end.

If I were very concerned with efficiency (i.e., collection size * number of attributes) I'd probably have a typed dict representation of my data, and go straight from msgspec to that, and that case, I would certainly appreciate the repository having methods to support.

Maybe we should mirror the sqlalchemy api for this stuff then - a bulk method for mappings, and a separate one for orm objects. It looks as though the `add_all` method should still participate in the benefits of the improved 'returning' support in 2.0: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#benchmarks. Our recommendation could be that if performance isn't a major factor, and collection sizes aren't very big, then use the ORM one, and where performance is critical, we direct users to the mappings method, and then suggest not parsing into ORM objects in the first place.

provinzkraut: > Maybe we should mirror the sqlalchemy api for this stuff then - a bulk method for mappings, and a separate one for orm objects.

> Our recommendation could be that if performance isn't a major factor, and collection sizes aren't very big, then use the ORM one, and where performance is critical, we direct users to the mappings method, and then suggest not parsing into ORM objects in the first place.

That sounds like the proper solution to this. Especially since `session.add_all` now uses `RETURNING` internally, performance should be still be quite good.

starlite/contrib/sqlalchemy/repository.py
cofin: Agreed. I put in a `asyncpg` based test into the `custom` branch on `saqlalchemy` as well as these updates to the repo pattern.  We can move those test suites over here when ready.

starlite/contrib/sqlalchemy/repository.py
cofin: I'd also argue that if there is a concern for performance, `sqlite` likely won't be used and the `insert` `returning` will be executed.

cofin: I also would need to verify that `add_all` doesn't increase the overhead too much.  I have used the bulk insert method above to insert thousands of objects at a time, and, in the past, I had issues with pretty much every other sqlalchemy method.  (This was obviously be fore the 2.0 change)

starlite/contrib/sqlalchemy/repository.py
cofin: I'm not sure there's an equivalent to `add_all` for bulk updates.  I'm thinking we leave this method as is and then extend this repository further with the `add_many_mapped` and `update_many_mapped` functions.  Thoughts?

cofin:
This last edit adds in a set of `asyncpg` tests using `pytest_docker`

JacobCoffee (commented):
I believe `NotFoundError` was renamed from `RepositoryNotFoundException`

Fixed by #1321

PR #1255, opened 2023-03-01
Fix CLI `--reload` flag not working as expected

### LINKED ISSUE #1191: Bug: Issue with reload
**Describe the bug**
When I start my app using starlite run -r I get the following error

INFO:     Will watch for changes in these directories: ['/Users/alc/starlite-apps/starlite_test']
ERROR:    [Errno 48] Address already in use

If I stop that and run it without -r it works, so I guess my port is not blocked

**To Reproduce**
Save this as `app.py`

```python

@get('/')
def route() -> str:
    return 'Hello'

app = Starlite([route])
```

**Additional context**
@provinzkraut in discord help suggested a workaround using ` uvicorn <app file>:<application object> --reload` and also linked https://github.com/encode/uvicorn/issues/1045

provinzkraut:
Fixes #1191 by invoking uvicorn with `subprocess.run`. This is the only actual workaround due to https://github.com/encode/uvicorn/issues/1045.

# PR Checklist

### DIFF
--- a/starlite/cli/commands/core.py
+++ b/starlite/cli/commands/core.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import inspect
 import multiprocessing
+import subprocess
+from typing import Any
 
 import click
 from click import command, option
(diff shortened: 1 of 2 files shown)


starlite/cli/commands/core.py
```diff
@@ -52,7 +66,7 @@ def run_command(
     """
 
     try:
-        import uvicorn
+        pass
```
Goldziher, line 69: empty try?

Goldziher (approved):

starlite/cli/commands/core.py
provinzkraut: Well, that can be removed :upside_down_face:

JacobCoffee (commented):
👍🏽

PR #1291, opened 2023-03-06
Fix docs references

provinzkraut:
Attempt to fix the 1000+ broken references in our documentation, caused by various refactors that didn't take the docs into account.

- Enable sphinx nitpicky node to pick up on all broken references
- Add a step to our CI that checks if the docs build correctly
- Add functions to our sphinx extension that can help ignore known false positive warnings when running sphinx in nitpicky mode

# PR Checklist


Goldziher (approved):

JacobCoffee:
😞 we should make it a hard requirement maybe to ensure documentation stays up to date inside of your code change PR

provinzkraut:
> we should make it a hard requirement maybe to ensure documentation stays up to date inside of your code change PR

Failing docs build should fail the CI with this.

provinzkraut:
Okay, about 50% of those stem from an issue with sphinx autodoc. I've been trying to resolve this with some effort and have been able to get the missing references down to ~300.

provinzkraut:
A few findings:

- Sphinx autodoc sometimes doesn't pick up type hints when the types are imported within an `if TYPE_CHECKING`  block.
  those warnings can be automatically silenced
- Sphinx autodoc also (seemingly at random) assumes wrong reference types, which we can fix locally within our sphinx extension by attempting to re-resolve the cross references
- Some errors will have to be ignored manually if we want to run in nitpicky mode, since they're things we intentionally don't want to document (like type variables) or no cross linking target exists (like objects from Pydantic)

PR #1296, opened 2023-03-08
Add configurable exception logging and handle traceback truncation 

Goldziher:
# PR Checklist

This PR does the following:

It adds three new attributes to the `BaseLoggingConfig` class, which allow users to configure if and how to log exceptions: 

a. `log_exceptions` which receives a string literal - `always`, `debug` and `never`. Always means always log exceptions. Debug means log when the Starlite app is in `debug=True`, and never means never log exceptions. 
b. `traceback_line_limit`, which allows users to limit the length of traceback. I gave it a default if 20 lines, which is pretty long. Also, the lines are taken from the END of the traceback, which has the most recent lines. 
c. `exception_logging_handler` - a callable that receives three parameters - the `app.logger`, the connection `scope` and the traceback list, and should handle logging. A default handler is provided. 

Currently the PR is missing documentation, because I am not sure where to add these - or if I even should.


starlite/logging/config.py
```diff
@@ -60,7 +60,14 @@
 }
 
 
-def get_default_handlers() -> dict[str, dict[str, Any]]:
+def get_logger_placeholder(_: str) -> NoReturn:  # pragma: no cover
+    """Raise an `ImproperlyConfiguredException."""
+    raise ImproperlyConfiguredException(
+        "To use 'app.get_logger', 'request.get_logger' or 'socket.get_logger' pass 'logging_config' to the Starlite constructor"
```
provinzkraut, line 66:
```suggestion
        "To use :meth:`Starlite.get_logger <.app.Starlite.get_logger>`, :meth`request.get_logger <.connection.Request.get_logger>` or :class:`socket.get_logger <.connection.WebSocket.get_logger>` pass ``logging_config`` to the Starlite constructor"
```

provinzkraut (changes requested):
Looks like a good solution to me. A few changes required in docstrings

provinzkraut:
> Currently the PR is missing documentation, because I am not sure where to add these - or if I even should.

The thing is, there is currently not really a place where this should go naturally. We don't have an explicit "logging" section. We have something contained within the "The Starlite app" section about logging, so maybe it could go there, but that one actually doesn't explain a whole lot about logging. 

If you feel like it you can extend the section and make it more general, or don't, and we'll just add that to an upcoming "Logging" usage guide chapter.

provinzkraut (approved):

PR #1302, opened 2023-03-09
feat(ci): fix ruff-generated errata

JacobCoffee:
# PR Checklist

There is one lingering issue 
https://github.com/JacobCoffee/starlite/blob/13de06e968b1f388d3139e6311a942018d63f73c/starlite/_openapi/responses.py#L186-L189

```
starlite/_openapi/responses.py:186:9: SIM108 Use ternary operator `schema = Schema(oneOf=exceptions_schemas) if len(exceptions_schemas) > 1 else exceptions_schemas[0]` instead of `if`-`else`-block
Found 1 error.
```

I tried `schema = Schema(oneOf=exceptions_schemas) if len(exceptions_schemas) > 1 else exceptions_schemas[0]
` but then mypy/pyright gets mad and i am not sure how to proceed.

I am more than certain that some things maybe should be undone and added to an ignore (❓) but I am not familiar enough to know what.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -278,17 +278,17 @@ select = [
 ]
 ignore = [
     "E501", # pycodestyle line too long, handled by black
-    "D100", # pydocstyle -missing docstring in public module
-    "D101", # pydocstyle -missing docstring in public class
-    "D102", # pydocstyle -missing docstring in public method
-    "D103", # pydocstyle -missing docstring in public function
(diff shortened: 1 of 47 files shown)


JacobCoffee:
This is failing in quite a few places... we may need to revert some things I put into TYPE_CHECKING blocks and add it to ruff ignore

PR #1304, opened 2023-03-10
RFC: Remove `Cache` abstraction

provinzkraut:
Remove the `Cache` abstraction, since it's become mostly superfluous with #1184.

- Move `starlite/cache/config/` to  `starlite/config/cache.py`
- Remove `starlite/cache` module
- Remove `ASGIConnection.cache` property
- Remove `Starlite.cache` attribute
- Add `Starlite.cache_config` attribute
- Change `RateLimitConfig` to accept a `Storage` and `RateLimitMiddleware` accordingly to use this instead of `app.cache`

# PR Checklist


Goldziher:
How would users now access the cache? 

I don't understand why remove the cache property from connection and from the app.

provinzkraut:
> How would users now access the cache?

> I don't understand why remove the cache property from connection and from the app.

IMO it doesn't make a lot of sense to special-case the storage backend that's being used for caching. We also don't expose e.g. the storage backend used for sessions. The pattern we do recommend there is simply to create the backend outside the app, and then just use it there.

Another issue with this cache thing is that it only gives you access to *one* storage you have defined for *one* purpose, but exposing it encourages to use it as a general key/value store. This is conflicting an non-intuitive. Users might want to use different storage methods for caching, sessions, storing the rate limiting information and a general "utility cache", and as soon as they want to, this pattern falls apart because only one of those is exposed directly.

Further reasons to not promote this is that it's actually not that useful, since you probably *want* different storages, or at least (in the case of redis) namespaced storages. Right now, if you're using e.g. the memory backend (i.e. don't configure anything) you might think "I'll just call `app.cache.delete_all` or `request.cache.delete_all` to clear my request cache", but also delete all active sessions. Not special casing the one storage used for caching simply promotes more useful patterns.

Simply put, having a `cache` property/attribute on the app/connection is rarely actually useful, since for anything more complex than an example application, you probably need explicit access to different storages / namespaces within a storage anyway, and we promote this pattern elsewhere. It's just a very niche use case that I don't think is worth keeping around.

We should instead promote to use dependency injection to be able to access the storages within route handlers.

provinzkraut:
> We should instead promote to use dependency injection to be able to access the storages within route handlers.

If we want deeper integration of storages, I have another proposal for that, which I'm going to add in a different PR based on this one. IMO *if* we want this level of integration, we should do it properly, in a way that actually supports and makes use of the features we offer in this area.

Goldziher:
You need to access cache everywhere that a connection could he, including inside middleware. DI is not a path forward here. 

As for namespaces Caches or multiple backends being Standard - i strongly disagree. For most apps, certainly Microservices, a single cache is sufficient. 

The alternative would be to instead of having a single cache property, having a caches property that allows you to call get. This is the Django pattern .

But, it might we'll be over engineered.

chris-telemetry:
> You need to be able to access cache everywhere that a connection could be, including inside middleware and hooks.

This is correct and you do need the full functionality of it. Is there not a world where `app.cache` and `app.storages` can coexist?

provinzkraut:
> You need to be able to access cache everywhere that a connection could be, including inside middleware and hooks.

I agree, but why does it *need* to be via a property on the connection / app instance? 

> As for namespaced caches or multiple backends being the standard for most apps - I strongly disagree. For most apps, certainly Microservices, a single cache is sufficient.

I disagree with your disagreement :upside_down_face: 
For example you'll likely care about different things when it comes to request caching and session storage. Request caching can be highly volatile and doesn't necessarily need to persist across processes, while session storage absolutely needs to be persistent. So you maybe would want to use the in-memory storage for request caching, and redis for session storage.

> forcing users to define the storage backend for middlewares is bad DX. It's good to support this, optionally, but a default in memory Cache should be instantiated in any case.

This is already the case.

> You need to be able to access cache everywhere that a connection could be, including inside middleware and hooks.

Do you need access to the cache or the storage backends? Because I really don't see a reason why you need access to the `Cache` abstraction. All it does is add a method to build a cache key from a request, which doesn't have any use in session storage or a general key/value store. If we only have the `.cache` property, we'd need to expose all the other methods of the storage backends there as well to cover the (very common IMO) use case @chris-telemetry brought up earlier, and I really don't see a reason why this abstraction is needed for that. 

Talking about DX, it's also bad DX IMO if you set up 3 middlewares with 2 different storages, and then use `app.cache` which returns the storage backend used by one of those middlewares and obscures most of its functionality. That's not a very logical thing really.

Having a singular `cache` property only makes sense if we don't support using different storages at all. Otherwise we need to include a method to access those. What do you propose for that if we neither want to promote DI or the pattern of declaring your storage backends independently and simply importing them wherever you need?

Goldziher:
there is no problem removing the `Cache` abstraction. There is a problem removing the `.cache` attribute / property. I think you don't like the term `.cache` and prefer storages, but this is not really a compelling reason (i mean, it does compel one, due to some aesthetic sense, but that is subjective).

provinzkraut:
> I think you don't like the term .cache and prefer storages

I'm fine with either, really. It's not about the name (although I think consistent terminology would be great, but I'm impartial to which one is being followed). Let's stick with cache now for this conversation to make it a bit easier.

> Why? Because DI cannot work outside of handlers and other dependencies. Its that simple

Which is why I didn't propose DI as the only solution. The easiest solution would be to not change anything about the app at all and simply promote the pattern of defining your caches outside the application and importing them wherever you need. It's dead simple and allows for maximum flexibility. We also already have this pattern in our docs; We recommend this to manage session storage.

> i propose `.cache` which would be an alias for `app.caches.default_cache` -- which would be one (or the only) cache instantiated by the user.

This would actually be what I envisioned when I was talking about adding a way to handle multiple caches, although with a slightly different interface. I was going to propose  `.default_cache` instead of just `.cache`, just because it's more explicit.

chris-telemetry:
I disagree. There is a problem moving the `Cache` abstraction. The `Cache` abstraction gives user the opportunity to define behaviors that are _specific to response caching_ -- removing it would require users to bake caching rules/behaviors into `Storage` which would then reduce the value of `Storage` as a general way to interact with their preferred storage system.

chris-telemetry:
> If you're talking about actually subclassing `Cache`

This is what I was talking about.

> If you're talking about actually subclassing Cache to change behaviour on that level, you can achieve the same with a subclass of a Storage

Okay, I can see that. I agree here. But now I'm running 2 RedisStorage subclasses -- one for caching, one for general Redis storage. But if this is the case, as a user, I'm still going to want to be able to access `app.cache`.

provinzkraut:
> But now I'm running 2 RedisStorage subclasses -- one for caching, one for general Redis storage

Could you provide some code for those? I'd be interested in your use case and adaptions here.

> But if this is the case, as a user, I'm still going to want to be able to access `app.cache`

Would an interface as described by @Goldziher work for you? I.e. having a "default cache" available on the app/connection, in addition to a "cache repository" through which you could access other caches you have defined?

provinzkraut:
I've given this some more thought. 

**Naming problems**

Contrary to what I said before, there actually exist some naming problems. 

1. If we keep the `.cache` properties around as-is, we'll have `cache` returning a `Storage`, which isn't ideal semantically
2. Adding a `caches` kwarg to `Starlite` clashes with the `cache_config`. It would imply they do similar things, but they don't. Passing instances of `Storage` to `caches` also implies they're being used solely for caching, which they're not

Those problems can be solved in two ways:

1. Using the term "storage" consistently for the key/value stores and `cache` only for the request cache. We could then have a `.default_storage` property and a `storages` kwarg. 
2. Using the term "cache" for the key/value stores: Renaming the `cache_config`/`CacheConfig` > `request_cache_config`/`RequestCacheConfig` would enable us to then use a `caches` kwarg on `Starlite` and a `.cache`/`.default_cache` property

I actually think "storage" is the better term, since they are key/value stores. They *might* be used for caching, but it's not the only use (But it's also not a hill I'm willing to die on). On the other hand, this terminology is used elsewhere as well for this type of thing and is kinda accepted and generally understood. Furthermore, I think `request_cache_config` is a better term in any case, since it's more explicit.


**Default cache**

Currently `.cache` exposes the `Cache` abstraction, and with it the underlying `Storage` that's being used for request caching. This is problematic because it's not user configurable, and `.cache` necessarily will always be the same as the requests cache, which in many cases, isn't desirable. The requests cache and a "default cache" should not be tied together, as one might want to use different storage methods for both and interact with them independently (e.g. clear them) or as  @chris-telemetry mentioned, use specialised subclasses of a `Storage` for each purpose.

The solution in this case I think should be to make this transparently configurable in a way that still enables a no-config setup. I'm thinking of something similar to `logging.getLogger`. `.cache` should always be available, even if not explicitly configured by the user. 

**Caches used by Starlite internally**

Currently middlewares such as `RateLimitMiddleware` and request caching will use `.cache`, which comes with problems I described in the section above. Introducing a "cache registry" could be used to address this, but I don't think this is necessary. The changes introduces in this PR already allow passing a `Storage` instance to the respective configurations, while defaulting to a `MemoryStorage`. If we were to add this functionality, for the sake of enabling user access via the app instance, it would be relatively straightforward to set this up in a way that provides similar functionality.

provinzkraut:
Playing around with a possible "cache repository" a bit actually furthered my belief that this is not the way to go, since it's unnecessary and does not enable anything that can't also easily be done without it. And the best implementation is always no implementation.

So why is it no needed?

As far as I can tell there are 4 problems it would solve:

1. A central location to register and access caches, making them available throughout all application contexts
2. Access to caches used internally by Starlite (`RateLimitMiddleware` and request caching)
3. Configuration of caches used by Starlite internally
4. Access to an (optionally user-defined) defined "default cache"

All of those can be (in my opinion more easily) solved in other ways, without adding additional abstractions.

**The solution**

Simply put caches in a module and import them from there. This is way more simple and additionally allows to use them outside the application context as well. It's not different than defining a `Redis` client instance somewhere and then importing this wherever needed, which is a very common practice. 

If we change the way `RateLimitConfig` and `CacheConfig` work (as proposed in this PR) to make them accept `Storage` instances directly and default to setting up their own `MemoryStorage` if none is provided, this also solves problem 2, 3 and 4 and in addition gets rid of the request cache leaking into other parts of the application without being able to configure it independently.

1. Solved by simply importing the caches
2. Solved by passing the centrally defined caches to the specific configurations
3. Solved by allowing to pass caches directly to configurations
4. Not needed since no "global default cache" is required anymore

My proposal would then result in something like this:

```python
from starlite import Starlite
from starlite.middlewares.rate_limit import RateLimitConfig
from starlite.config.request_cache import RequestCacheConfig

from my_package.storages import request_cache, rate_limit_cache

app = Starlite(
    ..., 
    rate_limit_config=RateLimitConfig(storage=rate_limit_cache),
    request_cache_config=RequestCacheConfig(storage=request_cache)
)
```

**The global default cache**

Having this non-configurable and depend on whatever cache is used for request caching is not very useful behaviour,  confusing and can lead to unwanted side effects (e.g. flushing request cache alongside the rate limit data). In order for this to be configurable we'd have to add e.g. a `default_cache` kwarg to `Starlite` which would then be returned by the `.cache` property. This however is essentially just setting an instance attribute and in my opinion not more useful that putting a reference to the cache passed to it into the state or a central module.

If we conclude that this is really needed, I'd say it should simply exist as a convenience feature for users who don't want to configure a cache backend or just need some one-off volatile storage, but it shouldn't be used by Starlite itself internally. This would enable to still go with the solution proposed above.

@starlite-api/maintainers I don't think I have any more thoughts on this and have explored most angles from my side (the ones I deem somewhat useful that is :upside_down_face:) so I'd appreciate some input.

provinzkraut:
Closed in favour of #1330

PR #1306, opened 2023-03-10
Update OpenAPI logic to not rely on pydantic

Goldziher:
# PR Checklist


peterschutt (approved):
Quite a big one to go through, so only a skim. I noticed that:

- appears to be a re-modelling of (mostly) the same datastructures from pydantic-openapi-schema.
- I noticed there is some attribute renaming e.g., camelCase to snake_case, and things like `param_schema` has become `schema` in places.
- I noticed that there are changes to the content assertions in tests, things like "security_schema_in" has become "in", and "media_type_schema" has become "schema" - stuff like that.

On the basis that is a fair summary of the gist of the changes, and that the changes I've mentioned are as intended, I think the most efficient way to get a better review on this is to get it in main and getting used.

Also, there was at least 1 `__all__ = [...]` with square brackets.

Issue #1299, opened 2023-03-09
Enhancement: Content negotiation and Accept header


NomAnor:
It would be nice to implement some kind of content negotiation based on the `Accept` header. 

I have an endpoint that should serve a graph in different formats. Depending on the requested mime type it can respond with an image, a PDF or the raw graph as a graphviz file. Currently I crudly parse the `Accept` header myself:

```python
def parse_accept(accept: str) -> List[str]:
    """
    Parse an Accept header string an return the list of accepted mime types.
    """
    mime_types = []
    for entry in accept.split(','):
        entry = entry.strip()
        type, _, extra = entry.partition(';')
        mime_types.append(type)
    return mime_types

def mime_type_matches(mime_type: str, pattern: str) -> bool:
    if pattern == '*/*':
        return True

    type, _, subtype = mime_type.partition('/')
    pattern_type, _, pattern_subtype = pattern.partition('/')

    return type == pattern_type and (subtype == pattern_subtype or pattern_subtype == '*')

def first_matching_mime_type(accept: str, available: List[str], default: Optional[str] = None) -> Optional[str]:
    accept_patterns = parse_accept(accept)
    for pattern in accept_patterns:
        for type in available:
            if mime_type_matches(type, pattern):
                return type
    return default

@get("/graph")
    async def graph(self, accept: str = Parameter(header="Accept")) -> Response[bytes]:
         mime_type = first_matching_mime_type(accept, ['image/png', 'application/pdf', 'text/vnd.graphviz'], default='image/png')
         with io.BytesIO() as f:
            if mime_type == 'image/png':
                render_graph_image(f)
            elif mime_type == 'application/pdf':
                render_graph_pdf(f)
            elif mime_type == 'application/pdf':
                raw_graph(f)
            return Response(content=f.getvalue(), media_type=mime_type)
```
I'm not sure what the best interface would be. Maybe just a special dependency that provides you with the requested type based on a list of available type. An integration into OpenAPI would also be nice, but I don't know if that is even possible.

The hard part is probably correctly parsing and matching the mime types, thats why i made this request. I haven't looked in the specs.

Django REST Framework has something [similar](https://www.django-rest-framework.org/api-guide/content-negotiation). But the render concept doesn't fit into starlite.

Goldziher:
Hi,

So this could be a nice ready made dependency packages with Starlite. I don't see how we could implement the content negotiation - this is for the user to do. But we could parse the header and offer a ready made dependency.

[...]

PR #1317, opened 2023-03-13
Implement #1299: Accept header parsing and content negotiation

NomAnor:
First iteration to implement Accept header parsing and content negotioation. Documentation is missing until the code is stable.

# Things to discuss/check

- I'm not sure where to put the helper types, `Accept` might reside in request.py, the other I don't know.
- Should `MediaType` be exposed or only used internally?
- Are the tests appropiate?
- Is the ordering correct based on [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-14.1)?

# MediaType class
Exposing the results as `str` is obviously easier. But when paramters are involved it gets more complicated:
```
Accept("text/plain,text/plain;p=test").best_match(["text/plain", "text/html"]) == "text/plain;p=test"
```
In a handler you would do:
```
preferred_type = request.accept.best_mach(["text/plain", "text/html"])
if preferred_type == "text/plain":
   return text_response
```
which would not match because of the missing parameter. 

Implementing parameters might just be too cumbersome and so rarely used that it might be best to ignore them, except for `q` and for specificity. Or my logic could be better and the method should only return elements from `provided_types`.

# PR Checklist

### DIFF
--- a/docs/usage/responses.rst
+++ b/docs/usage/responses.rst
@@ -144,20 +144,11 @@ Content Negotiation
 If your handler can return data with different media types and you want to use
 `Content Negotiation <https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation>`_
 to allow the client to choose which type to return, you can use the
-:class:`Request.accept <starlite.connection.request.Request>` property to
+:attr:`Request.accept <starlite.connection.Request.accept>` property to
 calculate the best matching return media type.
 
 .. literalinclude:: /examples/responses/response_content.py
     :language: python
-
-.. code-block:: console
-
-    $ curl -H "Accept: text/plain" http://127.0.0.1:8000/resource
-    Hello World!
-    $ curl -H "Accept: text/html" http://127.0.0.1:8000/resource
-    <h1>Hello World!</h1>
-    $ curl -H "Accept: application/xml" http://127.0.0.1:8000/resource
-    <xml><msg>Hello World!</msg></xml>
 
 
 Status Codes
--- a/docs/usage/responses.rst
+++ b/docs/usage/responses.rst
@@ -138,6 +138,28 @@ For ``MediaType.HTML``, route handlers should return a :class:`str` or :class:`b
    and to write the template itself in a separate file rather than a string.
 
 
+Content Negotiation
+-------------------
+
+If your handler can return data with different media types and you want to use
+`Content Negotiation <https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation>`_
+to allow the client to choose which type to return, you can use the
+:class:`Request.accept <starlite.connection.request.Request>` property to
+calculate the best matching return media type.
+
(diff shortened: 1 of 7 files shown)


provinzkraut:
> which would not match because of the missing parameter.

IMO this can be simplified. `best_match` should **always** only return one of the types passed in. If you pass in `text/plain;p=test`, you may handle that case and get back a `text/plain;p=test`, but if you pass in `text/plain`, all you should get back is `text/plain;p=test`, regardless of the header value.

provinzkraut:
> I'm not sure where to put the helper types, ``Accept` might reside in request.py, the other I don't know.

`datastructures`. We have an `ETag` header in `headers.py` there as well for example.

NomAnor:
Should the `Accept` class inherit from `Header` and become a pydantic model? 
It looks like those classes (ETag, CacheControlHeader) are only used for responses.

provinzkraut:
> Should the `Accept` class inherit from `Header` and become a pydantic model?

Hm. I think we're running into a bit of an issue here. Conceptually, they should have the same base class. But as you pointed out, `Header` is currently only used for response headers. It *should* be usable for request headers as well. On the other hand, we don't really want to introduce pydantic models unnecessarily, as we're currently trying to move away from those and towards dataclasses.

On the other hand, there's no reason you can't just implement an `Accept` class that doesn't inherit from any of those. I think this might be the way to go here for now.

NomAnor:
I moved the code to the dataclasses directory. I also changed it so the interface uses `str` and only internally the `_MediaType` class. 
I added a small initial paragraph for content negotiation to the response documentation,but it needs a bit more polish.

NomAnor:
I cleaned everything up, it should be ready for a final review.

I could not get sphinx to reference the ``accept`` property. Does anyone know why ``:attr:`Request.accept <starlite.connection.request.Request.accept>` `` does not work?

docs/usage/responses.rst
```diff
@@ -138,6 +138,28 @@ For ``MediaType.HTML``, route handlers should return a :class:`str` or :class:`b
...
+.. code-block:: console
+
+    $ curl -H "Accept: text/plain" http://127.0.0.1:8000/resource
+    Hello World!
+    $ curl -H "Accept: text/html" http://127.0.0.1:8000/resource
+    <h1>Hello World!</h1>
+    $ curl -H "Accept: application/xml" http://127.0.0.1:8000/resource
+    <xml><msg>Hello World!</msg></xml>
+
+
```
provinzkraut, line 162: We have automatic example execution that will keep this always up to date with the example and include nicely formatted output. Check it out: https://github.com/starlite-api/starlite/blob/main/CONTRIBUTING.rst#adding-examples

NomAnor: I missed that, but I got it working now.

Goldziher (approved):

peterschutt (approved):

PR #1320, opened 2023-03-14
Feature: Generic controller

peterschutt:
Controller type that supports generic inheritance hierarchies.

# PR Checklist


starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,46 @@
...
+T = TypeVar("T")
+
+
+class GenericController(Controller, Generic[T]):
+    """Controller type that supports generic inheritance hierarchies."""
+
+    _data_type: type[T]
+
+    def __class_getitem__(cls, data_type: type[T]) -> type[Self]:
+        return type(f"Controller[{data_type.__name__}]", (cls,), {"_data_type": data_type})
```
peterschutt, line 23:
```suggestion
        return type(f"GenericController[{data_type.__name__}]", (cls,), {"_data_type": data_type})
```

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def set_parameter_annotation(self, parsed_parameter: ParsedSignatureParameter) -> None:
+        """Substitute any ``TypeVar`` in annotation of ``parsed_parameter`` with narrowed type.
+
+        Args:
+            parsed_parameter: a parsed signature parameter - should have a ``TypeVar`` as its type.
```
peterschutt, line 29:
```suggestion
            parsed_parameter: a parsed signature parameter.
```

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        if isinstance(parsed_parameter.annotation, TypeVar):
+            parsed_parameter.annotation = self._data_type
+            return
+
+        args = get_args(parsed_parameter.annotation)
+        origin = get_origin(parsed_parameter.annotation)
+
+        if args and origin:
+            new_args = tuple(self._data_type if isinstance(arg, TypeVar) else arg for arg in args)
```
peterschutt, line 39: should return early if no typevars

starlite/_signature/parsing.py
```diff
@@ -145,6 +152,11 @@ def parse_fn_signature(
     parsed_params: list[ParsedSignatureParameter] = []
     dependency_names: set[str] = set()
     fn_type_hints = get_fn_type_hints(fn)
+    return_annotation = fn_type_hints.get("return", signature.empty)
+
+    owner_is_generic_controller = owner and isinstance(owner, GenericController)
```
provinzkraut, line 157:
```suggestion
    owner_is_generic_controller = isinstance(owner, GenericController)
```

should be good enough

starlite/utils/compat.py
```diff
@@ -43,8 +43,8 @@ def py_39_safe_annotations(annotated: Any) -> Generator[Any, None, None]:
     Yields:
```
provinzkraut, line 43: Not really related to this PR, but I recently ran into issue with this function because it only handled `Union`, and not the built-in subscriptable types. This causes route handlers defined with them to error out when used in a lower python version and with `__future__.annotations`. What do you think about handling those cases as well?

If not, could you elaborate why we only handle this specific case? (I can't recall anymore what the exact reason was you added this, sorry :grimacing:)

provinzkraut (commented):
Looking really good!

starlite/utils/compat.py
peterschutt: I regret having implemented it and want to remove it - its trying to do too much.

starlite/controller/generic_controller.py
Goldziher: Raise ImproperlyConfigured?

starlite/utils/compat.py
provinzkraut: Maybe we can remove it then? Or narrow its scope somewhat?

peterschutt: PR up to remove it

starlite/_signature/parsing.py
```diff
@@ -125,14 +128,18 @@ def get_type_annotation_from_plugin(
 
 
 def parse_fn_signature(
-    fn: AnyCallable, plugins: list[SerializationPluginProtocol], dependency_name_set: set[str]
+    fn: AnyCallable,
+    plugins: list[SerializationPluginProtocol],
+    dependency_name_set: set[str],
+    owner: Controller | Router | None = None,
```
Goldziher, line 134: Why pass owner here rather than simply pass in the annotations? It's better to inject these values from outside than to make this function more complex.

starlite/_signature/parsing.py
Goldziher, line 157: This logic can be done using a helper, and `return_annotation` can become a kwarg to this function.

starlite/_signature/parsing.py
```diff
@@ -145,6 +152,11 @@ def parse_fn_signature(
     parsed_params: list[ParsedSignatureParameter] = []
     dependency_names: set[str] = set()
     fn_type_hints = get_fn_type_hints(fn)
+    return_annotation = fn_type_hints.get("return", signature.empty)
+
+    owner_is_generic_controller = owner and isinstance(owner, GenericController)
+    if owner_is_generic_controller:
+        return_annotation = cast("GenericController", owner).get_parameter_annotation(return_annotation)
```
Goldziher, line 159: Same

starlite/_signature/parsing.py
```diff
@@ -174,24 +186,31 @@ def parse_fn_signature(
         if isinstance(parameter.default, ParameterKwarg) and parameter.default.value_type is not Empty:
             parameter.annotation = parameter.default.value_type
 
+        if owner_is_generic_controller:
+            parameter.annotation = cast("GenericController", owner).get_parameter_annotation(parameter.annotation)
```
Goldziher, line 190: Same

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,70 @@
...
+        new_args = []
+        for arg in args:
+            if isinstance(arg, TypeVar):
+                new_args.append(self._data_type)
+                type_var_found = True
+                continue
+            new_args.append(arg)
+
+        if not type_var_found:
+            return None
```
Goldziher, line 40: I think we should raise an ImproperlyConfiguredException - generic controllers should be used with geberics

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return tuple(new_args)
+
+    @staticmethod
+    def _rebuild_annotation(original_annotation: Any, origin: Any | None, args_tuple: tuple[Any, ...]) -> Any:
+        if origin is None:
+            raise RuntimeError("Unexpected origin value")
```
Goldziher, line 47: Why not ImproperlyConfigured?

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,70 @@
...
+                continue
+            new_args.append(arg)
+
+        if not type_var_found:
+            return None
+
+        return tuple(new_args)
+
+    @staticmethod
+    def _rebuild_annotation(original_annotation: Any, origin: Any | None, args_tuple: tuple[Any, ...]) -> Any:
```
Goldziher, line 45: I don't think this method needs to be attached to the class. It can and should go into a utils module, or be declared at the top of the file

starlite/controller/generic_controller.py
```diff
@@ -0,0 +1,70 @@
...
+        """Substitute any ``TypeVar`` in annotation of ``parsed_parameter`` with narrowed type.
+
+        Args:
+            annotation: a parsed signature parameter's annotation.
+        """
+        if isinstance(annotation, TypeVar):
+            return self._data_type
+
+        args = get_args(annotation)
+        origin = get_origin(annotation)
```
Goldziher, line 65: We have a `_get_origin` util

starlite/controller/generic_controller.py
peterschutt: This is called for every parameter to the handler though - so that would break something like:

```py
class GenCont(...):
   @put()
   def handle_put(item_id: int, generic_collection: list[ThingT]) -> ...:
       ...
```
The method is for identifying that `ThingT` needs to be swapped with a concrete `Thing` in the signature model parameter annotation, and the `return None` is for breaking early when we detect the `item_id` case - that is, where there is no chance of there being a type var in that parameter's annotation.

starlite/controller/generic_controller.py
peterschutt: I don't think it should ever hold true - but I am not confident enough to say that a user has improperly configured something if it does. 

I want to narrow the type of origin from optional to not-optional, and explicitly find out when my assumption that it should never be `None` if untrue. So it is an error that I want to be an internal error, but not necessarily a user-did-something-wrong error.

I'll just change it to StarliteException.

starlite/controller/generic_controller.py
peterschutt: yeh I agree - have moved to a helper

starlite/controller/generic_controller.py
peterschutt: Just had a look:

```py
def _get_origin(annotation: Any) -> Any:
    origin = get_origin(annotation)
    return origin if origin not in (Annotated, Required, NotRequired) else get_args(annotation)[0]
``` 

```python-console
>>> from starlite.utils.predicates import _get_origin
>>> from typing import List
>>> from typing_extensions import Annotated
>>> _get_origin(List[int])
<class 'list'>
>>> _get_origin(Annotated[List[int], "something"])
typing.List[int]
```
They should return the same thing I believe.

```py
def _get_origin(annotation: Any) -> Any:
    origin = get_origin(annotation)
    return origin if origin not in (Annotated, Required, NotRequired) else get_origin(get_args(annotation)[0])
```

Also, if it is intended to be used outside of that module, it shouldn't be a private function I think. I didn't know it existed, but given that it is private like that I wouldn't have used it anyway.

I can do a commit for these things and start to use it. There is definitely a set of operations to get at type info that should be centralized.

starlite/_signature/parsing.py
peterschutt: I use the owner directly, for both identifying that it is a subclass of `GenericController` and for calling a method on it.

I can remove the method call, but I still need to know that we are dealing with a signature from a handler that is defined on a generic controller before I can start attempting to swap out type vars in the signature.

peterschutt: The method called on the owner is `owner.get_parameter_annotation()` which is the one that detects if a type var exists and swaps it out for the type it has been narrowed with. So to remove everything I'm using the owner for here then I need to pass in a boolean indicating whether the owner is a generic controller, and the type that the generic controller has been narrowed with which is otherwise stored on the generic controller subtype.

starlite/controller/generic_controller.py
Goldziher: sure

Goldziher (approved):

PR #1322, opened 2023-03-14
Various CLI updates

### LINKED ISSUE #1227: Enhancement: Enable CLI autodiscovery for applications in arbitrarily named files
As discussed in #1221, the CLI should be able to pick up applications from arbitrarily named files within the current working directory. 

The discovery flow would then be:

1. `--app` option passed to the CLI
2. `STARLITE_APP` env var
3. In any of the canonical paths (`app.py`, `application.py`, `asgi.py`):
    1. A `Starlite` instance 
    2. A `create_app` factory
    3. Any callable that's annotated to return a `Starlite` instance
4. Any other `.py` files in the current working directory according to the logic described in step 3

provinzkraut:
Closes #1227.

- Update CLI autodiscovery logic to only consider canonical modules `app` and `application`, but every `Starlite` instance or application factory able to return a `Starlite` instance within those or one of their submodules, giving priority to the canonical names `app` and `application` for application objects and submodules containing them.
- Fix a bug introduced in #1218 that would print the application info twice
- Change `StarliteEnv` to be lazily loaded, only if a command requested it (via the `app` or `env` kwarg), enabling commands to run without an application being present
- Add a `version` command to display the currently installed Starlite version
- Some test tweaks

# PR Checklist


starlite/cli/_utils.py
```diff
@@ -5,9 +5,10 @@
 import sys
 from dataclasses import dataclass
 from functools import wraps
+from itertools import chain
 from os import getenv
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, Iterable, Sequence, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Callable, Iterable, Sequence, TypeVar, cast, Generator
```
peterschutt, line 11:
```suggestion
from typing import TYPE_CHECKING, Any, Callable, Generator, Iterable, Sequence, TypeVar, cast
```

peterschutt (approved):

starlite/cli/_utils.py
provinzkraut: Shouldn't this have been autofixed?

peterschutt: I'd have thought so..

peterschutt: think `cli` needs to be added to `tool.ruff.src` in pyproject

provinzkraut: Hm, no, it seems to run on that file in general.

provinzkraut: Well.. Turns out we didn't enable the isort rules for ruff.

peterschutt: ok well that was the only dart I had to throw :)

PR #1330, opened 2023-03-15
Refactor storages and add `StoreRegistry`

provinzkraut:
## Content

This builds upon #1304.

- Add `Starlite.cache_config` attribute
- Rename `starlite/storage` > `starlite/stores` and update general naming to "store" instead of "storage"
- Add `starlite/stores/registry.py` module, implementing a `StoreRegistry`
- Add `Starlite.stores` attribute, containing a `StoreRegistry`
- Add `stores` kwarg to `Starlite` and `AppConfig` to allow seeding of the `StoreRegistry`
- Change `RateLimitMiddleware` to use `app.storages`
- Change request caching to use `app.storages`
- Change server side sessions to use `app.storages`
- Move `starlite/cache/config/` to  `starlite/config/response_cache.py`
- Rename `CacheConfig` > `ResponseCacheConfig`
- Rename `Starlite.cache_config` > `response_cache_config`
- Rename `AppConfig.cache_config` > `response_cache_config`
- Remove `starlite/cache` module
- Remove `ASGIConnection.cache` property
- Remove `Starlite.cache` attribute

<hr> 

This fundamentally changes how different parts of the app interact with stores, and in its current state, requires less configuration by default.

At the core of this pattern is the `StoreRegistry`, which provides a central places where stores are registered under a name. The current implementation is not yet complete, and some details still have to be worked out. Open questions at the bottom.

<hr>

### Examples

```python
from starlite import Starlite

app = Starlite(...)
some_store = app.stores.get("some")
# we now have access to a newly created store instance. 
# This is provided by the registry via a factory function, 
# that gets called every time a `.get` call requests an unknown name.
assert app.stores.get("some") is some_store  # subsequent calls will return the registered instance
``` 

This allows configuration requiring very little boilerplate, but still offering access to the stores used:
```python
from starlite import Starlite
from starlite.middleware.rate_limit import RateLimitConfig

app = Starlite(..., rate_limit_config=RateLimitConfig(("second", 1))))
# RateLimitMiddleware will request a store via app.stores.get("rate_limit")
app.stores.get("rate_limit")  # and  we can access the store directly without having to set it up
```

By default, the default factory will return a new instance of `MemoryStore`, but we can configure it ourselves. 
In this example, we return the same `MemoryStore` instance every time.

```python
from starlite import Starlite
from starlite.stores.memory import MemoryStore

from starlite.stores.registry import StoreRegistry

memory_store = MemoryStore()
app = Starlite(..., stores=StoreRegistry(default_factory=lambda _: memory_store))
```

This pattern can also be used to conveniently build a store hierarchy. This simple configuration will use the same underlying redis connection, while making use of the namespacing feature. Every time a new store is requested via `StoreRegistry.get`, that hasn't been registered yet, the registry will call `root_store.with_namespace`, to create a new namespaced `RedisStore` instance:

```python
from starlite import Starlite, get
from starlite.stores.redis import RedisStore
from starlite.middleware.rate_limit import RateLimitConfig
from starlite.middleware.session.server_side import ServerSideSessionConfig
from starlite.stores.registry import StoreRegistry

root_store = RedisStore.with_client("redis://localhost")


@get(cache=True)
def cached_handler() -> str:
    # this will use app.stores.get("request_cache")
    return "Hello, world!"


app = Starlite(
    [cached_handler],
    stores=StoreRegistry(default_factory=root_store.with_namespace),
    middleware=[
        RateLimitConfig(("second", 1)).middleware,
        ServerSideSessionConfig().middleware,
    ],
)
```

the same can be done with a `FileStore`:

```python
from pathlib import Path

from starlite import Starlite, get
from starlite.middleware.rate_limit import RateLimitConfig
from starlite.middleware.session.server_side import ServerSideSessionConfig
from starlite.stores.file import FileStore
from starlite.stores.registry import StoreRegistry

root_store = FileStore(Path("data"))


@get(cache=True)
def cached_handler() -> str:
    # this will use app.stores.get("request_cache")
    return "Hello, world!"


app = Starlite(
    [cached_handler],
    stores=StoreRegistry(default_factory=root_store.with_namespace),
    # the line above is essentially equivalent to
    # stores=StoreRegistry(default_factory=lambda name: FileStore(root_store.path / name)),
    middleware=[
        RateLimitConfig(("second", 1)).middleware,
        ServerSideSessionConfig().middleware
    ],
)
```

It is also possible to explicitly set up stores:

```python
from pathlib import Path

from starlite import Starlite, get
from starlite.middleware.rate_limit import RateLimitConfig
from starlite.middleware.session.server_side import ServerSideSessionConfig
from starlite.stores.file import FileStore
from starlite.stores.redis import RedisStore
from starlite.stores.memory import MemoryStore

file_store = FileStore(Path("data"))
redis_store = RedisStore.with_client("redis://localhost")
memory_store = MemoryStore()


@get(cache=True)
def cached_handler() -> str:
    return "Hello, world!"


app = Starlite(
    [cached_handler],
    stores={
        "request_cache": redis_store,
        "sessions": file_store,
        "rate_limit": memory_store,
    },
    middleware=[
        RateLimitConfig(("second", 1)).middleware,
        ServerSideSessionConfig().middleware,
    ],
)
```

or by passing a store name to the configs, which also allows to easily re-use stores:

```python
from pathlib import Path

from starlite import Starlite, get
from starlite.middleware.rate_limit import RateLimitConfig
from starlite.middleware.session.server_side import ServerSideSessionConfig
from starlite.stores.memory import MemoryStore


@get(cache=True)
def cached_handler() -> str:
    return "Hello, world!"


app = Starlite(
    [cached_handler],
    stores={"root": MemoryStore()},
    middleware=[
        RateLimitConfig(("second", 1), store="root").middleware,
        ServerSideSessionConfig(store="root").middleware,
    ],
    response_cache_config=ResponseCacheConfig(store="root")
)
```

<hr>

## Open questions

~~**Should the store names used by various Starlite features be configurable or constant?**~~
(This has since been decided: Store names are configurable for each integration)

Currently, for the sake of keeping this RFC simple, the names of the stores used by Starlite internally (e.g. by `RateLimtiMiddleware` are hardcoded. 

A disadvantage of this is that it decouples the configuration value from where it's used. For example, there is a`RateLimitConfig`, but you can only specify which store it should use via another configuration value on the application.

```python
app = Starlite(
    ..., 
    stores={"rate_limit": some_store, "session": some_store}, 
    rate_limit_config=RateLimitConfig(("second", 1)),
    middleware=[SServerSideSessionConfig().middleware]
)
```

If we were to add a `store` field to those, it could be instead configure like:

```python
app = Starlite(
    ..., 
    stores={"some_store": some_store}, 
    rate_limit_config=RateLimitConfig(("second", 1), store="some_store"),
    middleware=[SServerSideSessionConfig(store="some_store").middleware]
)
```

**Should users be forced to go through the registry?**

As of now, the only way to configure a store for e.g. `RateLimitMiddleware` is by setting it via the `stores` kwarg to `Starlite` (e.g. `Starlite(..., stores={"rate_limit": my_store}`). 

The benefit of this is that all stores will always be registered, making them accessible everywhere, and configuration relatively simple, since there's only one way to set things up.

**Should the registry include a "default cache" property?**

Currently, the registry includes a `.default` property, which is equivalent to `registry.get("default")`. This is intended to mimic the `.cache` property of the `Starlite` app / `ASGIConnection`, which this PR removes.

The issue with using a default cache like this is that it returns the same instance, occupying the same namespace. This is not desirable in most cases, and can lead to issue if users do not pay attention to this. If we do not offer such a property, we would instead promote the pattern of using `registry.get(<my store>)`, to get a unique, namespaced store for each use case, while still allowing a user to override this behaviour (using the `default_factory`) if they so wish.

Especially for third party integrations that wish to make use of the stores provided by Starlite this pattern could prove very valuable, since no precautions have to be taken how to safely operate with a store without interfering with data not owned by this party. 

If we don't offer a "default cache", we will also ensure that a user is always able to control which integration uses which store, via the registry.

*Considerations*

For convenience (`connection.app.stores.get("foo")` is quite complicated), we could proxy `app.stores` in these cases, like `connection.cache` does in its current implementation.

**Should the registry be "frozen" after application startup?**

Currently, you can register (and possibly override) stores at any time through the registry. This might lead to side effects if a store is acquired via `registry.get("some store")`, and later on overridden. We could add an option that prevents store overrides after application startup.

### DIFF
--- a/docs/examples/tests/test_stores.py
+++ b/docs/examples/tests/test_stores.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from unittest.mock import MagicMock, patch
 
 import anyio
 
@@ -9,7 +10,8 @@ from starlite.stores.redis import RedisStore
 from starlite.testing import TestClient
 
 
-def test_configure_integrations_set_names() -> None:
+@patch("starlite.stores.redis.Redis")
+async def test_configure_integrations_set_names(mock_redis: MagicMock) -> None:
     from examples.stores.configure_integrations_set_names import app
 
     assert isinstance(app.stores.get("redis"), RedisStore)
@@ -46,14 +48,6 @@ async def test_delete_expired_on_startup(tmp_path) -> None:
         assert not await file_store.exists("foo")
 
 
-async def test_expiry(monkeypatch) -> None:
-    pass
-
-
-async def test_expiry_renew_on_get() -> None:
-    pass
-
-
 async def test_get_set(capsys) -> None:
     from examples.stores.get_set import main
 
@@ -62,11 +56,7 @@ async def test_get_set(capsys) -> None:
     assert capsys.readouterr().out == "None\nb'value'\n"
 
 
-def test_namespacing() -> None:
-    pass
-
-
-def test_registry() -> None:
+async def test_registry() -> None:
     from examples.stores.registry import app, memory_store, some_other_store
 
     assert app.stores.get("memory") is memory_store
@@ -75,7 +65,7 @@ def test_registry() -> None:
     assert some_other_store is not memory_store
 
 
-def test_registry_access_integration() -> None:
+async def test_registry_access_integration() -> None:
     from examples.stores.registry_access_integration import app, rate_limit_store
 
     assert app.stores.get("rate_limit") is rate_limit_store
@@ -83,7 +73,8 @@ def test_registry_access_integration() -> None:
     assert app.middleware[0].kwargs["config"].get_store_from_app(app) is rate_limit_store
 
 
-def test_configure_integrations() -> None:
+@patch("starlite.stores.redis.Redis")
+async def test_configure_integrations(mock_redis: MagicMock) -> None:
     from examples.stores.registry_configure_integrations import app
 
     session_store = app.middleware[0].kwargs["backend"].config.get_store_from_app(app)
@@ -94,14 +85,15 @@ def test_configure_integrations() -> None:
     assert cache_store.path == Path("request-cache")
 
 
-def test_registry_default_factory() -> None:
+async def test_registry_default_factory() -> None:
     from examples.stores.registry_default_factory import app, memory_store
 
     assert app.stores.get("foo") is memory_store
     assert app.stores.get("bar") is memory_store
 
 
-def test_default_factory_namespacing() -> None:
+@patch("starlite.stores.redis.Redis")
+async def test_default_factory_namespacing(mock_redis: MagicMock) -> None:
     from examples.stores.registry_default_factory_namespacing import app, root_store
 
     foo_store = app.stores.get("foo")


tests/caching/test_response_caching.py
```diff
@@ -72,13 +72,15 @@ async def test_custom_cache_key(sync_to_thread: bool, anyio_backend: str) -> Non
...
+    )(slow_handler)
+
+    app = Starlite(
+        [route_handler],
+    )
+
+    with TestClient(app) as client:
         client.get("/cached")
-        value = await client.app.cache.get("/cached:::cached")
+        value = await app.cache_config.get_store_from_app(app).get("/cached:::cached")
```
Goldziher, line 85: This is not good IMO.

starlite/app.py
```diff
@@ -207,6 +206,7 @@ def __init__(
         response_headers: OptionalSequence[ResponseHeader] = None,
         security: OptionalSequence[SecurityRequirement] = None,
         static_files_config: OptionalSequence[StaticFilesConfig] = None,
+        stores: StoreRegistry | dict[str, Store] | None = None,
```
Goldziher, line 209: I think this API is bad. 

1. It's not ergonomic at all. 
2. It's overly complex.
3. It's not junior friendly. 

Why not simply let the user specify either a single or a sequence if store configs and pass them to the app? Expose an `is_default_store` on the config, and also a `key`, and you can easily generate a mapping from this internally. 

You also allow plugins yo modify it, as needed.

starlite/app.py
```diff
@@ -101,10 +103,6 @@
 """The default OpenAPI config used if not configuration is explicitly passed to the
 :class:`Starlite <.app.Starlite>` instance constructor.
 """
-DEFAULT_CACHE_CONFIG = CacheConfig()
```
Goldziher, line 104: Why? What reason is there to not init an in memory cache by default? 

There were exactly 0 issues related to this, and no complaints. I don't get the reason for this change.

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+from .memory import MemoryStore
+
+__all__ = ("StoreRegistry",)
+
+
+def default_default_factory(name: str) -> Store:
+    return MemoryStore()
+
+
+class StoreRegistry:
```
Goldziher, line 18: I don't see why this class is required, can you elaborate?

Goldziher (changes requested):
Ok, gotta say I'm a bit annoyed. I feel like I'm returning to the same points again and again. 

One thing is to remove the cache abstraction, fine. But I'm really not on board with removing the access to the default store from request, app etc. I think there are a lot of over complications here, and am overall degradation of DX.

provinzkraut:
> Ok, gotta say I'm a bit annoyed. I feel like I'm returning to the same points again and again.
> 
> One thing is to remove the cache abstraction, fine. But I'm really not on board with removing the access to the default store from request, app etc. I think there are a lot of over complications here, and am overall degradation of DX.

Please wait until I'm done with the examples and reasoning. I specifically took care of addressing your particular concerns.

starlite/app.py
provinzkraut: > Why? What reason is there to not init an in memory cache by default? 

Please try to follow the examples and read through the PR carefully. This is not needed anymore and a default in-memory cache is provided in *every* case. I explained elsewhere why it is a very bad idea to use **the same** instance of an in memory cache, which is essentially a dictionary, throughout the whole application.

starlite/stores/registry.py
provinzkraut: I did. Read the initial comment on the PR. This specifically addresses several issues you raised.

starlite/middleware/rate_limit.py
```diff
@@ -69,20 +69,19 @@ async def __call__(self, scope: "Scope", receive: "Receive", send: "Send") -> No
         Returns:
             None
         """
-        if not hasattr(self, "cache"):
-            self.cache = scope["app"].cache
-
-        request: Request[Any, Any, Any] = scope["app"].request_class(scope)
+        app = scope["app"]
+        request: Request[Any, Any, Any] = app.request_class(scope)
+        store = self.config.get_store_from_app(app)
```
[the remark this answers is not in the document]
provinzkraut: No. Using the same instance of a store for different things by default is bad design. I explained why several times. If you have a good reason why this should be the case, please let me know.

starlite/connection/base.py
```diff
@@ -263,15 +262,6 @@ def logger(self) -> Logger:
...
         return self.app.get_logger()
 
-    @property
-    def cache(self) -> Cache:
-        """Return the ``Cache`` for this connection.
-
-        Returns:
-            A ``Cache`` instance.
-        """
-        return self.app.cache
```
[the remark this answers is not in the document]
provinzkraut: It is not, in my opinion, but if you feel strongly about this, we can add one that returns `self.app.stores.default`.

Goldziher:
> > Ok, gotta say I'm a bit annoyed. I feel like I'm returning to the same points again and again.
> > 
> > One thing is to remove the cache abstraction, fine. But I'm really not on board with removing the access to the default store from request, app etc. I think there are a lot of over complications here, and am overall degradation of DX.
> 
> Please wait until I'm done with the examples and reasoning. I specifically took care of addressing your particular concerns.

Yup, i feel a bit stupid. Didn't see the description at all. 

This is actually very good and a big step forward. I apologize for being an idiot. 

All Looks good. 

Regarding the open questions - I'd say yes, let's expose a store property on the configs.

starlite/app.py
provinzkraut: How is `[StoreConfig(store=MemoryStore(), key="memory"), StoreConfig(store=RedisStore(), key="redis")]` more ergonomic than `{"memory": MemoryStore(), "redis": RedisStore()}`?

provinzkraut:
@Goldziher thanks for the consideration. Please excuse the snarky tone in my comments to your remarks, I was a bit taken aback by your initial comment. But no hard feelings from my side :heart: 

 Let me get through with the examples and such and then let's see where this takes us.

>  I'd say yes, let's expose a store property on the configs.

Well, I too have an opinion on that, but I'll get to that in a separate comment :grimacing:

provinzkraut:
My personal thoughts on the two open questions:

> Should users be forced to go through the registry?

Yes. If we want to facilitate the patterns explained in the above section, we should make this non-optional. There is little to gain from adding a third way to define stores for an integration and it will only make things more confusing an complex. The current solution allows to express every configuration a user wishes. If they absolutely need to bypass the registry, they can still customize the configuration objects.

> Should the registry include a "default cache" property?

I don't actually think I have anything to add there :upside_down_face:

docs/lib/reference/stores/memory.rst
```diff
@@ -0,0 +1,5 @@
+memory
```
Goldziher, line 1: Why do we use small letters here and not title case? I noticed the references r like this

starlite/config/app.py
```diff
@@ -174,6 +176,12 @@ class AppConfig:
     """
     static_files_config: list[StaticFilesConfig] = field(default_factory=list)
     """An instance or list of :class:`StaticFilesConfig <.static_files.StaticFilesConfig>`."""
+    stores: StoreRegistry | dict[str, Store] | None = None
+    """Central registry of :class:`Store <.stores.base.Store>` to be made available and be used throughout the
```
Goldziher, line 180:
```suggestion
    """Central registry for :class:`Store <.stores.base.Store>` instances that is  available throughout the
```

starlite/config/app.py
```diff
@@ -174,6 +176,12 @@ class AppConfig:
...
+    application. Can be either dictionary mapping strings to :class:`Store <.stores.base.Store>` to be passed to a
```
Goldziher, line 181:
```suggestion
    application. Can be either a dictionary mapping strings to :class:`Store <.stores.base.Store>` instances 
```

starlite/config/app.py
```diff
@@ -174,6 +176,12 @@ class AppConfig:
...
+    :class:`StoreRegistry <.stores.registry.StoreRegistry>` or an instance of
```
Goldziher, line 182:
```suggestion
, or an instance of
```

starlite/config/response_cache.py
```diff
@@ -31,31 +29,20 @@ def default_cache_key_builder(request: Request[Any, Any, Any]) -> str:
...
-        """
-        return Cache(
-            backend=self.backend or MemoryStorage(),
-            default_expiration=self.expiration,
-            cache_key_builder=self.cache_key_builder,
-        )
+    key_builder: CacheKeyBuilder = field(default=default_cache_key_builder)
+    """:class:`CacheKeyBuilder <.types.CacheKeyBuilder>`. Defaults to :func:`default_cache_key_builder`."""
+    store: str = "request_cache"
+    """Name of the :class:`Store <.stores.base.Store>` to use"""
```
Goldziher, line 44:
```suggestion
    """Name of the :class:`Store <.stores.base.Store>` to use."""
```

starlite/config/response_cache.py
```diff
@@ -31,31 +29,20 @@ def default_cache_key_builder(request: Request[Any, Any, Any]) -> str:
...
+
+    def get_store_from_app(self, app: Starlite) -> Store:
+        """Get the store defined in :attr:`store` from an :class:`Starlite <.app.Starlite>` instance"""
```
Goldziher, line 47:
```suggestion
        """Get the store defined in :attr:`store` from an :class:`Starlite <.app.Starlite>` instance."""
```

starlite/middleware/rate_limit.py
```diff
@@ -266,3 +268,7 @@ def my_handler(request: Request) -> None:
             config kwarg value.
         """
         return DefineMiddleware(self.middleware_class, config=self)
+
+    def get_store_from_app(self, app: Starlite) -> Store:
+        """Get the store defined in :attr:`store` from an :class:`Starlite <.app.Starlite>` instance"""
```
Goldziher, line 273:
```suggestion
        """Get the store defined in :attr:`store` from an :class:`Starlite <.app.Starlite>` instance."""
```

starlite/stores/base.py
```diff
@@ -73,6 +77,21 @@ async def expires_in(self, key: str) -> int | None:
...
+class NamespacedStore(Store):
+    """A subclass of :class:`Store`, offering hierarchical namespacing.
+
+    Bulk actions on a parent namespace should affect all child namespaces, whereas other operations on all namespaces
+    should be isolated.
+    """
+
+    @abstractmethod
+    def with_namespace(self, namespace: str) -> Self:
+        """Return a new instance of :class:`NamespacedStore`, which exist in a child namespace of the current namespace.
```
Goldziher, line 89:
```suggestion
        """Return a new instance of :class:`NamespacedStore`, which exists in a child namespace of the current namespace.
```

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+
+    __slots__ = ("_stores", "_default_factory")
+
+    def __init__(
+        self, stores: dict[str, Store] | None = None, default_factory: Callable[[str], Store] = default_default_factory
+    ) -> None:
+        """Initialize ``StoreRegistry``.
+
+        Args:
+            stores: A dictionary of store names to stores used to initialize the registry
```
Goldziher, line 29:
```suggestion
            stores: A dictionary mapping store names to stores, used to initialize the registry
```

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+            default_factory: A callable use by :meth:`StoreRegistry.get` to provide a store if the requested name hasn't
```
Goldziher, line 30:
```suggestion
            default_factory: A callable used by :meth:`StoreRegistry.get` to provide a store, if the requested name hasn't
```

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+                been registered yet. This callable receives the requested name and should return a
+                :class:`Store <.base.Store>`
```
Goldziher, line 32:
```suggestion
                :class:`Store <.base.Store>` instance.
```

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+        self._stores = stores or {}
+        self._default_factory = default_factory
+
+    def register(self, name: str, store: Store, override: bool = False) -> None:
+        """Register a new :class:`Store <.base.Store>`.
+
+        Args:
+            name: Name to register the store under
+            store: The store to register
+            override: Whether to allow overriding an existing store of the same name
```
Goldziher, line 43: I would rename to `allow_override` or `override_namespace`

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Raises:
+            ValueError: If a store is already registered under this name and ``override`` is not ``True``
```
Goldziher, line 46: Why not ImproperlyConfiguredException?

starlite/stores/registry.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        if not override and name in self._stores:
+            raise ValueError(f"Store with the name {name!r} already exists")
```
Goldziher, line 49: See prev

Goldziher (approved):

docs/lib/reference/stores/memory.rst
provinzkraut: Because these names always reflect the module name verbatim

starlite/config/response_cache.py
provinzkraut: Didn't we have a linting rule in place for this?

starlite/stores/registry.py
provinzkraut: Because it's not really something to do with configuration. It's strictly a runtime thing with dynamic values

starlite/stores/registry.py
provinzkraut: Good idea

starlite/config/app.py
provinzkraut: Changed it to be less convoluted.

peterschutt (approved):

PR #1345, opened 2023-03-18
feat: repository enhancements and MySQL SQLAlchemy support

cofin:
This PR adds the following additional features:

- adds support and tests for MySQL using the `asyncmy` drivers.
- Basic repository docs.  A more comprehensive usage guide will be completed when the `Service` and `DTO` patterns have been merged
- `get_or_create` optionally accepts a `match_fields` argument.  This lets you lookup a model using a subset of the `kwargs` you've provided.  If the remaining `kwargs` are different from the retrieved model's stored values, an update is performed.
- Adds `OrderBy` and `SearchFilter` filter types.  `OrderBy` appends the `ORDER BY` clause to a statement and `SearchFilter` adds ` WHERE LIKE ..` clause.  It optionally has a case insensitive override.
- Renamed `base_select` to `statement`.
 
# PR Checklist


provinzkraut:
Since these database backed tests are quite slow, can we maybe add a pytest marker to skip them?

Goldziher (approved):
can we run the DB tests only on main? or maybe using a schedule (github action cron basically).

provinzkraut:
> can we run the DB tests only on main? or maybe using a schedule (github action cron basically).

Hmm. I think we generally do want to run those in our CI and locally (maybe optionally). It's not like they take ages to run (the asyncpg tests take 5 seconds once the container is set up). That's not too bad on its own, but if we start adding many other dbs, it will somewhat add up. For me, just having a marker to be able to skip them if I want would be good enough. For the CI, it definitely doesn't matter since we have steps that take significantly longer than our whole test suite.

Goldziher:
> > can we run the DB tests only on main? or maybe using a schedule (github action cron basically).
> 
> Hmm. I think we generally do want to run those in our CI and locally (maybe optionally). It's not like they take ages to run (the asyncpg tests take 5 seconds once the container is set up). That's not too bad on its own, but if we start adding many other dbs, it will somewhat add up. For me, just having a marker to be able to skip them if I want would be good enough. For the CI, it definitely doesn't matter since we have steps that take significantly longer than our whole test suite.

ok

provinzkraut:
@cofin is this currently blocked by something?

provinzkraut:
> No, it's not blocked by anything, and I need to wrap this up. Based on the last status, I need to add the Pytest marker and merge the latest changes.

Okay, cool, just checking in!

cofin:
I've been having issues with the `asyncmy` library, and it looks like there's an existing issue that references the problem: https://github.com/long2ice/asyncmy/issues/39

cofin:
I've excluded these mysql tests by default.  I can de the same with the `asyncpg` tests if that makes sense?

peterschutt:
> I've excluded these mysql tests by default.  I can de the same with the `asyncpg` tests if that makes sense?

I.e., exclude from the general CI tests?

cofin:
> > I've excluded these mysql tests by default.  I can de the same with the `asyncpg` tests if that makes sense?
> 
> I.e., exclude from the general CI tests?

Correct, i've added `-m 'not sqlalchemy_asyncmy'` to the pytest arguments in `pyproject.toml`.

We may need to actually remove the `asyncmy` support if we can't get a resolution to the above issue though.

cofin:
So, it looks like we have some issues with the mysql driver that may take a bit to resolve.  I'd like to merge this sooner rather than later.  I've left these mysql tests disabled and added the following features:

- Basic repository docs.  A more comprehensive usage guide will be completed when the `Service` and `DTO` patterns have been merged
- `get_or_create` optionally accepts a `match_fields` argument.  This lets you lookup a model using a subset of the `kwargs` you've provided.  If the remaining `kwargs` are different from the retrieved model's stored values, an update is performed.
- Renamed `base_select` to `statement`.

provinzkraut:
>  If the remaining kwargs are different from the retrieved model's stored values, an update is performed.

I'm not so sure about this. Sounds like a rather unexpected side effect of a `get_or_create` method to also perform updates. Maybe this could be its own method with a distinct name?

litestar/contrib/sqlalchemy/repository.py
```diff
@@ -230,26 +232,44 @@ async def get_one_or_none(self, **kwargs: Any) -> ModelT | None:
...
-            statement = kwargs.pop("base_select", self.statement)
+            statement = kwargs.pop("statement", self.statement)
             statement = self._filter_select_by_kwargs(statement=statement, **kwargs)
             instance = (await self._execute(statement)).scalar_one_or_none()
             if instance:
                 self.session.expunge(instance)
             return instance  # type: ignore
 
-    async def get_or_create(self, **kwargs: Any) -> tuple[ModelT, bool]:
+    async def get_or_create(self, match_fields: list[str] | str | None = None, **kwargs: Any) -> tuple[ModelT, bool]:
```
provinzkraut, line 242: After thinking about this, wouldn't this actually just be an upsert that also returns data? Maybe `upsert_returning` or something like that would be a good name.

docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
```diff
@@ -0,0 +1,206 @@
+from __future__ import annotations
```
provinzkraut, line 1: Should use 3.8 compatible annotations in the docs

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
+SQLAlchemy Models & Repository
+==============================
+
+Litestar comes with a built-in repository class (`SQLAlchemyAsyncRepository`) for `SQLAlchemy <https://docs.sqlalchemy.org/>`_ to make CRUD operations easier.
```
provinzkraut, line 4:
```suggestion
Litestar comes with a built-in repository class (:class:`SQLAlchemyAsyncRepository <litestar.contrib.sqlalchemy.repository.SQLAlchemyAsyncRepository>`) for `SQLAlchemy <https://docs.sqlalchemy.org/>`_ to make CRUD operations easier.
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Features
+--------
+
+* Pre-configured `DeclarativeBase` for `SQLAlchemy <https://docs.sqlalchemy.org/>`_ 2.0 that includes a UUID based primary-key and an optional version with audit columns.
```
provinzkraut, line 9:
```suggestion
* Pre-configured ``DeclarativeBase`` for `SQLAlchemy <https://docs.sqlalchemy.org/>`_ 2.0 that includes a UUID based primary-key and an optional version with audit columns.
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+* Generic asynchronous repository for select, insert, update, and delete operations on SQLAlchemy models
+* Implements optimized methods for bulk inserts, updates, and deletes.
+* Integrated counts, pagination, filtering with `LIKE`, `IN`, and dates before and/or after.
```
provinzkraut, line 12:
```suggestion
* Integrated counts, pagination, filtering with ``LIKE``, ``IN``, and dates before and/or after.
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+* Support for SQLite via `aiosqlite` and Postgres via `asyncpg`
```
provinzkraut, line 13:
```suggestion
* Support for SQLite via `aiosqlite` and Postgres via `asyncpg`
```

Should probably link either the SQLA dialects or libraries here

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Basic Use
+---------
+
+To use the `SQLAlchemyAsyncRepository` repository, you must first define your models using one of the included built-in `DeclarativeBase` ORM base implementations  (`Base` and `AuditBase`).  Both include a UUID based primary key and `AuditBase` includes an `updated` and `created` timestamp column.
```
provinzkraut, line 18:
```suggestion
To use the ``SQLAlchemyAsyncRepository`` repository, you must first define your models using one of the included built-in ``DeclarativeBase`` ORM base implementations  (:class:`Base <litestar.contrib.sqlalchemy.repository.Base>` and :class:`AuditBase <litestar.contrib.sqlalchemy.repository.AuditBase>`).  Both include a UUID based primary key and ``AuditBase`` includes an ``updated`` and ``created`` timestamp column.
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+* Pydantic BaseModel and Dict classes map to JSONB SQLAlchemy type map
+
+.. literalinclude:: /examples/contrib/sqlalchemy/sqlalchemy_declarative_models.py
+    :caption: sqlalchemy_declarative_models.py
+    :language: python
+
+Basic Controller Integration.
+-----------------------------
+
+Once you have declared your models, you are ready to use the `SQLAlchemyAsyncRepository` class with your controllers and function based routes.
```
provinzkraut, line 31:
```suggestion
Once you have declared your models, you are ready to use the ``SQLAlchemyAsyncRepository`` class with your controllers and function based routes.
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Models using these bases also include the following enhancements:
+* Auto-generated snake-case table name from class name
+* Pydantic BaseModel and Dict classes map to JSONB SQLAlchemy type map
+
+.. literalinclude:: /examples/contrib/sqlalchemy/sqlalchemy_declarative_models.py
+    :caption: sqlalchemy_declarative_models.py
+    :language: python
+
+Basic Controller Integration.
```
provinzkraut, line 28:
```suggestion
Basic Controller Integration
```

docs/usage/contrib/sqlalchemy.rst
```diff
@@ -0,0 +1,35 @@
...
+* Support for SQLite via `aiosqlite` and Postgres via `asyncpg`
+
+Basic Use
+---------
+
+To use the `SQLAlchemyAsyncRepository` repository, you must first define your models using one of the included built-in `DeclarativeBase` ORM base implementations  (`Base` and `AuditBase`).  Both include a UUID based primary key and `AuditBase` includes an `updated` and `created` timestamp column.
+
+Models using these bases also include the following enhancements:
+* Auto-generated snake-case table name from class name
+* Pydantic BaseModel and Dict classes map to JSONB SQLAlchemy type map
```
provinzkraut, line 22:
```suggestion
* Pydantic BaseModel and Dict classes map to :class:`JSONB <sqlalchemy.dialects.postgresql.JSONB>` SQLAlchemy type map
```

Not sure if that's the correct path for the reference

litestar/contrib/repository/testing/generic_mock_repository.py
```diff
@@ -179,18 +180,35 @@ async def get(self, item_id: Any, **kwargs: Any) -> ModelT:
...
-        existing = await self.get_one_or_none(**kwargs)
+        match_fields = match_fields if match_fields else self.match_fields
+        if isinstance(match_fields, str):
+            match_fields = [match_fields]
+        if match_fields:
+            match_filter = {
+                field_name: kwargs.get(field_name, None)
+                for field_name in match_fields
+                if kwargs.get(field_name, None) is not None
+            }
```
provinzkraut, line 202:
```suggestion
            match_filter = {
                field_name: field_value
                for field_name in match_fields
                if (field_value := kwargs.get(field_name)) is not None
            }
```

litestar/contrib/sqlalchemy/repository.py
```diff
@@ -59,23 +61,23 @@ def wrap_sqlalchemy_exception() -> Any:
...
+            statement: To facilitate customization of the underlying select query.
             session: Session managing the unit-of-work for the operation.
-            base_select: To facilitate customization of the underlying select query.
             **kwargs: Additional arguments.
 
         """
         super().__init__(**kwargs)
         self.session = session
-        self.statement = base_select if base_select is not None else select(self.model_type)
+        self.statement = statement if statement is not None else select(self.model_type)
```
provinzkraut, line 80:
```suggestion
        self.statement = statement or select(self.model_type)
```

I don't know if there's a condition under which a `Select` could ever be considered falsey. If there is, my suggestion won't work :grimacing:

pyproject.toml
```diff
@@ -199,12 +203,16 @@ exclude_lines = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "--ignore=examples"
+addopts = "--ignore=examples -m='not sqlalchemy_asyncmy'"
```
provinzkraut, line 206: This should then be explicitly selected in the CI and Makefile `test` command

provinzkraut (changes requested):

litestar/contrib/sqlalchemy/repository.py
cofin: I had it this way before but ran into a few errors, i think.  I'll change it back and do some testing.

docs/usage/contrib/sqlalchemy.rst
cofin: I think it should be :class: `sqlalchemy.types.JSON` this is the generic implementation that automatically selects the dialect specific version.

provinzkraut: Afaik for SQLA itself, `JSON` does not render as `JSONB` for postgres since postgres has both. Does the repo do that? If so, that should probably be mentioned (and is a sensible default).

cofin: well, it looks like you are right.  I was thinking that it would use the `JSONB` implementation for Postgres.  Maybe I should add a custom SQLAlchemy JSON type that handles this for dialects that support JSONB.

litestar/contrib/sqlalchemy/repository.py
provinzkraut: `upsert_returning` definitely isn’t ideal. `get_or_upsert` sounds good,  it actually, it could also just be `upsert` since it’s kinda just an upsert that also returns things, right?

cofin: We've already got an `upsert` that has a slightly different set of parameters.  Should it just replace that?

peterschutt: > wouldn't this actually just be an upsert

I think the kwargs and `match_fields` behavior make it different to the upsert method. Where `upsert()` receives a model instance, and purely matches on identity, this receives and matches on kwargs, and matches on all kwargs by default. Deserves a distinct method IMO and I'm OK with `get_or_create()`.

provinzkraut: > I think the kwargs and match_fields behavior make it different to the upsert method

Hmm, that's true. But from the way it works, it's still a type of upsert (at least that's what it feels like to me). I do get your point though, and @cofin, I don't think this should replace the already existing `upsert` method. 

How about `match_or_create`, `match_or_upsert`, `match_upsert` or something like that? I just think there ought to be something in the name that indicates this method does not only get or create, but also updates things.

tests/contrib/sqlalchemy/repository/test_sqlalchemy_asyncmy.py
```diff
@@ -0,0 +1,501 @@
...
+    """
+    conn = asyncmy.connect(host=host, port=3360, user="app", database="db", password="super-secret", echo=True)
+    conn = await conn.connect()
+    async with conn.cursor() as cursor:
+        await cursor.execute("select 1 as is_available")
+        resp = await cursor.fetchone()
+    return bool(resp[0] == 1)
+
+
+async def _containers(docker_ip: str, docker_services: Services) -> None:  # pylint: disable=unused-argument
```
provinzkraut, line 95:
```suggestion
@pytest.mark.sqlalchemy_asyncmy
@pytest.fixture(scope="session", autouse=True)
async def _containers(docker_ip: str, docker_services: Services) -> None:  # pylint: disable=unused-argument
```

tests/contrib/sqlalchemy/repository/test_sqlalchemy_asyncmy.py
```diff
@@ -0,0 +1,501 @@
...
+async def db_responsive(host: str) -> bool:
+    """
+    Args:
+        host: docker IP address.
+
+    Returns:
+        Boolean indicating if we can connect to the database.
+    """
+    conn = asyncmy.connect(host=host, port=3360, user="app", database="db", password="super-secret", echo=True)
+    conn = await conn.connect()
```
provinzkraut, line 88:
```suggestion
    conn = await asyncmy.connect(host=host, port=3360, user="app", database="db", password="super-secret", echo=True)
```

.github/workflows/test.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
         if: ${{ !inputs.coverage }}
         run: |
           source $VENV
-          pytest
+          pytest -m='not sqlalchemy_asyncpg'
```
provinzkraut, line 60: Should be the opposite, right? These are already disabled in `adopts` in `pyproject.toml`, so we'd want to explicitly run those here as well I think.

litestar/contrib/sqlalchemy/repository.py
cofin: The `match` part of this function is completely optional, I should say.  If you omit it, it functions exactly as it used to.  With that in mind, do we still want to rename?

provinzkraut: > With that in mind, do we still want to rename?

Well.. Personally I still think that it should somehow indicate that it can update things. But I can't really think of a succinct way to phrase that which does not cause confusion with other methods.

In the interest of not blocking this because of a single method name I'd say just leave it as it is :grimacing:

.github/workflows/test.yaml
```diff
@@ -58,6 +58,11 @@ jobs:
         run: |
           source $VENV
           pytest
+      - name: Test Database Backends
+        if: ${{ !inputs.coverage }}
```
provinzkraut, line 62: Do we not want coverage for those? If not, I think we need to exclude that in the config as well, since otherwise it will be counted as missing.

docs/conf.py
```diff
@@ -70,12 +70,31 @@
...
     ("py:class", "sqlalchemy.dialects.postgresql.named_types.ENUM"),
+    ("py:class", "_orm.Mapper"),
+    ("py:class", "_orm.registry"),
+    ("py:class", "_schema.MetaData"),
+    ("py:class", "_schema.Table"),
+    ("py:class", "_RegistryType"),
+    ("py:class", "abc.Collection"),
+    ("py:class", "TypeEngine"),
+    ("py:class", "ExternalType"),
+    ("py:class", "UserDefinedType"),
```
provinzkraut, line 81: These should definitely work. What was the issue with them?

docs/conf.py
```diff
@@ -70,12 +70,31 @@
...
     ("py:class", "RouteHandlerType"),
     ("py:obj", "litestar.security.base.AuthType"),
     ("py:class", "ControllerRouterHandler"),
     ("py:class", "PathParameterDefinition"),
     ("py:class", "BaseSessionBackendT"),
+    ("py:class", "litestar.contrib.repository.abc.CollectionT"),
+    ("py:obj", "ModelT"),
+    ("py:obj", "litestar.contrib.sqlalchemy.repository.ModelT"),
+    ("py:class", "litestar.contrib.sqlalchemy.repository.ModelT"),
+    ("py:class", "ModelT"),
```
provinzkraut, line 96: Instead of having both `py:class` and `py:obj` for `ModelT`, you can just use a regex in `nitpick_ignore_regex`: `("py:.*":, "ModelT")`

provinzkraut (approved):

docs/conf.py
cofin: Yep - i'll update this morning.

docs/conf.py
cofin: I got the following error when these weren't included:
```
/home/oracle/Code/Starlite/litestar/litestar/contrib/repository/filters.py:docstring of litestar.contrib.repository.filters.CollectionFilter.values:1: WARNING: py:class reference target not found: abc.Collection
/home/oracle/Code/Starlite/litestar/litestar/contrib/repository/filters.py:docstring of litestar.contrib.repository.filters.CollectionFilter.__init__:1: WARNING: py:class reference target not found: abc.Collection
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _schema.MetaData
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _schema.Table
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base.AuditBase.registry:1: WARNING: py:class reference target not found: _RegistryType
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _orm.registry
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _orm.Mapper
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _schema.MetaData
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _schema.Table
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base.Base.registry:1: WARNING: py:class reference target not found: _RegistryType
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _orm.registry
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/base.py:docstring of litestar.contrib.sqlalchemy.base:1: WARNING: py:class reference target not found: _orm.Mapper
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/types.py:docstring of litestar.contrib.sqlalchemy.types:1: WARNING: py:class reference target not found: ExternalType
/home/oracle/Code/Starlite/litestar/docstring of litestar.contrib.sqlalchemy.types.GUID.cache_ok:100: WARNING: py:class reference target not found: ExternalType
/home/oracle/Code/Starlite/litestar/docstring of litestar.contrib.sqlalchemy.types.GUID.cache_ok:100: WARNING: py:class reference target not found: UserDefinedType
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/types.py:docstring of litestar.contrib.sqlalchemy.types:1: WARNING: py:class reference target not found: TypeEngine
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/types.py:docstring of litestar.contrib.sqlalchemy.types:1: WARNING: py:meth reference target not found: type_engine
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/types.py:docstring of litestar.contrib.sqlalchemy.types:1: WARNING: py:class reference target not found: _types.TypeDecorator
/home/oracle/Code/Starlite/litestar/.venv/lib/python3.11/site-packages/sqlalchemy/sql/sqltypes.py:docstring of sqlalchemy.sql.sqltypes.CHAR:1: WARNING: py:meth reference target not found: _types.TypeDecorator.process_result_value
/home/oracle/Code/Starlite/litestar/litestar/contrib/sqlalchemy/types.py:docstring of litestar.contrib.sqlalchemy.types:1: WARNING: py:class reference target not found: _types.TypeDecorator
/home/oracle/Code/Starlite/litestar/.venv/lib/python3.11/site-packages/sqlalchemy/sql/sqltypes.py:docstring of sqlalchemy.sql.sqltypes.CHAR:1: WARNING: py:meth reference target not found: _types.TypeDecorator.process_bind_param
```

.github/workflows/test.yaml
cofin: Should I remove the whole entry or just the `if` clause?

docs/conf.py
provinzkraut: Hm. I need to look into that. Those should resolve, either natively or with our custom extension or be silenced by our custom extension. Let's just keep them on the suppression list for now.

.github/workflows/test.yaml
provinzkraut: Hmm. I think ideally all tests in the CI would include the database tests. You should be able to override the ini "adopts" option somewhere but I'm not 100% sure what the config for that would look.

cofin: I think the updated configuration accomplishes all of the above.

provinzkraut (approved):

JacobCoffee (approved):

PR #1346, opened 2023-03-18
update simple event emitter to use worker pattern

Goldziher:
# PR Checklist

This PR updates the `EventEmitterBackend` abstract class and the `SimpleEventEmitter` concrete implementation to be more useful. I used the `Broadcaster` library from encode as a source of inspiration for the implementation, which uses the worker pattern. It appears to be working with trio, without an issue.

### DIFF
--- a/starlite/events/emitter.py
+++ b/starlite/events/emitter.py
@@ -104,7 +104,7 @@ class SimpleEventEmitter(BaseEventEmitterBackend):
         """
         if sniffio.current_async_library() != "asyncio":
             return
-            
+
         self._queue = Queue()
         self._worker_task = create_task(self._worker())
 
--- a/starlite/events/emitter.py
+++ b/starlite/events/emitter.py
@@ -104,6 +104,7 @@ class SimpleEventEmitter(BaseEventEmitterBackend):
         """
         if sniffio.current_async_library() != "asyncio":
             return
+            
         self._queue = Queue()
         self._worker_task = create_task(self._worker())
 
@@ -138,7 +139,7 @@ class SimpleEventEmitter(BaseEventEmitterBackend):
         """
         if not (self._worker_task and self._queue):
             if sniffio.current_async_library() != "asyncio":
-                raise ImproperlyConfiguredException("This backend only supports asyncio")
+                raise ImproperlyConfiguredException("{type(self).__name__} only supports 'asyncio' based event loops")
 
             raise ImproperlyConfiguredException("Worker not running")
 
--- a/starlite/events/emitter.py
+++ b/starlite/events/emitter.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from asyncio import Queue, Task, create_task, gather
+from asyncio import CancelledError, Queue, Task, create_task
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Coroutine, DefaultDict, Sequence, cast
+from contextlib import suppress
+from typing import TYPE_CHECKING, Any, DefaultDict, Sequence
 
 from starlite.exceptions import ImproperlyConfiguredException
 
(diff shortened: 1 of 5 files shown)


starlite/events/emitter.py
```diff
@@ -49,10 +46,77 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:  # pragm
...
+
+        Args:
+            listeners: A list of listeners.
+        """
+        super().__init__(listeners=listeners)
+        self._queue = None
+        self._worker_task = None
+
+    async def _worker(self) -> None:
+        """Worker that runs in a separate thread and continuously pulls events from asyncio queue.
```
provinzkraut, line 87:
```suggestion
        """Worker that runs in a separate task and continuously pulls events from asyncio queue.
```

it doesn't actually run in a separate thread

starlite/events/emitter.py
```diff
@@ -49,10 +46,77 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:  # pragm
...
+
+        Returns:
+            None
+        """
+        while self._queue:
```
provinzkraut, line 92: Should this not be `while True`?

starlite/events/emitter.py
```diff
@@ -49,10 +46,77 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:  # pragm
...
+            coroutines = cast("tuple[Coroutine[Any, Any, Any], ...]", await self._queue.get())
+            await gather(*coroutines)
+
```
provinzkraut, line 95: This is missing a `task_done()` call, since we need to `join` the queue on shutdown

starlite/events/emitter.py
```diff
@@ -49,10 +46,77 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:  # pragm
...
+    async def on_shutdown(self) -> None:
+        """Hook called on application shutdown, used to perform cleanup.
+
+        Returns:
+            None
+        """
+        if self._worker_task:
+            self._worker_task.cancel()
+
+        self._queue = None
```
provinzkraut, line 117:
```suggestion
        await self._queue.join()
        self._queue = None
```

starlite/events/emitter.py
```diff
@@ -49,10 +46,77 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:  # pragm
...
+            pass
+
+    async def on_shutdown(self) -> None:
+        """Hook called on application shutdown, used to perform cleanup.
+
+        Returns:
+            None
+        """
+        if self._worker_task:
+            self._worker_task.cancel()
```
provinzkraut, line 115:
```suggestion
            self._worker_task.cancel()
            await self._worker_task
```

ensure the successful cancellation of the task

provinzkraut (changes requested):

starlite/events/emitter.py
```diff
@@ -6,6 +6,8 @@
 from contextlib import suppress
 from typing import TYPE_CHECKING, Any, DefaultDict, Sequence
 
+import sniffio
```
Goldziher, line 9: i think this can go, no?

starlite/events/emitter.py
```diff
@@ -100,6 +102,9 @@ async def on_startup(self) -> None:
         Returns:
             None
         """
+        if sniffio.current_async_library() != "asyncio":
+            return
```
Goldziher, line 106:
```suggestion
            return
            
```

starlite/events/emitter.py
```diff
@@ -129,6 +136,12 @@ async def emit(self, event_id: str, *args: Any, **kwargs: Any) -> None:
         Returns:
             None
         """
+        if not (self._worker_task and self._queue):
+            if sniffio.current_async_library() != "asyncio":
+                raise ImproperlyConfiguredException("This backend only supports asyncio")
```
Goldziher, line 141:
```suggestion
                raise ImproperlyConfiguredException("{type(self).__name__} only supports 'asyncio' based event loops")
```

Issue #1234, opened 2023-02-22
StaticFilesConfig and virtual directories


NomAnor:
I'm trying to write a ``FileSystemProtocol`` to load files from the package data using [importlib_resources](https://importlib-resources.readthedocs.io/en/latest/using.html#). But because ``directories`` is defined as ``DirectoryPath``, pydantic checks if the given directories exist in the local filesystem. 

This is not generally true, especially in any kind of virtual filesystem (e.g. a zipped package). I think this condition should be relaxed to support virtual filesystems.

https://github.com/starlite-api/starlite/blob/9bb6dcd57c10a591377cf8e3a537e9292566d5b9/starlite/config/static_files.py#L32

Goldziher:
I agree. Do you want to submit a PR?

PR #1374, opened 2023-03-23
Adds `signature_namespace` layered parameter.

peterschutt:
When we parse the signature models for handlers, names of the parameter types must be available. However, when using linters, names that are only referenced in type annotations will be flagged as being unnecessary at runtime and suggested to move into an `if TYPE_CHECKING` block.

This parameter will allow names to be declared outside the scope that handlers are defined, allowing modules defining handlers and dependencies to better integrate with tooling.

For example:

```py
# router.py
from starlite import Router

from domain.thing import Thing
from .controller import ThingController

router = Router("/", route_handlers=[ThingController], signature_namespace={"Thing": Thing})
```

```py
# controller.py

from starlite import Controller, post

if TYPE_CHECKING:
    from domain.thing import Thing

class ThingController(Controller):
    @post()
    def create_thing(data: Thing) -> Thing:
        return data
```

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/starlite/handlers/base.py
+++ b/starlite/handlers/base.py
@@ -230,8 +230,8 @@ class BaseRouteHandler(Generic[T]):
     def resolve_opts(self) -> None:
         """Build the route handler opt dictionary by going from top to bottom.
 
-        If multiple layers define the same key, the value from the closest layer to the response handler will take
-        precedence.
+        When merging keys from multiple layers, if the same key is defined by multiple layers, the value from the
+        layer closest to the response handler will take precedence.
         """
 
         opt: dict[str, Any] = {}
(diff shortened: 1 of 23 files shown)


starlite/handlers/base.py
```diff
@@ -234,6 +240,20 @@ def resolve_opts(self) -> None:
 
         self.opt = opt
 
+    def resolve_signature_namespace(self) -> dict[str, Any]:
+        """Build the route handler signature namespace dictionary by going from top to bottom.
+
+        Names from layers are merged, if multiple layers define the same key, the value from the closest layer to the
+        response handler takes precedence.
```
JacobCoffee, line 247:
```suggestion
        When merging names from multiple layers, if the same key is defined by multiple layers, the value from the
        layer closest to the response handler will takes precedence.
```

JacobCoffee (approved):

starlite/handlers/base.py
JacobCoffee: Just a suggestion, if it makes sense. It hurt my head reading the original but I am probably dumb 😛

peterschutt: Will absolutely update this. I copied the docstring from the `resolve_opts()` method on the same class and modified it a bit, so I'll change that one too.

Goldziher (approved):

PR #1381, opened 2023-03-25
RFC: Remove `InitPluginProtocol` in favor of `on_app_init` hook.

peterschutt:
We have two methods for automated app configuration, `on_app_init` hook, and `InitPluginProtocol`. This PR attempts to simplify this by refactoring plugins to utilise the `on_app_init` hook, and therefore removing the `InitPluginProtocol`.

In this PR, passing a plugin instance to `on_app_init` both adds the plugin to `AppConfig.plugins` and performs any application configuration, such as that which is done by the SQLAlchemy plugin.

The `on_app_init` approach for pre-baked app config is desirable due to it interacting with the `AppConfig` object. Plugin development is an area that we should encourage the community to participate in, and encouraging that work to interact with the `AppConfig` _before_ the majority of application configuration has taken place is more desirable than passing through an application object to the plugin at some arbitrary point inside the app's own `__init__()` method.

Additionally, the app configuration is something that may be desired independent of a serialization plugin. So it makes sense for that functionality to be defined separate from the serialization plugin, and leveraged by it. For example, if a user doesn't wish to leverage the serialization plugin functionality, they can use `on_app_init=[SQLAlchemyInit()]`. However if they want the serialization behavior _and_ the app initialization, should that be `on_app_init=[SQLAlchemyPlugin()]` or `plugins=[SQLAlchemyPlugin()]`, or `on_app_init=[SQLAlchemyInit()], plugins=[SQLAlchemyPlugin()]`?

In this PR, the `plugins` kwarg to app remains, so there is still two ways for a plugin to be registered, e.g., by either passing `on_app_init=[plugin_instance]` or `plugins=[plugin_instance]`. This isn't ideal, and definitely a downside of this approach.

Discussion points:
- is having plugin instances register themselves via `on_app_init` intuitive enough?
- does app configuration _have_ to be associated with serialization plugin behavior?
- an alternative would be to redefine `InitPluginProtocol.on_app_init()` to receive the app config object and call these immediately after `on_app_init` is called.
- other??

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_async.py
+++ b/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_async.py
@@ -64,5 +64,5 @@ async def get_company(company_id: int, async_session: AsyncSession) -> Company:
 app = Starlite(
     route_handlers=[create_company, get_company],
     on_startup=[on_startup],
-    plugins=[sqlalchemy_plugin],
+    on_app_init=[sqlalchemy_plugin],
 )
--- a/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_relationships.py
+++ b/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_relationships.py
@@ -10,8 +10,6 @@ from starlite.exceptions import HTTPException
 from starlite.status_codes import HTTP_404_NOT_FOUND
 
 engine = create_engine("sqlite+pysqlite://")
-sqlalchemy_config = SQLAlchemyConfig(engine_instance=engine, use_async_engine=False)
-sqlalchemy_plugin = SQLAlchemyPlugin(config=sqlalchemy_config)
 
 Base = declarative_base()
 
@@ -52,8 +50,10 @@ def get_user(user_id: int, db_session: Session) -> User:
     return user
 
 
+sqlalchemy_config = SQLAlchemyConfig(engine_instance=engine, use_async_engine=False)
+
 app = Starlite(
     route_handlers=[get_user],
     on_startup=[on_startup],
-    plugins=[sqlalchemy_plugin],
+    on_app_init=[SQLAlchemyPlugin(config=sqlalchemy_config)],
 )
--- a/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_relationships_to_many.py
+++ b/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_relationships_to_many.py
@@ -11,8 +11,6 @@ from starlite.exceptions import HTTPException
 from starlite.status_codes import HTTP_404_NOT_FOUND
 
 engine = create_engine("sqlite+pysqlite:///test.sqlite")
-sqlalchemy_config = SQLAlchemyConfig(engine_instance=engine, use_async_engine=False)
-sqlalchemy_plugin = SQLAlchemyPlugin(config=sqlalchemy_config)
 
 Base = declarative_base()
 
@@ -70,8 +68,10 @@ def get_user(user_id: int, db_session: Session) -> UserModel:
     return UserModel.from_orm(user)
 
 
+sqlalchemy_config = SQLAlchemyConfig(engine_instance=engine, use_async_engine=False)
+
 app = Starlite(
     route_handlers=[get_user],
     on_startup=[on_startup],
-    plugins=[sqlalchemy_plugin],
+    on_app_init=[SQLAlchemyPlugin(config=sqlalchemy_config)],
 )
--- a/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_sync.py
+++ b/docs/examples/plugins/sqlalchemy_plugin/sqlalchemy_sync.py
@@ -59,5 +59,5 @@ def get_company(company_id: str, db_session: Session) -> Company:
 app = Starlite(
     route_handlers=[create_company, get_company],
     on_startup=[on_startup],
-    plugins=[sqlalchemy_plugin],
+    on_app_init=[sqlalchemy_plugin],
 )
--- a/starlite/app.py
+++ b/starlite/app.py
@@ -27,7 +27,6 @@ from starlite.middleware.cors import CORSMiddleware
 from starlite.openapi.config import OpenAPIConfig
 from starlite.openapi.spec.components import Components
 from starlite.plugins import (
-    InitPluginProtocol,
     OpenAPISchemaPluginProtocol,
     SerializationPluginProtocol,
 )
@@ -407,9 +406,6 @@ class Starlite(Router):
             tags=config.tags,
             type_encoders=config.type_encoders,
         )
-
-        for plugin in (p for p in config.plugins if isinstance(p, InitPluginProtocol)):
-            plugin.on_app_init(app=self)
 
         for route_handler in config.route_handlers:
             self.register(route_handler)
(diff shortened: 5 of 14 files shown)


peterschutt:
> Discussion points:
> 
>     is having plugin instances register themselves via `on_app_init` intuitive enough?

I don't think so, and especially with there being the `plugins` app param and `AppConfig.plugins` - it feels too messy.

>     does app configuration _have_ to be associated with serialization plugin behavior?

This is where I'm landing on this and in retrospect, I think the approach in this PR is too extreme. I think we should ask users to do this: `on_app_init=[SQLAlchemyInit()], plugins=[SQLAlchemyPlugin()]` to get the same behavior as the v1 sqlalchemy plugin.

>     an alternative would be to redefine `InitPluginProtocol.on_app_init()` to receive the app config object and call these immediately after `on_app_init` is called.

Feels too much like `on_app_init` by a different name to me.

Goldziher:
I don't agree with the necessity or premise of the PR. 

Serialization and initialization for plugins have been separated into two different protocols. There is no coupling there.

The main purpose of plugin init is library author's, who can write a single class or multiple, as required.

peterschutt:
Thanks for taking a look.

> I don't agree with the necessity or premise of the PR.

Noted. Allow me to counter that it's an RFC - purely intended to promote discussion about an area where I think we have a lack of clarity.
 
> Serialization and initialization for plugins have been separated into two different protocols. There is no coupling there.

Why do we need _both_ `InitPluginProtocol` and `on_app_init` hooks? I feel we should be encouraging library authors to interact with the `AppConfig` object via `on_app_init` rather than the application object.

> The main purpose of plugin init is library author's, who can write a single class or multiple, as required.

This PR doesn't change that fact.

Goldziher:
> Thanks for taking a look.
> 
> > I don't agree with the necessity or premise of the PR.
> 
> Noted. Allow me to counter that it's an RFC - purely intended to promote discussion about an area where I think we have a lack of clarity.
> 
> > Serialization and initialization for plugins have been separated into two different protocols. There is no coupling there.
> 
> Why do we need _both_ `InitPluginProtocol` and `on_app_init` hooks? I feel we should be encouraging library authors to interact with the `AppConfig` object via `on_app_init` rather than the application object.
> 
> > The main purpose of plugin init is library author's, who can write a single class or multiple, as required.
> 
> This PR doesn't change that fact.

peterschutt:
@Goldziher why was this closed?

provinzkraut:
I agree with @peterschutt. The current situation isn't ideal, and `InitPluginProtocol` is indeed superfluous. 

If all your plugin does is be called `on_app_init`, then you can very well just pass it to that hook.

> Why do we need both `InitPluginProtocol` and `on_app_init` hooks? I feel we should be encouraging library authors to interact with the `AppConfig` object via `on_app_init` rather than the application object.

Fully agree here. In many cases it's actually easier to deal with the config object than the app instance, and having one "init hook" receive the config object and the other an application instance is inconsistent and really not necessary. If someone *needs* the application instance, we could think about adding an `after_app_init` hook. 

> `on_app_init=[SQLAlchemyInit()], plugins=[SQLAlchemyPlugin()]`

This isn't really how I as a user would expect a plugin to work. I'd expect it to do the setup for me. A base plugin protocol that only defines the application hooks is still necessary IMO, just for convenience.

Goldziher:
> > Thanks for taking a look.
> > > I don't agree with the necessity or premise of the PR.
> > 
> > 
> > Noted. Allow me to counter that it's an RFC - purely intended to promote discussion about an area where I think we have a lack of clarity.
> > > Serialization and initialization for plugins have been separated into two different protocols. There is no coupling there.
> > 
> > 
> > Why do we need _both_ `InitPluginProtocol` and `on_app_init` hooks? I feel we should be encouraging library authors to interact with the `AppConfig` object via `on_app_init` rather than the application object.
> > > The main purpose of plugin init is library author's, who can write a single class or multiple, as required.
> > 
> > 
> > This PR doesn't change that fact.

peterschutt:
How about we add `serialization_plugins` and `openapi_plugins` as application parameters, and replace `on_app_init` with `plugins`?

A "plugin" would then be any callable that receives and returns the `AppConfig` object, and is called just before app initialization.

A SQLAlchemy plugin example:

```py
Starlite(..., plugins=[SQLAlchemyPlugin()]`)
```
and inside `SQLAlchemyPlugin.__call__()`:

```py
class SQLAlchemyPlugin:
    def __call__(self, app_config: AppConfig) -> AppConfig:
        app_config.serialization_plugins.append(self)
        ...
```

Goldziher:
How does this make things simpler or cleaner? 

I mean, we already have a huge number of kwargs. I would suggest we try to make the API surface as compact as possible and reduce the process users need to go through.

Furthermore - in the current implementation plugins can implement multiple protocols. We use instance of checks to sort them. What you're suggesting will impact this negatively, no?

peterschutt:
You're right that it probably doesn't simplify things a great deal, if at all.

I just haven't found the place where all of this feels right to me yet.

Consider some scenarios:

1. user wants serialization plugin behavior and not application configuration
2. user wants both app configuration and serialization plugin behavior
3. user wants application configuration, but not serialization plugin behavior

So lets assume we keep both `plugins` and `on_app_init`, where `plugins` is typed `OpenAPISchemaPluginProtocol | SerializationPluginProtocol` and `on_app_init` is same as current.


Scenario 1:

```py
app = Starlite(..., plugins=[SQLAlchemyPlugin()])
```

Scenario 2:

```py
app = Starlite(..., on_app_init=[SQLAlchemyPlugin()])
```
Scenario 3:

We don't have support for this scenario in 1.x, but I'm intending on this being:

```py
app = Starlite(..., on_app_init=[SQLAlchemyInit()])
```
And `SQLAlchemyPlugin` would use `SQLAlchemyInit` under the hood.

Now I imagine trying to explain this to someone.

> If you don't want the app configuration behavior, but you do want the serialization plugin behavior, pass the `SQLAlchemyPlugin` instance to the starlite `plugins` parameter. If you want both application configuration behavior, and serialization plugin behavior, pass the plugin instance to `on_app_init` and _not_ to `plugins`, this is because the plugin will register itself as a plugin when it is called with the `on_app_init` hooks. If you only want application configuration pass an instance of `SQLAlchemyInit` to `on_app_init`.

That just doesn't work at all.

Now if I consider the same scenarios where we _do_ still have `InitPluginProtocol`:

Scenario 1:

```python
app = Starlite(..., plugins=[SQLAlchemySerializationPlugin()])
```

Scenario 2:

```python
app = Starlite(..., plugins=[SQLAlchemyPlugin()])
```

Scenario 3:

```python
app = Starlite(..., plugins=[SQLAlchemyInitPlugin()])
```

Explaining it to someone:

> If you don't want the app configuration behavior, but you do want the serialization plugin behavior, pass a `SQLAlchemySerializationPlugin` instance to the starlite `plugins` parameter. If you want both application configuration behavior, and serialization plugin behavior, pass an instance of `SQLAlchemyPlugin` to `plugins`. If you only want application configuration pass an instance of `SQLAlchemyInitPlugin`.

I've come around. I think we do still want the init plugin protocol, but it should be changed to:

```python
class InitPluginProtocol(Protocol):
        ...
```

And they should be called immediately after `on_app_init` hooks have been called, and so for all intents and purposes, they _are_ `on_app_init` hooks, but the redundancy of having two ways to skin the cat is well and truly offset by simplification of the interface.

PR #1384, opened 2023-03-25
Docs: help new contributors avoid "command not found" errors

zoni:
When following the contribution guide, this will have users installing dependencies with `poetry`, meaning commands like pytest and sphinx-{build,autobuild} get installed into a Python virtualenv where they're not on the user's $PATH by default.

If a user blindly follows these steps without invoking `poetry shell` first (or wrapping with `poetry run <command>`), they'll hit "command not found" errors, potentially obscured further by the use of makefiles.

For example, somebody wanting to make a quick documentation fix might encounter:

    $ make docs-serve
    sphinx-autobuild docs docs/_build/ -j auto --watch starlite,examples
    make: sphinx-autobuild: No such file or directory
    make: *** [Makefile:7: docs-serve] Error 127

I believe a hint towards executing `poetry shell` / prefixing with `poetry run` will help avoid this issue for newcomers, without going into too much irrelevant background detail on how `$PATH` works and how Poetry manages virtualenvs under the hood.

In addition to mentioning this in the environment setup, I repeat this reminder in places where commands are listed in case people later return to the docs and jump past the environment setup into the specific section they're interested in.


provinzkraut:
This is certainly a valid point, but I'm not so certain this is the best way to address it, since there are various ways to go about this. 

If you're using PyCharm for example, you will not have to run `poetry shell` at all, since it will be picked up by the IDE and a shell will be spawned in the venv automatically. This is at least the standard configuration, and what you have to do varies wildly depending on the IDE/editor used and how it's set up.

I left instructions about this out, since it's not intended to be a guide how to set up a Python development environment, and covering all bases would far exceed the scope of this document.

Maybe adding a simple, unspecific, note to activate the venv would suffice, or prefacing the section with a hint to follow the tutorials in the links provided if not familiar with the tools being used would be better?

CONTRIBUTING.rst
```diff
@@ -10,6 +10,7 @@ Setting up the environment
 3. If you're working on the documentation and need to build it locally, install the extra dependencies with ``poetry install --with docs``
 4. Install `pre-commit <https://pre-commit.com/>`_
 5. Run ``pre-commit install`` to install pre-commit hooks
+6. Optional: Run ``poetry shell`` to activate the Python virtual environment created by poetry so that you do not have to prefix commands with ``poetry run ..`` in your current shell session
```
provinzkraut, line 13:
```suggestion
6. Optional: If not done automatically by your IDE, run ``poetry shell`` to activate the virtual environment created in step 2
```

Something along those lines would also work. Wdyt?

CONTRIBUTING.rst
```diff
@@ -11,6 +11,13 @@ Setting up the environment
 4. Install `pre-commit <https://pre-commit.com/>`_
 5. Run ``pre-commit install`` to install pre-commit hooks
 
+.. tip::
```
provinzkraut, line 14: Does this render properly on GitHub?

If not, just use a subsection or bold text. It's not as pretty on our page but it should work when viewed on GitHub as well since it's linked there

zoni: For some definition of "properly" :stuck_out_tongue: You can see how this looks at https://github.com/starlite-api/starlite/blob/77c56f795b1bdcbc5d94490c22f4331923f59cc4/CONTRIBUTING.rst.

It's unstyled, but much better than say, admonitions for MkDocs would be which GitHub doesn't support at all.

Happy to change this to regular rst without the directive if you feel this still looks too badly on plain GitHub previews. Let me know what you'd prefer! :smile:

CONTRIBUTING.rst
```diff
@@ -11,6 +11,13 @@ Setting up the environment
...
+  Many modern IDEs like PyCharm or VS Code will enable the poetry-managed virtualenv that is created in step 2 for you automatically.
```
zoni, line 15: This is not a statement of endorsement for either, but I think these are the most well-known/commonly used IDEs and mentioning their names explicitly may help people who are really new to software development and might not recognize the term IDE.

CONTRIBUTING.rst
provinzkraut: Okay, it's good enough so we can keep it.

CONTRIBUTING.rst
```diff
@@ -11,6 +11,13 @@ Setting up the environment
...
+  If you're not using such an IDE then you will need to manually activate the virtualenv yourself otherwise you will encounter *"Command not found"* errors.
```
provinzkraut, line 16:
```suggestion
  If your IDE / editor does not offer this functionality, then you will need to manually activate the virtualenv yourself. Otherwise you may encounter errors or unexpected behaviour when trying to run the commands referenced within this document.
```

Making it a bit less specific.

CONTRIBUTING.rst
```diff
@@ -11,6 +11,13 @@ Setting up the environment
...
+
+  The easiest way to activate this virtualenv is by running ``poetry shell``, as described at `Using your virtual environment <https://python-poetry.org/docs/basic-usage/#using-your-virtual-environment>`_ in poetry's documentation.
```
provinzkraut, line 18:
```suggestion
  The easiest way to activate this virtualenv manually is by running ``poetry shell``, as described at `Using your virtual environment <https://python-poetry.org/docs/basic-usage/#using-your-virtual-environment>`_ in poetry's documentation.
```

and this a bit more specific :stuck_out_tongue:

provinzkraut (approved):

provinzkraut:
@all-contributors add @zoni doc

provinzkraut:
@JacobCoffee I think we need to propagate this change throughout the other repos as well.

JacobCoffee:
> If you're using PyCharm for example, you will not have to run poetry shell at all

I have to still do this for some reason, but probably a bug or something on my end 😢

PR #1402, opened 2023-03-29
Better websockets

provinzkraut:
Add enhanced websocket support, making websocket handling a lot easier and feel more "Starlite".

The main features are two new classes:

1. The `websocket_listener`: A special websocket handler, listening and sending incoming data in a loop
2. `WebsocketListener`: An abstract base class building upon `websocket_listener` and allowing to define things in a more OOP way

<hr>

## Examples

```python
from starlite.handlers.websocket_handlers import websocket_listener

@websocket_listener("/")
async def handler(data: str) -> str:
  return data
```

this is equivalent to:

```python
from starlite import websocket, WebSocket
from starlite.exceptions import WebSocketDisconnect


@websocket("/")
async def handler(socket: WebSocket) -> str:
  await socket.accept()
  while True:
    try:
      data = await socket.receive_text()
      await socket.send_text(data)
    except WebSocketDisconnect:
      break
```

It allows to send and receive either bytes, text or json, all configure by the type annotations of the handler function:

```python
from starlite.handlers.websocket_handlers import websocket_listener

@websocket_listener("/")
async def handler(data: list[str]) -> dict[str, str]:
  return {"message": data}
```

### Dependencies

Dependencies are handled on the route handler level, and passed into the listener function, so this works as expected:

```python
from starlite.di import Provide
from starlite.handlers.websocket_handlers import websocket_listener


@websocket_listener("/", dependencies={"foo": Provide(lambda: "foo")})
async def handler(data: str, foo: str) -> str:
  return data
```

And since this uses the regular signature/kwargs modelling, we can still inject things like `socket` or `state`:

```python
from starlite.handlers.websocket_handlers import websocket_listener
from starlite.datastructures import State

@websocket_listener("/")
async def handler(data: str, socket: WebSocket, state: State) -> str:
  return data
```

### Handling websockets synchronously

Since we don't need to handle the socket directly anymore, it's possible to use synchronous functions as well and this "just works":

```python
from starlite.handlers.websocket_handlers import websocket_listener

@websocket_listener("/")
def handler(data: str) -> str:
  return data
```

### Class based listeners

The class based listener build upon the `websocket_listener`, by providing an OOP way to handle websocket events, using synchronous or asynchronous methods as well:

```python
from starlite.handlers.websocket_handlers import WebsocketListener

class MyListener(WebsocketListener):
  path = "/"

  def on_connect(self, socket: WebSocket) -> None:
    print("socket connected")

  def on_receive(self, data: str) -> str:
    return "hello, world!"
  
  def on_disconnect(self, socket: WebSocket) -> None:
    print("socket disconnected")
```

## Todo

Ideally we want to integrate similar functionalities to route handlers in regards of data (de)serialisation. This is probably best done once #1295 is merged, as it would greatly simplify this. 

This is not a blocker though, and from a functionality standpoint this can be released as-is.

### DIFF
--- a/docs/examples/websockets/dependency_injection_yield.py
+++ b/docs/examples/websockets/dependency_injection_yield.py
@@ -1,8 +1,8 @@
 from typing import TypedDict
 
-from starlite import Starlite, websocket_listener
-from starlite.datastructures import State
-from starlite.di import Provide
+from litestar import Starlite, websocket_listener
+from litestar.datastructures import State
+from litestar.di import Provide
 
 
 class Message(TypedDict):
--- a/docs/examples/websockets/dependency_injection_yield.py
+++ b/docs/examples/websockets/dependency_injection_yield.py
@@ -1,0 +1,27 @@
+from typing import TypedDict
+
+from starlite import Starlite, websocket_listener
+from starlite.datastructures import State
+from starlite.di import Provide
+
+
+class Message(TypedDict):
+    message: str
+    client_count: int
+
+
+def socket_client_count(state: State) -> int:
+    if not hasattr(state, "count"):
+        state.count = 0
+
+    state.count += 1
+    yield state.count
+    state.count -= 1
+
+
+@websocket_listener("/", dependencies={"client_count": Provide(socket_client_count)})
+async def handler(data: str, client_count: int) -> Message:
+    return Message(message=data, client_count=client_count)
+
+
+app = Starlite([handler])
--- a/docs/examples/websockets/dependency_injection_simple.py
+++ b/docs/examples/websockets/dependency_injection_simple.py
@@ -1,5 +1,5 @@
-from starlite import Starlite, websocket_listener
-from starlite.di import Provide
+from litestar import Starlite, websocket_listener
+from litestar.di import Provide
 
 
 def some_dependency() -> str:
--- a/docs/examples/websockets/dependency_injection_simple.py
+++ b/docs/examples/websockets/dependency_injection_simple.py
@@ -1,0 +1,14 @@
+from starlite import Starlite, websocket_listener
+from starlite.di import Provide
+
+
+def some_dependency() -> str:
+    return "hello"
+
+
+@websocket_listener("/", dependencies={"some": Provide(some_dependency)})
+async def handler(data: str, some: str) -> str:
+    return data + some
+
+
+app = Starlite([handler])
--- a/docs/examples/websockets/listener_class_based.py
+++ b/docs/examples/websockets/listener_class_based.py
@@ -1,0 +1,18 @@
+from starlite import Starlite, WebSocket
+from starlite.handlers import WebsocketListener
+
+
+class Handler(WebsocketListener):
+    path = "/"
+
+    def on_accept(self, socket: WebSocket) -> None:
+        print("Connection accepted")
+
+    def on_disconnect(self, socket: WebSocket) -> None:
+        print("Connection closed")
+
+    def on_receive(self, data: str) -> str:
+        return data
+
+
+app = Starlite([Handler])
--- a/docs/examples/websockets/listener_class_based.py
+++ b/docs/examples/websockets/listener_class_based.py
@@ -1,5 +1,5 @@
-from starlite import Starlite, WebSocket
-from starlite.handlers import WebsocketListener
+from litestar import Starlite, WebSocket
+from litestar.handlers import WebsocketListener
 
 
 class Handler(WebsocketListener):
--- a/docs/examples/websockets/listener_class_based_async.py
+++ b/docs/examples/websockets/listener_class_based_async.py
@@ -1,0 +1,18 @@
+from starlite import Starlite, WebSocket
+from starlite.handlers import WebsocketListener
+
+
+class Handler(WebsocketListener):
+    path = "/"
+
+    async def on_accept(self, socket: WebSocket) -> None:
+        print("Connection accepted")
+
+    async def on_disconnect(self, socket: WebSocket) -> None:
+        print("Connection closed")
+
+    async def on_receive(self, data: str) -> str:
+        return data
+
+
+app = Starlite([Handler])
--- a/docs/examples/websockets/listener_class_based_async.py
+++ b/docs/examples/websockets/listener_class_based_async.py
@@ -1,5 +1,5 @@
-from starlite import Starlite, WebSocket
-from starlite.handlers import WebsocketListener
+from litestar import Starlite, WebSocket
+from litestar.handlers import WebsocketListener
 
 
 class Handler(WebsocketListener):
--- a/docs/examples/websockets/mode_receive_binary.py
+++ b/docs/examples/websockets/mode_receive_binary.py
@@ -1,4 +1,4 @@
-from starlite import Starlite, websocket_listener
+from litestar import Starlite, websocket_listener
 
 
 @websocket_listener("/", receive_mode="binary")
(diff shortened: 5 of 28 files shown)


v3ss0n:
i would be start testing this in a few days , really nice.

julio-34727:
Hi @provinzkraut , will there be a way with the new websocket to emit to frontend out of the route handle like in flask_socketio (https://flask-socketio.readthedocs.io/en/latest/) ?

provinzkraut:
> Hi @provinzkraut , will there be a way with the new websocket to emit to frontend out of the route handle like in flask_socketio (https://flask-socketio.readthedocs.io/en/latest/) ?

Can you be a bit more specific which of their features you're referring to?

Broadcasting (sending a message to all connected clients) won't be a built-in functionality implemented here, but it's planned. You can look at a very basic implementation of this here: https://github.com/starlite-api/starlite/pull/1378

Goldziher (commented):
Very nice all in all

starlite/handlers/websocket_handlers.py
```diff
@@ -94,3 +100,144 @@ def _validate_handler_function(self) -> None:
...
+                    elif isinstance(data_to_send, bytes):
+                        await socket.send_bytes(data_to_send)
+                    else:
+                        await socket.send_json(data_to_send)
+                except WebSocketDisconnect:
+                    if self._on_disconnect:
+                        await self._on_disconnect(socket)
+                    break
+
+        # make our listener_fn look like the callback, so we get a correct signature model
```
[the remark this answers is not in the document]
provinzkraut: Because no callable with the correct signature exists. 

There is the callback, which has the additional dependencies and `data` but not necessarily `socket`, and the handler function we create, which has `socket` but none of the additional dependencies defined by the callback.

So we need to create a signature from a function that looks exactly like the callback, minus `data`, and which is guaranteed to include `socket`.

Maybe I don't follow what you mean, but I don't think your suggestion would accomplish that?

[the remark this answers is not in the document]
Goldziher: Ok, lets revisit this code again in the future for refactoring - i think we can address this in a more cohesive way with some of the approaches that have been recently introduced by @peterschutt, but its not a blocker.

starlite/handlers/websocket_handlers.py
```diff
@@ -99,3 +112,252 @@ def _validate_handler_function(self) -> None:
...
+                    received_data = received_data.decode("utf-8")
+            elif wants_receive_type is bytes:
+                if isinstance(received_data, str):
+                    received_data = received_data.encode("utf-8")
+            else:
+                received_data = decode_json(received_data)
+            return received_data
+
+        async def handle_send(socket: WebSocket, data_to_send: Any) -> None:
+            if not can_send_data:
```
provinzkraut, line 211: I'm not sure if we should do this runtime validation, as we don't have this in place everywhere else. @starlite-api/maintainers?

starlite/handlers/websocket_handlers.py
```diff
@@ -99,3 +112,252 @@ def _validate_handler_function(self) -> None:
...
+                if data_to_send is None:
+                    raise RuntimeError(f"Expected {callback} to not return a value. Returned {data_to_send!r}")
+                return
+
+            if data_to_send is None:
```
provinzkraut, line 216: See comment above

starlite/handlers/websocket_handlers.py
Goldziher: im not in favor

docs/examples/websockets/dependency_injection_yield.py
```diff
@@ -0,0 +1,27 @@
+from typing import TypedDict
```
Goldziher, line 1: can you rebase this PR on the litestar stuff?

starlite/handlers/websocket_handlers.py
```diff
@@ -99,3 +112,252 @@ def _validate_handler_function(self) -> None:
...
+            guards=guards,
+            middleware=middleware,
+            name=name,
+            opt=opt,
+            signature_namespace=signature_namespace,
+            **kwargs,
+        )
+
+    def _validate_handler_function(self) -> None:
+        """Validate the route handler function once it's set by inspecting its return annotations."""
```
Goldziher, line 184: ?

Goldziher: why is this method empty?

starlite/handlers/websocket_handlers.py
```diff
@@ -99,3 +112,252 @@ def _validate_handler_function(self) -> None:
...
+            on_disconnect=self.on_disconnect,
+            opt=self.opt,
+            path=self.path,
+            return_dto=self.return_dto,
+            signature_namespace=self.signature_namespace,
+            type_encoders=self.type_encoders,
+        )(self.on_receive)
+
+    def on_accept(self, socket: WebSocket) -> SyncOrAsyncUnion[None]:  # noqa: B027
+        """Called after a WebSocket connection has been accepted"""
```
Goldziher, line 347: why not mark as abstract?

starlite/handlers/websocket_handlers.py
provinzkraut: Then let's not

starlite/handlers/websocket_handlers.py
provinzkraut: Because the validation needs to happen elsewhere and the validation of the base classes for this needs to be disabled since it does not apply in this case.

starlite/handlers/websocket_handlers.py
```diff
@@ -99,3 +112,252 @@ def _validate_handler_function(self) -> None:
...
+        This should take a ``data`` argument, receiving the processed WebSocket data,
+        and can additionally include handler dependencies such as ``state``, or other
+        regular dependencies.
+
+        Data returned from this function will be serialized and sent via the socket
+        according to handler configuration.
+        """
+        raise NotImplementedError
+
+    def on_disconnect(self, socket: WebSocket) -> SyncOrAsyncUnion[None]:  # noqa: B027
```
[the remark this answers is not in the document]
provinzkraut: Because then I'd have to check if this method has been implemented. By not making it abstract I can just use it as-is.

docs/examples/websockets/dependency_injection_yield.py
provinzkraut: I did.

provinzkraut:
I rebased this on, and set the target to `rename-target`. Once that one is merged, this PR can be merged into `main`.

starlite/handlers/websocket_handlers.py
Goldziher: so can you add a comment explaining this?

Goldziher (approved):

starlite/handlers/websocket_handlers.py
provinzkraut: Okay so the base class of this is `WebSocketRouteHandler`, which performs a bunch of validations, all of which do not apply to this subclass. For example, it checks that handler functions return `None`, have to be asynchronous, need to have a `socket` parameter, etc. All of this is not applicable in this case, so we just disable this completely.

We cannot however use this method to validate the handler passed in by the user, because that's already been wrapped by the time `_validate_handler_function` is called, since this happens *after* it has been replaced by the handler instance; Upon being registered on the application instance.

provinzkraut: Does that make sense?

Goldziher: i got it, i meant - add a comment in the code 😉

provinzkraut: Oh :cry:

PR #1415, opened 2023-03-30
move pytest-lazy-fixture to dev dependencies

bollwyvl:
### Pull Request Checklist

### Description

Moves `pytest-lazy-fixture` from a runtime dependency to development dependencies.

Noted on https://github.com/conda-forge/starlite-feedstock/pull/43

### Close Issue(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,6 @@ redis = { version = "*", optional = true, extras = ["hiredis"] }
 rich = {version = ">=13.0.0", optional = true}
 structlog = { version = "*", optional = true }
 typing-extensions = "*"
-pytest-lazy-fixture = "^0.6.3"
 
 [tool.poetry.group.dev.dependencies]
 aiomcache = "*"
@@ -89,6 +88,7 @@ pre-commit = "*"
 pytest = "*"
 pytest-asyncio = "*"
 pytest-cov = "*"
+pytest-lazy-fixture = "^0.6.3"
 pytest-mock = "*"
 python-dotenv = "*"
 python-jose = "*"


provinzkraut:
Thanks for the PR! Seems like this has been introduced with the latest patch. I'll get a new one out this evening.

provinzkraut:
@all-contributors add @bollwyvl code

provinzkraut:
@bollwyvl `v1.51.9` is out, containing this fix!

bollwyvl:
Great, thanks for the quick turn.

Shipped 1.51.9 without patches on https://github.com/conda-forge/starlite-feedstock/pull/44

There are a couple other _oddities_ in `pyproject.toml` which we [template](https://github.com/conda-forge/starlite-feedstock/blob/main/recipe/update_recipe.py#L79) around when consuming the `pyproject.toml`:
- `mako` is a hard dep, but also has a dedicated extra
-  the`tortoise-orm` compatibility pin is only listed in dev dependencies, not as an optional in the main `dependencies`

More broadly:
- there's a bunch of `sqlalchemy` stuff, but there is no extra declared
- it's hard to smoke test some of the extras, as they aren't actually imported until _used_
  - we just have `TODO`s

provinzkraut:
> `mako` is a hard dep, but also has a dedicated extra

Looks like `mako` is optional to me?

https://github.com/starlite-api/starlite/blob/6ad53abd3119b8acd63be26b77e4f9b7ef6c85a6/pyproject.toml#L51

> the `tortoise-orm` compatibility pin is only listed in dev dependencies

I *think* this was intentionaly, but it should also work as an extra. I'll investigate.

> there's a bunch of sqlalchemy stuff, but there is no extra declared

Afaik the dev dependency is intentionally for testing only. @peterschutt?

> it's hard to smoke test some of the extras, as they aren't actually imported until used
> we just have TODOs

Could you elaborate on those?

bollwyvl:
> Looks like `mako` is optional to me?

Welp, I don't know (or really trust, to be blunt) what `poetry` does, but the `METADATA` in the latest wheel looks like:

```email
Requires-Dist: aiomcache ; extra == "memcached" or extra == "full"
Requires-Dist: anyio (>=3)
Requires-Dist: brotli ; extra == "brotli" or extra == "full"
Requires-Dist: click ; extra == "cli" or extra == "standard" or extra == "full"
Requires-Dist: cryptography ; extra == "cryptography" or extra == "jwt" or extra == "full"
Requires-Dist: fast-query-parsers
Requires-Dist: httpx (>=0.22)
Requires-Dist: importlib-metadata ; python_version < "3.10"
Requires-Dist: jinja2 (>=3.1.2) ; extra == "jinja" or extra == "standard" or extra == "full"
Requires-Dist: jsbeautifier ; extra == "cli" or extra == "standard"
Requires-Dist: mako (>=1.2.4)
Requires-Dist: msgspec (>=0.11.0)
Requires-Dist: multidict (>=6.0.2)
Requires-Dist: opentelemetry-instrumentation-asgi ; extra == "opentelemetry" or extra == "full"
Requires-Dist: picologging ; extra == "picologging" or extra == "standard" or extra == "full"
Requires-Dist: pydantic
Requires-Dist: pydantic-factories
Requires-Dist: pydantic-openapi-schema (>=1.5.0)
Requires-Dist: python-jose ; extra == "jwt" or extra == "full"
Requires-Dist: pyyaml
Requires-Dist: redis[hiredis] ; extra == "redis" or extra == "full"
Requires-Dist: rich (>=13.0.0) ; extra == "cli" or extra == "standard" or extra == "full"
Requires-Dist: structlog ; extra == "structlog" or extra == "full"
Requires-Dist: typing-extensions
```

There's that `mako`, which `pip check` will duly point out.

But, of course, `pip check` is pretty much garbage for _validating_ `[extras]` (or keeping them up-to-date from successive interactive `pip installs`)... which brings us to...

> Could you elaborate on those?

So, for example, it's easy to test if the API of the upstream is _roughly_ correct of e.g `starlite[jinja]` by importing `starlite.contrib.jinja`. It explodes if it's wrong, and we don't ship something _knowingly_ incompatible.

`starlite[brotli]`, however, won't be exercised until its [_actually_ imported](https://github.com/starlite-api/starlite/blob/v1.51.9/starlite/middleware/compression.py#L49) and used inside a function. So there's no "cheap" way to roughly gauge whether it will work, as-shipped.

I haven't looked at running the full test suite, or even if `conda-forge` has all the test dependencies, which would give me more warm fuzzies, but that's a task for another day...

peterschutt:
Mako shows as `optional = false` in the lock file:

```
[[package]]
name = "mako"
version = "1.2.4"
description = "A super-fast templating language that borrows the best ideas from the existing templating languages."
category = "main"
optional = false
python-versions = ">=3.7"
files = [
    {file = "Mako-1.2.4-py3-none-any.whl", hash = "sha256:c97c79c018b9165ac9922ae4f32da095ffd3c4e6872b45eded42926deea46818"},
    {file = "Mako-1.2.4.tar.gz", hash = "sha256:d60a3903dc3bb01a18ad6a89cdbe2e4eadc69c0bc8ef1e3773ba53d44c3f7a34"},
]
```
compared to Alembic, another optional dep:

```
[[package]]
name = "alembic"
version = "1.10.2"
description = "A database migration tool for SQLAlchemy."
category = "main"
optional = true
python-versions = ">=3.7"
files = [
    {file = "alembic-1.10.2-py3-none-any.whl", hash = "sha256:8b48368f6533c064b39c024e1daba15ae7f947eac84185c28c06bbe1301a5497"},
    {file = "alembic-1.10.2.tar.gz", hash = "sha256:457eafbdc0769d855c2c92cbafe6b7f319f916c80cf4ed02b8f394f38b51b89d"},
]
```

I think I see the pattern - if we have a package marked as optional, but also in dev deps, it gets locked as `optional = false` and ends up in `Requires-Dist`. E.g, `brotli`, `click`, `cryptography`, `jinja2` are all the same, and I'm sure the others would be also if I were to keep going down the list. `alembic` is an example of a package that _isn't_ in dev deps and _is_ in extras, and it is locked as `optional = true` and doesn't end up as `Requires-Dist`.

The difference with `Mako` is that we don't have a `Mako` extra - and it is listed as an optional dependency.

Do we want optional extras to end up in `Requires-Dist`, or not? I'm not very familiar with this level of packaging infrastructure.

> the tortoise-orm compatibility pin is only listed in dev dependencies, not as an optional in the main dependencies

Given that we have the tortoise extra, we should have the optional.

> there's a bunch of sqlalchemy stuff, but there is no extra declared

Can you give some more info on this please? We have the `sqlalchemy` extra which includes `sqlalchemy` and `alembic`.

provinzkraut:
> I think I see the pattern - if we have a package marked as optional, but also in dev deps, it gets locked as `optional = false` and ends up in `Requires-Dist`

I would not expect this behaviour. Is this intended?

provinzkraut:
> I would not expect this behaviour. Is this intended?

Well I could not find anything in the documentation about this being intended so I opened up an issue: https://github.com/python-poetry/poetry/issues/7747

provinzkraut:
> `starlite[brotli]`, however, won't be exercised until its [actually imported](https://github.com/starlite-api/starlite/blob/v1.51.9/starlite/middleware/compression.py#L49) and used inside a function. So there's no "cheap" way to roughly gauge whether it will work, as-shipped.

I think we should be able to build in a check for those so they raise an error at application startup when this is being configured. Raising the error on import is going to be difficult since this is intentionally dynamic.

PR #1432, opened 2023-04-01
A parsed signature datastructure.

peterschutt:
# ParsedSignature

This datastructure is intended to be the primary source of parameter and annotation data for handler and dependency callables. Processing is only concerned with rudimentary preparation of annotations that should _always_ occur before using them, e.g., extracting annotations from `get_type_hints()`, extracting inner types and metadata from `Annotated`, etc.

This is intended to be different from signature modelling which has aims that conflict with having an absolute source of truth of the signature parameter and return value types. For example, code such as this:

```py
...
elif parameter.should_skip_validation:
    attribute = attr.attrib(type=Any)
```
...occurs _before_ the type is given to `SignatureField`, making `SignatureModel.fields` unreliable as a primary source of the form of a handler/ dependency function parameter or return annotation.

Signature modelling would instead use this datastructure as it's input instead of parsing the handler/dependency functions again itself.

## Integrated with Signature Modelling

Refactors `starlite._signature` to utilise `ParsedSignature` as input.

## Removes `BaseRouteHandler.signature` attribute

The `signature` attribute of the route handlers was the return value of `Signature.from_callable()`. This was removed in this PR in favor of `BaseRouteHandler.parsed_fn_signature` parameter, which has had forward refs resolved and wrapper types removed, among other things.

However, accessing the `parsed_fn_signature` property should only occur after the hierarchy of the application has been established (i.e., it needs to occur as part of the `Starlite.__init__()` procedure, _not_ in `BaseRouteHandler.__call__()`. This is to ensure that the `signature_namespace` is properly resolved for the handler before parsing the signature.

### BaseRouteHandler.on_registration()

This is a method that is called by the application during `__init__()` as handlers are registered. The `BaseRouteHandler._validate_handler_function()` call has been moved into this method, as have the method calls that pre-resolve certain handler properties on startup. 

### Pull Request Checklist

### Description

### Close Issue(s)


Goldziher:
I would like to see the DTO PR merged before we discuss any more RFCs or changes to the internals. I do not see the need for this now.

peterschutt:
This is _for_ it.

peterschutt:
I commented in your last PR that I was using the parsed parameters, and asked you to expose it. You said no, and told me to do it in a follow up PR.

Goldziher (approved):
Please increase coverage

provinzkraut (approved):
This looks like a nice addition and something I could use very well in an upcoming PR!

docs/reference/types.rst
```diff
@@ -189,3 +189,15 @@ File types
 
 .. autoclass:: starlite.types.FileSystemProtocol
     :members:
+
+Parsed signature types
```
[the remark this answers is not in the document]
peterschutt: I've removed it, but I wouldn't have a big issue with this being exposed as all it does is represent the users handler/dependency functions, there's not really much logic tied into it.

peterschutt:
> Please increase coverage

Still draft, will be fully covered by the time I'm finished.

starlite/_signature/utils.py
```diff
@@ -89,22 +72,171 @@ def create_signature_model(
...
+
+    Args:
+        dependency_name_set: A set of dependency names
+        fn: A callable.
+        parsed_signature: A parsed signature.
+
+    Returns:
+        A tuple containing the following values for generating a signature model: a mapping of field definitions, the
+        callable's return annotation, a mapping of field names to plugins - if any, and an updated dependency name set.
+    """
```
provinzkraut, line 227:
```suggestion
    """Validate dependencies of ``parsed_signature``.

    Args:
        dependency_name_set: A set of dependency names
        fn: A callable.
        parsed_signature: A parsed signature.

    Returns:
        A set of validated dependency names
    """
```

starlite/handlers/base.py
```diff
@@ -287,6 +299,13 @@ def _validate_dependency_is_unique(dependencies: dict[str, Provide], key: str, p
                     f"If you wish to override a provider, it must have the same key."
                 )
 
+    def on_startup(self) -> None:
```
provinzkraut, line 302: I think this should be called `on_registration` or some other name the reflects the point at which it is called. We already have "startup" event, which could lead to confusion, this this method will actually be called within `Starlite.register`, and not on application startup or init (it *may* be called on init, but only incidentally).

starlite/types/parsed_signature.py
```diff
@@ -0,0 +1,187 @@
...
+            raise ImproperlyConfiguredException(
+                f"'{parameter.name}' does not have a type annotation. If it should receive any value, use 'Any'."
+            ) from err
+
+        if parameter.name == "state" and not issubclass(annotation, ImmutableState):
+            raise ImproperlyConfiguredException(
+                f"The type annotation `{annotation}` is an invalid type for the 'state' reserved kwarg. "
+                "It must be typed to a subclass of `starlite.datastructures.ImmutableState` or "
+                "`starlite.datastructures.State`."
+            )
```
provinzkraut, line 142: I think this should go into the `BaseRouteHandler._validate_handler_function`. It logically belongs there, since this is where we do the other signature validations as well. It also makes `ParsedParameter` more generic.

provinzkraut (changes requested):
That's a very nice addition!

Just a few minor things to address from my side!

Goldziher:
Will review it soon - this blossomed quite a bit i c

starlite/_signature/utils.py
```diff
@@ -89,22 +72,171 @@ def create_signature_model(
...
+
+
+def _any_attrs_annotation(parsed_signature: ParsedSignature) -> bool:
+    any_attrs_annotation = False
+    for parameter in parsed_signature.parameters.values():
+        parsed_type = parameter.parsed_type
+        if any(is_attrs_class(t.annotation) for t in parsed_type.inner_annotations) or is_attrs_class(
+            parsed_type.annotation
+        ):
+            any_attrs_annotation = True
```
Goldziher, line 114:
```suggestion
            return True
```

Goldziher (approved):
Very nice. Approving again

starlite/handlers/base.py
peterschutt: Yes I was uneasy about this too, `on_registration` a much better option.

starlite/types/parsed_signature.py
peterschutt: I agree that this doesn't belong here - keeping this stuff as generic as possible is ideal. However, putting it in the handler validation only would mean that dependency functions aren't similarly validated as they are now.

I'll create a utility and look into adding a `Provide.on_registration()` callback as well, or something along those lines.

starlite/utils/typing.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 import re
-from collections import defaultdict, deque
+import typing as t
```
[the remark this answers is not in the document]
peterschutt: I like consistency as much as the next guy, but this is also the only place that we are pretty much importing the _entire_ typing module namespace. Doing that one-by-one seems unnecessarily laborious.

If anything, I think my crime here is not changing the existing code in the module to use the `t.` namespaced references.

I won't block on this, but will make a follow up PR one way or the other after I hear back from you.

starlite/types/parsed_signature.py
peterschutt: I'm going to make a separate issue for resolving this.

PR #1445, opened 2023-04-04
DTO Application Interface

peterschutt:
# DTO Application Interface

## DTO Functions

The interface has the following methods:

### Decoding 

Decoding and converting connection data into an application data type.

- `from_connection()` class method - create an instance of the dto given a connection. This is intended for parsing raw bytes into the type of the ``"data"`` kwarg annotation.
- `to_data_type()` - intended to return the data that is represented by the dto to inject into the handler.

### Encoding

Encoding data types back into bytes for responses.

- `from_data()` class method - create an instance of the dto given the return value of the handler function.
- `to_encodable_type()` - receives the request and returns anything that starlite can convert to bytes, or bytes.

## On Registration

- `on_registration()` class method - this is called by the route handler when it is registered on an application. It receives the `ParsedTyped` of the annotation it is being asked to manage, and the route handler.

## App Layers Interface

DTOs can be defined through all layers of the application structure, closest definition to the handler wins.

Each layer has the following parameters:

- `dto: type[AbstractDTOInterface] | None | Empty` - the DTO used to decode request data to be injected for the ``"data"`` kwarg. If `return_dto` not specified, this is also used as `return_dto`.
- `return_dto: type[AbstractDTOInterface] | None | Empty` - the DTO used to encode handler response data, if not resolved for the handler, the resolved `dto` type is used.

If a handler resolves no `dto` or `return_dto` type for a handler there is no change to current behavior.

For either parameter, to disable a DTO that has been declared higher up the branch, set it to `None` on the node below which you don't want it to apply.


peterschutt:
Going to start splitting out the work from the other dto branch by function.

starlite/dto/interface.py
```diff
@@ -0,0 +1,70 @@
...
+    from starlite.types import StarliteEncodableType
+    from starlite.types.parsed_signature import ParsedType
+
+__all__ = ("AbstractDTOInterface", "DataT")
+
+DataT = TypeVar("DataT")
+"""Type var representing data held by a DTO instance."""
+
+
+class AbstractDTOInterface(Generic[DataT], metaclass=ABCMeta):
```
Goldziher, line 20: you can simply extend ABC, why do you use an explicit metaclass instead?

starlite/dto/interface.py
```diff
@@ -0,0 +1,70 @@
...
+    from typing_extensions import Self
+
+    from starlite.connection import Request
+    from starlite.handlers import BaseRouteHandler
+    from starlite.types import StarliteEncodableType
+    from starlite.types.parsed_signature import ParsedType
+
+__all__ = ("AbstractDTOInterface", "DataT")
+
+DataT = TypeVar("DataT")
```
Goldziher, line 16: I prefer using `T` and single char for generics, unless there are a lot of them, which there arent.

starlite/dto/interface.py
Goldziher, line 20: this abstract class has no logic. Why not have it as a Protocol instead? You call it `Intefarce`, but the semantic name for this in python is `Protocol`

If not, please remove the work `Interface` and lets stick to the standard semantics of `AbstractDTO`.

Also, the rest of the codebase has a `base.py` file, not `interface.py`

starlite/dto/kwarg_extractor.py
```diff
@@ -0,0 +1,36 @@
...
+if TYPE_CHECKING:
+    from typing import Any, Callable, Coroutine
+
+    from starlite.connection import ASGIConnection, Request
+    from starlite.types.parsed_signature import ParsedParameter
+
+__all__ = ("create_dto_extractor",)
+
+
+def create_dto_extractor(
```
Goldziher, line 16: why have this logic outside the kwargs package? its where all extractors live.

starlite/dto/interface.py
Goldziher, line 20: missing slots

tests/openapi/typescript_converter/test_converter.py
```diff
@@ -10,7 +8,7 @@
 from tests.openapi.utils import PersonController, PetController
 
 
-@pytest.mark.skipif(version_info < (3, 10), reason="flaky in python 3.9 and lower")
+@pytest.mark.xfail
```
peterschutt, line 11: `xfail` as this test fails if the v1 `dto_factory` tests don't exist. To test, before this PR is merged, go to `main`, delete the `dto_factory` test directory and run the test suite.

peterschutt:
So how hard to we want to be with type checking the outputs of the DTO from the application side? Anywhere between blind trust and stringently instance check every value.

How about this? The dto gets the annotation given to it in the on registration hook. It can raise an "unsupported" type exception if it doesn't support the type - and the expectation is that DTO types do validate this at this point. The user has the option to revert to default serialisation behaviour, or raise if the DTO cannot handle the type.

What do you guys think about having a sample rate for runtime type checking? So you could configure a runtime type neck every 1000/10000 or whatever requests?

provinzkraut:
> It can raise an "unsupported" type exception if it doesn't support the type - and the expectation is that DTO types do validate this at this point. The user has the option to revert to default serialisation behaviour, or raise if the DTO cannot handle the type.

That sounds like a very good approach. I'm all for that.

> So you could configure a runtime type neck every 1000/10000 or whatever requests?

I'm all for a good type necking, but I'm not so sure if we should do this. Sounds like a lot of overhead. Also, couldn't users simply employ [beartype](https://github.com/beartype/beartype) for this?

<hr>

Maybe (in the somewhat distant future) we could somehow bridge this gap by employing some kind of static analysis - perhaps a mypy plugin or a standalone tool - that would help find those issues we can't easily detect at run-time.

peterschutt:
> Maybe (in the somewhat distant future) we could somehow bridge this gap by employing some kind of static analysis - perhaps a mypy plugin or a standalone tool - that would help find those issues we can't easily detect at run-time.

OK cheers, I'll leave it out. Not familiar with `beartype` - looks interesting!

peterschutt:
> > It can raise an "unsupported" type exception if it doesn't support the type - and the expectation is that DTO types do validate this at this point. The user has the option to revert to default serialisation behaviour, or raise if the DTO cannot handle the type.
> 
> That sounds like a very good approach. I'm all for that.

Sweet, to get this through, I'll provide the `UnsupportedType` exception, and I won't add any handling, so it will be effectively an `ImproperlyConfiguredException`. The fall-back stuff I'll circle back on.

Goldziher:
> What do you guys think about having a sample rate for runtime type checking? So you could configure a runtime type neck every 1000/10000 or whatever requests?

This sounds unpredictable to me. Also, to me we simply guarantee the static type checks, but dynamic type checking at runtime (except validation of our own constructors) is left in user land.

Goldziher:
> So how hard to we want to be with type checking the outputs of the DTO from the application side? Anywhere between blind trust and stringently instance check every value.

I agree we should expect users to implement the same validation we do. Imo we should trust users not to fuck this up, and instruct them in the docs accordingly.

peterschutt:
> instruct them in the docs accordingly.

Yes, I will ensure docs are in order.

PR #1447, opened 2023-04-05
Parsed signature utils

peterschutt:
A few utility functions to centralize type identification logic.

## `is_subclass_of()`

Determine if the `ParsedType` is a sub-type of some provided type, e.g, `ParsedType.from_annotation(bool).is_type_of(int) == True`

## `has_inner_subclass_of()`

Does the `is_subclass_of()` check on any inner types held by the `ParsedType`.

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/starlite/types/parsed_signature.py
+++ b/starlite/types/parsed_signature.py
@@ -40,7 +40,7 @@ class ParsedType:
         "is_required",
         "is_not_required",
         "safe_generic_origin",
-        "inner_annotations",
+        "inner_types",
     )
 
     raw: Any
@@ -64,7 +64,7 @@ class ParsedType:
 
     This is to serve safely rebuilding a generic outer type with different args at runtime.
     """
-    inner_annotations: tuple[ParsedType, ...]
+    inner_types: tuple[ParsedType, ...]
     """The type's generic args parsed as ``ParsedType``, if applicable."""
 
     @property
@@ -103,7 +103,7 @@ class ParsedType:
             is_required=Required in wrappers,
             is_not_required=NotRequired in wrappers,
             safe_generic_origin=get_safe_generic_origin(origin),
-            inner_annotations=tuple(cls.from_annotation(arg) for arg in args),
+            inner_types=tuple(cls.from_annotation(arg) for arg in args),
         )
 
 
--- a/starlite/types/parsed_signature.py
+++ b/starlite/types/parsed_signature.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from collections.abc import Collection
 from dataclasses import dataclass
 from inspect import Parameter, Signature
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Union
 
 from typing_extensions import Annotated, NotRequired, Required, get_args, get_origin
 
@@ -15,8 +15,6 @@ from starlite.utils.signature_parsing import get_fn_type_hints
 from starlite.utils.typing import get_safe_generic_origin, unwrap_annotation
 
 if TYPE_CHECKING:
-    from typing import Any
-
     from starlite.types import AnyCallable
 
 __all__ = (
@@ -75,7 +73,34 @@ class ParsedType:
     @property
     def is_collection(self) -> bool:
         """Whether the annotation is a collection type or not."""
-        return self.origin and issubclass(self.origin, Collection)
+        return bool(self.origin and self.origin is not Union and issubclass(self.origin, Collection))
+
+    def is_subclass_of(self, cl: type[Any] | tuple[type[Any], ...]) -> bool:
+        """Whether the annotation is a subclass of the given type.
+
+        Where ``self.annotation`` is a union type, this method will always return ``False``. While this is not
+        strictly correct, we intend on revisiting this once a concrete use-case is to hand.
+
+        Args:
+            cl: The type to check, or tuple of types. Passed as 2nd argument to ``issubclass()``.
+
+        Returns:
+            Whether the annotation is a subtype of the given type(s).
+        """
+        if self.origin:
+            return self.origin not in UNION_TYPES and issubclass(self.origin, cl)
+        return self.annotation is not Any and issubclass(self.annotation, cl)
+
+    def has_inner_subclass_of(self, cl: type[Any] | tuple[type[Any], ...]) -> bool:
+        """Whether any generic args are a subclass of the given type.
+
+        Args:
+            cl: The type to check, or tuple of types. Passed as 2nd argument to ``issubclass()``.
+
+        Returns:
+            Whether any of the type's generic args are a subclass of the given type.
+        """
+        return any(t.is_subclass_of(cl) for t in self.inner_types)
 
     @classmethod
     def from_annotation(cls, annotation: Any) -> ParsedType:
--- a/starlite/_signature/utils.py
+++ b/starlite/_signature/utils.py
@@ -107,9 +107,7 @@ def get_signature_model(value: Any) -> type[SignatureModel]:
 def _any_attrs_annotation(parsed_signature: ParsedSignature) -> bool:
     for parameter in parsed_signature.parameters.values():
         parsed_type = parameter.parsed_type
-        if any(is_attrs_class(t.annotation) for t in parsed_type.inner_annotations) or is_attrs_class(
-            parsed_type.annotation
-        ):
+        if any(is_attrs_class(t.annotation) for t in parsed_type.inner_types) or is_attrs_class(parsed_type.annotation):
             return True
     return False
 
@@ -120,7 +118,7 @@ def _any_pydantic_annotation(
     for parameter in parsed_signature.parameters.values():
         parsed_type = parameter.parsed_type
         if (
-            any(_is_pydantic_annotation(t.annotation) for t in parsed_type.inner_annotations)
+            any(_is_pydantic_annotation(t.annotation) for t in parsed_type.inner_types)
             or _is_pydantic_annotation(parsed_type.annotation)
             or field_plugin_mappings.get(parameter.name)
         ):
@@ -135,9 +133,7 @@ def _create_field_plugin_mappings(
     for parameter in parsed_signature.parameters.values():
         parsed_type = parameter.parsed_type
         if plugin := get_plugin_for_value(parameter.parsed_type.annotation, plugins):
-            type_value = (
-                parsed_type.inner_annotations[0].annotation if parsed_type.is_collection else parsed_type.annotation
-            )
+            type_value = parsed_type.inner_types[0].annotation if parsed_type.is_collection else parsed_type.annotation
             field_plugin_mappings[parameter.name] = PluginMapping(plugin=plugin, model_class=type_value)
     return field_plugin_mappings
 
--- a/starlite/handlers/asgi_handlers.py
+++ b/starlite/handlers/asgi_handlers.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING, Any, Mapping, Sequence
 
 from starlite.exceptions import ImproperlyConfiguredException
 from starlite.handlers.base import BaseRouteHandler
-from starlite.types.builtin_types import NoneType
 from starlite.utils import is_async_callable
 
 __all__ = ("ASGIRouteHandler", "asgi")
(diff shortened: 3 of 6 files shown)


starlite/types/parsed_signature.py
```diff
@@ -75,7 +75,60 @@ def is_optional(self) -> bool:
...
+
+        Args:
+            cl: The type to check.
+
+        Returns:
+            Whether the annotation is a subclass of the given type.
+        """
+        if self.origin:
+            if self.origin is Union:
+                return any(t.is_type_of(cl) for t in self.inner_types)
```
peterschutt, line 91:
```py
    assert ParsedType.from_annotation(Union[int, str]).is_type_of(int) is True
    assert ParsedType.from_annotation(Union[int, str]).is_type_of(str) is True
    assert ParsedType.from_annotation(Union[int, str]).is_type_of(bool) is False
    assert ParsedType.from_annotation(Optional[int]).is_type_of(int) is True
    assert ParsedType.from_annotation(Optional[int]).is_type_of(str) is False
```
@starlite-api/maintainers is this the right logic for unions?

provinzkraut: Hm. I wouldn't say it's wrong but maybe unexpected?

How about making `is_type_of` accept `cl: type[Any] | tuple[type[Any], ...]`, and if it's a tuple, check that it matches all items (if it's a union / generic alias), and if not a tuple behave like it does now?
The current behaviour is what I'd expect from `has_type_of` when operating on a union and a single type.

Then you'd have

```python
assert ParsedType.from_annotation(Union[int, str]).is_type_of(int) is False
assert ParsedType.from_annotation(Union[int, str]).is_type_of((int, str)) is True
assert ParsedType.from_annotation(Union[int, str]).has_type_of(str) is True
assert ParsedType.from_annotation(Union[int, str]).has_type_of(int) is True
```

starlite/types/parsed_signature.py
```diff
@@ -75,7 +75,60 @@ def is_optional(self) -> bool:
...
+            return issubclass(self.origin, cl)
+        return self.annotation is not Any and issubclass(self.annotation, cl)
+
+    def has_type_of(self, cl: type[Any]) -> bool:
```
provinzkraut, line 95: Maybe `has_any_type`? It doesn't fit in as good with the `is_<thing>_of` scheme, but it's a bit more explicit.

Goldziher (commented):
Interesting patterns!. Can you elaborate why this is cleaner in your view?

peterschutt:
Quite a few of the issues I've addressed recently have been related to our management of typing, and it stands out to me that stringent centralization of our interaction with type annotations will be of benefit.

Answering the question of "Is a parameter a type of 'x'" is something pretty fundamental that deserves centralized logic due to the complexities of it (dancing around `Union` and `All`, flattening and extracting metadata from `Annotated`, for example).

> Can you elaborate why this is cleaner in your view?

Well, if you already have `parsed_type` to hand, `parsed_type.is_type_of(Thing)` is a pretty nice API, in my opinion.

I assume you mean to compare to existing utilities?

I'm trying to avoid `is_class_and_subclass` because it can hide things, which is not necessarily a bad thing, but I simply don't know _what_ it is hiding. One example is that it has hidden where we've been passing an un-resolved string forward reference which was causing `str` return type handlers to be assigned `MediaType.JSON` media type if `__future__.annotations` was imported. This was the code:

```py
        if not self.media_type:
            if return_annotation in {str, bytes, AnyStr, Redirect, File} or any(
                is_class_and_subclass(return_annotation, t_type) for t_type in (str, bytes)  # type: ignore
            ):
                self.media_type = MediaType.TEXT
            else:
                self.media_type = MediaType.JSON
```

Prior to the `ParsedSignature` PR, if `__future__.annotation` was imported, `return_annotation` in that code was a `str` instance, so it was hitting the `else` block.

So, I'm preferring to be more explicit by checking for the `Any` and `Union` which are the only times that the `issubclass()` checks fail in my tests (so far, there may be more when/if these utilities get merged and more fully integrated). I think this explicitness will be of benefit to future contributors (and current) who can see exactly what the logic is to evaluate these sort of type checks.

Some of the existing typing utilities are a bit unclear to me when I should be reaching for them, and what their implications are. For example, `get_origin_or_inner_type()`. I added this test for it the other day to get coverage on a PR:

```py
def test_get_origin_or_inner_type() -> None:
    assert get_origin_or_inner_type(List[Person]) == list
    assert get_origin_or_inner_type(Annotated[List[Person], "foo"]) == List[Person]
```
It is unclear to me when that behavior would be useful as it forces the recursion to isolate the type from `Annotated` into the calling code - and I either want the annotated metadata at runtime, or I don't, and this function serves neither case well, as the annotation metadata is always thrown away, and I have to check for the recursive case in my calling code to get at the true origin. The function then goes on to return a mapped type from the `types_mapping` dict, which itself has no documentation, and there's no mention of why that is, and what problem it is solving in the doc string of the function, nor any tests of the behavior. I've got no doubt the code does exactly what it is supposed to do - it looks like it is to provide a concrete type to represent an abstract type in order to build real data structures at runtime. This just doesn't make the function look like a generic function that I should be pulling off the shelf to do checks for wrapped types. (We could add a `concrete_type` attribute to `ParsedType` to represent this mapping too, if it would be of benefit.)

Another example is the `annotation_is_iterable_of_type()`. This function looks to me like it is trying to identify a homogeneous iterable type, but there is a lot going on which makes it hard to take off the shelf for ad-hoc usage. It too has the opaqueness of the `is_class_and_subclass()` check, but also does regex string matching, and it does an `isinstance(annotation, (List, ...))` check to check for types, which feels strange as instances aren't valid annotations and `isinstance(List[int], List) is False`. Again, I'm sure it is doing what it needs to for a given context, but doesn't feel safe to use as a generic utility.

I bring these things up, just to say that it won't hurt us to make another pass over our handling of types and inference around them, and that is what this series of PRs aims to achieve.

Due to needing proper resolution of forward references we should always ensure that any type-annotation based logic occurs through these `ParsedSignature` and related data structures, and by having some utility methods for fundamental generic operations attached, we remove the need for future contributors to work this stuff out for themselves.

I wasn't so sure of the need for the predicate methods, but went with it  because it will add consistency in `Any` and `Union` and collection inner-type handling when we ask "is this thing a thing of this type".

Sorry, this got a bit long winded - hopefully it clears up my motivations.

starlite/types/parsed_signature.py
peterschutt: Yeh, it isn't right.

Actually, I think if the types are a tuple it should behave like a tuple 2nd arg to `is_subclass`.

What would you say to just making this return `False` for unions?

We could consider adding another predicate property, such as `ParsedType.is_union`, and then explicit union introspection would look something like `if parsed_type.is_union and parsed_type.has_type_of(Thing)`. That is probably the right amount of information to leave behind for future readers of code that happens to be calling out to these things. As it stands, it would be unclear that the underlying call was handling the union aspect of the type check, which would probably make calling code harder  to understand.

I'd say:

```py
assert ParsedType.from_annotation(Union[int, str]).is_type_of((int, str)) is False
assert ParsedType.from_annotation(Union[int, str]).is_type_of(Union[int, str]) is True
# and probably
assert ParsedType.from_annotation(Union[int, str]).is_type_of(Union[int, str, None) is True
```
Should this be named `is_subtype_of()`?

starlite/types/parsed_signature.py
peterschutt: Although, any (as in `Any`) is a bit of an overloaded word in the typing sphere. Let me sleep on it:)

starlite/types/parsed_signature.py
provinzkraut: The issue with `if parsed_type.is_union and parsed_type.has_type_of(Thing)` is that it only works for unions; If I want to compare to other generic aliases (e.g. `dict[str, str]`) )`is_union` this approach does not work.

How would I then check if e.g. an annotation is strictly equal to `dict[str, str]` (ignoring `__future__` and different typing versions of course)?

> Should this be named `is_subtype_of()`?

That would certainly be an option. And make the intent clearer.

starlite/types/parsed_signature.py
provinzkraut: > Although, any (as in `Any`) is a bit of an overloaded word in the typing sphere

Oh, that's right. Hadn't considered that. I was thinking of the `any` builtin.

starlite/types/parsed_signature.py
peterschutt: > How would I then check if e.g. an annotation is strictly equal to dict[str, str] (ignoring __future__ and different typing versions of course)?

It would be `parsed_type.origin is dict and all(t.is_type_of(str) for t in parsed_type.inner_types)`.

> That would certainly be an option. And make the intent clearer.

Yeh, I'll rename.

peterschutt: I think I'll just return `False` for `is_subtype_of()` union case for now, and we can add some handling for the unions when we have a concrete case to look at.

provinzkraut: > It would be `parsed_type.origin is dict and all(t.is_type_of(str) for t in parsed_type.inner_types)`.

Quite a mouthful :grimacing: And also not that great for error messages I think.

But let's leave that as is now. I have a concrete use case for this, but let's get this one merge for now and then I'll play around with it and see what works best in that case.

peterschutt: > Quite a mouthful

Maybe something like `parsed_type.has_origin_and_types(dict, (str, str))`

> I have a concrete use case for this

Awesome, happy for you to work out what is best in that case.

I'll get this polished up.

starlite/types/parsed_signature.py
peterschutt: `has_inner_type()` (we use "inner_type" for the `ParsedType` attribute)?

`contains_type()`?

starlite/types/parsed_signature.py
provinzkraut: > Maybe something like `parsed_type.has_origin_and_types(dict, (str, str))`

At that point, why not just make `parsed_type == dict[str, str]` or `parsed_type == ParseType(origin=dict, args=(str, str))` work? For strict equality checking this might be the most intuitive solution (this is also how `GenericAlias` does equality comparison).

starlite/types/parsed_signature.py
provinzkraut: `contains_type` sounds good. Also makes it clear that this is intended to work on container types.

starlite/types/parsed_signature.py
peterschutt: > `parsed_type == dict[str, str]`

Oh that's genius! If that is effectively an alias for `parsed_type.has_origin_and_types(dict, (str, str))` that would be an awesome API.

```py
class ParsedType:
    def __eq__(self, annotation):
        other = ParsedType.from_annotation(annotation)
        return self.has_origin_and_types(other.origin, tuple(t.annotation for t in other.inner_types))
```
Something like that? It could still be useful to have the `has_origin_and_types()` method for dynamic use when the args aren't known, to save constructing the dynamic generic type at runtime.

> `parsed_type == ParseType(origin=dict, args=(str, str))`

I'm not so warm on this as I don't want `ParsedType` to be easy to instantiate and would rather it is always done through the `from_annotation()` method so we have the consistency that brings.

peterschutt: > I'm not so warm on this as I don't want ParsedType to be easy to instantiate and would rather it is always done through the from_annotation() method so we have the consistency that brings.

I could always use `__post_init__()` too, and make the `__init__()` method only receive the annotation.

peterschutt:
OK, less is more. Got rid of the predicate methods, we can always add if turns out to be a common pattern.

Went with `is_subclass_of` and `has_inner_subclass_of` so that it is clear these are calls to `issubclass()` at their core.

Goldziher (approved):
cool

starlite/types/parsed_signature.py
peterschutt: > how do we handle Python 3.8 and 3.9? You can't do dict[str, str] there outside annotations, so ParsedType.from_annotation(dict[str, str]) would just be type error.

`ParsedType.from_annotation(Dict[str, str])` - same as always on those py versions.

> or can you think of a different way that supports the case of comparing against types from a constant that are not an annotation?

I don't understand what you mean - what is the constant, how is it different from a resolved type annotation?

peterschutt: Oh, are you are saying that you want to write code that supports `parsed_type == dict[str, str]` and have it work on 3.8? Obviously that isn't  possible, but why can't you do `parsed_annotation == Dict[str, str]` in our code - it won't matter if the user has defined their annotation `dict[str, str]`.

peterschutt: I think we are beyond the scope of this PR so I'll resolve this conversation so I can merge, however happy to keep talking about this - I'm obviously missing some understanding of the use-case, so would be good to get to the bottom of it.

PR #1462, opened 2023-04-08
Add attrs support to Partial

Goldziher:
This PR adds support for ATTRS base models to `Partial`. 

Additionally, because I encountered multiple test issues - it also fixes several other issues:

1. indeterminism in the ordering of `anyOf` values in OpenAPI schema
2. usage of the `tests/types` folder which causes import failures in tests (inside pytest there are lines that do from types import X, and these then fail when you have a folder called types)
3. issues with the Partial implementation- apparently a refactor we had in the past causes dataclass partials to modify the original dataclasses. This issue is also in the 1.51.x line, so we will need to upstream that fix there.
4. issues with datetime parsing (in pydantic/attrs) that require us to modify the attrs signature model to be identical to the pydantic behavior.


litestar/utils/predicates.py
```diff
@@ -306,3 +325,24 @@ def is_pydantic_constrained_field(
         )
     except ImportError:
         return False
+
+
+def is_class_var(annotation: Any) -> bool:
```
peterschutt, line 330: I'd love for something like this to be as simple as `ParsedType.from_annotation(annotation).origin is ClassVar` or similar.

Following on from discussion in #1447 - there's a bit more going on here with the forward ref handling etc, no tests that demonstrate behavior, and scant docs - all which make it hard to use as an off-the-shelf utility. I'm definitely not trying to say the behavior isn't correct, desired or necessary, I'm just trying to understand why, for example:

- What cases are we trying to exclude by checking for the `__class__` attribute?
- When does `annotation` have and not have a `_name` attribute?
- When are we expecting `get_type_hints()` to return forward references, and should we try to handle them recursively first?

On that last question, if not, why? For example, I know that this can occur if `__future__.annotations` are imported _and_ the annotation is a literal string, but should we make an effort to recursively resolve that inner forward ref? Then, if the name isn't available to be resolved, we can either choose to make that an error (think for parsing reserved kwargs out of handler sigs), or not (maybe such as this case, when parsing arbitrary user models - but at least we know in this case that the `ClassVar` should always be able to be recursively resolved as we always include the `typing` module in our namespace for resolution).

The code that calls this function in `partial.py` calls `get_type_hints(item)` (i.e., without `include_extras=True`) - so any `Annotated` are stripped from the types before calling. If that should always be the case, to aid in making this function more generally available, this should at least be documented somewhere. This test fails, for example:

```py
def test_is_class_var_with_annotated() -> None:
    assert is_class_var(Annotated[ClassVar[int], "meta"]) is True
```

Goldziher: addressed

litestar/partial.py
```diff
@@ -1,31 +1,33 @@
...
-    is_class_and_subclass,
+    is_attrs_class,
+    is_class_var,
     is_dataclass_class,
+    is_pydantic_model_class,
     is_typed_dict,
 )
 
+if TYPE_CHECKING:
+    import attrs
```
cofin, line 28: Is this a problem when we don't have both of these libraries installed?

cofin (approved):
lgtm

litestar/partial.py
Goldziher: no, this is for type checking only - on our code base

PR #1469, opened 2023-04-10
Add msgspec.Struct Partial support

Goldziher:
This PR adds support for Partial[msgpsec.Struct]

### DIFF
--- a/tests/test_partial.py
+++ b/tests/test_partial.py
@@ -3,6 +3,7 @@ from typing import Any, ClassVar, Optional, get_type_hints
 
 import pydantic
 import pytest
+from msgspec.inspect import type_info
 from pydantic import BaseModel
 from typing_extensions import TypedDict, get_args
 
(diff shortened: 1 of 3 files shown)


cofin (approved):
LGTM

tests/test_partial.py
```diff
@@ -82,6 +84,22 @@ class VanillaDataClassPersonWithClassVar(VanillaDataClassPerson):
             assert NoneType not in get_args(annotation)
 
 
+def test_partial_msgspec_struct() -> None:
+    class MsgspecPersonWithClassVar(MsgSpecStructPerson):
+        cls_var: ClassVar[int]
+
+    partial = Partial[MsgspecPersonWithClassVar]
+
+    assert len(type_info(MsgspecPersonWithClassVar).fields) == len(type_info(MsgSpecStructPerson).fields)  # type: ignore
```
jcrist, line 93: You might want to use `msgspec.structs.fields` ([docs](https://jcristharif.com/msgspec/api.html#msgspec.structs.fields)) here instead of `type_info`. The whole `msgspec.inspect` module is still marked as experimental, while `msgspec.structs.fields` is solid and not going anywhere. It also gets you the same info that you want here.

Goldziher: Cool, thanks

PR #1481, opened 2023-04-11
fix Partial modifying dataclasses

Goldziher:


pyproject.toml
```diff
@@ -115,7 +115,7 @@ sphinx-toolbox = "^3.2.0"
 sphinx-copybutton = "^0.5.1"
 sphinxcontrib-mermaid = "^0.8.1"
 auto-pytabs = {extras = ["sphinx"], version = ">=0.1.1"}
-starlite-sphinx-theme = { git = "https://github.com/starlite-api/starlite-sphinx-theme.git" }
+# starlite-sphinx-theme = { git = "https://github.com/litestar-org/litestar-sphinx-theme.git", rev = "72e2dca" }
```
provinzkraut, line 118:
```suggestion
litestar-sphinx-theme = { git = "https://github.com/litestar-org/litestar-sphinx-theme.git"}
```

I think this should work, but I'm not 100% sure this doesn't break anything for the v1.0 docs. Let's keep this commented out for now and I'll check the docs before I do the `1.51.10` release later.

provinzkraut (approved):

PR #1487, opened 2023-04-12
Support schema generation for attrs and msgspec classes

Goldziher:
This PR does as the title says - it adds OpenAPI Schema support for msgspec.Struct and attrs decorated classes. Additionally it ensures we generate identical schemas for all data container types, disregarding variations between dataclasses and pydantic etc.

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -5,7 +5,6 @@ from copy import copy
 from dataclasses import MISSING, fields
 from datetime import date, datetime, time, timedelta
 from enum import EnumMeta
-from inspect import getdoc
 from ipaddress import IPv4Address, IPv4Interface, IPv4Network, IPv6Address, IPv6Interface, IPv6Network
 from pathlib import Path
 from typing import (
@@ -50,6 +49,7 @@ from litestar.serialization import encode_json
 from litestar.types import DataclassProtocol, Empty, TypedDictClass
 from litestar.utils.predicates import (
     is_dataclass_class,
+    is_optional_union,
     is_pydantic_constrained_field,
     is_pydantic_model_class,
     is_typed_dict,
@@ -83,7 +83,6 @@ if TYPE_CHECKING:
         ConstrainedList = Any  # type: ignore
         ConstrainedSet = Any  # type: ignore
         ConstrainedStr = Any  # type: ignore
-
 try:
     import pydantic
 
@@ -535,9 +534,10 @@ def create_schema_for_pydantic_model(
     Returns:
         A schema instance.
     """
-    field_type_hints = get_type_hints(field_type, include_extras=False)
+
+    field_type_hints = get_type_hints(field_type)
     return Schema(
-        required=[field.alias or field.name for field in field_type.__fields__.values() if field.required],
+        required=sorted([field.alias or field.name for field in field_type.__fields__.values() if field.required]),
         properties={
             (f.alias or f.name): create_schema(
                 field=SignatureField.create(field_type=field_type_hints[f.name], name=f.alias or f.name),
@@ -549,7 +549,6 @@ def create_schema_for_pydantic_model(
         },
         type=OpenAPIType.OBJECT,
         title=_get_type_schema_name(field_type),
-        description=getdoc(field_type) or None,
     )
 
 
@@ -570,10 +569,19 @@ def create_schema_for_dataclass(
     Returns:
         A schema instance.
     """
+    field_type_hints = get_type_hints(field_type)
     return Schema(
-        required=[
-            field.name for field in fields(field_type) if field.default is MISSING and field.default_factory is MISSING
-        ],
+        required=sorted(
+            [
+                field.name
+                for field in fields(field_type)
+                if (
+                    field.default is MISSING
+                    and field.default_factory is MISSING
+                    and not is_optional_union(field_type_hints[field.name])
+                )
+            ]
+        ),
         properties={
             k: create_schema(
                 field=SignatureField.create(field_type=v, name=k),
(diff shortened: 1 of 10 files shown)


litestar/_openapi/schema_generation/schema.py
```diff
@@ -549,14 +558,99 @@ def create_schema_for_pydantic_model(
...
+
+    Returns:
+        A schema instance.
+    """
+
+    field_type_hints = get_type_hints(field_type)
+    return Schema(
+        required=sorted(
+            [
+                struct.name
```
jcrist, line 629: For encoding/decoding msgspec will use the `.encode_name` instead of `.name` attribute. By default these are the same, but if a user has [renamed fields](https://jcristharif.com/msgspec/structs.html#renaming-fields) then they will differ. You'll want the OpenAPI docs to match the behavior. The same `.encode_name` should also be used below when creating the `properties` info.

*Also, this variable should perhaps be called `field` instead of `struct`, but :shrug:.*

litestar/_openapi/schema_generation/schema.py
```diff
@@ -549,14 +558,99 @@ def create_schema_for_pydantic_model(
...
+        field_type: A msgspec.Struct class.
+        generate_examples: Whether to generate examples if none are given.
+        plugins: A list of plugins.
+        schemas: A mapping of namespaces to schemas - this mapping is used in the OA components section.
+
+    Returns:
+        A schema instance.
+    """
+
+    field_type_hints = get_type_hints(field_type)
```
jcrist, line 625: Instead of using `get_type_hints` here, you could rely on the `.type` attribute of the output of `msgspec_struct_fields` below. This will provide the same information - no sense in retrieving it twice. https://jcristharif.com/msgspec/api.html#msgspec.structs.FieldInfo

litestar/_openapi/schema_generation/schema.py
```diff
@@ -549,14 +558,99 @@ def create_schema_for_pydantic_model(
...
+        ),
+        properties={
+            k: create_schema(
+                field=SignatureField.create(field_type=v, name=k),
+                generate_examples=generate_examples,
+                plugins=plugins,
+                schemas=schemas,
+            )
+            for k, v in field_type_hints.items()
+        },
```
jcrist, line 644: This will exclude any [constraints](https://jcristharif.com/msgspec/constraints.html) set on types as interpreted by msgspec. Perhaps that's fine for your use case.

Note that we do provide our own JSON schema generation utility (https://jcristharif.com/msgspec/jsonschema.html) that properly handles all the types msgspec supports (including attrs classes) so the generated docs match the behavior. No pressure to use it, just letting you know it exists.

Goldziher: Thanks, I'm aware. But we want consistent schema's - that's also why we don't use the pydantic stuff.

litestar/_openapi/schema_generation/schema.py
Goldziher: done

litestar/_openapi/schema_generation/schema.py
Goldziher: done

litestar/_openapi/schema_generation/schema.py
```diff
@@ -549,14 +558,98 @@ def create_schema_for_pydantic_model(
...
+            ]
+        ),
+        properties={
+            k: create_schema(
+                field=SignatureField.create(field_type=v, name=k),
+                generate_examples=generate_examples,
+                plugins=plugins,
+                schemas=schemas,
+            )
+            for k, v in get_type_hints(field_type).items()
```
jcrist, line 642: You'll want to use the `encode_name` here as well. Something like (untested):

```suggestion
            field.encode_name: create_schema(
                field=SignatureField.create(field_type=field.type, name=field.encode_name),
                generate_examples=generate_examples,
                plugins=plugins,
                schemas=schemas,
            )
            for field in msgspec_struct_fields(field_type)
```

For mild efficiency you may want to pull the call to `msgspec_struct_fields` out so it's not called twice (once for `required` and once for `properties`.

A test case that you're doing the right thing would be checking that the output on this struct uses the camelCase names everywhere:

```python
class Test(msgspec.Struct, rename="camel"):
    field_one: int
    field_two: float
```

litestar/_openapi/schema_generation/schema.py
```diff
@@ -549,14 +558,98 @@ def create_schema_for_pydantic_model(
...
+    """
+
+    return Schema(
+        required=sorted(
+            [
+                field.encode_name
+                for field in msgspec_struct_fields(field_type)
+                if field.default is NODEFAULT
+                and field.default_factory is NODEFAULT
+                and not is_optional_union(field.type)
```
jcrist, line 632: msgspec doesn't treat `Optional[type]` as a non-required field (I believe pydantic v2 also will no longer do this). Only fields with default values aren't required.

Goldziher: Yes, but we provide a default of None to optional values even if they are not defined. It's on the layer of the application to ensure consistency.

litestar/_openapi/schema_generation/schema.py
Goldziher: Cheers

JacobCoffee (approved):
i think this looks good

litestar/testing/client/async_client.py
```diff
@@ -30,8 +31,11 @@
 
 T = TypeVar("T", bound=ASGIApp)
 
+# ensure that httpx logging is not interfering with our test client
+logging.getLogger("httpx").setLevel(logging.WARNING)
```
provinzkraut, line 35: This is going to break all httpx' logging for all users once they import this module.

It should go into our tests, not the test clients.

PR #1501, opened 2023-04-16
Alternate serialization plugin protocol that leverages DTOs

peterschutt:
This PR re-implements the `SerializationPluginProtocol` to leverage the `DTOInterface` protocol.

If a handler defines a plugin supported type as either the `data` kwarg type annotation, or as the return annotation for a handler function, and no DTO has otherwise been resolved to handle the type, the protocol creates a DTO implementation to represent that type which is then used to de-serialize into, and serialize from instances of that supported type.

Additionally this PR:

- removes the starlite v1 `SerializationPluginProtocol` and associated handling.
- removes the SQLAlchemy 1 contrib
- temporarily removes the Piccolo and Tortoise ORM plugin implementations and docs, these are to be re-implemented in the new format (tracked in #1533 and #1555).
- adds a new SQLAlchemy tutorial to demonstrate using the SQLAlchemy plugins (WIP).

### Pull Request Checklist

### Description

### Close Issue(s)


tests/contrib/sqlalchemy_1/sql_alchemy_plugin/test_sql_alchemy_plugin_integration.py
```diff
@@ -75,3 +78,29 @@ def test_serializing_a_list_of_sql_alchemy_instances() -> None:
...
+    def get_handler(foo: Foo) -> Foo:
+        assert foo is foo_instance
+        return foo
+
+    with create_test_client(
+        [get_handler], plugins=[SQLAlchemyPlugin()], dependencies={"foo": Provide(provide_foo)}, debug=True
+    ) as client:
+        response = client.get("/")
+        assert response.status_code == HTTP_200_OK
+        assert response.json() == {"id": foo_instance.id}
```
peterschutt, line 106: This test was added for demonstration purposes.

The test fails on `assert foo is foo_instance` in the handler, because the return value of `provide_foo` is used to populate a pydantic model that represents `Foo`, and then the values from the pydantic model are used to populate values on a new `Foo` instance that eventually gets injected.

I think this is surprising behavior for dependencies and that validation of individual attributes every time a plugin supported type is injected into some function is overkill.

peterschutt: @litestar-org/maintainers

litestar/handlers/base.py
```diff
@@ -337,6 +339,22 @@ def resolve_return_dto(self) -> type[DTOInterface] | None:
 
         return cast("type[DTOInterface] | None", self._resolved_return_dto)
 
+    def _set_dto(self, dto: type[DTOInterface]) -> None:
+        """Set the data_dto for the handler.
```
peterschutt, line 343:
```suggestion
        """Set the dto for the handler.
```

litestar/plugins.py
```diff
@@ -148,6 +151,36 @@ def from_data_container_instance(
...
+
+        Args:
+            parsed_type: A parsed type.
+
+        Returns:
+            Whether the type is supported by the plugin.
+        """
+        raise NotImplementedError()
+
+    def create_dto_for_type(self, parsed_type: ParsedType) -> type[AbstractDTOFactory]:
```
peterschutt, line 172:
```suggestion
    def create_dto_for_type(self, parsed_type: ParsedType) -> type[DTOInterface]:
```
Protocol shouldn't care what interface implementation it is.

litestar/testing/helpers.py
```diff
@@ -342,7 +342,7 @@ def create_async_test_client(
     opt: Mapping[str, Any] | None = None,
     parameters: ParametersMap | None = None,
     plugins: OptionalSequence[PluginProtocol] = None,
-    preferred_validation_backend: Literal["pydantic", "attrs"] = "attrs",
+    preferred_validation_backend: Literal["pydantic", "attrs"] | None = None,
```
Goldziher, line 345: Why are you modifying the test helpers and only them?

litestar/testing/helpers.py
```diff
@@ -89,7 +89,7 @@ def create_test_client(
...
```
Goldziher, line 92: Sand question as below

litestar/testing/helpers.py
peterschutt: Given that we recently reverted to using the pydantic signature models by default, it seemed strange to me that these would default to the attrs model. So my intention here was for the default to fall back to the application default, unless it is explicitly set to use one or the other.

Goldziher: I see, but the application does not expect None. 

Can you switch these to `pydantic` in both places?

Goldziher (approved):

litestar/testing/helpers.py
peterschutt: These signatures are exactly the same as the app:

https://github.com/litestar-org/litestar/blob/4f734ea5970d414d49b92cf3c49988aa158cad30/litestar/app.py#L208

Happy to change, but my intention was to just mirror exactly what the app accepts. So should we change the app signaturej to be `preferred_validation_backend: Literal["pydantic", "attrs"] = "pydantic"`. At the moment the default is being set at the time that the value is passed into the config object, here:

https://github.com/litestar-org/litestar/blob/4f734ea5970d414d49b92cf3c49988aa158cad30/litestar/app.py#L352

Goldziher: Lol, ok 😸

JacobCoffee (approved):

tests/contrib/sqlalchemy_1/sql_alchemy_plugin/test_sql_alchemy_plugin_integration.py
provinzkraut: > I think this is surprising behavior for dependencies and that validation of individual attributes every time a plugin supported type is injected into some function is overkill.

Absolutely agree. Not only is it surprising but also undesirable. Instances might be expensive to create, contain hidden state or users might simply - and rightfully so - rely on `a is b` logic and expect it to work.

Goldziher: Yup 👍

Goldziher (approved):

provinzkraut (approved):

Issue #1449, opened 2023-04-05
Enhancement: Add `--pythonpath` support to CLI


pawelad:
### Summary

Hi!

I'm a big fan of the `src/` based layouts, which unfortunately doesn't seem to be well supported in the `starlite` CLI interface. Adding `--pythonpath` (similarly to [`gunicorn`](https://docs.gunicorn.org/en/stable/settings.html#pythonpath), [`pytest`](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-pythonpath) and others) would help.

If there's support for the idea, I can try and help with a PR.

### Basic Example

Given this layout:

```
- src/
  - foobar/
    - __init__.py
    - app.py
```

This (predictably) doesn't work:

```
$ starlite --app src.foobar.app:application info
```

This works, but I don't think it's nice DX:

```
$ PYTHONPATH=src starlite --app foobar.app:application info
```

What I'd like is this:

```
$ starlite --pythonpath src --app foobar.app:application info
$ # Actually, I'd like to skip the `:application` part as well, but that's for another issue : )
```

### Drawbacks and Impact

I am not aware of any drawbacks.

### Unresolved questions

I am not aware of any unresolved questions.

patrickarmengol:
> This (predictably) doesn't work:
> 
> ```
> $ starlite --app src.foobar.app:application info
> ```

Did something change since this issue was created? That command seems to work for me. Running an identical project structure with a venv activated; on Litestar 2.0.0a4

Alternatively, you could install the module locally with `pip install -e .` and instead run `litestar --app foobar.app:application info`

pawelad:
> Adding `--pythonpath` would certainly be a good option, but I think we can also just support the `src` pattern by default.

That would work for me, but I guess could be considered "too magical" for some (I'm guessing that's why other tools use the more explicit `--pythonpath` option).

From my perspective, any option without the ugly `PYTHONPATH=src` is a win : )

> > Actually, I'd like to skip the `:application` part as well, but that's for another issue : )
> 
> You can! If you app is in any of the default locations (`app`, `application`), you won't need to specify `--app` at all. There are quite a few patterns supported for discovering the application instance: `app`, `application`, `create_app` callables, any callable that returns `Starlite`.

Sorry, I should've been more clear - if you specify the `--app` parameter, you also  _need_ to specify the name of the application instance as well, even if it's named `app` / `application` otherwise you get an error:

```
[...]

PR #1506, opened 2023-04-16
feat: add `--app-dir` to CLI

cofin:
### Pull Request Checklist

### Description

Exposes the `--app-dir` CLI flag to be passed into the [`uvicorn` command](https://www.uvicorn.org/deployment/#running-from-the-command-line).

### Close Issue(s)
Implements a solution for #1449


peterschutt:
Prob should be added to the docs too? https://github.com/litestar-org/litestar/blob/86b1173cc9012710e48d5ff66e1dd28f40a5b00a/docs/usage/cli.rst?plain=1#L102-L117

litestar/cli/commands/core.py
```diff
@@ -59,12 +61,19 @@ def info_command(app: Litestar) -> None:
 )
 @option("--host", help="Server under this host", default="127.0.0.1", show_default=True)
 @option("--debug", help="Run app in debug mode", is_flag=True)
+@option(
+    "--app-dir",
+    help="Look for APP in the specified directory, by adding this to the PYTHONPATH. Defaults to the current working directory.",
+    default=None,
+    show_default=False,
+)
```
provinzkraut, line 69: This should go on the root command. If we only inject this into uvicorn it won't work, since the autodiscovery logic will still only look in the CWD, and therefore not find the app in the first place. If it's on the root command that means it'll also work for commands besides `run`.

Maybe we can just set the `PYTHONPATH` env-var before the autodiscovery process starts?

tests/cli/test_core_commands.py
```diff
@@ -29,6 +29,7 @@ def mock_subprocess_run(mocker: MockerFixture) -> MagicMock:
 @pytest.mark.parametrize("port", [8081, None])
 @pytest.mark.parametrize("reload", [True, False, None])
 @pytest.mark.parametrize("web_concurrency", [2, None])
+@pytest.mark.parametrize("app_dir", [Path.cwd(), None])
```
provinzkraut, line 32: Should probably test for something that's not the default value and the application file be placed under that path.

provinzkraut (changes requested):

litestar/cli/commands/core.py
cofin: I've done exactly this.  It should now work with all of the commands.  The syntax looks something like:
```shell
litestar --app-dir /home/oracle/Code/Starlite/litestar-fullstack/src/ --app app.asgi:create_app  info
```

provinzkraut: Maybe we could also think about supporting `src/` by default, since it's relatively common? I'm thinking it could be as simple as detecting if it's there, and if no other ASGI app is found try again but set this as the `PYTHONPATH`. Wdyt?

JacobCoffee (approved):

litestar/cli/main.py
```diff
@@ -10,10 +14,17 @@
 
 @group(cls=LitestarExtensionGroup)
 @option("--app", "app_path", help="Module path to a Litestar application")
+@option(
+    "--app-dir",
+    help="Look for APP in the specified directory, by adding this to the PYTHONPATH. Defaults to the current working directory.",
```
JacobCoffee, line 19:
```suggestion
    help="Look for ``APP`` in the specified directory, by adding this to the ``PYTHONPATH``. Defaults to the current working directory.",
```

JacobCoffee (approved):

docs/usage/cli.rst
```diff
@@ -115,6 +117,9 @@ Options
 +-------------------------------------+---------------------------+-----------------------------------------------------------------+
 | ``--debug``                         | ``LITESTAR_DEBUG``        | Run the application in debug mode                               |
 +-------------------------------------+---------------------------+-----------------------------------------------------------------+
+| ``--app-dir``                       | N/A                       | Look for APP in the specified directory, by adding this to the  |
+|                                     |                           | PYTHONPATH. Defaults to the current working directory.          |
++-------------------------------------+---------------------------+-----------------------------------------------------------------+
```
provinzkraut, line 122:
```suggestion
```

provinzkraut (changes requested):
Looks good!

PR #1518, opened 2023-04-18
Mods for `DTOInterface` for `@websocket_listener`.

### LINKED ISSUE #1517: RFC: DTO Interface and websocket listeners
Decorating a handler with `@websocket_listener`, returns a new decorated handler that interacts with the "listener callback" (the  originally decorated handler).

The issue is that `DTOInterface.on_registration()` is retrieving the type that it represents for the purpose (i.e., parse data or serialize response) from the `parsed_fn_signature` property of the handler, but it is really interested in the type of the listener callback.

To resolve this, we can revert to the app passing in the type to the dto to the `on_registration()` handler, instead of it being retrieved from the handler.

### listener callback

These were just my notes while working through it, will leave them here as might be useful for others to get context if they want it.

- Function received by `@websocket_listener` decoraor
- Would be the route handler for other handler implementation types (asgi, original websocket, http)
- is decorated again for abstracting connection handling boilerplate.

### listener function

- this is the function that is actually decorated by `@websocket_listener`
- has usual websocket handler signature (i.e., receives `socket` kwarg)
- the signature of this function isn't the signature that the signature model and dto types should be initialized with.
- current implementation overwrites `__signature__` property of the listener function, so that `parsed_fn_signature` when generated models that, and in turn dto and signature model get the correct signature/types to interact with.

peterschutt:
Refactors the `websocket_listener` class and adds support for DTOs on these handler types.

Closes #1517

### Pull Request Checklist

### Description

### Close Issue(s)


litestar/handlers/websocket_handlers.py
```diff
@@ -246,7 +281,9 @@ def _update_listener_fn_signature(listener_fn: Callable, callback_signature: ins
         listener_fn.__annotations__ = {p.name: p.annotation for p in new_signature.parameters.values()}
 
     def __call__(self, listener_callback: AnyCallable) -> websocket_listener:
-        listener_callback_signature = ParsedSignature.from_fn(listener_callback, self.resolve_signature_namespace())
```
peterschutt, line 249: There is an issue with calling `resolve_signature_namespace()` inside `__call__()`, which is that the hierarchy of the handler hasn't been established yet.

peterschutt:
Hey @provinzkraut - LMK your thoughts with this. The crux of it is that the `DTOInterface.on_registration()` now receives the `ParsedType` of the annotation that it represents along with the other stuff.
, 
In `websocket_listener._init_handler_dtos()` it uses the types from the callback signature to initialize the DTO, instead of the handler signature as would occur if this was a http handler.

litestar/handlers/websocket_handlers.py
provinzkraut: How could we fix that? We do need that information there and I don't immediately see a way to delay this.

provinzkraut:
> types `DTOInterface` connection params to `ASGIConnection` instead of
Request.

So the idea is that the WS listener calls `from_bytes` and passes in the socket?
But what's the socket needed for in the DTO if we already received the bytes at that point?

provinzkraut:
> But what's the socket needed for in the DTO if we already received the bytes at that point?

Never mind. I remembered your explanation for why the connection is passed to `from_bytes` and that still applies here. 

However, I am uncertain if it *should* apply here. The connection is established once, and any handling specific to that connection should also only happen once, and not every time an event is received. So maybe doing it like this intentionally would promote a bad practice? 
I'm thinking that this should be similar to dependencies, which are also only resolved on the level of the route handler, not for every call of the listener callback (i.e. for every WS event).

litestar/handlers/websocket_handlers.py
```diff
@@ -232,6 +250,23 @@ async def listener_fn(socket: WebSocket, **kwargs: Any) -> None:
 
         return listener_fn
 
+    def _init_handler_dtos(self) -> None:
+        """Initialize the data and return DTOs for the handler."""
+        data_parameter = self._listener_callback_signature.parameters.get("data")
+        if data_parameter:
```
provinzkraut, line 256: The `data` parameter is required, so this should either use `parameters["data"]`, or throw an exception. Since we validate that a `data` parameter exists, I'd say we don't need to check for it again here or handle the case where it's not present.

provinzkraut (approved):

tests/handlers/websocket/test_listeners.py
```diff
@@ -133,6 +158,24 @@ def handler(data: str) -> Dict[str, str]:
         assert ws.receive_json(mode=send_mode) == {"data": "foo"}
 
 
+@pytest.mark.parametrize("send_mode", ["text", "binary"])
+def test_listener_send_with_dto(send_mode: WebSocketMode, mock: MagicMock) -> None:
+    @dataclass
+    class User:
+        name: str
+
+    user_dto = DataclassDTO[User]
```
provinzkraut, line 167: This should probably include some DTO specific things, since serialisation of dataclasses is supported by default, even without DTOs.

peterschutt:
> > But what's the socket needed for in the DTO if we already received the bytes at that point?
> 
> Never mind. I remembered your explanation for why the connection is passed to `from_bytes` and that still applies here.
> 
> However, I am uncertain if it _should_ apply here. The connection is established once, and any handling specific to that connection should also only happen once, and not every time an event is received. So maybe doing it like this intentionally would promote a bad practice? I'm thinking that this should be similar to dependencies, which are also only resolved on the level of the route handler, not for every call of the listener callback (i.e. for every WS event).

I see what you are saying, this might require a little re-architect of the interface in that case.

There is two use-cases:
1. parsing client data into the data model type (so far via `from_bytes()` or `from_builtins()`)
2. serializing return data via `from_data()`.

The current pattern is that we always instantiate a dto instance and get the data straight from it:

Parsing inbound connection data:

```py
dto_type.from_bytes(raw, connection).to_data_type()
```

Similarly, when we serialize return data:

```py
dto_type.from_data(data, connection).to_encodable_type()
```

I don't think there is any reason this cannot be:

```py
dto_instance = dto_type(connection)  # any connection related discrimination of backends etc can occur here
dto_instance.raw_to_data_type(raw)  # for data kwarg
dto_instance.data_to_encodable_type(data)  # for return value
```
... where the instance lives as long as the connection does. It would be virtually no different to what we are doing for http, and more suited to ws.

provinzkraut:
> where the instance lives as long as the connection does. It would be virtually no different to what we are doing for http, and more suited to ws.

Sounds like the ideal solution to me then.

peterschutt:
> ... where the instance lives as long as the connection does. It would be virtually no different to what we are doing for http, and more suited to ws.

This would also allow the same DTO instance to serve both the inbound and outbound data for http, if the handler is configured as such.

provinzkraut:
This should probably also update the usage docs on WebSockets which currently mention a limitation that data can only be parsed but not validated. That should be removed and an example using DTOs added. Do you want to do this within the scope of this PR or should I open a separate issue to track this? I can also just update the docs for this once this PR is merged.

litestar/handlers/websocket_handlers.py
peterschutt: I don't know that there _is_ a way to have the hierarchy resolvable from within `__call__()`.. for the same reason I pushed anything that accesses the `parsed_fn_signature` property into `on_registration()`.

peterschutt:
> They currently mention a limitation that data can only be parsed but not validated. That should be removed and an example using DTOs added. Do you want to do this within the scope of this PR or should I open a separate issue to track this? I can also just update the docs for this once this PR is merged.

I'm happy to do this within.

peterschutt:
Last commit was a bit of a large one @provinzkraut.

To get around the fact that we don't want to evaluate things in `__call__()` I've created a context object that holds values of all of the things that the generated handler callable needs to reference. The values on the context object, and the signature replacement can occur in the `on_registration()` handler.

For the handler signature  model, I wonder if moving the logic to create the handler's signature model should be moved on to the handler base class, i.e., instead of `Starlite._create_handler_signature_model()` we have `BaseRouteHandler._create_signature_model()`. That method could then be overridden on the `websocket_listener` class to produce whatever signature model we need.

Sorry about lumping the refactor in with all this, it just sorta happened :grimacing:

tests/handlers/websocket/test_listeners.py
peterschutt: This was a good catch, cheers!

litestar/handlers/websocket_handlers.py
peterschutt: Have addressed this in the latest commit.

peterschutt:
The missing coverage on this PR, is actually missing coverage from the `dto-factory` branch, which has been merged into this one. So will correct as it gets corrected in those other PRs.

Can this PR be merged into main as a part of `dto-factory`? Or, should we keep merging main into it as the other DTO PRs go in? I'd prefer the former, but either way doesn't really matter.

litestar/handlers/websocket_handlers/_utils.py
```diff
@@ -0,0 +1,184 @@
...
+                received_data = received_data.encode("utf-8")
+            received_data = dto.bytes_to_data_type(received_data)
+        elif listener_context.wants_receive_type is str:
+            if isinstance(received_data, bytes):
+                received_data = received_data.decode("utf-8")
+        elif listener_context.wants_receive_type is bytes:
+            if isinstance(received_data, str):
+                received_data = received_data.encode("utf-8")
+        else:
+            received_data = decode_json(received_data)
```
provinzkraut, line 98: To optimize this, instead of putting the outer `if`/`elif` tree inside `handle_receive`, `_create_handle_receive` could return a `handle_receive` function specific to the config, e.g. one for DTOs, one for strings etc. Kind of a micro-optimization but I don't see a reason not to do it. Wdyt?

litestar/handlers/websocket_handlers/listener.py
```diff
@@ -0,0 +1,234 @@
...
+            path=path,
+            dependencies=dependencies,
+            exception_handlers=exception_handlers,
+            guards=guards,
+            middleware=middleware,
+            name=name,
+            opt=opt,
+            signature_namespace=signature_namespace,
+            dto=dto,
+            return_dto=return_dto,
```
provinzkraut, line 118: I don't think the DTOs should be passed to `WebsocketRouteHandler`, since it can't do anything with those.

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    __slots__ = ()
+
+    def __init__(
+        self,
+        path: str | None | list[str] | None = None,
+        *,
+        dependencies: Dependencies | None = None,
+        dto: type[DTOInterface] | None | EmptyType = Empty,
```
provinzkraut, line 28: Same comment as above: Because DTOs don't do anything here, they shouldn't be defined as kwargs.

provinzkraut (approved):

litestar/handlers/websocket_handlers/listener.py
peterschutt: Well it does result in them being assigned to `self` - which is all any handler type does really, as any further use is done via `resolve_dto()`/`resolve_return_dto()`.

peterschutt: I see your point though - it is confusing defining them in `WebsockeRouteHandler` kwargs - I'll remove.

provinzkraut: Hmm. You have a point there. I'm just thinking that, from a user's perspective it might be confusing to be able to pass a dto and then that not having any effect at all.

peterschutt: I've come around to your PoV :)

litestar/handlers/websocket_handlers/_utils.py
peterschutt: I've done this. It means that we need `_ListenerContext.handle_receive` and `_ListenerContext.handle_send` so that we can build them in `on_registration()`, as the conditional needs resolved data.

LMK what you think.

litestar/handlers/websocket_handlers/listener.py
```diff
@@ -0,0 +1,236 @@
...
+        self._listener_context.listener_callback = listener_callback
+        self._listener_context.handler_function = handler_function = _create_handler_function(
+            listener_context=self._listener_context,
+            on_accept=self._on_accept,
+            on_disconnect=self._on_disconnect,
+        )
+        return super().__call__(handler_function)
+
+    def on_registration(self) -> None:
+        _set_listener_context(
```
provinzkraut, line 149: Why not make this a method? It's not being used anywhere else.

provinzkraut (approved):

litestar/handlers/websocket_handlers/_utils.py
provinzkraut: It does make things a bit more opaque but I guess it's a solid approach. I'm not 100% sure if the added complexity is worth it, but I'd be okay with having it like this. We can always revisit this should a better pattern for this emerge.

PR #1539, opened 2023-04-18
Refactor `DTOInterface` - a better pattern for websockets.

peterschutt:
Based on discussion in #1518.

This pattern emphasises and supports that the lifespan of a DTO instance should be bound to the lifespan of the connection.

Any connection-based logic can be performed once, and the DTO used to parse and serialize as often as necessary for the life of the connection.

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/litestar/dto/interface.py
+++ b/litestar/dto/interface.py
@@ -4,8 +4,6 @@ from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
-    from typing_extensions import Self
-
     from litestar.connection import Request
     from litestar.handlers import BaseRouteHandler
     from litestar.types import LitestarEncodableType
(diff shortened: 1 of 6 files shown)


litestar/dto/interface.py
```diff
@@ -19,43 +17,32 @@ class DTOInterface(Protocol):
...
 
         Returns:
             Either ``bytes`` or a type that Litestar can convert to bytes.
         """
 
-    @classmethod
     @abstractmethod
-    def from_bytes(cls, raw: bytes, connection: Request[Any, Any, Any]) -> Self:
-        """Construct an instance from a :class:`Request <.connection.Request>`.
+    def bytes_to_data_type(self, raw: bytes) -> Any:
```
provinzkraut, line 38: That's a good name!

litestar/serialization.py
```diff
@@ -55,12 +54,6 @@ def _enc_constrained_date(date: ConstrainedDate) -> str:
     return date.isoformat()
 
 
-def _enc_dto(dto: DTOInterface) -> Any:
-    if isinstance(ret := dto.to_encodable_type(), bytes):
-        return msgspec.Raw(ret)
```
provinzkraut, line 60: I don't fully understand why this is not needed anymore. Can you explain?

provinzkraut (approved):

litestar/serialization.py
peterschutt: In the previous pattern:

```py
dto_type.from_data(data, connection)
```
Would create a new dto instance and assign the data to it, (`self._data = data`). So then later, when the msgspec encoder would call `dto.to_encodable_type()` the data is retrieved from the instance and converted to an encodable type, and returned.

Now:

```py
dto_type(connection).data_to_encodable_type(data)
```
The data is handed to the DTO and converted to the encodable type in one step, so it just doesn't fit that pattern any more.

provinzkraut: Okay that's a bit easier. Thanks for the explanation!

peterschutt: Specifically for the case where the dto returns bytes, the value is given to the `Response` object which doesn't call `render()` on it if it is bytes.

PR #1550, opened 2023-04-19
Fix #1547: Remove bool coercion in URL parsing

### LINKED ISSUE #1547: Bug: Query parameter values of "true"/"false" become title-cased when used as string query parameter
### Description

When defining a query parameter as `param: str`, and passing it a value of `true`, the value received by the route handler is `True`, having been title cased. The same is true for the value of `false`.

The issue lies within the `parse_query_string_function`:

https://github.com/litestar-org/litestar/blob/ae2b8fcde70566bb9929c984525ee6bf98d16f0f/litestar/_parsers.py#L28-L38

The value is converted to a `bool` here, and later on back to a string since the parameter is annotated as such.

The conversion of query string parameters to `bool`s should not happen there, but instead be directed by the type annotation of the parameter, and occur within the signature model.

_No response_

### MCVE

```python
from litestar import get


@get("/")
def handler(param: str) -> str:
    return param


    assert client.get("/?param=true").text == "True"
```

### Steps to reproduce

_No response_

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

-

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
Fix #1547 by removing the unconditional coercion of bool-like strings (e.g. `"True"`) to booleans. 

The approach taken here is to let the signature model handle this, which Pydantic seems to do by default, whereas I couldn't get attrs to do the same thing. @Goldziher any insights on how to make them behave in the same way here?

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/litestar/_parsers.py
+++ b/litestar/_parsers.py
@@ -34,8 +34,7 @@ def parse_query_string(query_string: bytes) -> tuple[tuple[str, Any], ...]:
     Returns:
         A tuple of key value pairs.
     """
-    _bools = {"true": True, "false": False, "True": True, "False": False}
-    return tuple((k, v if v not in _bools else _bools[v]) for k, v in fast_parse_query_string(query_string, "&"))
+    return tuple(fast_parse_query_string(query_string, "&"))
 
 
 @lru_cache(1024)
--- a/tests/datastructures/test_url.py
+++ b/tests/datastructures/test_url.py
@@ -34,7 +34,7 @@ def test_url() -> None:
     assert url.password == "hunter2"
     assert url.port == 81
     assert url.hostname == "example.org"
-    assert url.query_params.dict() == {"query": ["param"], "bool": [True]}
+    assert url.query_params.dict() == {"query": ["param"], "bool": ["true"]}
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


Goldziher (approved):

litestar/_signature/models/attrs_signature_model.py
```diff
@@ -75,6 +75,16 @@ def _pass_through_unstructure_hook(value: Any) -> Any:
...
 
 
+def _structure_bool(value: Any, cls: type[bool]) -> bool:
+    if isinstance(value, cls):
+        return value
+
+    if isinstance(value, str):
+        return value.lower() in ("true", "1")
+
+    return cls(value)
```
peterschutt, line 85: Should this raise here @Goldziher?

Goldziher: It will raise a value error? If so it's gone as is, if not than yes

Goldziher:
Can your you increase coverage?

litestar/_signature/models/attrs_signature_model.py
peterschutt: It probably should raise for arbitrary strings too.

peterschutt:
> Can your you increase coverage?

Yeah I can next time I sit down.

Goldziher (approved):

litestar/_signature/models/attrs_signature_model.py
peterschutt: this is all the symbols they mention in the pydantic docs now

PR #1587, opened 2023-04-27
WebSocket broadcast / channels

provinzkraut:
Draft of a WebSocket broadcasting + channels implementation for Litestar.

Current features:

- In-memory backend
- Redis backend (for multi-process capabilities) 
- Automatic management of WebSockets (no need to manually collect and send messages)
- Optional: Creation of route handlers for channels + creating channels on-the-fly
- Optional: Sending a history of recent messages to a subscriber

## TODO

- [ ] ~~RabbitMQ backend~~
- [ ] ~~Kafka backend (?)~~
- [ ] ~~Other backends (?)~~
- [x] Extend testing
- [ ] ~~Possibly remove the Redis pub/sub backend~~

## Examples

**Basic functionality**

```python
from litestar import Litestar, websocket, WebSocket


@websocket("/")
async def handler(socket: WebSocket, channels: ChannelsPlugin) -> None:
    await socket.accept()
    # create a subscriber object to interact with the data stream
    subscriber = await channels.subscribe("some_channel")

    # fetch the channel history and put it in the subscriber's queue.
    # this will be published once the subscriber starts consuming
    await plugin.put_subscriber_history(subscriber, "some_channel")

    # iterate through the event stream. Useful if full control is needed
    async for event in subscriber.iter_events():  
        await socket.send_data(event)


channels_plugin = ChannelsPlugin(backend=MemoryChannelsBackend(), channels=["some_channel"])
app = Litestar([handler], plugins=[channels_plugin])

with TestClient(app) as client, client.websocket_connect("/") as ws:
    channels_plugin.publish({"message": "foo"}, "some_channel")  # publish something
    assert ws.receive_json() == {"message": "foo"}  # and it gets send to all subscribed sockets
```

**Let the subscriber send the data for you**

Instead of handling the sending of received data manually, this can also be done automatically
by the subscriber in a background task. This also gives the opportunity to freely interact with the socket without being bound by an infinite loop:

```python
from litestar import Litestar, websocket, WebSocket


@websocket("/")
async def handler(socket: WebSocket, channels: ChannelsPlugin) -> None:
    await socket.accept()
    # create a subscriber object to interact with the data stream
    subscriber = await channels.subscribe("some_channel")

    # fetch the channel history and put it in the subscriber's queue.
    # this will be published once the subscriber starts consuming
    await plugin.put_subscriber_history(subscriber, "some_channel")

    # start a task that publishes incoming messages to the socket
    async with subscriber.run_in_background(socket.send_data):  
        while True:
            await socket.receive() 


channels_plugin = ChannelsPlugin(backend=MemoryChannelsBackend(), channels=["some_channel"])
app = Litestar([handler], plugins=[channels_plugin])

with TestClient(app) as client, client.websocket_connect("/") as ws:
    channels_plugin.publish({"message": "foo"}, "some_channel")  # publish something
    assert ws.receive_json() == {"message": "foo"}  # and it gets send to all subscribed sockets
```

**Letting the plugin create the route handlers**

Alternatively, if all that's needed is to send received data to connected clients, you can also
let the plugin handle everything for you:

```python
from litestar import Litestar

channels_plugin = ChannelsPlugin(
    backend=MemoryChannelsBackend(), create_ws_route_handlers=True, arbitrary_channels_allowed=True
)
app = Litestar(plugins=[channels_plugin])

with TestClient(app) as client, client.websocket_connect("/some_channel") as ws:
    channels_plugin.publish({"message": "foo"}, "some_channel")
    assert ws.receive_json() == {"message": "foo"}
```

**Automatically sending a history once a client connects**

History as well can be sent automatically once a client has connected:

```python
from litestar import Litestar
from litestar.channels.memory import MemoryChannelsBackend


channels_plugin = ChannelsPlugin(
    backend=MemoryChannelsBackend(history=10),  # setting the limit of how many messages to store
    ws_handler_send_history=10,  # setting the limit of how many stored messages to send
    create_ws_route_handlers=True,
    channels=["some_channel"],
)
app = Litestar(plugins=[channels_plugin])


    channels_plugin.publish({"message": "foo"}, "some_channel")
    channels_plugin.publish({"message": "bar"}, "some_channel")

    with client.websocket_connect("/something") as ws:

        channels_plugin.publish({"message": "baz"}, "some_channel")
        assert ws.receive_json() == {"message": "foo"}
        assert ws.receive_json() == {"message": "bar"}
        assert ws.receive_json() == {"message": "baz"}

```

## Design decisions

**High level design**
The plugin is generally more high level than e.g. [encode/broadcaster](https://github.com/encode/broadcaster) and more opinionated. This was mainly done for performance and usability reasons. Shifting most of the work to the concrete implementation allows to make some useful abstractions and opimisations that wouldn't be possible if it was held more generic and low level. In addition, I wanted to provide something that would "just work", and I think especially in combination with the recently added websocket-listeners, this would be a really useful feature.

**Publishing using the worker pattern instead of async**
The publishing mechanism was implemented using queues and the worker pattern. This was done in order to provide a synchronous interface, and not having to block, effectively making calls to `.publish()` `O(1)`. 

**Single producer/consumer**
Instead of creating a producer/consumer streaming the events per connected socket, I chose to go with a central producer/consumer on the plugin instance. This reduces the performance overhead, but makes the implementation a bit more complex and less flexible; Users are not given the opportunity to decide how to handle events. I do not think this is a problem though, since this feature has a very clear purpose and already allows customization enough to be usable in most cases. 

This decision also allowed to - in case of structured data - only serialise it once, at the time of publishing, and subsequently only fan out already serialised data, greatly reducing impact of serialisation on performance.

## Open questions

**Are two separate Redis backends needed?**
There are currently two different Redis backends: A pub/sub based, and a stream based. The stream based implementation was added to enable the history feature, since pub/sub cannot provide this. I am however not sure of the performance impact of this, which should be tested. If it turns out that the streams backend performs as well as the pub/sub backend, the pub/sub backend can be dropped


Goldziher:
I like it! I agree with the decisions in general. 

As for backends - i would consider not implementing all of them. We just need a good ABC or protocol + a redis implementation and that should be enough. Users can than use any number of alternatives by implementing their own backends. 

Question regarding the name `channels`, why? Aside from the allusion yo Django, i don't think this is the best name.

provinzkraut:
> Question regarding the name channels, why?

No particular reason aside from me liking the name. I didn't want to just go for something like "broadcast" or "websocket_broadcast", since for one, "channels" sounds nicer (:stuck_out_tongue:) and more importantly, it highlights the channel features, which are the bulk of this implementation. A basic broadcasting thing is done with way less code if it doesn't need to be generic.

If you have other suggestions, I'm open.

JacobCoffee (commented):
🥲 beautiful

FWIW, i can't think of anything more clever and clean than _channels_

Goldziher:
> > Question regarding the name channels, why?
> 
> No particular reason aside from me liking the name. I didn't want to just go for something like "broadcast" or "websocket_broadcast", since for one, "channels" sounds nicer (😛) and more importantly, it highlights the channel features, which are the bulk of this implementation. A basic broadcasting thing is done with way less code if it doesn't need to be generic.
> 
> If you have other suggestions, I'm open.

channel is fine. The classic name is "room" i think, which is what is used in socket-io and a few others, but channel is equally ok.

litestar/channels/redis.py
```diff
@@ -0,0 +1,126 @@
...
+                approximate=self._cap_streams_approximate,
+            )
+
+    async def stream_events(self) -> AsyncGenerator[tuple[str, Any], None]:
+        stream_ids: dict[str, bytes] = {}
+        while True:
+            stream_keys = [self._make_key(c) for c in self._subscribed_channels]
+            if not stream_keys:
+                await asyncio.sleep(0)
+                continue
```
provinzkraut, line 106: This is not ideal, but the other solutions (e.g. using a `Future` or `Event`) would require a lock around `_subscribed_channels` to ensure their contents stay the same once execution resumes.

An alternative would be to use both an `Event` and still do the check as a fallback. This should lead to the event blocking in most cases where no channels are subscribed and the condition with `sleep(0)` should only be hit in case of a race conditions. @litestar-org/maintainers thoughts?

tests/channels/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+from redis import Redis
+from redis.asyncio import Redis as AsyncRedis
+
+from litestar.channels.memory import MemoryChannelsBackend
+from litestar.channels.redis import RedisChannelsPubSubBackend, RedisChannelsStreamBackend
+
+
+@pytest.fixture()
+def redis_client(docker_ip: str) -> AsyncRedis:
+    Redis(host=docker_ip, port=6397).flushall()
```
provinzkraut, line 13: For some reason pytest-asyncio goes haywire when this is an async function so we use a separate sync redis client to perform the flush

cofin: What's the purpose for the `flushall` call here?  Without storage allocated to the container, shouldn't the Redis database be empty already?

provinzkraut: The redis service is session scoped. This fixture is function scoped. By calling `flushall` it’s ensured that every test invocation gets an empty database.

provinzkraut:
I have decided to not add any more backends at this point. Redis and in-memory are good enough for this to be released and other backends can easily be added in a future PR.

litestar/channels/_redis_prune_streams.lua
provinzkraut, line 1: Is this the best way to handle this? I'm not sure.

The issue with how we use streams is that we want a history per channel, which means every channel has to be its own stream. This also means there's no native way to "expire" channels. This functionality is needed though, since streams for channels with no subscribers would be kept around indefinitely. 

The idea here is that periodically this script can be called, which then goes through all the streams, checks if it has any new IDs created after a certain cutoff point (a UNIX timestamp) and if not, deletes those.

After thinking about it for a while, this seems to be the least brittle option (and maybe even the most performant one?). The only viable alternative would be to additionally keep a log in redis (using HINCR for example) of how many global subscribers there are to a channel, and delete its stream once this number drops to zero. This though has significantly more overhead and the risk of improper cleanup, e.g. when a connected client crashes and does not properly decrease its subscriber count.

litestar/channels/plugin.py
```diff
@@ -214,15 +289,24 @@ async def unsubscribe(self, subscriber: Subscriber, channels: str | Iterable[str
                 channels_to_unsubscribe.add(channel)
 
         if not any(subscriber in queues for queues in self._channels.values()):
-            await subscriber.put(None)
+            await subscriber.put(None)  # this will stop any running task or generator by breaking the inner loop
```
JacobCoffee, line 292: Can you explain me this? Discord is okay

litestar/channels/base.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import AsyncGenerator, Iterable
+
+
+class ChannelsBackend(ABC):
```
Goldziher, line 7: Perhaps `BaseChannelsBackend` is better, since this is an ABC

litestar/channels/base.py
```diff
@@ -0,0 +1,41 @@
...
+        """Called by the plugin on application startup"""
+        ...
+
+    @abstractmethod
+    async def on_shutdown(self) -> None:
+        """Called by the plugin on application shutdown"""
+        ...
+
+    @abstractmethod
+    async def publish(self, data: bytes, channels: Iterable[str]) -> None:
```
Goldziher, line 19: Iterable or Sequence?

litestar/channels/base.py
```diff
@@ -0,0 +1,41 @@
...
+        """Publish the message ``data`` to all ``channels``"""
+        ...
+
+    @abstractmethod
+    async def subscribe(self, channels: Iterable[str]) -> None:
```
Goldziher, line 24: Sake question everywhere

litestar/channels/memory.py
```diff
@@ -0,0 +1,77 @@
...
+import asyncio
+import contextlib
+from asyncio import Queue
+from collections import defaultdict, deque
+from typing import Any, AsyncGenerator, Iterable
+
+from litestar.channels.base import ChannelsBackend
+
+
+class MemoryChannelsBackend(ChannelsBackend):
```
Goldziher, line 12:
```suggestion
class InMemoryChannelsBackend(ChannelsBackend):
```

litestar/channels/memory.py
```diff
@@ -0,0 +1,77 @@
...
+        Raises:
+            RuntimeError: If ``on_startup`` has not been called yet
+        """
+        if not self._queue:
+            raise RuntimeError("Backend not yet initialized. Did you forget to call on_startup?")
+
+        for channel in channels:
+            if channel not in self._channels:
+                continue
+            self._queue.put_nowait((channel, data))
```
Goldziher, line 47:
```suggestion

            self._queue.put_nowait((channel, data))
```

litestar/channels/memory.py
```diff
@@ -0,0 +1,77 @@
...
+            channels: Channels to publish to
+
+        Returns:
+            None
+
+        Raises:
+            RuntimeError: If ``on_startup`` has not been called yet
+        """
+        if not self._queue:
+            raise RuntimeError("Backend not yet initialized. Did you forget to call on_startup?")
```
Goldziher, line 42: `ImproperlyConfiguredException`

litestar/channels/memory.py
```diff
@@ -0,0 +1,77 @@
...
+
+    async def subscribe(self, channels: Iterable[str]) -> None:
+        """Subscribe to ``channels``, and enable publishing to them"""
+        self._channels.update(channels)
+
+    async def unsubscribe(self, channels: Iterable[str]) -> None:
+        """Unsubscribe from ``channels``"""
+        self._channels = self._channels - (set(channels))
+        for channel in channels:
+            with contextlib.suppress(KeyError):
```
Goldziher, line 60: You can move this above the for loop

litestar/channels/memory.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
```
Goldziher, line 1: I'd rename the file to `in_memory_backend.py`

litestar/channels/plugin.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+T = TypeVar("T")
+
+
+class ChannelsException(LitestarException):
+    pass
+
+
+class AsyncDeque(Queue, Generic[T]):
```
Goldziher, line 35: What makes this Async?

litestar/channels/plugin.py
```diff
@@ -0,0 +1,431 @@
...
+    def __init__(self, maxsize: int | None) -> None:
+        self._deque_maxlen = maxsize
+        super().__init__()
+
+    def _init(self, maxsize: int) -> None:
```
Goldziher, line 40: What's this underscored method? This is not convention used in our code

litestar/channels/plugin.py
```diff
@@ -0,0 +1,431 @@
...
+            return
+
+        async with anyio.create_task_group() as task_group:
+            for channel in channels:
+                task_group.start_soon(self._put_channel_history, channel, limit)
+
+    async def _put_channel_history(self, channel: str, limit: int | None = None) -> None:
+        history = await self._backend.get_history(channel, limit)
+        for entry in history:
+            self._queue.put_nowait(entry)
```
Goldziher, line 104: You can do this inside a comprehension

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
+from __future__ import annotations
```
Goldziher, line 1: This file should be renamed, otherwise there will be import errors. 

`redis_channels_backend.py`

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            redis: A :class:`redis.asyncio.Redis` instance
+            key_prefix: Key prefix to use for storing data in redis
+            stream_sleep_no_subscriptions: Amount of time in milliseconds to pause the
+                :meth:`stream_events` generator, should no subscribers exist
+        """
+        self._redis = redis
+        self._key_prefix = key_prefix
+        self._stream_sleep_no_subscriptions = stream_sleep_no_subscriptions / 1000
```
Goldziher, line 38: Use a constant for MS?

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _make_key(self, channel: str) -> str:
+        return f"{self._key_prefix}_{channel.upper()}"
+
+
+class RedisChannelsPubSubBackend(RedisChannelsBackend):
+    def __init__(
+        self, *, redis: Redis, stream_sleep_no_subscriptions: int = 1, key_prefix: str = "LITESTAR_CHANNELS"
```
Goldziher, line 46: Should be a const

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+        self._history_limit = history
+        self._subscribed_channels: set[str] = set()
+        self._cap_streams_approximate = cap_streams_approximate
+        self._stream_ttl = stream_ttl if isinstance(stream_ttl, int) else int(stream_ttl.total_seconds() * 1000)
+        self._flush_all_streams_script = self._redis.register_script(_FLUSHALL_STREAMS_SCRIPT)
+        self._publish_script = self._redis.register_script(_XADD_EXPIRE_SCRIPT)
+
+    async def on_startup(self) -> None:
```
Goldziher, line 144: Why do you need this boilerplate?

litestar/channels/base.py
provinzkraut: `Iterable` is correct, doesn't need to be sequence.

litestar/channels/memory.py
provinzkraut: Why? We also have the `MemoryStore` and not `InMemoryStore`.

litestar/channels/memory.py
provinzkraut: See comment below

litestar/channels/memory.py
provinzkraut: Not really. It's not a configuration error. We follow the same pattern elsewhere.

litestar/channels/memory.py
provinzkraut: Actually this should go completely. Ruff forces this usage, but I don't like introducing a context stack for just a single operation. Especially since we have zero-cost abstractions in newer Python versions, this seems like an antipattern.

litestar/channels/plugin.py
provinzkraut: It subclasses `Queue`, which is an async queue.

litestar/channels/plugin.py
provinzkraut: That's how the stdlib wants you to customise this. It's a weird, non-standard practice but that's how it is
 ¯\\_(ツ)_/¯

litestar/channels/plugin.py
provinzkraut: Sure, but using a comprehension to iterate without producing an iterable is pretty much considered an anti-pattern in my book.

litestar/channels/plugin.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def publish(self, data: LitestarEncodableType, channels: str | Iterable[str]) -> None:
+        """Publish ```data`` to ``channels``"""
+        if isinstance(channels, str):
+            channels = [channels]
+        data = self.encode_data(data)
+        try:
+            self._pub_queue.put_nowait((data, list(channels)))  # type: ignore[union-attr]
+        except AttributeError as e:
+            raise RuntimeError("Plugin not yet initialized. Did you forget to call on_startup?") from e
```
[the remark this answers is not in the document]
provinzkraut: Same comment as before.

litestar/channels/redis.py
provinzkraut: That's our established pattern. We have `stores.redis`, `contrib.sqlalchemy`, `contrib.jinja`, etc.

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
...
+    from redis.asyncio import Redis
+    from redis.asyncio.client import PubSub
+
+_resource_path = importlib_resources.files("litestar.channels")
+_PUBSUB_PUBLISH_SCRIPT = (_resource_path / "_redis_pubsub_publish.lua").read_text()
+_FLUSHALL_STREAMS_SCRIPT = (_resource_path / "_redis_flushall_streams.lua").read_text()
+_XADD_EXPIRE_SCRIPT = (_resource_path / "_redis_xadd_expire.lua").read_text()
+
+
+class RedisChannelsBackend(ChannelsBackend, ABC):
```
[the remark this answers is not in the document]
provinzkraut: Yes, linters complain otherwise because this does not implement all methods.

litestar/channels/redis.py
provinzkraut: Because it's an abstract method on the base class, and in this implementation it's a no-op. Alternative would be to implement this as no-op in `RedisChannelsBackend` and then overwrite again here but this seems more explicit since you won't have to potentially consider `super()` calls etc. It just seems cleaner to be to only implement abstract methods once.

litestar/channels/redis.py
```diff
@@ -0,0 +1,220 @@
...
+        self._subscribed_channels: set[str] = set()
+        self._cap_streams_approximate = cap_streams_approximate
+        self._stream_ttl = stream_ttl if isinstance(stream_ttl, int) else int(stream_ttl.total_seconds() * 1000)
+        self._flush_all_streams_script = self._redis.register_script(_FLUSHALL_STREAMS_SCRIPT)
+        self._publish_script = self._redis.register_script(_XADD_EXPIRE_SCRIPT)
+
+    async def on_startup(self) -> None:
+        """Called on application startup"""
+
+    async def on_shutdown(self) -> None:
```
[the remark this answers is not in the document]
provinzkraut: Same as above

Goldziher (approved):

litestar/channels/_redis_xadd_expire.lua
```diff
@@ -0,0 +1,13 @@
+local data = ARGV[1]
+local limit = ARGV[2]
+local exp = ARGV[3]
+local maxlen_approx = ARGV[4]
+
+for i, key in ipairs(KEYS) do
+    if maxlen_approx == 1 then
+        redis.call("XADD", key, "MAXLEN", "~", limit, "*", "data", data, "channel", ARGV[i + 4])
+    else
+        redis.call("XADD", key, "MAXLEN", limit, "*", "data", data, "channel", ARGV[i + 4])
```
provinzkraut, line 10: Can someone more experienced with lua tell me if there's a better way to do this?
I tried using `unpack` to construct a part of the args, but Redis throws a fit then and only seems to accept that when *all* the args are unpacked, which defeats the purpose.

JacobCoffee (approved):

litestar/channels/backends/redis.py
```diff
@@ -0,0 +1,222 @@
...
+
+    async def unsubscribe(self, channels: Iterable[str]) -> None:
+        """Unsubscribe from ``channels``"""
+        self._subscribed_channels = self._subscribed_channels - set(channels)
+
+    async def publish(self, data: bytes, channels: Iterable[str]) -> None:
+        """Publish ``data`` to ``channels``.
+
+        .. note::
+            This operation is performed atomically, using a lua script
```
JacobCoffee, line 164:
```suggestion
            This operation is performed atomically, using a Lua script
```

JacobCoffee (approved):

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+========
+
+**Channels** are a group of related functionalities, built to facilitate the routing
+of event streams, which for example can be used to broadcast messages to WebSocket
+clients.
+
+Channels provide:
+
+1. Independent :term:`broker` backends, optionally handling inter-process communication
+   and data persistence on demand
```
Goldziher, line 13: This looks like it's cut off mid sentence

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+2. "Channel" based subscription management
```
Goldziher, line 14:
```suggestion
2. "Channel" based subscription management.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+3. Subscriber objects as an abstraction over an individualized :term:`event stream`,
+   providing background workers and managed subscriptions
```
Goldziher, line 16:
```suggestion
   providing background workers and managed subscriptions.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+4. Synchronous and asynchronous data publishing
```
Goldziher, line 17:
```suggestion
4. Synchronous and asynchronous data publishing.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+5. Optional history management on a per-channel basis
```
Goldziher, line 18:
```suggestion
5. Optional history management on a per-channel basis.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+6. :doc:`WebSocket </usage/websockets>` integration, generating WebSocket route
+   handlers for an application, to handle the subscription and publishing of incoming
+   events to the connected client
```
Goldziher, line 21:
```suggestion
   events to the connected client.
```

Also, maybe break this point into two sentences, it's a bit long.

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+Basic concepts
+--------------
+
+Utilizing channels involves a few moving parts, of which the most important ones are:
+
+.. glossary::
+
+    event
+        A single piece of data published to, or received from a :term:`backend`, bound
```
Goldziher, line 32:
```suggestion
        A single piece of data published to, or received from a :term:`backend` bound
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        to the channel it was originally published to
```
Goldziher, line 33:
```suggestion
        to the channel it was originally published to.
```

This sentence is a bit elliptical

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+    event stream
+        A stream of :term:`events <event>`, consisting of events from all the channels a
+        :term:`Subscriber` has previously subscribed to
```
Goldziher, line 37:
```suggestion
        :term:`Subscriber` has previously subscribed to.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+    subscriber
+        A :class:`Subscriber <.subscriber.Subscriber>`: An object wrapping an
+        :term:`event stream` and providing access to it through various methods
```
Goldziher, line 41:
```suggestion
        :term:`event stream` and providing easy access to it through various methods.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+    backend
+        A :class:`ChannelsBackend <.backends.base.ChannelsBackend>`. This object
+        manages communication between the plugin and the :term:`broker`, publishing
+        messages to and receiving messages from it. Each plugin instance is associated
+        with exactly one backend.
```
Goldziher, line 47: Make this a note?

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+    broker
+        Responsible for receiving and publishing messages to all connected
+        :term:`backends <backend>`; All backends sharing the same broker will have
+        access to the same messages, allowing for inter-process communication. This is
+        typically handled by a separate entity like `Redis <https://redis.io/>`_
+
+    plugin
+        The :class:`ChannelsPlugin <.plugin.ChannelsPlugin>`, a central instance
```
Goldziher, line 56:
```suggestion
        The :class:`ChannelsPlugin <.plugin.ChannelsPlugin>`, which is a central instance
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        managing :term:`subscribers <subscriber>`, reading messages from the
```
Goldziher, line 57:
```suggestion
        managing :term:`subscribers <subscriber>`; reading messages from the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        :term:`backend`, putting them in the appropriate :term:`event stream` and
```
Goldziher, line 58:
```suggestion
        :term:`backend`; putting them in the appropriate :term:`event stream` and
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        publishing data to the backend
```
Goldziher, line 59:
```suggestion
        publishing data to the backend.
```

This entire point needs some attention. It's a bit complex and condensed

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        Plugin_2 --> Subscriber_6[Subscriber]
+
+
+The ``ChannelsPlugin``
+----------------------
+
+.. currentmodule:: litestar.channels.plugin
+
+The :class:`ChannelsPlugin` acts as the central entity for managing channels and
+subscribers. It's used to publish messages, control how data is stored, manage
```
Goldziher, line 106:
```suggestion
subscribers. It's used to publish messages, control how data is stored, and manage
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+subscribers, route handlers and configuration.
+
+
+.. tip::
+    The plugin makes itself available as a dependency under the ``channels`` key, which
+    means it's not necessary to import it and instead, it can be used from within route
```
Goldziher, line 112:
```suggestion
    means it's not necessary to import it and instead it can be used from within route
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    handlers or other callables within the dependency tree directly
```
Goldziher, line 113:
```suggestion
    handlers or other callables directly.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+
+Configuring the channels
++++++++++++++++++++++++++
+
+The channels manged by the plugin can be either defined upfront, passing them to the
```
Goldziher, line 119:
```suggestion
The channels managed by the plugin can be either defined upfront, passing them to the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+.. code-block:: python
+    :caption: Allowing arbitrary channels
+
+    from litestar.channels import ChannelsPlugin
+
+    channels_plugin = ChannelsPlugin(..., arbitrary_channels_allowed=True)
+
+
+If ``arbitrary_channels_allowed`` is not ``True``, trying to publish or subscribe to a
```
Goldziher, line 140: Make this a note or admonition?

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+This method is non-blocking, even though channels and the associated
+:term:`backends <backend>` are fundamentally asynchronous.
+
+Calling ``publish`` effectively enqueues a message to be sent to the backend, from which
+follows that there's no guarantee that an event will be available in the backend
+immediately after this call.
+
+Alternatively, the asynchronous :meth:`wait_published <ChannelsPlugin.wait_published>`
+method can be used, which  the internal message queue, publishing the data to the
```
Goldziher, line 167:
```suggestion
method can be used, which the internal message queue, publishing the data to the
```

This needs attention. "Which" is not working here

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+backend directly.
+
+.. note::
+    While calling :meth:`publish <ChannelsPlugin.publish>` does not guarantee the
```
Goldziher, line 171:
```suggestion
    Calling :meth:`publish <ChannelsPlugin.publish>` does not guarantee the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    message is sent to the backend immediately, it will be sent there *eventually*; On
```
Goldziher, line 172:
```suggestion
    message will sent to the backend immediately. Rather it will be sent there *eventually*; On
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    shutdown, the plugin will wait for all queues to empty
```
Goldziher, line 173: This is a separate note

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+
+Managing subscriptions
```
Goldziher, line 176:
```suggestion
Managing Subscriptions
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+++++++++++++++++++++++
+
+Another core functionality of the plugin is managing subscriptions, for which two
```
Goldziher, line 179:
```suggestion
Another core functionality of the plugin is managing subscriptions, for which there are two
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+different approaches exist:
```
Goldziher, line 180:
```suggestion
approaches:
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+1. Manually through the :meth:`subscribe <ChannelsPlugin.subscribe>` and
+   :meth:`unsubscribe <ChannelsPlugin.unsubscribe>` methods
```
Goldziher, line 183:
```suggestion
   :meth:`unsubscribe <ChannelsPlugin.unsubscribe>` methods.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+2. By using the :meth:`start_subscription <ChannelsPlugin.start_subscription>` context
+   manager
```
Goldziher, line 185:
```suggestion
   manager.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+Both :meth:`subscribe <ChannelsPlugin.subscribe>` and
+:meth:`start_subscription <ChannelsPlugin.start_subscription>` produce a
+:class:`Subscriber`, which can be used to interact with the streams of events subscribed
+to.
```
Goldziher, line 190:
```suggestion
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+different approaches exist:
+
+1. Manually through the :meth:`subscribe <ChannelsPlugin.subscribe>` and
+   :meth:`unsubscribe <ChannelsPlugin.unsubscribe>` methods
+2. By using the :meth:`start_subscription <ChannelsPlugin.start_subscription>` context
+   manager
+
+Both :meth:`subscribe <ChannelsPlugin.subscribe>` and
+:meth:`start_subscription <ChannelsPlugin.start_subscription>` produce a
+:class:`Subscriber`, which can be used to interact with the streams of events subscribed
```
Goldziher, line 189:
```suggestion
:class:`Subscriber`, which can be used to interact with the streams of events directly. 
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+to.
+
+The context manager should be preferred, and using the methods directly should only be
```
Goldziher, line 192: Admonition? Maybe explain why as well.

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+.. code-block:: python
+
+    async with channels.start_subscription(["foo", "bar"]) as subscriber:
+        ...  # do some stuff here
+        await channels.unsubscribe(subscriber, ["foo"])
+
+
+
+Managing history
```
Goldziher, line 235:
```suggestion
Managing History
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
++++++++++++++++++
+
+Some backends support a per-channel history, keeping a certain amount of
```
Goldziher, line 238:
```suggestion
Some backends support per-channel history, keeping a certain amount of
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+The plugin's :meth:`put_subscriber_history <ChannelsPlugin.put_subscriber_history>` can
+be used to fetch this history and put it into a subscriber's :term:`event stream`.
+
+.. literalinclude:: /examples/channels/put_history.py
+    :language: python
+
+
+.. note::
+    The publication of the history happens sequentially, one channel at a time, one
```
Goldziher, line 250:
```suggestion
    The publication of the history happens sequentially, one channel and one
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    event at a time. This is done to ensure correct ordering of the events and to not
```
Goldziher, line 251:
```suggestion
    event at a time. This is done to ensure the correct ordering of events and to avoid
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    fill up a subscriber's backlog, resulting in dropped history entries. Should the
```
Goldziher, line 252:
```suggestion
    filling up a subscriber's backlog, which would result in dropped history entries. Should the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        `Managing backpressure`
+
+
+The ``Subscriber``
+------------------
+
+.. py:currentmodule:: litestar.channels.subscriber
+
+The :class:`Subscriber` manages an individual :term:`event stream`, provided to it by
+the plugin, representing the sum of events from all channels the subscriber has
```
Goldziher, line 266:
```suggestion
the plugin. The event stream represents the sum of events from all channels the subscriber has
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+subscribed to.
+
+It can be considered the endpoint of all events, while the backends act as the source,
+and the plugin as a router, being responsible for supplying events gathered from the
```
Goldziher, line 270:
```suggestion
and the plugin as a router - being responsible for supplying events gathered from the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+backend into the appropriate subscriber's streams.
+
+In addition to being an abstraction of an :term:`event stream`, the :class:`Subscriber`
+provides different methods to handle this stream:
```
Goldziher, line 274:
```suggestion
provides different methods to interact with it:
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+:meth:`iter_events <Subscriber.iter_events>`
+    An asynchronous generator, producing one event from the stream at a time, waiting
+    until the next one becomes available
+
+:meth:`start_in_background <Subscriber.start_in_background>`
+    Starts a :class:`asyncio.Task` which runs in the background, consuming the
+    :term:`event stream` and invoking a provided callback with the received events
```
Goldziher, line 282:
```suggestion
    :term:`event stream` and invoking a provided callback with the received events.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+:meth:`run_in_background <Subscriber.run_in_background>`
+    A context manager, wrapping
+    :meth:`start_in_background <Subscriber.start_in_background>`. Upon exit, it will
+    attempt a graceful shutdown of the running task, waiting for all currently enqueued
+    events in the stream to be processed. Should the context be left with an error, the
```
Goldziher, line 288:
```suggestion
    events in the stream to be processed. If the context exits with an error, the
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+        It's possible to force the task to stop immediately, by passing ``join=False`` to
+        :meth:`run_in_background <Subscriber.run_in_background>`, which will
+        lead to the cancellation of the task. By default this only happens when the context is
+        left with an exception.
+
+
+.. important::
+    The :term:`events <event>` in the :term:`event streams <event stream>` are always
+    bytes; When calling :meth:`ChannelsPlugin.publish`, data will be serialized before
+    being sent to the backend
```
Goldziher, line 304:
```suggestion
    being sent to the backend.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+
+Consuming the event stream
```
Goldziher, line 307:
```suggestion
Consuming an Event Stream
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
++++++++++++++++++++++++++++
+
+There are two general methods of consuming the :term:`event stream`:
+
+1. By iterating over it directly, using :meth:`iter_events <Subscriber.iter_events>`
```
Goldziher, line 312:
```suggestion
1. :meth:`iter_events <Subscriber.iter_events>`, which iterates over it directly.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+.. important::
+    The :term:`events <event>` in the :term:`event streams <event stream>` are always
+    bytes; When calling :meth:`ChannelsPlugin.publish`, data will be serialized before
+    being sent to the backend
+
+
+Consuming the event stream
++++++++++++++++++++++++++++
+
+There are two general methods of consuming the :term:`event stream`:
```
Goldziher, line 310: Three?

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+1. By iterating over it directly, using :meth:`iter_events <Subscriber.iter_events>`
+2. By using a background task,
+   using :meth:`run_in_background <Subscriber.run_in_background>` or
+   :meth:`start_in_background <Subscriber.start_in_background>`
+
+Iterating over the stream directly is mostly useful if processing the events is the only
+concern, since :meth:`iter_events <Subscriber.iter_events>` is effectively an infinite
+loop. For all other applications, a background task is preferable, which also iterates
```
Goldziher, line 319:
```suggestion
loop. For all other applications, a background task is preferable, which also iterates
```

```suggestion
loop. 

For all other applications, a background task is preferable, which also iterates
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+
+In the above example, the stream is used to send data to a
+:class:`WebSocket <litestar.connection.WebSocket>`.
+
+The same can be achieve by passing
+:meth:`WebbSocket.send_text <litestar.connection.WebSocket.send_text>` as the callback
+to :meth:`run_in_background <Subscriber.run_in_background>`. This will cause the
+WebSocket's method to be invoked every time a new event becomes available in the stream,
+but gives control back to the application, providing an opportunity to perform other
```
Goldziher, line 335: Break this into a separate sentence

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+    Since :exc:`WebSocketDisconnect` is only raised after the corresponding ASGI event
+    has been *received*, it can result in an indefinitely suspended coroutine. This can
+    happen if for example the client disconnects, but no further events are received.
+    The generator will then wait for new events, but since it will never receive any,
+    no ``send`` call on the WebSocket will be made, which in turn means no exception
+    will be raised to break the loop.
+
+
+
+Managing backpressure
```
Goldziher, line 356:
```suggestion
Managing Backpressure
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+---------------------
+
+Each subscriber manages its own backlog: A queue of unprocessed :term:`events <event>`.
+By default, this backlog is unlimited in size, allowing it to grow indefinitely. For
+most applications, this should be no issue, but when the recipient consistently can't
+process messages faster than they come in, an application might opt to handle this case.
+
+The channels plugin provides two different strategies for managing this backpressure:
+
+1. A backoff strategy, dropping newly incoming messages as long as the backlog is full
```
Goldziher, line 366:
```suggestion
1. A backoff strategy, dropping newly incoming messages as long as the backlog is full.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+2. An eviction strategy, dropping the oldest message in the backlog when a new one is
+   added while the backlog is full
```
Goldziher, line 368:
```suggestion
   added while the backlog is full.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+:class:`RedisChannelsPubSubBackend <.redis.RedisChannelsPubSubBackend>`
+    A Redis based backend, using `Pub/Sub <https://redis.io/docs/manual/pubsub/>`_ to
+    delivery messages. This Redis backend has a low latency and overhead and is
+    generally recommended if history is not needed
+
+:class:`RedisChannelsStreamBackend <.redis.RedisChannelsStreamBackend>`
+    A redis based backend, using `streams <https://redis.io/docs/data-types/streams/>`_
+    to deliver messages. It has a slightly higher latency when publishing than the
+    Pub/Sub backend, but achieves the same throughput in message fanout. Recommended
+    when history is needed
```
Goldziher, line 419:
```suggestion
    when history is needed.
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+
+
+Integrating with websocket handlers
```
Goldziher, line 423:
```suggestion
Integrating with Websocket Handlers
```

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+-----------------------------------
+
+
+Generating route handlers
```
Goldziher, line 427:
```suggestion
Generating Route Handlers
```

Goldziher (commented):
Beautiful stuff 😍

docs/usage/channels.rst
provinzkraut: That's because it's not a full sentence. It's a continuation of the "Channels provide" frament, which is continued in the other list items. That's also why there should be no periods at the end. 

This applies to the rest of the document as well.

docs/usage/channels.rst
provinzkraut: This changes the meaning and makes it incorrect. It's a list of different functionalities, the latter items of which are not directly related to the management of subscribers.

docs/usage/channels.rst
provinzkraut: Well, it's a glossary, not an in-depth explanation. I go into more detail about the plugin in later sections.

docs/usage/channels.rst
provinzkraut: That's an important detail though. It does not automagically get injected into *any* callables.

docs/usage/channels.rst
```diff
@@ -0,0 +1,452 @@
...
+
+One of the core aspects of the plugin is publishing data, which is done through its
+:meth:`publish <ChannelsPlugin.publish>` method:
+
+.. code-block:: python
+
+    channels.publish({"message": "Hello"}, "general")
+
+
+The above example will publish the data to the channel ``general``, subsequently putting
```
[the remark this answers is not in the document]
provinzkraut: Sure, but not the one we use. The method is named `publish` after all. The term "putting" is used in this context as well, but only when referring to an event stream.

docs/usage/channels.rst
provinzkraut, line 167:
```suggestion
method can be used, which skips the internal message queue, publishing the data to the
```

docs/usage/channels.rst
provinzkraut: I don't think so. This note talks about when messages are sent to the backend, of which this is a detail. If this is moved the notes ends in talking about the messages being sent there eventually, which then begs the question *when*.

docs/usage/channels.rst
provinzkraut: I don't like this change in combination with the next one. It makes the sentence sound forced and chopped.

docs/usage/channels.rst
provinzkraut: Usual comments about periods

docs/usage/channels.rst
provinzkraut: There is no method to interact with the stream indirectly, so this is a needless clarification.

docs/usage/channels.rst
provinzkraut: See previous comment. There is only one method to interact with the stream, but different ways to handle it.

docs/usage/channels.rst
provinzkraut: Not really. You can either iterate over it or use a background task.

docs/examples/channels/create_route_handlers_send_history.py
```diff
@@ -0,0 +1,13 @@
+from litestar import Litestar
+from litestar.channels import ChannelsPlugin
+from litestar.channels.backends.memory import MemoryChannelsBackend
+
+channels_plugin = ChannelsPlugin(
+    backend=MemoryChannelsBackend(history=10),  # this number should be greater than
+    # or equal to the history to be sent
```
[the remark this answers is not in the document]
provinzkraut: I restructured it

docs/examples/channels/run_in_background.py
```diff
@@ -0,0 +1,19 @@
+from litestar import Litestar, WebSocket, websocket
+from litestar.channels import ChannelsPlugin
+from litestar.channels.backends.memory import MemoryChannelsBackend
+
+
+@websocket("/ws")
+async def handler(socket: WebSocket, channels: ChannelsPlugin) -> None:
+    await socket.accept()
+    async with channels.subscribe(["some_channel"]) as subscriber:
+        async with subscriber.run_in_background(socket.send_text):
```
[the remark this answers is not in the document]
provinzkraut: Could be, but I try to keep the line length down for the docs to make it easier to read on mobile, and multi-line context managers look even worse.

docs/usage/channels.rst
provinzkraut: Fixed

PR #1590, opened 2023-04-27
Implement an alias generator for DTO field names

### LINKED ISSUE #1526: DTO Factory: field renaming strategies
Support renaming strategies that apply to all model fields, e.g., `camelize`.

Given that msgspec natively supports this stuff, perhaps we should try to abstract it to the dto backend with a fallback strategy that we implement.

abdulhaq-e:
### Pull Request Checklist

- [X] New code has 100% test coverage

### Description

This implements an alias generator for DTO field names

Closes #1526 

### Close Issue(s)


peterschutt:
Hey @abdulhaq-e - thanks for this!

This is definitely a feature that we have planned for the DTOs - see #1526 - and pretty much how I foresee it being implemented as well.

What do you think about having some named strategies that are internally implemented, e.g. "upper", "camelize", etc - and a fallback for users to add their own transform function?

litestar/dto/factory/config.py
```diff
@@ -16,8 +16,10 @@ class DTOConfig:
     """Control the generated DTO."""
 
     exclude: AbstractSet[str] = field(default_factory=set)
-    """Explicitly exclude fields from the generated DTO, incompatible with ``include``."""
+    """Explicitly exclude fields from the generated DTO."""
     rename_fields: dict[str, str] = field(default_factory=dict)
     """Mapping of field names, to new name."""
+    fields_alias_generator: Callable[[str], str] | None = None
```
peterschutt, line 22: how about `rename_strategy`?

litestar/dto/factory/config.py
```diff
@@ -16,8 +16,10 @@ class DTOConfig:
     """Control the generated DTO."""
 
     exclude: AbstractSet[str] = field(default_factory=set)
-    """Explicitly exclude fields from the generated DTO, incompatible with ``include``."""
+    """Explicitly exclude fields from the generated DTO."""
```
peterschutt, line 19: Thanks!

abdulhaq-e:
@peterschutt 
I didn't see that thread, I only searched on Discord ^_^. It is interesting that you thought about implementing this as part of the backend, I spent some time trying to do it this way but it wasn't straightforward. Do you think a backend based approach is better?

Yeah I agree with having predefined strategies, makes it even better for the library user. For things like `camelize`, do you think writing a few lines of code is better or adding another external dependency is fine? I am using [this](https://github.com/nficano/humps/blob/master/humps/main.py#L42) in my project but since this is a framework we may want to keep dependencies at a minimum.

peterschutt:
> I spent some time trying to do it this way but it wasn't straightforward

I'm not totally sure how it would look either, and given that we want the feature to be backend agnostic, I think the approach that you've taken here is the right one for now, and we can iterate.

> do you think writing a few lines of code is better

Yes, we should roll our own - we can vendor that code from `humps` with appropriate attribution, or do our own. I would like to ensure that any implementation we introduce has the same behavior as the msgspec equivalent so that we don't have any issues if we do get the renaming strategy implementation pushed down into the backend.

abdulhaq-e:
> has the same behavior as the msgspec equivalent

I will check the msgspec implementation later on. The good thing is that pydantic doesn't offer out of the box implementations so we don't need to worry about that backend.

For now, are you ok with the `RenameStrategies` approach? I like to namespace small utils like this.

peterschutt:
> For now, are you ok with the `RenameStrategies` approach? I like to namespace small utils like this.

Yep, LGTM!

litestar/dto/factory/abc.py
```diff
@@ -225,6 +225,9 @@ def _parse_model(
 
         if rename := config.rename_fields.get(field_definition.name):
             field_definition = field_definition.copy_with(serialization_name=rename)  # noqa: PLW2901
+        elif rename_strategy := config.rename_strategy:
+            alias = RenameStrategies(rename_strategy)(field_definition.name)
+            field_definition = field_definition.copy_with(serialization_name=alias)  # noqa: PLW2901
```
peterschutt, line 230:
```suggestion
        elif config.rename_strategy:
            alias = RenameStrategies(config.rename_strategy)(field_definition.name)
            field_definition = field_definition.copy_with(serialization_name=alias)  # noqa: PLW2901
```
I don't think the walrus is necessary here, although if you do want to reference it, it can be done once before the loop.

litestar/dto/factory/config.py
```diff
@@ -16,8 +18,10 @@ class DTOConfig:
...
     rename_fields: dict[str, str] = field(default_factory=dict)
     """Mapping of field names, to new name."""
+    rename_strategy: RenameStrategy | None = None
+    """A callback for generating aliases of field names. Fields defined in ``rename_fields`` are ignored."""
```
peterschutt, line 25: Can you update this to include the literals please?

litestar/dto/factory/utils.py
```diff
@@ -51,3 +54,47 @@ def parse_configs_from_annotation(parsed_type: ParsedType) -> tuple[DTOConfig, .
...
+    """Useful renaming strategies than be used with :class:`DTOConfig`"""
+
+    def __init__(self, renaming_strategy: RenameStrategy) -> None:
+        self.renaming_strategy = renaming_strategy
+
+    def __call__(self, field_name: str) -> str:
+        if not isinstance(self.renaming_strategy, str):
+            return self.renaming_strategy(field_name)
+
+        return cast(str, self.__getattribute__(self.renaming_strategy)(field_name))
```
peterschutt, line 69:
```suggestion
        return cast(str, getattr(self, self.renaming_strategy)(field_name))
```

litestar/dto/factory/types.py
```diff
@@ -70,3 +70,6 @@ def make_field_type(self, inner_type: type) -> Any:
 
 FieldDefinitionsType: TypeAlias = "Mapping[str, FieldDefinition | NestedFieldDefinition]"
 """Generic representation of names and types."""
+
+RenameStrategy: TypeAlias = Union[Literal["lower", "upper", "camel", "pascal"], Callable[[str], str]]
```
peterschutt, line 74:
```suggestion
RenameStrategy: TypeAlias = 'Literal["lower", "upper", "camel", "pascal"] | Callable[[str], str]'
```
Using forward references means we can use the nicer union syntax.

abdulhaq-e: Neat, thanks. Didn't know that would work in py3.8

abdulhaq-e:
Please don't merge yet, I'm getting an error in my project using renaming and `SQLAlchemyDTO`. I'll try to write a test for that.

peterschutt:
> Please don't merge yet, I'm getting an error in my project using renaming and `SQLAlchemyDTO`. I'll try to write a test for that.

OK no worries, I just sat down to take a look - have you got a traceback or anything you can share?

abdulhaq-e:
I've just added failing tests.

For my project:

`.............src/litestar/litestar/dto/factory/backends/msgspec/utils.py", line 145, in _build_struct_from_model\n    model_val = getattr(model, model_name)\nAttributeError: \'AcademicTermRM\' object has no attribute \'entityId\'. Did you mean: \'entity_id\'?\n'`

peterschutt:
> I've just added failing tests.
> 
> For my project:
> 
> `.............src/litestar/litestar/dto/factory/backends/msgspec/utils.py", line 145, in _build_struct_from_model\n model_val = getattr(model, model_name)\nAttributeError: \'AcademicTermRM\' object has no attribute \'entityId\'. Did you mean: \'entity_id\'?\n'`

OK thanks, taking a look now - this seems to be a different error message to the one that the test is throwing though.

abdulhaq-e:
> > I've just added failing tests.
> > For my project:
> > `.............src/litestar/litestar/dto/factory/backends/msgspec/utils.py", line 145, in _build_struct_from_model\n model_val = getattr(model, model_name)\nAttributeError: \'AcademicTermRM\' object has no attribute \'entityId\'. Did you mean: \'entity_id\'?\n'`
> 
> OK thanks, taking a look now - this seems to be a different error message to the one that the test is throwing though.

Yh I just realised this. The difference is; the test is trying to decode first and it fails. In my project, I've only tried encoding. I'll add another test.

peterschutt:
> > > I've just added failing tests.
> > > For my project:
> > > `.............src/litestar/litestar/dto/factory/backends/msgspec/utils.py", line 145, in _build_struct_from_model\n model_val = getattr(model, model_name)\nAttributeError: \'AcademicTermRM\' object has no attribute \'entityId\'. Did you mean: \'entity_id\'?\n'`
> > 
> > 
> > OK thanks, taking a look now - this seems to be a different error message to the one that the test is throwing though.
> 
> Yh I just realised this. The difference is; the test is trying to decode first and it fails. In my project, I've only tried encoding. I'll add another test.

Want to talk in discord?

peterschutt (approved):
Thanks heaps @abdulhaq-e

Issue #689, opened 2022-10-27
Custom Middleware with a kwarg fails to load


ThinksFast:
**Describe the bug**
I am trying to use `APScheduler` with Starlite.  [APScheduler](https://github.com/agronholm/apscheduler) is a background task scheduling package.  I am currently using their latest alpha version `4.0.0a2`.  They have [a working example for Starlette](https://github.com/agronholm/apscheduler/blob/4.0.0a2/examples/web/asgi_starlette.py) (copied below), but when I adapt the code for Starlite (below), it seems the middleware is never called.  

When I try to debug, I can see the `SchedulerMiddleware` gets initialized, but the `__call__` dunder doesn't seem to get called, I never get to step through the code.

Based on the [Starlite middleware docs](https://starlite-api.github.io/starlite/usage/7-middleware/2-creating-middleware/2-using-middleware-protocol/), I believe I'm calling this middleware correctly [with the kwarg](https://starlite-api.github.io/starlite/usage/7-middleware/1-using-middleware/3-define-middleware/).  

Have I don't something wrong, or is this a bug?

**To Reproduce**
The code below should run as-is.   Only requirements are `Starlite=1.32.0`, `Uvicorn` and `APScheduler=4.0.0a2`. 

Expected result:  print or log statement every second from the `tick()` function.  Actual result is nothing.

```
from __future__ import annotations

import logging
from datetime import datetime

import uvicorn
from apscheduler.datastores.async_sqlalchemy import AsyncSQLAlchemyDataStore
from apscheduler.eventbrokers.asyncpg import AsyncpgEventBroker
from apscheduler.schedulers.async_ import AsyncScheduler
from apscheduler.triggers.interval import IntervalTrigger
from sqlalchemy.ext.asyncio import create_async_engine
from starlite import DefineMiddleware, Request, Starlite, State, get
from starlite.middleware.base import MiddlewareProtocol
from starlite.types import ASGIApp, Receive, Scope, Send

logger = logging.getLogger(__name__)


def tick():
    print("Hello, the time is", datetime.now())
    logger.info("Hello, the time is", datetime.now())


class SchedulerMiddleware(MiddlewareProtocol):
    def __init__(
        self,
        app: ASGIApp,
        scheduler: AsyncScheduler,
    ) -> None:
        super().__init__(app)
        self.app = app
        self.scheduler = scheduler

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        if scope["type"] == "lifespan":
[...]

PR #1635, opened 2023-05-08
Enhancement: Support application lifespan context managers

provinzkraut:
This PR changes how the ASGI lifespan is handled, in particular, it allows to use context managers instead of callables like `on_startup`. This enables to properly handle the startup/shutdown of objects that otherwise would needed to be persisted in a different place, or that don't even make themselves easily available in such a way (think task groups).

Support for a pattern like this is very much needed, since the lack of it is an issue that is currently raised and is also seeping into our own implementations (event emitters and channels backends for example). An example of an implementation problem that could be solved would be #689.

It would allow patterns such as:

```python

@asynccontextmanager
async def background_worker():
  async with anyio.create_task_group() as task_group
    task_group.start_soon(some_worker)
    yield

app = Litestar(..., lifespan=[background_worker])
```

which previously would have to be solved with something like (excluding error handling and other issues this sort of implementation brings)

```python

class Worker:
  def __init__(self):
    self._task = None

  def start():
    self._task = asyncio.create_task(some_worker())

  async def stop():
    self._task.cancel()
    await self._task
    self._task = None


worker = Worker()


app = Litestar(..., on_startup=[worker.start], on_shutdown=[worker.stop])
```

Furthermore, it easily allows us to integrate things like channels backends, which could benefit from such an implementation, without having to create a special case for them in the app; They could simply add their "lifespan manager" to the provided config object in the hook and wouldn't have to worry about the resource management themselves anymore.

<hr>

**Open questions**

This currently keeps all the hooks working as is. We might want to rethink these implementations as well and merge some of them, as they're partly redundant. Should we replace `before_startup`/`after_startup`, `before_shutdown`/`after_shutdown`, `on_startup`/`on_shutdown`? If yes which should be replaced?


litestar/_asgi/asgi_router.py
provinzkraut, line 1: The startup/shutdown logic to handle app events has been moved to the app object, cleaning things up a bit in the router

Goldziher:
I think this is the way to go. But i do think this will require breing more noob friendly - context managers are not as easy to understand as callbacks. 

This raises the question whether we should remove the lifespan hooks and support only this pattern of keep both? I think we might need to keep both. 

Also, what about this option:

```python
app Litestar(...)

@app.lifespan_context
async def handle_lifespan(app: Litestar) -> None:
   # x
   yield
   # y
```

We can even just do

```python
@app
async def ...
```

MatthewNewland:
For what it's worth, I think that adding context manager support without keeping *at least* `on_startup` and `on_shutdown` will be the point where a lot of people throw up their hands and choose a different framework. This seems like an "advanced" feature, not very user friendly unless you're already at least an intermediate Python developer.

provinzkraut:
> I think that adding context manager support without keeping at least `on_startup` and `on_shutdown`

No need to be concerned! As you can see from the conversation here, we pretty much agree on keeping some of those hooks.

> This seems like an "advanced" feature, not very user friendly unless you're already at least an intermediate Python developer.

I do disagree with this however. There are quite a few cases where it can actually be way more user friendly *especially* to an inexperienced developer. In fact, this is one of the motivators for this PR. The question "How to I run x for the lifetime of the app without blocking" or "How to I run x in the background" comes up quite frequently, and wrangling asyncio tasks (which is currently the only solution for this) in a way that deals with their lifetime properly is quite involved. The example I gave in the initial message here showcases this exact scenario. Context managers are easier in this regard because they "just work". You don't need to keep track of anything, care about the lifetime, garbage collection, cancellation and error handling. It's all just done for you.

Furthermore, I'd argue that the cases of using the lifetime hooks where you *do* need to keep some state around (e.g. a connection) far outnumber the ones where you don't. Think of establishing a connection to a database. Not having to store this state elsewhere makes things much simpler, and especially with patterns such as `@asynccontextmanager` around, context managers are not that unwieldy anymore. 

Another point to consider I think here is the ubiquity of context managers. `with open()` being probably the most commonly used one, and something you don't only start to learn at the intermediate level.

Goldziher:
> > I think that adding context manager support without keeping at least `on_startup` and `on_shutdown`
> 
> No need to be concerned! As you can see from the conversation here, we pretty much agree on keeping some of those hooks.
> 
> > This seems like an "advanced" feature, not very user friendly unless you're already at least an intermediate Python developer.
> 
> I do disagree with this however. There are quite a few cases where it can actually be way more user friendly *especially* to an inexperienced developer. In fact, this is one of the motivators for this PR. The question "How to I run x for the lifetime of the app without blocking" or "How to I run x in the background" comes up quite frequently, and wrangling asyncio tasks (which is currently the only solution for this) in a way that deals with their lifetime properly is quite involved. The example I gave in the initial message here showcases this exact scenario. Context managers are easier in this regard because they "just work". You don't need to keep track of anything, care about the lifetime, garbage collection, cancellation and error handling. It's all just done for you.
> 
> Furthermore, I'd argue that the cases of using the lifetime hooks where you *do* need to keep some state around (e.g. a connection) far outnumber the ones where you don't. Think of establishing a connection to a database. Not having to store this state elsewhere makes things much simpler, and especially with patterns such as `@asynccontextmanager` around, context managers are not that unwieldy anymore. 
> 
> Another point to consider I think here is the ubiquity of context managers. `with open()` being probably the most commonly used one, and something you don't only start to learn at the intermediate level. 

Well, let's agree to disagree here - it doesn't really matter since we agree we will need to support both patterns. 

Best are your thoughts on dropping the `on_` handlers or making them an alias with deprecation warning for the respective `after_` hooks? My thinking is that the before / after hooks are both more explicit and more granular.

provinzkraut:
> Best are your thoughts on dropping the `on_` handlers or making them an alias with deprecation warning for the respective `after_` hooks? My thinking is that the before / after hooks are both more explicit and more granular.

So, just from a cleanliness perspective, having as few of these hooks as possible would be nice, which makes me slightly favour the `on_` ones. You have a point with the granularity, but is that something that's actually needed? I think by far the most commonly used one is `on_startup`, and it's the one we also use the most. 

If someone *really* needs to run some shutdown after all other shutdown for example, they can still just subclass `Litestar` and handle it there.

Goldziher:
> > Best are your thoughts on dropping the `on_` handlers or making them an alias with deprecation warning for the respective `after_` hooks? My thinking is that the before / after hooks are both more explicit and more granular.
> 
> So, just from a cleanliness perspective, having as few of these hooks as possible would be nice, which makes me slightly favour the `on_` ones. You have a point with the granularity, but is that something that's actually needed? I think by far the most commonly used one is `on_startup`, and it's the one we also use the most.
> 
> If someone _really_ needs to run some shutdown after all other shutdown for example, they can still just subclass `Litestar` and handle it there.

We have before / after hooks in other places. I dont see why we need to have fewer ones at all. It also is not cleaner IMO. Its more consistent. ON is a reminder from starlette, thats it.

provinzkraut:
> We have before / after hooks in other places

That's a good point. Consistency is more important then.

provinzkraut:
@Goldziher My plan would be to first get this PR merged and then do the removal of the `on_` hooks in a different PR. Since it's a very commonly used hook (we use it as the default hook for our plugins etc. as well) I was thinking on actually deprecating it instead of removing it outright.

provinzkraut:
> I think that's a good approach, and under the hood merge it with `after_startup`

Conceptually, `before_` and `after_` should then refer to the ASGI `startup` and `shutdown` events, right?

But the flow would be

- ASGI `startup` event
- `before_startup`
- `after_startup`
- ASGI `startup.complete`
- ASGI `shutdown` event
- `before_shutdown`
- `after_shutdown`
- ASGI `shutodown.complete` event

Making the distinction between `before_` and `after_` kinda pointless, since it's effectively just different priorities for the hooks.

We also can't for example do `before_shutdown` *before* we receive the shutdown event, and while we *could* do that for `before_startup`, having the hooks tied to the ASGI events in a different order could be all kinds of confusing. 

So.. I'm not really sure what the best way to progress with this is. Should we just explicitly say the `before_` and `after_` are for the same event but different priorities?

provinzkraut:
> I'm not really sure what the best way to progress with this is. Should we just explicitly say the `before_` and `after_` are for the same event but different priorities?

Looking at our own docs, it says

> The `before_shutdown` and `after_shutdown` are basically identical, with the difference being that the callable they receive in callable is invoked respectively before or after the
list of callables defined in the `on_shutdown` list of callables.

So they are only distinct in relation to `on_shutdown`/`on_startup`. If we remove those, the distinction becomes obsolete, leaving us with only one set of hooks, which would then bring us full circle to only having the one, for which `on_startup`/`on_shutdown` makes more sense.

Goldziher:
> > I'm not really sure what the best way to progress with this is. Should we just explicitly say the `before_` and `after_` are for the same event but different priorities?
> 
> Looking at our own docs, it says
> 
> > The `before_shutdown` and `after_shutdown` are basically identical, with the difference being that the callable they receive in callable is invoked respectively before or after the
> list of callables defined in the `on_shutdown` list of callables.
> 
> So they are only distinct in relation to `on_shutdown`/`on_startup`. If we remove those, the distinction becomes obsolete, leaving us with only one set of hooks, which would then bring us full circle to only having the one, for which `on_startup`/`on_shutdown` makes more sense.

Aight, so make `before` and `on` synonyms. No?

provinzkraut:
> Aight, so make before and on synonyms. No?

And keep `after_`?

Goldziher:
> > Aight, so make before and on synonyms. No?
> 
> And keep `after_`?

I think so, why not?

provinzkraut:
> I think so, why not?

Well I'm not sure what the distinction would be between that and `on`? The only difference seems to be that one is called immediately before the other. That's what I meant earlier with the priority. I guess it's fine if we want to keep it like that but then we should also emphasize this in the docs.

litestar/app.py
```diff
@@ -482,6 +491,57 @@ async def __call__(
...
+        `on_startup``, ``before_shutdown``, ``on_shutdown`` and ``after_shutdown``
+        hooks, as well as custom lifespan managers.
+        """
+        async with AsyncExitStack() as exit_stack:
+            exit_callbacks: list[Callable[[], Awaitable[Any]]] = [
+                *[partial(hook, self) for hook in self.after_shutdown[::-1]],
+                *[partial(self._call_lifespan_handler, handler) for handler in self.on_shutdown[::-1]],
+                self.event_emitter.on_shutdown,
+                *[partial(hook, self) for hook in self.before_shutdown[::-1]],
+            ]
```
provinzkraut, line 524: This is a bit awkward currently, but will be cleaned up when we refactor the hook signatures.

provinzkraut:
> This is rather about naming conventions i think. We need to unify a pair of them.

So, our docs state the the purpose of the `before_` and `after_` is, that they run before and respectively after the `on_` hooks. If we remove `on_`, IMO, this does no longer make sense, since "before" would only mean "before the other", which could be done by putting the hook into the same list *before* another, so the functionality can be achieved with a single hook.

I think, we should either keep all three of them, `on_` only, or embrace the priority thing and come up with a different API for that (which could be made compatible with the current one e.g. by adding a wrapper).

Goldziher:
> > This is rather about naming conventions i think. We need to unify a pair of them.
> 
> So, our docs state the the purpose of the `before_` and `after_` is, that they run before and respectively after the `on_` hooks. If we remove `on_`, IMO, this does no longer make sense, since "before" would only mean "before the other", which could be done by putting the hook into the same list _before_ another, so the functionality can be achieved with a single hook.
> 
> I think, we should either keep all three of them, `on_` only, or embrace the priority thing and come up with a different API for that (which could be made compatible with the current one e.g. by adding a wrapper).

Well, looking at the code - 

```python
        for hook in self.app.before_startup:
            await hook(self.app)

        for handler in self.app.on_startup:
            await self._call_lifespan_handler(handler)

        for hook in self.app.after_startup:
            await hook(self.app)
```

I think you are right. This is somewhat redundant. We can achieve the same for both the startup and shutdown hooks with one ordered hook. We will need though to support two different signatures basically.

docs/usage/the-litestar-app.rst
```diff
@@ -68,6 +68,20 @@ establish the connection, and another to close it, and then pass them to the Lit
     :language: python
 
 
+Lifespan context managers
+-------------------------
+
+In addition to the lifespan hooks, Litestar also supports managing the lifespan of an
+application using asynchronous context managers. This can be useful when dealing with
```
JacobCoffee, line 75:
```suggestion
application using `asynchronous context managers <https://bbc.github.io/cloudfit-public-docs/asyncio/asyncio-part-3.html>`_. This can be useful when dealing with
```

JacobCoffee: This was a nice read around that, if we want to include it

Goldziher:
> > We will need though to support two different signatures basically.
> 
> Do we really? I think we talked about this before, how passing the app state is kinda redundant as well, since it's already available on the app instance. 
> 
> I think ideally, we'd only support one signature: A callable that receives the app as its first parameter. Alternatively, we can make this parameter optional, but I feel like having it be mandatory makes things a bit easier. Wdyt?

I think you're right it's redundant but we will break a lot of stuff. Although this is a good time to break stuff. So we can eliminate it as well imo.

JacobCoffee (approved):

peterschutt:
> So they are only distinct in relation to `on_shutdown`/`on_startup`. If we remove those, the distinction becomes obsolete, leaving us with only one set of hooks, which would then bring us full circle to only having the one, for which `on_startup`/`on_shutdown` makes more sense.

+1

> this is a good time to break stuff.

Yeah, lets simplify this stuff down as much as possible while we have the chance.

peterschutt (approved):

provinzkraut:
> Yeah, lets simplify this stuff down as much as possible while we have the chance.

Will do so in a separate PR then!

docs/usage/the-litestar-app.rst
provinzkraut: While a good read indeed, I don't think this belongs here. If we want we can link to the stdlib about those, but since this isn't a tutorial but rather a brief explanation, we assume the reader is already familiar with the topics being discussed.

Issue #1305, opened 2023-03-10
Enhancement: `provide` decorator 


Goldziher:
The idea is to add a new decorator `provide`. This decorator should decorate a route handler, and define a dependency. 

```python
def my_func(request: Request) -> str:
   ...

@provide("dep", my_func)
@get("/")
def handler(dep: str) -> str:
   ...
```

This should be equivalent to this:

```python
def my_func(request: Request) -> str:
   ...

@get("/", dependencies={"dep": Provide(my_func)})
def handler(dep: str) -> str:
   ...
```

Also- the decorator must ensure the proper order. This means the decorator MUST be applied after the route handler decorator (above it) or it will raise an exception. We could do stuff to go around this of course, but I'm not sure we should.

provinzkraut:
Would be a nice addition. Shall we add that to the 2.0 track?

> We could do stuff to go around this of course, but I'm not sure we should.

Working around that is going to be messy with typing and IMO it makes more sense semantically to enforce the "correct" order so I'm on the side of "let's not do that".

Goldziher:
> Shall we add that to the 2.0 track?

yhea lets do this

wassafshahzad:
Is this available to pick up ?

JacobCoffee:
@wassafshahzad yes I believe so, I can assign to you

wassafshahzad:
> @wassafshahzad yes I believe so, I can assign to you

Thank you

wassafshahzad:
@JacobCoffee @Goldziher, I have created the decorator however while running my local test I get the error "Improperly Configured". I do know why I get this, Since router.Register was passed an instance other then RouterHandler. However after RnD I could not figure out how to make the decorator work without changing the original check. Could you guide me a bit.

Goldziher:
> @JacobCoffee @Goldziher, I have created the decorator however while running my local test I get the error "Improperly Configured". I do know why I get this, Since router.Register was passed an instance other then RouterHandler. However after RnD I could not figure out how to make the decorator work without changing the original check. Could you guide me a bit.

Without seeing code we can't help you

provinzkraut:
Something like this was suggested on discord, and I think it would be a nice addition to have:


```python
class SomeController(Controller):
  @provide("foo")
  async def foo_dependency() -> str:
    ...

  @get("/")
  def handler(foo: str) -> str:
    ...
```

which would be equivalent to

```python
class SomeController(Controller):
  async def foo_dependency() -> str:
[...]

PR #1638, opened 2023-05-09
Enhancement: Declaring dependencies without Provide

provinzkraut:
As discussed in #1305, add support for declaring dependencies without using `Provide`. The usage of `Provide` internally remains unchanged, and dependencies declared without `Provide` will simply be wrapped in it upon resolution.

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -252,7 +252,7 @@ class BaseRouteHandler(Generic[T]):
             for layer in self.ownership_layers:
                 for key, value in (layer.dependencies or {}).items():
                     if not isinstance(value, Provide):
-                        value = Provide(value, sync_to_thread=True)  # noqa: PLW2901
+                        value = Provide(value)  # noqa: PLW2901
                     self._validate_dependency_is_unique(
                         dependencies=self._resolved_dependencies, key=key, provider=value
                     )
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, Generic, Mapping, Sequence, TypeVar, cast
 
 from litestar._signature import create_signature_model
 from litestar._signature.field import SignatureField
+from litestar.di import Provide
 from litestar.dto.interface import HandlerContext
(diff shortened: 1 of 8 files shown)


litestar/handlers/base.py
```diff
@@ -251,6 +251,8 @@ def resolve_dependencies(self) -> dict[str, Provide]:
 
             for layer in self.ownership_layers:
                 for key, value in (layer.dependencies or {}).items():
+                    if not isinstance(value, Provide):
+                        value = Provide(value, sync_to_thread=True)  # noqa: PLW2901
```
Goldziher, line 255: `sync_to_thread=True`? This will have a perf impact.

provinzkraut: Well, my thinking here was that this should "just work". Passing in a synchronous function which might block would result in blocking the main thread, which is worse than this performance degradation. It is a pitfall however, basically the opposite one of calling `Provide`. 

Perhaps we should handle `sync_to_thread` differently in general with `Provide`, and for example default to `False`, but give a warning if it's a sync function and it has not *explicitly* been set to `False`. The idea there would be that if a synchronous function is used that does not perform any blocking operation, it can easily just be replaced by an async function, but shooting yourself in the foot by accidentally missing a blocking operation is made harder due to the warning.

Wdyt?

Goldziher: I think we should either require explicit Provide or default to false and warn in the docs that this is the behvaior.

provinzkraut: I really don't feel like warning in the docs is enough in relationship to the size of this footgun. What are your concerns with a runtime warning?

Goldziher: i dunno, maybe its better - we simply dont have these in other places yet, but thats not a real reason tbh. So lets go with your suggestion.

provinzkraut: > we simply dont have these in other places yet, but thats not a real reason tbh

We should add more of those I think :upside_down_face:

provinzkraut: Implemented the warning functionality in #1648

winstxnhdw:
Was this scrapped?

PR #1641, opened 2023-05-09
Storage expiry bugfix

### LINKED ISSUE #1640: Bug: StorageObject doesn't return < 0 when using expiry
### Description

When the stored value is expired, the returned interval is set to 86400 and will therefore not expire. 

https://github.com/litestar-org/litestar/blob/main/litestar/stores/base.py#L122

### MCVE

```python
from pathlib import Path
from litestar.stores.file import FileStore

store = FileStore(path=Path("test.db"))

async def setstore() -> None:
  await store.set("test", "value", expires_in=5)
  return None

async def getstore() -> int:
  expiry = await store.expires_in("test")
  return expiry
```

### Steps to reproduce

_No response_

### Screenshots

```bash
"![expiry](https://cdn.discordapp.com/attachments/919193495690936353/1105483460677734501/image.png)"
```

### Logs

_No response_

### Litestar Version

`litestar==2.0.0a5`

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows
- [X] Other (Please specify in the description above)

SanderWegter:
### Pull Request Checklist

### Description

This PR fixes the bug in the StorageObject `expires_in` return value as described by #1640. 

### Close Issue(s)

Fixes #1640


provinzkraut:
@SanderWegter I think you might need to rebase

tests/test_stores.py
```diff
@@ -159,6 +160,9 @@ async def test_expires_in(store: Store) -> None:
     await store.set("foo", "bar", expires_in=10)
     assert math.ceil(await store.expires_in("foo") / 10) * 10 == 10  # type: ignore[operator]
 
+    frozen_datetime.tick(12)
+    assert await store.expires_in("foo") == -2
```
provinzkraut, line 164: So the reason this fails is because of a different bug. The redis backend is actually the only one behaving correctly here. 

If `expires_in` is called on an item that does not exist, it should return `None`. Since the item already expired here, Redis won't give us its TTL, since it was removed from memory.

The `expires_in` implementations of the other stores don't not go through the regular `get` method, which would not return an already expired item.

I suggest you xfail this for the other stores, since the test is actually correct, and we fix this in a different PR.

provinzkraut (approved):

provinzkraut (approved):

PR #1648, opened 2023-05-10
Enhancement: Warn about sync dependencies without an explicit `sync_to_thread` value

provinzkraut:
As discussed in https://github.com/litestar-org/litestar/pull/1638#discussion_r1188482488, this changes the functionality of `Provide` to issue a warning if used with a synchronous callable without setting `sync_to_thread` explicitly. 

This is meant to prevent pitfalls where a user might pass a synchronous dependency which is blocking and inadvertently blocks their whole application. Since the other option would be to use `sync_to_thread` by default, which would introduce a performance loss in cases where this isn't needed, simply issuing a warning is seen as a reasonable approach.

### DIFF
--- a/docs/examples/application_state/using_application_state.py
+++ b/docs/examples/application_state/using_application_state.py
@@ -27,7 +27,7 @@ def middleware_factory(*, app: "ASGIApp") -> "ASGIApp":
     return my_middleware
 
 
-def my_dependency(state: State) -> Any:
+async def my_dependency(state: State) -> Any:
     """Dependencies can receive state via injection."""
     logger.info("state value in dependency: %s", state.value)
 
--- a/docs/examples/dependency_injection/dependency_skip_validation.py
+++ b/docs/examples/dependency_injection/dependency_skip_validation.py
@@ -7,7 +7,7 @@ from litestar.di import Provide
 from litestar.params import Dependency
 
 
-def provide_str() -> str:
+async def provide_str() -> str:
     """Returns a string."""
     return "whoops"
 
--- a/docs/usage/dependency-injection.rst
+++ b/docs/usage/dependency-injection.rst
@@ -10,19 +10,19 @@ the application:
    from litestar.di import Provide
 
 
-   def bool_fn() -> bool:
+   async def bool_fn() -> bool:
        ...
 
 
-   def dict_fn() -> dict:
+   async def dict_fn() -> dict:
        ...
 
 
-   def list_fn() -> list:
+   async def list_fn() -> list:
        ...
 
 
-   def int_fn() -> int:
+   async def int_fn() -> int:
        ...
 
 
@@ -61,6 +61,21 @@ The above example illustrates how dependencies are declared on the different lay
 Dependencies can be either callables - sync or async functions, methods or class instances that implement the
 :meth:`object.__call__` method, or classes. These are in turn wrapped inside an instance of the
 :class:`Provide <.di.Provide>` class.
+
+.. admonition:: Sync vs. Async dependencies
+    :class: important
+
+    Litestar supports both synchronous and asynchronous dependencies. To ensure
+    To ensure synchronous dependencies don't block the main thread and therefore the
+    entire application (for example when they perform blocking I/O), the parameter
+    ``sync_to_thread`` will cause them to be run in a thread pool. If a synchronous
+    function is non-blocking, it should be considered to make it an async function
+    instead.
+
+.. tip::
+    Litestar will warn about the usage of synchronous dependency functions which don't
+    have ``sync_to_thread`` set.
+
 
 Pre-requisites and Scope
 ------------------------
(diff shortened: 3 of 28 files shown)


docs/examples/dependency_injection/dependency_skip_validation.py
```diff
@@ -7,7 +7,7 @@
 from litestar.params import Dependency
 
 
-def provide_str() -> str:
+async def provide_str() -> str:
     """Returns a string."""
     return "whoops"
```
Goldziher, line 12: I think it's a problem to have async non-blocking examples like this, since these are not good examples of using async.

docs/examples/application_state/using_application_state.py
```diff
@@ -27,7 +27,7 @@ async def my_middleware(scope: "Scope", receive: "Receive", send: "Send") -> Non
     return my_middleware
 
 
-def my_dependency(state: State) -> Any:
+async def my_dependency(state: State) -> Any:
     """Dependencies can receive state via injection."""
     logger.info("state value in dependency: %s", state.value)
```
Goldziher, line 32: Why convert it to async? Just wrap it in provide in the examples ), except one case that demoes the use without Provide.

docs/usage/dependency-injection.rst
```diff
@@ -10,19 +10,19 @@ the application:
    from litestar.di import Provide
 
 
-   def bool_fn() -> bool:
+   async def bool_fn() -> bool:
```
Goldziher, line 13: This file is a good place to demo the different patterns and explain them

docs/examples/dependency_injection/dependency_skip_validation.py
provinzkraut: You're right, it should not go without an explanation. How do you feel about leaving them like this now and updating the docs later? 

This whole chapter could use a re-write, so adding a section that goes a bit more in-depth about async dependencies and when and why, examples included, would fit well in there.

I'd like to not do this within the scope of this PR though if possible.

Goldziher: sure

Goldziher (approved):

docs/usage/dependency-injection.rst
```diff
@@ -62,6 +62,21 @@ Dependencies can be either callables - sync or async functions, methods or class
...
+    Litestar supports both synchronous and asynchronous dependencies. To ensure
+    To ensure synchronous dependencies don't block the main thread and therefore the
+    entire application (for example when they perform blocking I/O), the parameter
+    ``sync_to_thread`` will cause them to be run in a thread pool. If a synchronous
+    function is non-blocking, it should be considered to make it an async function
+    instead.
+
+.. tip::
+    Litestar will warn about the usage of synchronous dependency functions which don't
+    have ``sync_to_thread`` set.
```
JacobCoffee, line 77:
```suggestion
    Litestar will warn about the usage of synchronous dependency functions which do not
    have ``sync_to_thread`` set.
```

docs/usage/dependency-injection.rst
```diff
@@ -62,6 +62,21 @@ Dependencies can be either callables - sync or async functions, methods or class
...
 
+.. admonition:: Sync vs. Async dependencies
+    :class: important
+
+    Litestar supports both synchronous and asynchronous dependencies. To ensure
+    To ensure synchronous dependencies don't block the main thread and therefore the
+    entire application (for example when they perform blocking I/O), the parameter
+    ``sync_to_thread`` will cause them to be run in a thread pool. If a synchronous
+    function is non-blocking, it should be considered to make it an async function
+    instead.
```
JacobCoffee, line 73:
```suggestion
.. admonition:: Sync vs. Async dependencies
    :class: important

    Litestar supports both **synchronous** and **asynchronous** dependencies.
    To ensure **synchronous** dependencies don't block the main thread
    - *and therefore the entire application, for example when they perform blocking I/O* -
    the parameter ``sync_to_thread`` will ensure they are run in a thread pool.
    If a **synchronous** function is non-blocking, you should consider making it an async function instead.
```

litestar/di.py
```diff
@@ -49,7 +50,17 @@ def __init__(
...
         self.dependency = Ref["AnyCallable"](dependency)
         self.has_sync_callable = isclass(dependency) or not is_async_callable(dependency)
-        self.sync_to_thread = sync_to_thread
+        if self.has_sync_callable and sync_to_thread is None:
+            warnings.warn(
+                "Using a synchronous callable with Provide might block the main thread "
+                "if the callable performs blocking operations and sync_to_thread is not"
+                " set to True. If the callable is non-blocking, either make it "
+                "asynchronous or set sync_to_thread=False explicitly to silence this "
+                "warning.",
```
JacobCoffee, line 59:
```suggestion
                "Using a synchronous callable with Provide might block the main thread "
                "if the callable performs blocking operations and sync_to_thread is not "
                "set to True. If the callable is non-blocking, either make it "
                "asynchronous or set sync_to_thread=False explicitly to silence this "
                "warning.",
```

JacobCoffee (approved):

peterschutt (approved):
Nothing to add from me, beyond what the others have already mentioned.

PR #1650, opened 2023-05-11
Feature: `DTOData` datastructure.

### LINKED ISSUE #1579: RFC: Enhancement: DTO Partial Support
### Summary

# v2  `Partial` support

Starlite `Partial` uses `__class_getitem__()` to create a new model of the generic argument type with all fields made optional, and returns that from `__class_getitem__()`. There's two facets of this I'd like to change:

1. returning a totally different type from `__class_getitem__()` isn't ideal as static type checkers will still think the thing is an instance of `Partial`.
2. `None` used as missing sentinel doesn't allow to discriminate where an optional field with no default value has been omitted from client data.

## Proposal

We have a new `Partial` type, that is still the interface for supporting routes that can accept partial representations of entities, but less magic.

### Proposed usage example

```py
model_instance = MyModel()

@patch(dto=MyDTO)
def patch_handler(data: Partial[MyModel]) -> MyModel:
    assert isinstance(data, Partial)  # we do inject an instance of partial
    validated_data: dict[str, Any] = data.as_builtins()  # validated but unstructured data available
    new_instance: MyModel = data.create_instance()  # utility that will create an instance of `MyModel` from the validated client data
    return data.update_instance(model_instance)  # utility to do a `setattr` of the inbound client data onto some `MyModel` instance
```

### Partial methods

- `create_instance()`/`update_instance()` methods are there because there is no guarantee that the underlying library that we are abstracting away will even allow to instantiate the model type without all required attributes (SQLAlchemy does allow this, but I imagine most won't), and no guarantee that the client has to send complete enough data to instantiate an instance on every request. So, a pattern where the instance needs to be queried first and subsequently updated from the raw data should be generally more useful. 
- `as_builtins()`: Making unstructured, but validated data available to the developer means that they can roll their own solution to how the partial updates should be handled.

### Partial and AbstractDTOFactory

When the DTO receives the handler `"data"` kwarg type on startup, it can  detect that it is representing a `Partial` type, and build the DTO backend to suit it. Every model field that is built by the backed will be a union of its type and `EmptyType` with a default value of `Empty`. Any `Empty` values encountered when accessing one of the `Partial` methods documented above, will be omitted.

### Partial and OpenAPI

The DTO that represents the partial type should be able to produce an openapi schema that removes any `EmptyType` unions.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

peterschutt:
A datastructure for interacting with DTO validated data in its unstructured form.

This utility is to support the case where the amount of data that is available from the client request is not complete enough to instantiate an instance of the model that would otherwise be injected.

```py

from uuid import UUID, uuid4

from litestar import post
from litestar.dto.factory import DTOConfig, DTOData
from litestar.dto.factory.stdlib.dataclass import DataclassDTO

@dataclass
class Person:
    id: UUID
    name: str
    age: int

class WriteDTO(DataclassDTO[Person]):
    """Doesn't allow the client to set the id."""

    config = DTOConfig(exclude={"id"})

@post("/person", dto=WriteDTO, return_dto=None)
def create_person(data: DTOData[Person]) -> Person:
    """Create a person."""
    return data.create_instance(id=uuid4())
```

Closes #1579 

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/docs/usage/dto/1-dto-factory.rst
+++ b/docs/usage/dto/1-dto-factory.rst
@@ -134,3 +134,38 @@ is injected for our data parameter (line 35).
 We then add a ``B`` instance to the data (line 39), which includes a reference back to ``a``, and from inspection of the
 return data can see that ``b`` is included in the response data, however ``b.a`` is not, due to the default
 ``max_nested_depth`` of ``1``.
+
+DTO Data
+--------
+
+Sometimes we need to be able to access the data that has been parsed and validated by the DTO, but not converted into
+an instance of our data model.
+
+In the following example, we create a DTO that doesn't allow clients to set the ``id`` field on the ``Person`` model and
+declare it on the handler.
+
+.. literalinclude:: /examples/data_transfer_objects/factory/dto_data_problem_statement.py
+    :language: python
+    :emphasize-lines: 18,19,20,21,27
+    :linenos:
+
+Notice that we get a ``500`` response from the handler - this is because the DTO has attempted to convert the request
+data into a ``Person`` object and failed because it has no value for the required ``id`` field.
+
+One way to handle this is to create different models, e.g., we might create a ``CreatePerson`` model that has no ``id``
+field, and decode the client data into that. However, this method can become quite cumbersome when we have a lot of
+variability in the data that we accept from clients, for example, PATCH requests.
+
+This is where the :class:`DTOData <litestar.dto.factory.DTOData>` class comes in. It is a generic class that accepts the
+type of the data that it will contain, and provides useful methods for interacting with that data.
+
+.. literalinclude:: /examples/data_transfer_objects/factory/dto_data_usage.py
+    :language: python
+    :emphasize-lines: 7,25,27
+    :linenos:
+
+In the above example, we've injected an instance of :class:`DTOData <litestar.dto.factory.DTOData>` into our handler,
+and have used that to create our ``Person`` instance, after augmenting the client data with a server generated ``id``
+value.
+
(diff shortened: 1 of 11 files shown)


docs/usage/dto/1-dto-factory.rst
```diff
@@ -134,3 +134,38 @@ is injected for our data parameter (line 35).
...
+    :language: python
+    :emphasize-lines: 18,19,20,21,27
+    :linenos:
+
+Notice that we get a ``500`` response from the handler - this is because the DTO has attempted to convert the request
+data into a ``Person`` object and failed because it has no value for the required ``id`` field.
+
+One way to handle this is to create different models, e.g., we might create a ``CreatePerson`` model that has no ``id``
+field, and decode the client data into that. However, this method can become quite cumbersome when we have a lot of
+variability in the data that we accept from clients, for example, PATCH requests.
```
JacobCoffee, line 157:
```suggestion
variability in the data that we accept from clients, for example, 
`PATCH <https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH>`_ requests.
```

JacobCoffee (approved):
I think the same for the `POST` statement above (out of pr scope)

docs/usage/dto/1-dto-factory.rst
peterschutt: Thanks!

peterschutt: Fix for this is in #1651

PR #1655, opened 2023-05-12
Enhancement: Warn about sync handler functions without an explicit `sync_to_thread` value

provinzkraut:
As a continuation of #1648, which introduced a warning about the implicit usage of `sync_to_thread` in `Provide`, this adds the same warning for route handlers. 

Additionally this PR:

- Introduces an environment variable to toggle the emitted warnings: `LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD`
- Adds a new "topics" section to the documentation
- Adds a "Sync vs Async" article to the topics section, talking about basic considerations regarding the different modes of execution and when they are appropriate
- Adjusts the documentation of route handlers and `Provide` to reference that article
- Adjusts all examples to explicitly use `sync_to_thread=False` where appropriate
- Refactors a bunch of tests which did act on globally scoped objects and could therefore not be targeted by the setting of the `LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD` variable reliably

**Considerations**

Introducing this warning and including `sync_to_thread` by default in our documentation does not come without its downsides. The main concern is that it adds unnecessary noise, possibly making examples harder to follow. However, after a lengthy conversation with @peterschutt and weighing of the options, we agreed that handling it this way would be the best option overall.

The fact that Litestar does not implicitly run synchronous functions in a thread is a core design decision, and what allows us to be significantly faster for operations that are truly non-blocking, even being able to avoid async overhead. This is a detail that should be made explicit in the docs, and is simply the trade off being made in favour of more granular control over the mode of execution and better performance out of the box.

The impact of this on user code bases will be minimal, since this warning can be turned off at will. 

**Alternatives**

An alternative implementation would be to turn off the warning by default, which would then however place the burden of knowledge upon the user, making it easier to accidentally block their entire application by using a synchronous callable. Since this is exactly what the warning introduced here is trying to prevent, it was decided that this was not a viable option.

### DIFF
--- a/docs/topics/index.rst
+++ b/docs/topics/index.rst
@@ -1,7 +1,7 @@
 Topics
 ======
 
-A collection of articles talking about technical concepts, background information,
+A collection of articles about technical concepts, background information,
 design decisions and similar topics on a higher level.
 
 
--- a/docs/topics/index.rst
+++ b/docs/topics/index.rst
@@ -1,0 +1,10 @@
+Topics
+======
+
+A collection of articles talking about technical concepts, background information,
+design decisions and similar topics on a higher level.
+
+
+.. toctree::
+
+    sync-vs-async
--- a/docs/topics/sync-vs-async.rst
+++ b/docs/topics/sync-vs-async.rst
@@ -10,8 +10,7 @@ supported:
 - Running synchronous callables directly
 - Running synchronous callables in a thread pool
 
-There are a few important differences between those, which the following article tries
-to give an overview of.
+This article gives an overview of important differences between these modes.
 
 
 Blocking and non-blocking
@@ -33,7 +32,7 @@ would. More importantly, anything that happens inside an asynchronous function
 Technically speaking, this means that there are no non-blocking functions in Python,
 since async functions only "unblock" at each ``await``. As this is not a useful
 definition of the term, "blocking" usually refers to callables that *block for a long
-time*. long period of time*.
+time*.
 
 .. note::
 
--- a/docs/topics/sync-vs-async.rst
+++ b/docs/topics/sync-vs-async.rst
@@ -1,0 +1,147 @@
+Sync vs. Async
+==============
+
+
+Litestar supports synchronous as well as asynchronous callables in almost all places
+where it's possible to do so. In general, three different modes of execution are
+supported:
+
+- Running asynchronous callables directly
+- Running synchronous callables directly
+- Running synchronous callables in a thread pool
+
+There are a few important differences between those, which the following article tries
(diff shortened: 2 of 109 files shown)


docs/topics/index.rst
```diff
@@ -0,0 +1,10 @@
+Topics
+======
+
+A collection of articles talking about technical concepts, background information,
```
peterschutt, line 4:
```suggestion
A collection of articles about technical concepts, background information,
```

docs/topics/sync-vs-async.rst
```diff
@@ -0,0 +1,147 @@
...
+Litestar supports synchronous as well as asynchronous callables in almost all places
+where it's possible to do so. In general, three different modes of execution are
+supported:
+
+- Running asynchronous callables directly
+- Running synchronous callables directly
+- Running synchronous callables in a thread pool
+
+There are a few important differences between those, which the following article tries
+to give an overview of.
```
peterschutt, line 14:
```suggestion
This article gives an overview of important differences between these modes.
```

docs/topics/sync-vs-async.rst
```diff
@@ -0,0 +1,147 @@
...
+``await`` keyword.  This is a very important aspect to note, since an async function
+that never calls ``await`` and, for example, performs a computationally intensive task,
+*will* block the main thread for its entire runtime, just like a synchronous function
+would. More importantly, anything that happens inside an asynchronous function
+*between* the parts where it waits will also be blocked.
+
+Technically speaking, this means that there are no non-blocking functions in Python,
+since async functions only "unblock" at each ``await``. As this is not a useful
+definition of the term, "blocking" usually refers to callables that *block for a long
+time*. long period of time*.
```
peterschutt, line 36:
```suggestion
time*.
```

litestar/utils/warnings.py
```diff
@@ -0,0 +1,21 @@
...
+from litestar.types import AnyCallable
+
+
+def warn_implicit_sync_to_thread(source: AnyCallable, stacklevel: int = 2) -> None:
+    if os.getenv("LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD") == "0":
+        return
+
+    warnings.warn(
+        f"Use of a synchronous callable {source} without setting sync_to_thread is "
+        "discouraged. Since synchronous callables can block the main thread if they "
```
peterschutt, line 14:
```suggestion
        "discouraged since synchronous callables can block the main thread if they "
```

peterschutt (approved):
Great stuff!

PR #1656, opened 2023-05-12
Fix channels repeated unsubscribe bug

provinzkraut:
Fix a bug that would raise an error when repeatedly subscribing to and unsubscribing from the same channel while `arbitrary_channels_allowed=False`.


JacobCoffee (approved):
Hey this is that thing!

peterschutt (approved):

Goldziher (approved):

PR #1662, opened 2023-05-13
Infra: Set a timout for test workflow

provinzkraut:
Recently we've been introducing more integration tests that might hang indefinitely (behaviour observed multiple times already) due to a misconfigured tests, resource cleanup issues, issues with the docker containers being used etc. This PR introduces a hard limit of 10 minutes for the tests workflow, after which it will be cancelled.


peterschutt (approved):
Good idea

Issue #1634, opened 2023-05-08
Enhancement: Unify ASGI and application hooks


provinzkraut:
Currently there's two different sets of "lifespan" callables:

1. `before_shutdown` / `after_shutdown` and `before_startup` / `after_startup`
2.  `on_shutdown` and `on_startup`

The `on_` all receive the application as a positional arguments, while the `before_` / `after_` hooks optionally receive the application state. There is no actual need for this distinction, and the `before_` / `after_` hooks should also just receive the application instance.

Goldziher:
Whats preventing us from removing the `on_` events?

provinzkraut:
#1635

provinzkraut:
Closed in #1663

PR #1663, opened 2023-05-13
Refactor: Unify lifespan hooks

provinzkraut:
Implements #1634, as discussed in #1635.

The following application hooks were removed:

- `before_startup`
- `after_startup`
- `before_shutdown`
- `after_shutdown`

The hooks `on_startup` and `on_shutdown` have been altered to optionally accept an application instance as their only argument. The decision to keep this detail around was made to more easily allow to re-use existing callables and pass them to the hooks, without having to wrap them in another callable which accepts a positional argument.

In addition, the `lifespan` argument to `Litestar` et. al was changed to allow instances of `AbstractAsyncContextManager` alongside callables which return such context managers. This was done to allow e.g. plugins to inject themselves as a lifespan manager.


peterschutt (approved):
Very nice!

PR #1677, opened 2023-05-16
Exposes `ParsedType` as public api.

### LINKED ISSUE #1567: RFC: SerializationPluginProtocol.supports_type() argument
### Summary

The current `SerializationPluginProtocol.supports_type()` method receives an instance of the `ParsedType` signature util, which isn't exposed as public API.

We can either:
1. expose the `ParsedType` interface as public api
2. pass the annotation and force user implementations to do their own parsing of the annotation (we can then just use `ParsedType` for our own internal implementations.
3. try to unwrap any complex types to a single type to pass to the method (e.g., pass `Model` for `List[Model]`, etc) - but I'm not too keen on this as theoretically someone could define a protocol to manage `List[Model]`, where `Model` would be unsupported. Other examples might be `Union[ModelA, ModelB]`, we shouldn't be making a call on which type should be passed through in that case. So, this option is my least favorite.

Those suggestions are in order of preference IMO.

### Basic Example

1.
```py
def supports_type(parsed_type: ParsedType) -> bool:
```
This allows users to leverage the logic we have for parsing complex type annotations and inspecting inner types, etc.

2.
```py
def supports_type(annotation: Any) -> bool:
```
This puts the most work back onto the user. They need to handle unwrapping `Annotated`, recursively parsing inner types etc, which is all stuff that we've done.

3.
```py
def supports_type(simple_type: type[Any]) -> bool:
```
If annotation is `List[Model]` we'd pass `Model`, but this is too limiting on what users can do with the protocol IMO

### Drawbacks and Impact

By exposing `ParsedType` we impose stability constraints of the API of such, and being a fairly new type in the library, it may be early to do that. However, we are using it quite extensively already, so stability will also be of benefit to us.

### Unresolved questions

Other solutions?

peterschutt:
Motivation is described in #1567.

Closes #1567

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -39,3 +39,4 @@ API reference
     template
     testing
     types
+    typing
--- a/docs/reference/typing.rst
+++ b/docs/reference/typing.rst
@@ -1,0 +1,7 @@
+typing
+======
+
+.. py:currentmodule:: litestar.typing
+
+.. automodule:: litestar.typing
+   :members:
--- a/litestar/contrib/sqlalchemy/dto.py
+++ b/litestar/contrib/sqlalchemy/dto.py
@@ -19,8 +19,7 @@ if TYPE_CHECKING:
     from sqlalchemy.orm import RelationshipProperty
     from typing_extensions import TypeAlias
 
-    from litestar.utils.signature import ParsedType
-
+    from litestar.typing import ParsedType
 
 __all__ = ("SQLAlchemyDTO", "DataT")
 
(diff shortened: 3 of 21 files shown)


peterschutt:
I chose to add `litestar.typing` instead of documenting `ParsedType` in its original localtion as we aren't documenting _anything_ in the `utils` package - and so I though better to leave it that way.

JacobCoffee (approved):

PR #1695, opened 2023-05-19
feat: allows JWT tokens to have extra attributes specified. 

cofin:
### Pull Request Checklist

### Description

Without this fix, the user most override the Token and the decode method if the JWT token contains parameters not defined.  This fix moves any extra attributes on the token to an `_extras` attribute.

### Close Issue(s)


litestar/contrib/jwt/jwt_token.py
```diff
@@ -82,7 +85,12 @@ def decode(encoded_token: str, secret: str | dict[str, str], algorithm: str) ->
             payload = jwt.decode(token=encoded_token, key=secret, algorithms=[algorithm], options={"verify_aud": False})
             exp = datetime.fromtimestamp(payload.pop("exp"), tz=timezone.utc)
             iat = datetime.fromtimestamp(payload.pop("iat"), tz=timezone.utc)
-            return Token(exp=exp, iat=iat, **payload)
+            field_names = {f.name for f in dataclasses.fields(Token)}
+            extra_fields = set(payload.keys()).difference(field_names)
+            extras = {}
+            for key in extra_fields:
+                extras[key] = payload.pop(key)
+            return Token(exp=exp, iat=iat, **payload, _extras=extras)
```
peterschutt, line 93: Not really related - but this method looks like it could be a `@classmethod`, typed to return `Self` and `return cls(...)` etc.

litestar/contrib/jwt/jwt_token.py
```diff
@@ -42,6 +43,8 @@ class Token:
     """Audience - intended audience."""
     jti: str | None = field(default=None)
     """JWT ID - a unique identifier of the JWT between different issuers."""
+    _extras: dict[str, Any] | None = field(default=None)
```
peterschutt, line 46: Sorry if a dumb question - looking at this without any context. Are the extra fields intended to be accessed by users in any way?

Also, can you use `default_factory=dict` and not require the optional union for the type?

cofin: Yep, I can do the `default_factory` for sure.  

Re: the first question.  The answer is maybe. The JWT may contain any arbitrary claims, but they may only be used in certain contexts or may not be relevant for the application.  

As an example, [here](https://developers.google.com/identity/sign-in/web/backend-auth) is what the JWT token for an IAP authenticated request from Google.  These can change over time and could be different based on the defined oauth scopes.  

In my case, I only need 1 of the extra fields, but the bigger issue is that an exception is raised when I try to create a `Token` class when it has these extra fields.

peterschutt: > but the bigger issue is that an exception is raised when I try to create a Token class when it has these extra fields

Understood. Basically I'm querying whether the `_extra` attribute should have the leading underscore or not, as that implies to users that they shouldn't be accessing the collection.

cofin: It probably should not have the leading underscore since the user may need to get to these attributes.

litestar/contrib/jwt/jwt_token.py
```diff
@@ -63,8 +66,8 @@ def __post_init__(self) -> None:
         else:
             raise ImproperlyConfiguredException("iat must be a current or past time")
 
-    @staticmethod
-    def decode(encoded_token: str, secret: str | dict[str, str], algorithm: str) -> Token:
+    @classmethod
+    def decode(cls, encoded_token: str, secret: str | dict[str, str], algorithm: str) -> Token:
```
peterschutt, line 70:
```suggestion
    def decode(cls, encoded_token: str, secret: str | dict[str, str], algorithm: str) -> Self:
```

Import `Self` from `typing_extensions`. The reason this is important is that if someone subclasses `Token`, the current implementation will not type this method as returning an instance of the subtype, even though it does.

cofin:
I think this covered all of the open items on this.  It's ready for a final review.

litestar/contrib/jwt/jwt_token.py
```diff
@@ -82,7 +89,12 @@ def decode(encoded_token: str, secret: str | dict[str, str], algorithm: str) ->
             payload = jwt.decode(token=encoded_token, key=secret, algorithms=[algorithm], options={"verify_aud": False})
             exp = datetime.fromtimestamp(payload.pop("exp"), tz=timezone.utc)
             iat = datetime.fromtimestamp(payload.pop("iat"), tz=timezone.utc)
-            return Token(exp=exp, iat=iat, **payload)
+            field_names = {f.name for f in dataclasses.fields(Token)}
+            extra_fields = set(payload.keys()).difference(field_names)
```
provinzkraut, line 93:
```suggestion
            extra_fields = payload.keys() - field_names
```

that should work without creating a set explicitly.

provinzkraut (approved):

PR #1698, opened 2023-05-20
Infra: Fix run examples timeout

provinzkraut:
Attempt to mitigate the various timeout related application example failures during the documentation building process.

1. Fix a bug that would lead to application servers not being killed properly
2. Cycle through ports rather than re-use them immediately
3. Ignore missing outputs when building the preview


peterschutt:
> Documentation preview will be available shortly at https://litestar-org.github.io/litestar-docs-preview/1698

Oh this is shiny!

peterschutt (approved):

PR #1716, opened 2023-05-23
fix: don't emit sync to thread warnings for generator  dependencies

### LINKED ISSUE #1711: Bug: #1648/#1664 `sync_to_thread` warning shows on `async`-defined callables.
### Description

> I may be confused at the inner working here, but @provinzkraut said it was potentially a bug and to report it.

#1648 and #1664 introduced a new warning, but I am seeing it when all of the pieces along the way are in fact defined with `async def`.

I may be missing something here, though...

This is coming from a private repo, but the relevant bits are syntactically similar to [litestar-fullstack](https://github.com/cofin/litestar-fullstack/blob/81d33439a87d44ab1dcd93b8d824f83ef37f0658/src/app/lib/aiosql/service.py#L14) as @cofin and I worked through this and then I applied the same changes privately

- [Query Manager](https://github.com/cofin/litestar-fullstack/blob/81d33439a87d44ab1dcd93b8d824f83ef37f0658/src/app/lib/aiosql/service.py#L19)
- [Provides](https://github.com/cofin/litestar-fullstack/blob/81d33439a87d44ab1dcd93b8d824f83ef37f0658/src/app/domain/analytics/dependencies.py#LL24C24-L24C24)
- [Controller DI](https://github.com/cofin/litestar-fullstack/blob/81d33439a87d44ab1dcd93b8d824f83ef37f0658/src/app/domain/analytics/controllers.py#L33)

### MCVE

```python
async def provides_cool_thing(db_session: AsyncSession) -> AsyncGenerator[QueryManager, None]:
    async with QueryManager.from_session(cool_queries, db_session) as query_manager:
        logger.debug("providing cool queries %s", query_manager.queries)
        yield query_manager

class TestController(Controller):
    dependencies = {"cool_thing": Provide(provides_cool_thing)}
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

```bash
.../thing.py:51: LitestarWarning: Use of a synchronous callable 
<function provides_cool_thingat 0x7feedc123600> without setting sync_to_thread 
is discouraged since synchronous callables can block the main thread if they perform 
blocking operations. If the callable is guaranteed to be non-blocking, you can set 
sync_to_thread=False to skip this warning, or set the environment variable
 LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD=0 to disable warnings of this type entirely.

  dependencies = {"cool_thing": Provide(provides_cool_thing)}
```

### Litestar Version

2.0a7

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

peterschutt:
Closes #1711 

### Pull Request Checklist

### Description

### Close Issue(s)


peterschutt:
Initially went about this the wrong way - adding `isasyncgenfunction()` into the mix in `is_async_callable()` is wrong as async gens cannot be awaited like a coroutine:

```python-console
[GCC 9.4.0] on linux
Use "await" directly instead of "asyncio.run()".
>>> import asyncio
>>> async def f(): ...
... 
>>> async def g(): yield 1
... 
>>> f
<function f at 0x7f68d7e109d0>
>>> g
<function g at 0x7f68d7e10a60>
>>> f()
<coroutine object f at 0x7f68d7e072c0>
>>> g()
<console>:1: RuntimeWarning: coroutine 'f' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
<async_generator object g at 0x7f68d7e10af0>
>>> await f()
>>> await g()
  File "/home/peter/.pyenv/versions/3.8.15/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/home/peter/.pyenv/versions/3.8.15/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
TypeError: object async_generator can't be used in 'await' expression
>>> 
```
So everything else was behaving correctly, it really just is an issue of not warning for these cases. However, having an async generator dependency and `sync_to_thread` is `True` should maybe be a warning or an error, for as far as I can tell, they are never executed when called, so running them in a separate thread would be unnecessary, right?

```python-console
asyncio REPL 3.8.15 (default, Oct 17 2022, 12:54:29) 
[GCC 9.4.0] on linux
Use "await" directly instead of "asyncio.run()".
>>> import asyncio
>>> async def g():
...   print("in")
...   yield 1
... 
>>> g()
<async_generator object g at 0x7f09cc83da60>
>>> async for _ in g(): ...  # no execution until iterated, which would always be after injection
... 
in
Ellipsis
```

peterschutt:
Taking this even further, _any_ generator function doesn't execute until it is iterated, and so for our purposes of instantiating a dependency that is a generator function in order to inject it - irrespective of async or sync - we call it synchronously in either case. As nothing is ever executed at the point of calling these, it doesn't make sense to ever run this call in a thread, does it? 

So, should we just totally ignore `sync_to_thread` for generator dependencies? Or should we warn that it is not necessary to set `sync_to_thread=True` on a dependency that is a generator?

provinzkraut:
> Taking this even further, _any_ generator function doesn't execute until it is iterated, and so for our purposes of instantiating a dependency that is a generator function in order to inject it - irrespective of async or sync - we call it synchronously in either case. As nothing is ever executed at the point of calling these, it doesn't make sense to ever run this call in a thread, does it?

Interesting question. What should run in a thread would actually be the `next` call. I suppose we could add a utility that does this. While not ideal if the generator is advanced frequently (because a thread would have to be spawned for each invocation), in our case it would actually be fine since we only call `next` on it twice.

On the other hand, this is might over complicate things unnecessarily. 

> Or should we warn that it is not necessary to set sync_to_thread=True on a dependency that is a generator?

For now, I think we should warn about `sync_to_thread` with generators, just like we do for async functions, since it does not have any effect.

peterschutt:
> Do you want to add a warning for that in this PR?

Sure thing

litestar/di.py
```diff
@@ -50,11 +55,14 @@ def __init__(
...
+        if self.has_sync_callable and sync_to_thread is None and not is_sync_or_async_generator(dependency):
             warn_implicit_sync_to_thread(dependency, stacklevel=3)
 
-        if not self.has_sync_callable and sync_to_thread is not None:
-            warn_sync_to_thread_with_async_callable(dependency, stacklevel=3)
+        if sync_to_thread is not None:
+            if is_sync_or_async_generator(dependency):
+                warn_sync_to_thread_with_generator(dependency, stacklevel=3)
+            elif not self.has_sync_callable:
+                warn_sync_to_thread_with_async_callable(dependency, stacklevel=3)  # pyright: ignore
```
peterschutt, line 65: I couldn't work out how to make pyright realize that `dependency` could not be `AnyGenerator` here.. :cry:

provinzkraut: Isn't it right in that it can be? It's only narrowed to be a non-generator here:

https://github.com/litestar-org/litestar/blob/7a00aa71e08df7e5728e71aeff976a1ec1055714/litestar/di.py#L58

provinzkraut (approved):

litestar/di.py
peterschutt: yeah - but shouldn't this narrow the generators away for the `else` branch:
https://github.com/litestar-org/litestar/blob/7a00aa71e08df7e5728e71aeff976a1ec1055714/litestar/di.py#L62

provinzkraut: Uhm, yes. It should. I just missed that :grimacing:

PR #1729, opened 2023-05-25
Fix: `simple_asdict()` with attribute value having dataclass type.

peterschutt:
`simple_asdict()` was using `isinstance(value, DataclassProtocol)` to determine
if a value was a nested dataclass instance, however this would also return
`True` for dataclass types, which would cause an issue when the dataclass being
converted to a dict had an attribute value that was a dataclass type.

We add a utility function for specifically checking for dataclass instances,
`is_dataclass_instance()`. The logic of this function is pinched from the
private `_is_dataclass_instance()` function inside the stdlib `dataclasses`
module.

Additionally, this PR:

- reorganizes our utils so that all dataclass utils are grouped together in the `utils/dataclass.py` module.
- moves the tests for `ParsedType` into new module `tests/test_typing.py` to reflect the library structure.
- adds a regression test for `ParsedParameter.copy_with()` which is where the main issue addressed by this PR was exposed.

### Pull Request Checklist

### Description

### Close Issue(s)


litestar/utils/dataclass.py
```diff
@@ -1,21 +1,54 @@
...
 __all__ = (
     "extract_dataclass_fields",
     "extract_dataclass_items",
+    "is_dataclass_class",
+    "is_dataclass_instance",
     "simple_asdict",
 )
 
 
+def is_dataclass_instance(obj: Any) -> TypeGuard[DataclassProtocol]:
```
Goldziher, line 25: So, why not keep these inside the `predicates.py` namespace? It's where we have all type guards and similar predicates. I think consistency here is important.

Goldziher (changes requested):

litestar/utils/dataclass.py
peterschutt: We have `utils.dataclass` so that is the place I expect that dataclass utils exist.

The same argument about consistency can be made about having dataclass utilities across two modules.

peterschutt: There is precedent for predicates being grouped by functionality, e.g., `is_async_callable()` in the `utils.sync` module, `annotation_is_iterable_of_type()`  in `utils.typing`,

Goldziher: Well, i don't see the need for this change. I expect predicates to be in predicates since that's what we have there. 

The other utils should be moved there too if anything.

Goldziher (approved):

litestar/utils/dataclass.py
peterschutt: > The other utils should be moved there too if anything.

Created #1730

PR #1738, opened 2023-05-25
Feature: improved client error loc keys

peterschutt:
### Pull Request Checklist

[//]: # "Please review the [Litestar contribution guidelines](https://github.com/litestar-org/litestar/blob/main/CONTRIBUTING.rst) for this repository."

- [ ] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR

[//]: # "By submitting this issue, you agree to:"
[//]: # "- follow Litestar's [Code of Conduct](https://github.com/litestar-org/.github/blob/main/CODE_OF_CONDUCT.md)"
[//]: # "- follow Litestar's [Contribution Guidelines](https://litestar.dev/community/contribution-guide)"

### Description

[//]: # "Please describe your pull request for new release changelog purposes"
[//]: # "Example: 'This pr adds the capability to use server-sent events, and documents it in the usage docs'"

### Close Issue(s)

[//]: # "Please add in issue numbers this pull request will close, if applicable"
[//]: # "Examples: Fixes #4321 or Closes #1234"

PR #1774, opened 2023-06-02
Increase verbosity of validation error response keys

### LINKED ISSUE #1736: Enhancement: Make Pydantic exception keys more verbose
### Summary

To add clarity and avoid confusion when there are validation errors when using nested Pydantic models for input, it would be helpful to include the full `loc` path instead of just the child's key.

### Basic Example

```python
from litestar import Controller
from pydantic import BaseModel, ValidationError


class Child(BaseModel):
    my_value: str


class Parent(BaseModel):
    child: Child


class SomeController(Controller):
    @post('/some-route')
    def some_route(self, data: Parent) -> Parent:
        return data
```

The following request:

```bash
http POST localhost:8000/some-route child:='{}'
```
results in the following response:
```json
{
  "status_code": 400,
  "detail": "Validation failed for POST http://localhost:8000/some-route",
  "extra": [
    {
      "key": "my_value",
      "message": "field required"
    }
  ]
}
```

This is a bit confusing because it makes it seem like `my_value` is a `Parent`-level value, not a `Child`-level value. It may be helpful if the key were `child.my_value` (or some other similar solution), like:

```json
{
  "status_code": 400,
  "detail": "Validation failed for POST http://localhost:8000/some-route",
  "extra": [
    {
      "key": "child.my_value",
      "message": "field required"
    }
  ]
}
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

<a href="https://polar.sh/litestar-org/litestar/issues/1736">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1736/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

rseeley:
### Pull Request Checklist

### Description

This PR updates the keys in validation error responses to include the full path to the field with the issue. This gives the developer a better sense of where the failure happened.

It also adds an optional `source` key that designates whether the value is from the body, a cookie, a header, or a query param.

#### For example:

With the following controller

```python
from litestar import Controller
from pydantic import BaseModel, ValidationError


class Child(BaseModel):
    my_value: str


class Parent(BaseModel):
    child: Child


class SomeController(Controller):
    @post('/some-route')
    def some_route(
        self,
        data: Parent,
        int_param: int,
        int_header: int = Parameter(header="X-SOME-INT"),
        int_cookie: int = Parameter(cookie="int-cookie"), 
    ) -> Parent:
        return data
```

and the request

```bash
http POST localhost:8000/some-route child:='{}' int_param==asdf X-SOME-INT=asdf Cookie:int-cookie=asdf
```

the response would have previously been

```json
{
  "status_code": 400,
  "detail": "Validation failed for POST http://localhost:8000/some-route",
  "extra": [
    {"key": "int_param", "message": "value is not a valid integer"},
    {"key": "int_header", "message": "value is not a valid integer"},
    {"key": "int_cookie", "message": "value is not a valid integer"},
    {"key": "my_value", "message": "value is not a valid integer"}
  ]
}
```

This change will result in the following:

```json
{
  "status_code": 400,
  "detail": "Validation failed for POST http://localhost:8000/some-route",
  "extra": [
    {"key": "child.my_value", "message": "value is not a valid integer", "source": "body"},
    {"key": "int_param", "message": "value is not a valid integer", "source": "query"},
    {"key": "int_header", "message": "value is not a valid integer", "source": "header"},
    {"key": "int_cookie", "message": "value is not a valid integer", "source": "cookie"},
  ]
}
```

### Close Issue(s)

Fixes #1736

### DIFF
--- a/litestar/_signature/models/attrs_signature_model.py
+++ b/litestar/_signature/models/attrs_signature_model.py
@@ -10,6 +10,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    Union,
     cast,
 )
 from uuid import UUID
@@ -24,7 +25,6 @@ from litestar.datastructures import ImmutableState, MultiDict, State, UploadFile
 from litestar.exceptions import MissingDependencyException
 from litestar.params import BodyKwarg, DependencyKwarg, ParameterKwarg
 from litestar.types import Empty
-from litestar.utils import compact
 from litestar.utils.predicates import is_optional_union, is_union
 from litestar.utils.typing import get_origin_or_inner_type, make_non_optional_union, unwrap_union
 
@@ -49,7 +49,7 @@ if TYPE_CHECKING:
     from litestar.utils.signature import ParsedSignature
 
 __all__ = ("AttrsSignatureModel",)
-key_re = re.compile("@ attribute (.*)|'(.*)'")
+key_re = re.compile("@ (attribute|index) (.*)|'(.*)'")
 TRUE_SET = {"1", "true", "on", "t", "y", "yes"}
 FALSE_SET = {"0", "false", "off", "f", "n", "no"}
 
@@ -66,6 +66,11 @@ try:
     ]
 except ImportError:
     pydantic_hooks = []
+
+
+StructureException = Union[
+    cattrs.ClassValidationError, cattrs.IterableValidationError, ValueError, TypeError, AttributeError
+]
 
 
 def _pass_through_structure_hook(value: Any, _: type[Any]) -> Any:
@@ -238,33 +243,6 @@ class Converter(cattrs.Converter):
 _converter: Converter = Converter()
 
 
-def _extract_exceptions(e: Any) -> list[ErrorMessage]:
-    """Extracts and normalizes cattrs exceptions.
-
-    Args:
-        e: An ExceptionGroup - which is a py3.11 feature. We use hasattr instead of instance checks to avoid installing this.
-
-    Returns:
-        A list of normalized exception messages.
-    """
-    messages: list[ErrorMessage] = []
-    if hasattr(e, "exceptions"):
-        for exc in cast("list[Exception]", e.exceptions):
-            if hasattr(exc, "exceptions"):  # pragma: no cover
-                # cattrs raises an exception group, where each exception can potentially be an exception group.
-                # this case is not reproducible in any of our tests - and frankly I have no idea when it would occur,
-                # so this clause is defensive programming only.
-                messages.extend(_extract_exceptions(exc))
-                continue
-            if err_format := [
-                line
-                for line in traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__)
-                if key_re.search(line)
-            ]:
-                messages.append({"key": key_re.findall(compact(err_format)[-1])[0][0].strip(), "message": str(exc)})
-    return messages
-
-
 def _create_validators(
     annotation: Any, kwargs_model: BodyKwarg | ParameterKwarg
 ) -> list[Callable[[Any, attrs.Attribute[Any], Any], Any]] | Callable[[Any, attrs.Attribute[Any], Any], Any]:
@@ -304,12 +282,120 @@ class AttrsSignatureModel(SignatureModel):
         try:
             signature = _converter.structure(obj=kwargs, cl=cls)
         except (cattrs.ClassValidationError, ValueError, TypeError, AttributeError) as e:
-            raise cls._create_exception(messages=_extract_exceptions(e), connection=connection) from e
+            raise cls._create_exception(messages=cls._extract_exceptions(e, connection), connection=connection) from e
 
         return cast("dict[str, Any]", _converter.unstructure(obj=signature))
 
     def to_dict(self) -> dict[str, Any]:
         return attrs.asdict(self)
+
+    @classmethod
+    def _extract_exceptions(cls, e: StructureException, connection: ASGIConnection) -> list[ErrorMessage]:
+        """Extracts and normalizes cattrs exceptions.
+
+        Args:
+            e: An ExceptionGroup - which is a py3.11 feature. We use hasattr instead of instance checks to avoid installing this.
+            connection: The connection instance.
+
+        Returns:
+            A list of normalized exception messages.
+        """
+
+        error_messages: list[ErrorMessage] = []
+
+        if isinstance(e, cattrs.ClassValidationError):
+            for exc in cast("list[StructureException]", e.exceptions):
+                if messages := cls._get_messages_from_traceback(exc, connection):
+                    error_messages.extend(messages)
+
+        return error_messages
+
+    @classmethod
+    def _get_messages_from_traceback(cls, exc: StructureException, connection: ASGIConnection) -> list[ErrorMessage]:
+        """Gets a message from an attrs validation error.
+
+        The key will be a dot-separated string of the attribute path that failed
+        validation. The message will be the error message from the exception (or the
+        last exception in the exception group, when applicable).
+
+        Args:
+            exc: The exception to get the message from
+            connection: The connection instance
+
+        Returns:
+            An error message
+        """
+
+        error_data = cls._get_data_from_exception(exc=exc)
+        return cls._get_error_messages(error_data=error_data, connection=connection)
+
+    @classmethod
+    def _get_data_from_exception(
+        cls, exc: StructureException, prefix: str = "", error_data: dict | None = None
+    ) -> dict[str, str]:
+        """Gets the keys from an attrs validation error.
+
+        Handles nested structures (e.g. a model attribute references another
+        model) by going through all exceptions in the exception group
+        """
+
+        error_data = error_data or {}
+
+        if isinstance(exc, (cattrs.ClassValidationError, cattrs.IterableValidationError)):
+            formatted_exception = traceback.format_exception_only(type(exc), value=exc)
+            key = cls._get_key_from_formatted_exception(formatted_exception)
+
+            new_prefix = f"{prefix}.{key}" if prefix else key
+
+            for sub_exc in cast("list[StructureException]", exc.exceptions):
+                error_data = cls._get_data_from_exception(sub_exc, new_prefix, error_data)
+        # when using attrs as the preferred backend validation but
+        # pydantic as the model, you can still get pydantic
+        # validation errors.
+        elif isinstance(exc, pydantic.ValidationError):
+            formatted_exception = traceback.format_exception_only(type(exc), value=exc)
+            key = cls._get_key_from_formatted_exception(formatted_exception)
+
+            for error in exc.errors():
+                error_key = ".".join([key, *[str(loc) for loc in error["loc"]]])
+                error_data[error_key] = error["msg"]
+        else:
+            formatted_exception = traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__)
+            key = cls._get_key_from_formatted_exception(formatted_exception)
+            key = f"{prefix}.{key}" if prefix else key
+
+            error_data[key] = str(exc)
+
+        return error_data
+
+    @classmethod
+    def _get_key_from_formatted_exception(cls, formatted_exception: list[str]) -> str:
+        """Gets the key from a formatted exception."""
+        return next(
+            (key for line in formatted_exception if (match := key_re.findall(line)) and (key := match[0][1].strip())),
+            "",
+        )
+
+    @classmethod
+    def _get_error_messages(cls, error_data: dict[str, str], connection: ASGIConnection) -> list[ErrorMessage]:
+        """Build an error message.
+
+        Args:
+            error_data: A mapping of error location (dot-notated) to their error.
+            connection: An ASGI connection instance.
+
+        Returns:
+            An ErrorMessage
+        """
+
+        messages: list[ErrorMessage] = []
+
+        for key, error in error_data.items():
+            keys = key.split(".")
+            message = super()._build_error_message(keys=keys, exc_msg=error, connection=connection)
+            messages.append(message)
+
+        return messages
 
     @classmethod
     def populate_signature_fields(cls) -> None:
--- a/litestar/_signature/models/base.py
+++ b/litestar/_signature/models/base.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, ClassVar, TypedDict
+from typing import TYPE_CHECKING, Any, ClassVar, Literal, Sequence, TypedDict, cast
 
 from litestar.enums import ScopeType
 from litestar.exceptions import InternalServerException, ValidationException
+from litestar.params import ParameterKwarg
 
 if TYPE_CHECKING:
+    from typing_extensions import NotRequired
+
     from litestar._signature.field import SignatureField
     from litestar.connection import ASGIConnection
     from litestar.utils.signature import ParsedSignature
@@ -15,8 +18,12 @@ __all__ = ("SignatureModel",)
 
 
 class ErrorMessage(TypedDict):
-    key: str
+    # key may not be set in some cases, like when a query param is set but
+    # doesn't match the required length during `attrs` validation
+    # in this case, we don't show a key at all as it will be empty
+    key: NotRequired[str]
     message: str
+    source: NotRequired[Literal["cookie", "body", "header", "query"]]
 
 
 class SignatureModel(ABC):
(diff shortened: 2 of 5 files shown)


litestar/_signature/models/attrs_signature_model.py
```diff
@@ -238,30 +242,52 @@ def __init__(self) -> None:
...
+    if isinstance(exception, cattrs.ClassValidationError):
+        # ensure we get the actual cause of the exception by getting the last
+        # exception in the group
+        while hasattr(exception, "exceptions"):
+            exception = exception.exceptions[-1]
+
+    message = {"key": ".".join(key_matches), "message": str(exception)}
+
+    if key_matches and key_matches[0] != "data" and key_matches[0] in connection.query_params:
+        message["key"] = f"query_params.{message['key']}"
```
rseeley, line 268: Not sure how the team feels about this, but I prepended `query_params` (where applicable) in the spirit of adding clarity around the value that's got a problem.

litestar/_signature/models/attrs_signature_model.py
```diff
@@ -304,7 +344,7 @@ def parse_values_from_connection_kwargs(cls, connection: ASGIConnection, **kwarg
         try:
             signature = _converter.structure(obj=kwargs, cl=cls)
         except (cattrs.ClassValidationError, ValueError, TypeError, AttributeError) as e:
```
rseeley, line 346: This may not be the correct spot to ask this question, but I tested throwing a `TypeError` by not passing a required arg and it was still encapsulated in a `ClassValidationError`. Is this line catching the others as a defensive measure or did I potentially miss some scenarios?

litestar/_signature/models/attrs_signature_model.py
Goldziher: Hi @rseeley,

First off - thanks for contributing 😁

Secondly, I don't like this change. I think the parameter name is sufficient. 

Furthermore, you do not handle header and cookie parameters, no? 

Finally - whatever changes are done on the attrs side will have to also be done on the pydantic side. Our intention is to wrap error handling so it does not leak implementation details.

litestar/_signature/models/attrs_signature_model.py
Goldziher: I'm not sure tbh, it's been a while. You can try removing the other types and checking.

Goldziher: Also, can you make sure to update dependencies `poetry update`

litestar/_signature/models/attrs_signature_model.py
peterschutt: I suggest we add another field to `ErrorMessage` for reporting this:

```py
class ErrorMessage(TypedDict):
    key: str
    message: str
    source: Literal["body", "query", "cookie", "header"]
```
Something like that.

peterschutt:
I believe the issue demonstrated in #1738 - where the tests in `tests/signature/test_signature_validation.py` is still present here, where the integration tests are not selecting the attrs model.

E.g., I've raised a runtime error in the `_get_message_from_traceback()` function in the attrs module, yet the test still passes:

```py
def _get_message_from_traceback(exc: StructureException, connection: ASGIConnection) -> ErrorMessage:
    """Gets a message from an attrs validation error.

    The key will be a dot-separated string of the attribute path that failed
    validation. The message will be the error message from the exception (or the
    last exception in the exception group, when applicable).

    Args:
        exc: The exception to get the message from
        connection: The connection instance

    Returns:
        An error message
    """
    raise RuntimeError()
    err_format = [
        line for line in traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__) if key_re.search(line)
    ]
    ...
```
and
```console
(litestar-py3.8) peter@peter-Inspiron-15-5510:~/PycharmProjects/litestar$ pytest -s tests/signature/test_parsing.py -k test_validation_error_exception_key
============================================================================================ test session starts =============================================================================================
platform linux -- Python 3.8.15, pytest-7.3.1, pluggy-1.0.0
rootdir: /home/peter/PycharmProjects/litestar
configfile: pyproject.toml
plugins: asyncio-0.21.0, Faker-18.10.1, rerunfailures-11.1.2, anyio-3.7.0, mock-3.10.0, docker-1.0.1, cov-4.1.0, lazy-fixture-0.6.3, hypothesis-6.75.9
asyncio: mode=auto
collected 35 items / 33 deselected / 2 selected                                                                                                                                                              

tests/signature/test_parsing.py ..
```

peterschutt:
> I believe the issue demonstrated in #1738 - where the tests in `tests/signature/test_signature_validation.py` do not select the attrs model is still present here.
> 
> E.g., I've raised a runtime error in the `_get_message_from_traceback()` function in the attrs module, yet the test still passes:
> 
> ```python
> def _get_message_from_traceback(exc: StructureException, connection: ASGIConnection) -> ErrorMessage:
>     """Gets a message from an attrs validation error.
> 
>     The key will be a dot-separated string of the attribute path that failed
>     validation. The message will be the error message from the exception (or the
>     last exception in the exception group, when applicable).
> 
>     Args:
>         exc: The exception to get the message from
>         connection: The connection instance
> 
>     Returns:
>         An error message
>     """
>     raise RuntimeError()
>     err_format = [
>         line for line in traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__) if key_re.search(line)
>     ]
>     ...
> ```
> 
> and
> 
> ```
> (litestar-py3.8) peter@peter-Inspiron-15-5510:~/PycharmProjects/litestar$ pytest -s tests/signature/test_parsing.py -k test_validation_error_exception_key
> ============================================================================================ test session starts =============================================================================================
> platform linux -- Python 3.8.15, pytest-7.3.1, pluggy-1.0.0
> rootdir: /home/peter/PycharmProjects/litestar
> configfile: pyproject.toml
> plugins: asyncio-0.21.0, Faker-18.10.1, rerunfailures-11.1.2, anyio-3.7.0, mock-3.10.0, docker-1.0.1, cov-4.1.0, lazy-fixture-0.6.3, hypothesis-6.75.9
> asyncio: mode=auto
> collected 35 items / 33 deselected / 2 selected                                                                                                                                                              
> 
> tests/signature/test_parsing.py ..
> ```

Ah, ok, if pydantic is installed, the attrs backend is only selected if there is an attrs model in the signature.

https://github.com/litestar-org/litestar/blob/0af0038582e7e38ae85366166d725bee0fb989bf/litestar/_signature/utils.py#L124-L136

litestar/_signature/models/attrs_signature_model.py
```diff
@@ -238,30 +242,66 @@ def __init__(self) -> None:
...
+    err_format = [
+        line for line in traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__) if key_re.search(line)
+    ]
+
+    # handles nested structures (e.g. a model attribute references another
+    # model) by going through all exceptions in the exception group
+    key_matches = []
+    for line in compact(err_format):
+        if k := key_re.findall(line)[0][0].strip():
+            key_matches.append(k)
```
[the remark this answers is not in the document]
peterschutt: Also, running the regex over the lines twice, once with `re.search()` and again with `re.findall()`, stands out. Can this be avoided at all?

peterschutt (changes requested):

provinzkraut:
@peterschutt 

> Ah, ok, if pydantic is installed, the attrs backend is only selected if there is an attrs model in the signature.

Is that the correct behaviour though? I thought the purpose of being able to explicitly set a backend was to *not* have this behaviour? E.g. if I set the preferred validation backend to attrs, then it should use attrs, even if there's no attrs model being used.

peterschutt:
> @peterschutt
> 
> > Ah, ok, if pydantic is installed, the attrs backend is only selected if there is an attrs model in the signature.
> 
> Is that the correct behaviour though? I thought the purpose of being able to explicitly set a backend was to _not_ have this behaviour? E.g. if I set the preferred validation backend to attrs, then it should use attrs, even if there's no attrs model being used.

I don't know if it is the intended behavior - but it is the current behavior. It is surprising to me, as evidenced by these tests not working the way that I thought they would.

provinzkraut:
> I don't know if it is the intended behavior - but it is the current behavior. It is surprising to me, as evidenced by these tests not working the way that I thought they would.

Intention aside, I think it *should* be the behaviour. As you pointed out, the current behaviour is not inuitive.

litestar/_signature/models/attrs_signature_model.py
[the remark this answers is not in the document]
rseeley: Great feedback. I was able to do this without looping over it twice and only doing one regex search.

litestar/_signature/models/attrs_signature_model.py
rseeley: I like that idea. How do you feel on that, @Goldziher ?

Nice catch on the other data types (header, cookie). I'll update to include those.

Goldziher: Sure

litestar/_signature/models/attrs_signature_model.py
rseeley: I removed all but the `ClassValidationError` and all the tests passed, but it felt unnecessary as part of this PR to keep them removed. It'll make the code slightly cleaner in the message building, but it should probably be tested more thoroughly before doing so.

rseeley:
@all-contributors please add @rseeley for code

litestar/_signature/models/attrs_signature_model.py
```diff
@@ -238,30 +241,59 @@ def __init__(self) -> None:
...
+    Returns:
+        An error message
+    """
+
+    formatted_exception = traceback.format_exception(type(exc), value=exc, tb=exc.__traceback__)
+
+    # handles nested structures (e.g. a model attribute references another
+    # model) by going through all exceptions in the exception group
+    keys: list[str] = []
+    for line in formatted_exception:
```
Goldziher, line 264: This can be a list comprehension

litestar/_signature/models/attrs_signature_model.py
```diff
@@ -238,30 +241,59 @@ def __init__(self) -> None:
...
+        if (match := key_re.findall(line)) and (key := match[0][0].strip()):
+            keys.append(key)
+
+    exception = exc
```
Goldziher, line 268: Why?

litestar/_signature/models/base.py
```diff
@@ -17,6 +20,7 @@
 class ErrorMessage(TypedDict):
     key: str
     message: str
+    source: NotRequired[Literal["cookie", "body", "header", "query"]]
```
Goldziher, line 23: Why not required?

litestar/_signature/models/utils.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import TYPE_CHECKING, Literal, cast
+
+from litestar.params import ParameterKwarg
+
+if TYPE_CHECKING:
+    from litestar.connection.base import ASGIConnection
+    from litestar.utils.signature import ParsedParameter
+
+
+def get_param_source(key: str | int, connection: ASGIConnection) -> Literal["cookie", "header", "query"] | None:
```
Goldziher, line 12: So, the contents of this file are not really required as they are now. 

The signature model is responsible for modelling the required kwargs. We than have a pretty similar logic under the `_kwargs` namespace for modelling what you're doing here, basically. Take a look there.

Goldziher (commented):
Very nice progress

litestar/_signature/models/attrs_signature_model.py
rseeley: Will do.

litestar/_signature/models/attrs_signature_model.py
rseeley: Oh that's a bit silly looking at it now. I can just set this to `None` and maybe rename it to be more clear. Unless you were asking about the block as a whole?

litestar/_signature/models/base.py
rseeley: I opted to not include dependencies in the source handling since it's not part of the user request, but it probably makes sense to include that, too. Do you agree? If so, are there perhaps other scenarios I'm not considering that we'd want to include? It's probably better to be comprehensive.

litestar/_signature/models/utils.py
rseeley: Awesome, I'll take a look at that, thanks.

litestar/_signature/models/attrs_signature_model.py
rseeley: Do you mean why not use `exc` directly and overwrite that instead of creating a new variable?

litestar/_signature/models/base.py
peterschutt: I think you were correct to not include dependencies. Clients shouldn't know, or care, where a 500 originates from within the application as they cannot do anything about it, and so it would just unnecessarily leak implementation detail.

rseeley: Fair point.

litestar/_signature/models/attrs_signature_model.py
Goldziher: Indeed

litestar/_signature/models/base.py
```diff
@@ -47,6 +51,46 @@ def _create_exception(cls, connection: ASGIConnection, messages: list[ErrorMessa
...
+                key_start = 1
+                message["source"] = "body"
+
+            message["key"] = ".".join(keys[key_start:])
+        elif keys:
+            key = keys[0]
+            message["key"] = key
+
+            if key in connection.query_params:
+                message["source"] = "query"
```
rseeley, line 82: @Goldziher , made some changes here to use `kwarg_model`. I believe this is in line with the direction you were hoping to go. However, I noticed that when query params or dependencies are defined in the route handler using only a type annotation (i.e. `dep: int` or `param: int`), they don't get a `DependencyKwarg` or `ParameterKwarg` in the fields (the `kwarg_model` is `None`). As a result, I had to keep this `connection.query_params` check. Super possible that there's an easier way than what I've done here, but I was spinning my wheels and am running out of bandwidth. How do you feel about the current solution here?

Goldziher: this looks good. but why are you ignoring data? you can also validate the data kwarg, no?

Goldziher (approved):

Goldziher:
@all-contributors add @rseeley code

litestar/_signature/models/base.py
rseeley: I handled `data` in its own block as the keys will always be nested, and given its naming pattern, I figured we'll always know that `"body"` should be the source in that scenario. 

However, when looking at `cls.fields['data']`, `kwarg_model` for `data` is `None`. It doesn't seem to have anything indicating it's from the body at the time of building the message.

Goldziher: The field name is always data

rseeley: Yup, totally. Sorry, I meant there's nothing aside from the naming, from what I could tell. But because of that (it always being named `data`), I opted for just looking at that directly in the keys list.

Unrelated, but I saw that you approved (thanks!). I pulled in the latest upstream changes and something seems to have broken along the way with respect to my changes. There's a test that uses the preferred_validation_backend and I'm seeing some weird behavior with how the validation is being handled (it's going through the attrs signature model but the error is a pydantic error). I wasn't able to dig into it due to time constraints and will continue tomorrow. But j wanted to call that out since you approved when it was working.

Goldziher (approved):

tests/signature/test_parsing.py
```diff
@@ -284,3 +289,200 @@ def handler(param: bool) -> None:
...
+
+        assert response.status_code == HTTP_400_BAD_REQUEST
+
+        data = response.json()
+
+        assert data
+        assert data["extra"] == [
+            {"key": "child.val", "message": "value is not a valid integer", "source": "body"},
+            {"key": "child.other_val", "message": "value is not a valid integer", "source": "body"},
+            {"key": "other_child.val.1", "message": "value is not a valid integer", "source": "body"},
```
peterschutt, line 366: This isn't tested against attrs, does it behave the same?

peterschutt (approved):
This is good! I've one question there about the tests.

Have you given any thought to trying to standardize the messages as well? One approach could be to use a regex for each backend type to parse out pertinent info from the error message so that we could then reformat it.

tests/signature/test_parsing.py
rseeley: Good catch. Attrs actually fails this check when there are multiple nested models. Working on a fix now.

rseeley:
> This is good! I've one question there about the tests.
> 
> Have you given any thought to trying to standardize the messages as well? One approach could be to use a regex for each backend type to parse out pertinent info from the error message so that we could then reformat it.

It'd be awesome if that could be done! I think it'd fit better as part of a separate chunk of work, though, if that's alright. I don't know if I'll have the bandwidth soon to tackle that.

tests/unit/test_signature/test_parsing.py
```diff
@@ -284,3 +289,229 @@ def handler(param: bool) -> None:
...
+
+        data = response.json()
+
+        assert data
+        assert data["extra"] == [
+            {"key": "child.val", "message": "invalid literal for int() with base 10: 'a'", "source": "body"},
+            {"key": "child.other_val", "message": "invalid literal for int() with base 10: 'b'", "source": "body"},
+            {"key": "other_child.val.1", "message": "invalid literal for int() with base 10: 'c'", "source": "body"},
+            {"key": "int_param", "message": "invalid literal for int() with base 10: 'param'", "source": "query"},
+            {"key": "int_header", "message": "invalid literal for int() with base 10: 'header'", "source": "header"},
```
rseeley, line 418: Note that there's no `length_param` `extra` value here like there are for the other validation types. I'm noticing that when you have a param with no default followed by a param with a default and the no-default one fails validation, the one with a default won't validate fully (e.g., the min_length won't be validated against, but the type will).

This can be validated by changing `tests/unit/test_params.py::test_parsing_of_parameter_as_default_value` to the following and viewing the output of `response.json()`:

```python
@pytest.mark.parametrize("backend", ("pydantic", "attrs"))
def test_parsing_of_parameter_as_default_value(backend: Any) -> None:
    @get(path="/")
    def handler(int_param: int, other_int_param: int, param: int = Parameter(ge=1)) -> str:
        return param

    with create_test_client(handler, _preferred_validation_backend=backend) as client:
        response = client.get("/?int_param=a&other_int_param=a&param=a")
        print(f'{response.json()=}')

        response = client.get("/?int_param=4&other_int_param=4&param=a")
```

This fails the `int` check. But if you change it to `param: str = Parameter(min_length=1)`, `param` will not fail validation.

litestar/_signature/models/attrs_signature_model.py
rseeley: Good callout, this wasn't needed and I've removed it.

tests/signature/test_parsing.py
rseeley: Fixed, @peterschutt !

tests/unit/test_signature/test_parsing.py
Goldziher: Sure, make a ticket?

rseeley: Will do!

PR #1777, opened 2023-06-02
Feature: DTOConfig.underscore_fields_private.

### LINKED ISSUE #1768: Enhancement : DTO - private properties and methods  should be excluded by default.
### Summary

Identifiers which are considered private - starts with `_` should be mark as private by default , unless required.

### Basic Example

```py

class User(UUIDAuditBase):
    """User Model."""
    email: Mapped[str] = mapped_column(unique=True)
    name: Mapped[str | None]
    _private_name: Mapped[str | None]``` # <--- should be marked as `mark.private`

```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

peterschutt:
Closes #1768

### Pull Request Checklist

### Description

### Close Issue(s)


litestar/_signature/utils.py
```diff
@@ -56,6 +57,7 @@ def create_signature_model(
         fn: A callable.
         preferred_validation_backend: Validation/Parsing backend to prefer, if installed
         parsed_signature: A parsed signature for the handler/dependency function.
+        has_data_dto: Is a data DTO defined for the handler?
```
peterschutt, line 60: Added this param so that when a dto is available for the handler the "data" field on the signature model will be typed `Any`. I think this is correct behavior as the DTO is doing the raw data validation, so we just want whatever it returns to be passed through to the handler.

I noticed this when testing the behavior of the DTOs with `underscore_fields_private=False`, the signature model would emit

>   /home/peter/PycharmProjects/litestar/litestar/_signature/models/pydantic_signature_model.py:165: RuntimeWarning: fields may not start with an underscore, ignoring "_baz"
    model: type[PydanticSignatureModel] = create_model(  # type: ignore

Goldziher: Have you tested the OpenAPI schema?

litestar/contrib/sqlalchemy/dto.py
```diff
@@ -94,7 +94,7 @@ def _(
                 default=default,
                 parsed_type=parsed_type,
                 default_factory=default_factory,
-                dto_field=elem.info.get(DTO_FIELD_META_KEY),
+                dto_field=elem.info.get(DTO_FIELD_META_KEY, DTOField()),
```
Goldziher, line 97: Why not use Empty to signify this?

litestar/dto/factory/_backends/abc.py
```diff
@@ -162,6 +154,9 @@ def parse_model(
             if should_ignore_field(field_definition, self.context.dto_for):
                 continue
 
+            if should_mark_private(field_definition, self.context.config.underscore_fields_private):
+                field_definition.dto_field.mark = Mark.PRIVATE
```
Goldziher, line 158: I think `Mark` is not a good name. 

Additionally, instead of an enum - I'd use a sentinel class for this.

litestar/dto/factory/_backends/utils/renaming.py
```diff
@@ -0,0 +1,68 @@
...
+    def upper(field_name: str) -> str:
+        return field_name.upper()
+
+    @staticmethod
+    def lower(field_name: str) -> str:
+        return field_name.lower()
+
+    @staticmethod
+    def camel(field_name: str) -> str:
+        return RenameStrategies._camelize(field_name)
```
Goldziher, line 36: Why is this a private field?

litestar/dto/factory/_backends/utils/renaming.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def pascal(field_name: str) -> str:
+        return RenameStrategies._camelize(field_name, capitalize_first_letter=True)
+
+    @staticmethod
+    def _camelize(string: str, capitalize_first_letter: bool = False) -> str:
```
Goldziher, line 43: Why is this bound to the class? This is a classic utility function that should live in our utils folder. Same goes for the others.

litestar/dto/factory/data_structures.py
```diff
@@ -84,7 +84,7 @@ class FieldDefinition(ParsedParameter):
     """Unique identifier of model that owns the field."""
     default_factory: Callable[[], Any] | None
     """Default factory of the field."""
-    dto_field: DTOField | None
+    dto_field: DTOField
```
Goldziher, line 87: I'd type it as union `DTOField | Empty`

Goldziher (changes requested):

litestar/_signature/utils.py
peterschutt: Yes, schema for the request body is requested from the DTO if one has been declared:

https://github.com/litestar-org/litestar/blob/5d6cc88a06ff0e495467bde7468a5a60fd551003/litestar/_openapi/request_body.py#L33-L34

From this dataclass definition and the two handlers, one with and one without a DTO applied:

```py

@dataclass
class Thing:
    a: int
    _b: int


@post("/without")
async def without_dto(data: Thing) -> Thing:
    """Route Handler that outputs hello world."""
    return data


@post("/with", dto=DataclassDTO[Thing])
async def with_dto(data: Thing) -> Thing:
    """Route Handler that outputs hello world."""
    return data

```

Without docs:

image

With docs:

image

litestar/contrib/sqlalchemy/dto.py
peterschutt: Because `FieldDefinition.dto_field` is typed `DTOField`, not `DTOField | Empty` or `DTOField | None`. That is intentional to make it easier to use downstream. I would have given the field a default value on the dataclass but without `kw_only` for dataclasses in 3.8, that makes the type impossible to subclass.

litestar/dto/factory/_backends/abc.py
peterschutt: Well that's a little beyond the scope of this PR. It has been called `Mark` since the very first DTO PR.

litestar/dto/factory/_backends/utils/renaming.py
peterschutt: Not sure - I agree it is probably unnecessary.

litestar/dto/factory/data_structures.py
peterschutt: As I mentioned before, this is intentionally not a union with anything else.

litestar/dto/factory/_backends/utils/renaming.py
peterschutt: This stuff was only showed as changes in this PR because I refactored the utils module. I've reverted those refactors from this PR as I already was uneasy that I was lumping those changes in with the changes for this feature.

litestar/dto/factory/_backends/abc.py
v3ss0n: i like the `Mark` name tho.

PR #1782, opened 2023-06-04
DTO handling for handlers that return Response.

### LINKED ISSUE #1631: DTO: Implement DTO factory for Response[type]
### Summary

Sometimes, a concrete `Response` needs to be returned by a handler. For example, to add dynamic headers. The body of the `Response` should be something that can potentially be handled by `return_dto` of the handler.

### Basic Example

```python
from litestar import post

from .models import User, UserDTO

@get(user_id: str, return_dto=UserDTO)
def fetch_user() -> Response[User]:
    return  Response(
        User.get(user_id),
        headers={"Random-Header": str(randint(1, 100))},
    )
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

peterschutt:
- Delays rendering response content to bytes until the "response.start" message is sent. Previously, this occurred on instantiation of the response object, however, that did not give the DTO opportunity to overwrite the content.
- Moves call of the return dto into the `HTTPRouteHandler.to_response()` method. This removes the need for duplicated code in the response handlers.
- Modifies `DTOInterface.to_encodable_type()` return type to include `Response`.

Closes #1631

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/litestar/dto/factory/_backends/abc.py
+++ b/litestar/dto/factory/_backends/abc.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Final, Generic, TypeVar
 from litestar._openapi.schema_generation import create_schema
 from litestar._signature.field import SignatureField
 from litestar.dto.factory import DTOData, Mark
+from litestar.response import Response
 from litestar.utils.helpers import get_fully_qualified_class_name
 
 from .types import (
@@ -119,7 +120,7 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
         self.dto_data_type: type[DTOData] | None = None
         if context.parsed_type.is_subclass_of(DTOData):
             self.dto_data_type = context.parsed_type.annotation
-            annotation = self.dto_data_type.parsed_type.annotation
+            annotation = self.context.parsed_type.inner_types[0].annotation
         else:
             annotation = context.parsed_type.annotation
         self.annotation = build_annotation_for_backend(annotation, self.transfer_model_type)
@@ -279,7 +280,7 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
             self.context.parsed_type,
         )
 
-    def encode_data(self, data: Any, connection_context: ConnectionContext) -> LitestarEncodableType:
+    def encode_data(self, data: Any, connection_context: ConnectionContext) -> LitestarEncodableType | Response:
         """Encode data into a ``LitestarEncodableType``.
 
         Args:
@@ -289,6 +290,15 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
         Returns:
             Encoded data.
         """
+        if isinstance(data, Response):
+            data.content = transfer_data(
+                self.transfer_model_type,
+                data.content,
+                self.parsed_field_definitions,
+                "return",
+                self.context.parsed_type,
+            )
+            return data
         return transfer_data(
             self.transfer_model_type, data, self.parsed_field_definitions, "return", self.context.parsed_type  # type: ignore[arg-type]
         )
--- a/litestar/dto/factory/_backends/abc.py
+++ b/litestar/dto/factory/_backends/abc.py
@@ -22,13 +22,10 @@ from .types import (
     TupleType,
     UnionType,
 )
-from .utils import (
-    RenameStrategies,
-    build_annotation_for_backend,
-    should_exclude_field,
-    should_ignore_field,
-    transfer_data,
-)
+from .utils.predicates import should_exclude_field, should_ignore_field, should_mark_private
+from .utils.renaming import determine_serialization_name
+from .utils.transfer import transfer_data
+from .utils.typing import build_annotation_for_backend
 
 if TYPE_CHECKING:
     from typing import AbstractSet, Any, Callable, Generator
@@ -127,12 +124,7 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
             annotation = context.parsed_type.annotation
         self.annotation = build_annotation_for_backend(annotation, self.transfer_model_type)
 
-    def parse_model(
-        self,
-        model_type: Any,
-        exclude: AbstractSet[str],
-        nested_depth: int = 0,
-    ) -> FieldDefinitionsType:
+    def parse_model(self, model_type: Any, exclude: AbstractSet[str], nested_depth: int = 0) -> FieldDefinitionsType:
         """Reduce :attr:`model_type` to :class:`FieldDefinitionsType`.
 
         .. important::
@@ -162,11 +154,7 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
             if should_ignore_field(field_definition, self.context.dto_for):
                 continue
 
-            if (
-                self.context.config.underscore_fields_private
-                and field_definition.name.startswith("_")
-                and field_definition.dto_field.mark is None
-            ):
+            if should_mark_private(field_definition, self.context.config.underscore_fields_private):
                 field_definition.dto_field.mark = Mark.PRIVATE
 
             try:
@@ -180,16 +168,9 @@ class AbstractDTOBackend(ABC, Generic[BackendT]):
             except NestedDepthExceededError:
                 continue
 
-            if rename := self.context.config.rename_fields.get(field_definition.name):
-                serialization_name = rename
-            elif self.context.config.rename_strategy:
-                serialization_name = RenameStrategies(self.context.config.rename_strategy)(field_definition.name)
-            else:
-                serialization_name = field_definition.name
-
             transfer_field_definition = TransferFieldDefinition.from_field_definition(
                 field_definition=field_definition,
-                serialization_name=serialization_name,
+                serialization_name=determine_serialization_name(self.context.config, field_definition),
                 transfer_type=transfer_type,
                 is_partial=self.context.config.partial,
                 is_excluded=should_exclude_field(field_definition, exclude, self.context.dto_for),
--- a/litestar/dto/factory/_backends/abc.py
+++ b/litestar/dto/factory/_backends/abc.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, Final, Generic, TypeVar
 
 from litestar._openapi.schema_generation import create_schema
 from litestar._signature.field import SignatureField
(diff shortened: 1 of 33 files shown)


litestar/response/base.py
```diff
@@ -256,6 +259,31 @@ def render(self, content: Any) -> bytes:
...
+
+        Returns:
+            The response body as bytes.
+        """
+        if self._body is None:
+            self._body = self.content if isinstance(self.content, bytes) else self.render(self.content)
+        return self._body
+
+    @body.setter
+    def body(self, value: bytes) -> None:
```
peterschutt, line 276: Added a setter b/c the body can be modified by hooks on the way out, which wouldn't otherwise be possible.

Now that there is both `Response.content` which is the response content before rendering, and `Response.body` which is the content after rendering, I can see how it might be a little confusing. Any thoughts? One thought I had was to also use a `@property` and `@content.setter` for `content` and emit a warning if `content` is set after `body` has been accessed.

provinzkraut: How about using a property for `content` but not having a setter at all? Then it's not a warning but an error and type checkers and IDEs will warn about assigning to that directly.

peterschutt: Well the point of adding the `content` attribute is so that I can set it on the response instance from within the DTO.

provinzkraut: Okay so how about making `content` a property that also unsets `_body`?
That way `.body` is guaranteed to always be up to date.

Truth be told, I'm not a huge fan of this type of mutable responses like this in general, since it can lead to a lot of trouble. Maybe we should have a "frozen response" instead, which is a subclass of the `Response` and is what actually does the sending and stuff, and `Response` is simply a data container?

peterschutt: > Truth be told, I'm not a huge fan of this type of mutable responses like this in general

I can understand that.

We have tests that do explicitly modify the outgoing request in-place, e.g,:

```py
def sync_after_request_handler(response: Response) -> Response:
    assert isinstance(response, Response)
    response.body = response.render({"hello": "moon"})
    return response
```
At this point, would the response be immutable, or still the container?

Not totally related, but something else I've thought about recently is having a response context object available via injection, which can be modified in-place, and is merged with the response behind the scenes, after the handler has returned.

```py
@get()
def handler(response_context: ResponseContext) -> Thing:
    reponse_context.headers.update({"x-custom-header": "abc123"})
    return Thing()
```
I was thinking about this in relation to the `Response[Paginated[Nested[Hell]]]]` issue we were discussing, as a means to be able to modify the response without having to actually return one from the handler.. that thing would pretty much be a response container.

provinzkraut: > At this point, would the response be immutable, or still the container?

It needs to be a container since the point of these hooks is to be able to modify or even replace the response. But they are a whole other mess in general :grimacing: 

I like your suggestion with the `ResponseContext`. You would still need to be able to return a concrete Response in the hooks, since one of their uses is to outright replace a response.

It *could* solve our problem with immutability though. If we simply make `Response` immutable and offer a `ResponseContext` you can mutate, we could easily just make the return value of those hooks optional, and then you can either change something in a `ResponseContext`, or return a completely new `Response` instance, but not mutate the `Response` instance directly within the hook.

abdulhaq-e:
I've seen the `ContextBag` pattern before, it's a really good idea.

litestar/response/base.py
peterschutt: Some observations
---

- For all of the `Response` types, `Response` is the only one we advocate returning directly from the handlers - for the others, we have `ResponseContainer`s, which should instead be returned.
- In the docs we demonstrate using `after_request` to modify the response object in order to dynamically set both headers and cookies at any layer of the application.
- We also support returning a type-specialized `Response` to dynamically set headers and cookies at the handler layer.
- Wherever a dynamic header or cookie is set, we require users to also set a `ResponseHeader` or `Cookie` in the `response_headers` or `response_cookies` collection at the same layer as the dynamic header/cookie is set.

ideas for discussion
---

1. We end support for returning `Response` from handlers and replace with the response context idea from above. For the case of dynamically generating header/cookie values inside the handler, these can be set on the context object, and the handler can return the response content as it would irrespective of needing to set those dynamic response values. This change would render this PR obsolete.

For example, this:

```py
@get("/resources", sync_to_thread=False)
def retrieve_resource() -> Response[Resource]:
    return Response(
        Resource(
            id=1,
            name="my resource",
        ),
        headers={"MY-HEADER": "xyz"},
        cookies=[Cookie(key="my-cookie", value="abc")],
    )
```

Would become:

```py
@get("/resources", sync_to_thread=False)
def retrieve_resource(response_context: ResponseContext) -> Resource:
    response_context.headers.append(("MY-HEADER", "xyz"))
    response_context.cookies.append(Cookie(key="my-cookie", value="abc"))
    return Resource(id=1, name="my resource")
```

2. We add the ability to set a callable on `ResponseHeader` and `Cookie`. This removes the complexity of having to set _both_ the `after_request` and the associated `response_header` entry. In the below example I show that callable receiving the request and response context, but that's just for demo, we'd have to consider what it should receive in practice.

E.g, this:

```py
def after_request_handler(response: Response) -> Response:
    response.headers.update({"Random-Header": str(randint(1, 100))})
    return response

router = Router(
    path="/router-path",
    route_handlers=[retrieve_resource],
    after_request=after_request_handler,
    response_headers=[
        ResponseHeader(
            name="Random-Header",
            description="a random number in the range 1 - 100",
            documentation_only=True,
        )
    ],
)
```

Becomes

```py
def generate_random_header(request: Request, response_context: ResponseContext) -> Response:
    return "Random-Header", str(randint(1, 100))

router = Router(
    path="/router-path",
    route_handlers=[retrieve_resource],
    response_headers=[
        ResponseHeader(
            name="Random-Header",
            description="a random number in the range 1 - 100",
            value=generate_random_header,
        )
    ],
)
```

3. `after_request` hooks receive the response context object and can optionally return a response, or can modify the response context in place. We postpone construction of the `Response` object until _after_ the `after_request` hook is called. If the `after_request` hook returns a `Response` object, that is the response object used to return to the client, otherwise one is created from the context. This should then allow us to make response objects immutable.

provinzkraut: These are some great ideas and I think this is the functionality we should work towards to.

**Removing support for returning `Response` objects from handlers**

I don't think this is a good idea, as forces you to use layers of indirection, which might just not be necessary for many use cases. We should have a way for users to say "I want to return this exact response, with this body and these headers", without making them jump through various hoops.

If we want to simplify things, returning `Response` objects can be perfectly fine as long as we make it clear that this is a lower level functionality and don't attempt to support all the things we otherwise do. It could simply be an immutable instance of `Response` with a set content. We could maybe even restrict this to only accepting `bytes` / `str` as a value to draw a clear boundary.


**`ResponseContext`**

One issue I see with this is that `ResponseContext` is very similar to `ResponseContainer`; They do the same thing conceptually (from a user's perspective), that is, provide an object that receives additional information and will later be turned into a `Response`. Maybe instead of introducing yet another concept, we should look into ways to merge the two?

<hr>

Overall I think what we need is:

1. A low level ASGI callable that's responsible for sending a response to the server
2. A way for handlers to return "response encodable" objects, i.e. things that aren't responses but contain additional response information such as headers or cookies
3. A way for hooks to modify / replace the response information / replace the response entirely

I think all this, and the functionality you described above could be achieved by introducing an `ASGIResponse`, which would be responsible for doing the actual sending and only receives `str`/`bytes`, plus other information (headers, cookies).

We could then unify the concepts of `ResponseContainer` and `ResponseContext` in a `Response` class. This can be mutable, passed around, injected, replaced, etc., to allow the flexibility you described above. At the end it would then be turned into an `ASGIResponse`.

Since the `ResponseContainer`s simply exist to be able to construct `Response`s that make use of information not available at the time they're being constructed, they'd become obsolete.

A significant part of this though is semantics, and the same could be done by making `Response` taking the place of `ASGIResponse` and having `ResponseContainer` act as the intermediary value. I think though from a user's perspective, operating on a `Response` object is more intuitive, simply because it's more familiar.

litestar/dto/factory/utils.py
```diff
@@ -51,3 +56,26 @@ def parse_configs_from_annotation(parsed_type: ParsedType) -> tuple[DTOConfig, .
...
+        The data model type.
+    """
+    if parsed_type.is_optional:
+        return resolve_model_type(next(t for t in parsed_type.inner_types if not t.is_subclass_of(NoneType)))
+
+    if parsed_type.is_subclass_of((DTOData, Response)):
+        return resolve_model_type(parsed_type.inner_types[0])
+
+    if parsed_type.is_collection:
+        if len(parsed_type.inner_types) != 1:
```
Goldziher, line 77: I'd invert this - first return, than raise

litestar/dto/factory/_backends/abc.py
```diff
@@ -289,6 +290,15 @@ def encode_data(self, data: Any, connection_context: ConnectionContext) -> Lites
         Returns:
             Encoded data.
         """
+        if isinstance(data, Response):
+            data.content = transfer_data(
```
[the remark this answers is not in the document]
peterschutt: Good point - I'll address this in #1777 as I don't think this is going to go through.

litestar/dto/factory/utils.py
```diff
@@ -51,3 +56,26 @@ def parse_configs_from_annotation(parsed_type: ParsedType) -> tuple[DTOConfig, .
...
+    """Resolve the data model type from a parsed type.
+
+    Args:
+        parsed_type: A parsed type annotation that represents the annotation used to narrow the DTO type.
+
+    Returns:
+        The data model type.
+    """
+    if parsed_type.is_optional:
+        return resolve_model_type(next(t for t in parsed_type.inner_types if not t.is_subclass_of(NoneType)))
```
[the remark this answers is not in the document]
peterschutt: its the same as doing this:

`[t for t in parsed_type.inner_types if not t.is_subclass_of(NoneType)][0]`

except it doesn't need to build the intermediate list and breaks early.

peterschutt:
Switching the back to draft as it is now dependent upon #1790

peterschutt:
Closing as this is now handled by the generic approach in #1801

PR #1790, opened 2023-06-07
RFC: Response refactoring

peterschutt:
PR separates the user-configurable components of the response from the ASGI components.

This adds a new response layer below `Response`, that is intended to be an immutable response type. The current `Response` layer will remain mutable and be "the thing" that users interact with.

Still need to look at merging the response container concepts with the `Response` layer.

Based on [this discussion](https://github.com/litestar-org/litestar/pull/1782#discussion_r1216707199)

### Pull Request Checklist

### Description

### Close Issue(s)


litestar/handlers/http_handlers/_utils.py
```diff
@@ -77,13 +77,10 @@ async def create_response(data: Any) -> ASGIApp:
...
             type_encoders=type_encoders,
+            raw_headers=raw_headers,
         )
-        response.raw_headers = raw_headers
 
-        if after_request:
-            return await after_request(response)  # type: ignore
-
-        return response
+        return (await after_request(response) if after_request else response).to_asgi_response()  # type: ignore
```
provinzkraut, line 83: Is this necessary?

`Response` does `await self.to_asgi_response` in its `__call__` and this expression can't resolve to anything other than `Response` or `ASGIResponse`, right?

peterschutt: Oh, I meant to take the `__call__` off `Response` - I was just trying to change things gradually so that I didn't break all the tests in one go.

peterschutt: Thanks, good catch!

Goldziher:
Before finishing this, please make sure to benchmark. I optimized the response with a lot of work, we need to ensure there r no regressions

litestar/response/base.py
```diff
@@ -113,6 +113,41 @@ def __init__(
         self.is_head_response = is_head_response
         self.encoding = encoding
 
+    @staticmethod
+    def render(content: Any, media_type: str, enc_hook: Serializer, encoding: str) -> bytes:
```
provinzkraut, line 117: Why not move this to `Response` and have `ASGIResponse` accept `str`/`bytes` only?
`Response` could call it in `to_asgi_response`

provinzkraut: This would be an even stronger separation of concerns, since `ASGIResponse` would then *only* responsible for the sending part, and everything else would be handled externally, which is kinda of what we're going for, right?

provinzkraut: Thinking about it, it's also better from a user's perspective, since we plan to have them interact with `Response` for the most part, so customisation could happen there more easily.

peterschutt: OK good point.

litestar/response/base.py
```diff
@@ -377,6 +333,34 @@ def delete_cookie(
...
+        if isinstance(content, bytes):
+            return content
+
+        try:
+            if media_type.startswith("text/") or isinstance(content, str):
+                if not content:
+                    return b""
+
+                # TODO: refactor so this cast is unnecessary. The cast is necessary because the type of 'content'
+                #  has not been narrowed down to 'str' by this point. So, can it only be 'str' at this point?
```
provinzkraut: Can `content` not be something other than a `str` here still? We check above if it's a `str` *or* has a `text/` media type, so it could be anything really.

peterschutt: Exactly, we haven't narrowed `content` down to `str`, yet we cast to `str` and call encode on it.. what if it isn't a `str`? Can it not be a `str`? IDK, hence the TODO. Obviously no tests failing but I still want to clean this up, as it doesn't feel right.

peterschutt: what I wanted to check was if there was some earlier constraint that meant that a media type starting with `"text/" _had_ to be `str`... is the media type check even necessary at this point? Maybe its to catch a `"text/..."` media type with `None` content so we return `b""` in that case??

peterschutt: We have tests for empty collections with `"text/..."` media types not raising errors - so I think it is that.

provinzkraut: I think the change you made is correct. I am not aware of any assertions we're doing to narrow down the content before we pass it to the response that would catch this, and even if we did, we shouldn't. If it can be a `str` within this context, we should handle that case.

litestar/_openapi/responses.py
```diff
@@ -69,24 +75,29 @@ def create_success_response(  # noqa: C901
...
+    ):
+        if return_annotation is TemplateResponse:
             return_annotation = str
             route_handler.media_type = get_enum_string_value(MediaType.HTML)
-        elif is_class_and_subclass(get_origin(return_annotation), LitestarResponse):
-            return_annotation = get_args(return_annotation)[0] or Any
+        elif return_type.is_subclass_of(LitestarResponse):
+            return_annotation = return_type.inner_types[0].annotation if return_type.inner_types else Any
+            if not route_handler.media_type:
+                route_handler.media_type = get_enum_string_value(MediaType.JSON)
```
peterschutt, line 100: We don't actually know what the media type should be here, as it can be dynamically set on the response class returned by the handler. Not much we can do though, except suggest that the user set the appropriate media type on the handler if they want the docs correct.

litestar/handlers/http_handlers/base.py
```diff
@@ -499,12 +497,15 @@ def _validate_handler_function(self) -> None:
                 "If the function should return a value, change the route handler status code to an appropriate value.",
             )
 
+        if (
+            before_request := self.resolve_before_request()
+        ) and not before_request.parsed_signature.return_type.is_optional:
+            return_type = before_request.parsed_signature.return_type
```
peterschutt, line 503: Inspecting the return annotation of the `before_request` handler in order to determine what the _actual_ response of the handler is, is something that we do somewhere else also. However, I don't believe we take into account that the `before_request` handler may optionally return a response, in which case the true response type of the handler is the union of the non-optional component of the `before_request` handler, and the response of the handler itself.

peterschutt: Here:
https://github.com/litestar-org/litestar/blob/f9952573ba3a0f7e5aa742682c437bf412f1b5ab/litestar/handlers/http_handlers/base.py#L420-L423

This is to determine which response handler should be used. I believe that if the `before_request` handler has a return annotation of `RedirectResponse | None`, then we probably don't end up with the right response handler in this case.

litestar/handlers/http_handlers/base.py
```diff
@@ -499,12 +497,15 @@ def _validate_handler_function(self) -> None:
...
+
         if not self.media_type:
-            if return_type.annotation is AnyStr or return_type.is_subclass_of(
-                (str, bytes, Redirect, File, FileResponse)
-            ):
+            if return_type.is_subclass_of((str, bytes)) or return_type.annotation is AnyStr:
                 self.media_type = MediaType.TEXT
-            else:
+            elif not return_type.is_subclass_of(Response):
```
peterschutt, line 508: This change means that if a handler is annotated with a `Response` return type, then the handler media type is left empty. This lets me identify a media type that is explicitly set when merging response data with layered parameters for handlers that return `Response`, etc.

litestar/openapi/controller.py
```diff
@@ -11,34 +11,19 @@
...
-__all__ = ("OpenAPIController", "OpenAPISchemaResponse")
+__all__ = ("OpenAPIController",)
 
 
 if TYPE_CHECKING:
     from litestar.connection.request import Request
     from litestar.openapi.spec.open_api import OpenAPI
 
-
-class OpenAPISchemaResponse(Response):
```
peterschutt, line 26: This is removed as we return `ASGIResponse`s directly from the handlers on the controller and are passing encoded data directly to them.

litestar/response/base.py
```diff
@@ -24,28 +22,172 @@
...
+        *,
+        background: BackgroundTask | BackgroundTasks | None = None,
+        body: bytes = b"",
+        content_length: int | None = None,
+        cookies: list[Cookie] | None = None,
+        encoded_headers: list[tuple[bytes, bytes]] | None = None,
+        encoding: str = "utf-8",
+        headers: dict[str, Any] | None = None,
+        is_head_response: bool = False,
+        media_type: str | None = None,
```
peterschutt, line 58: I could still allow this to receive `MediaType` enum values - it would clean up usage a bit.

litestar/response/base.py
```diff
@@ -218,128 +339,83 @@ def delete_cookie(
...
+        background: BackgroundTask | BackgroundTasks | None,
+        cookies: list[Cookie] | None,
+        encoded_headers: list[tuple[bytes, bytes]] | None,
+        headers: dict[str, str] | None,
+        is_head_response: bool,
+        media_type: MediaType | str | None,
+        request: Request,
+        status_code: int | None,
+        type_encoders: TypeEncodersMap | None,
+    ) -> ASGIResponse:
```
peterschutt, line 381: Might make `app` and `request` positional and the rest kwargs with default values. It would make it easier to use this method downstream.

litestar/utils/sync.py
```diff
@@ -81,7 +81,7 @@ def set_parsed_signature(self, namespace: dict[str, Any]) -> None:
         """
         from litestar.utils.signature import ParsedSignature
 
-        self._parsed_signature = ParsedSignature.from_fn(self.ref.value, namespace)
+        self._parsed_signature = ParsedSignature.from_fn(unwrap_partial(self.ref.value), namespace)
```
peterschutt, line 84: I found we were generating the parsed signature on the `AsyncPartial` instance, causing the return of the `before_request` handlers to always appear as `Any`.

Whether this PR is merged, or not, this should go into main.

litestar/response/base.py
```diff
@@ -24,28 +22,172 @@
...
+        "encoding",
+        "is_head_response",
+        "status_code",
+    )
+
+    def __init__(
+        self,
+        *,
+        background: BackgroundTask | BackgroundTasks | None = None,
+        body: bytes = b"",
```
provinzkraut, line 51: I think this can be `str` as well. Makes the usage a bit more convenient and aligns with all the low-level de/encoders we have.

litestar/response/base.py
```diff
@@ -24,28 +22,172 @@
...
+        content_length: int | None = None,
+        cookies: list[Cookie] | None = None,
+        encoded_headers: list[tuple[bytes, bytes]] | None = None,
+        encoding: str = "utf-8",
+        headers: dict[str, Any] | None = None,
```
provinzkraut, line 56: I don't think this should have both header types. A single parameter that accepts encoded headers should be what we want. This is intended to be a low level implementation of the ASGI response protocol, so encoding and such should happen outside.

litestar/response/base.py
```diff
@@ -24,28 +22,172 @@
...
+                (f"{media_type}; charset={encoding}" if media_type.startswith("text/") else media_type).encode(
+                    "latin-1"
+                ),
+            ),
+        )
+
+        if content_length is None:
+            content_length = len(body)
+
+        if "content-length" not in headers and content_length:
```
provinzkraut, line 105: What if it's in `encoded_headers`? See comment above about those.

litestar/utils/helpers.py
```diff
@@ -70,3 +83,51 @@ def unwrap_partial(value: MaybePartial[T]) -> T:
...
+    """Given two sets of cookies, return a unique list of cookies, that are not marked as documentation_only.
+
+    Args:
+        local_cookies: Cookies returned from the local scope.
+        layered_cookies: Cookies returned from the layers.
+
+    Returns:
+        A unified list of cookies
+    """
+    # TODO: check this precedence, should local cookies be overwritten by layered cookies?
```
provinzkraut, line 120: Should be the other way around, right?

litestar/response/redirect.py
```diff
@@ -1,71 +1,139 @@
...
+RedirectStatusType = Literal[301, 302, 303, 307, 308]
+"""Acceptable status codes for redirect responses."""
+
+
+class ASGIRedirectResponse(ASGIResponse):
+    """A low-level ASGI redirect response class."""
+
+    def __init__(
+        self,
+        url: str | bytes,
```
provinzkraut, line 35: I think `path` is correct, since it does not have to be an actual URL.

litestar/handlers/http_handlers/base.py
```diff
@@ -421,23 +419,22 @@ def get_response_handler(self, is_response_type_data: bool = False) -> Callable[
...
                 self._response_handler_mapping["default_handler"] = response_type_handler
-            elif return_type.is_subclass_of(ResponseContainer):
-                self._response_handler_mapping["default_handler"] = create_response_container_handler(
-                    after_request=after_request,
-                    cookies=cookies,
-                    headers=headers,
-                    media_type=media_type,
-                    status_code=self.status_code,
-                )
             elif is_async_callable(return_annotation) or return_annotation is ASGIApp:
```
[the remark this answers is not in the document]
peterschutt: The tests that fail when I remove the cookie setting on the asgi responses are the Starlette response tests:

```
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[asyncio-expected_response0] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[asyncio-expected_response1] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[asyncio-expected_response2] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[asyncio-expected_response3] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[trio-expected_response0] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[trio-expected_response1] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[trio-expected_response2] - KeyError: 'set-cookie'
FAILED tests/handlers/http/test_to_response.py::test_to_response_returning_starlette_response[trio-expected_response3] - KeyError: 'set-cookie'
```
I still think we should remove this and let the users that are dealing in Starlette responses (I cannot imagine this is a big number) manage this themselves. WDYT?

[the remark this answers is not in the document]
peterschutt: I could also create another response handler that specifically deals with the starlette response types.

litestar/response/base.py
peterschutt: I do agree, and I did start with that but landed here for practical reasons.

First, centralizing the encoding logic - although that can be done in other ways so don't worry about that.

Second, in ASGIFileResponse, a few headers are set right before `http.response.start` is sent due to the source of those values potentially coming from an awaitable that is resolved within the `start_response()` coroutine. Given that those are unconditionally set by `ASGIFileResponse`, I `pop()` them from `headers` in `__init__()`.

litestar/response/base.py
peterschutt: This problem isn't unique to this PR though.

In main, we do care to not duplicate or overwrite the value using `setdefault()`:

https://github.com/litestar-org/litestar/blob/3c6480e5dca75e7df37247c548ac1fbb1425d717/litestar/response/base.py#L110-L113

So if `"content_type"` is in `headers` on instantiation of the `Response`, that is OK. But generally, headers are not passed to the response as a mapping, rather they are pre-encoded and we don't make any effort to record whether it is set on the layers or not in that case:

https://github.com/litestar-org/litestar/blob/3c6480e5dca75e7df37247c548ac1fbb1425d717/litestar/handlers/http_handlers/_utils.py#L67-L71

Nor any attempt to manage duplicate headers when they are merged with the response headers/cookies:

https://github.com/litestar-org/litestar/blob/3c6480e5dca75e7df37247c548ac1fbb1425d717/litestar/response/base.py#L269-L275

So we are paying a "correctness" price for the pre-encoding optimization. There might be ways to work around this however, without losing the pre-encoding optimization. E.g., we could keep a mapping of pre-encoded header name, to its index in the encoded headers collection that is passed to the asgi response... but IDK.. what if we make `encoded_headers` accept an object that keeps such state?

litestar/utils/helpers.py
peterschutt: There's actually no issue here - I just wrote the TODO to ensure I circled back to take a look, but the local headers unpacked into the set first means they take precedence. I see way more of this pattern to unpack into mappings where the 2nd unpacked takes precedence, and was looking at this with that frame of mind.

litestar/response/redirect.py
peterschutt: Agree - this was `path` on the container type and `url` on the response type. I'll change.

litestar/handlers/http_handlers/_utils.py
```diff
@@ -98,22 +78,31 @@ async def handler(
...
+            content=data,
+            media_type=media_type,
+            status_code=status_code,
+            type_encoders=type_encoders,
+        )
 
-def filter_cookies(local_cookies: frozenset[Cookie], layered_cookies: frozenset[Cookie]) -> list[Cookie]:
-    """Given two sets of cookies, return a unique list of cookies, that are not marked as documentation_only.
+        if after_request:
+            response = await after_request(response)  # type: ignore[arg-type,misc]
```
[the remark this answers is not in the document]
peterschutt: > IMO we _should_ have:
> 
>     * `before_request`: Receives and returns a `Request` instance, called before the handler function
> 
>     * `after_request`: Receives and returns a `Response` instance. Called after the handler function. Not called if the handler returns an ASGI application (including `ASGIResponse`)
> 
>     * `before_response`: Receives and returns an `ASGIApp`. Called after `Response` instances have been resolved to `ASGIResponse` and just before the response ASGI app is awaited

Should we do an RFC for this then?

docs/release-notes/whats-new-2.rst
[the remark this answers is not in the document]
peterschutt: Yes, absolutely - I'll do this once we have a green light on the rest.

litestar/handlers/http_handlers/_utils.py
```diff
@@ -157,58 +146,49 @@ def normalize_headers(headers: frozenset[ResponseHeader]) -> dict[str, str]:
...
         media_type: The response media type.
         status_code: The response status code.
+        type_encoders: A mapping of types to encoder functions.
 
     Returns:
         A handler function.
     """
+
     normalized_headers = normalize_headers(headers)
+    cookie_list = list(cookies)
```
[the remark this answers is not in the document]
peterschutt: I'm sure it doesn't need to be - I'll revisit.

litestar/handlers/http_handlers/base.py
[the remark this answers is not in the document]
provinzkraut: > I still think we should remove this and let the users that are dealing in Starlette responses (I cannot imagine this is a big number) manage this themselves. WDYT?

Agreed

litestar/handlers/http_handlers/_utils.py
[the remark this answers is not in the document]
provinzkraut: Yeah, I think so.

litestar/_openapi/responses.py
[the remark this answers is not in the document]
peterschutt: Created #1838

litestar/handlers/http_handlers/base.py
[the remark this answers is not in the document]
peterschutt: Created #1790

PR #1810, opened 2023-06-11
Make `preferred_validation_backend` private

provinzkraut:
- Make the `preferred_validation_backend` parameter private
- Make the `Litestar.preferred_validation_backend` attribute private
- Remove the `preferred_validation_backend` from `AppConfig`
- Simplify logic for selecting a validation backend: If the preferred backend is set to `attrs`, or the signature contains attrs types, `attrs` is selected. In all other cases, Pydantic will be used.

This exposes less of the implementation details to the user, so that we can more easily implement the proposed changes in preparation of the 2.0 release, while keeping current functionality intact.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -39,7 +39,7 @@ from litestar.static_files.base import StaticFiles
 from litestar.stores.registry import StoreRegistry
 from litestar.types import Empty
 from litestar.types.internal_types import PathParameterDefinition
-from litestar.utils import AsyncCallable, join_paths, unique
+from litestar.utils import AsyncCallable, join_paths, unique, warn_deprecation
 from litestar.utils.dataclass import extract_dataclass_items
 from litestar.utils.predicates import is_async_callable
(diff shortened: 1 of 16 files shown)


litestar/app.py
```diff
@@ -210,6 +209,7 @@ def __init__(
         websocket_class: type[WebSocket] | None = None,
         lifespan: list[Callable[[Litestar], AbstractAsyncContextManager] | AbstractAsyncContextManager] | None = None,
         pdb_on_exception: bool | None = None,
+        _preferred_validation_backend: Literal["attrs", "pydantic"] | None = None,
```
provinzkraut, line 212: This is not ideal, but the only good way to solve this. The alternative - removing this parameter outright - makes it impossible to change during testing, since it is being referenced at the time the signature models are created, which happens inside of `__init__`.

peterschutt: Still, I think it achieves the objective of not encouraging dependence on the feature, so SGTM.

peterschutt (approved):

PR #1811, opened 2023-06-11
Fix: Documentation for Authentication middleware has wrong imports

ctrl-Felix:
The documentation for the v2 contains
```python
from litestar.authentication import (
    AbstractAuthenticationMiddleware,
    AuthenticationResult,
)
```

which is not existing.

The correct imports are `from litestar.middleware`.

AbstractAuthenticationMiddleware,  AuthenticationResult are imported in : https://github.com/litestar-org/litestar/blob/main/litestar/middleware/__init__.py


provinzkraut (approved):

provinzkraut:
@all-contributors add @ctrl-Felix to doc

PR #1816, opened 2023-06-11
feat: Adds Timezone aware datetime.  Enforces UTC timestamps are stored and returned.

cofin:
### Pull Request Checklist

### Description

This PR modifies the enhances the handling of Timestamps in the database:
-  `created` and `updated` columns renamed to `created_at` and `updated_at` columns
- Adds a `DateTimeUTC` type that enforces UTC timestamps stored in the database.
- Sets the population of the `created_at` column to be generated by server side timestamps

### Close Issue(s)


litestar/contrib/sqlalchemy/base.py
```diff
@@ -52,23 +49,6 @@
 """Templates for automated constraint name generation."""
 
 
-@listens_for(Session, "before_flush")
-def touch_updated_timestamp(session: Session, *_: Any) -> None:
```
peterschutt, line 56: Is the idea that this is replaced by the `server_default`s?

This pattern was to support joined-table inheritance scenarios, where the `updated_at` field, and the modified model attribute value may not live in the same table.

cofin: That's correct, and it's one of the reasons I went to this pattern.  Is this behavior actually desired?  I noticed that when I added a child record to a model, this would cause the updated_at flag to be updated on the Parent  

This seemed off to me as there was no actual updated done to the records?

peterschutt: > I noticed that when I added a child record to a model

I.e., updated a collection attribute to add a child model to its collection? `model.things.append(thing)`, triggers the timestamp on the model that owns the collection?

cofin: Yeah, I can add a test case to confirm this behavior.  But, I originally started looking at this because I noticed that the `updated_at` timestamp changed without an actual update to that record.  For instance, if I had `User`, `TeamMembers`, and `Team`, the `User.updated_at` would change when I called `User.teams.append(TeamMember(...))`.

peterschutt: Yeh, that's not right - however I'd put that in the bug category, rather than invalidation of the actual intention of the concept.

Aside: is `server_default=func.current_timestamp()` enough to duplicate the behavior, even if the attribute changes are occurring on the base class? For example, in maria db I've had to do `server_default=db.text("CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP")` to get the timestamp to change, otherwise the column would be no different to `created_at`. However, it's been a while since I configured those, perhaps that's not required any more?

Back to the listener: its intended utility is to handle where a child type in a polymorphic inheritance hierarchy is modified, but the `updated_at` timestamp doesn't change because the attribute that was modified lives in a different table to the table that represents the child attribute.

```py
class Employee(Base):
    __tablename__ = "employee"
    id: Mapped[int] = mapped_column(primary_key=True)
    updated_at: Mapped[datetime] = mapped_column(server_default=datetime.now)

    __mapper_args__ = {
        "polymorphic_identity": "employee",
        "polymorphic_on": "type",
    }

class Engineer(Employee):
    __tablename__ = "engineer"
    id: Mapped[int] = mapped_column(ForeignKey("employee.id"), primary_key=True)
    subclass_attr: Mapped[str]

    __mapper_args__ = {
        "polymorphic_identity": "engineer",
    }
```
In this pattern, updating the `subclass_attr` attribute on an instance of `Engineer` will not update the timestamp in the employee table. At a database layer, that does make sense, nothing changed on that table. However, interpreting the models at the application layer, where the fact that the underlying data is spread across two tables is almost totally abstracted, it is surprising IMO. E.g., `engineer.subclass_attr = "updated"` doesn't change `engineer.updated_at`.

I'm not a fan of the fact that we have to iterate over every item in `session.dirty` on the way out to the database to achieve this, but I'm less of a fan of `engineer.updated_at` not changing after an attribute on the model was changed.

cofin: I honestly don't know how it will work in all of the inheritance cases, and I completely agree re: the need for it to reliably update.  Because of that, I think I decouple this change from this PR?  I'd like to add in some additional tests to make sure we handle everything properly.  

I'll bring back in the event listener and we can address at a later date.

peterschutt: Yep, I think that makes sense - open an issue for the collection append bug - hopefully there is a way we can distinguish that from changes to immediate attributes without too much overhead.

litestar/contrib/repository/testing/generic_mock_repository.py
JacobCoffee, line 1: Why do we do has created/updated at instead of "model_created_at"?

litestar/contrib/sqlalchemy/base.py
provinzkraut: > However, interpreting the models at the application layer, where the fact that the underlying data is spread across two tables is almost totally abstracted, it is surprising

I think that you can reasonably make an argument for either case. Some people will want to reflect the actual changes to the table, some won't care about that and prefer to have a level of abstraction.

It sounds to me like this *could* be solved by introducing a `cascade_audit_events` keywords somewhere, that controls this behaviour, but that would certainly add more complexity.

provinzkraut (approved):

litestar/contrib/repository/testing/generic_mock_repository.py
cofin: I think this has been updated?

litestar/contrib/sqlalchemy/base.py
cofin: I've reverted the change and we can dig into this one separately.  Thanks for all the input on this one.

PR #1823, opened 2023-06-13
Fix OpenAPI schema generation for aliased Pydantic field in responses

### LINKED ISSUE #1809: Bug: OpenAPI schema with Pydantic field aliases
### Description

The current situation with respect to Pydantic field aliases is the following:
1. Request data validation uses the alias names
2. Response data use the field (non-alias) names
3. OpenAPI spec documents both request and response with the aliased names

Whether (2) should change to be consistent with (1) (or at least be easier to override than suggested [here](https://github.com/orgs/litestar-org/discussions/1458#discussioncomment-5555105)) is debatable. (3) however is clearly wrong.

### MCVE

```python
from litestar import Controller, Litestar, post


class UserOrder(BaseModel):
    id: int = Field(alias="_id")
    name: str


    @post()
    async def create(self, data: UserOrder) -> UserOrder:
        return data


app = Litestar(route_handlers=[UserController])
```

### Steps to reproduce

```bash
curl localhost:8000/ -d '{"id":1, "name": "foo"}'
{"status_code":400,"detail":"Validation failed for POST http://localhost:8000/","extra":[{"key":"_id","message":"field required"}]}

curl localhost:8000/ -d '{"_id":1, "name": "foo"}'
{"id":1,"name":"foo"}
```

### Screenshots

Screenshot 2023-06-11 at 2 29 42 PM

### Litestar Version

Main branch

gsakkis:
### Description

- Add a `prefer_alias` parameter to `litestar._openapi.schema_generation.create_schema` & friends
- Set it to False for response schemas and True otherwise.

### Close Issue(s)

Fixes #1809

### DIFF
--- a/litestar/_openapi/parameters.py
+++ b/litestar/_openapi/parameters.py
@@ -48,6 +48,7 @@ def create_path_parameter_schema(
         generate_examples=generate_examples,
         plugins=[],
         schemas=schemas,
+        prefer_alias=True,
     )
 
 
@@ -128,7 +129,9 @@ def create_parameter(
         parameter_name = kwargs_model.query if kwargs_model and kwargs_model.query else parameter_name
 
     if not result:
-        result = create_schema(field=signature_field, generate_examples=generate_examples, plugins=[], schemas=schemas)
+        result = create_schema(
+            field=signature_field, generate_examples=generate_examples, plugins=[], schemas=schemas, prefer_alias=True
+        )
 
     schema = result if isinstance(result, Schema) else schemas[result.value]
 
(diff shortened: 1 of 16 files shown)


peterschutt:
I think you'd be right to just set this to `False` in the place where it is being called from within the DTO backend as that is virtually always sending through msgspec structs, not pydantic models.

cofin:
@gsakkis thanks for working this one!  I'm happy to see to see this rough edge resolved.

JacobCoffee (approved):

Goldziher:
@all-contributors add @gsakkis code

PR #1842, opened 2023-06-18
Refactor openapi create_schema

gsakkis:
### Pull Request Checklist

### Description

This PR was prompted by https://github.com/litestar-org/litestar/pull/1823 that introduced the `prefer_alias` parameter to `create_schema`. Although this parameter is actually used in [just one line](https://github.com/litestar-org/litestar/blob/main/litestar/_openapi/schema_generation/schema.py#L615), the diff was more than a hundred lines due to the need of propagating it over the (recursive) `create_schema_for_*` function calls.

To address this:
- The `litestar._openapi.schema_generation.schema.SchemaCreator` (data)class is introduced for encapsulating the set of `create_schema` parameters. 
- The numerous `create_schema_for_*` functions are converted to `SchemaCreator` methods with simplified signatures, often taking just a single parameter as all the rest are stored in `self`.


litestar/_openapi/schema_generation/schema.py
```diff
@@ -350,533 +354,421 @@ def create_schema_for_annotation(annotation: Any) -> Schema | None:
...
-    Args:
-        field: A signature field instance.
-        generate_examples: Whether to generate examples if none are given.
-        plugins: A list of plugins.
-        schemas: A mapping of namespaces to schemas - this mapping is used in the OA components section.
-        prefer_alias: Whether to prefer the alias name for the schema.
+    return Schema()
+
+
+@dataclass(frozen=True)
```
Goldziher, line 360: Why dataclass? I think this class shouldn't be a dataclass. 

Is it computed once per app? 

Also, please add `slots`

Goldziher (approved):
Excellent stuff. Glad to see this.

litestar/_openapi/schema_generation/schema.py
gsakkis: Mostly as a convenience to get `__init__` for free, plus the `frozen=True` doesn't hurt either. I tend to make most classes dataclasses these days for the free `__init__`, `__repr__`, `__eq__` etc, not just "dumb data" containers. Is there a downside?

I'll add `__slots__` (in 3.10+ that would be just `slots=True`).

Goldziher: It's not a problem as is.

PR #1846, opened 2023-06-18
Refactor openapi create schema v2

gsakkis:
### Pull Request Checklist

### Description

Follow up to https://github.com/litestar-org/litestar/pull/1842:
- Change `SchemaCreator` from dataclass to plain class with `__slots__`
- Replace `create_schema` with `SchemaCreator` across the board


litestar/_openapi/parameters.py
```diff
@@ -160,28 +134,18 @@ def get_recursive_handler_parameters(
...
-        return [
-            create_parameter(
-                generate_examples=generate_examples,
-                parameter_name=field_name,
-                path_parameters=path_parameters,
-                schemas=schemas,
-                signature_field=signature_field,
-            )
-        ]
+        return [create_parameter(signature_field, field_name, path_parameters, schema_creator)]
```
Goldziher, line 137: So, please use kwargs rather than positional arguments. The reason for this is two fold:

1. Kwargs are more resilient to refactoring.
2. Kwargs are more readable. 

The rule of thumb is - above two arguments always use kwargs. Below is optional but i personally prefer using them.

Goldziher (commented):
Very nice. Please see my comment and address it everywhere pertinent.

litestar/_openapi/parameters.py
gsakkis: Done.

FWIW I opted for positional arguments because (a) they are all required and (b) the passed parameters are (usually) variables named identically to the formal parameters so the meaning is obvious.

> Kwargs are more resilient to refactoring.

Not so sure about that; in my experience swapping the order of required parameters (which breaks positional args) is not more common than renaming parameters (which breaks kwargs).

> Kwargs are more readable.

This might be true in general, unless the caller names the passed parameters identically to the formal parameters. To me `return f(x, y, z, w)` is no less readable than `return f(x=x, y=y, z=z, w=w)`, and actually more readable than `return f(z=z, x=x, w=w, y=y)` which tends to happen over time once we stop caring about parameter order.

Goldziher (approved):

JacobCoffee:
Hello @gsakkis, 

You have been quite active in our project. 
Would you accept an invitation to the @litestar-org/members org?
This isn't an ask to do more or anything - This is mostly recognition for your continued contributions! 😄

It also is my opportunity to invite you to weigh in on project discussions on Discord more (if you would like)

JacobCoffee (approved):

gsakkis:
@JacobCoffee thank you for your kind words and invitation to join the org, much appreciated! Looking forward to more contributions :slightly_smiling_face:

Goldziher:
@gsakkis - you will need to rebase. i merged a PR that affected the same files.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -519,7 +538,7 @@ def for_builtin_generics(self, field: SignatureField) -> Schema:
                 },
             )
 
-        cursor_schema = replace(self, generate_examples=False).for_field(field.children[0])  # type: ignore[index]
+        cursor_schema = self.not_generating_examples.for_field(field.children[0])  # type: ignore[index]
```
Goldziher, line 541: so, this is a bit overcomplicated. 

Why not add `generate_examples` as a kwarg to for_field?

litestar/_openapi/schema_generation/schema.py
```diff
@@ -358,17 +357,37 @@ def create_schema_for_annotation(annotation: Any) -> Schema:
...
+            schemas: A mapping of namespaces to schemas - this mapping is used in the OA components section.
+            prefer_alias: Whether to prefer the alias name for the schema.
+        """
+        self.generate_examples = generate_examples
+        self.plugins = plugins if plugins is not None else []
+        self.schemas = schemas if schemas is not None else {}
+        self.prefer_alias = prefer_alias
+
+    @property
+    def not_generating_examples(self) -> SchemaCreator:
```
Goldziher, line 384: this will be wasteful memory wise - we dont need an instance for this. see my other comment.

litestar/_openapi/schema_generation/schema.py
gsakkis: Because it wouldn't be added just to `for_field` but to all (current and future) SchemaCreator methods, which is the whole point of this refactoring.

PR #1847, opened 2023-06-18
Add support for `annotated-types`

Goldziher:
This PR extends support for the [annotated-types](https://pypi.org/project/annotated-types/) library. This library is used in pydantic v2 internally and its good to support it when parsing metadata.

I will make a followup issue to add documentation for this - but I wont have capacity to do this for now (usage is clear from the tests ofc), so leaving it out of scope.


pyproject.toml
```diff
@@ -423,11 +426,8 @@ known-first-party = ["litestar", "tests", "examples"]
 "test_apps/**/*.*" = ["D", "TRY", "EM", "S", "PTH"]
 "tools/**/*.*" = ["D", "ARG", "EM", "TRY", "G", "FBT"]
 
-
 [tool.unasyncd]
 add_editors_note = true
```
provinzkraut, line 430:
```suggestion
add_editors_note = true
ruff_fix = true
```

provinzkraut (changes requested):

PR #1854, opened 2023-06-21
feat(internal): msgspec signature model and pydantic v2 support.

peterschutt:
This PR migrates to using msgspec as the basis for signature modelling. It removes the dependency on attrs and pydantic for this purpose and switching to using msgspec exclusively.


litestar/_signature/models/msgspec_signature_model.py
```diff
@@ -0,0 +1,132 @@
...
+                        k: v
+                        for k in (
+                            "gt",
+                            "ge",
+                            "lt",
+                            "le",
+                            "multiple_of",
+                            "pattern",
+                            "min_length",
+                            "max_length",
```
Goldziher, line 102: No title, description, examples etc? You can see what we copy in metadata.py from annotated stuff.

Goldziher (commented):
Looks good. 

The aim is to replace the other models with this, correct?

peterschutt:
> Looks good. 
> 
> The aim is to replace the other models with this, correct?

Yes that's the plan! I'm just trying to get the tests passing with as few changes as possible, but ran out of time. Will pick up again tomorrow.

litestar/_signature/models/msgspec_signature_model.py
```diff
@@ -0,0 +1,132 @@
...
+                default_value=Empty if default is NODEFAULT else default,
+                kwarg_model=kwargs_container,
+                extra=field_extra,
+            )
+
+        return defstruct(  # type:ignore[return-value]
+            f"{fn_name}_signature_model",
+            struct_fields,
+            bases=(MsgspecSignatureModel,),
+            module=fn_module or "",
```
provinzkraut, line 125: This workaround can be removed with the next msgspec release (https://github.com/jcrist/msgspec/issues/444)

litestar/serialization.py
```diff
@@ -135,9 +136,14 @@ def dec_hook(type_: Any, value: Any) -> Any:  # pragma: no cover
     Returns:
         A ``msgspec``-supported type
     """
+
```
provinzkraut, line 139: I think we just want the reverse of `enc_hook` here?

litestar/_signature/models/msgspec_signature_model.py
peterschutt: Nice!

tests/unit/test_kwargs/test_path_params.py
```diff
@@ -79,7 +79,7 @@ def test_method(
         assert user_id
         assert order_id
 
-    with create_test_client(test_method) as client:
+    with create_test_client(test_method, debug=True) as client:
```
peterschutt, line 82:
```suggestion
    with create_test_client(test_method) as client:
```

tests/unit/test_signature/test_parsing.py
```diff
@@ -118,21 +119,17 @@ async def health_check() -> None:
...
-            "attrs",
-            [{"key": "dep", "message": "invalid literal for int() with base 10: 'thirteen'"}],
-        ),
-        (
-            "pydantic",
-            [{"key": "dep", "message": "value is not a valid integer"}],
+        pytest.param(
+            "attrs", "invalid literal for int() with base 10: 'thirteen'", marks=pytest.mark.skip, id="pydantic"
         ),
+        pytest.param("pydantic", "value is not a valid integer", marks=pytest.mark.skip, id="attrs"),
```
provinzkraut: Skipping these for now, as they will always fail. If we are to drop the other backends, we can just remove them

tests/unit/test_signature/test_parsing.py
```diff
@@ -118,21 +119,17 @@ async def health_check() -> None:
...
-            "attrs",
-            [{"key": "dep", "message": "invalid literal for int() with base 10: 'thirteen'"}],
-        ),
-        (
-            "pydantic",
-            [{"key": "dep", "message": "value is not a valid integer"}],
+        pytest.param(
+            "attrs", "invalid literal for int() with base 10: 'thirteen'", marks=pytest.mark.skip, id="attrs"
         ),
+        pytest.param("pydantic", "value is not a valid integer", marks=pytest.mark.skip, id="pydantic"),
```
provinzkraut, line 127: Skipping these for now, as they will always fail. If we are to drop the other backends, we can just remove them

tests/unit/test_signature/test_parsing.py
```diff
@@ -147,17 +144,20 @@ def test(dep: int, param: int, optional_dep: Optional[int] = Dependency()) -> No
...
         "status_code": HTTP_500_INTERNAL_SERVER_ERROR,
     }
 
 
 @pytest.mark.parametrize(
-    "preferred_validation_backend, error_extra",
-    (("attrs", [{"key": "param", "message": "invalid literal for int() with base 10: 'thirteen'"}]),),
+    "preferred_validation_backend, error_message",
+    (
+        pytest.param("attrs", "invalid literal for int() with base 10: 'thirteen'", marks=pytest.mark.skip, id="attrs"),
```
provinzkraut, line 155: See comment above

tests/unit/test_signature/test_parsing.py
```diff
@@ -172,12 +172,21 @@ def test(dep: int, param: int, optional_dep: Optional[int] = Dependency()) -> No
...
     }
 
 
-def test_client_pydantic_backend_error_precedence_over_server_error() -> None:
+@pytest.mark.parametrize(
+    "backend,expected_error_msg",
+    [
+        pytest.param(
+            "pydantic", "invalid literal for int() with base 10: 'thirteen'", marks=pytest.mark.skip, id="pydantic"
+        ),
```
provinzkraut, line 185: Skipping for now, see comment above

peterschutt:
The remaining failures are from exception handling/error formatting.

litestar/serialization.py
```diff
@@ -125,6 +129,38 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
     raise TypeError(f"Unsupported type: {type(value)!r}")
 
 
+PydanticUUIDType = TypeVar("PydanticUUIDType", bound="UUID1")
+
+
+def _dec_pydantic_uuid(type_: type[PydanticUUIDType], val: Any) -> PydanticUUIDType:
```
peterschutt, line 135: This logic is copied out of pydantic's validators for UUID types - I'd imagine if we can see this to the end, then hooks like this could be added by plugin if users were working with pydantic.

litestar/serialization.py
```diff
@@ -125,6 +129,38 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
...
+        raise ValidationError(f"Invalid UUID version: {val!r}")
+
+    return val
+
+
+def _dec_pydantic(type_: type[BaseModel], value: Any) -> BaseModel:
+    try:
+        return type_.parse_obj(value)
+    except PydanticValidationError as e:
+        raise ValidationError from e
```
peterschutt, line 161: In [the docs](https://github.com/jcrist/msgspec/pull/445) it says we should raise `TypeError` in dec hooks, but that seemed to be raised directly and wouldn't be caught  by the error handling in the `parse_values_from_connection_kwargs()` method.

litestar/_signature/models/msgspec_signature_model.py
peterschutt: These are just the decoding constraints, nothing to do with documentation.

litestar/serialization.py
jcrist: > In https://github.com/jcrist/msgspec/pull/445 it says we should raise TypeError in dec hooks, but that seemed to be raised directly

*( the link you posted above is to a PR, I assume you meant to link something from [here](https://jcristharif.com/msgspec/extending.html#mapping-to-from-native-types))*

The docs don't actually take a stance on what you should raise on encode/decode error, the only error documented there is for raising on types that the `enc_hook`/`dec_hook` don't handle. I'm going to update the docs to recommend raising `NotImplementedError` in those cases instead of `TypeError`.

For fixing your issue here, I'll push a fix upstream to `msgspec` to wrap any `ValueError`/`TypeError`/`ValidationError` raised in a `msgspec.ValidationError` with the standard additional context. Since `PydanticValidationError` subclasses from `ValueError` this should be handled transparently.

I've opened https://github.com/jcrist/msgspec/issues/456 for tracking this.

JacobCoffee: Thanks @jcrist !

peterschutt: Thanks @jcrist.

The `TypeError` is referenced in the api docs for [`convert.dec_hook`](https://jcristharif.com/msgspec/api.html#msgspec.convert) (which is where the link was supposed to reference):

> dec_hook (callable, optional) – ... This hook should transform obj into type type, or raise a TypeError if unsupported.

litestar/_signature/utils.py
```diff
@@ -110,11 +110,7 @@ def _get_signature_model_type(
     preferred_validation_backend: Literal["pydantic", "attrs"],
     parsed_signature: ParsedSignature,
 ) -> type[SignatureModel]:
-    if preferred_validation_backend == "attrs" or _any_attrs_annotation(parsed_signature):
-        from litestar._signature.models.attrs_signature_model import AttrsSignatureModel
-
-        return AttrsSignatureModel
-    return PydanticSignatureModel
+    return MsgspecSignatureModel
```
Goldziher, line 113: We no longer need the abstraction layer which makes signature models generic. It can be removed..

peterschutt: absolutely - there is a lot to be tidied up before this PR is ready - just shooting for MVP at the moment

litestar/_signature/models/base.py
```diff
@@ -52,9 +56,7 @@ def _create_exception(cls, connection: ASGIConnection, messages: list[ErrorMessa
             if ("key" in err_message and err_message["key"] not in cls.dependency_name_set) or "key" not in err_message
         ]:
             return ValidationException(detail=f"Validation failed for {method} {connection.url}", extra=client_errors)
-        return InternalServerException(
-            detail=f"A dependency failed validation for {method} {connection.url}", extra=messages
-        )
+        return InternalServerException
```
peterschutt, line 59: Similar to #1857 - we shouldn't be leaking any implementation detail from ISEs.

This should be in a separate PR as we want this weather we get this one merged or not.

litestar/_signature/models/base.py
```diff
@@ -69,31 +71,25 @@ def _build_error_message(cls, keys: Sequence[str], exc_msg: str, connection: ASG
...
+        elif key in cls.fields and isinstance(cls.fields[key].kwarg_model, ParameterKwarg):
+            kwarg_model = cast("ParameterKwarg", cls.fields[key].kwarg_model)
+            if kwarg_model.cookie:
+                message["source"] = "cookie"
+            elif kwarg_model.header:
+                message["source"] = "header"
+            else:
+                message["source"] = "query"
+        else:
+            message["source"] = "body"
```
peterschutt, line 92: Refactor of this method has removed the need to have the key start with `data` to know that the source should be `"body"`.

litestar/_signature/models/msgspec_signature_model.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        messages: list[ErrorMessage] = []
+        try:
+            return convert(kwargs, cls, strict=False, dec_hook=dec_hook).to_dict()
+        except PydanticValidationError as e:
+            for exc in e.errors():
+                keys = [str(loc) for loc in exc["loc"]]
+                message = super()._build_error_message(keys=keys, exc_msg=exc["msg"], connection=connection)
+                messages.append(message)
+            raise cls._create_exception(messages=messages, connection=connection) from e
```
peterschutt, line 69: This is leveraging the fact that the pydantic error is raised out of the dec hook to format the exception as we otherwise do.

After https://github.com/jcrist/msgspec/issues/456 we will hopefully be able to access the pydantic exception type from an exception chain.

We will also need to make this behavior conditional upon whether pydantic is present in environment or not.

litestar/_signature/models/msgspec_signature_model.py
```diff
@@ -0,0 +1,160 @@
...
+        except ValidationError as e:
+            match = ERR_RE.search(str(e))
+            keys = [str(match.group(1)) if match else "n/a"]
```
peterschutt, line 72: Unlike the others, msgspec will only report one error: this is likely the biggest point of reckoning for the PR.

@jcrist is this something that can be explored on your side of the fence?

Using either pydantic or cattrs, we can report all validation errors, not just the first encountered. For example, if a client submits an invalid header value, and an invalid request body we can report both of those failures in the error response.

One possible way this could be handled on our side would be to have separate Structs to validate each of the data sources we have: body, query, cookies and headers (if DTOs are in use, we already do make a separate `convert()` call for body). This would allow us to report the _first_ error from each source, but would also require up to 4 separate calls to `convert()` for a request, which might be a hit to performance (although that would have to be bench marked before we lose sleep over it).

tests/unit/test_signature/test_parsing.py
```diff
@@ -145,44 +137,27 @@ def test(dep: int, param: int, optional_dep: Optional[int] = Dependency()) -> No
...
-    with create_test_client(
-        route_handlers=[test], dependencies=dependencies, _preferred_validation_backend=preferred_validation_backend
-    ) as client:
+    with create_test_client(route_handlers=[test], dependencies=dependencies) as client:
         response = client.get("/?param=thirteen")
 
     assert response.json() == {
         "detail": "Validation failed for GET http://testserver.local/?param=thirteen",
-        "extra": error_extra,
+        "extra": [{"key": "param", "message": "Expected `int`, got `str`", "source": "query"}],
```
peterschutt, line 155: If we parse out expected and actual, we can probably format our own error message, and maybe even include some structured data of the error, e.g., `{"expected": "integer", "received": "string"}`.

Related: https://github.com/jcrist/msgspec/issues/264

litestar/_signature/models/msgspec_signature_model.py
peterschutt: Actually, this will also affect our support for attrs (although we haven't had that in a proper release yet). The current attrs support comes from decoding with cattrs, which gives us all validation errors for a model as a validation group.

Take this example:

```py
import attrs
import cattrs
import msgspec
from exceptiongroup import ExceptionGroup

@attrs.define
class Foo:
    a: int
    b: str

try:
    cattrs.global_converter.structure({"data": {"one": 1, "b": 2}}, Foo)
except ExceptionGroup as e:
    print(str(e))
    print(e.exceptions)

try:
    msgspec.convert({"data": {"a": "one", "b": 2}}, type=Foo)
except Exception as e:
    print(str(e))
```

From cattrs we get an exception group and can access underlying exceptions for each sub-error:

> While structuring Foo (2 sub-exceptions)
(KeyError('a'), KeyError('b'))

From msgspec we only get the the one error:

> Object missing required field `a`

From my understanding, this cannot be solved with a dec hook either, as given msgspec's native support for attrs the dec hook would never be called.

jcrist: > @jcrist is this something that can be explored on your side of the fence?

No, this is not something I intend on supporting. For prior art, most typed JSON serialization libraries (across language ecosystems) don't support raising multiple errors. Neither golang's `json` nor rust's `serde` support this, and people generally love those tools.

When parsing JSON into a specified object type, there are multiple ways an error can occur:
- The JSON is invalid
- The JSON is valid, but the JSON type is incorrect (expected a `number`, got an `array`)
- The JSON type is correct, but the value is invalid (e.g. unknown Enum value, int out of range, invalid datetime str etc...)
- A JSON object is missing a field (or in certain situations has an extra field)

When parsing and validating, a failure in one location could be one error, or several. There's no clearly defined behavior for what should be expected. `pydantic` might group a few errors together and say "these are the things that are invalid". `cattrs` might find a different set of errors. But there's no clear definition of what "all the errors" means. Say for example you're parsing the value `"2022-03-"` into a type `datetime | date | None`. This string is not a valid datetime or date, and it isn't None. Is that one, two, or three errors?

Further, what if you receive a list of 1000 strings as an input for a field, but expected a list of ints. Is that 1000 errors (one for each item), or one for the list?

Maintaining a running structure of errors like this can get expensive, and make the failure mode in parsing significantly more expensive than the success mode. I'm skeptical of its general utility, and think if providing errors for multiple fields is important, you'd be better served by doing that validation yourself _after_ parsing, so you can ensure the kinds of user-facing errors you wish to provide are uniformly handled, rather than whatever errors the parsing framework happens to collect.

--- 

> One possible way this could be handled on our side would be to have separate Structs to validate each of the data sources we have: body, query, cookies and headers (if DTOs are in use, we already do make a separate convert() call for body).

I think this is a good idea, and is what I recommended y'all do several months ago: https://github.com/jcrist/msgspec/issues/301#issuecomment-1434842236.

Besides making it easier to group errors to include in a response, it makes it easier to make parsing more specific to the provided types. AFAICT right now you're combining all data sources (headers, body, ...) into a big dict, then passing that to an uberfunction to handle conversion into the types. This has a few downsides:

- It prevents optimizing the JSON decoding by parsing directly into the type using `msgspec.json.decode`, since you're decoding into a dict first then using `msgspec.convert` to convert. This has an unnecessary performance cost, and is missing out on the main reason I wrote msgspec in the first place.
- It prevents tailoring the decoding settings to the protocol at hand. When parsing data from untyped sources like headers/query parameters (which are all strings) you'll want very flexible coercion rules; you'll likely want to set `strict=False` (https://jcristharif.com/msgspec/usage.html#strict-vs-lax-mode). But when parsing something more structured like JSON those lax coercion rules can mask client-side issues, you likely want `strict=True` here. Pushing all parsing and validation into a single call prevents tailing the specifics of the parsing to the input data source.
- With dependency injection, some of the inputs to a handler likely don't need validation, since they're coming from within the litestar framework itself, not user data. Grouping these inputs into the same input dict and passing them to `msgspec.convert` is doing unnecessary work (though it may be pretty negligible in cost).

4 calls to `msgspec.convert` instead of a single call should have negligible performance impact, and if it lets you call a more specific parsing function for certain inputs (e.g. `msgspec.json.decode` for parsing json bodies) then this should have a net performance win.

> From my understanding, this cannot be solved with a dec hook either, as given msgspec's native support for attrs the dec hook would never be called.

This is correct, if you're using msgspec to parse a type containing an `attrs` type, msgspec will always handle this parsing itself rather than falling back to a `dec_hook`. The system was not designed to coexist with, but rather to supplant the need for `cattrs`.

If you separate out the input sources though like I recommended above, you could do something like:

- if the body type is an attrs object, use cattrs to parse it from json
- if the body type is a pydantic object, use pydantic to parse it from json
- otherwise use msgspec to parse it from json

tests/unit/test_signature/test_parsing.py
jcrist: If you have specific information you'd like to get in an error message, a comment in that issue adding additional context would be most useful.

litestar/_signature/models/msgspec_signature_model.py
peterschutt: Thanks for your detailed response @jcrist!

> No, this is not something I intend on supporting.

Understood, I can't argue with any of your rationale, and knowing this is off the table is just as helpful with moving the PR forward as it being on the table.

> I think this is a good idea, and is what I recommended y'all do several months ago

:grimacing: Sorry, I wasn't aware of that prior discussion, but good to know we are on the same page.

> 4 calls to msgspec.convert instead of a single call should have negligible performance impact

And, if we are happy enough to group the sources that we pre-decode together (query, headers, cookies), then this really only needs to be 2.

> If you separate out the input sources though like I recommended above, you could do something like:
>
>   - if the body type is an attrs object, use cattrs to parse it from json
>   - if the body type is a pydantic object, use pydantic to parse it from json
>   - otherwise use msgspec to parse it from json

I agree, this makes the most sense. Also then, we wouldn't need to worry about the exception chaining in order to access those pydantic exceptions, as I mentioned in another comment.

tests/unit/test_signature/test_parsing.py
peterschutt: done.

litestar/_signature/models/base.py
Goldziher: But this is not the place to do this. We should give access to this when `debug=true`

peterschutt: Oh, I see - I'll back this out then.

litestar/_signature/models/msgspec_signature_model.py
jcrist: > After https://github.com/jcrist/msgspec/issues/456 we will hopefully be able to access the pydantic exception type from an exception chain.

This has been resolved in https://github.com/jcrist/msgspec/pull/460. If you want to try it out before the next release, you can [install msgspec from GitHub](https://jcristharif.com/msgspec/install.html#installing-from-github).

peterschutt: Thank you @jcrist

Goldziher:
I rebased this on main and updated the implementation to align. I also squashed the previous commits together - so its easier to rebase. 

I am going to branch from this now.

Goldziher:
I removed the pydantic and attrs logic and refactored the code, also fixed most tests. The only remaining issue for now is the error handling.

Goldziher:
I merged the pydantic v2 support into this one. So it can be handled together.

provinzkraut:
This is not ready for review at all. There are several things that need to be handled before this can go in.

Goldziher:
> This is not ready for review at all. There are several things that need to be handled before this can go in.

such as?

provinzkraut:
The error handling will require a lot of substantial changes for example, and the support for Pydantic and attrs is only preliminary, since the intended solution (via plugins) hasn't been implemented yet.

I don't see why this has to be merged *now*, in an unfinished state, when we will then have to go back and alter is fundamentally again.

provinzkraut:
> we do not need to alter it

Yes we do, unless we don't care about breaking exceptions completely.

> It is fully functional now already

It is not. As I said before, the handling of the exception will require substantial changes, and those are *not ready at the moment*.
We can argue about this all we want, but that doesn't change the facts.

Goldziher:
> > we do not need to alter it
> 
> Yes we do, unless we don't care about breaking exceptions completely.
> 
> > It is fully functional now already
> 
> It is not. As I said before, the handling of the exception will require substantial changes, and those are _not ready at the moment_. We can argue about this all we want, but that doesn't change the facts.

the only thing that requires changes is the exception handling. But that is a given.

PR #1863, opened 2023-06-23
fix: added `session.refresh` to sqla repository update

patrickarmengol:
I haven't opened an issue for this, but ran into it when updating a model attribute that changes when sent to the db. In my current case, it is a Geography object that uses a WKT formatted string when sent to the db, and comes back as a WKBElement wrapped by [GeoAlchemy2](https://geoalchemy-2.readthedocs.io/en/latest/orm_tutorial.html#add-new-objects) when queried. 

By refreshing the instance after merging and flushing, we get the newest values of the updated attributes instead of keeping values sent to the db.

### DIFF
--- a/litestar/contrib/sqlalchemy/repository/_async.py
+++ b/litestar/contrib/sqlalchemy/repository/_async.py
@@ -286,6 +286,7 @@ class SQLAlchemyAsyncRepository(AbstractAsyncRepository[ModelT], Generic[ModelT]
             # this will merge the inbound data to the instance we just put in the session
             instance = await self._attach_to_session(data, strategy="merge")
             await self.session.flush()
+            await self.session.refresh(instance)
             self.session.expunge(instance)
             return instance
 
--- a/litestar/contrib/sqlalchemy/repository/_sync.py
+++ b/litestar/contrib/sqlalchemy/repository/_sync.py
@@ -288,6 +288,7 @@ class SQLAlchemySyncRepository(AbstractSyncRepository[ModelT], Generic[ModelT]):
             # this will merge the inbound data to the instance we just put in the session
             instance = self._attach_to_session(data, strategy="merge")
             self.session.flush()
+            self.session.refresh(instance)
             self.session.expunge(instance)
             return instance
 


cofin:
The summary/fix makes sense, and I believe it was an oversight.  Good catch!

Once the validations are passing, we can merge.

peterschutt:
> and I believe it was an oversight

@cofin - it looks like the `refresh()` calls were removed for both `update()` and `upsert()` in https://github.com/litestar-org/litestar/commit/aa626eb16e239d43ae797501b6867eaeba73fb4d - DYT just oversight, or might there have been a particular reason given they were both removed at the same time?

cofin:
@peterschutt  I was actually going to do a separate test with this branch before merging it.  I vaguely remember having some strangeness around this, but I'm thinking it may be an oversight that they weren't added back in.  I can't think of a reason why it shouldn't be there.

cofin:
@patrickarmengol do you mind adding it there as well?

patrickarmengol:
> @patrickarmengol do you mind adding it there as well?

on it

peterschutt:
> > @patrickarmengol do you mind adding it there as well?
> 
> on it

Thanks!

Integration tests for the refresh behavior would be nice too, if you have the bandwidth @patrickarmengol? You can check [the other repository tests](https://github.com/litestar-org/litestar/tree/main/tests/unit/test_contrib/test_sqlalchemy/test_repository) for patterns.

cofin:
I've tested the `update` method on this branch against a separate code base and it passes as well.  I've got a large chunk of code that uses the `upsert` as well, that I'll test once it's pushed.  If both of those pass, I'm pretty comfortable with these changes.

cofin:
`upsert` works as well.  I think this is good to go!

peterschutt:
> I've tested the `update` method on this branch against a separate code base and it passes as well. I've got a large chunk of code that uses the `upsert` as well, that I'll test once it's pushed. If both of those pass, I'm pretty comfortable with these changes.

 We should have a test for the behavior in the litestar suite to protect against future regressions.

patrickarmengol:
> > > @patrickarmengol do you mind adding it there as well?
> > 
> > 
> > on it
> 
> Thanks!
> 
> Integration tests for the refresh behavior would be nice too, if you have the bandwidth @patrickarmengol? You can check [the other repository tests](https://github.com/litestar-org/litestar/tree/main/tests/unit/test_contrib/test_sqlalchemy/test_repository) for patterns.

This is a good idea, but sorry, I won't be able to work on it for a few days. I can take a crack at it if nobody else has done it by next week. 

Trying to think of a more general case where the refresh is relevant....

peterschutt:
> Trying to think of a more general case where the refresh is relevant....

Server generated timestamps was my original motivation for adding them

tests/unit/test_contrib/test_sqlalchemy/test_repository/test_repo_refresh.py
```diff
@@ -0,0 +1,79 @@
...
+import pytest
+from sqlalchemy import text
+from sqlalchemy.dialects.mysql import TIMESTAMP
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, async_sessionmaker
+from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column
+from sqlalchemy.schema import FetchedValue
+
+from litestar.contrib.sqlalchemy.repository import SQLAlchemyAsyncRepository
+
+pytestmark = [pytest.mark.sqlalchemy_integration]
```
peterschutt, line 14: should this be marked with anything else?

cofin:
I think i have a case where the refresh may/may not be relevant.  I'll type out my thoughts in the discord channel though.

cofin:
I think there's one additional place we need to consider adding the refresh:

https://github.com/litestar-org/litestar/blob/main/litestar/contrib/sqlalchemy/repository/_async.py#L239

tests/unit/test_contrib/test_sqlalchemy/test_repository/test_repo_refresh.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@pytest.fixture(name="session")
+async def fixture_session(mysql_engine: AsyncEngine) -> AsyncGenerator[AsyncSession, None]:
+    session = async_sessionmaker(bind=mysql_engine, expire_on_commit=False)()
+    try:
+        yield session
+    finally:
+        await session.rollback()
+        await session.close()
```
cofin, line 47: I think we should run these tests against all of the backends intead of just mysql.

peterschutt: I have no problems if you want to do that - I don't know enough about all the other backends to do it myself.

My goal is to have an integration test that guards against future regressions of this on the repository layer so that anyone who tries to remove those `refresh()` calls in the future can see a real world example of why it is there.

cofin: I've made updates to the test cases and put together a model that uses fetchedvalues on all backends.

cofin:
I think this is ready for a final review.  Thanks @patrickarmengol for the patience and help on this one.

cofin (approved):
LGTM

peterschutt (approved):

PR #1865, opened 2023-06-23
chore: fix pydantic v1 regex validation

Goldziher:
This PR fixes a small regression in the pydantic signature model logic, which was caused by the renaming of `regex` -> `pattern` (pattern is used by pydantic v2 and everyone else).


cofin (approved):
LGTM

provinzkraut (approved):

PR #1869, opened 2023-06-24
feat: add `on_cli_startup` and `on_cli_shutdown` hooks.

### LINKED ISSUE #1835: Enhancement: add  CLI `on_cli_startup` and `on_cli_shutdown` hooks.
### Summary

It would be nice to be able to start processes outside of the ASGI lifespan.  This would allow us to run-processes such as Vite or SAQ/ARQ without having to hooking into the `on_startup` / `on_shutdown` hooks of the ASGI worker.  The benefit here is that you can launch a single process instead of a process for each uvicorn worker.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

<a href="https://polar.sh/litestar-org/litestar/issues/1835">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1835/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
Fixes #1835

### Pull Request Checklist

### Description

### Close Issue(s)


litestar/config/app.py
```diff
@@ -61,6 +61,8 @@ class AppConfig:
     The final attribute values are used to instantiate the application object.
     """
 
+    after_shutdown: list[Callable] = field(default_factory=list)
```
JacobCoffee, line 64: These wouldnt be `list[LifespanHook]`?

JacobCoffee (commented):
I think it looks good so far?

litestar/config/app.py
cofin: I originally had them as that.  It's not really part of the ASGI lifespan and I don't think these can be async functions, but it would be nice to leverage that type if we can.

peterschutt (commented):
I notice that the linked issue is for `on_cli_startup` and `on_cli_shutdown` - I agree with that. By nature, they rely on being called from outside the application instance, so should they really be a property of it? If I was to start an app through some means other than the cli, then I think it would be confusing that the hooks defined as arguments to the app itself don't get executed.

I think the problem is rooted in the fact that the Litestar application is tightly coupled to the ASGI application concept. Therefore, it doesn't make sense (at least, to me) to attempt to have the Litestar app manage anything at a higher level than that.

In the past, I've used [honcho](https://honcho.readthedocs.io/en/latest/) for this type of thing, but perhaps a plugin could have its own CLI that wraps ours, or use cli plugins to add this sort of functionality.

litestar/cli/commands/core.py
```diff
@@ -155,7 +173,7 @@ def run_command(
             process_args["fd"] = fd
         if uds is not None:
             process_args["uds"] = uds
-        if reload_dirs:
+        if reload:
```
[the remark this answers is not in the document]
cofin: No, I had been testing a few things out before actually changing things.  I'll be sure to revert this.

Goldziher:
We removed basically these two hooks previously 😁

cofin:
> I think the problem is rooted in the fact that the Litestar application is tightly coupled to the ASGI application concept. Therefore, it doesn't make sense (at least, to me) to attempt to have the Litestar app manage anything at a higher level than that.

This is where I view it slightly differently.  I think that once you move from calling `uvicorn` directly to using the CLI, it's reasonable to expect additional hooks.  Presumably, a user would be opt for the click CLI out of convenience or to extend application functionality.  In both of these cases, it seems like the two hooks would be useful? 

If there's an alternate method for me to execute things before/after startup/shutdown?  I'd be happy to use that instead. 
However, I don't think `honcho` is going to be a good option for me.  I'd like something not driven by a `Procfile`.  To be fair, it's been a while since I last used honcho, so I'll give it a fresh look.

peterschutt:
> I think that once you move from calling uvicorn directly to using the CLI, it's reasonable to expect additional hooks.

I'm not arguing this, but IMO they should be owned by the CLI, not the application.

> Presumably, a user would be opt for the click CLI out of convenience or to extend application functionality.

I don't think we can make presumptions like this unless we make the cli extras non-optional (although I wouldn't support that change either). For instance, I don't include the cli in production deployments to keep dependencies/artifacts size as small as possible.

cofin:
So, I took a refresh look at `honcho`.  It looks like I can call the process manager programatically.  I think this probably will work for my use case after all.

WIth that said, is there still value in leaving this as a lightweight hook if users want it?

provinzkraut:
> WIth that said, is there still value in leaving this as a lightweight hook if users want it?

As I said before, I think we generally want this. But maybe this isn't the best way to implement it? I agree with @peterschutt on that this shouldn't be part of the application config, since it can easily lead to confusion, and will just be ignored if the CLI isn't used.

One way to handle things like this is to add support for a configuration file (e.g. `litestar.run.toml` / `[tool.litestar.run]` inside `pyproject.toml`). Hooks could then be defined in arbitrary locations, and listed via that file, a bit like entrypoints. This would also allow to easily specify "CLI extensions", which is another thing we eventually want to support.

peterschutt:
> One way to handle things like this is to add support for a configuration file (e.g. `litestar.run.toml` / `[tool.litestar.run]` inside `pyproject.toml`). Hooks could then be defined in arbitrary locations, and listed via that file, a bit like entrypoints. This would also allow to easily specify "CLI extensions", which is another thing we eventually want to support.

Interesting idea!

cofin:
Why don't I close this PR and we can re-open once we have a better plan for implementation?

PR #1871, opened 2023-06-24
chore: clean sonar lint issues

Goldziher:
This PR cleans sonar lint issues


docs/conf.py
```diff
@@ -63,70 +65,71 @@
...
-    ("py:class", "_orm.registry"),
-    ("py:class", "_schema.MetaData"),
-    ("py:class", "_schema.Table"),
-    ("py:class", "_RegistryType"),
-    ("py:class", "abc.Collection"),
-    ("py:class", "TypeEngine"),
-    ("py:class", "ExternalType"),
-    ("py:class", "UserDefinedType"),
-    ("py:class", "_types.TypeDecorator"),
+    (PY_CLASS, "BaseModel"),
```
peterschutt, line 68: I understand why, but I don't think this really improves anything, as these directives aren't ever going to change.

If we are to go through with it,  `py:meth` and `py:obj` should be treated the same for sake of consistency.

peterschutt (approved):

docs/conf.py
Goldziher: its about repetation of values - sonar doesnt like copy pasting.

peterschutt: Yes, as I said, I understand why, but I don't think it adds anything. Not a blocker though, although I do think the consistency point is pertinent.

Goldziher: addressed

PR #1886, opened 2023-06-27
Allow customization of OpenAPIController path directly from OpenAPIConfig

### LINKED ISSUE #1486: Enhancement: Expose `path` from the `OpenAPIController` to the `OpenAPIConfig` object
### Summary

The [OpenAPIController](https://docs.litestar.dev/2/usage/openapi.html#the-openapicontroller) has a [`path`](https://docs.litestar.dev/2/reference/openapi/openapi.html#litestar.openapi.OpenAPIController.path) param that would be very nice to expose to the [OpenAPIConfig](https://docs.litestar.dev/2/reference/openapi/openapi.html#litestar.openapi.OpenAPIConfig)

Currently you can pass in a custom OpenAPICOntroller object into the OpenAPIConfig
```py
class MyOpenAPIController(OpenAPIController):
  path = "/api"

config = OpenAPIConfig(controller=MyOpenAPIController)
app = Litestar(openapi_config=config)
```

but having some abstraction would make this cleaner and more user-friendly

### Relevant Documentation
https://docs.litestar.dev/2/usage/openapi.html#the-openapicontroller
https://docs.litestar.dev/2/reference/openapi/openapi.html#litestar.openapi.OpenAPIController
https://docs.litestar.dev/2/reference/openapi/openapi.html#litestar.openapi.OpenAPIController.path
https://docs.litestar.dev/2/reference/openapi/openapi.html#litestar.openapi.OpenAPIConfig

### Basic Example

```py
config = OpenAPIConfig(
    title=settings.openapi.TITLE or settings.app.NAME,
    version=settings.openapi.VERSION,
    contact=Contact(name=settings.openapi.CONTACT_NAME, email=settings.openapi.CONTACT_EMAIL),
    use_handler_docstrings=True,
    root_schema_site="swagger",
    path="/api",  # <--- Add path param here to allow customization
)
```

### Drawbacks and Impact

If this is not done it will lead to user despair, many questions, and general unhappiness in the world

Current implementations can be done by overriding the `OpenAPIController`, but it must be subclassed.

```py
from litestar.openapi.controller import OpenAPIController

class OverridenController(OpenAPIController):
    path: str = settings.openapi.PATH


config = OpenAPIConfig(
    openapi_controller=OverridenController,
)
```

### Unresolved questions

_No response_

<a href="https://polar.sh/litestar-org/litestar/issues/1486">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1486/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

jaykv:
### Pull Request Checklist

### Description
* Exposes `OpenAPIController` path directly in `OpenAPIConfig`
* If both `OpenAPIController` and `path` params are set in `OpenAPIConfig`, it will use the path from `OpenAPIController`

### Close Issue(s)
Closes #1486 


litestar/openapi/config.py
```diff
@@ -41,7 +41,7 @@ class OpenAPIConfig:
 
     create_examples: bool = field(default=False)
     """Generate examples using the pydantic-factories library."""
-    openapi_controller: type[OpenAPIController] = field(default_factory=lambda: OpenAPIController)
+    openapi_controller: type[OpenAPIController] | None = field(default=None)
```
Goldziher, line 44: What's the reason for this change?

jaykv: Just so we can detect in `__post_init__` if openapi_controller is explicitly set or not and avoid the path patch if it is

JacobCoffee:
Hey @jaykv! Thanks for picking up this issue.

JacobCoffee:
Also i think you will need to run `poetry lock --no-update` or similar to fix the ci

jaykv:
Okay! I think I got everything now. Ended up reverting the poetry.lock change altogether to keep the pipeline happy

litestar/openapi/config.py
Goldziher: I don't think this is the right way to go. 

Thoughts @litestar-org/maintainers

peterschutt: I don't think so either.

With the controller typed as optional, everywhere that we reference the controller now needs to narrow away the `None` type union, e.g., `if self.openapi_config and self.openapi_config.openapi_controller:`. However, the controller isn't actually ever `None`, because it is always defined in `__post_init__`. So, yeh, doesn't feel quite right to me.

This is a convenience to save subclassing the openapi controller to define the path, right?

jaykv: Yeah that makes sense. I missed an easier check in `__post_init__`  with `if self.openapi_controller is OpenAPIController` so this change is gone now

litestar/openapi/config.py
```diff
@@ -94,6 +94,13 @@ class OpenAPIConfig:
     """A set of the enabled documentation sites and schema download endpoints."""
     operation_id_creator: OperationIDCreator = default_operation_id_creator
     """A callable that generates unique operation ids"""
+    path: str = "/schema"
+    """Base path for the OpenAPI documentation endpoints."""
+
+    def __post_init__(self) -> None:
+        if self.openapi_controller is OpenAPIController:
```
Goldziher, line 101: So what's the point here?

jaykv: It's to make sure we respect the custom OpenAPIController's path and only override the path for the default OpenAPIController, so:
```python
class MyOpenAPIController:
    path = "/customdocs"

app = Litestar([], openapi_config=OpenAPIConfig(title="Litestar API", version="1.0.0"), openapi_controller=MyOpenAPIController)
```
would still serve docs at `/customdocs` instead of `/schema`.

Does it feel hacky?

Goldziher: Why not simply check if the controller class has path defined on it? Use getattr etc

peterschutt: Is this something we should worry about? I believe a path customization would be overwritten in this case - it is a bit contrived though.

```py
from litestar.openapi import OpenAPIConfig, OpenAPIController

OpenAPIController.path = "/somewhere-else"

app = Litestar(..., openapi_config=OpenAPIConfig(openapi_controller=OpenAPIController))
```
What about making `OpenAPIConfig.path` `str | None = None` and only subclass and overwrite the path if the config path is not `None`? This would mean that a config defined path would always take precedence over a controller defined path.

Goldziher: Fair enough, but in this case we need validation

jaykv: For validation, aside from enforcing the path starts with `/`, is there anything else?

Goldziher: We have to validate the path is actually set. 

Furthermore, it cannot be `/` only but must be a proper sub path.

jaykv: thanks! I'll get it in sometime this weekend

Goldziher (approved):

Goldziher:
@all-contributors add @jaykv code

JacobCoffee:
This needs docs updates in /usage/openapi.rst

PR #1900, opened 2023-07-01
feat: Make repository `refresh` and `expunge` optional, Adds `auto_commit` option.

### LINKED ISSUE #1895: Question: problems with sqlalchemy repository update and lazy loading
I'm having trouble reading a field representing a many-to-many relationship after using the [repo `update` method](https://github.com/litestar-org/litestar/blob/main/litestar/contrib/sqlalchemy/repository/_async.py#L270). This is what I found out so far. 

My sqla repository has `statement=select(Item).options(selectinload(Item.tags))` in order to include the tags in the response, which works great for the `get` and `list` methods. 

When setting for the `lazy` arg for the `tags` relationship defined for the `Item` model, AKA configuring the loader strategy at [mapping time](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#configuring-loader-strategies-at-mapping-time) :

- If I set the `lazy` arg to `"noload"`, the one that I desire, the response ends up with an empty tags list in the case I'm updating a field other than `tags`. I only get a valid response when the `tags` field itself is being updated.  

- If I set the `lazy` arg to `"selectin"`, then it works as expected for `update`, but then I also get tags for regular `select(Item)` statement, which is undesired. 

- If I leave the `lazy` arg empty/default/`"select"`, then I get:

```
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Item at 0x7fb31aaea590> is not bound to a Session; lazy load operation of attribute 'tags' cannot proceed (Background on this error at: https://sqla
lche.me/e/20/bhk3)
```

I read through the above [error help](https://docs.sqlalchemy.org/en/20/errors.html#error-bhk3). I already have `expire_on_commit=False` and my sqla repository has `statement=select(Item).options(selectinload(Item.tags))`, but for some reason `update` doesn't care about the statement options even though it calls `get` internally, and, from what I can tell, instead uses the loader strategies configured at mapping time. 

Here's some of my code:

```python
item_tag = Table(
    "item_tag",
    UUIDBase.metadata,
    Column("item_id", ForeignKey("item.id"), primary_key=True),
    Column("tag_id", ForeignKey("tag.id"), primary_key=True),
)


class Item(UUIDBase):
    name: Mapped[str] = mapped_column(String(), unique=True)
    description: Mapped[str | None]
    tags: Mapped[list[Tag]] = relationship(
        secondary=item_tag,
        back_populates="items"
        lazy="noload"  # <-- here be problems
    )


class Tag(UUIDBase):
    name: Mapped[str] = mapped_column(String(), unique=True,
    items: Mapped[list[Item]] = relationship(
        secondary=item_tag,
        back_populates="tags",
        lazy="noload"
    )


@patch(
    path="/{item_id:uuid}",
)
async def update_item(
    item_repo: ItemRepository,
    tag_repo: TagRepository,
    data: ItemUpdate,
    item_id: UUID = Parameter(
        title="Item ID",
        description="The item to update",
    ),
) -> ItemDB:
    dd = data.dict(exclude_unset=True)
    dd.update({"id": item_id})
    if "tag_names" in dd:
        if len(dd["tag_names"]) != 0:
            dd["tags"] = await tag_repo.list(
                CollectionFilter("name", dd["tag_names"])
            )
        else:
            dd["tags"] = []
        del dd["tag_names"]
    obj = await item_repo.update(Item(**dd))
    await item_repo.session.commit()
    return parse_obj_as(ItemDB, obj)  # <-- here be errors

```

Hopefully that was clear and I provided enough context. I'm a bit lost on how to solve this and wondering if it may indicate an issue with the method's implementation itself. 

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below

<a href="https://polar.sh/litestar-org/litestar/issues/1895">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1895/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
Fixes #1895

This PR does the following:

- exposes the `attribute_names` and `with_for_update` parameters on functions that execute `session.refresh`
- Adds `auto_commit` parameter.  When this is true, the session will `commit` instead of `flush` before returning (default=False).
- Adds `auto_refresh` parameters.  When this is true, the session will execute `refresh` objects  before returning (default=True).
- Adds `auto_expunge` parameters.  When this is true, the session will execute `expunge` on all objects before returning (default=True).(default=True).

### Pull Request Checklist

### Description

### Close Issue(s)

### DIFF
--- a/litestar/contrib/sqlalchemy/repository/_async.py
+++ b/litestar/contrib/sqlalchemy/repository/_async.py
@@ -78,14 +78,14 @@ class SQLAlchemyAsyncRepository(AbstractAsyncRepository[ModelT], Generic[ModelT]
         Returns:
             The added instance.
         """
-        _auto_commit = kwargs.pop("auto_commit", self.auto_commit)
-        _auto_expunge = kwargs.pop("auto_expunge", self.auto_expunge)
-        _auto_refresh = kwargs.pop("auto_refresh", self.auto_refresh)
+        auto_commit = kwargs.pop("auto_commit", self.auto_commit)
+        auto_expunge = kwargs.pop("auto_expunge", self.auto_expunge)
+        auto_refresh = kwargs.pop("auto_refresh", self.auto_refresh)
         with wrap_sqlalchemy_exception():
             instance = await self._attach_to_session(data)
-            await self._flush_or_commit(auto_commit=_auto_commit)
-            await self._refresh(instance, auto_refresh=_auto_refresh)
-            self._expunge(instance, auto_expunge=_auto_expunge)
+            await self._flush_or_commit(auto_commit=auto_commit)
+            await self._refresh(instance, auto_refresh=auto_refresh)
+            self._expunge(instance, auto_expunge=auto_expunge)
             return instance
 
     async def add_many(
@@ -102,13 +102,13 @@ class SQLAlchemyAsyncRepository(AbstractAsyncRepository[ModelT], Generic[ModelT]
         Returns:
             The added instances.
         """
-        _auto_commit = kwargs.pop("auto_commit", self.auto_commit)
-        _auto_expunge = kwargs.pop("auto_expunge", self.auto_expunge)
+        auto_commit = kwargs.pop("auto_commit", self.auto_commit)
+        auto_expunge = kwargs.pop("auto_expunge", self.auto_expunge)
         with wrap_sqlalchemy_exception():
             self.session.add_all(data)
-            await self._flush_or_commit(auto_commit=_auto_commit)
+            await self._flush_or_commit(auto_commit=auto_commit)
             for datum in data:
-                self._expunge(datum, auto_expunge=_auto_expunge)
+                self._expunge(datum, auto_expunge=auto_expunge)
             return data
 
     async def delete(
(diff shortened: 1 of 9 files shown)


cofin:
Should we be able to override this on a method by method basis, similar to `select`?

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -31,16 +31,31 @@ class SQLAlchemyAsyncRepository(AbstractAsyncRepository[ModelT], Generic[ModelT]
 
     match_fields: list[str] | str | None = None
 
-    def __init__(self, *, statement: Select[tuple[ModelT]] | None = None, session: AsyncSession, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        *,
+        statement: Select[tuple[ModelT]] | None = None,
+        session: AsyncSession,
+        auto_expunge: bool = True,
```
peterschutt, line 39: Repo level config for this stuff is the right move IMO.

My only suggestion would be to consider whether the expunge call should be default, or opt-in. Once litestar is out of beta this won't be something that can be easily reversed.

Sorry I can't do a more thorough review of the PR.

cofin: In that case, perhaps we should disable `expunge` by default.  It can be a bit confusing when you hit edge cases like this.

cofin: After a bit more thought, I think this is the right choice.  If you want the additional protection that it offers, it's there to use, otherwise, you can expect it to work in all cases like this.

Goldziher (approved):
LGTM, but im not an expert....

provinzkraut:
@cofin Can you add docs for this somewhere?

PR #1911, opened 2023-07-02
feat(docs): add `sphinx-click` and adjust shorthand CLI options 

### LINKED ISSUE #1495: Docs: Add `sphinx-click` to the documentation.
### Summary

Sphinx has a contrib plugin for click. It would be nice to auto-doc it.
I have used it in my own app recently - example:
image

https://sphinx-click.readthedocs.io/en/latest/ 

1. `poetry add -G docs sphinx-click`
2. Add `sphinx-click` to the `conf.py`
3. Use it to document our CLI automatically

### Reference:
https://discord.com/channels/919193495116337154/1061269663750836234/1095917360591949904

<a href="https://polar.sh/litestar-org/litestar/issues/1495">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1495/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

JacobCoffee:
### Pull Request Checklist

### Description

This PR:

-  adds [`sphinx-click`](https://github.com/click-contrib/sphinx-click) and the relevant usage and reference docs.
- adjusts the click command options to use proper `-h` for help, and other shorthand commands

### Close Issue(s)

Closes #1495


provinzkraut (changes requested):
Don’t have time for a proper review at the moment, will take a look tomorrow. I just wanted to note that some of the changes to the CLI flags are breaking. These should be either adjusted to be non-breaking or reverted.

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
 )
 @option("--host", help="Server under this host", default="127.0.0.1", show_default=True)
 @option(
+    "-fd",
     "--fd",
```
cofin, line 75: Is the intent to remove all of the `--[shorthand]` options in favor of `-[shorthand]`?  I think we should do this before we leave beta if so.

tests/unit/test_cli/test_core_commands.py
```diff
@@ -244,7 +244,7 @@ def test_run_command_pdb(
     app_path = f"{path.stem}:create_app"
 
     with pytest.warns(LitestarWarning):
-        result = runner.invoke(cli_command, ["--app", app_path, "run", "--pdb"])
+        result = runner.invoke(cli_command, ["--app", app_path, "run", "--use_pdb"])
```
cofin, line 247:
```suggestion
        result = runner.invoke(cli_command, ["--app", app_path, "run", "--use-pdb"])
```
We use `-` in most of the other places.

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
...
     default=None,
     show_default=True,
 )
-@option("--uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
-@option("--debug", help="Run app in debug mode", is_flag=True)
-@option("--pdb", "use_pdb", help="Drop into PDB on an exception", is_flag=True)
-@option("--reload-dir", help="Directories to watch for file changes", multiple=True)
+@option("-uds", "--uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
+@option("-d", "--debug", help="Run app in debug mode", is_flag=True)
+@option("-pdb", "--use_pdb", help="Drop into PDB on an exception", is_flag=True)
```
cofin, line 84:
```suggestion
@option("-pdb", "--use-pdb", help="Drop into PDB on an exception", is_flag=True)
```
Maybe consider doing this since we use `-` for the others?

litestar/cli/commands/core.py
JacobCoffee: Most of this is because we have the long form commands (`--[command]`) but the shorthand versions are non-existent.

Click displays the shorthand appropriately when these are set, but without them we don't have that.

image

JacobCoffee: > Is the intent to remove all of the --[shorthand] options in favor of -[shorthand]? 

So the intent here is to properly set the shorthand commands in addition to their current long form versions.

current:
image

proposed:
(pictured in last comment)

JacobCoffee: Also, wrt `--uds` and `--fd` in particular, these seem like they shouldve been `-uds` and `-fd` from the start?
At least that is the convention (that i am used to).. it is pretty odd (to me at least) to have multiple long form arguments instead of one short/one long.

litestar/cli/commands/core.py
provinzkraut: The `use_pdb` was using an underscore because it wasn’t specifying a flag, but the argument name for click. As a flag, this should use a dash.

litestar/cli/commands/core.py
provinzkraut: Actually, I don't think so? This can get confusing IMO when combining multiple shorthands, e.g, `-uds` + `-R` would be `-udsR`, which looks like it's 4 flags, when it's actually just 2.

The proper way to have these is to reserve the `-` option for single letter options, and use `--` for everything else.

Also, the already short option names don't really *need* a shorthand. Wdyt?

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
...
     help="Bind to a socket from this file descriptor.",
     type=int,
     default=None,
     show_default=True,
 )
-@option("--uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
-@option("--debug", help="Run app in debug mode", is_flag=True)
-@option("--pdb", "use_pdb", help="Drop into PDB on an exception", is_flag=True)
-@option("--reload-dir", help="Directories to watch for file changes", multiple=True)
+@option("-uds", "--uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
```
provinzkraut, line 82: Same comment as above; Three options is one too many

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
...
+@option("-d", "--debug", help="Run app in debug mode", is_flag=True)
+@option("-pdb", "--use-pdb", help="Drop into PDB on an exception", is_flag=True)
```
provinzkraut, line 84: Just `--pdb` is fine here. It doesn't need a shorthand. *If* we were to add a shorthand though, it should be something like `-P`, which would be the canonical way of doing these things.

This is also a breaking change, which is something we should try to avoid during the beta.

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
...
+@option("-rd", "--reload-dir", help="Directories to watch for file changes", multiple=True)
```
provinzkraut, line 85:
```suggestion
@option("-R", "--reload-dir", help="Directories to watch for file changes", multiple=True)
```

litestar/cli/commands/core.py
```diff
@@ -71,24 +71,25 @@ def info_command(app: Litestar) -> None:
 )
 @option("--host", help="Server under this host", default="127.0.0.1", show_default=True)
 @option(
+    "-fd",
```
provinzkraut, line 74: Three options are a bit excessive :grimacing:

litestar/cli/commands/core.py
cofin, line 82:
```suggestion
@option("-uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
```

litestar/cli/commands/core.py
cofin: I'm ok with either way we choose to go on this.  

The user who chooses to use this particular syntax `-udsR` would likely know they are using 2 flags instead of 4, so I think it's probably fine to overcomplicate it.

litestar/cli/commands/core.py
cofin: I'm not sure I follow why there's a distinction between argument vs flag as user.  Wouldn't we want the CLI flags to be consistent or is there an expected naming convention that I've seemingly forgotten about (which is very possible)?

JacobCoffee:
I will finish this up by tomorrow - sorry I have been out of town

JacobCoffee:
> I will finish this up by tomorrow - sorry I have been out of town

5 days later 😅  this is ready for review..

docs/usage/cli.rst
```diff
@@ -116,26 +116,26 @@ Options
...
+| ``-d``\ , ``--debug``                 | ``LITESTAR_DEBUG``              | Run the application in debug mode                              |
 +---------------------------------------+---------------------------------+----------------------------------------------------------------+
-| ``--pdb``                             | ``LITESTAR_PDB``                | Drop into the Python debugger when an exception occurs         |
+| ``-P``\ , ``--use_pdb``             | ``LITESTAR_PDB``                | Drop into the Python debugger when an exception occurs         |
 +---------------------------------------+---------------------------------+----------------------------------------------------------------+
 
 
-``--reload-dir``
-++++++++++++++++
+``-R`` | ``--reload-dir``
```
provinzkraut, line 137: I would leave this as is. We don't include the shorthand for the other options either.

litestar/cli/commands/core.py
```diff
@@ -60,35 +60,35 @@ def info_command(app: Litestar) -> None:
...
     default=None,
     show_default=True,
 )
-@option("--uds", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
-@option("--debug", help="Run app in debug mode", is_flag=True)
-@option("--pdb", "use_pdb", help="Drop into PDB on an exception", is_flag=True)
-@option("--reload-dir", help="Directories to watch for file changes", multiple=True)
+@option("-U", "--unix-domain-socket", help="Bind to a UNIX domain socket.", default=None, show_default=True)
+@option("-d", "--debug", help="Run app in debug mode", is_flag=True)
+@option("-P", "--use-pdb", help="Drop into PDB on an exception", is_flag=True)
```
provinzkraut, line 84: This one is still breaking compatibility, as shown in the test case you had to adjust. I think just `--pdb` is absolutely fine here.

litestar/cli/commands/core.py
```diff
@@ -60,35 +60,35 @@ def info_command(app: Litestar) -> None:
...
     help="The number of HTTP workers to launch",
     type=click.IntRange(min=1, max=multiprocessing.cpu_count() + 1),
     show_default=True,
     default=1,
 )
-@option("--host", help="Server under this host", default="127.0.0.1", show_default=True)
+@option("-H", "--host", help="Server under this host", default="127.0.0.1", show_default=True)
 @option(
-    "--fd",
+    "-F",
```
provinzkraut, line 75: This is breaking backwards compatibility still. I'd leave it as `--fd`.

litestar/cli/commands/core.py
```diff
@@ -60,35 +60,35 @@ def info_command(app: Litestar) -> None:
 
 @command(name="run")  # type: ignore
 @option("-r", "--reload", help="Reload server on changes", default=False, is_flag=True)
+@option("-R", "--reload-dir", help="Directories to watch for file changes", multiple=True)
 @option("-p", "--port", help="Serve under this port", type=int, default=8000, show_default=True)
 @option(
-    "-wc",
+    "-W",
```
provinzkraut, line 66: This is breaking backwards compatibility.

provinzkraut (changes requested):

JacobCoffee:
Even with 3 switches, it looks and feels a lot better to me as a user than having less options.. but i reverted.. 😞 
```console
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
│ --reload              -r                            Reload server on changes                     │
│ --reload-dir              TEXT                      Directories to watch for file changes (TEXT) │
│ --port                -p  INTEGER                   Serve under this port (INTEGER)              │
│                                                     [default: 8000]                              │
│ --web-concurrency     -W  INTEGER RANGE [1<=x<=25]  The number of HTTP workers to launch         │
│                                                     (INTEGER RANGE)                              │
│                                                     [default: 1; 1<=x<=25]                       │
│ --host                -H  TEXT                      Server under this host (TEXT)                │
│                                                     [default: 127.0.0.1]                         │
│ --file-descriptor     -F  INTEGER                   Bind to a socket from this file descriptor.  │
│                                                     (INTEGER)                                    │
│ --unix-domain-socket  -U  TEXT                      Bind to a UNIX domain socket. (TEXT)         │
│ --debug               -d                            Run app in debug mode                        │
│ --use-pdb             -P                            Drop into PDB on an exception                │
│ --help                -h                            Show this message and exit.                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯


litestar on  cli_updates פֿ 1 🤷 3 is  v2.0.0beta2 via  v3.11.3 (litestar-py3.11) 🐠
➜ litestar run -h

 Usage: litestar run [OPTIONS]

 Run a Litestar app.
 The app can be either passed as a module path in the form of <module name>.<submodule>:<app
 instance or factory>, set as an environment variable LITESTAR_APP with the same format or
 automatically discovered from one of these canonical paths: app.py, asgi.py, application.py or
 app/__init__.py. When auto-discovering application factories, functions with the name
 ``create_app`` are considered, or functions that are annotated as returning a ``Litestar``
 instance.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
│ --reload                    -r                            Reload server on changes               │
│ --reload-dir                -R  TEXT                      Directories to watch for file changes  │
│                                                           (TEXT)                                 │
│ --port                      -p  INTEGER                   Serve under this port (INTEGER)        │
│                                                           [default: 8000]                        │
│ --wc,--web-concurrency          INTEGER RANGE [1<=x<=25]  The number of HTTP workers to launch   │
│                                                           (INTEGER RANGE)                        │
│                                                           [default: 1; 1<=x<=25]                 │
│ --host                      -H  TEXT                      Server under this host (TEXT)          │
│                                                           [default: 127.0.0.1]                   │
│ --fd,--file-descriptor          INTEGER                   Bind to a socket from this file        │
│                                                           descriptor.                            │
│                                                           (INTEGER)                              │
│ --uds,--unix-domain-socket      TEXT                      Bind to a UNIX domain socket. (TEXT)   │
│ --debug                     -d                            Run app in debug mode                  │
│ --pdb,--use-pdb                                           Drop into PDB on an exception          │
│ --help                      -h                            Show this message and exit.            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

cofin:
```
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────╮
│ --reload                -r                            Reload server on changes                   │
│ --reload-dir            -R  TEXT                      Directories to watch for file changes      │
│                                                       (TEXT)                                     │
│ --port                  -p  INTEGER                   Serve under this port (INTEGER)            │
│                                                       [default: 8000]                            │
│ --web-concurrency  -W  INTEGER RANGE [1<=x<=25]  The number of HTTP workers to launch       │
│                                                       (INTEGER RANGE)                            │
│                                                       [default: 1; 1<=x<=25]                     │
│ --host                  -H  TEXT                      Server under this host (TEXT)              │
│                                                       [default: 127.0.0.1]                       │
│ --file-descriptor  -F  INTEGER                   Bind to a socket from this file            │
│                                                       descriptor.                                │
│                                                       (INTEGER)                                  │
│ --unix-domain-socket    -U  TEXT                      Bind to a UNIX domain socket. (TEXT)       │
│ --debug                 -d                            Run app in debug mode                      │
│ --use-pdb               -P                            Drop into PDB on an exception              │
│ --help                  -h                            Show this message and exit.                │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```
Why don't we just do this?  I know it's a breaking change, but it's a change that I introduced recently in the beta.  There's no need to have 3 options for FD and UDS

JacobCoffee:
This would be what i prefer, but i just reverted back to not having the shorthand for those since it was 'breaking'

provinzkraut:
> This would be what i prefer, but i just reverted back to not having the shorthand for those since it was 'breaking'

I think there's some misunderstanding here. The shorthands weren't breaking, what was breaking was the removal of the previous ones (e.g. only `-w` instead of `-wc`).

To be honest, I am a bit confused now here as to what's what and what was reverted where, why and who is suggesting which changes and all :upside_down_face: 

@JacobCoffee I suggest you simply go ahead with this as you see fit. I feel like everyone involved here has stated their opinions on what's best enough times and I trust that you'll be able to figure something out ;)

provinzkraut (approved):

cofin (approved):

JacobCoffee:
Seems to be failing on merged in upstream changes.. going to merge this
image

PR #1917, opened 2023-07-04
Replace Headers with Dataclasses

Tsdevendra1:
### Pull Request Checklist

[//]: # "Please review the [Litestar contribution guidelines](https://github.com/litestar-org/litestar/blob/main/CONTRIBUTING.rst) for this repository."

- [x] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR

[//]: # "By submitting this issue, you agree to:"
[//]: # "- follow Litestar's [Code of Conduct](https://github.com/litestar-org/.github/blob/main/CODE_OF_CONDUCT.md)"
[//]: # "- follow Litestar's [Contribution Guidelines](https://litestar.dev/community/contribution-guide)"

### Description

Currently there are some left over pydantic models in the datastructure.headers namespace. This PR replaces them with dataclasses.

### Close Issue(s)

Closes #1386

PR #1926, opened 2023-07-08
Refactor headers code based on PR feedback

Tsdevendra1:
### Pull Request Checklist

### Description

This PR is for implementing feedback given in #1917. I had a look through the codebase for the usage of FieldDefinition, but didn't seem to see a similar case of casting from a string to the defined type in the FieldDefinition. I'm also not sure about the fact that I'm importing from `litestar.dto.factory.utils` to use `resolve_model_type`. I'd appreciate some feedback on this.


Goldziher:
hi @Tsdevendra1 - im not sure this is required anymore. I already did an update to the file on main.

Tsdevendra1:
From what I can see in [the current headers file on main](https://github.com/litestar-org/litestar/blob/main/litestar/datastructures/headers.py ) it still uses my old implementation of _get_type_annotations, your feedback mentioned that I should switch it to use FieldDefinition, is that not needed anymore?

Goldziher:
> From what I can see in [the current headers file on main](https://github.com/litestar-org/litestar/blob/main/litestar/datastructures/headers.py) it still uses my old implementation of _get_type_annotations, your feedback mentioned that I should switch it to use FieldDefinition, is that not needed anymore?

that would still be nice, pls continue

Tsdevendra1:
I've tried to improve the code using FieldDefinition in this PR, would you mind telling me if it's the right way? Or if there is a better approach?

Goldziher:
> I've tried to improve the code using FieldDefinition in this PR, would you mind telling me if it's the right way? Or if there is a better approach?

basically FieldDefinition does the parsing of the type for you - look inside the logic in that class. You can use a FieldDefinition instance to get a parsed annotation and a representation of any metadata it might include, see under `field_definition.kwarg_definition`. It also has a bunch of properties you can use, such as `is_optional` etc. so you do not need to do this stuff manually.

Tsdevendra1:
Is the current implementation not already relying on the correct methods?

```
    @classmethod
    def _get_field_definitions(cls) -> Dict[str, FieldDefinition]:
        if cls._field_definitions is None:
            cls._field_definitions = {}
            for key, value in get_type_hints(cls).items():
                definition = FieldDefinition.from_kwarg(annotation=value, name=key)
                # resolve_model_type so thatfield_definition.raw has the real raw type e.g. <class 'bool'>
                cls._field_definitions[key] = resolve_model_type(definition) 
        return cls._field_definitions
```

This seems to be close to what `get_model_type_hints` is doing which is used quite a bit through the codebase

Goldziher:
> Is the current implementation not already relying on the correct methods?
> 
> ```
>     @classmethod
>     def _get_field_definitions(cls) -> Dict[str, FieldDefinition]:
>         if cls._field_definitions is None:
>             cls._field_definitions = {}
>             for key, value in get_type_hints(cls).items():
>                 definition = FieldDefinition.from_kwarg(annotation=value, name=key)
>                 # resolve_model_type so that field_definition.raw has the real raw type e.g. <class 'bool'>
>                 cls._field_definitions[key] = resolve_model_type(definition) 
>         return cls._field_definitions
> ```
> 
> This seems to be close to what `get_model_type_hints` is doing which is used quite a bit through the codebase

this is good, but `get_type_hints` should be used with `include_extras=True` and be imported from `typing_extensions`

Tsdevendra1:
I've implemented your feedback, should be good to review again now

Goldziher (approved):
very nice!

PR #1943, opened 2023-07-12
feat(infra): add conventional commit pre-commit and semantic pull request workflow

Goldziher:
This PR does the following:

1. it adds conventional-commit pre-commit hook. This follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
2. it add s the semantic-PR github action, which enforces conventional PR titles following the same style guide.
3. it updates the readme and contribution guidelines to explain this.
4. it comments our the poetry action, since it currently bugs out.

### DIFF
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -1,0 +1,20 @@
+name: "Lint PR Title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,11 @@
 default_language_version:
   python: "3.11"
 repos:
+  - repo: https://github.com/compilerla/conventional-pre-commit
+    rev: v2.3.0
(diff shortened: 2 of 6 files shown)


Goldziher:
> How does the pre-commit hook work in conjunction with our "squash and merge" strategy? Wouldn't this result in whoever merges the PR having to write the commit message, which then would only get checked in the target branch ones the PR is merged?

We already discussed this internally - as I recall. Yes, it's not optimal. When squashing the squashed commit has all the messages appended into it. So you can see it as a sort of log.

provinzkraut:
> When squashing the squashed commit has all the messages appended into it. So you can see it as a sort of log.

That part I get. What I do not see working out however is the issue of having a pre-commit hook which would only fail once the PR is merged. The way I see it, this would simply add a lot of maintenance burden because of incorrectly squashed commits / commit messages in general. It also kinda defeats the purpose of such a fix if we cannot really automatically enforce it within the existing workflow.

Is there maybe a different step where we can run this check at that would solve this?

cofin (approved):
LGTM.

Goldziher:
> > When squashing the squashed commit has all the messages appended into it. So you can see it as a sort of log.
> 
> That part I get. What I do not see working out however is the issue of having a pre-commit hook which would only fail once the PR is merged. The way I see it, this would simply add a lot of maintenance burden because of incorrectly squashed commits / commit messages in general. It also kinda defeats the purpose if we cannot really automatically enforce it within the existing workflow.
> 
> Is there maybe a different step where we can run this check at that would solve this?

The hook will not run - to run the commit message hook you need to install it with a special command. See the contribution guide. 

I don't think this hook is critical at all, it's basically a guide for those who want it frankly. The PR action is more important.

JacobCoffee (approved):

PR #1950, opened 2023-07-14
feat(dto): re-introduce the `include` option to DTOConfig

Goldziher:
This PR reintroduces the removed `include` keyword. Include works the opposite of exclude. Namely, if exclude is specified, all fields that are not defined in exclude are assumed to be included by default - that is, unless they are marked as private fields. Include is the direct inverse - if a field is defined in include, it will be included, otherwise it is assumed to be excluded by default. This behavior is consistent with other libraries the expose a similar interface (e.g. pydantic).


cofin (approved):
Thanks for adding this one.  LGTM

litestar/dto/factory/base.py
provinzkraut, line 1: The rename is a breaking change and not within the described scope of the PR according to its title. Please revert.

provinzkraut (changes requested):

litestar/dto/factory/base.py
Goldziher: its a breaking change within a beta and for a type

provinzkraut: For entirely aesthetically reasons. And it doesn't matter if it's just "for a type". We said not to make breaking changes in the beta unless absolutely necessary. This is definitely not a necessity, and we shouldn't be breaking things for it.

PR #1959, opened 2023-07-15
style: Mark `Empty` and `CACHE_FOREVER` as `@typing.final`

sobolevn:
### Pull Request Checklist

### Description

mypy soon will have special narrowing for `@final` singleton types: https://github.com/python/mypy/pull/15646

So, it is a good idea to use `@final` for such types.

### DIFF
--- a/litestar/config/response_cache.py
+++ b/litestar/config/response_cache.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, final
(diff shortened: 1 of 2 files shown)


provinzkraut:
> mypy soon will have special narrowing for @final singleton types

Cool, didn't know about this special support!

provinzkraut (changes requested):
We should apply this to `CACHE_FOREVER` here as well:

https://github.com/litestar-org/litestar/blob/b5f31a9b3fe9b53bbfe786b78f67b3fbb36d8f96/litestar/config/response_cache.py#L17

PR #1963, opened 2023-07-15
feat(internal): Make Pydantic into an optional extra

Goldziher:
This PR does the following:

1. it makes pydantic into an optional extra.
2. it moves the pydantic logic into plugins (for ser/deser and schema)
3. it refacors the DTO logic because some stuff was no longer required there. It adds deprecation warnings for file moves.
4. it cleansup types.


provinzkraut:
So, I want to ask again here: Why not make Pydantic support a plugin and move everything Pydantic to `contrib`? Seems like the logical thing to do.

Goldziher:
> So, I want to ask again here: Why not make Pydantic support a plugin and move everything Pydantic to `contrib`? Seems like the logical thing to do.

you mean a serialization / schema plugin?

The problem with this is that currently the assumption we have is that if users have pydantic in place it simply works out of the box, same as attrs. 

We can implement these as plugins, but this is a pretty big effort, especially given the amount of changes to our tests that we will need to introduce. 

We will also need to rethink how we handle the enc/dec hooks for msgspec, and the very structure of plugins as well - currently the plugins are not designed for this use case at all. 

All in all I'd say doing this might well be a good future direction, but its a pretty big piece of work that would take us a while to achieve. We should consider doing this not for v2 at all but rather for v3 IMO.

provinzkraut:
> We can implement these as plugins, but this is a pretty big effort, especially given the amount of changes to our tests that we will need to introduce.

Agreed. But it is the proper way to handle it. What we currently have are quite a lot of workarounds that will be hard to maintain in the long run. One of the main reasons of having an msgspec signature model only was that we could move away from the hard-coded Pydantic support everywhere.

Regarding the tests, I think it will probably lead to better coverage since we currently don't properly test all cases for msgspec/pydantic/attrs equally. If the latter two are simply plugins we can more easily achieve a sensible testing structure for these, since their boundaries are more clearly defined.

> The problem with this is that currently the assumption we have is that if users have pydantic in place it simply works out of the box, same as attrs.

That's a pretty easy thing to overcome though. We can simply do a check on initialisation.

If Pydantic 1 is installed, add the `Pydantic1Plugin`, if Pydantic 2 is installed, add the `Pydantic2Plugin`, and if attrs is installed, add the `AttrsPlugin`.

> currently the plugins are not designed for this use case at all.

For which use case? Serialisation and OpenAPI schemas? Isn't that *specifically* what they were designed for?

Goldziher:
> For which use case? Serialisation and OpenAPI schemas? Isn't that _specifically_ what they were designed for?

They were not designed to be enc/dec hooks to msgspec. 

Which is basically what they will need to be changed to - so we can use them in the signature model when we call `convert` and internaly when we call `encode_json` and `encode_msgpack`. 

Frankly I'm not sure where and how we use them now for serialization, after the changes to DTOs etc.

litestar/dto/factory/_backends/pydantic/backend.py
provinzkraut, line 1: Why has this been removed entirely? Isn't this still needed?

Goldziher: No, it works without it.

provinzkraut (commented):
Why the breaking changes to the structure of the `dto` submodule?
Are they really necessary?

litestar/handlers/base.py
```diff
@@ -45,7 +56,9 @@ class BaseRouteHandler:
         "_resolved_layered_parameters",
         "_resolved_return_dto",
         "_resolved_signature_namespace",
+        "_resolved_type_decoders",
         "_resolved_type_encoders",
+        "_decoder",
```
Goldziher, line 61:
```suggestion
```

Goldziher:
> Why the breaking changes to the structure of the `dto` submodule?
> Are they really necessary?

Well, we are still in beta. We can change this stuff - and it's all changes from internal namespaces that are exported top level. 

Why? The structure was bad and made little sense given the redundancies.

provinzkraut:
> Well, we are still in beta. We can change this stuff - and it's all changes from internal namespaces that are exported top level.

It's not "internal namespaces". Those are public APIs that are documented. 
Yes, we are in beta. But that doesn't mean we can just break stuff. We explicitly stated that we **will not break things during the beta unless absolutely necessary*. This is not such a case, since the changes introduces within this PR do not require this restructuring and renaming.

> Why? The structure was bad and made little sense given the redundancies.

If the structure was bad to begin with, that should have been caught when the PRs that introduced it were initially reviewed. There's a whole bunch of other things that could be restructured but we don't do that either because the time to do that is not during beta, at least not with public facing APIs.

The whole point of a beta is that users can test out and find bugs in the library *as it is*. If we keep making breaking changes, we also break the contract and promise we have made to our users and defeat the purpose of a beta.

Goldziher:
> > Well, we are still in beta. We can change this stuff - and it's all changes from internal namespaces that are exported top level.
> 
> It's not "internal namespaces". Those are public APIs that are documented. It will break virtually all code that uses DTOs.
> 
> Yes, we are in beta. But that doesn't mean we can just break stuff. We explicitly stated that we **will not break things during the beta unless absolutely necessary*. This is not such a case, since the changes introduces within this PR do not require this restructuring and renaming.
> 
> > Why? The structure was bad and made little sense given the redundancies.
> 
> If the structure was bad to begin with, that should have been caught when the PRs that introduced it were initially reviewed. There's a whole bunch of other things that could be restructured but we don't do that either because the time to do that is not during beta, at least not with public facing APIs.
> 
> The whole point of a beta is that users can test out and find bugs in the library *as it is*. If we keep making breaking changes, we also break the contract and promise we have made to our users and defeat the purpose of a beta. 
> 

Well, there are breaking changes here - not many. Shoulda? Well, maybe, but they weren't and that's the way it is now. 

The structure is bad, the actual changes is to love files from the internal factory namespace to the outer DTO namespace, because it's the wrong place for them. 

The breakage is minimal because of the use of init files. 

Anyhow, yes - it will break stuff that explicitly relied on these export paths. I don't see it as a big deal because we are in beta. And as for the commitment to try and not break stuff, it's basically just that. 

Please review the PR otherwise, since I'm going to insist on these changes given I removed the redundant pydantic factory.

litestar/plugins.py
```diff
@@ -107,11 +103,13 @@ def is_plugin_supported_type(value: Any) -> TypeGuard[ModelT]:
         """
         raise NotImplementedError()
 
-    def to_openapi_schema(self, model_class: type[ModelT]) -> Schema:
-        """Given a model class, transform it into an OpenAPI schema class.
+    def to_openapi_schema(self, annotation: Any, schema_creator: SchemaCreator, dto_for: ForType | None) -> Schema:
```
Goldziher, line 106: this is a breaking change - reason for this is two-fold:

1. that we cannot actually create schemas only based on the annotation only, we need the schema_creator and dto_for info. 
2. the assumption of the plugin was that we are dealing with "models" - but pydantic actually has to serialize a lot of specialized annotations that are not models. Hence the plugin needs to be more generalized.

litestar/plugins.py
```diff
@@ -90,13 +86,13 @@ def create_dto_for_type(self, field_definition: FieldDefinition) -> type[DTOInte
...
 @runtime_checkable
-class OpenAPISchemaPluginProtocol(Protocol[ModelT]):
+class OpenAPISchemaPluginProtocol(Protocol):
     """Plugin to extend the support of OpenAPI schema generation for non-library types."""
 
     __slots__ = ()
 
     @staticmethod
-    def is_plugin_supported_type(value: Any) -> TypeGuard[ModelT]:
+    def is_plugin_supported_type(value: Any) -> bool:
```
Goldziher, line 95: minor breaking change - see the point regarding the other method. Namely, we are not dealing with a model here at all but rather with a lot of different types.

litestar/plugins.py
Goldziher: note - schema_creator is currently in a private namespace `_openapi`. I dont actually want to expose it to users TBH, but there is little we can do given we decided on this switch. We might need to document it. Thoughts?

litestar/serialization/msgspec_hooks.py
```diff
@@ -98,44 +88,48 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
...
 
-def encode_json(obj: Any, default: Callable[[Any], Any] | None = None) -> bytes:
+def encode_json(value: Any, serializer: Callable[[Any], Any] | None = None) -> bytes:
     """Encode a value into JSON.
 
     Args:
-        obj: Value to encode
-        default: Optional callable to support non-natively supported types.
+        value: Value to encode
+        serializer: Optional callable to support non-natively supported types.
```
Goldziher, line 132: i changes the name throughout - we are leaking implementation details IMO, and its not a very nice convention.

litestar/serialization/msgspec_hooks.py
```diff
@@ -194,71 +205,61 @@ def encode_msgpack(obj: Any, enc_hook: Callable[[Any], Any] | None = default_ser
...
-    except msgspec.DecodeError as msgspec_error:
-        raise SerializationException(str(msgspec_error)) from msgspec_error
+@overload
+def decode_msgpack(value: bytes, target_type: type[T], type_decoders: TypeDecodersSequence | None) -> T:
+    ...
 
 
-def decode_media_type(raw: bytes, media_type: MediaType | str, type_: Any) -> Any:
-    """Decode a raw value into an object.
+def decode_msgpack(value: bytes, target_type: type[T] | EmptyType = Empty, type_decoders: TypeDecodersSequence | None = None) -> Any:  # type: ignore[misc]
```
Goldziher, line 235: note in general here - i am not sure the use of overloads serves us. In the end we get a lot of typing issues still because of this implementation pattern. We can just drop the overloads and generic and return Any, which is pretty standard for dump/load logic.

litestar/testing/request_factory.py
```diff
@@ -209,7 +211,7 @@ def _create_request_with_data(
         user: Any = None,
         auth: Any = None,
         request_media_type: RequestEncodingType = RequestEncodingType.JSON,
-        data: dict[str, Any] | BaseModel | None = None,
+        data: dict[str, Any] | DataContainerType | None = None,  # pyright: ignore
```
Goldziher, line 214: not really in the same scope of the PR but it was either removing this or supporting everything else we support. I opted for the latter, since its easy enough to achieve. This is thus a micro feature- making the `request_factory` slightly better.

provinzkraut (approved):

PR #1993, opened 2023-07-17
chore: Add piccolo as a docs dependency

provinzkraut:
Piccolo has been removed in #1963, because it depends on Pydantic <2. Since we can't introduce that constraint to our dev dependencies, piccolo is moved to a doc dependency until it supports Pydantic 2.


Goldziher:
this was actually intentional - the issue is that they also have a pin on pydantic.

PR #1995, opened 2023-07-17
chore: Fix PyPi readme generation on all contributors bump

provinzkraut:
Remove the all contributors badge from the PyPi readme, since it will cause pre-commit fails after merging a PR from the all-contributors bot.


JacobCoffee:
Oh this is just the pypi.. sorry. Reading from my phone

JacobCoffee (approved):

provinzkraut:
> Oh this is just the pypi.. sorry. Reading from my phone

Yeah. I figured this is the most sane way to handle this. We don't really need the all contributors stuff on PyPi so stripping it out enables us to autogenerate the PyPi readme within pre-commit.

PR #2000, opened 2023-07-18
chore(internal): fix pydantic v1 serialization of bytes

Goldziher:
This PR fixes the serialization of bytes for pydantic v1. It also removes the `pydantic-extra-types` package from our optionals because this package causes pydantic v1 to restricted out from our lock file. 

This is a followup for a PR by @sobolevn


litestar/contrib/pydantic/__init__.py
```diff
@@ -16,7 +16,7 @@ def _model_dump(model: pydantic.BaseModel, *, by_alias: bool = False) -> dict[st
     return (
         model.model_dump(mode="json", by_alias=by_alias)
         if hasattr(model, "model_dump")
-        else model.dict(by_alias=by_alias)
+        else {k: v.decode() if isinstance(v, bytes) else v for k, v in model.dict(by_alias=by_alias).items()}
```
sobolevn, line 19: Maybe this can be moved to its own helper?

provinzkraut (approved):

PR #2001, opened 2023-07-18
fix(serialization): fixed url encoded serialization

Goldziher:
This PR fixes the handling of URL encoded strings - previously we would parse numerical values using rust and then let pydantic switch this to string if needed.


litestar/_parsers.py
```diff
@@ -21,7 +21,7 @@ def parse_url_encoded_form_data(encoded_data: bytes) -> dict[str, Any]:
     Returns:
         A parsed dict.
     """
-    return parse_url_encoded_dict(encoded_data)
+    return parse_url_encoded_dict(encoded_data, False)
```
provinzkraut, line 24: Kwarg for readability?

provinzkraut (approved):

litestar/_parsers.py
Goldziher: done

PR #2002, opened 2023-07-18
refactor: remove partial, fix partial DTO schema generation, update openapi controller versions

Goldziher:
This PR does the following:

1. it removes the `partial` module and updates the examples to use the DTO type for this.
2. after doing this i got failures in openapi schema generation and the DTO generation, which led to fixes there.
3. I also updated the versions in the open API controller. 

I realize that this PR is eclectic. apologies for this.


docs/release-notes/whats-new-2.rst
```diff
@@ -52,7 +52,7 @@ Imports
 +====================================================+========================================================================+
 | ``starlite.ASGIConnection``                        | :class:`.connection.ASGIConnection`                                    |
 +----------------------------------------------------+------------------------------------------------------------------------+
-| ``starlite.Partial``                               | :class:`.partial.Partial`                                              |
+| ``starlite.Partial``                               | replaced with the DTO                                                  |
```
provinzkraut, line 55:
```suggestion
| ``starlite.Partial``                               | replaced with DTOs                                                     |
```

litestar/partial.py
provinzkraut, line 1: Maybe a deprecation warning for this instead of outright removal?
Same way we handled the DTO refactoring. We can then remove this in the next beta / rc.

tests/unit/test_partial.py
provinzkraut, line 1: If we decide to deprecate the partial at first, we should keep this one around as well for the time being.

provinzkraut (approved):
Nice! You beat me to it :stuck_out_tongue:

litestar/partial.py
Goldziher: done

PR #2012, opened 2023-07-19
chore(internal): refactor dto factory

Goldziher:
This PR refactors the DTO backend and factory, reducing complexity and improving the handling of types and schemas.


litestar/dto/dataclass_dto_factory.py
```diff
@@ -47,7 +45,7 @@ def generate_field_definitions(
                     field_definition=field_definition,
                     default_factory=default_factory,
                     dto_field=dc_field.metadata.get(DTO_FIELD_META_KEY, DTOField()),
-                    unique_model_name=get_fully_qualified_class_name(model_type),
+                    model_name=model_type.__name__,
```
cofin, line 48: Should we use `__qualname__` here instead?

tests/unit/test_contrib/test_pydantic/test_pydanticdt_dto_factory.py
```diff
@@ -28,10 +28,7 @@ class TestModel(BaseModel):
         e: int = Field(default_factory=int_factory)
 
     field_defs = list(PydanticDTO.generate_field_definitions(TestModel))
-    assert (
-        field_defs[0].unique_model_name
-        == "tests.unit.test_contrib.test_pydantic.test_pydanticdt_dto_factory.test_field_definition_generation.<locals>.TestModel"
-    )
+    assert field_defs[0].model_name == "TestModel"
```
cofin, line 31: Not related to this specific line, but is there any special meaning to the the `dt` part in this filename?  Should it just be `test_pydantic_dto_factory.py`?

cofin (approved):
LGTM other than the file rename.  After reviewing the test cases and the shortened names, I don't think we need to use `__qualname__`.

litestar/dto/dataclass_dto_factory.py
peterschutt: The unique naming was brought in to prevent errors with openapi schema generation caused by having multiple transfer models with the same name. Has that been addressed some other way?

provinzkraut: > Has that been addressed some other way?

It has not, no. As far as I’m aware, the behaviour is still the same?

Goldziher: it has, yes. and unlike the previous version which created some weird random names with numbers and made the open api docs half usable, this is actually readable.

Issue #1185, opened 2023-02-10
Enhancement: Server sent events support


KestutisMa:
Amazing project, but I am little not sure if it is possible to server send events in Starlite? Or only way to send message is by pooling?
There is package sse-starlette for Starlette and FastApi https://github.com/sysid/sse-starlette

PR #2035, opened 2023-07-21
feat: add SSE support

Goldziher:
This PR adds SSE (Server Sent Event) support, closing #1185


cofin (approved):
LGTM.  I actually expected it to be a bit more complicated than this.

PR #2044, opened 2023-07-22
fix(multipart): serialization on multipart form

Goldziher:
This PR fixes a bug with multipart handling, where empty string values raised a validation error.

### DIFF
--- a/tests/unit/test_kwargs/test_multipart_data.py
+++ b/tests/unit/test_kwargs/test_multipart_data.py
@@ -1,13 +1,13 @@
-# flake8: noqa
 from collections import defaultdict
+from dataclasses import dataclass
 from os import path
 from os.path import dirname, join, realpath
 from pathlib import Path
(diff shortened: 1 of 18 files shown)


JacobCoffee (approved):

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -1,12 +1,11 @@
-# flake8: noqa
 from collections import defaultdict
+from dataclasses import dataclass
 from os import path
 from os.path import dirname, join, realpath
 from pathlib import Path
-from typing import Any, DefaultDict, Dict, List, Optional, Type
+from typing import Annotated, Any, Dict, List, Optional
```
JacobCoffee, line 6: This fails on 3.8, would we need to use `typing-extensions`?

JacobCoffee (changes requested):

PR #2045, opened 2023-07-22
feat: additional repository enhancements for bulk upsert operations

cofin:
feat: allow identity column to be more easily customized (this allows easier use of non-PK identity columns from routes)
feat: Adds on_or_before filters
feat: adds not in collection
feat: adds not like search function
fix: health function renders a valid statement for all dialects now

There's still an opportunity to optimize the upsert operation, but I think the best route is to work on a custom operation that will actually execute a SQL `MERGE` statement where possible.  That is a future enhancements.


cofin:
@provinzkraut do you want to merge this one before or after the type parameters PR?

cofin:
I will reopen individual PRs with each enhancement.

PR #2049, opened 2023-07-23
feat(testing): switch test clients to debug=True by default

Goldziher:
This PR sets `debug=True` in our test clients by default. It additionally ensures that traceback is returned in the appropriate media type for the request.


cofin (approved):
LGTM

PR #2066, opened 2023-07-25
feat(cli): CLI plugins

provinzkraut:
Add a new `CLIPluginProtocol` that can be used to create CLI plugins.

These contain a single method `on_cli_init`, which will be called when the CLI is initialised. This allows to dynamically register commands, without having to deal with the optional click dependency.

```python
from litestar import Litestar
from litestar.plugins import CLIPluginProtocol

    def on_cli_init(self, cli):
        @cli.command()
        def is_debug_mode(app: Litestar):
            print(app.debug)

app = Litestar(plugins=[CLIPlugin()])
``` 

```shell
$ litestar is-debug-mode

Using Litestar app from app:app
False
```

### DIFF
--- a/docs/usage/cli.rst
+++ b/docs/usage/cli.rst
@@ -252,11 +252,17 @@ This will result in:
 Extending the CLI
 -----------------
 
-Litestar's CLI is built with `click <https://click.palletsprojects.com/>`_ and can be easily extended.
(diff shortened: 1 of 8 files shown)


cofin (approved):
This will work perfectly for the alembic commands.  LGTM.

PR #2072, opened 2023-07-27
feat(tests): Parallelize test suite with pytest-xdist

provinzkraut:
Add [pytest-xdist](https://pytest-xdist.readthedocs.io) to speed up the tests.
A few tests had to be adjusted along the way to eliminate race conditions and combat lack of isolation (=

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -68,13 +68,13 @@ jobs:
         run: echo "PYTHONPATH=$PWD" >> $GITHUB_ENV
       - name: Test
         if: ${{ !inputs.coverage }}
-        run: poetry run pytest docs/examples tests
+        run: poetry run pytest docs/examples tests -n auto
       - name: Test with coverage
         if: inputs.coverage
-        run: poetry run pytest docs/examples tests --cov=litestar --cov-report=xml
+        run: poetry run pytest docs/examples tests --cov=litestar --cov-report=xml -n auto
       - name: Integration tests
         if: inputs.integration
-        run: poetry run pytest -m="sqlalchemy_integration"
+        run: poetry run pytest -m="sqlalchemy_integration" -n auto
       - uses: actions/upload-artifact@v3
         if: inputs.coverage
         with:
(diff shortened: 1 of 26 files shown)


provinzkraut:
I'm not yet sure whether to enable this by default. The issue is that `pytest-xdist` doesn't work well with most debuggers, which means you'd have to set either the `--pdb` flag or `-n=0` when running pytest to get the debuggers working.

Now instead, you'll have to explicitly set `-n=auto` to enable `pytest-xdist`.

I guess it depends on the individual workflows which is more convenient.
@litestar-org/maintainers @litestar-org/members please provide your input! (=

Goldziher:
please enable it in CI, not locally.

JacobCoffee (approved):
pytest go brrrrrrrr

provinzkraut:
> please enable it in CI, not locally.

Hm. I quite like having it locally, but I didn't want to force people to use it, so it left it out of the general pytest strategy and only added it to the `Makefile`. Do you guys use those?

My thinking was that this gives - with relative ease - easy access to both:

- Run `make test` to run the full speed, complete suite
- Run `pytest <whatever>` to debug things, without having to override anything

This made sense to me as I assumed you'd usually run individual parts while developing / debugging, where `pytest-xdist` would get in the way, and the whole suite *not* while debugging, where `pytest-xdist` could be useful since it cuts down the time.

If I'm in the minority here with this and it's more convenient for you guys I'm happy to change the setup though (=

JacobCoffee:
Yeah that works, i think. I use the makefile a lot but I can use pytest by itself if i run into issues. I am not worried really about issues I run into but onboarding new contributors that may have 50 questions. but this solves it, i think

provinzkraut:
> onboarding new contributors that may have 50 questions

I actually thought it'd make it easier, since it's somewhat of a sane default, and you won't have to add special flags / options to make stuff work!

Goldziher:
Ok, fine for me. I dont use the make file but the IDE and i constantly use the debugger.

provinzkraut:
> Ok, fine for me. I dont use the make file but the IDE and i constantly use the debugger.

Well then you're just like me and everything should work as expected since these runs won't use pytest-xdist (=

PR #2074, opened 2023-07-27
feat: pre-calculate decoders in signature model

Goldziher:
This PR improves our deserialization by pre-calculating the required decoders for the signature. 

I added a hack to bypass msgspec, which strips Annotated before hitting the hooks. I hope this is temporary - see: https://github.com/jcrist/msgspec/issues/497

### DIFF
--- a/litestar/_signature/types.py
+++ b/litestar/_signature/types.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Callable, ClassVar
 
 from msgspec import ValidationError
 
@@ -9,3 +9,7 @@ class ExtendedMsgSpecValidationError(ValidationError):
     def __init__(self, errors: list[dict[str, Any]]) -> None:
         self.errors = errors
         super().__init__(errors)
+
+
+class SerializationWrapper:
+    _decoder: ClassVar[Callable[[type, Any], Any]]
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -23,14 +23,9 @@ documentation = "https://docs.litestar.dev/"
 keywords = [
     "api",
     "rest",
-    "http",
```
cofin, line 26: What's the reason for removing some of these keywords?

litestar/_signature/types.py
```diff
@@ -9,3 +9,7 @@ class ExtendedMsgSpecValidationError(ValidationError):
     def __init__(self, errors: list[dict[str, Any]]) -> None:
         self.errors = errors
         super().__init__(errors)
+
+
+class SerializationWrapper:
```
provinzkraut, line 14: This doesn't seem to get used anywhere? Did you accidentally drop some commits along the way? :upside_down_face:

pyproject.toml
Goldziher: linting errors in my IDE letting me know that keywords are limited to 5 max. intellij got updated.

litestar/_signature/types.py
Goldziher: ha yes, lemme remove it.

provinzkraut (approved):

PR #2078, opened 2023-07-27
feat(schema): support configuring operation_id_creator on the route h…

Goldziher:
This PR allows `operation_id` on route handlers to be a callable. Previously we supported this only on the global open api configuration, but it also makes sense to allow this to be locally overriden.


cofin (approved):
LGTM

JacobCoffee (approved):

PR #2080, opened 2023-07-28
refactor(tests): use in-memory SQLite databases for testing where possible

provinzkraut:
Refactor test that use SQLite (SQLAlchemy / piccolo) to use an in-memory SQLite database instead of a file

### DIFF
--- a/tests/examples/test_contrib/test_piccolo/test_piccolo_crud.py
+++ b/tests/examples/test_contrib/test_piccolo/test_piccolo_crud.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 from unittest import TestCase
 
 import pytest
@@ -24,6 +25,9 @@ class TestCrud(TestCase):
 
     def tearDown(self):
         Task.alter().drop_table().run_sync()
+        from docs.examples.contrib.piccolo.piccolo_conf import DB
+
+        Path(DB.path).unlink()
 
     def test_get_tasks(self):
         with TestClient(app=app) as client:
--- a/tests/examples/test_contrib/test_sqlalchemy/plugins/test_example_apps.py
+++ b/tests/examples/test_contrib/test_sqlalchemy/plugins/test_example_apps.py
@@ -3,10 +3,10 @@ from __future__ import annotations
 from typing import Any
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
+from sqlalchemy import Engine, StaticPool, create_engine
 
 from litestar.testing import TestClient
-
-pytestmark = pytest.mark.xdist_group("sqla-plugin-examples")
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


tests/examples/test_contrib/test_piccolo/test_piccolo_crud.py
provinzkraut, line 1: I couldn't get piccolo to work with an in-memory SQLite database because apparently there's some hidden global state somewhere that leads to it not properly initialising tables between test runs. If anyone knows how to do this properly, please submit a PR.

Also, this should be refactored to use pytest instead of unittest.

sinisaos: @provinzkraut Sorry for the interruption, but you could try using a fixture for database setup and teardown like this
```python
@pytest.fixture(autouse=True, scope="module")
def create_test_data():
    Task.create_table(if_not_exists=True).run_sync()
    ModelBuilder.build_sync(Task)
    yield
    Task.alter().drop_table().run_sync()
```
Here is the modified file (with pytest instead unittest) which passed the tests. You can use that in this PR (if that's OK with you and I haven't missed the point) or do I need to make another PR?
```python
import pytest

try:
    import piccolo  # noqa: F401
except ImportError:
    pytest.skip("Piccolo not installed", allow_module_level=True)

from docs.examples.contrib.piccolo.app import Task, app
from piccolo.testing.model_builder import ModelBuilder


@pytest.fixture(autouse=True, scope="module")
def create_test_data():
    Task.create_table(if_not_exists=True).run_sync()
    ModelBuilder.build_sync(Task)
    yield
    Task.alter().drop_table().run_sync()


@pytest.mark.xdist_group("examples-piccolo")
def test_get_tasks():
        response = client.get("/tasks")
        assert len(response.json()) == 1


@pytest.mark.xdist_group("examples-piccolo")
def test_task_crud():
        payload = {
            "name": "Task 1",
            "completed": False,
        }

        response = client.post(
            "/tasks",
            json=payload,
        )
        assert response.status_code == 201
        assert response.json()["name"] == "Task 1"

        response = client.get("/tasks")
        assert len(response.json()) == 2

        task = Task.select().first().run_sync()

        payload = {
            "id": task["id"],
            "name": "Task 2",
            "completed": True,
        }

        response = client.patch(
            f"/tasks/{task['id']}",
            json=payload,
        )
        assert response.json()["name"] == "Task 2"
        assert response.json()["completed"] is True

        response = client.delete(
            f"/tasks/{task['id']}",
        )
        assert response.status_code == 204

        response = client.get("/tasks")
        assert response.json()[0]["name"] == "Task 1"
        assert len(response.json()) == 1
```

provinzkraut: It should be a separate PR. Feel free to open one though (=

provinzkraut: > Sorry for the interruption, but you could try using a fixture for database setup and teardown like this

A fixture wouldn't solve the issue though, since this would effectively do the same thing as `setUp` and `tearDown`. If you want to give it a go, try patching the `DB` to use `path=":memory":`.

sinisaos: > If you want to give it a go, try patching the `DB` to use `path=":memory":`.

We can use `shared_cache` to achieve that like this (but I don't understand why we don't just use the `test.sqlite` that already exists?)

```python
# in docs/examples/contrib/piccolo/piccolo_conf.py
DB = SQLiteEngine(path=":memory:?cache=shared")
```
Here is updated example

```python
from pathlib import Path
from docs.examples.contrib.piccolo.app import DB, Task, app
from piccolo.testing.model_builder import ModelBuilder


@pytest.fixture(autouse=True, scope="module")
def create_test_data():
    Task.create_table(if_not_exists=True).run_sync()
    ModelBuilder.build_sync(Task)
    yield
    Task.alter().drop_table().run_sync()
    Path(DB.path).unlink()
```
If you are happy with the updated example above, I can do a PR.

provinzkraut: > but I don't understand why we don't just use the test.sqlite that already exists?

*Because* it already exists, not despite. The main reason here is simply isolation. Using an in-memory database means you won't:

1. Have to make sure the database is empty for each run
2. Ensure that it runs on in the same process to avoid race conditions

In general it's best practice to isolate your tests as good as you can, and using an ephemeral in-memory database is simply a very convenient way to achieve this!

<hr>

I'll update this PR with the snippet you provided to get the in-memory stuff working. For the rest, you're welcome create a separate PR (you can base it off of this branch if you want).

sinisaos: OK. Feel free to use whatever you find useful. Cheers.

provinzkraut:
> Importing from docs is wild to me

You mean the test-testing things? I don't like it either, but the examples are already tests, so it kinda makes sense that we just run those.

Issue #2018, opened 2023-07-20
Bug: Pagination breaks serialization of SQL Alchemy models


audiovention:
### Description

Returning directly sql models (with the SQLAlchemySerializationPlugin) is broken when trying to paginate the results with some of the built-in classes. Same issue happens with return_dtos as well. See code for more details

### MCVE

```python
from litestar import Litestar, get
from litestar.pagination import ClassicPagination
from litestar.contrib.sqlalchemy.plugins import SQLAlchemySerializationPlugin

from litestar.contrib.sqlalchemy.base import UUIDAuditBase
from sqlalchemy.orm import Mapped

class User(UUIDAuditBase):
    first_name: Mapped[str]


users= [User(first_name="ASD"), User(first_name="qwe")]


@get("/broken")
async def broken() -> ClassicPagination[User]:
    return ClassicPagination[User](items=users, page_size=2, current_page=1, total_pages=1)

@get("/working")
async def working() -> list[User]:
    return users


app = Litestar(route_handlers=[broken, working], 
    plugins=[SQLAlchemySerializationPlugin()]
)
```

### Steps to reproduce

```bash
1. Go to '/schema/swagger'
2. Inspect documentation of both endpoint - notice that the paginated one doesn't show the actual model, but just "string"
3. Go to '/broken' - observe the stack trace - litestar.exceptions.base_exceptions.SerializationException: Unsupported type: <class 'app.User'>
```

### Litestar Version

litestar==2.0.0b2
python 3.10.7

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

## Funding
* If you would like to see an issue prioritized, make a pledge towards it!
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/litestar-org/litestar/issues/2018">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2018/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2018/pledge.svg">
</picture>
</a>

Issue #1929, opened 2023-07-10
Bug: Using a bound generic type with msgspec DTO fails to decode


abdulhaq-e:
### Description

Please see the MCVE and note the two possible ways to circumvent this. I'm not sure if this is expected behaviour. Regardless, the test failure happens at the assertion. The actual logic doesn't throw and silently returns: `b'{"age":10,"data":{}}` which it shouldn't do silently imo.

### MCVE

```python
def test_dto_with_msgspec_and_generic_inherited_models(create_module: Callable[[str], ModuleType]) -> None:
    module = create_module(
        """
from typing import Dict, Generic, TypeVar
from typing_extensions import Annotated

from litestar import Litestar, post
from litestar.dto.factory import DTOConfig
from litestar.contrib.msgspec import MsgspecDTO

from msgspec import Struct

T = TypeVar("T", bound=Struct) # removing the bound argument solves the issue

class ClassicNameStyle(Struct):
    first_name: str
    surname: str

class User(Struct, Generic[T]):
    age: int
    data: T

class Superuser(User[ClassicNameStyle]):
    # uncommenting the line below also solves the issue (while keeping bound)
    # data: ClassicNameStyle 
    pass

class SuperUserDTO(MsgspecDTO[Superuser]):
    pass

@post(dto=SuperUserDTO)
def handler(data: Superuser) -> Superuser:
    return data

app = Litestar(route_handlers=[handler])
"""
    )
    with TestClient(app=module.app) as client:
        data = module.Superuser(data=module.ClassicNameStyle(first_name="A", surname="B"), age=10)
        headers = dict()
        headers["Content-Type"] = "application/json; charset=utf-8"
        received = client.post(
            "/",
            content=msgspec.json.encode(data),
            headers=headers,
        )
        print(received.content)
        assert msgspec.json.decode(received.content, type=module.Superuser) == data
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Litestar Version

latest

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

## Funding
* You can sponsor this specific effort via a [Polar.sh](https://polar.sh) pledge below
* We receive the pledge once the issue is completed & verified

[...]

PR #2088, opened 2023-07-29
feat: rework and simplify DTOs

Goldziher:
This PR does the following:

1. It removes the `DTOInterface` class, which was both unnecessary and problematic - all DTOs must include the logic that is on the base DTO class, and this was wrong. 
2. It renames `AbstractDTOFactory` into `AbstractDTO` - this is more correct since all its subclasses are called `DataclassDTO`, `MsgspecDTO` etc. 
3. It removes the `ForType` and the logic associated with it - this is no longer necessary and it just added complication. 
4. It removes the unnecessary layers of abstraction added in multiple places - these were not helping and made the code more complicated and harder to grasp.
5. It simplifies and merges the encoding/decoding logic on the DTO classes.
6. It simplifies and conventionalizes the registration logic in handlers, which was a bit messy and hacky.
7. It fixes multiple issues with the registration of DTOs and the handling of nested types. 
8. It cleansup some parts of the tests.
9. It fixes an issue with pagination and DTOs. 

The following are breaking changes:

* removal of `DTOInterface`
* Removal of `ForType`
* Rename of `AbstractDTOFactory` -> `AbstractDTO`
* Replacement of the DTO methods `builtins_to_data_type` and `bytes_to_data_type` with `decode`
* Replace of the DTO method `data_to_encodable_type` with `encode`

It closes:

#2018 
#1929

This is the last change to DTOs before we go for release. From this point onward its only bug fixes. 

Apologies for the scale of this PR - it required a rather deep dive to fully work out the code and simplify it.


JacobCoffee:
This is a monster ._. 👹

Goldziher:
> This is a monster ._. 👹 

Not quite done yet...

litestar/dto/_backend.py
```diff
@@ -102,39 +99,13 @@ def __init__(
...
-                and not self.is_data_field
-                or field_definition.dto_for == "return"
-                and self.is_data_field
-            ):
-                continue
+            # if field_definition.name and (
+            #     and not self.is_data_field
+            #     or field_definition.name != "data"
+            #     and self.is_data_field
+            # ):
```
JacobCoffee, line 108: can we remove this?

Goldziher: Yes, good catch

tests/unit/test_contrib/test_sqlalchemy/test_serialization_plugin.py
```diff
@@ -57,3 +63,19 @@ def get_a() -> A:
         assert response.json() == [{"id": 1, "a": "test"}, {"id": 2, "a": "test2"}]
         response = client.get("/a/1")
         assert response.json() == {"id": 1, "a": "test"}
+
+
+class User(UUIDAuditBase):
+    first_name: Mapped[str]
```
cofin, line 69:
```suggestion
    first_name: Mapped[str] = mapped_column(String(200))
```

tests/unit/test_contrib/test_sqlalchemy/test_serialization_plugin.py
```diff
@@ -2,7 +2,13 @@
 
 from typing import TYPE_CHECKING
 
+from sqlalchemy.orm import Mapped
```
cofin, line 5:
```suggestion
from sqlalchemy import String
from sqlalchemy.orm import Mapped, mapped_column
```

litestar/handlers/base.py
```diff
@@ -131,33 +134,66 @@ def __init__(
...
 
     def __call__(self, fn: AsyncAnyCallable) -> Self:
         """Replace a function with itself."""
         self._fn = Ref["MaybePartial[AsyncAnyCallable]"](fn)
         return self
 
+    @property
+    def handler_id(self) -> str:
+        """A unique idenrtifier used for generation of DTOs and schemas."""
+        return self.name or str(self)
```
provinzkraut, line 158: I think this might be an issue in cases where the same handler is registered multiple times but inherits different configs from upper layers. Maybe we should somehow take the ownership layers into account here and / or ensure that explicitly given names are globally unique?

litestar/handlers/base.py
```diff
@@ -170,7 +206,7 @@ def fn(self) -> Ref[MaybePartial[AsyncAnyCallable]]:
             Handler function
         """
         if not hasattr(self, "_fn"):
-            raise ImproperlyConfiguredException("Handler has not decorated a function")
+            raise ImproperlyConfiguredException("handler does not decorate a function")
```
provinzkraut, line 209: This error message is ambiguous. 

How about
```suggestion
            raise ImproperlyConfiguredException("No callable has been registered for this handler")
```
?

litestar/handlers/base.py
```diff
@@ -222,6 +270,10 @@ def ownership_layers(self) -> list[Self | Controller | Router]:
 
         return list(reversed(layers))
 
+    @property
+    def app(self) -> Litestar:
```
provinzkraut, line 274: Is this being used anywhere?

litestar/handlers/base.py
```diff
@@ -287,14 +339,31 @@ def resolve_dependencies(self) -> dict[str, Provide]:
...
+                            data_dto=self.resolve_data_dto(),
+                            type_decoders=self.resolve_type_decoders(),
+                        )
+                    if provider.sync_to_thread and not is_async_callable(provider.dependency.value):
+                        provider.dependency.value = async_partial(provider.dependency.value)
+                        provider.has_sync_callable = False
+                    else:
+                        provider.has_sync_callable = not is_async_callable(provider.dependency.value)
+
+                    self._resolved_dependencies[key] = provider
```
provinzkraut, line 366: This should not be a side effect of resolving layered dependencies.

provinzkraut, line 366: Why is this happening in here? Isn't this already being handled within `Provide`?

litestar/handlers/base.py
```diff
@@ -345,76 +414,72 @@ def resolve_signature_namespace(self) -> dict[str, Any]:
...
+            ):
+                data_dto = plugins_for_data_type[0].create_dto_for_type(self.parsed_data_field)
+            else:
+                data_dto = None
+
+            if self.parsed_data_field and data_dto:
+                data_dto.create_for_field_definition(
+                    field_definition=self.parsed_data_field, handler_id=self.handler_id
+                )
+
```
provinzkraut, line 446: This should happen somewhere else within a separate method. This method is supposed to resolve DTOs from the ownership layers. Currently it does:

- Resolve existing DTOs from ownership layers
- Handle serialization plugins
- Create a new DTO if needed

An appropriate name would therefore be `resolve_and_create_data_dto_and_handle_serialization_plugins`, which is a bit lengthy in my opinion :upside_down_face:

litestar/handlers/base.py
```diff
@@ -345,76 +414,72 @@ def resolve_signature_namespace(self) -> dict[str, Any]:
...
-    def _init_handler_dtos(self) -> None:
-        """Initialize the data and return DTOs for the handler."""
-        if (dto := self.resolve_dto()) and (data_parameter := self.parsed_fn_signature.parameters.get("data")):
-            dto.on_registration(
-                HandlerContext(
-                    dto_for="data",
-                    handler_id=str(self),
-                    field_definition=data_parameter,
-                    request_encoding_type=infer_request_encoding_from_field_definition(data_parameter),
+            if return_dtos := cast(
```
provinzkraut, line 460: See comment above

litestar/handlers/base.py
```diff
@@ -431,76 +496,19 @@ def _validate_dependency_is_unique(dependencies: dict[str, Provide], key: str, p
                     f"If you wish to override a provider, it must have the same key."
                 )
 
-    def on_registration(self, app: Litestar) -> None:
```
provinzkraut, line 434: Why this breaking change?

litestar/handlers/websocket_handlers/_utils.py
provinzkraut, line 1: I think most of these changes don't belong into this PR as they don't directly have anything to do with the DTO related things, but are simply a refactoring, so they're out of scope IMO.

litestar/handlers/websocket_handlers/listener.py
```diff
@@ -49,49 +42,61 @@
...
+class WebsocketListenerRouteHandler(WebsocketRouteHandler):
     """A websocket listener that automatically accepts a connection, handles disconnects,
     invokes a callback function every time new data is received and sends any data
     returned
     """
 
     __slots__ = {
-        "connection_accept_handler": "Callback to accept a WebSocket connection. By default, calls WebSocket.accept",
-        "on_accept": "Callback invoked after a WebSocket connection has been accepted",
-        "on_disconnect": "Callback invoked after a WebSocket connection has been closed",
```
provinzkraut, line 69: Why have the docstrings been removed?

litestar/typing.py
```diff
@@ -205,6 +206,9 @@ class FieldDefinition:
     name: str
     """Field name."""
 
+    def __deepcopy__(self, memo: dict[str, Any]) -> Self:
```
provinzkraut, line 209: What was the issue that necessitated this change?

litestar/typing.py
```diff
@@ -529,3 +557,14 @@ def from_parameter(cls, parameter: Parameter, fn_type_hints: dict[str, Any]) ->
             name=parameter.name,
             default=Empty if parameter.default is Signature.empty else parameter.default,
         )
+
+    def test_predicate_recursively(self, predicate: Callable[[FieldDefinition], bool]) -> bool:
```
provinzkraut, line 561:
```suggestion
    def matches_predicate(self, predicate: Callable[[FieldDefinition], bool]) -> bool:
```

`test_` doesn't really match the naming convention of the rest of the methods.

litestar/utils/signature.py
```diff
@@ -92,6 +93,13 @@ class ParsedSignature:
     original_signature: Signature
     """The raw signature as returned by :func:`inspect.signature`"""
 
+    def __deepcopy__(self, memo: dict[str, Any]) -> Self:
```
provinzkraut, line 96: What problem does this solve?

tests/examples/test_contrib/test_sqlalchemy/plugins/test_tutorial_example_apps.py
provinzkraut, line 1: Outside the scope of this PR

tests/examples/test_contrib/test_sqlalchemy/plugins/test_tutorial_example_apps.py
```diff
@@ -12,42 +13,34 @@
...
+    [
         full_app_no_plugins,
         full_app_with_init_plugin,
         full_app_with_plugin,
         full_app_with_serialization_plugin,
         full_app_with_session_di,
-    ]
+    ],
 )
-async def app(monkeypatch: MonkeyPatch, request: FixtureRequest) -> Litestar:
```
provinzkraut, line 28: The point of this being a fixture was to ensure we get a new engine per app per module, so the tests are properly isolated. Please revert

tests/examples/test_dto/test_tutorial.py
provinzkraut, line 1: Outside the scope of this PR

tests/unit/test_contrib/test_piccolo_orm/test_piccolo_orm_dto.py
```diff
@@ -79,7 +80,8 @@ class _Table(Table):
 
 
 def test_piccolo_dto_openapi_spec_generation() -> None:
-    app = Litestar(route_handlers=[retrieve_studio, retrieve_venues, create_concert], dto=PiccoloDTO)
+    DTOBackend._seen_model_names = set()
```
provinzkraut, line 83: If this is needed to ensure test isolation, it should be a global fixture with `autouse=True`

tests/unit/test_contrib/test_prometheus.py
provinzkraut, line 1: Outside the scope of this PR

tests/unit/test_dto/test_factory/test_integration.py
```diff
@@ -93,7 +95,7 @@ class Spam:
 
 
 @dataclass
-class Foo:
+class Fzop:
```
provinzkraut, line 98: Why?

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
provinzkraut, line 1: Outside the scope of this PR

tests/unit/test_handlers/test_websocket_handlers/test_validations.py
provinzkraut, line 1: Outside the scope of this PR

provinzkraut (changes requested):

litestar/enums.py
```diff
@@ -30,8 +30,6 @@ class MediaType(str, Enum):
     MESSAGEPACK = "application/x-msgpack"
     HTML = "text/html"
     TEXT = "text/plain"
-    CSS = "text/css"
-    XML = "application/xml"
```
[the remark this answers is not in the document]
Goldziher: This was introduced by someone to support in testing. 

This is not a breaking change - rather whoever added this added a problem. We do not support these Media types and they shouldn't have been declared there to begin with. 

Whoever merged this did a bad job. I can spin this out to a different pr.

litestar/handlers/base.py
Goldziher: Fair point.

litestar/handlers/base.py
Goldziher: Aight

litestar/handlers/base.py
Goldziher: Yes, I added it to use it. It's used in the resolve DTO properties

litestar/handlers/base.py
Goldziher: Why?

litestar/enums.py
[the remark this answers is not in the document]
provinzkraut: 1. The someone who added this was me
2. The someone who approved this was you
3. It was not to support things in testing, but templates which can indeed have CSS and XML as a media type. There's not really anything for us to support here, apart from the convenience. The tests case you removed further down below was explicitly added to test for the bug regarding these media types.

https://github.com/litestar-org/litestar/pull/970

litestar/handlers/base.py
Goldziher: Yes, it's better to move the logic into probably provide. And no, it's not happening there but happened in the handler.

provinzkraut: Side effects like these from properties are unexpected and obscure what's actually going on. I'm not sure what the motivation was to pull these into the resolve methods instead of having explicit methods for this stuff?

It strikes me as more of a personal style preference.

litestar/handlers/base.py
Goldziher: I disagree. It should resolve the DTOs for the handler. This logic means actually setting the providers for the handler, doing it once and caching.

litestar/handlers/base.py
Goldziher: DTOs are instantiated using a class method after resolution, and are cached. 

You are right that it can be abstracted into a sub method, but it still needs to be called at this point in the control flow and exactly once. 

I prefer having the logic in one method rather than spreading it into two, with one calling the other.

litestar/enums.py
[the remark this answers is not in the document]
Goldziher: Ok, I'll revert

litestar/handlers/base.py
Goldziher: It's not a breaking change. This is internal API. It's not meant to be user facing at all. 

Anyhow, we do not need to inject the app during registration at all, and the stuff there was hacky.

litestar/handlers/websocket_handlers/_utils.py
Goldziher: They are not. The listener context stuff was causing issues with the DTO stuff, and then everything there was overly complex. I'm either case, they are there now - so please stop with this out of scope stuff.

litestar/typing.py
Goldziher: See the copy on controller. The listener logic with the replacement of the signature did not work with deep copy. This caused issues for owner registration.

tests/examples/test_dto/test_tutorial.py
Goldziher: What?

litestar/utils/signature.py
Goldziher: Same

litestar/handlers/base.py
provinzkraut: It is a breaking change. This API is public and documented.

https://docs.litestar.dev/2/reference/handlers.html#litestar.handlers.BaseRouteHandler.on_registration

If we the changes you made make usage of this parameter obsolete than you can simply not use it. No need to possibly break things for users.

Goldziher: ok

litestar/handlers/websocket_handlers/listener.py
Goldziher: lemme see.

litestar/typing.py
Goldziher: `matches_predicate_recursively`? i want to make it clear there is recursion involved here.

tests/unit/test_handlers/test_websocket_handlers/test_validations.py
Goldziher: reverted

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
Goldziher: reverted

tests/examples/test_contrib/test_sqlalchemy/plugins/test_tutorial_example_apps.py
Goldziher: done

litestar/pagination.py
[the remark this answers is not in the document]
Goldziher: reverted

litestar/handlers/websocket_handlers/listener.py
Goldziher: fixed

docs/usage/dto/2-dto-interface.rst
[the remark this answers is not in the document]
Goldziher: ok, will revert

litestar/handlers/websocket_handlers/listener.py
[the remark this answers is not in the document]
Goldziher: Perhaps, but they are there now. The reason I did them was because it was very hard for me to follow the logic when I was fixing stuff with the DTO changes, and things were breaking in cascade. The end result is much simpler to understand IMO. Also - I cant really separate them out now without a lot of work, which I don't think is needed.

litestar/dto/_backend.py
```diff
@@ -102,39 +100,14 @@ def __init__(
     def parse_model(
         self, model_type: Any, exclude: AbstractSet[str], include: AbstractSet[str], nested_depth: int = 0
     ) -> FieldDefinitionsType:
-        """Reduce :attr:`model_type` to :class:`FieldDefinitionsType`.
```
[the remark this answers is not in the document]
Goldziher: because we no longer subclass this class, not do we support subclassing it. I will return it there, rewritten.

litestar/dto/_backend.py
```diff
@@ -696,10 +662,7 @@ def _create_msgspec_field(field_definition: TransferDTOFieldDefinition) -> Any:
 
 def _create_struct_for_field_definitions(model_name: str, field_definitions: FieldDefinitionsType) -> type[Struct]:
     struct_fields: list[tuple[str, type] | tuple[str, type, type]] = []
-    for field_definition in field_definitions:
-        if field_definition.is_excluded:
-            continue
-
+    for field_definition in (f for f in field_definitions if not f.is_excluded):
```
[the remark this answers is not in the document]
Goldziher: fixed

litestar/dto/base_dto.py
```diff
@@ -84,18 +84,15 @@ def __class_getitem__(cls, annotation: Any) -> type[Self]:
 
         return type(f"{cls.__name__}[{annotation}]", (cls,), cls_dict)
 
-    def builtins_to_data_type(self, builtins: Any) -> Any:
+    def decode(self, value: bytes | dict[str, Any]) -> Any:
```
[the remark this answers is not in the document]
Goldziher: ok

litestar/handlers/base.py
Goldziher: done

litestar/handlers/base.py
Goldziher: same

litestar/handlers/base.py
Goldziher: done

cofin (approved):
Sorry for the long review time on this.  It's a big one, but other than the current comments, it LGTM.

JacobCoffee (approved):

JacobCoffee:
Sorry for the delay

provinzkraut (approved):

PR #2093, opened 2023-07-30
test(sqla repo): unify UUID and BigInt tests

provinzkraut:
Unify the tests for UUID and BigInt model repositories by introducing fixture hell to the test suite (=

The main motivator behind this was to make the tests easier to reason about (if you exclude the fixtures) and to increase coverage (which gets easier if you don't have to write all tests twice).

@cofin Let me know what you think.

<hr>

The only issue here is that mypy is not happy *at all* with this approach, but I also don't want to type everything as `Any`. I'll probably have to sit down again and fine a more sane way to do this. Open to suggestions..
@sobolevn maybe? (=


tests/unit/test_contrib/test_sqlalchemy/test_repository/conftest.py
```diff
@@ -57,7 +57,7 @@ def fx_raw_authors_uuid() -> list[dict[str, Any]]:
             "updated_at": "2023-05-11T00:00:00",
         },
         {
-            "id": "5ef29f3c-3560-4d15-ba6b-a2e5c721e4d2",
+            "id": UUID("5ef29f3c-3560-4d15-ba6b-a2e5c721e4d2"),
```
cofin, line 60: In this case, I left it stringified to verify that it would be properly cast to UUID.

cofin: Maybe it should be its own test case though?

cofin:
I'm going to have to study how you've done this, but this is really helpful for reducing duplication.

tests/unit/test_contrib/test_sqlalchemy/test_repository/conftest.py
provinzkraut: Yeah, that's what I assumed. There's a few things that need to be cleaned up still, I just wanted to see if I could get it working at all first :D
I just did this so it's easier to handle the case where we compare the PKs directly.

Should probably be its own test, yes.

tests/unit/test_contrib/test_sqlalchemy/test_repository/test_repository.py
```diff
@@ -28,30 +25,101 @@
...
 
+RepositoryPKType = Literal["uuid", "bigint"]
+AuthorModel = Type[Union[models_uuid.UUIDAuthor, models_bigint.BigIntAuthor]]
+RuleModel = Type[Union[models_uuid.UUIDRule, models_bigint.BigIntRule]]
+ModelWithFetchedValue = Type[Union[models_uuid.UUIDModelWithFetchedValue, models_bigint.BigIntModelWithFetchedValue]]
+ItemModel = Type[Union[models_uuid.UUIDItem, models_bigint.BigIntItem]]
+TagModel = Type[Union[models_uuid.UUIDTag, models_bigint.BigIntTag]]
+
+AnyAuthor = Union[models_uuid.UUIDAuthor, models_bigint.BigIntAuthor]
+AuthorRepository = SQLAlchemyAsyncRepository[AnyAuthor]
```
provinzkraut, line 42: This *should* be `SQLAlchemyAsyncRepository[AnyAuthor] | SQLAlchemySyncRepository[AnyAuthor]`, but I can't get mypy to correctly narrow the return values of `maybe_async` then. Not sure why since it does work correctly in other places.

Anyway, this isn't that big of a deal since we inherit the typing for the async/sync mostly from SQLA and type check the repository code.

tests/unit/test_contrib/test_sqlalchemy/test_repository/conftest.py
provinzkraut: The same test should be done for the BigInt model I think, where a string is cast to an int?

provinzkraut:
> I'm going to have to study how you've done this, but this is really helpful for reducing duplication.

Basically, I created a top level fixture that indicates what repo type should be used, and then put everything else into fixtures that request that one, giving back the different models and repo types.

sobolevn (commented):
Thanks for tagging! Looks like a good work on unifing these two tests :)

provinzkraut:
> Looks like mypy is happy now :)

It is, but only because I'm lying to it :grimacing: 

You'll notice that the params in all the functions receiving a repository (e.g. `author_repository`) are typed as `SQLAlchemyAsyncRepository` only, where it should be `SQLAlchemyAsyncRepository | SQLAlchemySyncRepository` (since that what they actually receive).

tests/unit/test_contrib/test_sqlalchemy/test_repository/test_repository.py
```diff
@@ -28,30 +25,111 @@
...
+    if repository_pk_type == "uuid":
+        return models_uuid.UUIDBook
+    return models_bigint.BigIntBook
+
+
+@pytest.fixture()
+def new_pk_id(repository_pk_type: RepositoryPKType) -> Any:
+    if repository_pk_type == "uuid":
+        return uuid4()
+    return 10
```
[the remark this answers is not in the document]
provinzkraut: Oh, thanks for the reminder. I meant to replace that UUID with a constant value as well.

I guess you suggested a generator so we can generate more values as needed?
The reason for not doing it like this is that the tests never actually require more than one unused ID so this was "the easy" way. I guess the correct way would be to have a generator producing UUIDs, and for ints to generate a random one and check if it's in the range of known int PKs but.. I just thought it's overkill :P

tests/unit/test_contrib/test_sqlalchemy/test_repository/conftest.py
provinzkraut: Okay, looks like I need to be more selective about this. There are a few failures which I *think* are correct, since we don't explicitly handle those cases (e.g. passing in a `str` as a PK a bigint model). 

I guess the question is if we *do* want to handle those cases, or if we leave it up to the user to provide a PK value with an appropriate type?

tests/unit/test_contrib/test_sqlalchemy/test_repository/test_repository.py
```diff
@@ -315,15 +448,21 @@ async def test_repo_created_updated(author_repo: AuthorAsyncRepository) -> None:
     assert author.updated_at is not None
     original_update_dt = author.updated_at
 
-    author.books.append(UUIDBook(title="Testing"))
+    # looks odd, but we want to get correct type checking here
+    if repository_pk_type == "uuid":
+        author = cast(models_uuid.UUIDAuthor, author)
+        book_model = cast("type[models_uuid.UUIDBook]", book_model)
+        author.books.append(book_model(title="Testing"))
+    else:
```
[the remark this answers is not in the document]
provinzkraut: Oh what. Since when does that exist? I missed that completely!

tests/unit/test_contrib/test_sqlalchemy/test_repository/conftest.py
provinzkraut: Shall we spin that off into a separate PR then?

provinzkraut: As for the `str` > `int` conversion: If we simply handle this as a method on the repo (e.g. some method that depends on the type declared for the id column), then supporting this alongside UUIDs would basically be zero effort, so unless there's a reason not do to this, I think we might as well support it. Wdyt?

cofin: I like this idea

cofin:
Nice work on this!  LGTM

PR #2095, opened 2023-07-30
feat(sqlalchemy): extend SQLAlchemy Plugin to include Alembic

cofin:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-


litestar/contrib/sqlalchemy/plugins/init/plugin.py
```diff
@@ -2,19 +2,22 @@
 
 from typing import TYPE_CHECKING
 
+from litestar.contrib.sqlalchemy.cli import database_group
```
provinzkraut, line 5: You'll need to put this import inside the `on_cli_init` method, else users will get an `ImportError` when they don't have the CLI dependencies installed

litestar/contrib/sqlalchemy/alembic/commands.py
provinzkraut, line 1: What is the purpose of these commands being async?
They don't seem to be doing anything asynchronous

cofin: I'm not sure exactly.  My original intent was to try and create a single set of functions that work for sync and async.  I'll remove the async if there's no reason to have it.

provinzkraut: I've just never seen async migration functions :shrug: 

I guess it makes some sort of sense if you're calling them from an async context, but then you'd have to wrap the alembic commands themselves in something like `anyio.to_thread.run_sync` to make them non-blocking.

provinzkraut:
Maybe stamp would be another good command to add? I find that quite useful for development sometimes.

litestar/contrib/sqlalchemy/alembic/commands.py
```diff
@@ -0,0 +1,255 @@
...
+    plugin = app.plugins.get(SQLAlchemyInitPlugin)
+    alembic_cfg = get_alembic_command_config(config=plugin._config)
+    migration_command.history(
+        config=alembic_cfg, rev_range=rev_range, verbose=verbose, indicate_current=indicate_current
+    )
+
+
+def merge(
+    app: Litestar,
+    revisions: str,
```
provinzkraut, line 157: Is this correct? Shouldn't this be `tuple[str, str]` or something like that?

provinzkraut:
> Maybe stamp would be another good command to add? I find that quite useful for development sometimes.

Never mind, it's already in there (=

litestar/contrib/sqlalchemy/alembic/commands.py
cofin: I'll take a look at this one.  I started by making wrappers around each alembic command and I'm slowly adding all the commands to the CLI.

cofin:
@provinzkraut I may need your input on this one.  I need a way to inject the database URL into the `env.py`.  The method you see that i've implemented results in the value of the DB URL being `None`.  This is one of the last pieces to solve to make it run.

provinzkraut:
> @provinzkraut I may need your input on this one. I need a way to inject the database URL into the `env.py`. The method you see that i've implemented results in the value of the DB URL being `None`. This is one of the last pieces to solve to make it run.

Do you mean this?
https://github.com/litestar-org/litestar/blob/94afae2fc4eb9dc54c60f6fbfd8453b63976a237/litestar/contrib/sqlalchemy/alembic/commands.py#L76

cofin:
> 

Yep.  This is the section that doesn't seem to be working.  I also tried passing the engine itself and that didn't work.

PR #2096, opened 2023-07-30
feat: plugin registry

provinzkraut:
Add a "plugin registry", that allows to retrieve plugins by their type.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -5,10 +5,10 @@ import logging
 import os
 from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager, suppress
 from datetime import date, datetime, time, timedelta
-from functools import lru_cache, partial
+from functools import partial
 from itertools import chain
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, AsyncGenerator, Callable, Mapping, Sequence, TypedDict, TypeVar, cast
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Callable, Mapping, Sequence, TypedDict, cast
 
 from litestar._asgi import ASGIRouter
 from litestar._asgi.utils import get_route_handlers, wrap_in_exception_handler
@@ -34,6 +34,7 @@ from litestar.plugins import (
     InitPluginProtocol,
     OpenAPISchemaPluginProtocol,
     PluginProtocol,
+    PluginRegistry,
     SerializationPluginProtocol,
 )
 from litestar.router import Router
@@ -42,7 +43,7 @@ from litestar.static_files.base import StaticFiles
 from litestar.stores.registry import StoreRegistry
 from litestar.types import Empty, TypeDecodersSequence
 from litestar.types.internal_types import PathParameterDefinition
-from litestar.utils import AsyncCallable, join_paths, unique
+from litestar.utils import AsyncCallable, deprecated, join_paths, unique
 from litestar.utils.dataclass import extract_dataclass_items
 from litestar.utils.predicates import is_async_callable
 from litestar.utils.warnings import warn_pdb_on_exception
@@ -96,8 +97,6 @@ if TYPE_CHECKING:
     from litestar.types.callable_types import LifespanHook
 
 
-PluginT = TypeVar("PluginT", bound=PluginProtocol)
-
 __all__ = ("HandlerIndex", "Litestar", "DEFAULT_OPENAPI_CONFIG")
 
 DEFAULT_OPENAPI_CONFIG = OpenAPIConfig(title="Litestar API", version="1.0.0")
@@ -134,7 +133,7 @@ class Litestar(Router):
         "_lifespan_managers",
         "_debug",
         "_openapi_schema",
-        "_plugins",
+        "plugins",
         "after_exception",
         "allowed_hosts",
         "asgi_handler",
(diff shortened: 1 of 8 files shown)


litestar/app.py
```diff
@@ -450,6 +451,33 @@ def __init__(
             config.stores if isinstance(config.stores, StoreRegistry) else StoreRegistry(config.stores)
         )
 
+    @property
+    def cli_plugins(self) -> list[CLIPluginProtocol]:
+        return self.get_plugins_of_type(CLIPluginProtocol)
```
provinzkraut, line 456: I have no idea why the typing doesn't work here.. I feel like I made some very obvious mistake that I'm missing completely..

litestar/app.py
```diff
@@ -450,6 +451,33 @@ def __init__(
...
+        """
+        try:
+            return cast(PluginT, self._plugins[type_])
+        except KeyError as e:
+            raise KeyError(f"No plugin of type {type_!r} registered") from e
+
+    @lru_cache  # noqa: B019
+    def get_plugins_of_type(self, type_: type[PluginT]) -> list[PluginT]:
+        """Return all registered plugins that are subclasses of ``type_``."""
+        return [p for p in self._plugins.values() if isinstance(p, type_)]
```
provinzkraut, line 479: Ideally this should be an immutable type like a frozenset, but it's not for reasons of backwards compatibility.

cofin (approved):

litestar/app.py
gsakkis: I don't have a complete and coherent explanation either but I think it comes down to the annotation of `self._plugins`. `Mapping[type[PluginProtocol], PluginProtocol]` is not strict enough based on the intended usage: it allows mapping a type `A` to an instance of a different type `B` as long as both `A` and `B` are PluginProtocols. I don't know if the desired stricter annotation is currently expressible.

Other observations:
- The properties give a different error based on the presence or absence of `lru_cache`:
   - With `lru_cache`: `Incompatible return value type (got "list[PluginT]", expected "list[SerializationPluginProtocol]")  [return-value]`
   - Without `lru_cache`: `Only concrete class can be given where "type[SerializationPluginProtocol]" is expected  [type-abstract]`
- Changing the signature of `get_plugins_of_type` from `(self, type_: type[PluginT]) -> list[PluginT]` to `(self, type_: type[PluginProtocol]) -> list[PluginProtocol]` is more consistent with the `self._plugins` annotation. In this case the error is the same with and without `lru_cache`: `Incompatible return value type (got "list[SerializationPluginProtocol | InitPluginProtocol | OpenAPISchemaPluginProtocol | CLIPluginProtocol]", expected "list[SerializationPluginProtocol]")  [return-value]` 

As for what to do about it I can think of two options:
1)  
7u8dh4

2) Abandon the generic attempt and inline `get_plugins_of_type` on every property:
```py
    @cached_property
    def cli_plugins(self) -> list[CLIPluginProtocol]:
        return [p for p in self._plugins.values() if isinstance(p, CLIPluginProtocol)]

    @cached_property
    def openapi_schema_plugins(self) -> list[OpenAPISchemaPluginProtocol]:
        return [p for p in self._plugins.values() if isinstance(p, OpenAPISchemaPluginProtocol)]

    @property
    def serialization_plugins(self) -> list[SerializationPluginProtocol]:
        return [p for p in self._plugins.values() if isinstance(p, SerializationPluginProtocol)]
```

JacobCoffee (approved):

litestar/app.py
provinzkraut: Thanks for the input @gsakkis!

So, I did some more digging and found something curious. Removing `@lru_cache` reveals a different error:

```python
    @property
    def cli_plugins(self) -> list[CLIPluginProtocol]:
        return self.get_plugins_of_type(CLIPluginProtocol) #  error: Only concrete class can be given where "type[CLIPluginProtocol]" is expected  [type-abstract]
```

I'm now wondering whether this is an issue with mypy, `@lru_cache`'s typing, or my code..
Probably a mix? :upside_down_face: 

Interestingly enough, pyright doesn't report issues anymore without the `@lru_cache`.

<hr>

Regarding your suggestions: They would work, but kinda defeat the purpose. The idea I had was that users could use this function to retrieve these plugins in a type safe manner. The properties are just for backwards compatibility.

If it turns out this simply can't be typed properly (which I'm not yet sure, since it *looks* like this should work!), I'd rather just remove the method in favour of exposing the `plugins` mapping directly, so users can iterate over it themselves (and get correct typing).

litestar/plugins.py
```diff
@@ -161,3 +159,41 @@ def to_openapi_schema(self, annotation: Any, schema_creator: SchemaCreator, dto_
     OpenAPISchemaPluginProtocol,
     CLIPluginProtocol,
 ]
+
+PluginT = TypeVar("PluginT", bound=PluginProtocol)
+
+
+class PluginRegistry:
```
gsakkis, line 166: This can extend the [Mapping ABC](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) to get a few more standard mapping methods for free.

litestar/plugins.py
```diff
@@ -161,3 +159,41 @@ def to_openapi_schema(self, annotation: Any, schema_creator: SchemaCreator, dto_
...
+    __slots__ = {
+        "init": "Plugins that implement the InitPluginProtocol",
```
gsakkis, line 168: I haven't seen `__slots__` as a dict, is this new?

provinzkraut: It's not new, but their usage for documentation has only been added in 3.9. Before that it was "left up for future implementation" to decide what the meaning of this was iirc.

litestar/plugins.py
provinzkraut: I guess that could be done. I was thinking we leave the interface minimal for now to make it a bit more flexible for future use cases, and I didn't really see a use case for `items()`, `values()`, etc. as this is now.

gsakkis (approved):

PR #2129, opened 2023-08-06
feat(dto): optimise _transfer_instance_data

provinzkraut:


litestar/dto/_backend.py
```diff
@@ -579,34 +579,42 @@ def _transfer_instance_data(
     """
     unstructured_data = {}
 
+    should_use_serialization_name = is_data_field and not is_dto_data_type
+    nested_as_dict = destination_type is dict
+
+    # we assume that in most cases we will call these more than one so assigning them
+    # here is going to be more efficient in most cases
+    if isinstance(source_instance, Mapping):
```
Goldziher, line 587: I don't like declaring functions conditionally like this, I think this is a bad pattern.

Goldziher (changes requested):
This is not optimized, or more readable. In fact, it's a step back imo.

provinzkraut:
> This is not optimized, or more readable. In fact, it's a step back imo.

It is optimised. It's pulling out a lot of unnecessary function calls that would happen on every iteration and instead just calls them once.

litestar/dto/_backend.py
provinzkraut: It's not ideal in most situations, yes. But since this is a performance critical part, it's worth doing it like this, since it means we won't have to do multiple unnecessary `isinstance` checks on every iteration in the loop below.

This is also not done yet, as indicated by the work in progress.

Goldziher: ok

provinzkraut:
Closing this in favour of #2175

PR #2130, opened 2023-08-06
fix: moved application/octet-stream from contentEncoding to contentMediaType

Goldziher:
This PR fixes a bug in the response schema where `contentEncoding` had a default of `application/octet-stream`, which actually should be the default of `contentMediaType`.


cofin (approved):
LGTM

PR #2145, opened 2023-08-10
fix: add default receive and send parameter values for HTMXRequest

tompin82:
I noticed there seemed to be an issue setting the apps `request_class` to `HTMXRequest`.

The CSRF middleware doesn't initialize the `request_class` with a `send` argument, which causes the `HTMXRequest` class to fail its initialization since it expects `send` to exist.

This PR simply adds the `empty_receive` and `empty_send` functions as default vaules, which is consistent with the signature in the request class it derives from.


cofin (approved):
LGTM

cofin:
@all-contirbutors add @tompin82 for code and bugs

PR #2152, opened 2023-08-12
feat: Ensures deterministic ordering of handlers

gsakkis:
Currently the order of controller handlers (e.g. as listed in the generated OpenAPI schema) is not only random but unstable between different runs unless [PYTHONHASHSEED](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED) is fixed (which is typically not).

This PR ensures deterministic (even if somewhat arbitrary) order.

Potential future feature: Make handler ordering configurable by the user.

### DIFF
--- a/litestar/controller.py
+++ b/litestar/controller.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections import defaultdict
 from copy import deepcopy
 from functools import partial
+from operator import attrgetter
 from typing import TYPE_CHECKING, Any, Mapping, cast
 
 from litestar._layers.utils import narrow_response_cookies, narrow_response_headers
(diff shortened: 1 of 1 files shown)


provinzkraut:
@gsakkis can you retarget this to the `v2.1` branch?

gsakkis:
@provinzkraut I can but why? This is (admittedly a small but annoying) bug, why defer the fix to 2.1?

provinzkraut:
I would argue that this is more of a feature than a bug fix. Everything is working as expected, it's just that the behaviour isn't desirable.

Unless I'm mistaken and this is a regression / does not work as expected?

Goldziher:
> I would argue that this is more of a feature than a bug fix. Everything is working as expected, it's just that the behaviour isn't desirable.
> 
> Unless I'm mistaken and this is a regression / does not work as expected?

I wouldn't consider this a feature tbh. It's a change in implementation from something that can be considered a bug. I don't think if needs to wait for a minor version since it can't break anything.

Goldziher:
@gsakkis also please reformat the PR title to follow conventional commits

PR #2154, opened 2023-08-12
fix(routing): support serving static files from a route used by a route handler

Goldziher:
This PR fixes an issue with static routing where static file serving failed if a route handler was defined for the base path of the static file serving.

Additionally two fixes for tests are added.

### DIFF
--- a/tests/unit/test_openapi/test_parameters.py
+++ b/tests/unit/test_openapi/test_parameters.py
@@ -104,15 +104,15 @@ def test_create_parameters(person_controller: Type[Controller]) -> None:
                 "items": {
                     "type": "string",
                     "enum": ["M", "F", "O", "A"],
-                    "examples": [{"description": "Example  value", "value": "A"}],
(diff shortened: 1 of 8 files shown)


tests/unit/test_openapi/test_parameters.py
provinzkraut, line 1: What lead to the change here?

provinzkraut (approved):
Nice! :heart:

tests/unit/test_openapi/test_parameters.py
Goldziher: they were failing locally. Reason is an update of polyfactory which included a change to the random seed.

provinzkraut: So the fix is to bump polyfactory? If yes, could you include that in this PR?

Goldziher: i did... dunno why its failing here, its working locally.

PR #2160, opened 2023-08-13
fix(auth-middleware): add configuration for HTTP methods to skip when authenticating

Goldziher:
This PR fixes an issue where OPTIONS requests got 401. This issue emerged due to a fix for the OPTIONS handler registration - which surfaced it. The solution was to add configuration for the HTTP methods to exclude from authentication and give defaults to the JWT auth middleware that mirrors the previous (buggy) behavior.

### DIFF
--- a/litestar/contrib/jwt/jwt_auth.py
+++ b/litestar/contrib/jwt/jwt_auth.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from dataclasses import asdict, dataclass, field
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING, Any, Callable, Generic, Iterable, Literal, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Generic, Iterable, Literal, Sequence, TypeVar, cast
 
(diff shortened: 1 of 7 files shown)


provinzkraut:
> and give defaults to the JWT auth middleware that mirrors the previous (buggy) behavior.

Just so I understand: The previous behaviour of the JWT auth was buggy, then it was "fixed", which broke things, and this PR now adds an option which restores the buggy behaviour?

Goldziher:
> > and give defaults to the JWT auth middleware that mirrors the previous (buggy) behavior.
> 
> Just so I understand: The previous behaviour of the JWT auth was buggy, then it was "fixed", which broke things, and this PR now adds an option which restores the buggy behaviour?

not exactly. the options handlers were not registered correctly, so the auth middleware skipped them. The skipping them part was not by design, but it seems to be the desired behavior. The fix for the registration suddenly caused the auth middleware to authenticate these handlers, hence the fix here.

provinzkraut (approved):

v3ss0n:
Great!!

Issue #1311, opened 2023-03-11
Enhancement: Generic controllers


ghost:
**What's the feature you'd like to ask for.**
First off, lovely framework. I would like for it to be possible to "inherit routes" when I use `starlite.Controller`, in code speak this roughly translates to

A "Base Controller" that can have common routes such as 

- Get by ID
- Get by Name
- Delete by ID
- Get All

All these would return models (in my case, they would be SQLAlchemy objects)

```python
T = TypeVar('T', bound=models.Base)

class BaseController(Controller, Generic[T]):
    model = models.Base

    @get('{resource_id:uuid}')
    async def get_by_id(
        self, resource_id: UUID, session: AsyncSession
    ) -> T | None:
        return await session.get(self.model, resource_id)
```

Any inherited class can then use this like so

```python
class Foo(BaseController[type[models.Foo]]):
    path = 'foo'
    tags = [path]
    model = models.Foo

class Bar(BaseController[type[models.Bar]]):
    path = 'Bar'
    tags = [path]
    model = models.Bar
```

**Expected Behavior**

I should now be able to do `/foo/{resource_id}` and `/bar/{resource_id}` and of course, any other routes in the base class should be added. 

**Additional context**

- Right now the code example I've shown results in an `ImproperlyConfiguredException`

```
starlite.exceptions.http_exceptions.ImproperlyConfiguredException: 500: Parameter '' with type '~T' could not be mapped to an Open API type. This can occur if a user-defined generic type is resolved as a parameter. If '' should not be documented as a parameter, annotate it using the `Dependency` function, e.g., `: ... = Dependency(...)`.
```

- OpenAPI integration would be a plus.

peterschutt:
I'm thinking that we have an alternate controller class called `GenericController` that accepts a single type argument. In `__class_getattr__` we take that type arg, create a subclass of `GenericController` that has the type bound to it, and return that subclass.

Internally, Starlite would have to know to substitute out any type vars in the handler signatures for the referenced type as it models the handler function signatures.

Lets say the generic base looked like this:

```py
class GenericController(Controller, Generic[T]):
    _data_type: type[T]
    def __class_getitem__(cls, data_type: type[T]) -> Self:
        return type(f"Controller[{data_type.__name__}]", (cls,), {"_data_type": data_type})
```
Then, you'd want to be able to subclass it to add generic methods.

```py
class MyGenericController(GenericController[T], Generic[T]):
    @get()
[...]

PR #2162, opened 2023-08-13
feat: add generic controller

Goldziher:
This PR will add a generic controller. Its still an initial draft. 

Related to #1320


provinzkraut:
@Goldziher I think you'll need to rebase this from `v2.1`

pomarec:
I'm trying to make something like @iamrinshibuya describes in the first comment of https://github.com/litestar-org/litestar/issues/1311

I get a `litestar.exceptions.base_exceptions.SerializationException: Unsupported type: `. I guess the return type of `get_by_id` (the actual type behind `T`) is not "registered to match with dto".

Does it make sense ? Do you have any clue on how to make this work ?

iamrinshibuya:
> I'm trying to make something like @iamrinshibuya describes in the first comment of #1311
> 
> I get a `litestar.exceptions.base_exceptions.SerializationException: Unsupported type: `. I guess the return type of `get_by_id` (the actual type behind `T`) is not "registered to match with dto".
> 
> Does it make sense ? Do you have any clue on how to make this work ?

Hello @pomarec I'd suggest not regarding my code as anywhere close to ideal. I simply created something that felt appropriate to me at that time.

pomarec:
I just found out that the issues comes when using a sqlalchemy model.
To makes things clearer, here are two tests, first passing, second failing.

```python
def test_generic_sub_controller() -> None:
    T = TypeVar("T")

    class GenericSubController(GenericController[T], Generic[T]):
        model_type = T

        @get("/{id:int}")
        def get_handler(self, id: int) -> T:
            return VanillaDataClassPersonFactory.build(id=id)

    class GenericPersonController(GenericSubController[VanillaDataClassPerson]):
        model_type = VanillaDataClassPerson
        path = "/"

    with create_test_client(GenericPersonController) as client:
        response = client.get("/1")
        assert response.json()

def test_generic_sub_controller_sqlalchemy() -> None:
    T = TypeVar("T")

    class GenericSubController(GenericController[T], Generic[T]):
        model_type = T

        @get("/{id:int}")
        def get_handler(self, id: int) -> T:
            return self.model_type(id=id)

    class Base(DeclarativeBase):
        pass

    class Person(Base):
        __tablename__ = "user"
        id: Mapped[int] = mapped_column(primary_key=True)

    PersonDTO = SQLAlchemyDTO[Person]

    class GenericPersonController(GenericSubController[Person]):
        model_type = Person
        dto = PersonDTO
        path = "/"

    with create_test_client(GenericPersonController) as client:
        response = client.get("/1")
        assert response.json()
```

Error for `test_generic_sub_controller_sqlalchemy ` : `litestar.exceptions.base_exceptions.SerializationException: Unsupported type: <class 'tests.unit.test_controller.test_generic_controller.test_generic_sub_controller_sqlalchemy.<locals>.Person'>`

pomarec:
Adding this to `GenericController` passes the test
```python
    def __init__(self, owner: Router) -> None:
        super().__init__(owner=owner)
        self.signature_namespace[T.__name__] = self.model_type
```
But my feeling is that it's just a dirty workaround. 
Looking at `GenericController.get_parameter_annotation()`, I feel that it should be used somehow in `BaseRouteHandler.parsed_fn_signature()`

What do you guys think ?

Goldziher:
The place where I'd add this is in `__init_subclass__`. 

Btw this PR is just a starting point, it's pretty far from what it will need to be. I think it needs to be something like DRF mixins basically.

pomarec:
Disclaimer : i'm not familiar with python typing system nor litestar codebase

I took some time to dig the situation, found out the generic types are not infered in the signature of a function.
I tried to hack a draft of resolution :

> https://github.com/akago-dev/litestar/commit/50e0358a5a86fc275148ef80ab7255c42a183aad

This leads to a usage for CRUD generic controller :

> https://github.com/akago-dev/litestar/commit/19ebaa1e8269b8d995d9c7280c62480aaee7c86a

If you guys think its the right direction, please reach me on discord with some guidance as i'm a bit lost or take it from there :)
If the direction is not right, please say it aswell !

cheers

Goldziher:
Ok, I set this to have a clearer direction. Please give you input (also @litestar-org/maintainers and @litestar-org/members please). 

The idea is to use mixins - which are generic. You can see the example in the test. Its still very small. We will then offer an SQLA generic version of these as well, which will already include some ready to use logic (or repository logic).

PR #2175, opened 2023-08-17
feat(DTO): Improve performance with code generation - WIP

provinzkraut:
**Work in progress**

Drastically improve DTO overhead by generating otimised transfer functions ahead of time.

This needs to be cleaned up a fair bit so approach with caution (=

<hr>

Preliminary benchmarking results:

- ~500% performance increase when parsing from a raw connection (simple)
- ~200% performance increase when parsing from a raw connection (collection)
- ~500% performance increase when encoding structured data into bytes
- ~300% performance increase when populating from builtins
- ~250% performance increase when populating from builtins (collection)

<hr>

The codegen is implemented with the `DTOCodegenBackend` and will not be enabled by default. It can be switched on by using the `experimental_codegen_backend` feature flag on the `DTOConfig`.

This allows us to experiment with it while still having the opportunity to roll it back at any point should it not work out as expected or we decide it's too hard to maintain.

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -336,11 +336,7 @@ class DTOBackend:
                 is_data_field=self.is_data_field,
                 override_serialization_name=self.override_serialization_name,
             )
-            setattr(
-                data,
-                self.wrapper_attribute_name,
-                wrapped_transfer,
-            )
+            setattr(data, self.wrapper_attribute_name, wrapped_transfer)
             return cast("LitestarEncodableType", data)
 
         return cast(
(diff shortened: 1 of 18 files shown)


Goldziher:
I like the idea of the experimental flag. Maybe we can add it top level on the app though. Maybe a sequence of literals?

provinzkraut:
My only concern with that would be that it wouldn't allow for fine-grained control (i.e. enabling an experimental feature only for part of an application).

Goldziher:
> My only concern with that would be that it wouldn't allow for fine-grained control (i.e. enabling an experimental feature only for part of an application).

I'd do `experimental_features: Sequence[Literal["x", "y"]] | bool`

PR #2179, opened 2023-08-19
feat(sqlalchemy): use `lambda_stmt` in the repository

cofin:
### Pull Request Checklist
This PR converts the repository to use `[lambda_stmt](https://docs.sqlalchemy.org/en/20/core/connections.html#engine-lambda-caching)` instead of the normal `select`

### Description

-

### Close Issue(s)

-


litestar/contrib/sqlalchemy/base.py
```diff
@@ -42,7 +43,7 @@
 UUIDBaseT = TypeVar("UUIDBaseT", bound="UUIDBase")
 BigIntBaseT = TypeVar("BigIntBaseT", bound="BigIntBase")
 
-convention = {
+convention: NamingSchemaParameter = {
```
provinzkraut, line 46: Oh, that's good fix. Didn't know about this secret type alias when I added the `type: ignore` :grimacing:

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -64,7 +75,15 @@ def __init__(
...
-        self.statement = statement if statement is not None else select(self.model_type)
+        if isinstance(statement, Select):
+            self.statement = lambda_stmt(lambda: statement)
+        elif statement is None:
+            statement = select(self.model_type)
+            self.statement = lambda_stmt(lambda: statement)
+        else:
+            self.statement = statement
+
+        self.statement = self._to_lambda_stmt(statement)
```
provinzkraut, line 86: Couldn't this be simplified?

```suggestion
        self.statement = self._to_lambda_stmt(statement if statement is not None else select(self.model_type))
```

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -245,8 +271,13 @@ async def get(  # type: ignore[override]
         """
         with wrap_sqlalchemy_exception():
             id_attribute = id_attribute if id_attribute is not None else self.id_attribute
-            statement = statement if statement is not None else self.statement
-            statement = self._filter_select_by_kwargs(statement=statement, kwargs=[(id_attribute, item_id)])
+            statement = self._to_lambda_stmt(statement)
+            statement = self._filter_select_by_kwargs(
+                statement=statement,
+                kwargs={
+                    id_attribute: item_id,  # pyright: ignore[reportGeneralTypeIssues]
```
provinzkraut, line 278: pyright is actually right here. `id_attribute` is not guaranteed to be hashable. That's why I added support for passing this as an iterable of tuples.

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -219,11 +238,18 @@ async def exists(self, **kwargs: Any) -> bool:
         existing = await self.count(**kwargs)
         return existing > 0
 
+    def _to_lambda_stmt(
```
provinzkraut, line 241:
```suggestion
    def _get_base_stmt(
```

would be more appropriate I think. The lambda statement conversion is more of an implementation detail and the current name hides the fact that this also optionally returns the class instance base statement.

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -300,9 +331,9 @@ async def get_one_or_none(
             The retrieved instance or None
         """
         with wrap_sqlalchemy_exception():
-            statement = statement if statement is not None else self.statement
+            statement = self._to_lambda_stmt(statement)
             statement = self._filter_select_by_kwargs(statement=statement, kwargs=kwargs)
-            instance = (await self._execute(statement)).scalar_one_or_none()
+            instance = cast("Result[tuple[ModelT]]", (await self._execute(statement))).scalar_one_or_none()
```
provinzkraut, line 336: Lambda statements don't support proper type inference?

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -605,15 +635,18 @@ async def _list_and_count_basic(
...
-        statement = statement if statement is not None else self.statement
+        statement = self._to_lambda_stmt(statement)
         statement = self._apply_filters(*filters, statement=statement)
         statement = self._filter_select_by_kwargs(statement, kwargs)
-        count_statement = statement.with_only_columns(
-            sql_func.count(self.get_id_attribute_value(self.model_type)),
-            maintain_column_froms=True,
-        ).order_by(None)
+
+        def count_statement(statement: StatementLambdaElement) -> StatementLambdaElement:
```
provinzkraut, line 642: This should be a method instead of an inline function sicne it does not capture anything from the outer functions scope

litestar/contrib/sqlalchemy/repository/_async.py
```diff
@@ -860,57 +896,85 @@ def _apply_filters(self, *filters: FilterTypes, apply_pagination: bool = True, s
...
-            statement = statement.where(field >= on_or_after)
+            statement += lambda s: s.where(field >= on_or_after)
         return statement
 
     def _filter_select_by_kwargs(
-        self, statement: SelectT, kwargs: dict[Any, Any] | Iterable[tuple[Any, Any]]
-    ) -> SelectT:
-        for key, val in kwargs.items() if isinstance(kwargs, dict) else kwargs:
-            statement = statement.where(get_instrumented_attr(self.model_type, key) == val)  # pyright: ignore
+        self, statement: StatementLambdaElement, kwargs: dict[Any, Any]
```
provinzkraut, line 938: See comment above about the non-hashable keys we use here

provinzkraut (approved):

litestar/contrib/sqlalchemy/repository/_async.py
cofin: I didn't do this based on[ these guidelines here.](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#adding-relationships-to-mapped-classes-after-declaration)

litestar/contrib/sqlalchemy/repository/_async.py
cofin: I'll do some additional testing on this before we merge it.  I want to say it doesn't, but I tried quite a few things and I now am a little unsure :)

litestar/contrib/sqlalchemy/repository/_async.py
provinzkraut: Oh. I was not aware of that. Seems like you need to be quite careful when implementing this lambda stuff.

litestar/contrib/sqlalchemy/repository/_async.py
cofin: So, we can't do the type inference the same way we do for a `select`.  It raises an exception when it's hinted the same way.

provinzkraut: Sad. Could you find out if SQLA intends to support this at some point?

litestar/contrib/sqlalchemy/repository/_async.py
cofin: We'd need a slightly different implementation of `to_lambda_stmt` because it expects `self.statement` to be set when it's called.

litestar/contrib/sqlalchemy/repository/_async.py
cofin: Is there any action to take on this one?  I wasn't exactly sure.

litestar/contrib/sqlalchemy/repository/_async.py
cofin: Take a look at the refactoring when you have a moment.  I've also modified the `delete_many` and `update_many` methods to use lambdas.

PR #2189, opened 2023-08-21
feat: make `302` the default `status_code` for redirect responses

### LINKED ISSUE #2138: Enhancement: make 302 the default status_code for Redirect
### Summary

I wish `Redirect()` would have a default status_code of 302. This is, I believe, the most common use case.

### Basic Example

```
Redirect(url)
```
instead of
```
Redirect(url, status_code=HTTP_302_FOUND)
```

### Drawbacks and Impact

No impact on existing code.

Some people may use 302 by default when they actually meant to use something else.

### Unresolved questions

_No response_

---

## Funding

<a href="https://polar.sh/litestar-org/litestar/issues/2138">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2138/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
Fixes #2138

### Pull Request Checklist

### Description
This PR will set the Redirect status code to 302 if it is left unset.
-

### Close Issue(s)
Closes #2138 
-


guacs (commented):
Shouldn't the default in `ASGIRedirectResponse` be changed as well?

provinzkraut:
Hmm. It might be a bit unexpected if the two different redirects available use different defaults for the status code. I do believe though that the `307` in the `ASGIRedirectResponse` is a bug, and it *should* be `302` as well. 

But this raises the question whether this ought to be considered a breaking change. It's more of a breaking bugfix really :thinking:

guacs:
I agree that having different defaults for the different redirects seems a bit confusing. 

On the topic of the breaking change, it seems this change is the one that will actually document that the default is 302, and so the previous use of 307 was undocumented and could be considered an implementation detail. I don't know...I may be stretching this a bit too much :P

cofin:
> I agree that having different defaults for the different redirects seems a bit confusing.
> 
> On the topic of the breaking change, it seems this change is the one that will actually document that the default is 302, and so the previous use of 307 was undocumented and could be considered an implementation detail. I don't know...I may be stretching this a bit too much :P

Well, it is true.  The original default was undocumented.  However, i think it could be considered breaking if anyone's existing code is set to look for the 307 instead of the 302.

I'm fine with any approach y'all are good with on this.

provinzkraut:
> However, i think it could be considered breaking if anyone's existing code is set to look for the 307 instead of the 302.

I think we should treat this as such, which means this would have to be postponed to the next major version.

guacs:
Then maybe we should document that the current default is 307?

cofin:
So how do we handle these types of changes, do we need to create the next major version branch now?

provinzkraut:
> So how do we handle these types of changes, do we need to create the next major version branch now?

Well.. I don't really want to have another branch like that for a year. I'd say we just put this into the `3.0` milestone and then address it when the time comes?

guacs:
> > So how do we handle these types of changes, do we need to create the next major version branch now?
> 
> Well.. I don't really want to have another branch like that for a year. I'd say we just put this into the `3.0` milestone and then address it when the time comes?

Does that mean this PR and the associated bug has to be kept open as well?

provinzkraut:
> Does that mean this PR and the associated bug has to be kept open as well?

I'd close the PR but keep the issue. Wdyt?

abdulhaq-e:
Merge + feature flag is a good combo imo.

provinzkraut:
> Merge + feature flag is a good combo imo.

The issue I see with that is that the feature flag implementation would require more code than the actual change. It's a really trivial change, so imo we should just postpone it.

docs/usage/responses.rst
```diff
@@ -589,7 +589,7 @@ In Litestar, a redirect response looks like this:
 
 To return a redirect response you should do the following:
 
-- set an appropriate status code for the route handler (301, 302, 303, 307, 308)
+- set an appropriate status code for the route handler (301, 302, 303, 307, 308). The default status code is 302
```
Goldziher, line 592:
```suggestion
- optionally: set an appropriate status code for the route handler (301, 302, 303, 307, 308). If not set the default of 302 will be used.
```

JacobCoffee (approved):

PR #2223, opened 2023-08-25
fix: use os.replace instead of shutil.move for renaming file

guacs:
### Pull Request Checklist

### Description

- This PR uses `os.replace` instead of `shutil.move` for renaming files to ensure atomicity.

### Close Issue(s)

### DIFF
--- a/litestar/stores/file.py
+++ b/litestar/stores/file.py
@@ -64,7 +64,7 @@ class FileStore(NamespacedStore):
                 finally:
                     os.close(tmp_file_fd)
 
-                shutil.move(tmp_file_name, target_file)
+                os.replace(tmp_file_name, target_file)  # noqa: PTH105
                 renamed = True
             finally:
                 if not renamed:
--- a/tests/unit/test_stores.py
+++ b/tests/unit/test_stores.py
@@ -344,7 +344,7 @@ def test_registry_register_exist_override(memory_store: MemoryStore) -> None:
 
 
 async def test_file_store_handle_rename_fail(file_store: FileStore, mocker: MockerFixture) -> None:
-    mocker.patch("litestar.stores.file.shutil.move", side_effect=OSError)
+    mocker.patch("litestar.stores.file.os.replace", side_effect=OSError)
     mock_unlink = mocker.patch("litestar.stores.file.os.unlink")
 
     await file_store.set("foo", "bar")


guacs:
The current usage of `shutil.move` for renaming a file is not ensured to be atomic. It uses `os.rename` semantics under the hood, but the atomicity is not guaranteed across all platforms (Windows to be more specific) as per this [issue](https://github.com/python/cpython/issues/53074).

Alc-Alc:
> The current usage of `shutil.move` for renaming a file is not ensured to be atomic. It uses `os.rename` semantics under the hood, but the atomicity is not guaranteed across all platforms (Windows to be more specific) as per this [issue](https://github.com/python/cpython/issues/53074).
> 
> Also, a point to note is that, if the destination file exists `os.rename` will raise a `FileExistsError` in Windows, but `os.replace` will silently overwrite the existing file.

Can you point me where it says this is atomic on Windows? The docs don't seem to mention that

Goldziher:
@JacobCoffee - this is a good issue for a ruff rule. I see there is a new noqa added by this pr, so it should be brought to their attention I'd say

provinzkraut:
> @JacobCoffee - this is a good issue for a ruff rule. I see there is a new noqa added by this pr, so it should be brought to their attention I'd say

I don't think this would make a good rule really? Ruff can't know if you need the atomicity, and aside from that there are differences in behaviour of the functions that might otherwise be relevant.

JacobCoffee:
as an aside, I would be interested in enabling all ruff rules rather than the 75 line list we have now.

Reason being is that as new ruff rules are added, we wouldn't know.. if we enabled them all and selectively disabled then we would get the newest new there is, and possibly greatly shorten our pyproject.toml

Goldziher:
> as an aside, I would be interested in enabling all ruff rules rather than the 75 line list we have now.
> 
> Reason being is that as new ruff rules are added, we wouldn't know.. if we enabled them all and selectively disabled then we would get the newest new there is, and possibly greatly shorten our pyproject.toml

why not do it?

Goldziher (approved):

provinzkraut (changes requested):
@guacs I second what @Alc-Alc said. I tried to confirm that `os.replace` is indeed (more) atomic than `shutil.move` but couldn't find anything. If this is indeed the case, this can be merged!

Alc-Alc:
> > The current usage of `shutil.move` for renaming a file is not ensured to be atomic. It uses `os.rename` semantics under the hood, but the atomicity is not guaranteed across all platforms (Windows to be more specific) as per this [issue](https://github.com/python/cpython/issues/53074).
> > Also, a point to note is that, if the destination file exists `os.rename` will raise a `FileExistsError` in Windows, but `os.replace` will silently overwrite the existing file.
> 
> Can you point me where it says this is atomic on Windows? The docs don't seem to mention that

If I understood what I read, internally `MoveFileEx` is used, which is also used by JDK [at least in this version](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/windows/classes/sun/nio/fs/WindowsFileCopy.java#L291) for atomic move. Also as per one of the answer [here](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/449bb49d-8acc-48dc-a46f-0760ceddbfc3/movefileexmovefilereplaceexisting-ntfs-same-volume-atomic) (not the entire answer)

> It is very possible for MoveFile to be atomic, but it is also possible for something to go wrong and for it to fall back to a non-atomic behavior even if the underlying filesystem supports atomic rename.

Just posting this here for documentation purposes.

guacs:
I wasn't able to find direct documentation that states that `os.replace` is atomic in Windows, but I think `os.replace` is the best attempt at an atomic rename. It's what is used by Python in its `importlib` for an atomic write as seen [here](https://github.com/python/cpython/blob/8ba47146111d714c7b61825d43b52311d9be366d/Lib/importlib/_bootstrap_external.py#L195). The following is a comment in that `_write_atomic` function:

>   We first write data to a temporary file, and then use os.replace() to perform an atomic rename. 

Also, this [issue](https://github.com/python/cpython/issues/58169) is the one that changes from using `os.unlink` followed by `os.rename` to instead use `os.replace`.

provinzkraut (approved):

provinzkraut:
Thanks @guacs and @Alc-Alc for investigating this!

PR #2229, opened 2023-08-26
chore: Fix typing errors and outdated lockfile

provinzkraut:
Fix some typing errors and an outdated lockfile, currently causing conflicts in the `main` branch.


tests/unit/test_openapi/test_integration.py
```diff
@@ -209,7 +211,7 @@ async def example_route() -> Lookup:
 
 def test_schema_for_optional_path_parameter() -> None:
     @get(path=["/", "/{test_message:str}"], media_type=MediaType.TEXT, sync_to_thread=False)
-    def handler(test_message: str | None) -> str:
+    def handler(test_message: Optional[str]) -> str:  # noqa: UP007
```
guacs, line 214: I think it might be better to add UP007 to Ruff's ignore list.

guacs (changes requested):

tests/unit/test_openapi/test_integration.py
provinzkraut: This is just for testing. In certain cases we want to use the backwards compatible annotations. For the general code base, and most of the tests as well the rule should be applied.

guacs: Aah got it

guacs (approved):

PR #2238, opened 2023-08-27
refactor(structure): move the repository module from contrib to top-level

Goldziher:
This PR refactors `repository` by moving the module from contrib to be top-level. That is, instead of `litestar.contrib.repository` -> `litestar.repository`.

The reasoning is as follows: the logic in repository is abstract and generalized. its meant to be used to implement third-party integrations, but is itself devoid of 3rd party components. As such it is not a contrib package, but rather an integral part of the library. Furthermore, I wish to build on this the generic controllers, which I also do not wish to place in contrib.


docs/reference/contrib/repository/abc.rst
provinzkraut, line 1: We don't want to break links in the documentation. These files should be kept in place, with a link to the new page.

litestar/contrib/repository/__init__.py
```diff
@@ -1,14 +1,20 @@
...
-    "NotFoundError",
-    "RepositoryError",
-)
+from litestar.utils import warn_deprecation
+
+
+def __getattr__(attr_name: str) -> object:
+    from litestar import repository
+
+    for k in repository.__all__:
```
provinzkraut, line 7: Why warn for all members instead of just the one that's actually being accessed? Getting a warning about a deprecated import if you didn't actually try to import that thing sounds confusing. 
Applies to all the other files as well.

litestar/contrib/repository/__init__.py
```diff
@@ -1,14 +1,20 @@
...
+        warn_deprecation(
+            deprecated_name=f"litestar.contrib.repository.{k}",
+            version="2.1",
+            kind="import",
+            removal_in="3.0",
+            info=f"importing {attr_name} from 'litestar.contrib.repository' is deprecated, please"
```
provinzkraut, line 13:
```suggestion
            alternative=f"litestar.repository.{k}"
```

applies to the other files as well.

provinzkraut (changes requested):

cofin (approved):
LGTM

litestar/contrib/repository/__init__.py
Goldziher: fixed

litestar/contrib/repository/__init__.py
Goldziher: done

docs/reference/contrib/repository/abc.rst
Goldziher: can you do this? i am not sure how to handle this sphinx stuff properly

provinzkraut (approved):

PR #2245, opened 2023-08-27
feat: add generic controller

Goldziher:
This PR adds support for generic controllers, closing #1311 

Still pending:

* DTO logic
* more tests
* documentation


benjamin-kirkbride:
@Goldziher I would like to take a stab at re-creating the [Django tutorial](https://docs.djangoproject.com/en/4.2/intro/tutorial01/) for Litestar, which may then be used as a stepping point for a "Migrating from Django" guide.

Does it make sense to wait for this to be done first? If so, is there anything I can do to help with this?

Thanks! :)

Goldziher:
> @Goldziher I would like to take a stab at re-creating the [Django tutorial](https://docs.djangoproject.com/en/4.2/intro/tutorial01/) for Litestar, which may then be used as a stepping point for a "Migrating from Django" guide.
> 
> Does it make sense to wait for this to be done first? If so, is there anything I can do to help with this?
> 
> Thanks! :)

That's awesome 😎. I suggest you start - I need to do a few more hours on this, and then it would be great if you could actually test it properly.

benjamin-kirkbride:
@Goldziher you mean start by using this branch as a basis? Or start with the way Starlite works today?

Assuming the former, I'll admit the limited tests and docs make it daunting to know where to start :sweat_smile: but I am willing to take a stab at it :)

Goldziher:
> @Goldziher you mean start by using this branch as a basis? Or start with the way Starlite works today?
> 
> Assuming the former, I'll admit the limited tests and docs make it daunting to know where to start :sweat_smile: but I am willing to take a stab at it 

Well, I should get you wait. I need a few more hours to bring this PR to a functional state with some tests, and that will make it easier for you

Goldziher:
> @Goldziher you mean start by using this branch as a basis? Or start with the way Starlite works today?
> 
> Assuming the former, I'll admit the limited tests and docs make it daunting to know where to start 😅 but I am willing to take a stab at it :)

ok, still no documentation at all and the tests are not going to clarify usage, but this should be fully operational at this point. I will complete the tests and documentation in the coming days.

Goldziher:
tests are in place, @benjamin-kirkbride - take a look and tell me what you think. I will be adding docs as a final step before review.

litestar/controller/generic.py
```diff
@@ -0,0 +1,226 @@
...
+    "update_many",
+    "delete_many",
+    "get_many",
+)
+
+
+class GenericController(Controller, Generic[ModelT, IdAttrT]):
+    repository_type: type[AbstractAsyncRepository[ModelT]] | type[AbstractSyncRepository[ModelT]]
+    """Repository for the controller's model."""
+    id_attribute: str = "id"
```
abdulhaq-e, line 44: The`id_attribute` is already defined on the repository. Can't that be used via `repository_type` 

I think it would be less ambiguous if it's defined once only.

litestar/controller/generic.py
```diff
@@ -0,0 +1,226 @@
...
+    "delete_instance",
+    "get_instance",
+    "create_many",
+    "update_many",
+    "delete_many",
+    "get_many",
+)
+
+
+class GenericController(Controller, Generic[ModelT, IdAttrT]):
```
abdulhaq-e, line 41: I'd prefer an API were the controller has 1 generic type only. I can see why you opted for `IdAttrT`, would it be possible to extract the type of id attribute using `ModelT` and `id_attribute`.

`GenericController[MyModel]` looks much better that `GenericControlller[MyModel, str]`

abdulhaq-e (commented):
I made some suggestions, not sure how feasible they are.

litestar/controller/generic.py
Goldziher: Ok, I'll look into it

Goldziher: done

litestar/controller/generic.py
Goldziher: i went with your input, this is the API now

litestar/controller/generic.py
```diff
@@ -0,0 +1,243 @@
...
+                route_handler.owner = self
+                route_handlers.append(route_handler)
+
+        route_handlers.sort(key=attrgetter("handler_id"))
+        return route_handlers
+
+    @classmethod
+    def get_generic_annotations(cls) -> tuple[type, type]:
+        try:
+            if (generic_bases := getattr(cls, "__orig_bases__", None)) and (
```
sobolevn, line 155: `__orig_bases__` generally contain `tuple[Any, ...]`, not `tuple[type, ...]`, because there are multiple special forms that are allowed.

sobolevn: See https://github.com/python/typeshed/blob/703ed36d5a5c9505c903ea2182e6eed679d9bddb/stdlib/types.pyi#L586 and https://github.com/python/typeshed/issues/7811

Goldziher: Cool,  addressed

abdulhaq-e (approved):
LGTM

Maybe add docs only for this PR and we can leave DTO handling for another PR.

Goldziher:
Ok, I will merge this branch into a feature branch. 

I implemented DTO logic, but it uncovers issues with the DTO stuff. So I will do this in a followup.

PR #2294, opened 2023-09-10
feat: add DTOs in generic controllers

Goldziher:
This PR adds the DTO support to generic controllers. 

Tasks:
- [ ] more tests
- [ ] documentation


litestar/controller/generic.py
```diff
@@ -188,20 +206,106 @@ def path_param_type(self) -> str:
...
     def create_repository(
         self, *, request: Request[Any, Any, Any], **kwargs: Any
     ) -> AbstractAsyncRepository[ModelT] | AbstractSyncRepository[ModelT]:
         kwargs["request"] = request
         return self.repository_type(**kwargs)
 
-    async def create_instance(self, data: ModelT, request: Request[Any, Any, Any]) -> ModelT:
-        result = self.create_repository(request=request).add(data=data)
+    async def create_instance(self, data: DTOData[ModelT], request: Request[Any, Any, Any]) -> ModelT:
+        result = self.create_repository(request=request).add(data=data.as_builtins())
```
abdulhaq-e, line 302: I must be missing something here, why not `DTOData.create_instance()` That to me looks like it will save you from all the edits on the repositories.

Goldziher: instance is missing an ID or is partial (depending on the method), if its pydantic etc. it will fail validation

abdulhaq-e: I see what is happening now.

I would suggest an alternative API that would require sightly more input from the user (but leads to less code for the library):

1. Have a required input to the generic controller that is an "id_factory". The results of calling the factory will be passed to `create_instance` hence no validation errors.
2. This one is more tricky, we need to obtain the whole model from the repository and pass it to `update_instance`. This can be done without requiring any input from the user, we already have the `repository` and the `id` (from the url path).

What do you think?

Goldziher: You are welcome to make a parallel PR with these, let's see what's what and decide between the patterns.

provinzkraut:
This won't be worked on at the moment. We'll revisit at a later point.

PR #2295, opened 2023-09-11
feat(layered-include_in_schema): Add include_in_schema option for all layers

### LINKED ISSUE #2267: Enhancement: Add include_in_schema option on Routers and/or Controllers
### Summary

I'd like to exclude entire Controllers from my OpenAPI docs, with the current state I need to add include_in_schema=False on every single route handler in the controller.
It would be very convenient if I could specify this option on the Controller level (and possibly on the router level as well), and it would be propagated downwards. 

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

Alc-Alc:
Add include_in_schema option for all layers

### Pull Request Checklist

### Description

- https://github.com/litestar-org/litestar/issues/2267

### Close Issue(s)

- Closes #2267


v3ss0n:
include_in_schema sounds a bit long and duplicated .
since it is a Flag , how about `in_schema` or `in_openapi`

guacs:
> `include_in_schema` sounds a bit long and duplicated . since it is a Flag , how about `in_schema` or `in_openapi`

In the decorators, the argument's name is `include_in_schema` so it's probably best to keep it the same in this case as well for consistency.

v3ss0n:
> > `include_in_schema` sounds a bit long and duplicated . since it is a Flag , how about `in_schema` or `in_openapi`
> 
> In the decorators, the argument's name is `include_in_schema` so it's probably best to keep it the same in this case as well for consistency.

okay.

litestar/handlers/http_handlers/base.py
```diff
@@ -387,6 +389,23 @@ def resolve_after_response(self) -> AsyncCallable | None:
...
+
+        If 'include_in_schema' is found in any of the ownership layers, the last value found is returned.
+        If not found in any layer, the default value ``True`` is returned.
+
+        Returns:
+            bool | None: The resolved 'include_in_schema' property.
+        """
+        if self._resolved_include_in_schema is Empty:
+            include_in_schemas = [
+                i.include_in_schema for i in self.ownership_layers if i.include_in_schema is not Empty
```
Goldziher, line 403: Wouldn't it be better to check that the value is an instance of bool? We want to filter out none and empty and just have a list of bool values.

litestar/handlers/http_handlers/base.py
```diff
@@ -387,6 +389,23 @@ def resolve_after_response(self) -> AsyncCallable | None:
...
+            ]
+            self._resolved_include_in_schema = include_in_schemas[-1] if include_in_schemas else True
+
+        return cast("bool | None", self._resolved_include_in_schema)
```
Goldziher, line 407: I think this should return only bool

litestar/handlers/http_handlers/decorators.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         content_media_type: str | None = None,
         deprecated: bool = False,
         description: str | None = None,
-        include_in_schema: bool = True,
+        include_in_schema: bool | EmptyType | None = Empty,
```
Goldziher, line 85: Why none? 

I get why the empty type but not what's none here

litestar/handlers/http_handlers/decorators.py
```diff
@@ -246,7 +246,7 @@ def __init__(
...
```
Goldziher, line 249: Same everywhere regarding none

Goldziher (commented):
All in all very good, a few comments

litestar/handlers/http_handlers/decorators.py
Alc-Alc: I followed the type hints used in `return_dto: type[AbstractDTO] | None | EmptyType = Empty`, I did find the usage of None for `include_in_schema` rather pointless, so I will make it `bool | Empty`

Goldziher: Yes, the typing there is meant to allow explicitly overriding the DTO on lower layers and setting it the none. It's not the best API tbh.

Goldziher (approved):

PR #2296, opened 2023-09-11
feat: handling DTOs in generic controller

abdulhaq-e:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-


Goldziher:
So, looking at the code there are a few problems I see. 

1. The idea of a repository is to be the layer where data persistence is handled. The pattern you are suggesting basically overrides the separation. 

2. The data persistence layer should handle id creation, and ensure there is no collision. The handling in code doesn't ensure there are no collisions in the persistence layer, and it's additionally way less performant. 

3. The update operation logic is inherently not performant. A database allows you to update partial values without previously retrieving the value. 

The problem we have is, IMO, a problematic design of the repository Interface.

abdulhaq-e:
> The problem we have is, IMO, a problematic design of the repository Interface.

The points you raise are valid but, IMO, repository design varies, repositories as defined in DDD get a domain model, not a dictionary-like data structure. 

Since the repositories in Litestar are more aligned towards CRUD, maybe the interface doesn't suit having a domain model passed to the methods.

Goldziher:
> The points you raise are valid but, IMO, repository design varies, repositories as defined in DDD get a domain model, not a dictionary-like data structure.

On is this based? DDD as far as I am familiar with it, is not a specification. The interface for repository methods is arbitrary. 

In either case - I don't think this PR here can proceed. We have to address the repository interface we have, since its not workable as such.

abdulhaq-e:
Yes it's not a specification and of course there is no one correct way. 

The idea is that a domain model is either created or loaded from the db then modified. This domain model should in theory be rich in behaviour (or else it's just CRUD), after that, it's passed to the repository that persists the model in an abstracted way.

Your design implies that primitive validated data is passed to the repository and the repository will create/update. I'm with you that this is the way to go for CRUD.

I will close this and will try to finish reviewing your PR asap.

PR #2344, opened 2023-09-23
fix: `iter_data` must be async in `litestar/connection/websocket.py`

sobolevn:
I've also filled a mypy bug: https://github.com/python/mypy/issues/16166

Thanks for this :)


provinzkraut:
This breaks mypy though. Iirc this was the reason I didn't make the overload async when I added them. Would this be covered by the fix you proposed?

sobolevn:
Async support is a mess 🤦‍♂️ 

I will investigate what is going on.

sobolevn:
Closing for now.

Issue #2302, opened 2023-09-14
Enhancement: Validate dependency types when registering dependencies and handlers


ccrvlh:
### Description

When injecting a dependency, if the type annotation on the handler is different than the one on the callable, the handler fails silently and just returns a 500. The reason I'm posting this as a bug, is because this is silent, so it can be very hard to trace the issue.

### MCVE

This works fine, note that `get_host` returns a `str` and `host_context` is also annotated with a string.
```python
from litestar import get
from litestar import Controller
from litestar.di import Provide
from litestar import Response
from litestar import Request


def get_host(request: Request) -> str:
    return request.headers.dict().get("host", ["empty list"])[0]


class MyController(Controller):
    path = "/controller"

    dependencies = {"host_context": Provide(get_host)}

    @get("/{param:int}")
    def handler_get_id(self, param: int, host_context: str) -> Response:
        print(host_context)
        return Response(dict(param=param))
```

If the `get_host` is annotated with a `str` and the handler is taking a `dict`, the route fails silently, and the handler just returns a 500.

```python
from litestar import get
from litestar import Controller
from litestar.di import Provide
from litestar import Response
from litestar import Request


def get_host(request: Request) -> str:
    return request.headers.dict().get("host", ["empty list"])[0]


class MyController(Controller):
    path = "/controller"

    dependencies = {"host_context": Provide(get_host)}

    @get("/{param:int}")
    def handler_get_id(self, param: int, host_context: dict) -> Response:
        print(host_context)
        return Response(dict(param=param))
```

The handler failing is not an issue per se, consider this is built with type safety in mind, but the lack of warning, error message or something that says what's happening, seems like a bug.

### Litestar Version

2.0.1

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---

## Funding
* If you would like to see an issue prioritized, make a pledge towards it!
* We receive the pledge once the issue is completed & verified

<a href="https://polar.sh/litestar-org/litestar/issues/2302">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2302/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2302/pledge.svg">
</picture>
</a>

guacs:
[...]

PR #2349, opened 2023-09-24
feat: check consistency between dependency type and handler parameter type

ccrvlh:
### Pull Request Checklist

### Description

This is a first draft, wanted to validate the approach before moving forward, I think drafts are the easiest way to do that.
When declaring a dependency, if the type of the return statement on the dependency is different than what the handler expects, the server will silently return a 500 error. As this is enforced at runtime, it seems to make sense to enforce this when starting up.

The approach was based on @peterschutt suggestion to work with the `KwargsModel`. It does seems to be a good place to put this validation. There's a catch though: without raising the exception, the error will be found both ways, meaning first there will be a mismatch between the dependency and handler, and afterwards between the handler and the dependency (exactly the same thing, but once through the handler's lenses and then through the dependency lenses). This is not an issue per se, since the app would fail at the very first inconsistency, but it would make sense to make this cleaner. One way to fix that though of only doing that for handlers, but I didn't spend much time looking into how to check whether what was being checked was indeed a handler.

Also, I used a very hacky approach to get the handler name, couldn't find it in the `signature_model`. So I just used `signature_model.__name__.replace("_signature_model", "")`. I'm sure there's a cleaner (and less error prone) way of doing that.

What's missing:
- [ ] Validating the approach
- [ ] Fix logic to get the handler name 
- [ ] Ensure validation only happens one way (validates the handler signature and compare to the dependency vs doing both)
- [ ] Making sure this works regardless of how dependencies were injected, or routes were declared
- [ ] Adding to docs that now dependency/handler typing is enforced during setup. This is specially relevant since type annotations don't usually affect runtime behavior. 
- [ ] Deleting test app (any interest in keeping this?)

### Close Issue(s)
- [ ] #2302 

### Extra Thoughts

Even though `KwargsModel` is a fairly generic class, and the the `create_for_signature_model` is a fairly generic method, it is being used only twice: one for the route handler, and the other for dependencies. This makes the method a bit tricky to deal with IMHO: `signature_model`, `parsed_signature` are generic terms - perhaps it could make sense to `@overload` this, or even completely split it into two different methods (one for the handlers, and the other for the dependencies), while sharing private helpers. That way we could have something a bit easier to reason about, like `handler_parameters` / `handler_signature`. Also `signature_model` is the same name as the class it receives, so having something slighly more precise (business-wise) seems to make sense, since we would already have the `type[SignatureModel]` annotation anyways.


litestar/_kwargs/kwargs_model.py
```diff
@@ -359,6 +365,41 @@ def create_for_signature_model(
...
+
+        Returns:
+            True if the annotations are consistent, False otherwise
+        """
+        for dependency_field_name, dependency_field in dependencies.items():
+            if dependency_field_name not in handler_params:
+                continue
+            handler_param = handler_params[dependency_field_name]
+            dependency_return_annotation = dependency_field.signature_model._return_annotation
+            if dependency_return_annotation == handler_param.annotation:
```
peterschutt, line 397: I think we should be checking for sub-types, not just straight equality.

E.g:

```py

class MyStr(str):
    pass


def test(s: str) -> None:
    ...


test(MyStr("hello"))  # OK
```

```suggestion
            if FieldDefinition.from_annotation(dependency_return_annotation).is_subclass_of(handler_param.annotation)
```

It's a shame that the signature model throws away the `FieldDefinition` of the return annotation such that we have to recreate it, here:

https://github.com/litestar-org/litestar/blob/645e41e787b9cb0697bdb32e2997e9149cc49705/litestar/_signature/model.py#L267

ccrvlh: > I think we should be checking for sub-types, not just straight equality.

Absolutely. I didn't think about that.

> the signature model throws away the FieldDefinition of the return annotation 

Yeah. It also doesn't keep the `fn_name` right?
This struct seems to be used in quite a few places, so I didn't consider changing it.

peterschutt:
> Also, I used a very hacky approach to get the handler name, couldn't find it in the `signature_model`. So I just used `signature_model.__name__.replace("_signature_model", "")`. I'm sure there's a cleaner (and less error prone) way of doing that.

In retrospect, doing this is `KwargsModel` is probably not the best place.

On registration, we call each handler's `on_registration()` method:

https://github.com/litestar-org/litestar/blob/645e41e787b9cb0697bdb32e2997e9149cc49705/litestar/app.py#L601-L602

As part of that we call `BaseRouteHandler._validate_handler_function()`:

https://github.com/litestar-org/litestar/blob/645e41e787b9cb0697bdb32e2997e9149cc49705/litestar/handlers/base.py#L493-L502

Which on `BaseRouteHandler` is just a noop impl:

https://github.com/litestar-org/litestar/blob/645e41e787b9cb0697bdb32e2997e9149cc49705/litestar/handlers/base.py#L510-L512

 Perhaps this should go in there?

Because, where `KwargsModel.create_for_signature_model()` is called, all of the things we need to check the dependency types are sourced from the route handler anyway:

https://github.com/litestar-org/litestar/blob/645e41e787b9cb0697bdb32e2997e9149cc49705/litestar/routes/base.py#L135-L138

So, if you move this validation to `BaseRouteHandler._validate_handler_function()` you'd have access to the `BaseRouteHandler.handler_name`, and it probably does make more sense to have the check as its own thing, away from the kwags model construction.

peterschutt:
There's also the question of how far we'd want to go checking the types of collections/generics etc, and I fear this could get a bit messy.

Say for a dependency that returns a list of something:

```py
def provide_list_int() -> list[int]:
    ...

@get(dependencies={"the_list": provide_list_int})
def receive_list_str(the_list: list[str]) -> None:
    ...
```
That case would be fairly simple, I think, if we check that both the outer and pairwise inner types are sub-types, and I suppose that could be done recursively for things like `list[dict[str, bool]]` <: `Sequence[Mapping[str, int]]`.

Given the potential complexity, and the number of tests that now fail due to the change in startup behavior, I wonder if we shouldn't develop this behind a feature flag so that we can leave default behavior for now and iterate on this outside of the normal sem-ver guarantee. E.g., `Litestar(..., validate_dependency_types=True)`, which could have a warning on startup about the feature being experimental/alpha/subject to breaking changes etc.

Thoughts @litestar-org/maintainers

ccrvlh:
> In retrospect, doing this is KwargsModel is probably not the best place.

Seems to make sense, I'll try to spend more time on it this week and check the full life cycle, I confess I went straight into the `kwargs_model`, so haven't looked into other parts as much.

> There's also the question of how far we'd want to go checking the types of collections/generics etc, and I fear this could get a bit messy.

Isn't this being done already? I mean, this _is_ being validated during runtime (and returns 500 if it fails), so wouldn't the behavior stay exactly the same? I do a agree that a feature flag would be helpful indeed.

provinzkraut:
> Isn't this being done already? I mean, this is being validated during runtime (and returns 500 if it fails), so wouldn't the behavior stay exactly the same (only changing when it happens rather than if it happens)? 

No, this is not done. There's not actually any type checking happening there, it's just runtime validation, provided by msgspec (and Pydantic / attrs). This is *very* different from checking types ahead of time. This feature would be more similar to a classic static type checker like mypy.

After having thought about this some more, I am very convinced we should not do this, for a couple of reasons:

- There is a great number of hidden complexities involved in this, and once we have this built in, and people start to rely on it, we can't easily take it out anymore.
- There are already great type checkers out there, like mypy. It would make much more sense to provide a plugin for those that enable them to type check Litestar apps than to rebuilt everything they've done from the ground up
- It's not a really strong use case, as everything covered by this would also be caught in testing. Actually, the examples brought forth so far would only catch things if there were *no tests at all*, since every request would lead to a validation error. 


> There's also the question of how far we'd want to go checking the types of collections/generics etc, and I fear this could get a bit messy.


This is a good example of what I'm talking about. How do we handle generics and collections? Covariance? Inheritance? Multiple inheritance? And I don't think we could actually just ignore those things, since if people are using this, they'd expect it to work like mypy does, since it's the de-facto standard, which means we'll have to re-implement significant parts of it to be useful.

Also, consider an example like this:

```python
from litestar import get
from typing import Any

def get_thing() -> int:
    return "hello"

@get("/", dependencies={"thing": get_thing})
def handler_get_id(thing: int) -> None:
    pass
```

this would pass naive type checking, but would fail at runtime and would fail in mypy.


```python
from litestar import get
from typing import Any

def get_thing() -> int:
    return 0.1

@get("/", dependencies={"thing": get_thing})
def handler_get_id(thing: float) -> None:
    pass
```

This would fail naive type checking, pass at runtime and in mypy.

peterschutt:
> There is a great number of hidden complexities involved in this, and once we have this built in, and people start to rely on it, we can't easily take it out anymore.

Agree. At best this could only start out as an experimental feature with appropriate warnings if enabled.

> There are already great type checkers out there, like mypy. It would make much more sense to provide a plugin for those that enable them to type check Litestar apps than to rebuilt everything they've done from the ground up

I'm not even sure that a mypy plugin could help us in this case, as the names of dependencies on the handler's graph aren't going to be known statically. Also, mypy plugins don't help non-mypy users.

> It's not a really strong use case, as everything covered by this would also be caught in testing. Actually, the examples brought forth so far would only catch things if there were no tests at all, since every request would lead to a validation error.

This is the strongest argument against the feature - we should encourage writing tests. Coupled with maintenance burden, I agree that this is not worth pursuing. Take everything else I've said here as comment.

> ```python
> from litestar import get
> from typing import Any
> 
> def get_thing() -> int:
>     return "hello"
> 
> @get("/", dependencies={"thing": get_thing})
> def handler_get_id(thing: int) -> None:
>     pass
> ```
>
> this would pass naive type checking, but would fail at runtime and would fail in mypy.

In this case, that is sufficient. I don't think we'd ever want to say that we are replacing the need for static type checkers. If a user chooses not to use one and experiences this class of error, I'm fine with that.

> ```python
> from litestar import get
> from typing import Any
> 
> def get_thing() -> int:
>     return 0.1
> 
> @get("/", dependencies={"thing": get_thing})
> def handler_get_id(thing: float) -> None:
>     pass
> ```
> 
> This would fail naive type checking, pass at runtime and in mypy.

Actually, this fails `mypy` b/c a `float` is not an `int`, and for the same reason, this would pass naive checking as we would test that the return type of the dependency is a sub-type of the handler type, and an `int` is a `float`.

provinzkraut:
> I'm not even sure that a mypy plugin could help us in this case, as the names of dependencies on the handler's graph aren't going to be known statically.

They can be known once we have an app object though, right? I was thinking that the mypy plugin would need a definitive entry point to work, so it could resolve everything correctly.

> In this case, that is sufficient. I don't think we'd ever want to say that we are replacing the need for static type checkers.

I too think that it is sufficient and we don't *want* that, but I'm very certain that if we introduce this feature, there will inevitably requests for things like this, and we're going to have to continue discussing the scope of the type checking we're doing.

> Actually, this fails mypy b/c a float is not an int, and for the same reason, this would pass naive checking as we would test that the return type of the dependency is a sub-type of the handler type, and an int is a float.

Actually actually, this does not fail with mypy, because `int`s and `float`s are special cased. Quoting what PEP 484 calls ["The numeric tower"](https://peps.python.org/pep-0484/#the-numeric-tower):

> this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type `float`, an argument of type `int` is acceptable.

peterschutt:
> Actually actually

I think maybe we aren't talking about the same thing - I mean the dependency provider function can't return a float when it is annotated to return an int.

peterschutt:
> > > Actually actually
> > 
> > 
> > I think maybe we aren't talking about the same thing - I mean the dependency provider function can't return a float when it is annotated to return an int.
> 
> Oh. Well that was just an error in my example 🙃

Also, I had an error in my reasoning, given that `issubclass(int, float)` is `False` the naive type checking would most likely fail unless we used a method like `is_compatible_with()` that included the duck type logic of int <: float, etc.

So close this, and mark issue as "wontfix"?

provinzkraut:
> So close this, and mark issue as "wontfix"?

Yeah, I think so

peterschutt:
Thanks for your interest and efforts toward this @ccrvlh, but we've decided the complexity and ongoing maintenance burden outweigh the potential benefit.

ccrvlh:
@peterschutt sure. Been following the discussion and I do agree.

I still wonder about the current approach. From what I understand it does _some_ type of validation, even if indirectly (through msgspec Structs) - the type annotation affects the validation of the dependency logic. I wonder how all of those cases are being handled (even if implicitly) right now, and if the current approach should indeed fail.

PR #2352, opened 2023-09-24
chore: use `Final` instead of `Literal` for constants

sobolevn:
There are several differences:
- `Final` does not need value duplication, it is inferenced by type-checkers to be `Literal[value]` automagically
- `Final` does not allow value re-assigning (`x: Literal[1] = 1; x = 1` is allowed)


peterschutt (approved):
Nice, thanks! I have a few code bases I need to go and update now.

PR #2361, opened 2023-09-25
ci: Use dmypy to speed up type checking

provinzkraut:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,6 +65,7 @@ repos:
     rev: "v1.5.1"
     hooks:
       - id: mypy
+        entry: dmypy run --
         exclude: "test_apps|tools|docs|tests/examples|tests/docker_service_fixtures|tests/unit/test_contrib/test_pydantic/test_beanie_integration"
         additional_dependencies:
           [
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ target/
 /docs/_build/
 coverage.*
 setup.py
+/.dmypy.json


JacobCoffee (approved):

.pre-commit-config.yaml
```diff
@@ -65,6 +65,7 @@ repos:
     rev: "v1.5.1"
     hooks:
       - id: mypy
+        entry: dmypy run --
```
sobolevn, line 68: > If you have a large codebase to check, running mypy using the mypy daemon can be 10 or more times faster than the regular command-line mypy tool, especially if your workflow involves running mypy repeatedly after small edits – which is often a good idea, as this way you’ll find errors sooner.

dmypy gives a significant speed up when you have incremental changes. Regular mypy runs can be the same or even slower.

provinzkraut: Yeah, that's why I was thinking it would be great if we could use dmypy locally and regular mypy in CI. But after digging a bit, there doesn't seem to be a great solution for this when using pre-commit :/

Only way to do this would be to not use the pre-commit hook and instead install mypy locally and invoke it with a script that checks whether it's running in CI and then can issue the appropriate command. This script can then be used in a pre-commit hook.

Not very elegant though.

sobolevn: In my opinion, `pre-commit` is not suitable for mypy. Reasons:
1. No cache between runs
2. It alters mypy's defaults
3. It requires an extra set of dependencies
4. Mypy does not fit the "run on changes" rule of pre-commit, mypy still reports errors in other files and it needs all of them to build a graph

PR #2378, opened 2023-09-28
fix: remove runtime assignment of type aliases to `Any`.

peterschutt:
This PR replaces this pattern:

```py
from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:
    from typing_extensions import TypeAlias

    from litestar import Thing

    ThingAlias: TypeAlias = Thing
else:
    ThingAlias: TypeAlias = Any
```

My best guess is that this pattern was some workaround for circular dependencies before we introduced `__future__.annotations`.

I believe these changes will be beneficial as:
- the current pattern is non-standard (at least in my experience)
- the current pattern causes the types to be documented as `Any` in the API docs.

Example of current API docs for alias types:

image

### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/litestar/types/internal_types.py
+++ b/litestar/types/internal_types.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING, Any, Callable, Literal, NamedTuple
 
 __all__ = (
     "ControllerRouterHandler",
@@ -15,9 +15,6 @@ __all__ = (
 
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-    from typing import Any, Literal
-
     from typing_extensions import TypeAlias
 
     from litestar.app import Litestar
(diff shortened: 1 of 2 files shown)


peterschutt:
Type API docs improve somewhat:

image

Any @litestar-org/maintainers or @litestar-org/members know how to prevent this?

litestar/types/internal_types.py
```diff
@@ -23,6 +15,9 @@
 
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Any, Literal
```
sobolevn, line 19: Nit: since we already import `typing`, we can keep these imports there. This will allow to un-stringify some aliases like `ReservedKwargs`.

But, this is not really important, we can also keep this PR the way it is!

sobolevn (approved):

litestar/types/internal_types.py
peterschutt: Thanks!

JacobCoffee (approved):

PR #2380, opened 2023-09-28
docs: remove references to `SQLAlchemyAsyncConfig.create_engine`

### LINKED ISSUE #2374: Docs: errors in the SQLAlchemy Repository Tutorial
### Summary

Running the examples in the SQLAlchemy Repository Tutorial give the following error:

```
    |   File "toto.py", line 144, in on_startup
    |     async with sqlalchemy_config.create_engine().begin() as conn:
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | AttributeError: 'SQLAlchemyAsyncConfig' object has no attribute 'create_engine'
```

I suspect the API has changed.

---

## Funding

<a href="https://polar.sh/litestar-org/litestar/issues/2374">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2374/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

lsanpablo:
### Pull Request Checklist

### Description

- Remove references to `SQLAlchemyAsyncConfig.create_engine` from docs; use `SQLAlchemyAsyncConfig.get_engine` in examples instead.

### Close Issue(s)

- Closes #2374

### DIFF
--- a/docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
+++ b/docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
@@ -204,7 +204,7 @@ sqlalchemy_plugin = SQLAlchemyInitPlugin(config=sqlalchemy_config)
 
 async def on_startup() -> None:
     """Initializes the database."""
-    async with sqlalchemy_config.create_engine().begin() as conn:
+    async with sqlalchemy_config.get_engine().begin() as conn:
         await conn.run_sync(UUIDBase.metadata.create_all)
 
 
--- a/docs/examples/contrib/sqlalchemy/sqlalchemy_declarative_models.py
+++ b/docs/examples/contrib/sqlalchemy/sqlalchemy_declarative_models.py
@@ -39,7 +39,7 @@ sqlalchemy_plugin = SQLAlchemyInitPlugin(config=sqlalchemy_config)
 
 async def on_startup() -> None:
     """Initializes the database."""
-    async with sqlalchemy_config.create_engine().begin() as conn:
+    async with sqlalchemy_config.get_engine().begin() as conn:
         await conn.run_sync(UUIDBase.metadata.create_all)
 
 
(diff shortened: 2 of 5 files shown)


lsanpablo:
According to the [advanced-alchemy docs](https://docs.advanced-alchemy.jolt.rs/latest/reference/config/common.html#advanced_alchemy.config.common.GenericSQLAlchemyConfig.get_engine:~:text=Return%20an%20engine.%20If%20none%20exists%20yet%2C%20create%20one.), `get_engine` seems like the only way to get what we want in those examples.

peterschutt:
`create_engine()` was part of the interface in the 2.0 release, so really should still exist.

It existed in the commit prior to the `advanced-alchemy` commit:

https://github.com/litestar-org/litestar/blob/a463aab9eec73266e10db71a83d3ede95e1832c1/litestar/contrib/sqlalchemy/plugins/init/config/common.py#L174-L195

lsanpablo:
It seems like the best thing to do is to add a deprecated `create_engine` function which just calls `get_engine` and leave the docs as they are in the PR. Correct?

cofin:
> I agree, `create_engine` should still be part of the interface. It seems like the best thing to do is to add a deprecated `create_engine` method which just calls `get_engine` and leave the docs as they are in the PR. Correct?

Yes, I think this is correct.  This means we should probably do this over on the `advanced_alchemy` side.

guacs:
> > I agree, `create_engine` should still be part of the interface. It seems like the best thing to do is to add a deprecated `create_engine` method which just calls `get_engine` and leave the docs as they are in the PR. Correct?
> 
> Yes, I think this is correct. This means we should probably do this over on the `advanced_alchemy` side.

Wouldn't it be better to just subclass the config and add that method in `litestar` till v3?

provinzkraut:
> Wouldn't it be better to just subclass the config and add that method in `litestar` till v3?

Good point. We re-export things anyway from Litestar, so we might as well just do that.

peterschutt (approved):

PR #2389, opened 2023-09-30
feat(DTO): deterministic transfer model names

provinzkraut:
We currently ensure uniqueness of the generated transfer model names by appending a random string to a base name. This makes debugging and testing harder in cases where you want to reference / compare the models between runs, since the names differ on every run.

This PR introduces a helper function that generates a unique name by simply appending an increasing integer to the base name, ensuring that, given nothing else changed, transfer models will have the same names.


JacobCoffee (approved):

litestar/utils/helpers.py
```diff
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Generic, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Container, Generic, TypeVar, cast
```
Alc-Alc, line 5: IIRC, `typing.Container` is deprecated since 3.9
```suggestion
from collections.abc import Container
```

Alc-Alc (approved):

PR #2390, opened 2023-09-30
feat(OpenAPI): lazy import example creator

provinzkraut:
Only import function to create OpenAPI examples when needed; In the case where users have turned off generating examples this import is not needed at all.

This cuts the time it takes to `import litestar` in half by avoiding an expensive `polyfactory` import and is the first step in making polyfactory optional.


cofin (approved):
LGTM and very welcomed.

PR #2393, opened 2023-10-01
fix: #1301 - Apply compression before caching

### LINKED ISSUE #1301: Bug: Compression is applied after request caching
The CompressionMiddleware applies compression after CacheMiddleware caches the response and, on subsequent requests that hit the cache, applies compression every time.

I think in a perfect world the response would be cached after it is sent though I realize this comes with significant implementation issues.

Here's an easy way to reproduce:
```python
from typing import Any
from starlite import (
    Request,
    Starlite,
    get,
)
from starlite.config.cache import CacheConfig
from starlite.config.compression import CompressionConfig
from starlite.config.logging import LoggingConfig
from starlite.datastructures.headers import MutableScopeHeaders
from starlite.middleware.base import MiddlewareProtocol
import uvicorn
from rich import print

from starlite.types.asgi_types import ASGIApp, Message, Receive, Scope, Send

@get('/test_txt', cache=True)
async def handle(request: Request) -> Any:
    with open('test.txt', 'r') as f:
        return f.read()
    
async def after_response(request: Request) -> None:
    cached_val = await request.app.cache.get('/test_txt')
    print('Cached val length', len(cached_val))

class InspectBodyMiddleware(MiddlewareProtocol):
    def __init__(self, app: ASGIApp) -> None:
        super().__init__(app)
        self.app = app

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        if scope["type"] == "http":
            async def send_wrapper(message: Message) -> None:
                if message['type'] == "http.response.body":
                    print('Response val length',len(message['body']))
                await send(message)

            await self.app(scope, receive, send_wrapper)
        else:

app = Starlite(
    route_handlers=[handle],
    after_response=after_response,
    logging_config=LoggingConfig(),
    cache_config=CacheConfig(),
    compression_config=CompressionConfig(backend="gzip", gzip_compress_level=9),
    middleware=[InspectBodyMiddleware]
)

def main() -> None:
    uvicorn.run("main:app", host="0.0.0.0", port=8001, reload=True) 

if __name__ == '__main__':
    main()


```

The output for me, using a 1MB file for `test.txt` is:
```shell
Response val length 499774
Cached val length 1057092
```

<a href="https://polar.sh/litestar-org/litestar/issues/1301">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1301/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

provinzkraut:
Fix #1301 by introducing a moving caching to a hybrid approach that records outgoing ASGI messages within a middleware and hydrates them within the route when responding to a request.

There are some additional benefits to this approach:

- Improved performance and reduced size of the stored responses by switching from pickle to MessagePack and storing already compressed responses and only compressing them once
- Reduced overhead because the middleware can be applied ahead of time to only those routes that have caching enabled instead of checking each request
- Every response can now be cached without issues; Previously only those that could be pickled could be cached

### DIFF
--- a/litestar/_asgi/routing_trie/mapping.py
+++ b/litestar/_asgi/routing_trie/mapping.py
@@ -186,6 +186,8 @@ def build_route_middleware_stack(
     from litestar.middleware.allowed_hosts import AllowedHostsMiddleware
     from litestar.middleware.compression import CompressionMiddleware
     from litestar.middleware.csrf import CSRFMiddleware
+    from litestar.middleware.response_cache import ResponseCacheMiddleware
+    from litestar.routes import HTTPRoute
 
     # we wrap the route.handle method in the ExceptionHandlerMiddleware
     asgi_handler = wrap_in_exception_handler(
@@ -197,6 +199,10 @@ def build_route_middleware_stack(
 
     if app.compression_config:
         asgi_handler = CompressionMiddleware(app=asgi_handler, config=app.compression_config)
+
+    if isinstance(route, HTTPRoute) and any(r.cache for r in route.route_handlers):
+        asgi_handler = ResponseCacheMiddleware(app=asgi_handler, config=app.response_cache_config)
+
     if app.allowed_hosts:
         asgi_handler = AllowedHostsMiddleware(app=asgi_handler, config=app.allowed_hosts)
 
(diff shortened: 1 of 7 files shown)


cofin (approved):
LGTM

PR #2394, opened 2023-10-02
fix: don't implicitly parse URL encoded form data as JSON

provinzkraut:
We currently parse URL encoded form data as JSON. This is arguably not the correct behaviour and can lead to serious bugs if something looks like JSON but isn't meant to be. It is also unexpected and there's no way for a user to control this.

This PR fixes the behaviour and adjusts the tests accordingly.

As an aside, it allows us to make `fast-query-parsers` optional as a speedup, which has now been moved to the `standard` and `full` extra.

### DIFF
--- a/litestar/_parsers.py
+++ b/litestar/_parsers.py
@@ -1,18 +1,25 @@
 from __future__ import annotations
 
+from collections import defaultdict
 from functools import lru_cache
 from http.cookies import _unquote as unquote_cookie
-from typing import Any, Iterable
+from typing import Iterable
 from urllib.parse import unquote
 
-from fast_query_parsers import parse_query_string as fast_parse_query_string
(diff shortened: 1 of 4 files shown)


cofin (approved):
LGTM and another welcomed change

provinzkraut:
No, this has nothing to do with that. How the JSON presented in the browser is entirely up to your browser. The change only affects how incoming form data is parsed, not how data is rendered as JSON. As you can see, it is still JSON; Your browser just displays it differently than the one you're seeing a screenshot of in the tutorial.

MaxenceBouvier:
@provinzkraut Thanks for your fast and detailed answer.
That actually helps a lot!
I would not have looked on the browser side for this json interface (non)issue.

PR #2398, opened 2023-10-03
fix(connection): Simplify connection header parsing

provinzkraut:
~~- Use `MultiDict` instead of a `dict[str, str]` to inject into route handlers. This is non-breaking as `MultiDict` behaves exactly like a regular dict outside for its additional methods~~ (Turns out this is not non-breaking)
- Remove redundancies in headers parsing

### DIFF
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -9,7 +9,7 @@ from litestar._parsers import (
     parse_query_string,
     parse_url_encoded_form_data,
 )
-from litestar.datastructures import Headers
+from litestar.datastructures import Headers, MultiDict
 from litestar.datastructures.upload_file import UploadFile
 from litestar.enums import ParamType, RequestEncodingType
 from litestar.exceptions import ValidationException
@@ -155,7 +155,7 @@ def parse_connection_query_params(connection: ASGIConnection, kwargs_model: Kwar
     )
 
 
-def parse_connection_headers(connection: ASGIConnection, _: KwargsModel) -> Headers:
+def parse_connection_headers(connection: ASGIConnection, _: KwargsModel) -> MultiDict:
     """Parse header parameters and cache the result in scope.
 
     Args:
@@ -165,7 +165,7 @@ def parse_connection_headers(connection: ASGIConnection, _: KwargsModel) -> Head
     Returns:
         A Headers instance
     """
-    return Headers.from_scope(connection.scope)
+    return MultiDict(Headers.from_scope(connection.scope))
 
 
 def state_extractor(values: dict[str, Any], connection: ASGIConnection) -> None:
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Callable, Coroutine, NamedTuple, Mapping, cast
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, Mapping, NamedTuple, cast
 
 from litestar._multipart import parse_multipart_form
 from litestar._parsers import (
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -2,14 +2,14 @@ from __future__ import annotations
 
 from collections import defaultdict
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Callable, Coroutine, NamedTuple, cast
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, NamedTuple, Mapping, cast
 
 from litestar._multipart import parse_multipart_form
 from litestar._parsers import (
-    parse_headers,
     parse_query_string,
     parse_url_encoded_form_data,
 )
+from litestar.datastructures import Headers
 from litestar.datastructures.upload_file import UploadFile
 from litestar.enums import ParamType, RequestEncodingType
 from litestar.exceptions import ValidationException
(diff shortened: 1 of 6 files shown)


litestar/_kwargs/extractors.py
```diff
@@ -131,20 +131,17 @@ def parse_connection_query_params(connection: ASGIConnection, kwargs_model: Kwar
...
 
     Returns:
-        A dictionary of parsed values
+        A Headers instance
     """
-    parsed_headers = connection.scope["_headers"] = (  # type: ignore
-        connection._headers if connection._headers is not Empty else parse_headers(tuple(connection.scope["headers"]))
-    )
-    return cast("dict[str, Any]", parsed_headers)
+    return Headers.from_scope(connection.scope)
```
provinzkraut, line 144: This parsing was redundant; `Headers.from_scope` now stores itself in the `scope`

euri10:
> This is non-breaking as `Headers` behaves exactly like a regular dict outside for its additional methods
 
genuine question: is it really non-breaking since one doesnt care of case-sensitivity and the other does ?

litestar/_parsers.py
```diff
@@ -66,25 +65,3 @@ def parse_cookie_string(cookie_string: str) -> dict[str, str]:
         )
     }
     return output
-
```
provinzkraut, line 69: No need for these anymore. `Headers` does the conversion itself while also caching itself to the `scope` if invoked with `Headers.from_scope`. In addition, it retains duplicate keys, which were dropped in these methods.

provinzkraut:
> > This is non-breaking as `Headers` behaves exactly like a regular dict outside for its additional methods
> 
> genuine question: is it really non-breaking since one doesnt care of case-sensitivity and the other does ?

Good point. I guess you could say this is breaking, but it's only breaking the buggy behaviour. The headers that are coming in are all lower case, and when encoding them for the response they will all be lowercased as well. The only issue would be if you're relying on the case sensitivity within your route handler.

provinzkraut:
@euri10 On second thought, yes, you're right, this is breaking. Which is unfortunate, since it's also a bugfix. I guess this raises the general question of how we want to handle breaking bugfixes.

@litestar-org/members @litestar-org/maintainers, thoughts?

JacobCoffee:
> If it's something that has to be absolutely fixed, then we should fix it and release a new major version.

provinzkraut:
I have updated the PR to be non-breaking. This keeps most of the refactoring but does not change the type of the `headers` passed into route handlers.

cofin (approved):
lgtm

eldano1995:
I think this caused a breaking change in v2.2.1 of Litestar. Here's the config from my **pipfile.lock**:
```
        "litestar": {
            "hashes": [
                "sha256:cf93b913d620866c5de6b85ef4480a85816c83f9efacb49d2a06fb4f7d802d5a",
                "sha256:f8b6100bb6a0f4204d9b9a187a0dc5ea1a88dac29f6d0d83136497979f8a7e74"
            ],
            "index": "pypi",
            "markers": "python_version >= '3.8' and python_version < '4.0'",
            "version": "==2.2.1"
        },
```

Basically, I had something like this:
```
from msgspec import Struct

   class GetOrCreatePropertyOwners(Struct):
      property_id: int | None
      owner_names: list[str]

    @put("property-owners", guards=[valid_api_key_guard])
    async def get_or_create_property_owners(
        self,
        data: GetOrCreatePropertyOwners,
        headers: Headers,
    ) -> Response[PropertyOwnersResponse]:
        pass
```

and this is the error I was getting:
```
[my-app] Traceback (most recent call last):
[my-app]   File "/root/.local/share/virtualenvs/backends-Aysw1nYw/lib/python3.11/site-packages/litestar/_signature/model.py", line 93, in _deserializer
[my-app]     return default_deserializer(target_type, value)
[my-app]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[my-app]   File "/root/.local/share/virtualenvs/backends-Aysw1nYw/lib/python3.11/site-packages/litestar/serialization/msgspec_hooks.py", line 118, in default_deserializer
[my-app]     raise TypeError(f"Unsupported type: {type(value)!r}")
[my-app] TypeError: Unsupported type: <class 'dict'>
[skipsherpa-dev-public-api]
[my-app] The above exception was the direct cause of the following exception:
[skipsherpa-dev-public-api]
[my-app] Traceback (most recent call last):
[my-app]   File "/root/.local/share/virtualenvs/backends-Aysw1nYw/lib/python3.11/site-packages/litestar/_signature/model.py", line 190, in parse_values_from_connection_kwargs
[my-app]     return convert(kwargs, cls, strict=False, dec_hook=deserializer, str_keys=True).to_dict()
[my-app]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[my-app] msgspec.ValidationError: Unsupported type: <class 'dict'> - at `$.headers`
```

provinzkraut:
@eldano1995 Thanks for bringing this up! I think it's actually the docs that are wrong though; `headers` will receive a `dict[str, str]`, and not an instance of `Headers`. This was already the case before this change, it's just that this is now getting validated correctly.

eldano1995:
@provinzkraut Thanks for you quick reply!

Indeed, having it as `headers: dict[str,str]` fixed it. 

> I think it's actually the docs that are wrong though

However, I can confirm that before, at least on version 2.1.0, it was working with `headers: Headers`. But no problem: I'll just go with `dict[str, str]` from now on.

provinzkraut:
> @provinzkraut Thanks for you quick reply!
> 
> Indeed, having it as `headers: dict[str,str]` fixed it.
> 
> > I think it's actually the docs that are wrong though
> 
> However, I can confirm that before, at least on version 2.1.0, it was working with `headers: Headers`. But no problem: I'll just go with `dict[str, str]` from now on.
> 
> Thanks again!

Maybe I'm missing something but I think it *should* not have worked before, so if it did that was kind of a bug or unexpected behaviour, since it should have been passed as a `dict` as well there.

PR #2404, opened 2023-10-05
feat: allow customization of Pydantic integration

### LINKED ISSUE #2373: Bug: schema not respecting alias generator or alias on field
### Description

When using a Pydantic model with an alias generator function, like camel case, or a field with an alias, and using `model_dump(by_alias=True)` in `type_encoders` for `BaseModel`, the schema doesn't show the aliased key format. It'll still use snake case, matching the class attributes. This seems to be a regression from `2.0.0alpha7` (currently testing this on `2.1.1`).

https://github.com/rseeley/litestar-mcves/blob/mcve/camel-schema/app.py

### MCVE

_No response_

### Steps to reproduce

```bash
1. Clone repo above
2. Run installation/run steps in the README
3. Create a Pydantic model with either an alias generator or aliased Field (or both, as in the GH reproduction)
4. Go to `/async` to see that the alias is returning properly in actual responses
5. Go to `/schema` to see that the keys are snake case
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.1.1

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

---

## Funding

<a href="https://polar.sh/litestar-org/litestar/issues/2373">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2373/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
### Pull Request Checklist

### Description

- This PR adds a new `PydanticPlugin` that allows for customization of the Pydantic integration.  Currently, the only feature is to let a user export a model by alias.  When setting this, it applies to OpenAPI and responses.

### Close Issue(s)

- Closes #2373

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -341,9 +341,7 @@ class Litestar(Router):
             opt=dict(opt or {}),
             parameters=parameters or {},
             pdb_on_exception=pdb_on_exception,
-            plugins=[
-                *self._get_default_plugins(list(plugins or [])),
-            ],
+            plugins=self._get_default_plugins(list(plugins or [])),
             request_class=request_class,
             response_cache_config=response_cache_config or ResponseCacheConfig(),
             response_class=response_class,
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -341,7 +341,9 @@ class Litestar(Router):
             opt=dict(opt or {}),
             parameters=parameters or {},
             pdb_on_exception=pdb_on_exception,
-            plugins=[*(plugins or []), *self._get_default_plugins()],
+            plugins=[
+                *self._get_default_plugins(list(plugins or [])),
+            ],
             request_class=request_class,
             response_cache_config=response_cache_config or ResponseCacheConfig(),
             response_class=response_class,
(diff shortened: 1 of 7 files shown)


cofin:
Addresses regression in #2150 

Resolves #2373

litestar/contrib/pydantic/__init__.py
```diff
@@ -20,5 +24,32 @@ def _model_dump(model: pydantic.BaseModel, *, by_alias: bool = False) -> dict[st
...
+class PydanticPlugin(InitPluginProtocol):
+    """A plugin that provides Pydantic integration."""
+
+    __slots__ = ("prefer_alias",)
+
+    def __init__(self, prefer_alias: bool = False) -> None:
+        """Initialize ``PydanticPlugin``.
+
+        Args:
+            prefer_alias: OpenAPI and type_encoders will export by alias
```
JacobCoffee, line 42:
```suggestion
            prefer_alias: OpenAPI and ``type_encoders`` will export by alias
```

litestar/app.py
```diff
@@ -466,18 +468,24 @@ def serialization_plugins(self) -> list[SerializationPluginProtocol]:
         return list(self.plugins.serialization)
 
     @staticmethod
-    def _get_default_plugins() -> list[PluginProtocol]:
-        default_plugins: list[PluginProtocol] = []
+    def _get_default_plugins(plugins: list[PluginProtocol] | None = None) -> list[PluginProtocol]:
```
JacobCoffee, line 471: Is there a better place to do this it seems like we would keep adding things to it and this would become quite long.. or just it just stop at `attrs` and `pydantic` plugs?

JacobCoffee (approved):

litestar/app.py
cofin: I think we should refactor this if we add more than these, but, at the moment, it's just these two.

Alc-Alc (approved):
Existing and new tests are passing, LGTM

litestar/app.py
```diff
@@ -466,18 +466,24 @@ def serialization_plugins(self) -> list[SerializationPluginProtocol]:
...
+        if plugins is None:
+            plugins = []
         with suppress(MissingDependencyException):
-            from litestar.contrib.pydantic import PydanticInitPlugin, PydanticSchemaPlugin
-
-            default_plugins.extend((PydanticInitPlugin(), PydanticSchemaPlugin()))
+            from litestar.contrib.pydantic import PydanticInitPlugin, PydanticPlugin, PydanticSchemaPlugin
 
+            pre_configured = any(
+                isinstance(plugin, (PydanticPlugin, PydanticInitPlugin, PydanticSchemaPlugin)) for plugin in plugins
```
guacs, line 476: If the user provides a `PydanticInitPlugin` but not a `PydanticSchemaPlugin` (or vice-versa) then it won't get added right? Should this be handled?

cofin: My thought process was that if the user has supplied any plugin, then they likely want to override the default behavior.  Maybe that's not a good way to think about it though?

provinzkraut (approved):
This seems to be the (more) correct way to address these. Good work Mr. @cofin!

litestar/app.py
guacs: If they give `PydanticPlugin` then yeah I agree with you. As a user I wouldn't expect passing in a custom `PydanticInitPlugin` to disable the `PydanticSchemaPlugin`.

cofin: Let me see whatI an do to address this.

cofin: @guacs I think this should handle this scenario properly now.

PR #2417, opened 2023-10-07
fix: add `uvicorn[standard]` as a dependency of the `standard` extra

provinzkraut:
Add `uvicorn[standard]` as a dependency of the `standard` extra, replacing the previously manually added uvicorn extras. This resolves an install issue with Windows.


cofin (approved):
lgtm

PR #2418, opened 2023-10-07
feat(params): Improve error messages for missing required parameters

provinzkraut:
Add the source of the expected parameter to the error message.

Before:
```json
{
  "status_code": 400,
  "detail": "Missing required parameter foo for url http://testerver.local"
}
```

After:
```json
{
  "status_code": 400,
  "detail": "Missing required header parameter 'foo' for url http://testerver.local"
}
```


Alc-Alc (approved):
that was fast 🙂

PR #2427, opened 2023-10-12
chore: upgrade tooling & type fixes.

peterschutt:
Upgrades mypy to 1.6 & pyright to 1.331.

### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, Callable, Generator, Iterable, Sequence, 
 
 from rich import get_console
 from rich.table import Table
-from typing_extensions import Concatenate, ParamSpec, get_type_hints
+from typing_extensions import ParamSpec, get_type_hints
 
 from litestar import Litestar, __version__
 from litestar.middleware import DefineMiddleware
@@ -247,7 +247,7 @@ class LitestarExtensionGroup(LitestarGroup):
         return super().list_commands(ctx)
 
 
-def _inject_args(func: Callable[P, T]) -> Callable[Concatenate[Context, P], T]:
+def _inject_args(func: Callable[P, T]) -> Callable[P, T]:
     """Inject the app instance into a ``Command``"""
     params = inspect.signature(func).parameters
 
@@ -271,7 +271,7 @@ def _inject_args(func: Callable[P, T]) -> Callable[Concatenate[Context, P], T]:
 
         return func(*args, **kwargs)
 
-    return pass_context(wrapped)  # type: ignore[arg-type]
+    return pass_context(wrapped)
 
 
 def _wrap_commands(commands: Iterable[Command]) -> None:
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -271,7 +271,7 @@ def _inject_args(func: Callable[P, T]) -> Callable[Concatenate[Context, P], T]:
 
         return func(*args, **kwargs)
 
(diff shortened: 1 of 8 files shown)


peterschutt:
Regards the `disableBytesTypePromotions` flag ([ref](https://github.com/microsoft/pyright/issues/6129#issuecomment-1758697477)):

> The type annotation bytes has historically implied bytes | bytearray | memoryview (similar to how the type annotation float implies float | int). That means a value annotated as bytes might actually be a bytearray or memoryview at runtime. If you use an isinstance type guard to validate that it is actually a bytes value, it may still be a bytesarray | memoryview in the fall-through (else) case. That's where these types arise (seemingly out of nowhere).
> 
> [PEP 688](https://peps.python.org/pep-0688/#no-special-meaning-for-bytes) recommends that type checkers no longer treat bytes in this manner, but some existing code bases still rely on this behavior. I recently added a configuration switch called disableBytesTypePromotions. If you set this to true in your config file, pyright will no longer assume that a bytes annotation should also imply bytearray | memoryview. The disableBytesTypePromotions setting is true by default in "strict" type checking mode, but it's false by default in normal type checking mode. At some point in the future, I hope to default this to true all the time.

litestar/cli/_utils.py
```diff
@@ -271,7 +271,7 @@ def wrapped(ctx: Context, /, *args: P.args, **kwargs: P.kwargs) -> T:
 
         return func(*args, **kwargs)
 
-    return pass_context(wrapped)  # pyright: ignore
+    return pass_context(wrapped)  # type: ignore[arg-type]
```
peterschutt, line 274:
```
litestar/cli/_utils.py:274: error: Argument 1 to "pass_context" has incompatible type "Callable[[Context, **P], T]"; expected "Callable[[Context, Context, **P], T]"  [arg-type]
```
I ran out of time to debug this, but I'm wondering if a mypy bug given that pyright is OK with this now. Only guessing though.

peterschutt: The issue was that the decorator `pass_context()` returns a callable that doesn't include `Context` as the first argument, but the `_inject_args()` func return annotation was `Callable[Concatenate[Context, P], T]`.

litestar/security/session_auth/middleware.py
```diff
@@ -113,7 +113,7 @@ async def authenticate_request(self, connection: ASGIConnection[Any, Any, Any, A
         Returns:
             :class:`AuthenticationResult <.middleware.authentication.AuthenticationResult>`
         """
-        if not connection.session or connection.session is Empty:
+        if not connection.session or connection.scope["session"] is Empty:
```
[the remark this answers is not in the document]
peterschutt: If setting `scope["session"]` to `Empty` is the same as the session not being set, then perhaps we should change `ASGIConnection.session` to raise if it is accessed when `scope["session"] == Empty`. At least that way we are sure to honor the type of the property. E.g.,:

```py
    @property
    def session(self) -> dict[str, Any]:
        """Return the session for this connection if a session was previously set in the ``Scope``

        Returns:
            A dictionary representing the session value - if existing.

        Raises:
            ImproperlyConfiguredException: if session is not set in scope.
        """
        if "session" not in self.scope or self.scope["session"] is Empty:  # <---- add this "or" clause
                "'session' is not defined in scope, install a SessionMiddleware to set it"
            )

        return cast("dict[str, Any]", self.scope["session"])
```

JacobCoffee (approved):

PR #2437, opened 2023-10-13
refactor: Centralize config of pyright.

peterschutt:
We should favor setting all exclusions and other configs in pyproject.toml. As such, this PR removes the `exclude` hook config for the pre-commit pyright. Ideally, there should be no difference between running `pre-commit run pyright --all-files` and `pdm run --venv 3.11 pyright`, and I propose that centralizing the config in this manner is the best way to achieve that.

This PR makes an effort to minimize the amount of excluded modules. So, in addition to moving the exclude config from .pre-commit-config.yaml to pyproject.toml, it also omits some of the exclusions and addresses the arising type issues.

### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/litestar/testing/client/async_client.py
+++ b/litestar/testing/client/async_client.py
@@ -156,7 +156,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def get(  # type: ignore [override]
@@ -195,7 +195,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def options(
@@ -234,7 +234,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def head(
@@ -273,7 +273,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def post(
@@ -324,7 +324,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def put(
@@ -375,7 +375,7 @@ class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
         )
 
     async def patch(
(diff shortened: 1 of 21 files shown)


peterschutt:
> Ideally, there should be no difference between running `pre-commit run pyright --all-files` and `pdm run --venv 3.11 pyright`

This is far from my experience, however. Even after removing `exclude` from pre-commit, `pdm run --venv 3.11 pyright` would report a lot more errors than pre-commit would.

litestar/testing/client/async_client.py
```diff
@@ -156,7 +156,7 @@ async def request(
             auth=auth,
             follow_redirects=follow_redirects,
             timeout=timeout,
-            extensions=extensions,
+            extensions=None if extensions is None else dict(extensions),
```
peterschutt, line 159: The reason for this, and all the others like it, is that we have typed the `extensions` param to the method `Mapping`, however httpx has typed it `MutableMapping` in their method that we call through.

pyproject.toml
```diff
@@ -98,6 +98,7 @@ dev = [
   "fsspec",
   "greenlet",
   "hypothesis",
+  "pydantic[email]>=2.0",
```
peterschutt, line 101: Are we happy doing this @litestar-org/maintainers? This means that we lock pydantic @ v2.4ish but suffer a lower version of piccolo than is current.

If good to go, I'll split this into a separate PR.

tests/unit/test_dto/test_factory/test_dataclass_dto.py
```diff
@@ -25,7 +25,7 @@ def fx_dto_type() -> type[DataclassDTO[Model]]:
     return DataclassDTO[Model]
 
 
-@pytest.mark.skipif(sys.version_info > (3, 8), reason="generic builtin collection")
+@pytest.mark.skipif(sys.version_info >= (3, 9), reason="generic builtin collection")
```
peterschutt, line 28: `> (3, 8)` means we skip this test for any py version gt 3.8.0, which is basically all of them.

pyproject.toml
peterschutt: hmm, maybe we can't: https://github.com/litestar-org/litestar/actions/runs/6507707828/job/17675629575?pr=2437#step:6:1

litestar/types/builtin_types.py
```diff
@@ -14,11 +14,12 @@
...
 
+NoneType: type[None] = type(None)
+
 try:
-    from types import NoneType, UnionType  # pyright: ignore
+    from types import UnionType  # pyright: ignore
 except ImportError:
-    NoneType: TypeAlias = type(None)  # type: ignore
-    UnionType: TypeAlias = Union  # type: ignore
+    UnionType: TypeAlias = Union  # type:ignore[no-redef]
```
sobolevn, line 22: nit

```suggestion
    UnionType: TypeAlias = Union  # type: ignore[no-redef]
```

litestar/testing/client/async_client.py
sobolevn: Do we prefer runtime overhead or `type: ignore` in this case?

sobolevn:
Thanks for doing this!

litestar/testing/client/async_client.py
peterschutt: The thing that swayed me is that its not technically safe if we allow an immutable mapping into our method and pass it through where a mutable one is expected, and being a test client I figure that the performance implications aren't significant.

However, I also get that there is likely to be few, if not zero occurrences of users actually passing an immutable mapping through to this in the wild. But if that is the argument that sways, we should probably just change our methods to `MutableMapping` also.

`httpx` only changed this from `Mapping` to `MutableMapping` a month ago - they did this on a minor version bump, but they are also zero-ver ATM.

litestar/types/builtin_types.py
peterschutt: thanks!

pyproject.toml
peterschutt: Created https://github.com/litestar-org/litestar/pull/2438 to discuss this.

provinzkraut (approved):

provinzkraut:
@peterschutt any reason you're not targeting `main` with this?

peterschutt:
> @peterschutt any reason you're not targeting `main` with this?

Having v2 in the env fixes quite a few type errors - I had the v2 pinning logic in here first before opening #2438

PR #2439, opened 2023-10-14
fix: Prevent `ASGIConnection.session` from returning `Empty`.

### LINKED ISSUE #2435: Bug: ASGIConnection.session incorrect type - may return `Empty`
If setting `scope["session"]` to `Empty` is the same as the session not being set, then perhaps we should change `ASGIConnection.session` to raise if it is accessed when `scope["session"] == Empty`. At least that way we are sure to honor the type of the property. E.g.,:

```py
    @property
    def session(self) -> dict[str, Any]:
        """Return the session for this connection if a session was previously set in the ``Scope``

        Returns:
            A dictionary representing the session value - if existing.

        Raises:
            ImproperlyConfiguredException: if session is not set in scope.
        """
        if "session" not in self.scope or self.scope["session"] is Empty:  # <---- add this "or" clause
                "'session' is not defined in scope, install a SessionMiddleware to set it"
            )

        return cast("dict[str, Any]", self.scope["session"])
```

_Originally posted by @peterschutt in https://github.com/litestar-org/litestar/pull/2427#discussion_r1356810821_

peterschutt:
The property is typed to return `dict[str, Any]` but we use `Empty` as a sentinel to instruct session middleware to clear the session.

This PR changes the behavior of `ASGIConnection.session` to raise an exception on session access, just as if the session key is not set at all if it would otherwise return `Empty`.

In addition, includes some related refactorings:

- Defines a new constant, `SCOPE_SESSION_KEY: Final = "session"` and replaces uses of `"session"` literal with that.
- Refactors tests/unit/test_connection/test_base.py to utilise fixtures.

Closes #2435

### Pull Request Checklist

### Description

-

### Close Issue(s)

-


litestar/connection/base.py
```diff
@@ -234,12 +234,12 @@ def session(self) -> dict[str, Any]:
...
             ImproperlyConfiguredException: if session is not set in scope.
         """
-        if "session" not in self.scope:
+        if SCOPE_SESSION_KEY not in self.scope or self.scope[SCOPE_SESSION_KEY] is Empty:
             raise ImproperlyConfiguredException(
                 "'session' is not defined in scope, install a SessionMiddleware to set it"
             )
 
-        return cast("dict[str, Any]", self.scope["session"])
+        return cast("dict[str, Any]", self.scope[SCOPE_SESSION_KEY])
```
peterschutt, line 242: This `cast()` hides quite a bit more, it seems.

`BaseScope.session` is typed as `EmptyType | Dict[str, Any] | DataContainerType | None`, where `DataContainerType` is `Struct | BaseModel | AttrsInstance | TypedDictClass | DataclassProtocol`.

I'm not sure where the scope might be swapped out for one of the model types like Struct et al., but it seems as though type checkers won't prevent us from letting that happen.. nor will they prevent this property from returning `None`.

litestar/security/session_auth/middleware.py
```diff
@@ -113,15 +114,15 @@ async def authenticate_request(self, connection: ASGIConnection[Any, Any, Any, A
...
             # the assignment of 'Empty' forces the session middleware to clear session data.
-            connection.scope["session"] = Empty
+            connection.clear_session()
             raise NotAuthorizedException("no session data found")
 
         user = await self.retrieve_user_handler(connection.session, connection)
 
         if not user:
-            connection.scope["session"] = Empty
+            connection.scope[SCOPE_SESSION_KEY] = Empty
```
peterschutt, line 125:
```suggestion
            connection.clear_session()
```

litestar/connection/base.py
provinzkraut: Hm. I'm wondering if we might want to change the scope type?

provinzkraut (approved):

litestar/connection/base.py
peterschutt: Or, we _could_ just type it correctly:

```py
    @property
    def session(self) -> ScopeSession:
```

It throws a stack of type errors but only one of those is outside of the test suite.

I'm pretty green when it comes to session related aspects of the code base, so I think I'll take a bit more time to understand the problem before pulling the trigger on anything here.

PR #2396, opened 2023-10-03
feat: add support for `Python@3.12`

JacobCoffee:
### Pull Request Checklist

[//]: # "By submitting this issue, you agree to:"
[//]: # "- follow Litestar's [Code of Conduct](https://github.com/litestar-org/.github/blob/main/CODE_OF_CONDUCT.md)"
[//]: # "- follow Litestar's [Contribution Guidelines](https://litestar.dev/community/contribution-guide)"

- [ ] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR

### Description
[//]: # "Please describe your pull request for new release changelog purposes"
[//]: # "Example: 'This pr adds the capability to use server-sent events, and documents it in the usage docs'"

- Add Python 3.12 as a target - [3.12 Release Schedule](https://peps.python.org/pep-0693/#schedule) is slated to arrive within the next 3 months, with RC 1 coming at the first of next week

### Close Issue(s)
[//]: # "Please add in issue numbers this pull request will close, if applicable"
[//]: # "Examples: Fixes #4321 or Closes #1234"

- Closes #1862

PR #2453, opened 2023-10-16
build: specify uvloop only for non-windows builds

peterschutt:
### Pull Request Checklist

### Description

uvloop does not support windows.

### Close Issue(s)

-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,6 @@ jobs:
       python-version: ${{ matrix.python-version }}
 
   test-platform-compat:
-    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       python-version: ${{ matrix.python-version }}
 
   test-platform-compat:
+    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ Twitter = "https://twitter.com/LitestarAPI"
 annotated-types = ["annotated-types"]
 attrs = ["attrs"]
 brotli = ["brotli"]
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -73,7 +73,7 @@ Twitter = "https://twitter.com/LitestarAPI"
 annotated-types = ["annotated-types"]
 attrs = ["attrs"]
 brotli = ["brotli"]
-cli = ["jsbeautifier", "uvicorn[standard]", "uvloop>=0.18.0"]
+cli = ["jsbeautifier", "uvicorn[standard]", "uvloop>=0.18.0; sys_platform != 'win32'"]
```
guacs, line 76: Why are we even specifying `uvloop`? Is there a reason that it needs to be `>=0.18.0`? If not, we can just allow `uvicorn` to decide whether to install `uvloop` or not since it's part of the `standard` extras of `uvicorn` as seen [here](https://github.com/encode/uvicorn/blob/40b99b8436c0c261e3a85d10e291424072946292/pyproject.toml#L36-L45).

.github/workflows/ci.yml
```diff
@@ -45,7 +45,6 @@ jobs:
       python-version: ${{ matrix.python-version }}
 
   test-platform-compat:
-    if: github.event_name == 'push'
```
guacs: Why is this being removed?

peterschutt: Temporary, just to get the windows compat CI to run - it has been reverted in latest commit.

pyproject.toml
peterschutt: It is for 3.12 compatibility. Uvicorn doesn't have the same constraint, hence the spec.

peterschutt: Found the commit for you: https://github.com/litestar-org/litestar/pull/2396/commits/34bc1957d39155a30d54de2b65d3ebfe21953630

guacs (approved):

.github/workflows/ci.yml
peterschutt: Interesting! Cheers.

PR #2457, opened 2023-10-16
refactor: centralize config of mypy

peterschutt:
### Pull Request Checklist

### Description

We should favor setting all exclusions and other configs in pyproject.toml. As such, this PR removes the exclude hook config for the pre-commit pyright. Ideally, there should be no difference between running `pre-commit run mypy --all-files` and `pdm run mypy`, and I propose that centralizing the config in this manner is the best way to achieve that.

This PR makes an effort to minimize the amount of excluded modules. So, in addition to moving the exclude config from .pre-commit-config.yaml to pyproject.toml, it also omits some of the exclusions and addresses the arising type issues.

### Close Issue(s)

-


pyproject.toml
```diff
@@ -92,6 +92,7 @@ sqlalchemy = ["advanced-alchemy>=0.2.2,<0.4.0"]
 standard = ["jinja2", "jsbeautifier", "uvicorn[standard]", "uvloop>=0.18.0; sys_platform != 'win32'", "fast-query-parsers>=1.0.2"]
 structlog = ["structlog"]
 
+[tool.pdm]
```
peterschutt, line 95:
```suggestion
```

pyproject.toml
```diff
@@ -120,19 +120,20 @@ docs = [
...
+    "slotscheck",
+    "sourcery",
+    "pyright",
+    "blacken-docs",
+    "asyncpg-stubs",
+    "types-beautifulsoup4",
+    "types-pytest-lazy-fixture",
+    "types-python-jose",
+    "types-pyyaml",
+    "types-redis",
```
peterschutt, line 136:
```suggestion
  "ruff",
  "black",
  "mypy",
  "pre-commit",
  "slotscheck",
  "sourcery",
  "pyright",
  "blacken-docs",
  "asyncpg-stubs",
  "types-beautifulsoup4",
  "types-pytest-lazy-fixture",
  "types-python-jose",
  "types-pyyaml",
  "types-redis",
```

pyproject.toml
```diff
@@ -235,20 +237,23 @@ warn_unused_configs = true
 warn_unused_ignores = true
 
 [[tool.mypy.overrides]]
-disable_error_code = "attr-defined"
-module = ["tests.*.test_sqlalchemy", "tests.*.test_sqlalchemy"]
-
-[[tool.mypy.overrides]]
-disable_error_code = ["arg-type", "misc", "valid-type", "var-annotated"]
-module = "tests.unit.test_contrib.test_sqlalchemy.test_dto"
+module = ["tests.examples.*", "tests.docker_service_fixtures"]
```
guacs, line 240: Why are these two modules being ignored?

guacs (approved):

guacs:
> Can I propose we remove mypy and pyright from pre-commit @litestar-org/maintainers @litestar-org/members?
> 
> Pre-commit passes the names of all files in the project to mypy individually, which overrides the `packages = ["litestar", "tests"]` config in `[tool.mypy]`. This causes errors like this one:
> 
> image
> 
> ...which aren't an issue if the command is `pdm run mypy`, because the packages config is respected.
> 
> Also, it would mean we'd no longer have to maintain 2 extra lists of our dependencies in the pre-commit file!
> 
> We can add individual jobs for these in actions that can run parallel, and add to the `lint` command in `Makefile`.

I'm not in favor of this tbh. I think there might be two other alternatives that we could consider:

1. Use the `files` as documented [here](https://[pre-commit](https://pre-commit.com/#hooks-files).com/#hooks-files) to only include `litestar` and `tests`.
2. Run a local pre-commit hook using `pdm run mypy` and `pdm run pyright`. This has the benefit of not needing to maintain the separate dependency lists.

pyproject.toml
peterschutt: They were always ignored by the pre-commit config. When I did the pyright version of this last week there were a stack more issues to resolve, and so picking my battles I chose to leave these two ignored. I've left them ignored in this PR so we are consistent with the pyright config.

If I remove:

image

Will be better to remove the excludes from both pyright and mypy in the same PR and fix for both I think. I'll open up an issue for it.

peterschutt:
> > Can I propose we remove mypy and pyright from pre-commit @litestar-org/maintainers @litestar-org/members?
> > Pre-commit passes the names of all files in the project to mypy individually, which overrides the `packages = ["litestar", "tests"]` config in `[tool.mypy]`. This causes errors like this one:
> > image
> > ...which aren't an issue if the command is `pdm run mypy`, because the packages config is respected.
> > Also, it would mean we'd no longer have to maintain 2 extra lists of our dependencies in the pre-commit file!
> > We can add individual jobs for these in actions that can run parallel, and add to the `lint` command in `Makefile`.
> 
> I'm not in favor of this tbh. I think there might be two other alternatives that we could consider:
> 
>     1. Use the `files` as documented [here](https://%5Bpre-commit%5D(https://pre-commit.com/#hooks-files).com/#hooks-files) to only include `litestar` and `tests`.
> 
>     2. Run a local pre-commit hook using `pdm run mypy` and `pdm run pyright`. This has the benefit of not needing to maintain the separate dependency lists.

I'm not a fan of 1. as it means that we are moving back toward multiple configurations for the same thing.

I would prefer 2. over nothing I suppose.

Due to having mypy and pyright in pre-commit I've disabled the hook here, its just too slow & disruptive for my liking.

tests/helpers.py
```diff
@@ -56,8 +56,10 @@ def maybe_async_cm(obj: ContextManager[T] | AsyncContextManager[T]) -> AsyncCont
 def get_exception_group() -> type[BaseException]:
     """Get the exception group class with version compatibility."""
     try:
-        return cast("type[BaseException]", globals()["__builtins__"]["ExceptionGroup"])
+        eg = globals()["__builtins__"]["ExceptionGroup"]
     except KeyError:
```
sobolevn, line 60:
```suggestion
        eg = ExceptionGroup
    except NameError:
```

?

Or even `if sys.version_info >= (3, 11):`?
Current way is cryptic.

pyproject.toml
```diff
@@ -235,20 +237,23 @@ warn_unused_configs = true
...
+ignore_errors = true
 
 [[tool.mypy.overrides]]
 disallow_untyped_decorators = false
-module = "tests.*"
+module = ["tests.unit.test_kwargs.test_reserved_kwargs_injection"]
```
sobolevn, line 245: nit: let's first declare `module` and then all other settings. Right now it is inconsistent.

.pre-commit-config.yaml
```diff
@@ -61,7 +61,6 @@ repos:
     rev: "v1.6.0"
     hooks:
       - id: mypy
-        exclude: "test_apps|tools|docs|tests/examples|tests/docker_service_fixtures|tests/unit/test_contrib/test_pydantic/test_beanie_integration"
         additional_dependencies:
```
sobolevn, line 64: It would also be nice to somehow get rid of this duplication in the next PRs.

sobolevn (commented):
Thanks!

guacs:
>  it is non-obvious that when you run pre-commit run --all-files that the packages configuration from [tool.mypy] is not used

This is weird. I thought it would pick up the config from `pyproject.toml` since the default for `--config-file` includes `pyproject.toml`. For example, `ruff` does pick up from the `pyproject.toml`.

tests/helpers.py
peterschutt: You are right, it is weird to do it this way.

It originates from the pyright PR as for some reason they report
```
/litestar/tests/helpers.py:59:14 - error: "ExceptionGroup" is unbound (reportUnboundVariable)
```

I'll ignore the pyright error.

peterschutt: Oh, actually, it is a runtime error:

```
Traceback:
tests/helpers.py:59: in get_exception_group
    eg = ExceptionGroup  # pyright: ignore[reportUnboundVariable]
E   UnboundLocalError: cannot access local variable 'ExceptionGroup' where it is not associated with a value
```

peterschutt:
Another example of the complexity that pre-commit adds to running mypy.

pre-commit automagically runs mypy with the `--ignore-missing-imports` flag, which is not what we want, as we are explicitly building the exact environment via `additional_dependencies`.

This means we get an error for an `unused-ignore` when running through pre-commit, or if we remove the ignore, then we get an error for a missing import when not running via pre-commit. The latter being the correct warning, b/c the import is explicitly not in our 3.12 environment, which is the version we are running linting on.

In the readme for the pre-commit mirror it says to do this to override the args:

``` yaml
    hooks:
    -   id: mypy
        args: [--strict, --ignore-missing-imports]
```
Which I take to mean that if you want to preserve the `--ignore-missing-imports` and add other configs, then you need to ensure that you include it in the args.

We already have `strict = True` in pyproject.toml, so my hope is by simply passing `args: [--strict]` we will avoid the import ignore config. Unfortunately, that causes all of our top-level mypy configs from pyproject to be ignored, resulting in some 600+ errors. And so, now we are back to having to fully duplicate the configuration before we can indiscriminately run mypy inside and outside pre-commit.

Now, there might be a better way that I've missed, but the point is that the mental overhead all of this adds to what is otherwise a pretty simple tool to configure and run is just.... not fun.

Additionally, running mypy using a `local` hook, still means that we need to add pdm stuff to the validate hook so that we build the local environment. At this point, and with all the other arguments I've made, I'm -10 on running mypy via pre-commit.

PR #2463, opened 2023-10-17
fix(open-api): support generic types for OpenAPI schema generation

### LINKED ISSUE #2383: Bug: Missing API Schema for generic response type annotations
### Description

When I want to use a msgspec Struct to define a standard response struct for my project (`ResponseStruct`), after typed `Response[ResponseStruct[str]]` for return type, I got an empty API Schema.

### MCVE

```python
from msgspec import Struct
from litestar import Litestar, get, Response
from typing import TypeVar, Generic, Optional

T = TypeVar("T")


class ResponseStruct(Struct, Generic[T]):
    code: int
    data: Optional[T]


@get("/")
def test_handler() -> Response[ResponseStruct[str]]:
    return Response(
        ResponseStruct(code=200, data="Hello World"),
    )


app = Litestar([test_handler])
```

### Steps to reproduce

1. Run the MCVE
2. head on to [http://localhost:8000/schema/swagger](http://localhost:8000/schema/swagger)
3. You can now see an empty API Schema

### Screenshots

Empty API Schema

### Logs

```bash
Using Litestar app from app:app
Starting server process 
───────────────────────────────────────────────────────
┌──────────────────────────────┬──────────────────────┐
│ Litestar version             │ 2.1.1                │
│ Debug mode                   │ Disabled             │
│ Python Debugger on exception │ Disabled             │
│ CORS                         │ Disabled             │
│ CSRF                         │ Disabled             │
│ OpenAPI                      │ Enabled path=/schema │
│ Compression                  │ Disabled             │
└──────────────────────────────┴──────────────────────┘
INFO:     Started server process [100844]
```

### Litestar Version

```toml
litestar = {extras = ["cli"], version = "2.1.1"}
msgspec = "0.18.2"
```

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

guacs:
### Pull Request Checklist

### Description

- This PR adds support for generating the OpenAPI schema for generic types. 

### Close Issue(s)

- Fixes #2383


guacs:
The reason the generic types were not working was because `get_type_hints` does not resolve the generic types as seen [here](https://github.com/python/typing/issues/156). So a helper function was created to get the type hints after resolving for the generic types as much as possible. The relevant tests should make the expected behavior clear.

For the following handlers, the resulting schema will be as follows:

```python
from litestar import get, post
from litestar.app import Litestar

T = TypeVar("T")


@dataclass
class Foo(Generic[T]):
    foo: T


@get("/")
async def get_foo() -> Foo[int]:
    return Foo(1)


@get("/no-type")
async def get_foo_no_type() -> Foo:
    return Foo(1)


@post("/generic-type")
async def get_foo_generic(foo: Foo[T]) -> Foo[T]:
    return foo


app = Litestar([get_foo, get_foo_no_type, get_foo_generic])
```

image

guacs:
Generic Pydantic models do not work currently due to this [issue](https://github.com/pydantic/pydantic/issues/7837). I just realized (as I was typing this :P) a simple workaround for this would be to just get the `__parameters__` and `__args__` using the pydantic API and then use the `_substitute_typevars` to get the correct type hints.

I'll add this in along with tests to this PR itself.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -148,7 +155,19 @@ def _get_type_schema_name(value: Any) -> str:
...
+    if name := getattr(value, "__schema_name__", None):
+        return cast("str", name)
+
+    name = value.__name__
+    if args := getattr(value, "__args__", None):
+        name_parts = [name, "["]
+        name_parts.extend(_get_type_schema_name(a) for a in args)
+        name_parts.append("]")
+
+        return "".join(name_parts)
```
peterschutt, line 168: Is there a reason to not use `get_args()` here? Also, f-string, and do we need commas to separate the inner names?

```suggestion
    if args := get_args(value):
        inner_parts = ", ".join(_get_type_schema_name(a) for a in args)
        return f"{name}[{inner_parts}]"
```

litestar/utils/predicates.py
```diff
@@ -124,6 +124,9 @@ def is_dataclass_class(annotation: Any) -> TypeGuard[type[DataclassProtocol]]:
         ``True`` if instance or type of ``dataclass``.
     """
     try:
+        origin = get_origin_or_inner_type(annotation)
```
peterschutt, line 127: This is just for removing wrapper types like `Annotated` et al. isn't it? I.e., if `annotation` is `Annotated[list[int]]` then this will return `list[int]`? When I think about annotations and inner types, I think of `int` in `list[int]` being the inner type.

Not your issue, but do you think the name `get_origin_or_inner_type()` is a big confusing? `unwrap_and_get_origin()` would maybe be a better name. WDYT?

peterschutt (commented):
Sorry, ran out of time to complete my review, but a couple of comments so far. I'll pick up again later. Cheers!

litestar/_openapi/schema_generation/schema.py
guacs: You're absolutely right on both accounts. I should use `get_args` as well as f-strings.

litestar/utils/predicates.py
guacs: I definitely agree. I had to go read the code to make sure it really is only giving me back the origin class :P

Should this be included in this PR itself or a separate one? Also, is this public API? It's not there in the documentation, but it's also not under a private namespace of `_utils`. I'm leaning towards it's not public since it's not present in the docs.

peterschutt: > Should this be included in this PR itself or a separate one?

I think better ti do it separately  and keep this PR focussed, i was just curious what your thoughts on it were👍

> Also, is this public API? It's not there in the documentation, but it's also not under a private namespace of `_utils`. I'm leaning towards it's not public since it's not present in the docs.

I think we can rename the function that we use internally and alias it to the current name, with a deprecation for 3.0.

guacs: Yeah this sounds like a good plan. On that note, what do you think about moving some of the pydantic predicates etc into a `litestar.contrib.pydantic.predicates` module?

litestar/utils/typing.py
```diff
@@ -133,6 +135,12 @@
 wrapper_type_set = {Annotated, Required, NotRequired}
 """Types that always contain a wrapped type annotation as their first arg."""
 
+# `strict` was only introduced in 3.10
+if sys.version_info >= (3, 10):
+    _zip = partial(zip, strict=True)
```
provinzkraut: I think we shouldn't do this, since this means that the behaviour will be different for 3.10 and above.

litestar/utils/predicates.py
peterschutt: It would be nice if all of the pydantic stuff could exist inside the contrib/plugin eventually, but I don't think we should be importing from contrib into our core. So, while we are using funcs like `is_pydantic_constrained_field` and `is_pydantic_model_instance` outside of the contrib package, then we really need to keep them in there.

If we can get it to the point where our core doesn't have to care at all about third party libraries, it means we have the abstraction right and no coupling between our core and pydantic, or attrs or whatever. :pinched_fingers:

litestar/contrib/pydantic/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    unwrap_annotation,
+    wrapper_type_set,
+)
+
+if PYDANTIC_VERSION.startswith("2"):
+    # These utility functions are the same as the ones in litestar.utils.typing but this has to be
+    # done due to this issue: https://github.com/pydantic/pydantic/issues/7837. This is only for
+    # v2.
+
+    def pydantic_unwrap_and_get_origin(annotation: Any) -> Any | None:
```
peterschutt, line 22: We could save the duplication by making the callable that gets the origin an argument to `get_origin_or_inner_type()` in `utils.typing` without any impact on existing functionality, e.g,:

```py 
def get_origin_or_inner_type(annotation: T, get_origin_callable: Callable[[Any], Any] = get_origin) -> Any:
    origin = get_origin_callable(annotation)
    ...
```

Then could do something like:

```py
if PYDANTIC_VERSION.startswith("2"):
    get_origin_or_inner_type = partial(get_origin_or_inner_type(get_origin_callable=special_get_origin_for_pydantic_v2)
```

litestar/contrib/pydantic/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        origin = annotation.__pydantic_generic_metadata__["origin"]
+        if origin in wrapper_type_set:
+            inner, _, _ = unwrap_annotation(annotation)
+            origin = get_origin_or_inner_type(inner)
+        return instantiable_type_mapping.get(origin, origin)
+
+    def pydantic_get_type_hints_with_generics_resolved(
```
peterschutt, line 29: Same with this one, if the new `get_type_hints_with_generics_resolved()` had arguments that allowed overriding getters for args and parameters then I think you could avoid the duplication.

litestar/contrib/pydantic/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    pydantic_unwrap_and_get_origin = get_origin_or_inner_type
+
+    def pydantic_get_type_hints_with_generics_resolved(
+        annotation: Any,
+        globalns: dict[str, Any] | None = None,
+        localns: dict[str, Any] | None = None,
+        include_extras: bool = False,
+    ) -> dict[str, Any]:
+        if not is_generic(annotation):
+            return get_type_hints(annotation, globalns=globalns, localns=localns, include_extras=include_extras)
```
peterschutt, line 57: Why do we do this is this version of the func but not in `utils.typing` version?

litestar/utils/predicates.py
peterschutt: The more I look at this, even `unwrap_and_get_origin()` wouldn't be appropriate, as this implies that it behaves the same as `get_origin()` while ignoring "wrapper" types, but it doesn't because it returns the value from the `instantiable_type_mapping`. So it would have to be more like `get_unwrapped_instantiable_origin()` or something.

litestar/contrib/pydantic/utils.py
guacs: I missed updating this. I had changed the behavior to be the same as the one in `utils.typing` for Pydantic v2, but not v1. Nice catch!

litestar/utils/predicates.py
guacs: Well, this doesn't necessarily return an instantiable origin. It defaults to the origin if it's not part of the instantiable origin map. So, then the name would have to be `unwrap_and_get_instantiable_origin_if_it_exists_else_the_origin` 😆

guacs: I created issue #2467 so we can track this.

litestar/contrib/pydantic/utils.py
guacs: I do like this, but should we add this configurability to handle solely a bug with Pydantic. Imo, the `get_origin` call should return the right origin even for third-party libraries and if it doesn't it's a bug on their side? The reason I'm a bit reluctant is because I don't see a reason for using that `get_origin_callable` anywhere else.

Also, if and when Pydantic does fix this issue, then we could just remove all this Pydantic specific code and just use the original ones in `utils.typing`.

litestar/contrib/pydantic/utils.py
guacs: I think we should consider the same points I've mentioned above for this one as well. However, if we're going to add the arguments to get those details, how about just one callable `get_generic_metadata(annotation: Any) -> ??` which just returns a dictionary or tuple containing the origin, args and parameters?

litestar/utils/typing.py
guacs: That makes sense. I'll remove the strict check, because the other option is to implement zip ourselves with strict check for versions lower than 3.10, which I don't think is a good idea.

litestar/contrib/pydantic/utils.py
guacs: Just to be clear, I'm fine either way, but I think these are points we should consider before making the decision.

guacs:
The tests are failing for 3.12 with pydantic v1 because `get_type_hints` returns `__slots__` for reasons I'm not able to understand. Anyone have any ideas on how to handle this? Like should we special case this as well?

guacs:
> The tests are failing for 3.12 with pydantic v1 because `get_type_hints` returns `__slots__` for reasons I'm not able to understand. Anyone have any ideas on how to handle this? Like should we special case this as well?
> 
> This is only for Python 3.12 and pydantic v1. It's not some bug with Python 3.12 since `__slots__` are not returned for `dataclasses` with `slots=True`, or `msgspec`.
> 
> EDIT: added extra context

All the tests are passing if the relevant tests are skipped for this version combination.

litestar/contrib/pydantic/utils.py
peterschutt: Your point is valid - I did consider this and can see both sides. You are right, it is a shame to have to modify the method to support this, but at the same time, the modification really makes no functional difference to the current behaviour.

What about making the parameters to the functions private with leading underscore and undocumented so we can remove them freely if/when pydantic repair it.

Or, leave it - I'm OK with that too 👍

peterschutt:
> The tests are failing for 3.12 with pydantic v1 because `get_type_hints` returns `__slots__` for reasons I'm not able to understand. Anyone have any ideas on how to handle this? Like should we special case this as well?
> 
> This is only for Python 3.12 and pydantic v1. It's not some bug with Python 3.12 since `__slots__` are not returned for `dataclasses` with `slots=True`, or `msgspec`.
> 
> EDIT: added extra context

In pydantic v1, `BaseModel` inherits from `Representation` ([ref](https://github.com/pydantic/pydantic/blob/2e939dc3bfb88f54efb66f8f1a031ff22e4f9865/pydantic/main.py#L310)):

```py
class BaseModel(Representation, metaclass=ModelMetaclass):
    ...
```

`Representation` types `__slots__` as `Tuple[str, ...]` ([ref](https://github.com/pydantic/pydantic/blob/2e939dc3bfb88f54efb66f8f1a031ff22e4f9865/pydantic/utils.py#L368)):

```py
class Representation:
    """
    Mixin to provide __str__, __repr__, and __pretty__ methods. See #884 for more details.

    __pretty__ is used by [devtools](https://python-devtools.helpmanual.io/) to provide human readable representations
    of objects.
    """

    __slots__: Tuple[str, ...] = tuple()
```

I've confirmed that `__slots__` doesn't return in output from `get_type_hints()` in 3.8, but it does in 3.12.

The most recent changes to the `get_type_hints()` func are ~1 year old [ref](https://github.com/python/cpython/blame/3.12/Lib/typing.py#L2139) - so doubt the answer is in there..

On 3.12:

```py
Python 3.12.0 (main, Oct  7 2023, 15:26:19) [GCC 11.4.0] on linux
>>> from pydantic.utils import Representation
>>> Representation.__annotations__
{'__slots__': typing.Tuple[str, ...]}
>>> 

```

On 3.8:
```python-console
Python 3.8.17 (default, Sep 14 2023, 13:58:34) 
[GCC 11.4.0] on linux
>>> from pydantic.utils import Representation
>>> Representation.__annotations__
AttributeError: type object 'Representation' has no attribute '__annotations__'

```
It doesn't seem to be a case of cpython < 3.12 explicitly exluding `__slots__` from `__annotations__` or anything like that:

```python-console
Python 3.8.17 (default, Sep 14 2023, 13:58:34) 
[GCC 11.4.0] on linux
>>> from typing import Tuple
>>> class WithSlots:
...     __slots__: Tuple[str, ...]
... 
>>> WithSlots.__annotations__
{'__slots__': typing.Tuple[str, ...]}
```

So, somehow, `__annotations__` seems to get getting deleted from `Representation` in pydantic on < 3.12.. 

Looks like a difference between whether pydantic is compiled, or not:

```py
[GCC 11.4.0] on linux
>>> from pydantic.version import compiled
>>> compiled
False
>>> from pydantic.utils import Representation
>>> Representation.__annotations__
{'__slots__': typing.Tuple[str, ...]}
```

vs

```py
Python 3.8.17 (default, Sep 14 2023, 13:58:34) 
[GCC 11.4.0] on linux
>>> from pydantic.version import compiled
>>> compiled
True
>>> from pydantic.utils import Representation
>>> Representation.__annotations__
AttributeError: type object 'Representation' has no attribute '__annotations__'
```

So, maybe this is cython related..

However, even if we ignore this slots annotation, e.g., change return of `get_type_hints_with_Generics_resolved()` to:

```py
    return {n: _substitute_typevars(type_, typevar_map) for n, type_ in type_hints.items() if n != "__slots__"}
```

I can still get this error to raise by adding any concrete collection to the test model, e.g.:

```py
class PydanticGeneric(BaseModel, Generic[T]):
    foo: T
    optional_foo: Optional[T]
    annotated_foo: Annotated[T, object()]
    concrete_collection: Tuple[str]  # added this
```

```python-traceback
    def _check_generic(cls, parameters, elen=_marker):
        """Check correct count for parameters of a generic cls (internal helper).
        This gives a nice error message in case of count mismatch.
        """
        if not elen:
            raise TypeError(f"{cls} is not a generic class")
        if elen is _marker:
            if not hasattr(cls, "__parameters__") or not cls.__parameters__:
>               raise TypeError(f"{cls} is not a generic class")
E               TypeError: typing.Tuple[str] is not a generic class
```

So there's going to be 2 things to address here still, 1. how do we handle the variable inclusion/exclusoin of `__slots__` annotations, and 2., we should only be trying to substitute types that actually are generic.

guacs:
@peterschutt first of all, thank you for such a fantastic analysis!

Yeah so the reason for this is that in `_substitute_typevars`, the check for whether there are any values for `__parameters__` was incorrect. Essentially the substitution should only happen if there are any `__parameters__`. The current `_substitute_typevars` implementation should be changed to:

```python

def _substitute_typevars(obj: Any, typevar_map: Mapping[Any, Any]) -> Any:
    params = getattr(obj, "__parameters__", None)

    if params: # this was `if params is None` before but list[str].__parameters__ returns an empty tuple instead
        args = tuple(_substitute_typevars(typevar_map.get(p), typevar_map) for p in params)
        return obj[args]

    ....
```

This also resolves the issue related to `__slots__` being returned in `get_type_hints`! It doesn't matter even if `__slots__` is included in the type hints since the OpenAPI generation will be using the respective APIs to get the field names and then use that to get the annotation from the resolved type hints dictionary we get from `get_type_hints_with_generics_resolved`.

peterschutt:
> This also resolves the issue related to __slots__ being returned in get_type_hints! It doesn't matter even if __slots__ is included in the type hints since the OpenAPI generation will be using the respective APIs to get the field names and then use that to get the annotation from the resolved type hints dictionary we get from get_type_hints_with_generics_resolved

Oh, of course! Good stuff:)

tests/unit/test_openapi/test_schema.py
```diff
@@ -331,3 +340,84 @@ class MyDataclass:
         FieldDefinition.from_kwarg(name="MyDataclass", annotation=MyDataclass)
     )
     assert schemas["MyDataclass"].properties["bar"].items.const == 1  # type: ignore
+
+
+@dataclass
+class DataclassGeneric(Generic[T]):
+    foo: T
+    optional_foo: Optional[T]
+    annotated_foo: Annotated[T, object()]
```
peterschutt, line 349: Can we add a generic collection with optional variant, and perhaps even a concrete annotation, and concrete collection to these tests?

```suggestion
@dataclass
class DataclassGeneric(Generic[T]):
    concrete: int
    concrete_list: List[int]
    foo: T
    optional_foo: Optional[T]
    annotated_foo: Annotated[T, object()]
    bar: List[T]
    optional_bar: Optional[List[T]]
```

peterschutt (approved):
Nice work, thanks!

litestar/utils/predicates.py
peterschutt, line 127: Why are we using `get_origin_or_inner_type()` here, is it just to remove `Annotated` or other wrapper types?

Can we not just use `utils.typing.unwrap_annotation` for `annotation, _, _ = unwrap_annotation(annotation)`?

guacs: This was to handle cases like `GenericDataclass[int]`. Before it returned `False`, but now I think about it, it could be fine to get the caller to this function to ensure that it's the origin that is being passed. I'm fine with either approach tbh.

peterschutt: Oh ok, got you, so we actually do want some flavor of `get_origin()`.

guacs: Yeah, this same issue was there for `TypedDict` as well.

guacs: What do you think about splitting `unwrap_and_get_instantiable_origin_if_it_exists_else_the_origin` into two different functions: `unwrap_and_get_origin` and `unwrap_and_get_instantiable_origin`? In both cases `None` would be returned if there's not origin or instantiable origin, respectively.

peterschutt: So I guess the question is, should we consider a `_GenericAlias` that represents a dataclass type, as a dataclass type?

peterschutt: > 

What do you think about splitting unwrap_and_get_instantiable_origin_if_it_exists_else_the_origin into two different functions: unwrap_and_get_origin and unwrap_and_get_instantiable_origin? In both cases None would be returned if there's not origin or instantiable origin, respectively.

The thing is, this work has already been done on `FieldDefinition`, which is what we are using to call these methods anyway:

```py
# _openapi.schema_generation.schema
...
        elif is_dataclass_class(field_definition.annotation):
            result = self.for_dataclass(field_definition.annotation)
```

Here `field_definition` is an instance of `typing.FieldDefinition` and we have already calculated `origin`, it is available at `field_defintion.origin`. Also, `field_definition.annotation` is by definition already unwrapped (`field_definition.raw` being the original untouched annotation).

So, I'm starting to think we should be using that. Even maybe adding some extra predicates to that type, e.g., `FieldDefinition.is_dataclass_type()`, which could be:

```py
class FieldDefinition:
    def is_dataclass_type(self) -> bool:
        return is_dataclass_class(self.origin or self.annotation)
```

guacs: Oh, I hadn't looked too much into the `FieldDefinition`. I think adding the `is_dataclass_type` and the `is_typeddict_type` might be nice. The other predicates like `is_struct_class`, `is_pydantic_class` etc. all work since it does a `is_class_and_subclass` which does the unwrapping if needed.

guacs: Unfortunately, we will still be doing these checks again in the `for_*` methods in the `get_unwrapped_annotation_and_type_hints` helper function. To avoid this we could do the following:

- add a `get_type_hints` method to the `FieldDefinition` which will call `get_type_hints` or `get_type_hints_with_generics_resolved` since it already knows whether the annotation is generic or not.
- pass in the `FieldDefinition` instance to the `for_*` functions which is what we do for the plugins anyway

litestar/_openapi/schema_generation/schema.py
```diff
@@ -260,10 +277,21 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
...
+        # has to come before the `is_dataclass_check` since the the Pagination classes are dataclasses
+        # but we want to handle them differently from how dataclasses are normally handled.
         if field_definition.is_optional:
             result = self.for_optional_field(field_definition)
         elif field_definition.is_union:
             result = self.for_union_field(field_definition)
+        elif field_definition.is_generic and (
+            get_origin_or_inner_type(field_definition.annotation)
+            in (ClassicPagination, CursorPagination, OffsetPagination)
+        ):
```
peterschutt, line 291: `field_definition.is_generic` will return `True` for any of the pagination types.

If the pagination type isn't specialized, e.g., `ClassicPagination` not `ClassicPagination[int]` then `get_origin_or_annotation()` returns `None`, which is the same behavior as `field_definition.origin`. That means this branch will only be executed if `field_definition.origin` is one of the pagination types. Also, we don't need the unwrapping behavior of `get_origin_or_inner_type()` as `field_definition.annotation` is already unwrapped.

So I think this can just be:

```suggestion
        elif field_definition.origin in (ClassicPagination, CursorPagination, OffsetPagination):
```

litestar/_openapi/schema_generation/schema.py
```diff
@@ -260,10 +277,21 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
...
+            result = self.for_builtin_generics(field_definition)
+        elif isinstance(field_definition.annotation, TypeVar):
```
peterschutt, line 293:
```suggestion
        elif field_definition.is_type_var:
```

litestar/_openapi/schema_generation/schema.py
peterschutt: Another general discussion point, interested to hear your thoughts. It was surprising to me to find that `FieldDefinition.is_generic` is specifically a subclass check for `typing.Generic`.

When I think about whether a type is generic or not, that's not usually what I'm thinking of - maybe this should be `is_typing_generic` or `is_user_defined_generic`, b/c I consider `list[T]` generic, but  this property would return `False` for that.

litestar/utils/predicates.py
peterschutt: That sounds reasonable. Given a type, `FieldDefinition(typ).get_type_hints()` would basically be the same as calling `typing_extensions.get_type_hints(typ)`?

You could make the generic resolution optional via flag?

```py
class FieldDefinition:
    def get_type_hints(self, include_extras: bool = False, resolve_generics: bool = False) -> dict[str, Any]: ...
```

guacs: Okay so I was thinking of the following changes:
- remove the `get_origin_or_inner_type` from the `is_dataclass_class` and `is_typed_dict` and document in the docstrings that if the type is generic, then the origin is what has to be given
- add `is_dataclass_type` and `is_typeddict_type` helpers to the `FieldDefinition`
- add `get_type_hints` with the above signature you've mentioned
- pass the `FieldDefinition` to the `for_*` methods on the schema generation which means the `get_unwrapped_annotation_and_type_hints` can be removed

Sounds good?

litestar/_openapi/schema_generation/schema.py
guacs: You're right that this does feel a bit misleading. I would expect the `list[T]` to be considered generic as well. We could rename the property, but I think a more appropriate thing would be to change the `is_generic` to return `True` for `list` and other in-built generic collections etc. as well.

litestar/utils/predicates.py
peterschutt: > remove the get_origin_or_inner_type from the is_dataclass_class and is_typed_dict and document in the docstrings that if the type is generic, then the origin is what has to be given

How about we leave them alone and deprecate them in favor of `FieldDefinition.{is_dataclass_type,is_typeddict_type}` in-case anyone is using them, and re-implement the logic in `FieldDefinition` methods. They are pretty light-on on functionality.

The rest SGTM!

guacs: If we deprecate them, what about cases where we or the user needs to check if a type is a dataclass/typeddict class, but it's not wrapped in a `FieldDefinition` instance?

peterschutt: My opinion is that we should strive to have one way to interact with types internally, via `FieldDefinition`. Much of the untidiness we've seen in this area of the code-base (pre-existing, not yours) stems from the fact that we have both the OO approach to reasoning about our types (`FieldDefinition`), and the functional approach via `utils.typing`/`utils.predicates` etc. And we've seen in our discussion here how that is leading to things like duplication of functionality and making code hard to reason about.

While the functions I propose to deprecate aren't in our API docs, they also aren't in any sort of private namespace - which is why I don't think we should change their behavior. And, if we aren't using functionality ourselves internally (the only place we use them is for this openapi logic), I don't think there is merit in carrying them forward ad infinitum, hence my suggestion to deprecate in favor of the `FieldDefinition` properties.

We'll still have a public interface through which users can leverage our type introspection logic, which would be via `FieldDefinition` and users can easily implement the functions themselves if they don't want to go that route.

guacs: Yeah, that sounds good to me. I think it will be good if we start enforcing the use of `FieldDefinition` in our internal API.

PR #2474, opened 2023-10-20
build: configure type checkers to run on min supported version

peterschutt:
### Pull Request Checklist

### Description

This PR modifies our type-checking to run on the minimum supported version, instead of the latest.

Rationale:
- Reduce incidence of backward incompatible changes.
- Improve local development workflow - prior to this change, local development requires 2 python environments, 1 for 3.8 for dev/running tests, and one for 3.12 to run the type checkers. After this change developers will only have to run a single python environment for running both tests and type-checks.

### Close Issue(s)

-


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.12"
+          python-version: "3.8"
```
peterschutt, line 18:
```suggestion
          python-version: "3.12"
```

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
   "multidict>=6.0.2",
   "polyfactory>=2.6.3",
   "pyyaml",
-  "typing-extensions",
+  "typing-extensions>=4.6.0",
```
guacs, line 42: Why do we need this constraint?

tests/helpers.py
```diff
@@ -16,7 +16,7 @@
 if sys.version_info < (3, 9):
 
     def randbytes(n: int) -> bytes:
-        return bytearray(RANDOM.getrandbits(8) for _ in range(n))
+        return bytes(bytearray(RANDOM.getrandbits(8) for _ in range(n)))
```
guacs, line 19:
```suggestion
        return bytes(RANDOM.getrandbits(8) for _ in range(n))
```

Is there a reason it needs to be put into a `bytearray` and then into `bytes`?

guacs (approved):

JacobCoffee (commented):
Maybe can suggest RTX as well

JacobCoffee (approved):
Are these required for end users or just available? Sorry I haven't caught up to this new change

.github/workflows/ci.yml
```diff
@@ -38,13 +38,13 @@ jobs:
...
```
sobolevn, line 41: We can add this to `mypy.ini` / `pyproject.toml` instead. So, we won't have to duplicate it and add a detailed doc about it.

pyproject.toml
sobolevn: I propose to leave this alone, because typing-exntensions is a common source of resolution problems. **Unless** we relly on some details in the implementation.

.github/workflows/ci.yml
peterschutt: Are you referring to this? https://github.com/actions/setup-python/pull/669? Given that we specify a range in `requires-python`, will it always select the lowest or is there some way to configure that?

pyproject.toml
peterschutt: We use `Buffer` in an isinstance check which was added in 4.6. If the use of the constraint outweighs the utility of using the type I'll remove its use. Let me know what you think.

.github/workflows/ci.yml
sobolevn: No, I was thinking about adding this flag https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version

sobolevn: Together / instead of this one

tests/helpers.py
peterschutt: > Is there a reason it needs to be put into a `bytearray` and then into `bytes`?

Nothing but absentmindedness, unfortunately.

peterschutt:
> Maybe can suggest RTX as well

In `CONTRIBUTING`? I don't have any experience with it - if you'd like to add something, I'm open to it.

pyproject.toml
guacs: This check is only being done in the `pydantic_init_plugin` right? If that's the case, I think we can safely remove the constraint since `pydantic` sets the minimum constraint for `typing_extensions` to `4.6.1`. Or we could change that to `isinstance(value, (bytes, bytearray, memoryview))`.

peterschutt: > sets the minimum constraint for typing_extensions to 4.6.1

This is the clincher for me, we'll just remove the constraint on our side :heavy_check_mark:

CONTRIBUTING.rst
```diff
@@ -4,6 +4,15 @@ Contribution guide
...
 --------------------------
 
+.. tip:: Choosing your Python version.
+
+   We strongly encourage setting up your environment to use the lowest version of Python that is supported by
+   Litestar. This will ensure that the changes you made are backward compatible and will not fail in CI. This will also
+   ensure that when you run type checkers locally, you will get the same results as CI. Doing this will save you time!
+
+   The lowest currently supported version is Python 3.8. You can use `pyenv <https://github.com/pyenv/pyenv>`_ to manage
+   multiple Python versions on your system.
```
sobolevn, line 14:
```suggestion
.. tip:: Supported Python versions.

   The lowest currently supported version is Python 3.8. You can use `pyenv <https://github.com/pyenv/pyenv>`_ to manage
   multiple Python versions on your system.

   We also use the lowest supported version in our type-checking CI,
   this ensures that the changes you made are backward compatible.
```

sobolevn (commented):
Last final nit :)

sobolevn (approved):
Thank you! Great work!

PR #2482, opened 2023-10-21
fix: change URL.with_replacements to use Empty

nilsso:
### Pull Request Checklist

### Description

Small modification to `URL.with_replacements` using the `Empty`/`EmptyType` sentinel as the `query` parameter default, allowing for scenarios like replacing after having popped from an existing dict with only one item. For this example

```py
url = URL.from_components(query="foo=1")
query_params = url.query_params
query_params.pop("foo")
print(query_params)  # "<MultiDict()>"
```

We currently have this behavior:

```py
print(url.with_replacements(query=query_params)  # "?foo=1"
```

But with change:
```py
print(url.with_replacements(query=query_params)  # ""
```


peterschutt (changes requested):
So the change is that passing an empty dict to `query` will force replacement of the existing URL query string with an empty one? This makes sense to me.

Can you add a test that demonstrates the behavior please? You should be able to add another parametrized test to `tests.unit.test_datastructures.test_url.test_url_with_replacements()`.

nilsso:
> So the change is that passing an empty dict to `query` will force replacement of the existing URL query string with an empty one? This makes sense to me.
> 
> Can you add a test that demonstrates the behavior please? You should be able to add another parametrized test to `tests.unit.test_datastructures.test_url.test_url_with_replacements()`.

Added three more parameterized tests. But sorry I messed up the commit naming a bit.

peterschutt (approved):
Thanks!!

PR #2483, opened 2023-10-22
refactor: improve `template` module typing

peterschutt:
### Pull Request Checklist

### Description

We have issues with typing in our `template` module, as evidenced by the number of `type: ignore`'s, here: https://github.com/litestar-org/litestar/blob/56eaeeb4fbe49962d0c45e8a3b2bb15014fec78d/litestar/contrib/jinja.py#L52-L54

And here: https://github.com/litestar-org/litestar/blob/56eaeeb4fbe49962d0c45e8a3b2bb15014fec78d/litestar/contrib/mako.py#L88-L90

These, and other issues are caused by problems with the abstraction we've laid out for `TemplateEngineProtocol`.

The main issue is that the typing of `TemplateEngineProtocol.register_template_callable()` expects the `template_callable` parameter to be a callable that accepts a single `dict[str, Any]` positional argument. However, this is not an accurate constraint, in fact, the functions that we are passing to that method ourselves have more complex signatures than that. Not to mention, the `minijinja` engine doesn't even use a mapping type for its context object.

This PR:
- makes `TemplateEngineProtocol` generic on the context type in addition to the template type.
- Introduces a new type alias `TemplateCallableType` that concatenates a first argument context type with the remainder of a callables signature, so that we can express "callable that accepts context as first argument, with arbitrary additional args and kwargs".
- Changes the typing of `url_for`, `csrf_token` and `url_for_static_asset` to receive context typed as `Mapping[str, Any]` instead of the typed-dict `TemplateContext`. This lines the signature of these functions up with that accepted by `register_template_callable()`.
- Deprecates `TemplateContext` typed-dict. It is no-longer used internally.
- Adds some types to aid in typing minijinja (a stub for typing the `pass_state` decorator, a protocol for the state object).
- Adds a decorator for transforming minijinja state into a mapping required by `url_for` et al.
- Deprecates `minijinja_from_state` as it is no longer used internally.

### Close Issue(s)

-


litestar/contrib/minijnja.py
```diff
@@ -2,37 +2,73 @@
...
     raise MissingDependencyException("minijinja") from e
 
-from minijinja import Environment, pass_state
-from minijinja import TemplateError as MiniJinjaTemplateNotFound
-
 if TYPE_CHECKING:
-    from minijinja._lowlevel import State
+    from typing import Callable
+
     from pydantic import DirectoryPath
```
peterschutt, line 29: The fact that we are using a pydantic type in here means that our `minijinja` extra should really include pydantic too - but I think preferably we should look at why this is needed and try to decouple.

litestar/template/base.py
```diff
@@ -55,10 +65,10 @@ def csrf_token(context: TemplateContext) -> str:
     Returns:
         A CSRF token if the app level ``csrf_config`` is set, otherwise an empty string.
     """
-    return context["request"].scope.get("_csrf_token", "")  # type: ignore
+    return _get_request_from_context(context).scope.get("_csrf_token", "")  # type: ignore[no-any-return]
```
peterschutt, line 68: `Any` return here, because `Request.scope` is an instance of the typed-dict `HTTPScope` which doesn't have a `_csrf_token` key on it. I think that would be better off stored/retrieved from scope state using `utils.scope.{get_litestar_scope_state,set_litestar_scope_state}`.

pyproject.toml
```diff
@@ -250,8 +250,6 @@ module = [
   "brotli.*",
   "fsspec.*",
   "jsbeautifier.*",
-  "mako.*",
-  "minijinja.*",
```
peterschutt, line 254: Favored ignoring these in-line as more explicit, but on the fence. Happy to revert if others feel differently.

litestar/template/base.py
```diff
@@ -55,10 +65,10 @@ def csrf_token(context: TemplateContext) -> str:
     Returns:
         A CSRF token if the app level ``csrf_config`` is set, otherwise an empty string.
     """
-    return context["request"].scope.get("_csrf_token", "")  # type: ignore
+    return _get_request_from_context(context).scope.get("_csrf_token", "")  # type: ignore[return-value]
```
peterschutt, line 68: `object` return here, because Request.scope is an instance of the typed-dict HTTPScope which doesn't have a `_csrf_token` key on it. I think `_csrf_token` would be better off stored/retrieved from scope state using `utils.scope.{get_litestar_scope_state,set_litestar_scope_state}`.

litestar/contrib/minijnja.py
```diff
@@ -130,7 +160,9 @@ def get_template(self, template_name: str) -> MiniJinjaTemplate:
         except MiniJinjaTemplateNotFound as exc:
             raise TemplateNotFoundException(template_name=template_name) from exc
 
-    def register_template_callable(self, key: str, template_callable: Callable[[dict[str, Any]], Any]) -> None:
+    def register_template_callable(
+        self, key: str, template_callable: TemplateCallableType[StateProtocol, P, T]
+    ) -> None:
         """Register a callable on the template engine.
```
peterschutt, line 166: I believe the typing of this method was incorrect before these changes. Reason being that `template_callable` is constrained to receive a `dict` but we decorate with `pass_state` which means it will receive minijinja's `State` object.

provinzkraut (approved):

PR #2458, opened 2023-10-16
test: Remove Pydantic as default model from tests

provinzkraut:
Removes Pydantic as the default model from tests and replaces it with dataclasses.
Pydantic is now only used in cases where Pydantic functionality is being tested explicitly.

This is a preparation for #2419, to make testing a bit easier.

Issue #2278, opened 2023-09-04
Bug: forms validation discrepency between underlying object and form type


euri10:
### Description

Given this little app that basically enables in a browser to post a form to 6 different routes:  atrts/msgspec/pydantic times url-encoded/multi-part version, I have isues to understand why the multi-part pydantic vs attrs/msgspec different behaviour. 

I would expect a 201 on all and get a 400 on multi-part msgspec and attrs (see logs), is that intended ?

Input I pass is purposedely 1 and 1 for name and amount and the same for all POST logs below.

### MCVE

```python
from typing import Annotated

import msgspec
from attr import define, field
from attr.validators import ge, instance_of, lt
from jinja2 import Environment, DictLoader
from pydantic import BaseModel, ConfigDict, Field

from litestar import Litestar, Request, post, get
from litestar.contrib.jinja import JinjaTemplateEngine
from litestar.enums import RequestEncodingType
from litestar.params import Body
from litestar.response import Template
from litestar.template import TemplateConfig

MAX_INT_POSTGRES = 10


@define
class AddProductFormAttrs:
    name: str
    amount: int = field(validator=[instance_of(int), ge(1), lt(MAX_INT_POSTGRES)])


class AddProductFormPydantic(BaseModel):
    model_config = ConfigDict(arbitrary_types_allowed=True)
    name: str
    amount: int = Field(ge=1, lt=MAX_INT_POSTGRES)


class AddProductFormMsgspec(msgspec.Struct):
    name: str
    amount: Annotated[int, msgspec.Meta(lt=MAX_INT_POSTGRES, ge=1)]


@get(path="/")
async def form_index() -> Template:
    return Template(template_name="form.html", context={"templates": ["attrs", "pydantic", "msgspec"]})


@post(path="/form/attrs")
async def form_attrs(
        request: Request, data: Annotated[AddProductFormAttrs, Body(media_type=RequestEncodingType.URL_ENCODED)]
) -> int:
    print(type(data.name))
    return data.amount


@post(path="/form/pydantic")
async def form_pydantic(
        request: Request, data: Annotated[AddProductFormPydantic, Body(media_type=RequestEncodingType.URL_ENCODED)]
) -> int:
    print(type(data.name))
    return data.amount


@post(path="/form/msgspec")
async def form_msgspec(
        request: Request, data: Annotated[AddProductFormMsgspec, Body(media_type=RequestEncodingType.URL_ENCODED)]
) -> int:
    print(type(data.name))
[...]

PR #2486, opened 2023-10-22
refactor(tests): Refactor Pydantic and attrs tests

provinzkraut:
Follow-up to #2458.

- Move all Pydantic related tests to `tests/unit/test_contrib/test_pydantic`
- Move all attrs related tests to `tests/unit/test_contrib/test_attrs.py`
- Remove some outdated tests that were no longer needed


litestar/testing/request_factory.py
```diff
@@ -280,16 +286,7 @@ def _create_request_with_data(
...
-                data = to_builtins(data)
-            elif is_dataclass_instance(data):
-                data = asdict(data)
-            elif is_attrs_class(type(data)):
-                from attr import asdict as attrs_as_dict
-
-                data = attrs_as_dict(data)  # type: ignore[arg-type]
-            elif is_pydantic_model_instance(data):
-                data = data.model_dump(mode="json") if hasattr(data, "model_dump") else json.loads(data.json())
+            data = json.loads(encode_json(data, serializer=get_serializer_from_scope(scope)))
```
provinzkraut, line 289: Do a JSON roundtrip here (because Pydantic) with the regular encoder, so we can ensure this is actually encoded the same as a request, and we don't need special casing for every data type. Incidentally, this make it easier to test.

tests/unit/test_handlers/test_http_handlers/test_media_type.py
provinzkraut, line 1: We don't actually need to test the `pydantic.types.PaymentCardBrand` here, as it's just a `(str, Enum)` subtype, so we can simply test that directly

tests/unit/test_kwargs/test_attrs_data.py
provinzkraut, line 1: Moved to `tests/unit/test_contrib/test_attrs.py`

tests/unit/test_kwargs/test_header_params.py
```diff
@@ -39,19 +37,3 @@ def test_method(special_header: t_type = param) -> None:  # type: ignore
             assert response.status_code == HTTP_400_BAD_REQUEST, response.json()
         else:
             assert response.status_code == HTTP_200_OK, response.json()
-
-
-def test_header_param_example() -> None:
```
provinzkraut, line 44: This example no longer exists

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -493,31 +487,15 @@ def handler(
...
-    amount: int = Field(ge=1, lt=MAX_INT_POSTGRES)
-
-
 class AddProductFormMsgspec(msgspec.Struct):
     name: str
-    amount: Annotated[int, msgspec.Meta(lt=MAX_INT_POSTGRES, ge=1)]
+    amount: Annotated[int, msgspec.Meta(lt=10, ge=1)]
 
 
-@pytest.mark.parametrize("form_object", [AddProductFormMsgspec, AddProductFormPydantic, AddProductFormAttrs])
```
provinzkraut, line 516: We don't actually need to test the different libraries here. This test was added for #2278, but the discrepancies displayed there were expected and a result of the different strictness of the libraries, a difference we do want to preserve. The reason this test was failing before the fix was simply due to invalid input.

tests/unit/test_kwargs/test_validations.py
```diff
@@ -58,13 +57,10 @@ def test_raises_when_reserved_kwargs_are_misused(reserved_kwarg: str) -> None:
     with pytest.raises(ImproperlyConfiguredException):
         Litestar(route_handlers=[handler_with_dependency])
 
-    # these kwargs are set to Any when the signature model is generated,
-    # because pydantic can't handle generics for non pydantic classes. So these tests won't work for aliased parameters.
-    if reserved_kwarg not in SKIP_VALIDATION_NAMES:
```
provinzkraut, line 63: This restriction no longer exists, since we moved away from the Pydantic signature model.

tests/unit/test_middleware/test_session/test_integration.py
```diff
@@ -16,15 +15,17 @@
 from litestar.testing import create_test_client
 
 
-class User(BaseModel):
+@dataclass
+class User:
     id: UUID
     name: str
-    email: EmailStr
```
provinzkraut, line 22: No need for Pydantic specific tests here; We test handling of those custom types in the Pydantic plugin tests

tests/unit/test_signature/test_signature_validation.py
provinzkraut, line 1: This was meant to test the different signature modelling backends, which no longer exist.

tests/unit/test_testing/test_request_factory.py
```diff
@@ -81,6 +76,20 @@ async def test_request_factory_create_with_data(data_cls: DataContainerType) ->
     assert json.loads(body) == person_data
 
 
+async def test_request_factory_create_with_data_with_custom_encoder(mocker: MockerFixture) -> None:
```
provinzkraut, line 79: Instead of testing with different modelling libraries (for which the serialisation functionality is already tested elsewhere), what we actually want to test here is that the `RequestFactory` picks up the correct `type_encoders`.

tests/unit/test_dto/test_factory/test_integration.py
```diff
@@ -29,7 +27,7 @@
 
 
 def test_url_encoded_form_data(use_experimental_dto_backend: bool) -> None:
-    @dataclass
+    @dataclass()
```
peterschutt, line 30: Just a nit, but why braces for these decorators and not others?

peterschutt (approved):
Thanks!!

tests/unit/test_dto/test_factory/test_integration.py
provinzkraut: To be honest, I have no idea. Mypy was complaining when the braces were missing here :shrug:

peterschutt: 🤷‍♂️

Issue #2419, opened 2023-10-07
Enhancement: Support Pydantic 1 and 2 within the same app


provinzkraut:
### Summary

Since Pydantic 2 contains Pydantic 1, we should support Pydantic 1 and 2 within the same application so users can make use of Pydantic's backwards compatibility layer. 

We already have everything we need for this in the Pydantic plugin, it just need to be split up.
This could also simplify testing, since we could test v1 and v2 within the same environment.

peterschutt:
> This could also simplify testing, since we could test v1 and v2 within the same environment.

This would be nice!

Would this mean we would have to constrain to pydantic>=2? Or we would support v1 in either case that v1 and v2 is installed, and only v2 if v2 installed?

provinzkraut:
The latter. I did some testing with this a while ago and it seemed to work just fine.

peterschutt:
> only v2 if v2 installed?

Should read "only v1 if v1 installed"

provinzkraut:
> > only v2 if v2 installed?
> 
> Should read "only v1 if v1 installed"

That's actually how I read it :upside_down_face: 

So yes, the plan is to support v1 if v1 is installed, and v1 + v2 if v2 is installed.

geeshta:
@provinzkraut this should be closed because of #2487 no?

JacobCoffee:
Closed in 2487

PR #2487, opened 2023-10-22
feat: Support Pydantic 1 and 2 within the same application

provinzkraut:
Implements #2419.

Support Pydantic 1 and 2 by making use of its backwards compatibility layer; Pydantic 2 ships with version 1 in the `pydantic.v1` namespace. This allows us to support either both 1 & 2 if version 2 is installed or just version 1 if version 1 is installed.

In addition, this also simplifies our testing setup somewhat, as we can test for both Pydantic 1 & 2 without having to re-install dependencies.


litestar/contrib/pydantic/pydantic_dto_factory.py
```diff
@@ -16,21 +17,36 @@
...
+
+    ValidationErrorV2 = ValidationErrorV1  # type: ignore[assignment, misc]
+
+
+except ImportError as e:  # noqa: B025
     raise MissingDependencyException("pydantic") from e
+# isort: on
 
-__all__ = ("PydanticDTO",)
+if pydantic_v2 is not Empty:  # type: ignore[comparison-overlap]
```
provinzkraut, line 41: Mypy doesn't like this at all, but it also doesn't like "redefining" type aliases within an `if`/`else` branch so I'm not really sure how to best deal with this.

pyproject.toml
```diff
@@ -268,7 +268,9 @@ exclude = [
   "tools",
   "docs",
   "tests/examples",
-  "tests/docker_service_fixtures.py"
+  "tests/docker_service_fixtures.py",
+  "litestar/contrib/pydantic/pydantic_dto_factory.py",
```
provinzkraut, line 272: pyright complains about every instance where a pydantic import is used here because for some reason it think that it's a "variable"

litestar/constants.py
```diff
@@ -25,16 +25,3 @@
 UNDEFINED_SENTINELS: Final = {Signature.empty, Empty, Ellipsis, MISSING, UnsetType}
 WEBSOCKET_CLOSE: Final = "websocket.close"
 WEBSOCKET_DISCONNECT: Final = "websocket.disconnect"
-
-try:
```
provinzkraut, line 29: Moved this logic into `contrib.pydantic`

litestar/contrib/pydantic/utils.py
provinzkraut, line 1: Some utils and predicates were moved here so we don't have to deal with the pydantic imports in other modules

tests/unit/test_contrib/test_pydantic/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+from pytest import FixtureRequest
+
+
+@pytest.fixture(params=["v1", "v2"])
+def pydantic_version(request: FixtureRequest) -> str:
+    return request.param  # type: ignore[no-any-return]
+
+
+@pytest.fixture()
+def base_model(pydantic_version: str) -> type[pydantic.BaseModel | pydantic_v1.BaseModel]:
```
provinzkraut, line 15: Not sure if doing it like this is the best way or if we want to write out all tests explicitly for v1 and v2. 
I now used a mix of both; Separated tests where the body of the test diverges and parametrised tests where just the setup is different, as long as the setup doesn't get to messy.

@litestar-org/members?

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -5,22 +5,35 @@
...
 from litestar.types import Empty
 from litestar.typing import FieldDefinition
-from litestar.utils import is_class_and_subclass, is_pydantic_model_class, is_undefined_sentinel
+from litestar.utils import is_class_and_subclass, is_undefined_sentinel
 
 try:
-    import pydantic
+    import pydantic as _pydantic
+
+    PYDANTIC_VERSION = "1" if _pydantic.VERSION.startswith("1") else "2"
```
guacs, line 19: Is this being used anywhere?

litestar/contrib/pydantic/utils.py
```diff
@@ -0,0 +1,108 @@
...
+    from pydantic import v1 as pydantic_v1
+    import pydantic as pydantic_v2
+except ImportError:
+    import pydantic as pydantic_v1  # type: ignore[no-redef]
+
+    pydantic_v2 = Empty  # type: ignore[assignment]
+
+except ImportError:  # noqa: B025  # pyright: ignore
+    pydantic_v1 = Empty  # type: ignore[assignment]
+    pydantic_v2 = Empty  # type: ignore[assignment]
```
guacs, line 20: This should be handled in an inner try-except block right??

guacs (commented):
I haven't gone through all of it yet.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -259,6 +258,8 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
         Returns:
             A schema instance.
         """
+        from litestar.contrib.pydantic.utils import is_pydantic_constrained_field
```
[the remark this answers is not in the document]
provinzkraut: Unfortunately not, as this is needed to get around a circular dependency issue which can't be resolved otherwise without breaking something

litestar/contrib/pydantic/pydantic_dto_factory.py
```diff
@@ -16,21 +17,36 @@
...
+    import pydantic as pydantic_v1  # type: ignore[no-redef]
+
+    pydantic_v2 = Empty  # type: ignore[assignment]
+    from pydantic import ValidationError as ValidationErrorV1  # type: ignore[assignment]
+
+    ValidationErrorV2 = ValidationErrorV1  # type: ignore[assignment, misc]
+
+
+except ImportError as e:  # noqa: B025  # pyright: ignore
     raise MissingDependencyException("pydantic") from e
```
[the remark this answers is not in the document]
provinzkraut: No. The logic is:

1. Try to import `pydantic.v1`. If this succeeds, we have pydantic version 2 installed and can import both Pydantic 1 and 2. If this fails
2. Try to import just `pydantic`, if this succeeds, it means we have Pydantic 1 installed and will set `pydantic_v2 = Empty`. If this fails
3. We don't have Pydantic installed and want to raise a `MissingDependencyException`

litestar/contrib/pydantic/pydantic_init_plugin.py
```diff
@@ -5,39 +5,48 @@
...
-from litestar.exceptions import MissingDependencyException
 from litestar.plugins import InitPluginProtocol
-from litestar.utils import is_class_and_subclass, is_pydantic_model_class
+from litestar.utils import is_class_and_subclass
+
+try:
+    import pydantic as pydantic_v2
+    from pydantic import v1 as pydantic_v1
+except ImportError:
+    import pydantic as pydantic_v1  # type: ignore[no-redef]
```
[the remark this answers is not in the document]
provinzkraut: Probably, since we're doing this elsewhere as well

litestar/contrib/pydantic/pydantic_schema_plugin.py
provinzkraut: It's not, no. I'll remove it

litestar/contrib/pydantic/pydantic_dto_factory.py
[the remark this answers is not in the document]
guacs: Yes, but an import error during the handling of the first import error will not result in the `except ImportError as e` code being called.

```python
try:
    raise Exception()
except Exception:
    raise Exception()
except Exception:
    print("Hello, world") # never gets printed due to the exception raised in the first `except Exception` block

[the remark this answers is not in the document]
Comment (provinzkraut) [reply litestar/contrib/pydantic/pydantic_dto_factory.py]: I simplified the whole process; We now only check for the presence of Pydantic in the top-level `contrib.pydantic` module, since we re-export everything from there.

File: litestar/contrib/pydantic/utils.py [hunk 10]
```diff
@@ -0,0 +1,115 @@
...
+    from typing_extensions import TypeGuard
+
+PYDANTIC_UNDEFINED_SENTINELS = set()
+
+if pydantic_v1 is not Empty:  # type: ignore[comparison-overlap]
+    PYDANTIC_UNDEFINED_SENTINELS.add(pydantic_v1.fields.Undefined)  # pyright: ignore
+    if pydantic_v2 is not Empty:  # type: ignore[comparison-overlap]
+        PYDANTIC_UNDEFINED_SENTINELS.add(pydantic_v2.fields.PydanticUndefined)  # type: ignore[attr-defined]
+
+UNDEFINED_SENTINELS.update(PYDANTIC_UNDEFINED_SENTINELS)
```
provinzkraut, line 35: Not sure if this is a good idea actually. This only modifies that global when its imported, which is guaranteed to happen if you're using the Pydantic plugin, but it is a side effect.

On the other hand, I don't really like having the Pydantic stuff in the regular namespace, but maybe we should take care of that with a separate PR

litestar/contrib/pydantic/pydantic_init_plugin.py
```diff
@@ -55,22 +64,41 @@ def _dec_pydantic_uuid(
...
+        }
+    )
+
+
+def is_pydantic_v1_model_class(annotation: Any) -> TypeGuard[type[pydantic_v1.BaseModel]]:
+    return is_class_and_subclass(annotation, pydantic_v1.BaseModel)
+
+
+def is_pydantic_v2_model_class(annotation: Any) -> TypeGuard[type[pydantic_v1.BaseModel]]:
+    return is_class_and_subclass(annotation, pydantic_v2.BaseModel)
```
peterschutt, line 100:
```suggestion
def is_pydantic_v2_model_class(annotation: Any) -> TypeGuard[type[pydantic_v2.BaseModel]]:
    return is_class_and_subclass(annotation, pydantic_v2.BaseModel)
```

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -62,73 +66,122 @@
...
-            pydantic.NonPositiveFloat: Schema(type=OpenAPIType.NUMBER, maximum=0.0),
-            pydantic.PaymentCardNumber: Schema(type=OpenAPIType.STRING, min_length=12, max_length=19),
-            pydantic.PositiveFloat: Schema(type=OpenAPIType.NUMBER, exclusive_minimum=0.0),
-            pydantic.PositiveInt: Schema(type=OpenAPIType.INTEGER, exclusive_minimum=0),
         }
     )
 
-_supported_types = (pydantic.BaseModel, *list(PYDANTIC_TYPE_MAP.keys()))
+
+_supported_types = (pydantic_v1.BaseModel, *list(PYDANTIC_TYPE_MAP.keys()))
```
peterschutt, line 182:
```suggestion
_supported_types = (pydantic_v1.BaseModel, *(k for k in PYDANTIC_TYPE_MAP.keys()))
```
Pedantic, but no need to build the list. Could also be `*iter(PYDANTIC_TYPE_MAP.keys())`.

pyproject.toml
peterschutt: Funny story. I was removing `# pyright: ignore`'s in that module to see what pyright was complaining about as I was going through the review, and it would never report an issue. So I convinced myself that something was wrong with my pyright installation :D Perhaps we can remove all those ignore directives then?

tests/unit/test_contrib/test_pydantic/conftest.py
peterschutt: sounds reasonable to me :+1:

litestar/_openapi/schema_generation/examples.py
```diff
@@ -14,9 +14,9 @@
 from polyfactory.utils.predicates import is_union
 from typing_extensions import get_args
 
+from litestar.contrib.pydantic.utils import is_pydantic_model_instance
```
guacs, line 17: Would this cause the `MissingDependencyError` to be raised if pydantic is not installed?

litestar/_openapi/schema_generation/schema.py
guacs, line 261: Same doubt as above.

tests/unit/test_contrib/test_pydantic/test_beanie_integration.py
```diff
@@ -1,13 +1,15 @@
...
+if TYPE_CHECKING:
+    from pydantic import BaseModel
 
 from litestar.contrib.pydantic import PydanticDTO
 
 
-def test_generate_field_definitions_from_beanie_models() -> None:
-    class Category(pydantic.BaseModel):
+def test_generate_field_definitions_from_beanie_models(base_model: "Type[BaseModel]") -> None:
+    class Category(base_model):  # type: ignore[valid-type, misc]
```
peterschutt, line 12: What about ignoring these in pyproject for the `test_pydantic` module?

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -62,73 +66,122 @@
...
+    pydantic_v1.NonPositiveFloat: Schema(type=OpenAPIType.NUMBER, maximum=0.0),
+    pydantic_v1.PaymentCardNumber: Schema(type=OpenAPIType.STRING, min_length=12, max_length=19),
+    pydantic_v1.PositiveFloat: Schema(type=OpenAPIType.NUMBER, exclusive_minimum=0.0),
+    pydantic_v1.PositiveInt: Schema(type=OpenAPIType.INTEGER, exclusive_minimum=0),
 }
 
-if pydantic.VERSION.startswith("1"):  # pragma: no cover
-    # pydantic v1 values only - some are removed in v2, others are Annotated[] based and require a different
-    # logic
+if pydantic_v2 is not None:
```
guacs, line 127: Is this check correct? The previous check was for whether it's pydantic v1.

peterschutt (approved):
Huge effort! LGTM.

My only concern is we aren't actually testing anywhere with actual v1.10 installed and v2 not installed. What do you think about making another acceptance test that tests a light pydantic v1 only application alongside our minimal app test in the publish CI? Just to trap any major stuff like import errors before a release?

litestar/contrib/pydantic/utils.py
guacs: Yeah I agree. I think it would be really nice if both `PYDANTIC_UNDEFINED_SENTINELS` and `UNDEFINED_SENTINTELS` where instances of `frozenset`.

litestar/typing.py
```diff
@@ -225,10 +219,12 @@ def __hash__(self) -> int:
     def _extract_metadata(
         cls, annotation: Any, name: str | None, default: Any, metadata: tuple[Any, ...], extra: dict[str, Any] | None
     ) -> tuple[KwargDefinition | None, dict[str, Any]]:
+        from litestar.contrib.pydantic.utils import is_pydantic_constrained_field, is_pydantic_field_info
```
guacs, line 222: Same doubt as mentioned before. I think this would cause the `MissingDepdendencyError` to be raised in the case of `pydantic` not being installed.

tests/unit/test_contrib/test_pydantic/test_beanie_integration.py
```diff
@@ -1,13 +1,15 @@
...
+
+if TYPE_CHECKING:
+    from pydantic import BaseModel
 
 from litestar.contrib.pydantic import PydanticDTO
 
 
-def test_generate_field_definitions_from_beanie_models() -> None:
-    class Category(pydantic.BaseModel):
+def test_generate_field_definitions_from_beanie_models(base_model: "Type[BaseModel]") -> None:
```
guacs, line 11: nit: Maybe we could use `from __future__ import annotations` so that we can remove the stringified type annotation.

guacs (commented):
I think this looks good, but it's probably best if someone else took a look this as well just to make sure we don't miss anything.

litestar/contrib/pydantic/utils.py
provinzkraut: I'm thinking of leaving it like this for now, and then do a larger refactoring that removes the need for this alltogether. Wdyt?

litestar/_openapi/schema_generation/examples.py
provinzkraut: Good catch. Yeah, it would. Oh my, this whole stuff is a mess. I'll try to figure out a way that actually works correctly and is a bit nicer :)

litestar/contrib/pydantic/pydantic_schema_plugin.py
provinzkraut: Check is correct, the definitions are the ones for v2. The types below are false though :grimacing:

tests/unit/test_contrib/test_pydantic/test_beanie_integration.py
provinzkraut: If it was just the `valid-type`, but `misc` is really broad, so I'd rather not

litestar/_openapi/schema_generation/examples.py
provinzkraut: I fixed it. I think :grimacing:

tests/unit/test_contrib/test_pydantic/test_beanie_integration.py
provinzkraut: Generally we want to test these without I think

litestar/contrib/pydantic/utils.py
guacs: @provinzkraut, sorry I missed this. I think leaving it for a later refactoring is a good idea as well to keep this PR focused.

provinzkraut:
This PR will bring down coverage a bit, as a lot of previously uncovered code will now be included in the coverage report. This should be addressed, but I didn't want to blow up this PR even more.

PR #2492, opened 2023-10-22
fix: surface fields for single parameter 'query'

### LINKED ISSUE #2015: Enhancement: Support for Typing the `query` keyword (and other reserved kwargs) to generate OpenAPI Spec
### Summary

```json
  "paths": {
    "/": {
      "get": {
        "summary": "Foo",
        "operationId": "Foo",
        "responses": {
          "200": {
            "headers": {},
            "content": {
              "application/json": {
                "schema": {
                }
              }
            }
          }
        },
        "deprecated": false
      }
    }
  }
```

### Basic Example

```python
from typing import Annotated


class EvenOdd(BaseModel):


@get()
    return query


app = Litestar([foo])


with TestClient(app) as test:
    data = {'even': 0, 'odd': 1}

    data = {'even': 1, 'odd': 0}
```

### Drawbacks and Impact

### Unresolved questions

jw-can:
### Pull Request Checklist

### Description

- surface fields for single parameter 'query'

### Close Issue(s)

- Fixes #2015


jw-can:
There was a suggestion to raise an exception if there are other parameters than 'query' (which should be included within query). However, test cases (e.g. test_create_parameters in test_parameters.py) suggest a different behaviour. I therefore decided to only surface the fields if 'query' is the only parameter.

litestar/_openapi/parameters.py
```diff
@@ -201,6 +203,25 @@ def create_parameter_for_handler(
...
     layered_parameters = route_handler.resolve_layered_parameters()
 
+    # if there's a single parameter 'query', surface its fields
+    if "query" in handler_fields and len(handler_fields) == 1:
+        if pydantic.VERSION.startswith("1"):  # pragma: no cover
+            model_fields: dict[str, pydantic.fields.FieldInfo] = {k: model_field.field_info for k, model_field in handler_fields["query"].annotation.__fields__.items()}  # type: ignore
+        else:
+            model_fields = dict(handler_fields["query"].annotation.model_fields)
+        for k in model_fields:
+            t = model_fields[k]
```
JacobCoffee, line 213:
```suggestion
        for k, t in model_fields.items():
```
suggested sourcery edit

JacobCoffee: [`b22e57c` (#2532)](https://github.com/litestar-org/litestar/pull/2532/commits/b22e57cdd565702254702f5183a1837315b9ad4b)

cofin (approved):
LGTM

litestar/_openapi/parameters.py
```diff
@@ -206,6 +208,24 @@ def create_parameter_for_handler(
     dependency_providers = route_handler.resolve_dependencies()
     layered_parameters = route_handler.resolve_layered_parameters()
 
+    # if there's a single parameter 'query', surface its fields
+    if "query" in handler_fields and len(handler_fields) == 1:
```
cofin, line 212: I don't want to hold this one up since it's been open for a while, but is `query` the only parameter this applies to?  What about something like `state`?  Does that fall into this same category?

litestar/_openapi/parameters.py
```diff
@@ -2,6 +2,8 @@
 
 from typing import TYPE_CHECKING
 
+import pydantic
```
guacs, line 5: We can't import `pydantic` here since users might not be using it.

litestar/_openapi/parameters.py
```diff
@@ -206,6 +208,24 @@ def create_parameter_for_handler(
     dependency_providers = route_handler.resolve_dependencies()
     layered_parameters = route_handler.resolve_layered_parameters()
 
+    # if there's a single parameter 'query', surface its fields
+    if "query" in handler_fields and len(handler_fields) == 1:
+        if pydantic.VERSION.startswith("1"):  # pragma: no cover
+            model_fields: dict[str, pydantic.fields.FieldInfo] = {k: model_field.field_info for k, model_field in handler_fields["query"].annotation.__fields__.items()}  # type: ignore
+        else:
+            model_fields = dict(handler_fields["query"].annotation.model_fields)
```
guacs, line 216: It seems adding the `query` only works with `pydantic` models. I think this should be refactored to allow for passing in any of the models that are supported i.e. `msgpsec`, `pydantic`, `attrs`, `dataclasses`.

guacs (changes requested):

provinzkraut:
@jw-can any updates here?

JacobCoffee:
Closing as stale, feel free to reopen

Issue #2468, opened 2023-10-18
Docs: How to run the Todo app tutorial with sqlite?


orgarten:
### Summary

I am currently trying to get started with litestar with the [todo app with sqlalchemny tutorial](https://docs.litestar.dev/2/tutorials/sqlalchemy/0-introduction.html). 

Running the code provided by this page does not run currently because the table `todo_items` is missing in the database.

```
(sqlite3.OperationalError) no such table: todo_items [SQL: SELECT todo_items.title, todo_items.done FROM todo_items] (Background on this error at: https://sqlalche.me/e/20/e3q8)
```

This is triggered by `result = await session.execute(query)` in the function `get_todo_list`.

It does make sense to me, because no table is visibly created. I am not very familiar with SQLalchemy and wasn't able to figure out what to do. Not sure if this is a bug in the code or if it's just a missing hint in the docs that one has to create the table before.

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh Litestar dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2468">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2468/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2468/pledge.svg">
</picture>
</a>

peterschutt:
Thanks, a contribution to add this would be very welcome!

orgarten:
I will add this within the next couple of days. 

When I build the docs with `make docs-serve` I don't see different code for different python versions. Where is that configured?

orgarten:
Closed by merge of #2497

PR #2497, opened 2023-10-23
docs: Added database creation on startup to the todo app with SQLAlchemy tutorial

orgarten:
### Pull Request Checklist

### Description

- This PR adds an `on_startup` function to the tutorial `Improving the TODO app with SQLAlchemy`  to create database.
- This PR also adds apostrophes to for pip install for improved platform independence

### Close Issue(s)

- Addresses issue #2468.


docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py
```diff
@@ -64,6 +64,11 @@ async def get_todo_list(done: Optional[bool], session: AsyncSession) -> List[Tod
     return result.scalars().all()
 
 
+async def on_startup(app: Litestar) -> None:
+    async with app.state.engine.begin() as conn:
+        await conn.run_sync(Base.metadata.create_all)
```
peterschutt, line 69: This would be more succinct if we perform it inside the `db_connection` lifespan context:

```py
@asynccontextmanager
async def db_connection(app: Litestar) -> AsyncGenerator[None, None]:
    engine = getattr(app.state, "engine", None)
    if engine is None:
        engine = create_async_engine("sqlite+aiosqlite:///todo.sqlite", echo=True)
        app.state.engine = engine

    async with engine.begin() as conn:
        await conn.run_sync(Base.metadata.create_all)

    try:
        yield
    finally:
        await engine.dispose()
```

This way we don't need the extra `on_startup` method and it functionality-wise it should be to the same effect.

peterschutt (changes requested):
Thanks for this!

I think we might be able to make the example more succinct by including the schema creation inside the lifespan context manager.

LMK what you think.

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py
orgarten: The `asnyccontextcontext` is only there for the first part of the tutorial. For the 2 last parts we need to have the `on_startup` method anyway and I figured I just introduce it there to decrease the amount of changes along the way.

peterschutt: Good point.

Got me thinking we should add a `create_all` option to the configs, e.g., `db_config = SQLAlchemyAsyncConfig(connection_string="sqlite+aiosqlite:///todo.sqlite", create_all=metadata)`.

I think use the lifecycle hook in the examples where you can, b/c it is nicer, and switch to using the `on_shutdown` hook when you have to, with an explanation.

I'll add a feature request to `advanced-alchemy` to see if we can get the `create_all` config parameter across the line.

peterschutt: https://github.com/jolt-org/advanced-alchemy/issues/65

cofin:
@all-contributors, add @orgarten for docs

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py
peterschutt: Perfect, thanks!

orgarten: Just updated the first parts of the tutorial. 

@peterschutt and @cofin, is there any reason to limit `advanced-alchemy` to `<0.4.0` in `pyproject.toml`: `sqlalchemy = ["advanced-alchemy>=0.2.2,<0.4.0"]`?

The [aforementioned PR](https://github.com/jolt-org/advanced-alchemy/pull/70) pushed `advanced-alchemy` to v0.4.0. 

For this to work I would limit it to a version below v0.5.0 and then continue?

orgarten: I had to bump the version of `advanced-alchemy` in pdm lock as well to make the CI/CD run. 

LMK if you are happy with the current state.

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_init_plugin.py
```diff
@@ -74,7 +74,7 @@ async def update_item(item_title: str, data: TodoItem, transaction: AsyncSession
     return todo_item
 
 
-db_config = SQLAlchemyAsyncConfig(connection_string="sqlite+aiosqlite:///todo.sqlite")
+db_config = SQLAlchemyAsyncConfig(connection_string="sqlite+aiosqlite:///todo.sqlite", create_all=True)
```
peterschutt, line 77: If I understand the changes to advanced-alchemy correctly, we'd also need to add `metadata=Base.metadata` to the config.

docs/tutorials/sqlalchemy/3-init-plugin.rst
```diff
@@ -19,10 +19,11 @@ Here's the updated code:
 .. literalinclude:: /examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_init_plugin.py
     :language: python
     :linenos:
-    :emphasize-lines: 11,29,77,84
+    :emphasize-lines: 10-11,29,77,84
 
 The most notable difference is that we no longer need the ``db_connection()`` lifespan context manager - the plugin
-handles this for us.
+handles this for us. It also handles the creation of the tables in our database if we set ``create_all = True`` when
+creating a ``SQLAlchemyAsyncConfig`` instance.
```
peterschutt, line 26:
```suggestion
handles this for us. It also handles the creation of the tables in our database if we supply our metadata and set ``create_all=True`` when
creating a ``SQLAlchemyAsyncConfig`` instance.
```

peterschutt (changes requested):
We are getting close, thanks for your efforts:)

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_init_plugin.py
peterschutt: https://github.com/jolt-org/advanced-alchemy/blob/bc51751897ff6ca56e2a21ddc9897248cc6e62ef/advanced_alchemy/config/common.py#L120-L126

orgarten: It does run without specifying which metadata to use. However, one could add it specifically. I am not too familiar with the details of SQLAlchemy, so not 100% sure what's needed.

[Here](https://github.com/jolt-org/advanced-alchemy/blob/05ea56e1d7f985f09596fa8948f9171b2ad97f2d/advanced_alchemy/config/common.py#L211) the default metadata is specified as `orm_registry.metadata`. I am not sure what exactly that entails.

docs/tutorials/sqlalchemy/3-init-plugin.rst
orgarten: Same as above I think. Not needed, but possible. Maybe @cofin can add his thoughts?

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_init_plugin.py
peterschutt: It runs, but have you confirmed that the tables are being created? 

We aren't using the advanced-alchemy base classes in this example. We create our own `Base` by creating the `class Base(DeclarativeBase): ...` class which all of the models in the example inherit from.

That means that `advanced_alchemy.base.orm_registry.metadata` doesn't know anything about the tables represented by the models in the example. And this is why we need to tell the config object about _our_ metadata.

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py
cofin: @orgarten We definitely should bump up the advanced alchemy version now.  I'll make a PR to do this today.

peterschutt: I did this as part of a PR yesterday to update deps - just capped AA at <1.0

docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_init_plugin.py
peterschutt: Awsome, after this change we'll be right to merge! Thanks.

orgarten: Ok, added the code and committed your change in the wording, too. Had to fix a couple of formatting issues that didn't show up in my local pre-commit run. Should work now.

peterschutt (approved):

peterschutt:
Thank you @orgarten!

orgarten:
> Thank you @orgarten!

My pleasure!

PR #2498, opened 2023-10-23
refactor: my last pass

peterschutt:
### Pull Request Checklist

### Description

- makes `_get_type_schema_name()` receive the field definition. This removes the need to call `get_args()` on the annotation as we've already done it.
- removed a use of `get_origin_or_inner_type()`
- fixed type error

### Close Issue(s)

-

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -142,22 +142,22 @@ TYPE_MAP: dict[type[Any] | None | Any, Schema] = {
 }
 
 
-def _get_type_schema_name(value: Any) -> str:
+def _get_type_schema_name(field_definition: FieldDefinition) -> str:
     """Extract the schema name from a data container.
 
     Args:
-        value: A data container
+        field_definition: A field definition instance.
 
     Returns:
         A string
     """
 
-    if name := getattr(value, "__schema_name__", None):
+    if name := getattr(field_definition.annotation, "__schema_name__", None):
         return cast("str", name)
 
-    name = get_name(value)
-    if args := get_args(value):
-        inner_parts = ", ".join(_get_type_schema_name(a) for a in args)
+    name = get_name(field_definition.annotation)
+    if field_definition.inner_types:
+        inner_parts = ", ".join(_get_type_schema_name(t) for t in field_definition.inner_types)
         return f"{name}[{inner_parts}]"
 
     return name
@@ -274,7 +274,7 @@ class SchemaCreator:
         result: Schema | Reference
 
         # NOTE: The check for whether the field_definition.annotation is a Pagination type
-        # has to come before the `is_dataclass_check` since the the Pagination classes are dataclasses
+        # has to come before the `is_dataclass_check` since the Pagination classes are dataclasses,
         # but we want to handle them differently from how dataclasses are normally handled.
         if field_definition.is_optional:
             result = self.for_optional_field(field_definition)
@@ -495,7 +495,7 @@ class SchemaCreator:
                 for field in fields
             },
             type=OpenAPIType.OBJECT,
-            title=_get_type_schema_name(field_definition.annotation),
+            title=_get_type_schema_name(field_definition),
         )
 
     # noinspection PyDataclass
@@ -525,7 +525,7 @@ class SchemaCreator:
             ),
             properties={k: self.for_field_definition(FieldDefinition.from_kwarg(v, k)) for k, v in type_hints.items()},
             type=OpenAPIType.OBJECT,
-            title=_get_type_schema_name(field_definition.annotation),
+            title=_get_type_schema_name(field_definition),
         )
 
     # noinspection PyTypedDict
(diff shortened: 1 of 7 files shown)


litestar/_openapi/schema_generation/schema.py
```diff
@@ -142,22 +142,22 @@
...
-    if name := getattr(value, "__schema_name__", None):
+    if name := getattr(field_definition.annotation, "__schema_name__", None):
         return cast("str", name)
 
-    name = get_name(value)
-    if args := get_args(value):
-        inner_parts = ", ".join(_get_type_schema_name(a) for a in args)
+    name = get_name(field_definition.annotation)
+    if field_definition.inner_types:
+        inner_parts = ", ".join(_get_type_schema_name(t) for t in field_definition.inner_types)
```
peterschutt, line 160: This means we don't need to call `get_args()` again for the type and each of its inner types.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -274,7 +274,7 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
         result: Schema | Reference
 
         # NOTE: The check for whether the field_definition.annotation is a Pagination type
-        # has to come before the `is_dataclass_check` since the the Pagination classes are dataclasses
+        # has to come before the `is_dataclass_check` since the Pagination classes are dataclasses,
```
peterschutt, line 277: Sorry, just nit, but IDE was highlighting some grammar stuff.

litestar/typing.py
```diff
@@ -397,7 +397,7 @@ def generic_types(self) -> tuple[FieldDefinition, ...] | None:
     def is_dataclass_type(self) -> bool:
         """Whether the annotation is a dataclass type or not."""
 
-        return is_dataclass(self.origin or self.annotation)
+        return is_dataclass(cast("type", self.origin or self.annotation))
```
peterschutt, line 400: The cast is necessary b/c `self.origin or self.annotation` is typed `Any` b/c `self.origin` and `self.annotation` are both typed `Any`.

I think we should be able to change the type of `origin` to `type[Any] | None` because where `annotation` is one of the type alias things, it will always return a class.

Then, I'm thinking I will add a method `FieldDefinition.origin_or_annotation()` which will be typed to return `type[Any]`. We use the `field_def.origin or field_def.annotation` pattern quite a bit, and whenever `annotation` is not a type (ie. special form thing, or whatever) then origin will be non-none, otherwise, `annotation` will always be a type.

I'll do that separate though.

peterschutt (commented):
These are just some final things I noticed sweeping over the code once more.

Take, or leave, but I'm happy to go when you are with the rest.

Great stuff!

litestar/typing.py
guacs: Oh so this is why mypy was complaining.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -142,22 +142,22 @@
...
     Args:
-        value: A data container
+        field_definition: A field definition instance.
 
     Returns:
         A string
     """
 
-    if name := getattr(value, "__schema_name__", None):
+    if name := getattr(field_definition.annotation, "__schema_name__", None):
```
guacs, line 155: Would this have to be `field_definition.origin or field_definition.annotation` since `Foo[T]` might not have `__schema_name__`?

litestar/_openapi/schema_generation/schema.py
```diff
@@ -142,22 +142,22 @@
...
         return cast("str", name)
 
-    name = get_name(value)
-    if args := get_args(value):
-        inner_parts = ", ".join(_get_type_schema_name(a) for a in args)
+    name = get_name(field_definition.annotation)
```
guacs, line 158: Same doubt as above.

guacs (approved):
Passing in `FieldDefinition` to `_get_type_schema_name` is a fantastic idea :)

litestar/typing.py
peterschutt: Yeah - the overloads for `is_dataclass` don't have a variant for `Any` arg: https://github.com/python/typeshed/blob/ba3f23b2040927cb278dcb4d5ea5f8c43eb5c8b3/stdlib/dataclasses.pyi#L226-L230

litestar/_openapi/schema_generation/schema.py
peterschutt: Well I don't think this is any more broken than it was, is it? B/c in every case that we called the function before this change we were passing `field_definition.annotation` through.

However, if you mean this:

```python-console
>>> from typing import Generic
>>> from typing import TypeVar
>>> T = TypeVar("T")
>>> class GenType(Generic[T]):
...     __schema_name__ = "abc"
... 
>>> GenType[int].__schema_name__
  File "/home/peter/.pyenv/versions/3.8.17/lib/python3.8/typing.py", line 760, in __getattr__
    raise AttributeError(attr)
AttributeError: __schema_name__
```
Then, yeah - I think you are right.

PR #2502, opened 2023-10-23
docs: Add documentation for creating a standard library logging logger

ddxv:
Added an example for how to use the standard library logger as opposed to LoggingConfig. This works in Python3.12 as well. The provided example is mostly taken from https://docs.python.org/3/howto/logging.html

I think the main point here, is that to use the standard logger, you need to not use LoggingConfig. So if you'd like the code example to be reduced a bit and a sentence to reflect that please let me know.

### Pull Request Checklist

### Description

-

### Close Issue(s)

-


docs/usage/logging.rst
```diff
@@ -32,6 +32,55 @@ Application and request level loggers can be configured using the :class:`~lites
...
+        # add ch to logger
+        logger.addHandler(ch)
+        return logger
+
+
+    logger = get_logger(__name__)
+
+
+    @get("/")
+    def my_router_handler(self, request: Request) -> None:
```
cofin, line 75:
```suggestion
    def my_router_handler(request: Request) -> None:
```

docs/usage/logging.rst
```diff
@@ -32,6 +32,55 @@ Application and request level loggers can be configured using the :class:`~lites
     is keyed as ``queue_listener`` in the logging configuration. The above example is using this handler,
     which is optimal for async applications. Make sure to use it in your own loggers as in the above example.
 
+
+
+Standard Library Logging
```
cofin, line 37:
```suggestion
Standard Library Logging (Manual Configuration)
```

docs/usage/logging.rst
```diff
@@ -32,6 +32,55 @@ Application and request level loggers can be configured using the :class:`~lites
...
+    @get("/")
+    def my_router_handler(self, request: Request) -> None:
+        logger.info("inside a request")
+        return None
+
+
+    app = Litestar(route_handlers=[my_router_handler])
+
+
+
```
cofin, line 83: I think we should also include an example of how to use the `LoggingConfig`.  This would be the preferred method for using the logger.

```python
import logging

from litestar.logging.config import LoggingConfig

log_config = LoggingConfig(
    root={"level": logging.getLevelName(logging.INFO), "handlers": ["console"]},
    formatters={"standard": {"format": "%(asctime)s - %(name)s - %(levelname)s - %(message)s"}},
)

logger = log_config.configure()()


@get("/")
def my_router_handler(request: Request) -> None:
    request.logger.info("inside a request")
    logger.info("here too")


app = Litestar(
    route_handlers=[my_router_handler],
    logging_config=log_config,
)
```

cofin:
Thanks for the additions!  We've had many questions on logging configuration, so this is much appreciated.

cofin:
@all-contributors add @ddxv for docs

docs/usage/logging.rst
ddxv: Ah nice, the above works with Python3.12 and uvicorn. I'm happy to just replace my example with this one actually since it uses the litestar LoggingConfig and works as expected. 

What do you think? The other idea would be to take elements of your example and incorporate them into the very first (not yet changed in this PR) logging example so that the first example users see is one that works on 3.12 and uvicorn?

ddxv (commented):
Made changes based on  @cofin comments

cofin (approved):
LGTM

PR #2506, opened 2023-10-23
docs: correct `blog_post_repo` dependency and replaced deprecated  calls

pbaletkeman:
Connected blog_post_repo dependency and replaced deprecated data.dict to data.model_dump

### Pull Request Checklist

### Description

-

### Close Issue(s)

-


pbaletkeman:
I am unsure how to solve this failing test https://github.com/litestar-org/litestar/actions/runs/6620348740/job/17982583981?pr=2506.
To me, it seems you need to import Provide to allow for the dependency to be injected.

Please let me know how I can fix this.

peterschutt:
> I am unsure how to solve this failing test https://github.com/litestar-org/litestar/actions/runs/6620348740/job/17982583981?pr=2506. To me, it seems you need to import Provide to allow for the dependency to be injected.
> 
> Please let me know how I can fix this.

Just an issue with the ordering of your imports - imports must be defined in alphabetical order, so the linter wants you to move the `litestar.di` import after the `litestar.contrib` imports.

docs/examples/contrib/sqlalchemy/sqlalchemy_repository_extension.py
```diff
@@ -180,6 +181,7 @@ async def create_blog(
 
 app = Litestar(
     route_handlers=[create_blog, get_blogs, get_blog_details],
+    dependencies={"blog_post_repo": Provide(provide_blog_post_repo, sync_to_thread=True)},
```
peterschutt, line 184:
```suggestion
    dependencies={"blog_post_repo": Provide(provide_blog_post_repo, sync_to_thread=False)},
```
I don't think we have any need to call this dependency in a separate thread.

The only time this would be necessary is if there was some synchronous I/O occurring in the `BlogPostRepository.__init__()` method. I don't think there is, so this will be much more efficient to be called as a normal sync function.

@cofin can you confirm if I'm missing something here? I also notice we do this in the `sqlalchemy_sync_repository.py` tutorial example.

peterschutt (changes requested):

pbaletkeman:
> > I am unsure how to solve this failing test https://github.com/litestar-org/litestar/actions/runs/6620348740/job/17982583981?pr=2506. To me, it seems you need to import Provide to allow for the dependency to be injected.
> > Please let me know how I can fix this.
> 
> Just an issue with the ordering of your imports - imports must be defined in alphabetical order, so the linter wants you to move the `litestar.di` import after the `litestar.contrib` imports.

Problem solved.
Thank you.

docs/examples/contrib/sqlalchemy/sqlalchemy_repository_extension.py
cofin: I agree that this should be updated.  We should not be doing IO in the `__init__` method now (or in the future).

cofin:
@all-contributors add @pbaletkeman for docs

docs/examples/contrib/sqlalchemy/sqlalchemy_repository_extension.py
peterschutt: thanks for confirming - I'll open an issue for the other module :+1:

cofin (approved):
LGTM

PR #2515, opened 2023-10-24
test: query param consistency with Annotated

### LINKED ISSUE #2514: Bug: Annotated optional query parameter inconsistency
### Description

Annotated optional query parameters are inconsistent compared to non annotated and required query parameters.

_No response_

### MCVE

```python
from typing import Annotated

@get("/optional", sync_to_thread=False)
def optional(key: str | None = None) -> dict[str, str | None]:
    return {"key": key}


@get("/annotated_required", sync_to_thread=False)
def annotated_required(param: Annotated[str, Parameter(query="key")]) -> dict[str, str]:
    return {"key": param}


@get("/annotated_optional", sync_to_thread=False)
def annotated_optional(
    param: Annotated[str | None, Parameter(query="key")] = None
) -> dict[str, str | None]:
    return {"key": param}


app = Litestar([optional, annotated_required, annotated_optional])
```

### Steps to reproduce

```bash
# optional - non annotated
$ curl -w '\n' -s 'localhost:8000/optional?param=a'
{"key":null}
$ curl -w '\n' -s 'localhost:8000/optional?key=a'
{"key":"a"}

# annotated - non optional
$ curl -w '\n' -s 'localhost:8000/annotated_required?param=a'
{"status_code":400,"detail":"Missing required query parameter 'key' for url http://localhost:8000/annotated_required?param=a"}
$ curl -w '\n' -s 'localhost:8000/annotated_required?key=a'
{"key":"a"}

# annotated - optional
$ curl -w '\n' -s 'localhost:8000/annotated_optional?param=a'
{"key":"a"}
$ curl -w '\n' -s 'localhost:8000/annotated_optional?key=a'
{"key":null}
```
```


### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.2.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2514">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2514/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
### Pull Request Checklist

### Description

This PR adds tests for consistency of behavior of query params between kwargs declared with and without `typing.Annotated`.

These tests were created when debugging #2514, however that issue was solved by PR #2516. I figure we may as well still add the tests.

### DIFF
--- a/tests/unit/test_params.py
+++ b/tests/unit/test_params.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Generator, List, Optional
 
 import pytest
 from typing_extensions import Annotated
@@ -8,7 +8,7 @@ from litestar.di import Provide
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.params import Body, Dependency, Parameter
 from litestar.status_codes import HTTP_200_OK, HTTP_201_CREATED, HTTP_400_BAD_REQUEST, HTTP_500_INTERNAL_SERVER_ERROR
-from litestar.testing import create_test_client
+from litestar.testing import TestClient, create_test_client
 
 
 def test_parsing_of_parameter_as_annotated() -> None:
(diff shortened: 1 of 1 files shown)


tests/unit/test_params.py
```diff
@@ -219,3 +219,78 @@ async def regex_val(text: Annotated[str, Parameter(title="a or b", pattern="[a|b
...
+        return {"key": param}
+
+    with create_test_client(route_handlers=[handle_optional, handle_optional_annotated]) as client:
+        yield client
+
+
+def test_optional_query_parameter_consistency_no_default_queried_without_param(
+    optional_no_default_client: TestClient,
+) -> None:
+    assert optional_no_default_client.get("/optional-no-default", params={}).json() == {"key": None}
```
peterschutt, line 241: This behavior is a bit undesirable IMO @litestar-org/maintainers - if a query param is `<smth> | None`, and it automatically defaults to `None` if not provided, that could cause problems. E.g.,:

```py
@get("/top-secret-government-cloud-using-litestar-route")
def armageddon_dead_man_switch(do_not_launch_nukes: Literal["thanks"] | None) -> None:
    if not do_not_launch_nukes:
        big_red_button.press()

# intern forgets to add query parameter to the query.
```
Stupid example, but YKWIM..

tests/unit/test_params.py
```diff
@@ -219,3 +219,78 @@ async def regex_val(text: Annotated[str, Parameter(title="a or b", pattern="[a|b
...
+    assert optional_no_default_client.get("/optional-annotated-no-default", params={"key": "a"}).json() == {"key": "a"}
+
+
+def test_optional_query_parameter_consistency_no_default_queried_with_other_param(
+    optional_no_default_client: TestClient,
+) -> None:
+    assert optional_no_default_client.get("/optional-no-default", params={"param": "a"}).json() == {"key": None}
+    assert optional_no_default_client.get("/optional-annotated-no-default", params={"param": "a"}).json() == {
+        "key": None
+    }
```
peterschutt, line 258: Same, these should really be 400s IMO

tests/unit/test_params.py
JacobCoffee: What would the alternative be to default to? the first `smth` value (`True`, `Literal#1`, etc)?

cofin: So what's the desirable outcome here?  

Should `do_not_launch_nukes` default to `thanks` instead?

or this?

```python
assert optional_no_default_client.get("/optional-annotated-no-default", params={}).json() == {}
```

peterschutt: No default at all IMO, it should be a client error.

peterschutt: Explicit better than implicit.

If the developer wants default `None`, they can declare it `key: Optional[str] = None`.

Where the type of the query parameter is `non-optional` and without default, e.g.,:

```py
    @get("/not-optional-no-default")
    def handle_not_optional(key: str) -> Dict[str, str]:
        return {"key": key}
```
And the query omits the parameter, that is a client error already. But for some reason we apply this implicit default when the parameter is optional (which I suspect is a hangover from pydantic signature model behavior).

JacobCoffee (approved):

PR #2521, opened 2023-10-26
build: run code coverage on min and max py vers

peterschutt:
We do a lot of `pragma: no cover` due to coverage only being reported on the highest version of python.

This PR modifies coverage reporting so that we run coverage tests on the lowest and highest supported versions, and combines the coverage reports.

### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -191,7 +191,7 @@ exclude_lines = [
   'except ImportError as e:',
   'except ImportError:',
   '\.\.\.',
-  'raise NotImplementedError',
+  'raise NotImplementedError.*',
   'if VERSION.startswith("1"):',
   'if pydantic.VERSION.startswith("1"):',
 ]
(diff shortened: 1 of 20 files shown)


pyproject.toml
```diff
@@ -191,7 +191,7 @@ exclude_lines = [
   'except ImportError as e:',
   'except ImportError:',
```
sobolevn, line 192:
```suggestion
  'except ImportError.*',
```

sobolevn: Also, consider using https://github.com/asottile/covdefaults
It also has conditional coverage.

provinzkraut (approved):
Great stuff!

pyproject.toml
peterschutt: > Also, consider using https://github.com/asottile/covdefaults It also has conditional coverage.

Thanks for the suggestion. I'll experiment with adopting it in a follow up.

PR #2522, opened 2023-10-26
feat: add rapidoc for openapi schema visualisation

betaboon:
### Pull Request Checklist

### Description
Add support for using [RapiDoc](https://github.com/rapi-doc/RapiDoc) for openapi schema visualisation.

### Close Issue(s)

-


peterschutt (approved):
This looks good to me, thanks very much!

peterschutt:
@all-contributors add @betaboon for code

provinzkraut:
Looks good in general but maybe we'd want to add this as a standalone plugin/package?

We have plenty of different UIs already available for this and I feel like we don't necessarily need to support *all* of them within the core framework.

@litestar-org/maintainers?

peterschutt:
> Looks good in general but maybe we'd want to add this as a standalone plugin/package?
> 
> We have plenty of different UIs already available for this and I feel like we don't necessarily need to support _all_ of them within the core framework.
> 
> @litestar-org/maintainers?

I'm also happy to go this route.

I actually mentioned in discord chat earlier that for 3.0, we should have all of the different UIs implemented by plugin. 

Thinking about that more, there's probably no reason why we couldn't do that for 2.0, and if a plugin is registered for OpenAPI UI then not register the OpenAPI controller, and if no plugin registered, then register the current controller as defined. We'd need to be able to detect that an OpenAPI UI plugin was added for this to work though..

betaboon:
i also mentioned on discord that i think this class could really use a refactor.
i guess users rarely would even want to expose more than one frontend at a time.

The original reason for wanting to add this is actually the non-functional state of swagger-ui for file-uploads.
i discussed with @cofin about the neccesary schema-changes in order to make the file-upload openapi-schema correct, which would then lead to it working in rapidoc.

see [this discord thread](https://discord.com/channels/919193495116337154/1165355806401757184) for further information.

cofin:
> i also mentioned on discord that i think this class could really use a refactor. i guess users rarely would even want to expose more than one frontend at a time.
> 
> The original reason for wanting to add this is actually the non-functional state of swagger-ui for file-uploads. i discussed with @cofin about the neccesary schema-changes in order to make the file-upload openapi-schema correct, which would then lead to it working in rapidoc.
> 
> see [this discord thread](https://discord.com/channels/919193495116337154/1165355806401757184) for further information.

I think we should move these to plugins.  But, i also don't want to hold up this fix that addresses a long standing file upload rendering issue.

My $0.02 is that we do this in two steps.  Merge this and then submit a refactored PR.

peterschutt:
> My $0.02 is that we do this in two steps. Merge this and then submit a refactored PR.

I too am fine with merging this as is. The implementation conforms with existing patterns, and is fairly small.

I'll open an issue to track the refactor.

JacobCoffee:
This is exciting :)

peterschutt:
> > My $0.02 is that we do this in two steps. Merge this and then submit a refactored PR.
> 
> I too am fine with merging this as is. The implementation conforms with existing patterns, and is fairly small.
> 
> I'll open an issue to track the refactor.

https://github.com/litestar-org/litestar/issues/2541

PR #2528, opened 2023-10-26
build: implement covdefaults coverage plugin.

peterschutt:
### Pull Request Checklist

### Description

This PR adds the [covdefaults](https://github.com/asottile/covdefaults) plugin to our coverage config.

The plugin offers default configs for `[tool.coverage.run]` and `[tool.coverage.report]`.

For "run", the main difference is that it activates branch coverage. This is something I always run on other projects and I think would be a nice addition to our config. Our overall coverage score takes a hit, but it's not as bad as I thought it might be.

For reporting, it adds a `fail_under` value - which we don't currently have as we let sonar handle that. We can override their default of 100, so I set it to 96 which is our current baseline after branch coverage activated.

### Close Issue(s)

-

### DIFF
--- a/litestar/__main__.py
+++ b/litestar/__main__.py
@@ -1,10 +1,10 @@
-from litestar.cli.main import litestar_group  # pragma: no cover
+from litestar.cli.main import litestar_group
 
 
-def run_cli() -> None:  # pragma: no cover
+def run_cli() -> None:
     """Application Entrypoint."""
     litestar_group()
 
 
-if __name__ == "__main__":  # pragma: no cover
+if __name__ == "__main__":
     run_cli()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,6 +135,7 @@ linting = [
   "types-redis",
 ]
 test = [
+  "covdefaults",
   "pytest",
   "pytest-asyncio",
   "pytest-cov",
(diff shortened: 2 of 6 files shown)


peterschutt:
RFC @litestar-org/maintainers - WDYT about adding this plugin?

pyproject.toml
```diff
@@ -183,17 +184,16 @@ skip = 'pdm.lock,docs/examples/contrib/sqlalchemy/us_state_lookup.json'
...
-  'pragma: no cover',
-  'if TYPE_CHECKING:',
-  'except ImportError.*',
-  '\.\.\.',
-  'raise NotImplementedError.*',
+  'except ImportError\b',
   'if VERSION.startswith("1"):',
   'if pydantic.VERSION.startswith("1"):',
 ]
+fail_under = 96
```
sobolevn, line 196: `pytest-cov` also can do this, even if this PR won't be merged: `--cov-fail-under=96` in `addopts`

sobolevn (commented):
I think that this plugin is very nice, removes a lot of boilerplate.
There's also https://github.com/wemake-services/coverage-conditional-plugin for cases where you need to have conditional coverage based on your own rules. For example: with `msgspec` and with `pydantic2`

In other words: lines and scopes can be conditionally ignored.

pyproject.toml
peterschutt: So far, we've been relying on Sonar to manage this metric for us - can't hurt to fail earlier though if it dips below. Also, sonar reports the coverage only on new contribution code for a PR, so measuring overall at this point can be useful too.

litestar/__main__.py
```diff
@@ -1,10 +1,10 @@
...
 
 
-def run_cli() -> None:  # pragma: no cover
+def run_cli() -> None:
     """Application Entrypoint."""
     litestar_group()
 
 
-if __name__ == "__main__":  # pragma: no cover
+if __name__ == "__main__":
```
peterschutt, line 9: I thought these should not be required given the plugin adds this to "omit" ([ref](https://github.com/asottile/covdefaults/blob/6e03d07652445d94683f97d14953c81559bd0ba3/covdefaults.py#L84)):

```py
    ('run:omit', ['*/__main__.py', '*/setup.py']),
```

JacobCoffee (approved):

JacobCoffee:
sgtm

PR #2537, opened 2023-10-27
feat: ResponseCacheConfig.cache_response_filter

### LINKED ISSUE #2501: Enhancement: Ignore cache when the cached response is an error
### Summary

Hello,

Following a thread on discord and the suggestion from @Alc-Alc 

When cache is enabled on the controller, it caches errors. So if we fix the issue and start again the app, the response sent by Litestar is still the error and not the correct answer.

```python
@get(
        operation_id="ListCustomer",
        name="customers:list",
        summary="List Customers",
        description="Retrieve the customers.",
        path=urls.CUSTOMER_LIST,
        cache=60,
    )
```

@Alc-Alc suggests the following:  adding `no_cache_on_exception` on `ResponseCacheConfig` (or any relevant place)

Thank you

### Basic Example

1. Create a route with cache=60
2. Create an error that route
3. Get the route through the appropriate method (GET)
4. Fix the issue and query the route again, the error is returned
5. (optional) if we set the cache to 0, the correct response is returned

### Drawbacks and Impact

Adding a new property/config
Documentation needed

### Unresolved questions

_No response_

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2501">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2501/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### LINKED ISSUE #2508: Documentation: Unnecessary use of sync_to_thread=True
The following are calling object constructors with no I/O, so should be right to run sync.

https://github.com/litestar-org/litestar/blob/0d4fe01a9d361e09748f71aa7adf70672b97c822/docs/examples/contrib/sqlalchemy/sqlalchemy_sync_repository.py#L120

https://github.com/litestar-org/litestar/blob/0d4fe01a9d361e09748f71aa7adf70672b97c822/docs/examples/contrib/sqlalchemy/sqlalchemy_sync_repository.py#L154

https://github.com/litestar-org/litestar/blob/0d4fe01a9d361e09748f71aa7adf70672b97c822/docs/examples/contrib/sqlalchemy/sqlalchemy_sync_repository.py#L170

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2508">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2508/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
### Pull Request Checklist

### Description

A predicate called by the response cache middleware that discriminates whether a response should be cached, or not.

PR includes a default implementation of the predicate.

### Close Issue(s)

Closes #2501


litestar/config/response_cache.py
```diff
@@ -35,6 +41,22 @@ def default_cache_key_builder(request: Request[Any, Any, Any]) -> str:
...
+        _: ASGI scope.
+        status_code: status code of the response.
+
+    Returns:
+        A boolean indicating whether the response should be cached.
+    """
+    return HTTP_200_OK <= status_code < HTTP_300_MULTIPLE_CHOICES or status_code in (
+        HTTP_301_MOVED_PERMANENTLY,
+        HTTP_308_PERMANENT_REDIRECT,
+    )
```
peterschutt, line 57: Any thoughts on what a default for this should be?

litestar/config/response_cache.py
```diff
@@ -49,6 +71,9 @@ class ResponseCacheConfig:
     """:class:`CacheKeyBuilder <.types.CacheKeyBuilder>`. Defaults to :func:`default_cache_key_builder`."""
     store: str = "response_cache"
     """Name of the :class:`Store <.stores.base.Store>` to use."""
+    do_cache_predicate: Callable[[HTTPScope, int], bool] = field(default=default_do_cache_predicate)
```
provinzkraut, line 74: Maybe `should_cache_response` would be a better name for this? I know technically the `predicate` is more precise, but this feels more intuitive to me

peterschutt: Yeh, I hate the name `do_cache_predicate` too.. still used it though :smile:

peterschutt: WDYT about `cache_filter`?

peterschutt: Well, I went with `cache_filter` for now, I think it is succinct and fairly obvious what it is.. happy to change again if you don't like it.

provinzkraut: Hmm. That kinda sounds like it would filter in the other direction, i.e. checking if an already cached response should be used

peterschutt: `cache_response_filter`?

litestar/config/response_cache.py
cofin: I'm not sure what you mean by default, but these specific status codes conform to the the [PRG pattern](https://en.wikipedia.org/wiki/Post/Redirect/Get)

cofin (approved):
LGTM

litestar/config/response_cache.py
peterschutt: I suppose I just mean, what responses should we not cache by default. E.g., this will cache a 201 CREATED which I imagine wouldn't be common.

But then again, a user would have to set `cache=True` on a POST route for this to occur, and at that point, they are just getting what they ask for I guess.

JacobCoffee: `201`, `202`, `204`, `206`, `304`? Not really sure the right choice here for defaults though..

JacobCoffee (approved):

litestar/config/response_cache.py
peterschutt: One thing to consider is that where caching is enabled, we've been caching _everything_. While we obviously don't want to cache error responses, the success responses like 201 - it is not so clear to me that we can or should change that behavior. I just know I can be a bit conservative on this type of stuff, so IDK..

guacs: The caching of everything has been a source of bugs though. I think the current default you have is fine, but we should also exclude caching of responses to "OPTIONS" requests. Then maybe in v3 we could just go ahead and only cache responses to "GET" requests by default?

guacs (approved):

litestar/config/response_cache.py
provinzkraut: > Then maybe in v3 we could just go ahead and only cache responses to "GET" requests by default?

I don't think that's needed, as you can already selectively enable/disable the caching per route handler, which also means per method. This is only an issue for `OPTIONS` because it doesn't necessarily have an explicit route handler.

However, there is another bug at play here which is causing issues; We don't take the request method into account when calculating the cache key. This means that, not just for `OPTIONS` requests, requests that go to the same route will overwrite each other's caches.

guacs: > you can already selectively enable/disable the caching per route handler, which also means per method

Makes sense to me.

> We don't take the request method into account when calculating the cache key

Aah you're right.

peterschutt: > We don't take the request method into account when calculating the cache key.

Yeh this is prob the best solution to #2573

peterschutt: I'll leave this as is for now then.

PR #2544, opened 2023-10-27
fix: FieldDefinition.from_kwarg() with None default

peterschutt:
### Pull Request Checklist

### Description

This PR fixes an issue where parameters with `None` defaults would have `FieldDefinition.default == Empty`.

Fixing this only broke 2 tests, which were both changed to test that the field def. has an empty default when they clearly shouldn't (i.e., default defined in the test func signature for the param) :shrug:

This PR changes the logic in `FieldDefinition.from_kwarg()` to not filter out the default value from params it passes to `FieldDefinition.from_annotation` when its value is `None`. It also reverts the behavior of the tests to test that the `FieldDefinition.default` value is `None` when the default is declared in the function signature.

### Close Issue(s)

-


cofin (approved):
LGTM

PR #2547, opened 2023-10-27
fix(channels): Performance issue with polling without subscriptions

### LINKED ISSUE #2503: Litestar consumes high amount of CPU
### Description

Running litestar `litestar app run` with uvicorn consumes in an idling state on my machines 40-45% CPU.

Is this normal or is there a way to reduce the idling CPU consumption?

Attached a picture of cProfile. `python -m cProfile -o litestar.pstats ./.venv/bin/litestar run`

_No response_

### MCVE

```python
UVICORN_LOOP=uvloop litestar run                            
Using Litestar app from env: 'app.asgi:create_app'
Starting server process ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
┌──────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Litestar version             │ 2.2.1                                                                                                        │
│ Debug mode                   │ Disabled                                                                                                     │
│ Python Debugger on exception │ Disabled                                                                                                     │
│ CORS                         │ Enabled                                                                                                      │
│ CSRF                         │ Disabled                                                                                                     │
│ OpenAPI                      │ Enabled path=/schema                                                                                         │
│ Compression                  │ Disabled                                                                                                     │
│ Template engine              │ ViteTemplateEngine                                                                                           │
│ Static files                 │ path=/static dirs=.../src/app/domain/web/public html_mode=Enabled │
│ Middlewares                  │ JWTCookieAuthenticationMiddleware, SentryLitestarASGIMiddleware, middleware_factory                          │
└──────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
2023-10-23T19:12:38.780745Z [info     ] Started server process [4257]  [uvicorn.error] 
2023-10-23T19:12:38.781923Z [info     ] Waiting for application startup. [uvicorn.error] 
2023-10-23T19:12:38.789916Z [info     ] Application startup complete.  [uvicorn.error] 
2023-10-23T19:12:38.792650Z [info     ] Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit) [uvicorn.error] 
^C2023-10-23T19:15:04.770399Z [info     ] Shutting down                  [uvicorn.error] 
^C2023-10-23T19:15:04.910332Z [info     ] Finished server process [4257] [uvicorn.error] 
^C2023-10-23T19:15:04.974529Z [info     ] ASGI 'lifespan' protocol appears unsupported. [uvicorn.error] 
```
```


### Steps to reproduce

```bash
- Run `litestar run`
- Look process ressources
```

### Screenshots

```bash
![litestar_profile](https://github.com/litestar-org/litestar/assets/1131885/ab23a525-f85e-479e-8a5d-981fe68c46e3)
```

### Logs

_No response_

### Litestar Version

2.2.1
python3.11

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

provinzkraut:
Fix an issue with the channels backends where they would poll for new subscribers too frequently when no channel subscriptions were present, resulting in a high CPU load.

Fixes #2503

### DIFF
--- a/litestar/channels/backends/redis.py
+++ b/litestar/channels/backends/redis.py
@@ -98,7 +98,9 @@ class RedisChannelsPubSubBackend(RedisChannelsBackend):
         """
 
         while True:
-            message = await self._pub_sub.get_message(ignore_subscribe_messages=True, timeout=self._stream_sleep_no_subscriptions)
+            message = await self._pub_sub.get_message(
+                ignore_subscribe_messages=True, timeout=self._stream_sleep_no_subscriptions
+            )
             if message is None:
                 continue
 
--- a/litestar/channels/backends/redis.py
+++ b/litestar/channels/backends/redis.py
@@ -60,6 +60,7 @@ class RedisChannelsPubSubBackend(RedisChannelsBackend):
         )
         self.__pub_sub: PubSub | None = None
         self._publish_script = self._redis.register_script(_PUBSUB_PUBLISH_SCRIPT)
+        self._has_subscribed = asyncio.Event()
 
     @property
     def _pub_sub(self) -> PubSub:
@@ -77,10 +78,14 @@ class RedisChannelsPubSubBackend(RedisChannelsBackend):
     async def subscribe(self, channels: Iterable[str]) -> None:
         """Subscribe to ``channels``, and enable publishing to them"""
         await self._pub_sub.subscribe(*channels)
+        if not self._has_subscribed.is_set():
+            self._has_subscribed.set()
 
     async def unsubscribe(self, channels: Iterable[str]) -> None:
         """Stop listening for events on ``channels``"""
         await self._pub_sub.unsubscribe(*channels)
+        if not self._pub_sub.subscribed:
+            self._has_subscribed.clear()
 
     async def publish(self, data: bytes, channels: Iterable[str]) -> None:
         """Publish ``data`` to ``channels``
@@ -98,6 +103,7 @@ class RedisChannelsPubSubBackend(RedisChannelsBackend):
         """
 
         while True:
+            await self._has_subscribed.wait()
             message = await self._pub_sub.get_message(
                 ignore_subscribe_messages=True, timeout=self._stream_sleep_no_subscriptions
             )
@@ -199,7 +205,7 @@ class RedisChannelsStreamBackend(RedisChannelsBackend):
         stream_ids: dict[str, bytes] = {}
         while True:
             # We wait for subscribed channels, because we can't pass an empty dict to
-            # xread and block for susbcribers
+            # xread and block for subscribers
             stream_keys = [self._make_key(c) for c in await self._get_subscribed_channels()]
             if not stream_keys:
                 continue
--- a/litestar/channels/backends/redis.py
+++ b/litestar/channels/backends/redis.py
@@ -35,7 +35,7 @@ class RedisChannelsBackend(ChannelsBackend, ABC):
         """
         self._redis = redis
         self._key_prefix = key_prefix
-        self._stream_sleep_no_subscriptions = stream_sleep_no_subscriptions / 1000
+        self._stream_sleep_no_subscriptions = stream_sleep_no_subscriptions
 
     def _make_key(self, channel: str) -> str:
         return f"{self._key_prefix}_{channel.upper()}"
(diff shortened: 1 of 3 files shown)


litestar/channels/backends/redis.py
```diff
@@ -23,6 +23,28 @@
 _XADD_EXPIRE_SCRIPT = (_resource_path / "_redis_xadd_expire.lua").read_text()
 
 
+class _LazyEvent:
```
peterschutt, line 26: What problem does this solve?

provinzkraut: It ensures that the `Event` is created within a running event loop. If it’s not, you’ll sometimes get issues where a future within the event itself is attached to a different loop.

This happens in our tests and could also happen when you create an instance of the backend outside an event loop in your application.

peterschutt: thanks

litestar/channels/backends/redis.py
```diff
@@ -23,6 +23,28 @@
...
+    """A lazy proxy to asyncio.Event that only creates the event once it's accessed"""
```
peterschutt, line 27:
```suggestion
    """A lazy proxy to asyncio.Event that only creates the event once it's accessed.

    It ensures that the Event is created within a running event loop. If it's not, there can be an issue where a future
    within the event itself is attached to a different loop.

    This happens in our tests and could also happen when a user creates an instance of the backend outside an event loop
    in their application.
    """
```
Might be good to add this context for future readers.

litestar/channels/backends/redis.py
```diff
@@ -149,6 +174,7 @@ def __init__(
         self._stream_ttl = stream_ttl if isinstance(stream_ttl, int) else int(stream_ttl.total_seconds() * 1000)
         self._flush_all_streams_script = self._redis.register_script(_FLUSHALL_STREAMS_SCRIPT)
         self._publish_script = self._redis.register_script(_XADD_EXPIRE_SCRIPT)
+        self._has_subscribed_channels = _LazyEvent()
```
peterschutt, line 177:
```suggestion
        self._has_subscribed = _LazyEvent()
```
Nit, but this is consistent with the pub/sub backend.

peterschutt: Could this be assigned in the super class instead of in each subclass?

peterschutt (approved):
Asyncio events seem like a good solution for this!

litestar/channels/backends/redis.py
provinzkraut: I should probably put this in the docstring somewhere for future reference

litestar/channels/backends/redis.py
provinzkraut: Nit on the nit: they represent slightly different things. For the pubsub this guards the `PubSub.subscribed` property, and here it’s whether the channel count is positive.

Although conceptually they’re the same.

litestar/channels/backends/redis.py
```diff
@@ -23,6 +23,35 @@
 _XADD_EXPIRE_SCRIPT = (_resource_path / "_redis_xadd_expire.lua").read_text()
 
 
+class _LazyEvent:
+    """A lazy proxy to asyncio.Event that only creates the event once it's accessed.
```
JacobCoffee, line 27:
```suggestion
    """A lazy proxy to ``asyncio.Event`` that only creates the event once it's accessed.
```
or maybe
```rst
asyncio.Event :class:`~asyncio.Event`
```
 or similar

litestar/channels/backends/redis.py
provinzkraut: > Could this be assigned in the super class instead of in each subclass?

I thought about handling the `_LazyEvent` thing in the superclass, but it felt nicer this way. I’ll sleep on it.

litestar/channels/backends/redis.py
JacobCoffee: oh its private

JacobCoffee (approved):

litestar/channels/backends/redis.py
peterschutt: well don't lose sleep on it - all good either way

PR #2550, opened 2023-10-28
fix: openapi schema for literal and enum unions with None.

### LINKED ISSUE #2546: Bug: Optionally typed parameters do not render in OpenAPI docs properly
### Description

When using an arbitrary enum as a type (or `Literal` assumedly), it generates a dropdown in the OpenAPI docs.

```py
environment: Environment = Parameter(
        title="Environment",
        description="The environment to filter by.",
        default=None,
        required=False,
    ),
```

but hitting this endpoint gives a validation error:

```py
 File ".venv/lib/python3.10/site-packages/litestar/_signature/model.py", line 190, in parse_values_from_connection_kwargs
    return convert(kwargs, cls, strict=False, dec_hook=deserializer, str_keys=True).to_dict()
msgspec.ValidationError: Expected `str`, got `null` - at `$.environment`
```

This makes sense, but the fix of adding ` | None` moves the dropdown field to just a plain textfield.

_No response_

### MCVE

```python
from enum import StrEnum

class Environment(StrEnum):
    DEVELOPMENT = "dev"
    TEST = "test"
    QA = "qa"

@get(path="/")
async def test(
    environment: Environment
    | None = Parameter(
        title="Environment",
        description="The environment to filter by.",
        default=None,
        required=False,
    ),
) -> None:
    return None

app = Litestar([test])
```

### Steps to reproduce

```bash
1. Write mcve.py
2. Browse to OpenAPI schema
3. Observe dropdown fields
4. Execute on the test route
5. See error
6. Apply fix `Environment | None`
7. Reload OpenAPI schema
8. See no dropdown, just text field
9. Execute on the test route
10. See it work, but be deeply saddened that there is no longer a dropdown box and contemplate deleting your entire application.
```

### Screenshots
`: Environment`
image

`: Environment | None`
image

### Logs

_No response_

### Litestar Version

2.x

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows
- [X] Other (Please specify in the description above)

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2546">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2546/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
### Pull Request Checklist

### Description

Existing behavior is to make the schema for every type that is a union with `None` a "one_of" schema, that includes `OpenAPIType.NULL` in the "one_of" types.

https://github.com/litestar-org/litestar/blob/5db8d81fbf24788d38d210e90300e5f0926f830a/litestar/_openapi/schema_generation/schema.py#L319-L340

When a `Literal` or `Enum` type is in a union with `None`, this behavior is not desirable, as we want to have `null` available in the list of available options on the type's schema.

This PR changes literal and enum schema generation so that it can be identified that the types are in a union with `None`, allowing `null` to be included in `Schema.enum` values.

### Close Issue(s)

Closes #2546

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -228,10 +228,7 @@ def create_schema_for_annotation(annotation: Any) -> Schema:
         A schema instance or None.
     """
 
-    if annotation in TYPE_MAP:
-        return copy(TYPE_MAP[annotation])
-
-    return Schema()
+    return copy(TYPE_MAP[annotation]) if annotation in TYPE_MAP else Schema()
 
 
 class SchemaCreator:
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -42,6 +42,11 @@ from litestar._openapi.schema_generation.constrained_fields import (
     create_numerical_constrained_field_schema,
     create_string_constrained_field_schema,
 )
+from litestar._openapi.schema_generation.utils import (
+    _do_enum_schema,
+    _do_literal_schema,
+    _type_or_first_not_none_inner_type,
+)
 from litestar.datastructures.upload_file import UploadFile
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.openapi.spec import Reference
@@ -159,40 +164,6 @@ def _get_type_schema_name(field_definition: FieldDefinition) -> str:
     return name
 
 
-def _first_not_none_inner_type(field_definition: FieldDefinition) -> FieldDefinition:
-    """Get the first inner type that is not None.
-
-    Args:
-        field_definition: A field definition instance.
-
-    Returns:
-        A field definition instance.
-    """
-    return next(t for t in field_definition.inner_types if not t.is_none_type)
-
-
-def _do_enum_schema(field_definition: FieldDefinition) -> bool:
-    """Predicate to determine if we should create an enum schema for the field def, or not.
-
-    This returns true if the field definition is an enum, or if the field definition is a union
-    of an enum and ``None``.
-
-    When an annotation is ``SomeEnum | None`` we should create a schema for the enum that includes ``null``
-    in the enum values.
-
-    Args:
-        field_definition: A field definition instance.
-
-    Returns:
-        A boolean
-    """
-    return field_definition.is_subclass_of(Enum) or (
-        field_definition.is_optional
-        and len(field_definition.args) == 2
-        and field_definition.has_inner_subclass_of(Enum)
-    )
-
-
 def create_enum_schema(annotation: EnumMeta, include_null: bool = False) -> Schema:
     """Create a schema instance for an enum.
 
@@ -208,28 +179,6 @@ def create_enum_schema(annotation: EnumMeta, include_null: bool = False) -> Sche
         enum_values.append(None)
     openapi_type = OpenAPIType.STRING if isinstance(enum_values[0], str) else OpenAPIType.INTEGER
     return Schema(type=openapi_type, enum=enum_values)
-
-
-def _do_literal_schema(field_definition: FieldDefinition) -> bool:
-    """Predicate to determine if we should creat a literal schema for the field def, or not.
-
-    This returns ``True`` if the field definition is an literal, or if the field definition is a union
-    of a literal and None.
-
-    When an annotation is `Literal["anything"] | None` we should create a schema for the literal that includes `null`
-    in the enum values.
-
-    Args:
-        field_definition: A field definition instance.
-
-    Returns:
-        A boolean
-    """
-    return (
-        field_definition.is_literal
-        or field_definition.is_optional
-        and all(inner.is_literal for inner in field_definition.inner_types if not inner.is_none_type)
-    )
 
 
 def _iter_flat_literal_args(annotation: Any) -> Iterable[Any]:
@@ -354,18 +303,10 @@ class SchemaCreator:
         if plugin_for_annotation := self.get_plugin_for(field_definition):
             result = self.for_plugin(field_definition, plugin_for_annotation)
         elif _do_enum_schema(field_definition):
-            annotation = (
-                _first_not_none_inner_type(field_definition).annotation
-                if field_definition.is_optional
-                else field_definition.annotation
-            )
+            annotation = _type_or_first_not_none_inner_type(field_definition)
             result = create_enum_schema(annotation, include_null=field_definition.is_optional)
         elif _do_literal_schema(field_definition):
-            annotation = (
-                _first_not_none_inner_type(field_definition).annotation
-                if field_definition.is_optional
-                else field_definition.annotation
-            )
+            annotation = _type_or_first_not_none_inner_type(field_definition)
             result = create_literal_schema(annotation, include_null=field_definition.is_optional)
         elif field_definition.is_optional:
             result = self.for_optional_field(field_definition)
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -306,7 +306,11 @@ class SchemaCreator:
             annotation = _type_or_first_not_none_inner_type(field_definition)
             result = create_enum_schema(annotation, include_null=field_definition.is_optional)
         elif _do_literal_schema(field_definition):
-            annotation = _type_or_first_not_none_inner_type(field_definition)
+            annotation = (
+                make_non_optional_union(field_definition.annotation)
+                if field_definition.is_optional
+                else field_definition.annotation
+            )
             result = create_literal_schema(annotation, include_null=field_definition.is_optional)
         elif field_definition.is_optional:
             result = self.for_optional_field(field_definition)
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -43,8 +43,8 @@ from litestar._openapi.schema_generation.constrained_fields import (
     create_string_constrained_field_schema,
 )
 from litestar._openapi.schema_generation.utils import (
-    _do_enum_schema,
-    _do_literal_schema,
+    _should_create_enum_schema,
+    _should_create_literal_schema,
     _type_or_first_not_none_inner_type,
 )
 from litestar.datastructures.upload_file import UploadFile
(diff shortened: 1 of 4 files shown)


litestar/_openapi/schema_generation/schema.py
```diff
@@ -274,14 +277,24 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
 
         result: Schema | Reference
 
-        # NOTE: The check for whether the field_definition.annotation is a Pagination type
-        # has to come before the `is_dataclass_check` since the Pagination classes are dataclasses,
-        # but we want to handle them differently from how dataclasses are normally handled.
-        if field_definition.is_optional:
+        if _do_enum_schema(field_definition):
+            annotation = _type_or_first_not_none_inner_type(field_definition)
```
peterschutt, line 281: I didn't use `make_non_optional_union()` like we do for literal types, b/c the literal schema generation has handling for flattening out nested literals, and the enum handling does not.

So a union like `EnumA | EnumB | None` is still going to be rendered as Object.

litestar/typing.py
```diff
@@ -318,12 +318,12 @@ def is_required(self) -> bool:
     @property
     def is_annotated(self) -> bool:
         """Check if the field type is Annotated."""
-        return Annotated in self.type_wrappers  # type: ignore[comparison-overlap]
+        return bool(self.metadata)
```
peterschutt, line 321: This is a non-related change - so I can split out into a separate PR if anyone feels necessary - it just seems better to infer this property this way, as the only way the object can have metadata is if it is annotated, and a type cannot be annotated without at least one metadata argument.

litestar/typing.py
```diff
@@ -318,12 +318,12 @@ def is_required(self) -> bool:
...
 
     @property
     def is_literal(self) -> bool:
         """Check if the field type is Literal."""
-        return get_origin(self.annotation) is Literal
+        return self.origin is Literal
```
peterschutt, line 326: Similar to above, this isn't related to the PR, but a minor improvement to previous.

tests/unit/test_openapi/test_schema.py
```diff
@@ -397,3 +401,39 @@ def test_schema_generation_with_pagination(annotation: Any) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ("in_type", "out_type"),
+    [
+        (FieldDefinition.from_annotation(Optional[int]), int),
+        (FieldDefinition.from_annotation(Union[None, int]), int),
+        (FieldDefinition.from_annotation(int), int),
+        # hack to create a union of NoneType, NoneType to hit a branch for coverage
+        (FieldDefinition.from_annotation(_GenericAlias(Union, (NoneType, NoneType))), ValueError),
```
peterschutt, line 431: This is the dark side of branch coverage.. the only way that the condition of exhausting the generator in the function under test could be hit is if there is no non-None type in the union with None. Not possible in the real world, but here we are.

peterschutt: To preoccupied to see if I could, didn't stop to think if I should.. etc etc.

litestar/_openapi/schema_generation/utils.py
```diff
@@ -15,3 +26,67 @@ def sort_schemas_and_references(values: list[Schema | Reference]) -> list[Schema
...
+    """
+    return field_definition.is_subclass_of(Enum) or (
+        field_definition.is_optional
+        and len(field_definition.args) == 2
+        and field_definition.has_inner_subclass_of(Enum)
+    )
+
+
+def _do_literal_schema(field_definition: FieldDefinition) -> bool:
+    """Predicate to determine if we should creat a literal schema for the field def, or not.
```
JacobCoffee, line 74:
```suggestion
    """Predicate to determine if we should create a literal schema for the field def, or not.
```

JacobCoffee (approved):
Im going to try this out at work since i hit this issue

litestar/typing.py
cofin: I don't think there's any need to split this out into separate PR.

JacobCoffee:
I don't like the extra `null` option on top of `--`, is that required?
image

JacobCoffee:
Works good for me otherwise, except the issue discussed here https://discord.com/channels/919193495116337154/1167686060650799165

JacobCoffee:
Running with `field: Enum | None` works well, extra `none` option, and execute works. But without `| None` im hitting this still?  
```json
{
  "status_code": 400,
  "detail": "Validation failed for GET http://127.0.0.1:8000/",
  "extra": [
    {
      "message": "Expected `str`, got `null`",
      "key": "environment",
      "source": "query"
    }
  ]
}
```

litestar/_openapi/schema_generation/schema.py
```diff
@@ -274,14 +277,24 @@ def for_field_definition(self, field_definition: FieldDefinition) -> Schema | Re
...
+            result = create_enum_schema(annotation, include_null=field_definition.is_optional)
```
JacobCoffee, line 282: I think I dislike this because we already generate `--` which is `Null` enough, but we add an ugly `null` dropdown option

peterschutt: If you look at the request that is sent when you select either `---` or `null` they are different. One excludes the param, one sends an explicit `null`.

peterschutt:
> Running with `field: Enum | None` works well, extra `none` option, and execute works. But without `| None` im hitting this still?  
> 
> ```json
> 
> {
> 
>   "status_code": 400,
> 
>   "detail": "Validation failed for GET http://127.0.0.1:8000/",
> 
>   "extra": [
> 
>     {
> 
>       "message": "Expected `str`, got `null`",
> 
>       "key": "environment",
> 
>       "source": "query"
> 
>     }
> 
>   ]
> 
> }
> 
> ```

Is this with `default=None` still declared in `Parameter`?

litestar/_openapi/schema_generation/schema.py
JacobCoffee: i think see how this would be needed if the default were not `None` and it was an optional field.

provinzkraut:
> Running with `field: Enum | None` works well, extra `none` option, and execute works. But without `| None` im hitting this still?
> 
> ```json
> {
>   "status_code": 400,
>   "detail": "Validation failed for GET http://127.0.0.1:8000/",
>   "extra": [
>     {
>       "message": "Expected `str`, got `null`",
>       "key": "environment",
>       "source": "query"
>     }
>   ]
> }
> ```

Can you elaborate? If it's not `| None` then that error seems to be correct?

litestar/_openapi/schema_generation/schema.py
provinzkraut: `str | None` -> Accepts `String` or `null`, param can't be omitted 
`str | None` with a default of `None` -> Accepts `String` or `null`, param can be omitted.

At least that's how I understand it.

litestar/_openapi/schema_generation/utils.py
```diff
@@ -15,3 +26,67 @@ def sort_schemas_and_references(values: list[Schema | Reference]) -> list[Schema
...
+    """
+    if not field_definition.is_optional:
+        return field_definition.annotation
+    inner = next((t for t in field_definition.inner_types if not t.is_none_type), None)
+    if inner is None:
+        raise ValueError("Field definition has no inner type that is not None")
+    return inner.annotation
+
+
+def _do_enum_schema(field_definition: FieldDefinition) -> bool:
```
provinzkraut, line 51:
```suggestion
def _should_create_enum_schema(field_definition: FieldDefinition) -> bool:
```

I think that's the language we usually use for things like these. It's also consistent with the docstring :)

litestar/_openapi/schema_generation/utils.py
```diff
@@ -15,3 +26,67 @@ def sort_schemas_and_references(values: list[Schema | Reference]) -> list[Schema
...
+        A boolean
+    """
+    return field_definition.is_subclass_of(Enum) or (
+        field_definition.is_optional
+        and len(field_definition.args) == 2
+        and field_definition.has_inner_subclass_of(Enum)
+    )
+
+
+def _do_literal_schema(field_definition: FieldDefinition) -> bool:
```
provinzkraut, line 73:
```suggestion
def _should_use_literal_schema(field_definition: FieldDefinition) -> bool:
```

provinzkraut (approved):
Good stuff, just some nits you can feel free to ignore :)

litestar/_openapi/schema_generation/schema.py
peterschutt: We have a matrix of conditions between the annotation of the parameter and its default value. For the annotation, we need to consider whether it is a union with `None`, or not. And we consider whether the param has a default value, or not.

- If annotation is not union with `None`:
    - in any case with default, we don't add `null` to enum members in the schema.
- If annotation is union with `None`:
  - if parameter has default `None`:
      - we wouldn't necessarily _have_ to add the `null` to the enum options in this case as omission is same outcome.
  - if parameter has not `None` default, or parameter has no default:
      - we have to add `null` to the available enums in either of these cases, as it needs to be able to be specified, and isn't default.

> i think see how this would be needed if the default were not None and it was an optional field.

So in the matrix, this is where annotation is union with `None` and has a default `None`, we could omit the `null` value from the list of options. I'm open to being convinced that we should do this, but my preference would be to not do this, for a few reasons:
- inconsistency with the not `None`/no default case
- more complexity, its another condition we have to handle when generating the docs
- I'm not sure that we can police it - i.e., the signature model is going to be configured with `param: MyEnum | None = None` as the field annotation, so if a client sends `?param=null` in their request, the signature model will allow it.

The argument I can think of for allowing a field to be `null` but not allowing it to be provided as `null` is if there should be some distinction between "parameter not provided", and "parameter provided", where `None` is being used as a sentinel value. For this, maybe we could support allowing params to be annotated as union with our `Empty` type, and we could add explicit handling to remove that from openapi spec. One benefit of that is that there is no actual way to specify `Empty` via a query param as it has no encodable value.

JacobCoffee (approved):

litestar/_openapi/schema_generation/schema.py
JacobCoffee: i think what we have here is the most sensible, im just being weird 😅

peterschutt: > im just being weird

this i can relate to

PR #2553, opened 2023-10-28
fix(pydantic): Decouple Pydantic integration

provinzkraut:
Decouple the Pydantic integration from the core logic.

In previous versions, Pydantic integration was baked into the core schema generation, causing issues when trying to run Litestar without it. A recent PR (#2487) surfaced some of these issues, causing an `ImportError` when Pydantic was not installed.

This PR is an attempt to remove all direct references to Pydantic from Litestar's core, although due to other constraints, some of these should be seen as temporary workarounds.

<hr>

One major issue preventing a clean solution here is the merging of the `ParseType` util into `FieldDefinition` that happened in #1912. This causes `FieldDefinition` to require information specific to Pydantic, which is not available to it, since this information is contained within the plugin, `FieldDefinition` however is an ubiquitous utility, used in places where an application context and therefore installed plugins aren't available (nor should they be).

A permanent solution to this would require unmerging these utilities again, creating a separation of concerns between field definitions and metadata and type information.

### DIFF
--- a/litestar/plugins.py
+++ b/litestar/plugins.py
@@ -15,6 +15,7 @@ __all__ = (
     "SerializationPluginProtocol",
     "InitPluginProtocol",
     "OpenAPISchemaPluginProtocol",
+    "OpenAPISchemaPlugin",
     "PluginProtocol",
     "CLIPluginProtocol",
     "PluginRegistry",
@@ -122,7 +123,7 @@ class SerializationPluginProtocol(Protocol):
 
 @runtime_checkable
 class OpenAPISchemaPluginProtocol(Protocol):
-    """Plugin to extend the support of OpenAPI schema generation for non-library types."""
+    """Plugin protocol to extend the support of OpenAPI schema generation for non-library types."""
 
     __slots__ = ()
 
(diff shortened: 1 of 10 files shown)


litestar/typing.py
```diff
@@ -35,6 +35,19 @@
...
+    @staticmethod
+    def matches(annotation: Any, name: str | None, default: Any) -> bool:
+        ...
+
+    @staticmethod
+    def extract(annotation: Any, default: Any) -> Any:
+        ...
+
+
+_KWARG_META_EXTRACTORS: set[_KwargMetaExtractor] = set()
```
provinzkraut, line 48: This is not a long term solution, but the only somewhat sane way of doing this at the moment. I really don't like using global variables like this, but we were relying on global state before as well, it was just hidden behind the pydantic utility functions. This at least makes it visible and decouples the core utilities from the pydantic utils.

litestar/plugins.py
```diff
@@ -151,10 +152,27 @@ def to_openapi_schema(self, field_definition: FieldDefinition, schema_creator: S
         raise NotImplementedError()
 
 
+class OpenAPISchemaPlugin(OpenAPISchemaPluginProtocol):
```
provinzkraut, line 155: I added a new plugin class here, since adding these methods to the protocol would be breaking. I want to open up a new PR that adds abc's for all of these instead of the protocols, as it makes extension of the plugin interface in a backwards compatible way much easier.

Thoughts @litestar-org/maintainers?

JacobCoffee (approved):

litestar/plugins.py
JacobCoffee: asked probably dumb question in DMs, otherwise i think its good

provinzkraut:
The coverage reports are not on new code but on already existing code, it's just that I touched these files. Did something change there recently? @peterschutt I think you made some changes to how we report coverage, but I don't think they should affect this?

I would prefer to not address these coverage issues in this PR though, since they don't really fit this scope, so I'd like to ignore them and still go ahead with the merge. That fine with everyone @litestar-org/maintainers?

peterschutt:
> I think you made some changes to how we report coverage

The main change was https://github.com/litestar-org/litestar/pull/2528 where we adopted the covdefaults plugin. The plugin enables branch coverage, and it might just be that there hasn't been a PR merged into main that has had the branch coverage reported for this file, so it is reporting it here.

We have min 80% coverage configured here: https://sonarcloud.io/organizations/litestar-api/quality_gates/show/9 but I think that is referring to line coverage. Condition coverage is 100% by default so we could look at adding a rule to lessen that constraint, which would probably make this transition period easier.

peterschutt (approved):
Lgtm!

PR #2554, opened 2023-10-28
feat: SSL Context for the litestar CLI

### LINKED ISSUE #2335: Enhancement: Self-signed certificate generation for CLI
### Summary

This is along the lines what Flask (actually Werkzeug) and Django-extensions have. A new CLI option (or two) would be added to the `run` command
- `--ssl-certfile` to specify the location of the certificate
- `--ssl-keyfile` to specify the location of the keyfile
- If only one of them is provided, the second file is assumed to be in the same directory
- If the files are not found, they are automatically generated

The logic is taken from [`runserver_plus` in Django Extensions](https://django-extensions.readthedocs.io/en/latest/runserver_plus.html#ssl) and the names are taken from [Uvicorn](https://www.uvicorn.org/settings/#https)

The paths would be then forwarded to Uvicorn.

It would require the `cryptography` extras (which are already used for client sessions) and the `cryptography` module would be used to generate the self-signed certificates. 

[HERE](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/serving.py#L540) is how Werkzeug handles the cert generation and [HERE](https://github.com/django-extensions/django-extensions/blob/main/django_extensions/management/commands/runserver_plus.py#L404) is how Django-Extensions handles the logic.

It is only meant to be used for development purposes on a HTTPS connection of course and would only generate a self-signed certificate.

### Basic Example

`litestar run --ssl-certfile=cert/server.crt`  - looks for `server.key` inside of the `cert/` directory
`litestar run --ssl-keyfile=cert/server.key` - looks for `server.crt` inside of the `cert/` directory
`litestar run --ssl-cerfile=cert/server.crt --ssl-keyfile=security/keys/server.key`
- if both files exist, use them
- otherwise generate new ones using `cryptography`
- if `cryptography` extras are not installed, terminate and inform user

### Drawbacks and Impact

_No response_

### Unresolved questions

The exact logic and option names are open for discussion.

---

## Funding

<a href="https://polar.sh/litestar-org/litestar/issues/2335">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2335/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

geeshta:
### Pull Request Checklist

### Description
This PR implements two functionalities. I haven't touched tests and documentation yet, please let me know if this implementation makes sense to you.

#### Passing SSL cert files to Uvicorn
The `--ssl-certfile` and `--ssl-keyfile` flags are passed to uvicorn when using `litestar run`. Uvicorn requires both to be passed (or neither) but additional validation was added to generate a more user friendly CLI errors.

The other SSL-related flags (like password or CA) were not added (yet). See [uvicorn CLI docs](https://www.uvicorn.org/#command-line-options)

#### Generation of a self-signed certificate
One more CLI flag was added (--create-devcert) that uses the `cryptography` module to generate a self-signed development certificate. Both of the previous flags must be passed when using this flag. Then the following logic is used:
- If both files already exists, they are used and nothing is generated
- If neither file exists, the dev cert and key are generated
- If only one file exists, it is ambiguous what to do so an exception is raised

#### Usage
`litestar run --ssl-certfile=cert.pem --ssl-keyfile=key.pem`
Validates that both flags were given, both files exist and are not directories. Then pass the resolved paths to uvicorn

`litestar run --ssl-certfile=cert.pem --ssl-keyfile=key.pem --create-devcert`
If both files already exist, pass them to uvicorn. If only one exists, raise an error. If neither exists, create a new self-signed cert and a passwordless private key at the given location.

### Close Issue(s)
Closes #2335


cofin:
Nice work.  What's the best way for us to test this?

cofin:
Should we name the parameter `--create-self-signed-cert` instead of `--create-devcert`?

I think there may be some valid use cases outside of dev.

geeshta:
> Nice work. What's the best way for us to test this?

Well I'm not quite sure about that. Intuitively after running the server, I would connect with a test client using https and check if I get a valid response.

> I think there may be some valid use cases outside of dev.

I'm not sure about that, most clients like curl will refuse connections to such server without using a flag to explicitely allow it and browsers display a warning. Or do you know about any other use cases?

geeshta:
Okay now I see that the calls to uvicorn.run and subprocess.run are mocked in the cli tests. I am sorry unit testing is not my greates strength :D

cofin:
> I'm not sure about that, most clients like curl will refuse connections to such server without using a flag to explicitely allow it and browsers display a warning. Or do you know about any other use cases?

I was thinking maybe auto-securing service-to-service communication.  I can't think of a specific use case immediately though.  Certainly in the context of a public facing web server, it should signed.

geeshta:
> Nice work. What's the best way for us to test this?
I've written tests for cli argument handling and cert generation. 
I'm not sure how to write tests that run the app using the `CliRunner` and then use the `TestClient` against it. If I remove the uvicorn mock, the app runs and the execution of the tests hangs. But the ssl files get accepted by uvicorn without raising an exception.
If there is a need to actually run the app or even fire some requests against it, I need some help with that.

litestar/cli/_utils.py
```diff
@@ -410,3 +423,127 @@ def show_app_info(app: Litestar) -> None:  # pragma: no cover
...
+    """Validate if a path's parent directory exist. Return the resolved path."""
+    path = Path(file_path).resolve()
+
+    if not (parent_dir := path.parent).exists():
+        raise LitestarCLIException(f"Unable to create file. Directory doesn't exist: {parent_dir}")
+
+    return path
+
+
+def validate_and_create_ssl_files(
```
provinzkraut, line 454: This should be split into two different functions. You have to branches here, guarded by the `create_devcert` bool, that have completely different behaviour.

Additionally, I think the logic can be simplified a bunch. We don't need to do the whole path validation thing if we're not creating a devcert. Uvicorn will validate the file paths provided anyway. So all we need to do is, in case the `create-devcert` flag is passed, check if both paths provided do not exist, in which case we generate a cert, or if only one of them exists, in which case we raise an exception.

tests/unit/test_cli/test_core_commands.py
```diff
@@ -48,6 +48,9 @@ def mock_show_app_info(mocker: MockerFixture) -> MagicMock:
         (False, None, 2),
     ],
 )
+@pytest.mark.parametrize(
+    "ssl_certfile, ssl_keyfile, create_devcert", [(None, None, False), ("cert.pem", "key.pem", True)]
+)
```
provinzkraut, line 53: This should go into a separate test case, as adding this parametrization adds another 128 test cases. This test case should only have options that are known to / could interfere with each other, since we want to test all combinations of those. Since the certs don't interact with anything else, they don't need to be included in here.

provinzkraut (changes requested):
Thanks for the addition @geeshta! Just a couple of things I think that need to be changed before this can go in

provinzkraut:
> Should we name the parameter `--create-self-signed-cert` instead of `--create-devcert`?

+1 on this, also because it's more explicit. 

Additionally, I think we should raise `UserWarning` that a cert is being generated / a self-signed cert is being used. Maybe even a big warning in the CLI, with bold, red text in a box.

Also, SSL being active / whether it's a self-signed cert should be displayed in the app info.

litestar/cli/_utils.py
geeshta: @provinzkraut
> You have to branches here

The problem is after adding a single `if` to the function, it doesn't pass the pre-commit checks, ruff complains that it is too complex ([C901](https://www.flake8rules.com/rules/C901.html))

> I think the logic can be simplified a bunch. We don't need to do the whole path validation thing if we're not creating a devcert. Uvicorn will validate the file paths provided anyway

The uvicorn validation is very bare bones and the error messages are not nice and not even consistent for when the key file is missing vs cert file is missing. Given that it might be not intuitive what different combinations of flags do, I wanted to give nice and explicit CLI errors that would guide the user. 
Do you really want to get rid of that?

geeshta:
> > Should we name the parameter `--create-self-signed-cert` instead of `--create-devcert`?
> 
> +1 on this, also because it's more explicit.
> 
> Additionally, I think we should raise `UserWarning` that a cert is being generated / a self-signed cert is being used. Maybe even a big warning in the CLI, with bold, red text in a box.
> 
> Also, SSL being active / whether it's a self-signed cert should be displayed in the app info.

I was thinking the same. I am just not sure if we can recognize whether a self-signed or a proper CA-signed certificate is being used.

litestar/cli/_utils.py
provinzkraut: > The problem is after adding a single `if` to the function, it doesn't pass the pre-commit checks, ruff complains that it is too complex ([C901](https://www.flake8rules.com/rules/C901.html))

That's why I'm suggesting to split this into 2 functions. This is a classic "boolean trap"; A function that has two completely different behaviours that depend on a boolean argument passed to the function. The only thing these branches have in common is that they return the same thing. This complexity can be avoided by taking what's currently in the branches, putting it into two separate functions, and calling them based on the boolean you'd pass to the flag. 


> The uvicorn validation is very bare bones and the error messages are not nice and not even consistent for when the key file is missing vs cert file is missing.

Fair point. If we're adding value here then it's worth having this validation :+1:

provinzkraut:
> I was thinking the same. I am just not sure if we can recognize whether a self-signed or a proper CA-signed certificate is being used.

Well, we know if the user passes the `--create-devcert` flag. If they supply a self-signed certificate outside of that, it's their own responsibility and I'd assume they know what they're doing. This is more to prevent accidentally using this flag / env var, or for users who might not have read the documentation properly / are not aware they shouldn't be using this to actually deploy public-facing SSL.

litestar/cli/_utils.py
geeshta: I'll deal with the linter and try to simplify it

geeshta: @provinzkraut I understand that and hopefully have fixed it. The complexity warning was on the already existing `run_command` function so I silenced it there. My code should be a cleaner now

geeshta:
> Also, SSL being active / whether it's a self-signed cert should be displayed in the app info.

@provinzkraut This would require extending the Litestar and AppConfig objects to carry this information, setting it [like this](https://github.com/litestar-org/litestar/blob/main/litestar/cli/commands/core.py#L125) in the `run_command` function and then getting it [like this](https://github.com/litestar-org/litestar/blob/main/litestar/cli/_utils.py#L380) in the `show_app_info` function.

Do I understand this correctly?

provinzkraut:
> > Also, SSL being active / whether it's a self-signed cert should be displayed in the app info.
> 
> @provinzkraut This would require extending the Litestar and AppConfig objects to carry this information, setting it [like this](https://github.com/litestar-org/litestar/blob/main/litestar/cli/commands/core.py#L125) in the `run_command` function and then getting it [like this](https://github.com/litestar-org/litestar/blob/main/litestar/cli/_utils.py#L380) in the `show_app_info` function.
> 
> Do I understand this correctly?

I was thinking we could just add a `ssl_enabled` parameter to the `show_app_info` function. If we call it from the `run` command, we pass it, and in other cases we don't. 

Thinking about it again now though, it's not a configuration on the app object, and therefore shouldn't be shown there. We should probably instead show it in a dedicated "server info" section, which would also include information about the workers and such. That would be a separate PR then, so IMO this can be left out here.

Issue #2535, opened 2023-10-26
Enhancement: Improve test coverage


JacobCoffee:
### Summary

Coverage is around 90-98%, depending on who you ask.

https://sonarcloud.io/component_measures?id=litestar-org_litestar&metric=new_coverage&view=list

peterschutt:
If #2528 gets a green light these numbers will get fractionally worse due to branch coverage being enabled.

We also have a lot of `# pragma: no cover` that should really be covered instead of ignored.

JacobCoffee:
Anyone is free to find uncovered tests and add them, even if its just 1-2 :)

provinzkraut:
I like what @sobolevn suggested: https://github.com/wemake-services/coverage-conditional-plugin

This looks to be quite useful.

PR #2583, opened 2023-11-01
test: coverage for contrib.pydantic.

peterschutt:
### Pull Request Checklist

### Description

This PR adds tests for coverage in the `contrib.pydantic` namespace.

### Close Issue(s)

For #2535

### DIFF
--- a/litestar/contrib/pydantic/utils.py
+++ b/litestar/contrib/pydantic/utils.py
@@ -12,8 +12,6 @@ from litestar.utils.typing import (
     get_origin_or_inner_type,
     get_type_hints_with_generics_resolved,
     instantiable_type_mapping,
-    unwrap_annotation,
-    wrapper_type_set,
 )
 
 # isort: off
@@ -116,9 +114,6 @@ def pydantic_unwrap_and_get_origin(annotation: Any) -> Any | None:
         return get_origin_or_inner_type(annotation)
 
     origin = annotation.__pydantic_generic_metadata__["origin"]
-    if origin in wrapper_type_set:
-        inner, _, _ = unwrap_annotation(annotation)
-        origin = get_origin_or_inner_type(inner)
     return instantiable_type_mapping.get(origin, origin)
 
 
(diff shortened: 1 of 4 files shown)


litestar/contrib/pydantic/utils.py
```diff
@@ -116,9 +114,6 @@ def pydantic_unwrap_and_get_origin(annotation: Any) -> Any | None:
         return get_origin_or_inner_type(annotation)
 
     origin = annotation.__pydantic_generic_metadata__["origin"]
-    if origin in wrapper_type_set:
-        inner, _, _ = unwrap_annotation(annotation)
-        origin = get_origin_or_inner_type(inner)
```
peterschutt, line 121: @guacs can you review this?

I couldn't get coverage on these lines b/c the wrapper types (e.g., `Annotated`) don't have a `__pydantic_generic_metadata__` attribute. I.e., sending in `Annotated[SomePydanticModel, ...]` just fails with attribute error.

If removing the lines isn't the right thing to do, the function will need a refactor.

tests/unit/test_contrib/test_pydantic/test_integration.py
```diff
@@ -1,26 +1,36 @@
...
+from litestar.params import Body, Parameter
 from litestar.status_codes import HTTP_400_BAD_REQUEST
 from litestar.testing import create_test_client
 from tests.unit.test_contrib.test_pydantic.models import PydanticPerson, PydanticV1Person
 
 from . import PydanticVersion
 
 
-def test_pydantic_v1_validation_error_raises_400() -> None:
+@pytest.mark.parametrize(("meta",), [(None,), (Body(media_type=RequestEncodingType.URL_ENCODED),)])
```
peterschutt, line 19: testing with form encoded data in addition to json sees that we hit the overridden `decode_builtins()` method on the `PydanticDTO`.

JacobCoffee (approved):

provinzkraut (approved):

litestar/contrib/pydantic/utils.py
guacs: I think this should be fine since we're moving all the parsing of the `Annotated` logic to `FieldDefinition`. I'm assuming by the time this function ever gets called, the `Annotated` would have been stripped.

peterschutt: sweet!

PR #2587, opened 2023-11-01
refactor: openapi default values

peterschutt:
### Pull Request Checklist

### Description

`Litestar.update_openapi_schema()` is called whenever we first generate the openapi schema for an app object, and whenever we register a handler on the app.

Inside, we perform checks to ensure that the `OpenAPI` object has a not-none `components` attribute, and that the components attribute has a not-none `schema` attribute.

This PR modifies the definition of the `OpenAPI` type, and the `Component` type to remove the `None` union and instantiate an instance of the respective instance on construction.

Background to the change is that when addressing uncovered lines of code in the `app.py` module, I was unsure of how to hit this line: https://github.com/litestar-org/litestar/blob/fd06486e2ad4ed0a41636659fec4f093a09e3dd0/litestar/app.py#L835

Then realised that perhaps this whole block of code: https://github.com/litestar-org/litestar/blob/fd06486e2ad4ed0a41636659fec4f093a09e3dd0/litestar/app.py#L829-L835 might be unnecessary if we adjust these default values.

### Close Issue(s)

-

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -28,7 +28,6 @@ from litestar.exceptions import (
 from litestar.logging.config import LoggingConfig, get_logger_placeholder
 from litestar.middleware.cors import CORSMiddleware
 from litestar.openapi.config import OpenAPIConfig
-from litestar.openapi.spec.components import Components
 from litestar.plugins import (
     CLIPluginProtocol,
     InitPluginProtocol,
@@ -826,14 +825,6 @@ class Litestar(Router):
 
         operation_ids: list[str] = []
 
-        if not self._openapi_schema.components:
-            self._openapi_schema.components = Components()
-            schemas = self._openapi_schema.components.schemas = {}
-        elif not self._openapi_schema.components.schemas:
-            schemas = self._openapi_schema.components.schemas = {}
-        else:
-            schemas = {}
-
         for route in self.routes:
             if (
                 isinstance(route, HTTPRoute)
@@ -848,7 +839,7 @@ class Litestar(Router):
                     plugins=self.plugins.openapi,
                     use_handler_docstrings=self.openapi_config.use_handler_docstrings,
                     operation_id_creator=self.openapi_config.operation_id_creator,
-                    schemas=schemas,
+                    schemas=self._openapi_schema.components.schemas,
                 )
                 self._openapi_schema.paths[route.path_format or "/"] = path_item
 
--- a/litestar/openapi/config.py
+++ b/litestar/openapi/config.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+from copy import deepcopy
 from dataclasses import dataclass, field, fields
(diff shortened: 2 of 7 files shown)


guacs (approved):
This is a simple, but great change imo!

litestar/app.py
```diff
@@ -826,14 +825,6 @@ def update_openapi_schema(self) -> None:
 
         operation_ids: list[str] = []
 
-        if not self._openapi_schema.components:
-            self._openapi_schema.components = Components()
-            schemas = self._openapi_schema.components.schemas = {}
-        elif not self._openapi_schema.components.schemas:
-            schemas = self._openapi_schema.components.schemas = {}
-        else:
-            schemas = {}
```
peterschutt, line 835: The thing I didn't understand here was the case where the openapi config object defined a `components` object with a `schemas` dict that had values in it. In that case, we'd drop into the `else` clause and use a dict that isn't bound to the openapi object at all.

litestar/openapi/config.py
```diff
@@ -133,7 +134,7 @@ def to_openapi_schema(self) -> OpenAPI:
         return OpenAPI(
             external_docs=self.external_docs,
             security=self.security,
-            components=cast("Components", self.components),
+            components=deepcopy(self.components),  # deepcopy prevents mutation of the config's components
```
peterschutt, line 137: The main reason we need this is that `DEFAULT_OPENAPI_CONFIG` is a global, and if `components.schemas` is defined on the config object, then it is otherwise the same dict that makes its way onto the openapi schema. That dict is then mutated as the schema is built, which is actually mutating the components value on the global config.

In the implementation that is superseded by this PR, this was not required b/c the schemas dict on the openapi object would be overwritten every time an application is constructed, if `OpenAPIConfig.components`, or `OpenAPIConfig.components.schemas` were not defined.

litestar/app.py
guacs: I think that was quite simply a bug.

litestar/openapi/config.py
provinzkraut: > The main reason we need this is that `DEFAULT_OPENAPI_CONFIG` is a global

I'd really like to get rid of that global, but I think that would technically breaking. Is there any reason besides this to keep it around?

litestar/app.py
peterschutt: yep, definitely looks that way!

litestar/openapi/config.py
guacs: > > The main reason we need this is that `DEFAULT_OPENAPI_CONFIG` is a global
> 
> I'd really like to get rid of that global, but I think that would technically breaking. Is there any reason besides this to keep it around?

Then the default would have to be some special marker instance and then we'll have to create the default config after checking if the value is the marker. I don't think this is breaking since the default config will not change and the user can still set `None` to disable OpenAPI.

PR #2590, opened 2023-11-01
chore: simplify unwrap_partial

provinzkraut:
Simplify `utils.helpers.unwrap_partials`

### DIFF
--- a/litestar/utils/helpers.py
+++ b/litestar/utils/helpers.py
@@ -82,7 +82,9 @@ def unwrap_partial(value: MaybePartial[T]) -> T:
     Returns:
         Callable
     """
-    return cast("T", value.func) if isinstance(value, partial) else value
+    from litestar.utils.sync import async_partial
+
+    return cast("T", value.func if isinstance(value, (partial, async_partial)) else value)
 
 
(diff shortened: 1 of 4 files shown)


litestar/utils/helpers.py
```diff
@@ -81,10 +82,9 @@ def unwrap_partial(value: MaybePartial[T]) -> T:
     Returns:
         Callable
     """
-    output: Any = value.func if hasattr(value, "func") else value  # pyright: ignore
-    while hasattr(output, "func"):
-        output = output.func
-    return cast("T", output)
+    if isinstance(value, partial):
```
provinzkraut, line 85: We don't actually need that loop, since `partial` never does the nested wrapping. Internally, it reuses the original func when a `partial` instance is passed.

peterschutt (approved):
Nice, TIL!

provinzkraut:
> Nice, TIL!

Or maybe not, because it seems to fail in some cases. Not sure why tho, will investigate later.

JacobCoffee (approved):

provinzkraut:
> > Nice, TIL!
> 
> Or maybe not, because it seems to fail in some cases. Not sure why tho, will investigate later.

It was our `async_partial` function that was making a bit of a mess :)

PR #2618, opened 2023-11-04
refactor(utils): Remove unused utils

provinzkraut:
Remove unused utils


JacobCoffee (approved):
what if someone is using these out in the wild? These don't seem privatized

provinzkraut:
> what if someone is using these out in the wild? These don't seem privatized

They are not, they're just "implicitly private". We should move them to a private module. At the moment they are simply intentionally undocumented because we have been treating them as if they were private; We usually don't care about backwards compatibility there, don't advertise their usage in the docs and they are missing from the reference docs.

I've put up a discussion on discord about this already but feel free to share your thoughts here as well.

tests/unit/test_utils/test_typing.py
```diff
@@ -68,93 +40,3 @@ def test_get_origin_or_inner_type() -> None:
...
-            {
-                "bound_foo": BoundFoo[int],
-                "constrained_foo": ConstrainedFoo[Union[int, str]],  # type: ignore[type-var]
-                "constrained_foo_with_t": ConstrainedFoo[int],
-            },
-        ),
-    ),
-)
-def test_get_type_hints_with_generics(annotation: Any, expected_type_hints: dict[str, Any]) -> None:
-    assert get_type_hints_with_generics_resolved(annotation, include_extras=True) == expected_type_hints
```
guacs, line 160: Why is this test being removed?

guacs (approved):
LGTM.

tests/unit/test_utils/test_typing.py
provinzkraut: Because I messed something up while rebasing :)

PR #2619, opened 2023-11-04
chore: simplify _unwrap_implicit_optional_hints

provinzkraut:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-


litestar/utils/signature.py
```diff
@@ -76,16 +76,6 @@ def _is_two_arg_optional(origin: Any, args: Any) -> bool:
...
 
-    def _is_any_optional(origin: Any, args: Any) -> bool:
-        """Detect if a type is a union with `NoneType`.
-
-        After detecting that a type is a two-argument optional type, this function can be used to detect if the
-        inner type is a union with `NoneType` at all.
-
-        We only want to perform the unwrapping of the optional union if the inner type is optional as well.
-        """
-        return origin is Union and any(arg is type(None) for arg in args)
```
provinzkraut, line 87: I think this may be unnecessary. I discovered this while working on improving the test coverage.

We call this function with an unwrapped annotation, and then check of any of the inner types is a `NoneType`. Iiuc this can't happen though because the only case where this could be true if there were some nested unions with `NoneType`, which have already been flattened by `get_type_hints`, which is where we originally get the hints from that are processed here. So effectively all this is doing is check if the unwrapped notation of the first argument is a `Union`.

I'm not entirely sure of the intent here @peterschutt, so please let me know if I'm missing something here. All tests seem to pass fine though with this change.

peterschutt: From memory it is to distinguish between the following cases, prior to 3.11:

First, if the type is an `Annotated` type containing an optional union, it isn't fully flattened resulting in an optional optional union - this is the one we want to remove the implicit optional union from:

```python-console
Python 3.8.18 (default, Oct 26 2023, 08:29:21) 
[GCC 11.4.0] on linux
>>> from typing_extensions import Annotated, Union, get_type_hints
>>> def fn(t: Annotated[Union[int, None], ...] = None) -> None: ...
... 
>>> get_type_hints(fn)
{'t': typing.Union[typing.Union[int, NoneType], NoneType], 'return': <class 'NoneType'>}
```

The next case is where we have a 2 arg union, that should keep the implicit optional:

```python-console
>>> def fn(t: Annotated[Union[int, str], ...] = None) -> None: ...
... 
>>> get_type_hints(fn)
{'t': typing.Union[typing.Union[int, str], NoneType], 'return': <class 'NoneType'>}
```

peterschutt: I'm very happy to simplify this down though if you think it is unnecessary - the 2nd case that leverages the implicit optional isn't recommended anymore, but it was back in 3.8 days.

And, sorry for not having a test to hit the condition.

provinzkraut: > From memory it is to distinguish between the following cases, prior to 3.11:
> 
> First, if the type is an `Annotated` type containing an optional union, it isn't fully flattened resulting in an optional optional union - this is the one we want to remove the implicit optional union from:
> 
> ```
> Python 3.8.18 (default, Oct 26 2023, 08:29:21) 
> [GCC 11.4.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> from typing_extensions import Annotated, Union, get_type_hints
> >>> def fn(t: Annotated[Union[int, None], ...] = None) -> None: ...
> ... 
> >>> get_type_hints(fn)
> {'t': typing.Union[typing.Union[int, NoneType], NoneType], 'return': <class 'NoneType'>}
> ```
> 
> The next case is where we have a 2 arg union, that should keep the implicit optional:
> 
> ```
> >>> def fn(t: Annotated[Union[int, str], ...] = None) -> None: ...
> ... 
> >>> get_type_hints(fn)
> {'t': typing.Union[typing.Union[int, str], NoneType], 'return': <class 'NoneType'>}
> ```

Ah, thanks for the explanation! I couldn't quite figure out what this is supposed to accomplish and since it didn't break anything I figured it must have simply outlived its usefulness. But I'm fine with keeping it as is now, we should just add a test case for this!

peterschutt: > we should just add a test case for this!

Yep, leave that with me!

peterschutt:
closed in favor of #2698

PR #2621, opened 2023-11-04
refactor(utils): Simplify sync utils

provinzkraut:
### Pull Request Checklist

### Description

- Remove unused functionality from `AsyncCallable`
- Remove `async_partial`
- Introduce `ensure_async_callable` helper function

Previously, we used the `AsyncCallable` to ensure a callable was async, always wrapping the receive callable. This PR moves to the `ensure_async_partial` function which simply returns the original callable if it's already an async callable, and only wraps it in a greatly simplified `AsyncPartial` if necessary.

This removes some layers of indirection and allows us to simplify how we're dealing with the callables passed to us.

### Close Issue(s)

-

### DIFF
--- a/litestar/routes/asgi.py
+++ b/litestar/routes/asgi.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING, Any
 from litestar.connection import ASGIConnection
 from litestar.enums import ScopeType
 from litestar.routes.base import BaseRoute
-from litestar.utils.helpers import unwrap_partial
 
 if TYPE_CHECKING:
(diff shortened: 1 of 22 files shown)


litestar/routes/asgi.py
```diff
@@ -33,7 +32,7 @@ def __init__(
         super().__init__(
             path=path,
             scope_type=ScopeType.ASGI,
-            handler_names=[unwrap_partial(route_handler.handler_name)],
```
provinzkraut, line 36: Not sure why this was here in the first place; `router_handler.handler_name` is a string

JacobCoffee (approved):

PR #2624, opened 2023-11-04
refactor(utils): Remove `Ref` 

provinzkraut:
Remove the `Ref` helper as it is not actually needed, saving us quite a bit of extra attribute lookups and complexity. 

<hr>

*Note: This branch is currently targetting `simplify-async-utils` (#2621) but should be merged into `main` after that branch is merged*

### DIFF
--- a/litestar/controller.py
+++ b/litestar/controller.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
+import types
 from collections import defaultdict
 from copy import deepcopy
-from functools import partial
 from operator import attrgetter
 from typing import TYPE_CHECKING, Any, Mapping, Sequence, cast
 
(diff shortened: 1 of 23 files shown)


litestar/controller.py
```diff
@@ -209,7 +209,9 @@ def get_route_handlers(self) -> list[BaseRouteHandler]:
         self_handlers.sort(key=attrgetter("handler_id"))
         for self_handler in self_handlers:
             route_handler = deepcopy(self_handler)
-            route_handler.fn.value = partial(route_handler.fn.value, self)
+            # at the point we get a reference to the handler function, it's unbound, so
+            # we replace it with a regular bound method here
+            route_handler._fn = types.MethodType(route_handler._fn, self)
```
provinzkraut, line 214: This has probably the same outcome in most cases but is technically more correct.

We should improve the way controllers are registered in general so that this + the `deepcopy` isn't necessary. Afaik there's nothing fundamentally preventing this.

litestar/middleware/compression.py
```diff
@@ -173,8 +173,8 @@ def create_compression_send_wrapper(
         bytes_buffer = BytesIO()
         facade = CompressionFacade(buffer=bytes_buffer, compression_encoding=compression_encoding, config=self.config)
 
-        initial_message = Ref[Optional["HTTPResponseStartEvent"]](None)
```
provinzkraut, line 176: Again, I don't know why this was introduced here in the first place other than that it was. I also don't see a reason why simply using `nonlocal` wouldn't work here so I went with that.

JacobCoffee (approved):

PR #2654, opened 2023-11-12
feat: expose `advanced-alchemy` at `litestar.plugins.sqlalchemy`

cofin:
### Pull Request Checklist

### Description

** RFC **

- This PR re-exports the official SQLAlchemy plugins at `advanced_alchemy.extension.litestar` to `litestar.plugins.sqlalchemy`

### Close Issue(s)

-


JacobCoffee (approved):

JacobCoffee:
I think this is okay

guacs:
A few doubts from my end:

1. Are we confident regarding the stability of the API of `advanced-alchemy`? (I know we've discussed this before, and if @cofin is confident, then I think we can confidently move forward).
2. Do we need to add any sanity tests or something to ensure things are working as expected?

cofin:
> 2\. Do we need to add any sanity tests or something to ensure things are working as expected?

The only thing that I could think of would be creating mocks of all the expected method and signatures.  We can go down this path if everyone thinks it's best.  I was hoping to avoid this as long as possible by making sure we follow semver strictly, but maybe that's not really feasible.  Are there other options?

cofin:
> 1. Are we confident regarding the stability of the API of `advanced-alchemy`? (I know we've discussed this before, and if @cofin is confident, then I think we can confidently move forward).

I feel like we've got a very solid set of repo and service APIs.  The rename of `get_or_create` to `get_or_upsert` has been only change to the existing method names over the last several months.  We've changed the signatures a bit from the original `*args, **kwargs`, but they've only gotten more restrictive.  Additionally, the CLI implements all of the alembic commands with their exact parameters, so this makes the CLI a bit easier to deal with.

I think we are OK re: the stability here.

JacobCoffee:
@cofin AA 1.0.0 soon? 👀

guacs:
> > 1. Are we confident regarding the stability of the API of `advanced-alchemy`? (I know we've discussed this before, and if @cofin is confident, then I think we can confidently move forward).
> 
> I feel like we've got a very solid set of repo and service APIs. The rename of `get_or_create` to `get_or_upsert` has been only change to the existing method names over the last several months. We've changed the signatures a bit from the original `*args, **kwargs`, but they've only gotten more restrictive. Additionally, the CLI implements all of the alembic commands with their exact parameters, so this makes the CLI a bit easier to deal with.
> 
> I think we are OK re: the stability here.

That's great! Then I think we can definitely add this and provide "first-class" support for `sqlalchemy`.

guacs:
> > 2. Do we need to add any sanity tests or something to ensure things are working as expected?
> 
> The only thing that I could think of would be creating mocks of all the expected method and signatures. We can go down this path if everyone thinks it's best. I was hoping to avoid this as long as possible by making sure we follow semver strictly, but maybe that's not really feasible. Are there other options?

I'm not really sure either. @litestar-org/members, any ideas?

guacs (approved):

provinzkraut (approved):

PR #2658, opened 2023-11-12
feat: add `server_lifespan` hook

cofin:
### Pull Request Checklist

### Description

-

This is an alternate implementation of [this](https://github.com/litestar-org/litestar/pull/1869/files) PR. 

This allows for a way to add arbitrary callables that run at on CLI server startup and shutdown 

### Close Issue(s)

-


JacobCoffee (approved):

litestar/cli/commands/core.py
```diff
@@ -31,12 +32,27 @@
...
+__all__ = ("info_command", "routes_command", "run_command", "server_lifespan")
 
 if TYPE_CHECKING:
     from litestar import Litestar
 
 
+@contextmanager
+def server_lifespan(app: Litestar) -> Generator[None, Any, None]:
+    for hook in app.on_cli_startup:
+        _call_on_cli_startup_shutdown_hook(hook)
```
guacs, line 44: Is there a reason this is needed instead of directly doing `hook()`?

litestar/app.py
```diff
@@ -538,7 +548,7 @@ async def __call__(
     async def _call_lifespan_hook(self, hook: LifespanHook) -> None:
         ret = hook(self) if inspect.signature(hook).parameters else hook()  # type: ignore
 
-        if is_async_callable(hook):  # pyright: ignore[reportGeneralTypeIssues]
+        if is_async_callable(hook):  # type: ignore[arg-type] # pyright: ignore[reportGeneralTypeIssues]
```
guacs, line 551: Is the pyright ignore needed since there's a `type: ignore`?

guacs (approved):

provinzkraut (changes requested):
Great! One thing though: You’re using the `server_lifespan` context-manager internally that we talked about, but expose individual hooks with the `on_cli_` wording in the public API. Is there a reason for this? I thought we came to the conclusion that there’s not really a downside to having a single `server_lifespan` context manager.

If there’s no particular reason for this, I’d prefer the context manager to be exposed in the public API since it’s more succinct, more useful in a lot of cases (easier to manage things like keeping connections open, calling other context managers, etc.) and more closely mirrors the lifespan context managers for applications.

litestar/app.py
```diff
@@ -566,6 +581,20 @@ async def lifespan(self) -> AsyncGenerator[None, None]:
 
             yield
 
+    @contextmanager
+    def server_lifespan(self) -> Generator[None, None, None]:
```
provinzkraut, line 585: Can this be private for now or just moved to the CLI? I feel like we shouldn't commit to this API to soon

litestar/plugins/base.py
```diff
@@ -96,6 +101,17 @@ def is_debug_mode(app: Litestar):
...
+class ServerLifespanPluginProtocol(Protocol):
+    """Plugin protocol to extend the CLI Server Lifespan."""
+
+    __slots__ = ()
+
+    @contextmanager
+    def server_lifespan(self, app: Litestar) -> Generator[None, None, None]:
+        yield
+
+
```
provinzkraut, line 114:
```suggestion
class CLIPlugin(CLIPluginProtocol):
    """Plugin protocol to extend the CLI Server Lifespan."""

   __slots__ = ()
    
   def on_cli_init(self, cli: Group) -> None:
      """Called when the CLI is initialized. 

    @contextmanager
    def server_lifespan(self, app: Litestar) -> Generator[None, None, None]:
        yield
```

This way we don't need to add another protocol and can extend the base class further without breaking backwards compatibility.

litestar/app.py
JacobCoffee: sounds like a candidate for one of those experimental features sortve

litestar/config/app.py
```diff
@@ -131,6 +131,10 @@ class AppConfig:
         default_factory=list
     )
     """A list of callables returning async context managers, wrapping the lifespan of the ASGI application"""
+    server_lifespan: list[Callable[[Litestar], AbstractContextManager] | AbstractContextManager] = field(
+        default_factory=list
+    )
+    """A list of callables returning async context managers, wrapping the lifespan of the ASGI Server when called from the CLI."""
```
JacobCoffee, line 137: Would a better name for this be service_cli_lifespan or something to match what the docstring states?

litestar/plugins/base.py
JacobCoffee: I think you are missing the trailing `"""` there in the docstring unless it is hiding from me

tests/unit/test_cli/test_core_commands.py
```diff
@@ -294,3 +296,51 @@ def test_version_command(short: bool, runner: CliRunner) -> None:
...
+    runner: CliRunner, mock_uvicorn_run: MagicMock, create_app_file: CreateAppFileFixture
+) -> None:
+    path = create_app_file("_create_app_with_path.py", content=APP_FACTORY_FILE_CONTENT_SERVER_LIFESPAN_PLUGIN)
+    app_path = f"{path.stem}:create_app"
+    result = runner.invoke(cli_command, ["--app", app_path, "run"])
+
+    assert result.exception is None
+    assert result.exit_code == 0
+    assert "i_run_before_startup_plugin" in result.stdout
+    assert "i_run_after_shutdown_plugin" in result.stdout
```
JacobCoffee, line 336: How do we know these happen at their desired points (before startup, after shutdown) and dont possibly leak into stdout somewhere else?

JacobCoffee (changes requested):

tests/unit/test_cli/test_core_commands.py
cofin: I added a test to check the order of the print statements, but i'm not seeing a very easy way to modify our CLI mocks to check that there's a call to uvicorn run in between these.

provinzkraut (approved):

JacobCoffee (approved):

PR #2660, opened 2023-11-12
fix: correct the format of the OpenAPI examples

### LINKED ISSUE #2272: Bug: Incorrectly serialized examples in OpenAPI schema
### Description

The provided example values are serialized incorrectly into the OpenAPI schema.
image

The issue seem to be that due to inheritance simply the [`BaseSchemaObject.to_schema()` method is called for `Example` instances during the schema serialization](https://github.com/litestar-org/litestar/blob/3b3ed502de26bb945857fa8b94af45afdde7e075/litestar/openapi/spec/base.py#L22-L23), because of this all of the fields of [`Example`](https://github.com/litestar-org/litestar/blob/3b3ed502de26bb945857fa8b94af45afdde7e075/litestar/openapi/spec/example.py#L9-L33) ends up being serialized under the example "section" instead of just the value of the [`Example.value`](https://github.com/litestar-org/litestar/blob/3b3ed502de26bb945857fa8b94af45afdde7e075/litestar/openapi/spec/example.py#L20-L25) field.

https://github.com/litestar-org/litestar/blob/3b3ed502de26bb945857fa8b94af45afdde7e075/litestar/openapi/spec/base.py#L22-L23

### MCVE

```python
from typing import Annotated


class TestBody(Struct):
    field1: Annotated[str, Meta(examples=["the answer"])]
    field2: Annotated[int, Meta(examples=[42])]


@post(sync_to_thread=False)
def test(data: TestBody) -> None:
    ...


if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app=Litestar(route_handlers=[test]), port=8080)
```

### Steps to reproduce

```bash
1. Run the MCVE
2. Open any of the auto-generated docs (redoc, swagger, elements)
3. See the request or schema example (depending on docs)
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.0.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

---

## Funding

<a href="https://polar.sh/litestar-org/litestar/issues/2272">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2272/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

guacs:
### Pull Request Checklist

### Description

- This PR fixes the format in which the examples are generated.

### Close Issue(s)

- Fixes #2272.

### DIFF
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -424,6 +424,12 @@ class FieldDefinition:
 
         return is_typeddict(self.origin or self.annotation)
 
+    @property
+    def type_(self) -> Any:
+        """The type of the annotation with all the wrappers removed, including the generic types."""
+
+        return self.origin or self.annotation
+
     def is_subclass_of(self, cl: type[Any] | tuple[type[Any], ...]) -> bool:
         """Whether the annotation is a subclass of the given type.
 
--- a/litestar/_openapi/parameters.py
+++ b/litestar/_openapi/parameters.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from litestar._openapi.schema_generation.utils import get_formatted_examples
 from litestar.constants import RESERVED_KWARGS
 from litestar.enums import ParamType
 from litestar.exceptions import ImproperlyConfiguredException
@@ -18,7 +19,7 @@ if TYPE_CHECKING:
     from litestar._openapi.schema_generation import SchemaCreator
     from litestar.di import Provide
     from litestar.handlers.base import BaseRouteHandler
-    from litestar.openapi.spec import Example, Reference
+    from litestar.openapi.spec import Reference
     from litestar.types.internal_types import PathParameterDefinition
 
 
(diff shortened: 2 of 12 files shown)


guacs:
The OpenAPI schema requires `examples` to be in one of the following formats:
- a list of the example values (this should be the values directly and must not include the metadata of an example)
- a dictionary of the examples where the keys are the example name and the values are the example with the other metadata like the description, summary etc.

Reference: https://spec.openapis.org/oas/latest.html#requestBodyObject

guacs:
For the following model:

```python
@dataclass
class Foo:
    foo: int
```

Before:
image


After:
image

litestar/typing.py
```diff
@@ -424,6 +424,12 @@ def is_typeddict_type(self) -> bool:
 
         return is_typeddict(self.origin or self.annotation)
 
+    @property
+    def type_(self) -> Any:
```
guacs, line 428: I'm not sure if this is the best name for this field. I'm open to suggestions for a better name or keeping this is also fine by me.

provinzkraut: How is this different from `self.origin`? `origin` already has all the wrapper removed, and `origin` can't be `None` (afaik), so the right side of this `or` will never be triggered.

guacs: `origin` is `None` for non-generic types.

provinzkraut: Yes.. Totally missed that. Sorry :upside_down_face:

cofin (approved):

JacobCoffee (approved):

PR #2661, opened 2023-11-12
feat(di): `provide` decorator

provinzkraut:
WIP


provinzkraut:
@guacs feel free to take it from here

guacs:
@provinzkraut is there a reason we need a `BoundProvide`? I was thinking we could just add a `dependency_key` field directly to `Provide`.

provinzkraut:
> @provinzkraut is there a reason we need a `BoundProvide`? I was thinking we could just add a `dependency_key` field directly to `Provide`.

Yes, there is a reason. `dependency_key` doesn't make a lot of sense on `Provide`, since it will be ignored in most cases. This would mean that the kwarg would have to accept `str | None = None`, which in turn means the instance attribute can be `None`. This means we'd have to either validate it at the point where it's being used and raise an exception if it's `None`, or require the name to always be passed, leading to duplication: `dependencies={"foo": Provide(some_foo, dependency_key="foo")`.

guacs:
> > @provinzkraut is there a reason we need a `BoundProvide`? I was thinking we could just add a `dependency_key` field directly to `Provide`.
> 
> Yes, there is a reason. `dependency_key` doesn't make a lot of sense on `Provide`, since it will be ignored in most cases. This would mean that the kwarg would have to accept `str | None = None`, which in turn means the instance attribute can be `None`. This means we'd have to either validate it at the point where it's being used and raise an exception if it's `None`, or require the name to always be passed, leading to duplication: `dependencies={"foo": Provide(some_foo, dependency_key="foo")`.

You're right. The need to allow `dependency_key` to be `str | None` brings in unnecessary complications in `Provide`. I'll stick to using `BoundProvide.`

guacs:
@JacobCoffee is it possible to disable the tests from running on draft PRs?

JacobCoffee:
@guacs yeah i think we can do something like `if: github.event.pr.draft == false` or similar

provinzkraut:
Closing this for now @guacs

Issue #2609, opened 2023-11-03
Docs: Add deployment documentation


JacobCoffee:
### Summary

Create documentation around deployment of Litestar applications.

This would include deployment via

- [x] Manually via `uvicorn`, `hypercorn`, `daphne`, `granian`, etc.
	- [x] https://github.com/litestar-org/litestar/pull/2663
- [ ] Baremetal/VM with something like `systemd` unit files
- [x] Via `supervisor` https://github.com/litestar-org/litestar/pull/2724/files
- [ ] Containers/LXC
- [ ] K8s
- [ ] Behind a proxy (NGINX/Traefik/HaProxy/etc)
  - [x] ``nginx-unit`` (Done in https://github.com/litestar-org/litestar/pull/2615)
- [ ] Deployment through various cloud providers

Any other deployment mechanism that may be beneficial.

## It should probably go in the `Topics` section

JacobCoffee:
This can be done in steps by many people. If you know how to deploy with `systemd` unit files, then right just that section. We can add on in pieces.

provinzkraut:
> It should probably go in Usage section documentation, or topics.

It should go into "Topics"; That's what this section was originally intended for.

JacobCoffee:
Inspiration from various good resources:

- https://flask.palletsprojects.com/en/3.0.x/deploying/
- https://fastapi.tiangolo.com/deployment/
- https://docs.djangoproject.com/en/4.2/howto/deployment/

WilliamStam:
so let me give this a bit of a stab with supervisor (i stopped using systemctl cause you cant actualy "watch" what its doing)

im on ubuntu and i installed (git pulled lol) the repo with the api into `/opt/api` so please change your paths accordingly. i call my app "api" as well so again change as you need. all references to api here refer to the service and not the actual json thing

this is assuming the venv is setup (`python3.12 install -m venv venv`) and the requirements.txt is loaded (`pip install -r requirements.txt` once activating the venv)

----------

place the actual litestar code in an `app.py` file. `app = Litestar()` however your app is setup

create a main.py  file with

```
import uvicorn

if __name__ == "__main__":
    uvicorn.run(
        "app:app",
        port=80,
        host="0.0.0.0",
    )
```
if you use a proxy infront of your api then use this replacing your proxy servers ip

```
import uvicorn

if __name__ == "__main__":
    uvicorn.run(
        "app:app",
        port=80,
        host="0.0.0.0",
        proxy_headers=True,
[...]

PR #2679, opened 2023-11-15
docs: deployment guides for Docker and Supervisor

WilliamStam:
### Description
Added guides for:

- Docker
- Supervisor
- Supervisor cool alias
- Setup on ubuntu 22.04

### Close Issue(s)

Enhances: https://github.com/litestar-org/litestar/issues/2609

(this is an interim pr. i haven't used rst before and looking at formatting and cleanups - i also fail at proper punctuation)

### DIFF
--- a/litestar/datastructures/url.py
+++ b/litestar/datastructures/url.py
@@ -8,11 +8,12 @@ from litestar._parsers import parse_query_string
 from litestar.datastructures import MultiDict
 from litestar.types import Empty
 
-__all__ = ("Address", "URL")
-
-
 if TYPE_CHECKING:
+    from typing_extensions import Self
+
     from litestar.types import EmptyType, Scope
+
+__all__ = ("Address", "URL")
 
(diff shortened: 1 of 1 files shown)


provinzkraut (changes requested):
Thanks @WilliamStam for this! 

As this is, it's a bit bare though for our documentation. The documentation, the topical guide in particular, should not assume a lot of prior knowledge about these things and are meant to be in-depth guides, focusing on technical backgrounds, explaining why things are done a certain way, weighing alternatives, etc. These are meant to be learning oriented, trying to broaden the knowledge about a particular topic.

Currently, this is more of a condensed "how to" guide. I think this should either be am independent blog post (which we could link to from the docs) or, if you're up for it, changed to better fit the purpose of the documentation.

WilliamStam:
i dont mind beefing it out a bit more. would just need some feedback with stuff. the assuming that the setup is done gets covered in the long filename 1. i kinda thought to split it like that out incase somone wants to just see the supervisor stuff etc but aggreed

PR #2689, opened 2023-11-17
feat(template): add ``template_str`` for rendering from strings

### LINKED ISSUE #2687: Enhancement: Allow creating templates from strings
### Summary

Especially when using HTMX, there are lots of little html partials that need to be returned. I could set these all up in a Jinja templates but it's a layer of indirection I don't particularly want to deal with - particularly with HTMX's focus on locality. It would be easier to just do something like `Template.from_str(...)`.

The alternative would be to use something like `yattag` or `airium`, which I'll probably try. If this is the best way to do it maybe `litestar` could suggest one of them as part of being "opinionated".

### Basic Example
````python
@get("/foo")
def my_route() -> Reswap:
    table = f"""
        <table>
        {% for item in items %}
        <TR>
           <TD class="c1"><IMG src="favicon.ico"></TD>
           <TD class="c2">{{item.date}}</TD>
           <TD class="c3">{{item.id}}</TD>
           <TD class="c4"><SPAN>{{item.position}}</SPAN></TD>
           <TD class="c5"><SPAN>{{item.status}}</SPAN></TD>
        {% endfor %}
        </table>
        </TR>
      """
    return Template.from_str(table, context=get_items())
````

### Drawbacks and Impact

I guess there other ways of doing this, perhaps with a custom template engine? I can't see any other way in which it would have much impact

### Unresolved questions

_No response_

### PR DESCRIPTION
**Pull request by JacobCoffee**
### Pull Request Checklist

### Description

- Adds a method to render HTML templates from strings

### Close Issue(s)

- Closes #2687

### DIFF
--- a/docs/examples/templating/returning_templates_jinja.py
+++ b/docs/examples/templating/returning_templates_jinja.py
@@ -1,5 +1,4 @@
 from pathlib import Path
-from typing import Optional
 
 from litestar import Litestar, get
 from litestar.contrib.jinja import JinjaTemplateEngine
@@ -8,14 +7,11 @@ from litestar.template.config import TemplateConfig
 
 
 @get(path="/{template_type: str}", sync_to_thread=False)
-def index(name: str, template_type: Optional[str]) -> Template:
+def index(name: str, template_type: str) -> Template:
     if template_type == "file":
         return Template(template_name="hello.html.jinja2", context={"name": name})
     elif template_type == "string":
         return Template(template_str="Hello <strong>Jinja</strong> using strings", context={"name": name})
-    elif not template_type:
-        # Return something that should raise an error
-        return Template(template_str=None)
 
 
 app = Litestar(
--- a/docs/examples/templating/returning_templates_jinja.py
+++ b/docs/examples/templating/returning_templates_jinja.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from pathlib import Path
 
 from litestar import Litestar, get
@@ -7,11 +9,14 @@ from litestar.template.config import TemplateConfig
 
 
 @get(path="/{template_type: str}", sync_to_thread=False)
-def index(name: str, template_type: str) -> Template:
+def index(name: str, template_type: str | None) -> Template:
     if template_type == "file":
         return Template(template_name="hello.html.jinja2", context={"name": name})
     elif template_type == "string":
         return Template(template_str="Hello <strong>Jinja</strong> using strings", context={"name": name})
+    elif not template_type:
+        # Return something that should raise an error
+        return Template(template_str=None)
 
 
 app = Litestar(
--- a/docs/examples/templating/returning_templates_jinja.py
+++ b/docs/examples/templating/returning_templates_jinja.py
@@ -1,6 +1,5 @@
-from __future__ import annotations
-
 from pathlib import Path
+from typing import Optional
 
 from litestar import Litestar, get
 from litestar.contrib.jinja import JinjaTemplateEngine
@@ -9,7 +8,7 @@ from litestar.template.config import TemplateConfig
 
 
 @get(path="/{template_type: str}", sync_to_thread=False)
-def index(name: str, template_type: str | None) -> Template:
+def index(name: str, template_type: Optional[str]) -> Template:
     if template_type == "file":
         return Template(template_name="hello.html.jinja2", context={"name": name})
     elif template_type == "string":
--- a/docs/examples/templating/returning_templates_jinja.py
+++ b/docs/examples/templating/returning_templates_jinja.py
@@ -6,9 +6,12 @@ from litestar.response import Template
 from litestar.template.config import TemplateConfig
 
 
-@get(path="/", sync_to_thread=False)
-def index(name: str) -> Template:
-    return Template(template_name="hello.html.jinja2", context={"name": name})
+@get(path="/{template_type: str}", sync_to_thread=False)
+def index(name: str, template_type: str) -> Template:
+    if template_type == "file":
+        return Template(template_name="hello.html.jinja2", context={"name": name})
+    elif template_type == "string":
+        return Template(template_str="Hello <strong>Jinja</strong> using strings", context={"name": name})
 
 
 app = Litestar(
--- a/litestar/contrib/mako.py
+++ b/litestar/contrib/mako.py
@@ -121,7 +121,7 @@ class MakoTemplateEngine(TemplateEngineProtocol[MakoTemplate, Mapping[str, Any]]
         self._template_callables.append((key, template_callable))
 
     @staticmethod
-    def render_string(template_string: str, context: Mapping[str, Any] | None = None) -> str:
+    def render_string(template_string: str, context: Mapping[str, Any]) -> str:
         """Render a template from a string with the given context.
 
         Args:
@@ -131,7 +131,6 @@ class MakoTemplateEngine(TemplateEngineProtocol[MakoTemplate, Mapping[str, Any]]
         Returns:
             The rendered template as a string.
         """
-        context = context or {}
         template = _MakoTemplate(template_string)
         return template.render(**context)  # type: ignore[no-any-return]
 
--- a/litestar/contrib/mako.py
+++ b/litestar/contrib/mako.py
@@ -18,13 +18,12 @@ from litestar.template.base import (
 try:
     from mako.exceptions import TemplateLookupException as MakoTemplateNotFound  # type: ignore[import-untyped]
     from mako.lookup import TemplateLookup  # type: ignore[import-untyped]
+    from mako.template import Template as _MakoTemplate  # type: ignore[import-untyped]
 except ImportError as e:
     raise MissingDependencyException("mako") from e
 
 if TYPE_CHECKING:
     from pathlib import Path
-
-    from mako.template import Template as _MakoTemplate  # type: ignore[import-untyped]
 
 __all__ = ("MakoTemplate", "MakoTemplateEngine")
 
@@ -64,7 +63,7 @@ class MakoTemplate(TemplateProtocol):
 
 
 class MakoTemplateEngine(TemplateEngineProtocol[MakoTemplate, Mapping[str, Any]]):
-    """Mako based TemplateEngine."""
+    """Mako-based TemplateEngine."""
 
     def __init__(self, directory: Path | list[Path] | None = None, engine_instance: Any | None = None) -> None:
         """Initialize template engine.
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by JacobCoffee** [comment]
Tested so far using this MCVE:
<details>
<summary>MCVE</sumary>
```py
from litestar.contrib.jinja import JinjaTemplateEngine
from litestar.response import Template
from litestar.template import TemplateConfig

jinja_config = TemplateConfig(engine=JinjaTemplateEngine, directory="templates")


@get()
async def index() -> Template:
    table = """
            <table>
            {% for item in items %}
            <TR>
               <TD class="c1"><IMG src="favicon.ico"></TD>
               <TD class="c2">{{item.date}}</TD>
               <TD class="c3">{{item.id}}</TD>
               <TD class="c4"><SPAN>{{item.position}}</SPAN></TD>
               <TD class="c5"><SPAN>{{item.status}}</SPAN></TD>
            {% endfor %}
            </table>
            </TR>
          """
    return Template(
        template_str=table,
        context={"items": [{"date": "2023-11-14", "id": "1", "position": "jacob", "status": "birthday"}]},
    )


app = Litestar(route_handlers=[index], template_config=jinja_config, debug=True)

if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app)

```
</details>

JacobCoffee:
Can do the tests and docs if it works out

litestar/response/template.py
```diff
@@ -28,13 +28,15 @@ class Template(Response[bytes]):
...
     __slots__ = (
         "template_name",
+        "template_str",
         "context",
     )
 
     def __init__(
         self,
-        template_name: str,
+        template_name: str | None = None,
```
JacobCoffee, line 37: I'm not fully committed to the API being called `template_name`, but I, as a person masquerading as a developer, am bad at naming things

tests/examples/test_templating/test_returning_templates.py
```diff
@@ -1,23 +1,36 @@
+import pytest
 from docs.examples.templating.returning_templates_jinja import app as jinja_app
-from docs.examples.templating.returning_templates_jinja import app as minijinja_app
```
JacobCoffee, line 2: This was apparently an uncaught bug until now :)

litestar/contrib/mako.py
```diff
@@ -121,6 +120,21 @@ def register_template_callable(
...
+        Args:
+            template_string: The template string to render.
+            context: A dictionary of variables to pass to the template.
+
+        Returns:
+            The rendered template as a string.
+        """
+        context = context or {}
+        template = _MakoTemplate(template_string)
+        return template.render(**context)  # type: ignore[no-any-return]
```
JacobCoffee, line 136: would prefer not to blanket `type ignore` these if anyone can provide some guidance

JacobCoffee:
Note to add @Alc-Alc as co-contributor for his suggestion on #[681e87d](https://github.com/litestar-org/litestar/pull/2689/commits/681e87d37075516674167700386a389488627504)

litestar/contrib/mako.py
peterschutt: from memory the mako lib isn't typed itself so its either this or use `cast` i think

litestar/response/template.py
peterschutt: sgtm

litestar/response/template.py
```diff
@@ -45,8 +47,12 @@ def __init__(
     ) -> None:
         """Handle the rendering of a given template into a bytes string.
 
+            .. note:: Either ``template_name`` or ``template_str`` must be provided.
+                If both are provided, ``template_str`` will be used.
```
peterschutt, line 51: should we make them mutually exclusive up-front and raise an error if both are provided?

litestar/response/template.py
```diff
@@ -142,3 +156,16 @@ def to_asgi_response(
...
+
+        Args:
+            template_str: A string representing the template.
+            request: A :class:`Request <.connection.Request>` instance.
+
+        Returns:
+            Rendered content as bytes.
+        """
+        context = self.create_template_context(request)
+        return request.app.template_engine.render_string(template_str, context).encode(self.encoding)  # type: ignore[no-any-return]
```
peterschutt, line 171: should this type ignore be required? I think if the method is added to the template protocol this should be OK?

peterschutt (approved):
Looks good, cheers!

litestar/contrib/minijinja.py
```diff
@@ -174,6 +174,32 @@ def register_template_callable(
         """
         self.engine.add_global(key, pass_state(template_callable))
 
+    def render_string(self, template_string: str, context: Mapping[str, Any] | None = None) -> str:
```
guacs, line 177: Does the type for `context` need to be an optional mapping? From what I could tell, it's passed from the context we get from `create_template_context` which will always return a dictionary.

litestar/response/template.py
guacs: I think it'd be better to raise an error if both are provided as well.

docs/examples/templating/returning_templates_jinja.py
```diff
@@ -6,9 +8,15 @@
 from litestar.template.config import TemplateConfig
 
 
-@get(path="/", sync_to_thread=False)
-def index(name: str) -> Template:
-    return Template(template_name="hello.html.jinja2", context={"name": name})
+@get(path="/{template_type: str}", sync_to_thread=False)
+def index(name: str, template_type: str | None) -> Template:
```
guacs, line 12: Maybe the type for `template_type` can be a `Literal["file", "string"]`? This way the final returning with `template_str` set to `None` wouldn't be needed.

litestar/response/template.py
```diff
@@ -126,9 +133,16 @@ def to_asgi_response(
             else:
                 media_type = MediaType.TEXT
 
-        template = request.app.template_engine.get_template(self.template_name)
-        context = self.create_template_context(request)
-        body = template.render(**context).encode(self.encoding)
+        if self.template_str is not None:
+            body = self._render_from_string(self.template_str, request)
+            media_type = media_type or "text/html"
```
guacs, line 138:
```suggestion
            media_type = media_type or MediaType.HTML
```

guacs:
This is a really nice feature to add!

litestar/contrib/minijinja.py
JacobCoffee: I'm not smart enough to know the answer here 😅

docs/examples/templating/returning_templates_jinja.py
JacobCoffee: I'm not sure I am understanding but I think this is not allowed per my failed tests and https://docs.litestar.dev/latest/usage/routing/parameters.html#supported-path-parameter-types

JacobCoffee: Maybe it is too late for me 😅 If you mean in the func. signature then I can do this I think im just having a disconnect on if its no longer testing the Template being empty but will just throw an error on the `Literal` enums mismatching once it hits none? Again... may be sleepy 😪

guacs: Aah my bad, I meant that you could try setting the type of `template_type` in the handler signature itself. Something like this:

```python
@get(path="/{template_type: str}", sync_to_thread=False)
def index(name: str, template_type: Literal["file", "string"]) -> Template:
    ...
```

litestar/contrib/minijinja.py
```diff
@@ -174,6 +174,31 @@ def register_template_callable(
...
+
+        Returns:
+            The rendered template as a string.
+
+        Raises:
+            TemplateNotFoundException: if no template is found.
+        """
+
+        if template_string is None:
+            raise TemplateNotFoundException("No template string provided", template_name="template_from_string")
```
guacs, line 192: Is this needed? The check for whether `template_string` would have happened before this is ever called.

guacs (approved):
LGTM!

litestar/response/template.py
peterschutt: The issue is the typing of `AppConfig.template_config`: https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/litestar/config/app.py#L199

Mypy infers this as `TemplateConfig[Any]`, it should be `TemplateConfig[TemplateEngineProtocol]` so that mypy can infer `Litestar.template_engine` to be `TemplateEngineProtocol | None` from assignment here: https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/litestar/app.py#L412

I really think we should work toward changing https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/pyproject.toml#L225 to `true` so these types of errors cannot occur.

peterschutt: After that change:

image

JacobCoffee:
@peterschutt / @guacs can you check the latest here: [532cac9](https://github.com/litestar-org/litestar/pull/2689/commits/532cac9c77c8ae22910dbc69096f058b502ca2a5)?

I am only running into `litestar/response/template.py:163: error: Item "None" of "Optional[TemplateEngineProtocol[Any, Any]]" has no attribute "render_string"  [union-attr]` but i cant fight it any longer :\

litestar/response/template.py
```diff
@@ -117,31 +120,32 @@ def to_asgi_response(
...
+                suffixes = PurePath(self.template_name).suffixes
+                for suffix in suffixes:
+                    if _type := guess_type(f"name{suffix}")[0]:
+                        media_type = _type
+                        break
+                else:
+                    media_type = MediaType.TEXT
             else:
-                media_type = MediaType.TEXT
+                media_type = MediaType.HTML
```
peterschutt: I think better to do all of the media type inference stuff in one place.

So we assume that if the media type isn't specified the string it HTML?

litestar/response/template.py
```diff
@@ -117,31 +120,32 @@ def to_asgi_response(
...
-            media_type = media_type or MediaType.HTML
+            body = template_engine.render_string(self.template_str, context)
         else:
-            if not self.template_name and not self.template_str:
-                raise ValueError("Either template_name or template_str must be provided")
-
-            template = request.app.template_engine.get_template(self.template_name)
-            context = self.create_template_context(request)
+            # cast to str b/c we know that either template_name cannot be None if template_str is None
+            template = template_engine.get_template(cast("str", self.template_name))
```
peterschutt: Using the same computation of context in both scenarios, then there is only one line of work required in the `_render_from_string()` method, so I opted to remove instead of passing the context in.

litestar/response/template.py
```diff
@@ -62,6 +62,9 @@ def __init__(
                 the media type based on the template name. If this fails, fall back to ``text/plain``.
             status_code: A value for the response HTTP status code.
         """
+        if not (template_name or template_str):
+            raise ValueError("Either template_name or template_str must be provided.")
+
```
peterschutt, line 67: moved this from the create asgi response method to here as better to raise it asap, and consistent with having the "not both" check in here as well.

I started to experiment with using `overload` to type this method so that type checkers will pick up where both are not `None`, but prob getting carried away.

guacs:
> @peterschutt / @guacs can you check the latest here: [e653101](https://github.com/litestar-org/litestar/pull/2689/commits/e653101fd213e8e7c2f13b2f6e47cf3bd46eebba)?
> 
> I am only running into `litestar/response/template.py:163: error: Item "None" of "Optional[TemplateEngineProtocol[Any, Any]]" has no attribute "render_string" [union-attr]` but i cant fight it any longer :\

@JacobCoffee, sorry, I just saw this now :P

PR #2690, opened 2023-11-17
refactor: ws receive disconnect exc type

peterschutt:
### Pull Request Checklist

### Description

Modifies `WebSocket.wrapped_receive()` to raise `WebSocketDisconnect` instead of `WebSocketException` where `receive()` is called after a disconnect message has already been received.

### Close Issue(s)

-

### DIFF
--- a/litestar/connection/websocket.py
+++ b/litestar/connection/websocket.py
@@ -11,7 +11,7 @@ from litestar.connection.base import (
     empty_send,
 )
 from litestar.datastructures.headers import Headers
-from litestar.exceptions import WebSocketDisconnect, WebSocketException
+from litestar.exceptions import WebSocketDisconnect
 from litestar.serialization import decode_json, decode_msgpack, default_serializer, encode_json, encode_msgpack
 from litestar.status_codes import WS_1000_NORMAL_CLOSURE
 
@@ -72,7 +72,7 @@ class WebSocket(Generic[UserT, AuthT, StateT], ASGIConnection["WebsocketRouteHan
 
         async def wrapped_receive() -> ReceiveMessage:
             if self.connection_state == "disconnect":
-                raise WebSocketException(detail=DISCONNECT_MESSAGE)
+                raise WebSocketDisconnect(detail=DISCONNECT_MESSAGE)
             message = await receive()
             if message["type"] == "websocket.connect":
                 self.connection_state = "connect"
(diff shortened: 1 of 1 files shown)


litestar/connection/websocket.py
```diff
@@ -167,8 +167,6 @@ async def receive_data(self, mode: WebSocketMode) -> str | bytes:
         event = cast("WebSocketReceiveEvent | WebSocketDisconnectEvent", await self.receive())
         if event["type"] == "websocket.disconnect":
             raise WebSocketDisconnect(detail="disconnect event", code=event["code"])
-        if self.connection_state == "disconnect":
-            raise WebSocketDisconnect(detail=DISCONNECT_MESSAGE)  # pragma: no cover
```
peterschutt, line 171: FWICT - I don't think this code is reachable - which probably explains the pragma.

This method calls `self.receive()` which is really `self.receive_wrapper()`, which will either set `self.connection_state` to `"disconnect"` and return the message if its type is `websocket.disconnect`, or (prior to this PR) raise `WebsocketException` if `self.connection_state == "disconnect"` is already true.

litestar/connection/websocket.py
```diff
@@ -72,7 +72,7 @@ def receive_wrapper(self, receive: Receive) -> Receive:
 
         async def wrapped_receive() -> ReceiveMessage:
             if self.connection_state == "disconnect":
-                raise WebSocketException(detail=DISCONNECT_MESSAGE)
```
peterschutt, line 75: This is raised when `receive()` is called on the connection object after we have already received a `websocket.disconnect` message. Should we raise `WebSocketDisconnect` as I've done here (which is consistent with the unreachable code that I've removed in `receive_data()`), or leave it as is and continue to raise the base exception?

JacobCoffee: Since `WebSocketDisconnect` is a subclass of `WebSocketException` I think its okay?

JacobCoffee (approved):

litestar/connection/websocket.py
peterschutt: It's OK for backward compat - but is is correct - i.e., is calling `receive()` on an already closed disconnected connection a different error? Maybe I'm over thinking.

guacs: I think it makes sense to raise a `WebSocketDisconnect` here. This way the users can handle disconnection related errors differently from the more generic `WebSocketException` if needed. If the `WebSocketException` is raised, then they would have to check the error code to determine if it was a disconnection or not.

guacs (approved):

litestar/connection/websocket.py
peterschutt: OK, will roll with it!

PR #2692, opened 2023-11-17
feat: move Piccolo integration into a separate library

cofin:
### Pull Request Checklist

### Description

- This PR creates a `PiccoloPlugin` and `PiccoloSerializationPlugin` and consolidates the existing code under the `litestar.plugins.piccolo` namespace.

### Close Issue(s)

-


litestar/contrib/piccolo.py
```diff
@@ -21,12 +21,22 @@
...
 
+warn_deprecation(
+    deprecated_name="litestar.contrib.piccolo",
+    version="2.3.2",
+    kind="import",
+    removal_in="3.0",
+    info="importing from 'litestar.contrib.piccolo' is deprecated, please import from 'litestar.plugins.piccolo' instead",
+)
+
+
```
peterschutt, line 39: can `PiccoloDTO` be imported from `litestar.plugins.piccolo` so we don't have to duplicate the rest of the module?

peterschutt (approved):
In the project for migrating away from contrib, this is earmarked to be split out to a separate library: https://github.com/orgs/litestar-org/projects/12?pane=issue&itemId=40891907 - is that out of date?

docs/reference/plugins/sqlalchemy/base.rst
provinzkraut, line 1: These should all have a redirect so existing links don't break

provinzkraut:
> In the project for migrating away from contrib, this is earmarked to be split out to a separate library: https://github.com/orgs/litestar-org/projects/12?pane=issue&itemId=40891907 - is that out of date?

Thinking the same. We should probably go straight to the external library instead of creating another module which we're going to deprecate in 3.

cofin:
> > In the project for migrating away from contrib, this is earmarked to be split out to a separate library: https://github.com/orgs/litestar-org/projects/12?pane=issue&itemId=40891907 - is that out of date?
> 
> Thinking the same. We should probably go straight to the external library instead of creating another module which we're going to deprecate in 3.

I'll start the project today and push to the litestar-org.

cofin:
I've created a new `litestar-piccolo` plugin, moved all of the tests over there, and published the initial release.

I've got the updated referencing `litestar.plugins`, should I just totally remove the `litestar.plugin` re-rexport and leave the deprecation in place?  If that's the case, should I also remove the tests?

peterschutt:
> I've got the updated referencing `litestar.plugins`, should I just totally remove the `litestar.plugin` re-rexport and leave the deprecation in place?

Yeh, I think we shouldn't create a `litestar.plugins.piccolo` namespace, and just deprecate the contrib namespace with the dto in it.

> If that's the case, should I also remove the tests?

Yep, if they exist in the new plugin lib - prob just leave a test for the deprecation and availability of the dto from contrib.

provinzkraut:
I know we didn't do this for advanced-alchemy, but I think it's what we should have done in the first place; I think we should keep the piccolo implementation in Litestar for now and only deprecate it, instead of re-exporting. I know that comes with a higher maintenance cost attached, but the alternative would be to eventually have a whole bunch of required extra dependencies added that most people don't actually want to use. If we were to just keep this as-is and add a deprecation warning pointing to the new package, I think that would make for a better DX.

cofin:
Ah, just saw this.  I'm ok with this solution.  Let me close and re-open this PR at this point.

PR #2516, opened 2023-10-25
fix: get_type_hints() optional param compatibility

peterschutt:
Prior to python 3.11, if a func or method param has a default of `None`, then `get_type_hints()` will automagically wrap the annotation in `Union[T, NoneType]`.

Sometimes this is flattened out before we receive the type back:

```python-console
Python 3.8.17 (default, Sep 14 2023, 13:58:34)
[GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import Optional
>>> from typing_extensions import Annotated, get_type_hints
>>>
>>> def simple(p: Optional[str] = None) -> None: ...
...
>>> get_type_hints(simple)
{'p': typing.Union[str, NoneType], 'return': <class 'NoneType'>}
```
However, if that type is annotated:

```python-console
>>> def annotated(p: Annotated[Optional[str], ...] = None) -> None: ...
...
>>> get_type_hints(annotated)
{'p': typing.Union[typing.Union[str, NoneType], NoneType], 'return': <class 'NoneType'>}
```

Which gets more convoluted if `include_extras=True`:

```python-console
>>> get_type_hints(annotated, include_extras=True)
{'p': typing.Union[typing_extensions.Annotated[typing.Union[str, NoneType], Ellipsis], NoneType], 'return': <class 'NoneType'>}
```

This means that when we parse one of these types into a `FieldDefinition`, we get different behavior b/w <3.11 and 3.11+:

```python
hints = get_type_hints(annotated, include_extra=True)
t = FieldDefinition.from_annotation(hints["p"])
# on 3.8-3.10
print(t.metadata)  # ()
# on 3.11+
print(t.metadata)  # (Ellipsis,)
```

...which I suspect may be playing a part in #2514.

This PR reverses the magic optional wrapping performed by `get_type_hints()` on versions < 3.11, so that we can treat the annotation the same downstream.

### Pull Request Checklist

- [ ] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR
- [ ] Pre-Commit Checks were ran and passed
- [ ] Tests were ran and passed

### Description

-

### Close Issue(s)

-

PR #2698, opened 2023-11-18
refactor: improve _unwrap_implicit_optional_hints impl

peterschutt:
### Pull Request Checklist

### Description

This PR changes the implementation `_unwrap_implicit_optional_hints()` so that it will flatten or remove any outer optional union, if an inner annotated union exists.

E.g., `Union[Annotated[Union[str, int], ...], NoneType]` becomes `Annotated[Union[str, int, Nonetype], ...]`.

Also, the case where a non-union annotated annotation is wrapped in an optional union, we move the outer union into the annotated type. e.g., the following func def:

```py
def fn(a: Annotated[int, ...] = None) -> ...
```
Will be parsed as `Union[Annotated[int, ...], NoneType]` by `get_type_hints()` prior to py 3.11. We convert it to `Annotated[Union[int, NoneType], ...]`.

Refactors the test suite to test that for any given annotation, after processing, the type argument to annotated is the same as the type that would be returned from `get_fn_type_hints()` if the annotation was not annotated.

See https://github.com/litestar-org/litestar/pull/2516 for context on why `_unwrap_implicit_optional_hints()` exists in the first place.

### Close Issue(s)

-

### DIFF
--- a/litestar/utils/signature.py
+++ b/litestar/utils/signature.py
@@ -120,14 +120,16 @@ def _unwrap_implicit_optional_hints(defaults: dict[str, Any], hints: dict[str, A
             inner_args = get_args(unwrapped_inner)
 
             if not _is_any_optional(get_origin(unwrapped_inner), inner_args):
-                # this is where hint is like `Union[Union[str, int], NoneType]`, we flatten it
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):
Nice improvement @peterschutt! Much easier to follow now!

PR #2705, opened 2023-11-18
chore: use `pdm run` where possible

cofin:
### Pull Request Checklist

### Description

- This PR does the following:
- remove the `install-pdm.py` file that was fetched for installation
- Set the expected .venv directory as a string so that there is no errror printed when there is no existing venv.
- Set commands to run with `pdm run` when possible to ensure they are executed from the venv.

### Close Issue(s)

-

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
 default_language_version:
   python: "3.8"
+default_stages: ["pre-push"]
 repos:
   - repo: https://github.com/compilerla/conventional-pre-commit
     rev: v2.4.0
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ SHELL := /bin/bash
 .DEFAULT_GOAL:=help
 .ONESHELL:
 USING_PDM		=	$(shell grep "tool.pdm" pyproject.toml && echo "yes")
-ENV_PREFIX		=	$(shell python3 -c "if __import__('pathlib').Path('.venv/bin/pip').exists(): print('.venv/bin/')")
+ENV_PREFIX		=  .venv/bin/
 VENV_EXISTS		=	$(shell python3 -c "if __import__('pathlib').Path('.venv/bin/activate').exists(): print('yes')")
 PDM_OPTS 		?=
 PDM 			?= 	pdm $(PDM_OPTS)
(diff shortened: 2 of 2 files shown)


provinzkraut (approved):

.pre-commit-config.yaml
```diff
@@ -1,5 +1,6 @@
 default_language_version:
   python: "3.8"
+default_stages: ["pre-push"]
```
provinzkraut, line 3: Why this change?

cofin: This is the one that may be able to be left alone.  This sets pre-commit to run only on push instead of after every commit.  I find having the pre-commit process run after every local commit a burden.  It feels like the rules should only be applied when you are pushing something back into the main repo.

provinzkraut: That actually makes a lot of sense, I like it!

Makefile
```diff
@@ -107,20 +108,20 @@ lint: pre-commit type-check 						## Run all linting
...
-	@$(ENV_PREFIX)coverage xml
+	@$(PDM) run pytest tests --cov -n auto
+	@$(PDM) run coverage html
+	@$(PDM) run coverage xml
 	@echo "=> Coverage report generated"
 
 .PHONY: test
 test:  												## Run the tests
 	@echo "=> Running test cases"
-	@$(ENV_PREFIX)pytest tests -n auto
```
peterschutt, line 118: Why do we keep removing the `-n auto` from this? I only made a PR to add it back in yesterday and haven't seen any discussion about it.

Running the tests on multiple cores saves >30s per run.

.pre-commit-config.yaml
peterschutt: @cofin I backed this change out because it caused us to only run a couple of the hooks in CI.

>  It feels like the rules should only be applied when you are pushing something back into the main repo.

Agree - this is what I do but have manually managed the process by not installing the hook at all and running `make lint` once when I'm ready.

Just to experiment I created this as `litestar/.git/hooks/pre-push` and it seems to work:

```bash
#!/bin/sh

pdm run --venv 3.8 make lint
```

...although I get warnings like:

> Inside an active virtualenv /home/peter/.local/share/pdm/venvs/litestar-dj-FOhMr-3.8, reusing it.
Set env var PDM_IGNORE_ACTIVE_VENV to ignore it.

in the output - so maybe specifying the venv isn't necessary.

PR #2712, opened 2023-11-19
test(typing): Improve `litestar.typing` coverage

provinzkraut:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-

### DIFF
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -21,6 +21,7 @@ from litestar.utils.predicates import (
     is_generic,
     is_non_string_iterable,
     is_non_string_sequence,
+    is_union,
 )
 from litestar.utils.typing import (
     get_instantiable_origin,
@@ -125,7 +126,7 @@ def _traverse_metadata(
 
     Args:
         metadata: A list of metadata values from annotation, namely anything stored under Annotated[x, metadata...]
-        is_sequence_container: Whether or not the container is a sequence container (list, tuple etc...)
+        is_sequence_container: Whether the container is a sequence container (list, tuple etc...)
         extra: Extra key values to parse.
 
     Returns:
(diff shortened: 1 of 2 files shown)


provinzkraut:
https://github.com/litestar-org/litestar/blob/88bc82ba6fabbdcca16f6b74a5004053806e0b13/litestar/typing.py#L250

This line is still missing coverage since I'm not sure when this would ever be hit and I didn't just want to create a brute-force test for this. It checks if the annotation is an instance of `AbstractDTO`, which I don't think should ever happen here.

@peterschutt do you maybe have any insights?

peterschutt:
> https://github.com/litestar-org/litestar/blob/88bc82ba6fabbdcca16f6b74a5004053806e0b13/litestar/typing.py#L250
> 
> This line is still missing coverage since I'm not sure when this would ever be hit and I didn't just want to create a brute-force test for this. It checks if the annotation is an instance of `AbstractDTO`, which I don't think should ever happen here.
> 
> @peterschutt do you maybe have any insights?

No insight, no.

However, `_extract_metadata()` is only called in `from_annotation()` and after looking at all of the usages of that, along with `from_kwarg()` and `from_parameter()` I haven't identified a case where we could be passing an  `AbstractDTO` type to it - and as you say, its not something we'd expect to see.

Further, if it is an `AbstractDTO` type, why does the type get wrapped in list and passed as a `metadata` item to `_extract_metadata()`? This is what is hanging me up, b/c it looks kind of deliberate - but there is no constraint or example info stored directly on an `AbstractDTO`.

provinzkraut:
> No insight, no.
> 
> However, `_extract_metadata()` is only called in `from_annotation()` and after looking at all of the usages of that, along with `from_kwarg()` and `from_parameter()` I haven't identified a case where we could be passing an `AbstractDTO` type to it - and as you say, its not something we'd expect to see.
> 
> Further, if it is an `AbstractDTO` type, why does the type get wrapped in list and passed as a `metadata` item to `_extract_metadata()`? This is what is hanging me up, b/c it looks kind of deliberate - but there is no constraint or example info stored directly on an `AbstractDTO`.

Let's remove it then? Doing so doesn't break any tests, which is not surprising so.. 🤷

litestar/typing.py
```diff
@@ -392,8 +393,8 @@ def is_non_string_collection(self) -> bool:
     def bound_types(self) -> tuple[FieldDefinition, ...] | None:
         """A tuple of bound types - if the annotation is a TypeVar with bound types, otherwise None."""
         if self.is_type_var and (bound := getattr(self.annotation, "__bound__", None)):
-            if is_non_string_sequence(bound):
-                return tuple(FieldDefinition.from_annotation(t) for t in bound)
+            if is_union(bound):
```
peterschutt, line 396: This was a bug?

peterschutt (approved):

litestar/typing.py
provinzkraut: Ah, yeah. I was trying to test this but found it didn't even work so I added this small fix

PR #2715, opened 2023-11-20
fix: revert setting "default_stages" in pre-commit

peterschutt:
### Pull Request Checklist

### Description

This change has caused `make lint` and pre-commit in CI to only run two of the hooks.

E.g., https://github.com/litestar-org/litestar/actions/runs/6916362683/job/18816353425#step:6:29

If there is a more nuanced way to solve this then feel free to close this off in favor of another approach.

### Close Issue(s)

-

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,5 @@
 default_language_version:
   python: "3.8"
-default_stages: ["pre-push"]
 repos:
   - repo: https://github.com/compilerla/conventional-pre-commit
     rev: v2.4.0


JacobCoffee (approved):

guacs:
@peterschutt There was some discussion regarding this [here](https://github.com/litestar-org/litestar/pull/2705#discussion_r1398424120). Maybe we can have a discussion to kind of figure out a workflow that everyone's comfortable with?

Personally, I prefer pre-commit running on every commit rather than just before pushing.

JacobCoffee:
Same here, I am most saddened by this and also by type checking being removed. I never remember to do this until i waste time watching the PR actions run and then fail causing me to waste more time

peterschutt:
> @peterschutt There was some discussion regarding this [here](https://github.com/litestar-org/litestar/pull/2705#discussion_r1398424120).

Thanks, I did read that but I don't think its the right solution in light of it breaking CI and `make lint` workflows.

> a workflow that everyone's comfortable with

This is very optimistic :smile_cat: 

> Personally, I prefer pre-commit running on every commit rather than just before pushing.

My workflow is to not install the hook at all. I iterate with tests and mypy, and then run `make lint` once before I push up - sometimes I forget that last bit, but whatever.

Not installing the hook is a habit I formed when we were running mypy and pyright and all sorts of other things in pre-commit and it would take way too long. Even though we've cut down the execution time I still find I like being explicit about when it runs.

Aside from execution time, with the hooks installed and committing through the pycharm IDE, if pre-commit detects an error you get a "push failed" popup but the full output is hidden behind the source control tab and within that hidden again by the console tab, so its click-city to find out why the push failed.

PR #2706, opened 2023-11-19
fix: Explicit grouping for regex in OpenAPI type name normalization

provinzkraut:
### Pull Request Checklist

- [ ] New code has 100% test coverage
- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR
- [ ] Pre-Commit Checks were ran and passed
- [ ] Tests were ran and passed

### Description

#2475 introduced a new regex with a potential for "catastrophic backtracking" ([SonarCloud analysis](https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=BUG&id=litestar-org_litestar&open=AYvgiwx-edsUYnEP43pU)). This is fixed by introducing explicit grouping.

PR #2717, opened 2023-11-20
refactor: OpenAPI type name trimming regex

peterschutt:
### Pull Request Checklist

### Description

This regex is used to remove superfluous chars from the start and end of the schema keys that we generate for openapi.

Sonar reports two issues to do with the original implementation:

1. a reliability issue to do with operator precedence
2. a security issue to do with backtracking

This PR refactors the regex so that it has less parts and less scope for backtracking.

A follow up to #2706 

### Close Issue(s)

-


JacobCoffee (approved):

peterschutt:
As you can see it didn't report anything on this pr either @provinzkraut - but still isn't happy with the pattern on main 🤷‍♂️

provinzkraut:
That's quite annoying. Fixing something like this by just guessing isn't ideal 🙃 

I was thinking of maybe using their IDE integration to figure this out. Ideally though I'd like to be able to just run the tests from the command line or something.

PR #2718, opened 2023-11-20
feat: scope bound connection object

peterschutt:
### Pull Request Checklist

### Description

-

### Close Issue(s)

-


peterschutt:
I've realized that this approach won't work because it binds the request to the `receive()` and `send()` coros before they are passed through the middleware stack where they might later be wrapped.

PR #2725, opened 2023-11-20
refactor: use scope state for caching connection data

peterschutt:
### Pull Request Checklist

### Description

Instead of using private names in the top-level scope namespace, we instead store and retrieve our internal data using the `get_litestar_scope_state()` and `set_litestar_scope_state()` utilities.

This PR is a continuation of work started in #2540.

### Close Issue(s)

-


litestar/connection/base.py
```diff
@@ -115,7 +126,8 @@ def url(self) -> URL:
             A URL instance constructed from the request's scope.
         """
         if self._url is Empty:
-            self._url = self.scope["_url"] = URL.from_scope(self.scope)  # type: ignore[typeddict-unknown-key]
+            self._url = URL.from_scope(self.scope)
```
provinzkraut, line 129: Why don't we adapt the same thing we have in the `Headers.from_scope` method? It feels a bit nicer to me that all those methods kinda follow the same logic

peterschutt: Sorry, can you demonstrate what you mean?

provinzkraut: The `Headers.from_scope` classmethod contains the "cache in scope" functionality:

https://github.com/litestar-org/litestar/blob/a2e5b7854edce5b40d23bc3493d47f579bdcddef/litestar/datastructures/headers.py#L74-L88

Since we already have `URL.from_scope`, I think it would make sense that it works in the same way. Or we make `Headers.from_scope` not store itself in the scope; Either way, IMO it's a good idea if they both do the same thing :)

peterschutt: OK I follow and agree, we'll use the `from_scope()` methods to also do the caching of the values in scope where required :+1:

litestar/testing/client/base.py
```diff
@@ -83,10 +99,16 @@ def __init__(
...
+
         self._session_backend: BaseSessionBackend | None = None
         if session_config:
             self._session_backend = session_config._backend_class(config=session_config)
-        self.app = app
+
+        if not isinstance(app, Litestar):
+            app = _add_state_app(app)  # type: ignore[assignment]
+
+        self.app = cast("T", app)  # type: ignore[redundant-cast]  # pyright needs this
```
peterschutt, line 110: We depend on the `"state"` key existing in `scope`, and setting it is one of the first things that we do with every connection:

https://github.com/litestar-org/litestar/blob/a2e5b7854edce5b40d23bc3493d47f579bdcddef/litestar/app.py#L541

However, `httpx` doesn't inject `state` with its scope ([ref](https://github.com/encode/httpx/blob/87f39f12c9ecdd0ebcd8d017235698e72567e9c7/httpx/_transports/asgi.py#L98-L111)):

```py
    async def handle_async_request(
        self,
        request: Request,
    ) -> Response:
        assert isinstance(request.stream, AsyncByteStream)

        # ASGI scope.
        scope = {
            "type": "http",
            "asgi": {"version": "3.0"},
            "http_version": "1.1",
            "method": request.method,
            "headers": [(k.lower(), v) for (k, v) in request.headers.raw],
            "scheme": request.url.scheme,
            "path": request.url.path,
            "raw_path": request.url.raw_path,
            "query_string": request.url.query,
            "server": (request.url.host, request.url.port),
            "client": self.client,
            "root_path": self.root_path,
        }
```
... that scope is constructed totally within that function scope and there is no interface to add extra keys to it.

This is why I'm wrapping arbitrary non-litestar asgi-apps with another app that injects the "state" key into the scope.

litestar/_kwargs/extractors.py
```diff
@@ -145,11 +147,13 @@ def parse_connection_query_params(connection: ASGIConnection, kwargs_model: Kwar
     Returns:
         A dictionary of parsed values.
     """
-    parsed_query = connection.scope["_parsed_query"] = (  # type: ignore
+    parsed_query = cast(
```
peterschutt, line 150: this is a bit of a hot path, should we prefer using `# type: ignore` over `cast()` in these cases?

litestar/utils/scope.py
```diff
@@ -46,13 +46,6 @@ def get_serializer_from_scope(scope: Scope) -> Serializer:
 def get_litestar_scope_state(scope: Scope, key: str, default: Any = None, pop: bool = False) -> Any:
     """Get an internal value from connection scope state.
 
-    Note:
-        If called with a default value, this method behaves like to `dict.set_default()`, both setting the key in the
-        namespace to the default value, and returning it.
-
-        If called without a default value, the method behaves like `dict.get()`, returning ``None`` if the key does not
-        exist.
-
```
peterschutt, line 55: The behavior of the util was changed, so this is stale info.

tests/unit/test_connection/test_request.py
```diff
@@ -30,42 +31,47 @@ def _route_handler() -> None:
...
 
-async def test_request_empty_body_to_json(anyio_backend: str) -> None:
+@pytest.fixture(name="scope")
+def scope_fixture(create_scope: Callable[..., Scope]) -> Scope:
+    return create_scope(type="http", route_handler=_route_handler)
+
+
+async def test_request_empty_body_to_json(anyio_backend: str, scope: Scope) -> None:
     with patch.object(Request, "body", return_value=b""):
-        request_empty_payload: Request = Request(scope={"type": "http", "route_handler": _route_handler})  # type: ignore
```
peterschutt, line 35: Now that we aren't caching connection things in the top-level of `scope`, when we create connection instances, we need to ensure that the scope that is passed to the connection has the "state" key in it, like it would at runtime.

That means this change has had to cover a bit more surface area to accommodate this, but I think it is a better pattern overall to use the `create_scope` fixture in our tests anyway.

tests/unit/test_connection/test_request.py
```diff
@@ -354,26 +366,26 @@ async def receiver() -> dict:
...
+    @get("/", signature_namespace={"dict": Dict})
+    def handler(request: Request[Any, Any, Any]) -> dict[Any, Any]:
         request.state.test = 1
         assert request.state.test == 1
         return request.state.dict()  # type: ignore
 
     with create_test_client(handler) as client:
         response = client.get("/")
-        assert response.json() == {"test": 1}
+        assert response.json()["test"] == 1
```
peterschutt, line 384: This test changed because the state also has the `__litestar__` key in it.

litestar/connection/base.py
peterschutt: Thinking about this more, we'd have to add a `scope_cache_key` parameter to the `from_scope()` methods, b/c we cache URLs under both `url` and `base_url` keys.

litestar/_kwargs/extractors.py
peterschutt: changed it to the comment

litestar/connection/base.py
provinzkraut: > Thinking about this more, we'd have to add a `scope_cache_key` parameter to the `from_scope()` methods, b/c we cache URLs under both `url` and `base_url` keys.

Those would be the same though if constructed from the same scope, right?

peterschutt: In `base_url()` we unpack the connection scope into a new scope, modify the path, query etc and construct the url with `from_scope()`. Even though they are separate scopes, they share the same state, and now that we are caching in state, the first one to access would be the one that is cached and returned for either `url()` or `base_url()`. 

I'm AFK though, so I've not actually ran the scenario but I think that's right, isn't it?

provinzkraut: > I'm AFK though, so I've not actually ran the scenario but I think that's right, isn't it?

Same here but that sounds plausible!

litestar/connection/request.py
```diff
@@ -177,18 +185,23 @@ async def form(self) -> FormMultiDict:
...
+            form_values = parse_url_encoded_form_data(
                 await self.body(),
             )
-            return FormMultiDict(form_values)
-        return FormMultiDict()
+        else:
+            form_values = {}
+
+        if form_values:
+            self._form = form_values
```
JacobCoffee, line 201:
```suggestion
        if self._form := form_values:
```
Would this be okay?
Muh walrus!

JacobCoffee (approved):

litestar/connection/base.py
peterschutt: I'm going to leave this as is for now, so that this PR is only changing where we cache stuff within scope, which is its original intent.

Changing the behavior of either `URL.from_scope()` or `Headers.from_scope()` should be looked at, but probably best in its own PR as it will likely need some discussion around it.

litestar/testing/client/base.py
peterschutt: There's only one pattern of tests in our own test suite that this affects, that is the one where we construct a simple asgi app and pass it to `TestClient` and then use the scope to construct a connection instance, like this:

https://github.com/litestar-org/litestar/blob/8e4b50ceb79de292410d8d9600ed15000191ecb7/tests/unit/test_connection/test_request.py#L158-L167

So, while I'm not over the moon about this, the alternative is to do something like `scope.setdefault("state", {})` in `ASGIConnection.__init__()`, and I don't like having to change to library code just to support running tests.

PR #2734, opened 2023-11-22
refactor: change type of `Empty` and `EmptyType`

peterschutt:
### Pull Request Checklist

### Description

This PR changes the type of `Empty` and `EmptyType` in `litestar.utils.empty` to types that better support narrowing.

There is no change to the API of empty, and identity checks against `Empty` will still work as they always have.

Use of `EmptyType` as a type in annotations will continue to work like it always has.

Type checking of `Empty` will improve, as evidenced by the bug that this PR exposed in our use of `ParameterDefinition.default_factory`. It is typed `Callable | None` and in multiple places we've been passing `Empty` to it which was passing type checks because the type `Empty` _is_ callable. As a single enum variant, this type of error should no longer be possible.

Where downstream users are importing and using `Empty` as one would expect, i.e., `if thing is Empty: ...`, this PR should only improve their typing experience.

An issue that may arise is if `Empty` is being used in an `issubclass` check, e.g., `issubclass(thing, Empty)`. Under this PR, this would now be a runtime error. However, given that the previous implementation of `Empty` was marked as `@final`, it cannot have been subclassed in user-code, and so I think safe to assume that this scenario would be rare, if not non-existent.

### Close Issue(s)

-


JacobCoffee (approved):

litestar/contrib/pydantic/utils.py
```diff
@@ -1,3 +1,4 @@
+# mypy: strict-equality=False
```
peterschutt, line 1: this disables the mypy comparison overlap just for this module

litestar/contrib/pydantic/utils.py
```diff
@@ -52,13 +53,15 @@ def is_pydantic_model_class(
...
-    if pydantic_v2 is Empty:  # type: ignore[comparison-overlap] # pragma: no cover
-        return is_class_and_subclass(annotation, pydantic_v1.BaseModel)
+    if pydantic_v1 is not Empty:  # pragma: no cover
+        tests.append(is_class_and_subclass(annotation, pydantic_v1.BaseModel))
 
-    return is_class_and_subclass(annotation, (pydantic_v1.BaseModel, pydantic_v2.BaseModel))
+    if pydantic_v2 is not Empty:  # pragma: no cover
+        tests.append(is_class_and_subclass(annotation, pydantic_v2.BaseModel))
+
+    return any(tests)
```
peterschutt, line 64: there shouldn't be any functional change, just pyright was a lot happier with this pattern.

litestar/dto/msgspec_dto.py
```diff
@@ -40,7 +43,7 @@ def default_or_empty(value: Any) -> Any:
                     field_definition=field_definition,
                     dto_field=dto_field,
                     model_name=model_type.__name__,
-                    default_factory=default_or_empty(msgspec_field.default_factory),
+                    default_factory=default_or_none(msgspec_field.default_factory),
```
peterschutt, line 46: This was a bug.

The type of `default_factory` is `Callable | None`, so `Empty` was being allowed b/c  as a type it is a callable.

litestar/response/base.py
```diff
@@ -371,6 +372,9 @@ def render(self, content: Any, media_type: str, enc_hook: Serializer = default_s
         if isinstance(content, bytes):
             return content
 
+        if content is Empty:
+            raise RuntimeError("The `Empty` sentinel cannot be used as response content")
```
peterschutt, line 376: we have a test that checks that when `Empty` is returned from a handler it raises a 500, so this preserves that behavior. Otherwise, as an enum `Empty` can be serialized.

provinzkraut (approved):
That's some nice cleanup!

guacs (approved):

PR #2735, opened 2023-11-22
feat: set app env on startup

cofin:
### Pull Request Checklist

### Description

- This PR allows for registered CLI plugins to be visible in the `help` menus of the click commands.  Without it, you must set the `app` and `app-dir` setting on the command before the extra CLI commands are visible.

### Close Issue(s)

-

### DIFF
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -120,9 +120,7 @@ class LitestarEnv:
             import dotenv
 
             dotenv.load_dotenv()
-        app_path = app_path or getenv("LITESTAR_APP")
-        if app_path and getenv("LITESTAR_APP") is None:
-            os.environ["LITESTAR_APP"] = app_path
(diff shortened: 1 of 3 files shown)


litestar/cli/_utils.py
```diff
@@ -120,6 +121,8 @@ def from_env(cls, app_path: str | None, app_dir: Path | None = None) -> Litestar
 
             dotenv.load_dotenv()
         app_path = app_path or getenv("LITESTAR_APP")
+        if app_path and getenv("LITESTAR_APP") is None:
+            os.environ["LITESTAR_APP"] = app_path
```
provinzkraut, line 125:
```suggestion
        app_path = os.environ["LITESTAR_APP"] = app_path or getenv("LITESTAR_APP")
```

No need for the check really. The value either comes from the env, in which case we can override it with its current value without changing anything, or it doesn't come from the env, in which case we want to set it.

provinzkraut (approved):

PR #2736, opened 2023-11-22
fix: connection cache access

### LINKED ISSUE #2727: Bug: multiple connection objects can cause the app to hang
### Description

If multiple request objects are created before the first one calls one of the cached data access methods, e.g., `Request.json()` et. al, then the 2nd one to attempt to access the data will cause the request to block on `receive()` for a 2nd time, waiting for message that will never come.

The reason is that we only check whether the data has been cached upon instantiation of the connection objects, not when the connection object attempts to access the request data.

_No response_

### MCVE

```python
from typing import Dict

from litestar import Litestar, Request, post
from litestar.types.asgi_types import ASGIApp, Message, Receive, Scope, Send


def some_middleware(app: ASGIApp) -> ASGIApp:
    async def middleware(scope: Scope, receive: Receive, send: Send) -> None:
        request = Request(scope, receive, send)

        async def wrapped_send(message: Message) -> None:
            await request.json()
            # do something with data
            await send(message)

        await app(scope, receive, wrapped_send)

    return middleware


@post(middleware=[some_middleware])
async def root(request: Request) -> Dict[str, str]:
    return request.json()


app = Litestar(route_handlers=[root], debug=True)
```

### Steps to reproduce

```bash
1. run the above application
2. make any POST request to `/`
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

main

### Platform

- [X] Linux
- [X] Mac
- [X] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2727">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2727/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
### Pull Request Checklist

### Description

It is currently possible to make handling of an HTTP request hang indefinitely by first creating two connection objects, and then accessing the connection data on each one. E.g., if we await `request_1.json()`, we'll receive the data, and then if we await `request_2.json()` it will not identify that the data has been cached and will indefinitely await on `receive()` for a message that will never arrive.

This PR resolves this issue by moving the check for cached data from the connection's `__init__()` method, to within the data access methods.

### Close Issue(s)

Closes #2727

### DIFF
--- a/litestar/testing/client/base.py
+++ b/litestar/testing/client/base.py
@@ -51,7 +51,7 @@ def fake_asgi_connection(app: ASGIApp, cookies: dict[str, str]) -> ASGIConnectio
         "app": app,  # type: ignore[typeddict-item]
         "state": {},
         "path_params": {},
-        "route_handler": None,  # type: ignore[typeddict-item]
+        "route_handler": None,  # type ignore[typeddict-item]
         "asgi": {"version": "3.0", "spec_version": "2.1"},
         "auth": None,
         "session": None,
--- a/litestar/testing/client/base.py
+++ b/litestar/testing/client/base.py
@@ -51,7 +51,7 @@ def fake_asgi_connection(app: ASGIApp, cookies: dict[str, str]) -> ASGIConnectio
         "app": app,  # type: ignore[typeddict-item]
         "state": {},
         "path_params": {},
-        "route_handler": None,  # type ignore[typeddict-item]
+        "route_handler": None,  # type: ignore[typeddict-item]
         "asgi": {"version": "3.0", "spec_version": "2.1"},
         "auth": None,
         "session": None,
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import importlib
 import inspect
+import os
 import sys
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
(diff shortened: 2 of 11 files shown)


litestar/testing/client/base.py
```diff
@@ -51,7 +51,7 @@ def fake_asgi_connection(app: ASGIApp, cookies: dict[str, str]) -> ASGIConnectio
         "app": app,  # type: ignore[typeddict-item]
         "state": {},
         "path_params": {},
-        "route_handler": None,  # type: ignore[typeddict-item]
+        "route_handler": None,  # type ignore[typeddict-item]
```
provinzkraut, line 54:
```suggestion
        "route_handler": None,  # type: ignore[typeddict-item]
```

probably an oversight?

provinzkraut (approved):
Nice!

~~I assume you went with this approach over the "singletons with `__new__`" thing we talked about for the sake of simplicity?~~

Never mind, you explained it on the issue already :)

peterschutt:
> Nice!
> 
> ~I assume you went with this approach over the "singletons with `__new__`" thing we talked about for the sake of simplicity?~
> 
> Never mind, you explained it on the issue already :)

I still think we should look at that. I've got #2718 there which is something of a half way approach to the singleton solution, where it just creates a request object for us to use upfront, stashes it in scope and we can reuse it where ever we need it (e.g., in middle wares, DI etc.).

The singleton change feels a bit 3.0-ish to me.

peterschutt:
I'll round these tests out tomorrow.

provinzkraut:
> The singleton change feels a bit 3.0-ish to me.

That's reasonable!

PR #2740, opened 2023-11-23
fix: _get_normalized_schema_key regexes

provinzkraut:
Follow up to #2717, replacing the regex logic with a more solid one, avoiding backtracking issues.


provinzkraut:
@peterschutt I *think* this ought to do it for real this time

JacobCoffee (approved):

peterschutt:
> @peterschutt I *think* this should fix it for real this time

Thank you!

PR #2746, opened 2023-11-24
feat: stronger typing for scope state utils.

peterschutt:
### Pull Request Checklist

### Description

We use connection scope state to store a lot of different things for caching, auth, logging etc.

This PR improves the type safety of writing to, and retrieving from the scope state via the scope state utils and in doing so allows us to make another dozen or so improvements to our typing by removing casts and fixing existing type errors.

While this pattern of overloading the scope util methods for literal key arguments is quite verbose, the pattern is clear so I don't believe incremental changes to the data that we store in scope state will cause too much headache. 

To add a new key to store in scope state:

1. add a new key literal to `litestar.types.scopes`
2. add a new constant to `litestar.constants` typed as the literal key name
3. add appropriate overloads to the get/pop/set utilities.

Additionally, the PR removes the "pop" behavior from `get_litestar_scope_state()` in favor of adding a new utility `pop_litestar_scope_state()` so that both the get and pop utils share the same semantics as `dict.get()` and `dict.pop()`.

### Close Issue(s)

-

### DIFF
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -7,6 +7,8 @@ from typing import TYPE_CHECKING, Any, Iterable
 from litestar.constants import (
     HTTP_RESPONSE_BODY,
     HTTP_RESPONSE_START,
+    SCOPE_STATE_HTTP_RESPONSE_BODY_KEY,
+    SCOPE_STATE_HTTP_RESPONSE_START_KEY,
     SCOPE_STATE_RESPONSE_COMPRESSED,
 )
 from litestar.data_extractors import (
@@ -19,9 +21,10 @@ from litestar.enums import ScopeType
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.middleware.base import AbstractMiddleware, DefineMiddleware
 from litestar.serialization import encode_json
-from litestar.utils import (
+from litestar.utils.scope import (
     get_litestar_scope_state,
     get_serializer_from_scope,
+    pop_litestar_scope_state,
     set_litestar_scope_state,
 )
 
@@ -196,8 +199,8 @@ class LoggingMiddleware(AbstractMiddleware):
         serializer = get_serializer_from_scope(scope)
         extracted_data = self.response_extractor(
             messages=(
-                get_litestar_scope_state(scope, HTTP_RESPONSE_START, pop=True),
-                get_litestar_scope_state(scope, HTTP_RESPONSE_BODY, pop=True),
+                pop_litestar_scope_state(scope, SCOPE_STATE_HTTP_RESPONSE_START_KEY),
+                pop_litestar_scope_state(scope, SCOPE_STATE_HTTP_RESPONSE_BODY_KEY),
             ),
         )
         response_body_compressed = get_litestar_scope_state(scope, SCOPE_STATE_RESPONSE_COMPRESSED, default=False)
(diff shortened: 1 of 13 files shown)


litestar/middleware/logging.py
```diff
@@ -196,8 +199,8 @@ def extract_response_data(self, scope: Scope) -> dict[str, Any]:
         serializer = get_serializer_from_scope(scope)
         extracted_data = self.response_extractor(
             messages=(
-                get_litestar_scope_state(scope, HTTP_RESPONSE_START, pop=True),
-                get_litestar_scope_state(scope, HTTP_RESPONSE_BODY, pop=True),
+                pop_litestar_scope_state(scope, SCOPE_STATE_HTTP_RESPONSE_START_KEY),
+                pop_litestar_scope_state(scope, SCOPE_STATE_HTTP_RESPONSE_BODY_KEY),
```
peterschutt, line 203: This was a type error because the old `pop=True` behavior would automatically default to return `None` if the keys don't exist, but the response extractor is not typed to receive optional messages.

This new function behaves the same as `dict.pop()` so these cannot be `None` if this doesn't raise.

peterschutt:
@cofin I think this will cause some typing issues in AA where the session/engine etc are stored in state and we don't have the literal state keys/overloads defined for those in here. It should be a simple workaround to have AA create and use its own namespace within scope state (i.e., we store stuff in `scope["state"]["__litestar__"]`, AA can use `scope["state"]["__aa__"]` or whatever), and we'll just re-implement versions of the scope access utils for over there.

We'd need to do that and release it before this PR makes it into a Litestar release - I'll send through a PR for that when I get a chance over the next day or so.

provinzkraut (commented):
I like the idea, but what if instead of a module, why not make it a class? That would be a somewhat nicer interface IMO.

Something like:
```python
scope_state = LitestarScopeState(scope)

cookies = scope_state.cookies
scope_state.cookies = {}
del scope_state.cookies
```

peterschutt:
> but what if instead of a module

Do you mean instead of a dict? I.e., the current `scope["state"]["__litestar__"] = {}` would instead be `scope["state"]["__litestar__"] = LitestarScopeState()`?

If so, it would probably be better, yeh. Although to enforce the state namespace we'd still need an access function to retrieve the state from the scope, so the pattern would be `if get_scope_state(scope).url is Empty: ...` smth like that anyway.

provinzkraut:
> > but what if instead of a module
> 
> Do you mean instead of a dict? I.e., the current `scope["state"]["__litestar__"] = {}` would instead be `scope["state"]["__litestar__"] = LitestarScopeState()`?
> 
> If so, it would probably be better, yeh. Although to enforce the state namespace we'd still need an access function to retrieve the state from the scope, so the pattern would be `if get_scope_state(scope).url is Empty: ...` smth like that anyway.

Oh, that's also a neat idea. But not what I was thinking!

The snippet I showed was meant literally; You'd instantiate a `LitestarScopeState`, which then internally defines the attributes, e.g. as properties:

```python
class LitestarScopeState:
    def __init__(self, scope: Scope) -> None:
        self._scope = scope

    @property
    def accept(self) -> Accept | None:
        namespace = self._scope["state"].setdefault(constants.SCOPE_STATE_NAMESPACE, {})
        return namespace.get(constants.SCOPE_STATE_ACCEPT_KEY)

    @accept.setter
    def accept(self, value: Accept) -> None:
        self._scope["state"].setdefault(constants.SCOPE_STATE_NAMESPACE, {})[constants.SCOPE_STATE_NAMESPACE] = value
```

PR #2751, opened 2023-11-25
feat: ConnectionState object to replace litestar scope state.

peterschutt:
### Pull Request Checklist

### Description

We use connection scope state to store a lot of different things for caching, auth, logging etc.

This PR improves the type safety of interacting with our scope state by introducing a `ConnectionState` datastructure that is stored and cached in the connection scope.

### Close Issue(s)

-


peterschutt:
I think object approach is a big improvement @provinzkraut!

guacs:
I didn't look through it in detail, but this is really nice!

I had a doubt somewhat related to this. For a lot of this internal stuff we're using a `dataclass`, but since we're already using `msgspec` is there a reason we can't switch to using `Strcut` over dataclasses? If we switch to `Struct`, then we can reap the benefits as stated [here](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs).

litestar/connection/base.py
```diff
@@ -121,11 +116,10 @@ def url(self) -> URL:
             A URL instance constructed from the request's scope.
         """
         if self._url is Empty:
-            if url := get_litestar_scope_state(self.scope, SCOPE_STATE_URL_KEY):
-                self._url = cast("URL", url)
+            if (url := self._connection_state.url) is not Empty:
+                self._url = url
```
guacs, line 120: Not exactly related to this, but do we need this `self._url`? I don't think it's being used anywhere else (within the class at least).

guacs: The same for `_base_url`. Maybe we can remove them in another PR? I don't really see the point of keeping them in the connection if we're already maintaining it in the `ConnectionState`.

EDIT: There are a few more like `_cookies`, `_parsed_query` etc.

litestar/connection/request.py
guacs, line 1: I think the same comment I made above applies here as well. Am I missing some reason as to why we need to keep this data in both the `Request` and the `ConnectionState`.

litestar/constants.py
```diff
@@ -17,26 +18,8 @@
...
-SCOPE_STATE_CSRF_TOKEN_KEY: Final = "csrf_token"  # possible hardcoded password
-SCOPE_STATE_DEPENDENCY_CACHE: Final = "dependency_cache"
-SCOPE_STATE_DO_CACHE: Final = "do_cache"
-SCOPE_STATE_FORM_KEY: Final = "form"
-SCOPE_STATE_IS_CACHED: Final = "is_cached"
-SCOPE_STATE_JSON_KEY: Final = "json"
-SCOPE_STATE_MSGPACK_KEY: Final = "msgpack"
-SCOPE_STATE_PARSED_QUERY_KEY: Final = "parsed_query"
-SCOPE_STATE_RESPONSE_COMPRESSED: Final = "response_compressed"
-SCOPE_STATE_URL_KEY: Final = "url"
```
guacs, line 42: I know we have a comment that it's for internal stuff, but would this be considered public API since it's not under a private namespace?

litestar/datastructures/internal.py
guacs, line 1: Should we rename this to `_internal.py` to indicate this is private?

peterschutt:
> I didn't look through it in detail, but this is really nice!

Thanks!

> I had a doubt somewhat related to this. For a lot of this internal stuff we're using a `dataclass`, but since we're already using `msgspec` is there a reason we can't switch to using `Strcut` over dataclasses? If we switch to `Struct`, then we can reap the benefits as stated [here](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs).

I've thought about the same thing myself, but I think we should choose one or the other, and the migration (if there is to be one) should be done in a single effort.   Should we make an rfc?

litestar/utils/scope.py
guacs, line 1: Shouldn't the `get_litestar_scope_state` and other functions be changed to according to the new implementation? Also, should it be allowed to set arbitrary values in the `ConnectionState` to account for things like `advanced-alchemy` and `litestar-asyncpg` using it. 

While I think those plugins should be changed to not use the internal stuff of `litestar`, the reason I was asking is that some users might have created their own plugins that use these functions for handling the plugin state. Especially if they referred to any of these existing plugins that in order to create their own.

Also, these weren't technically placed in a private namespace. However, they are not documented either.

guacs:
> I've thought about the same thing myself, but I think we should choose one or the other, and the migration (if there is to be one) should be done in a single effort. Should we make an rfc?

Yeah, I think an RFC would be good. And I agree, we should stick to one all throughout in a single PR.

litestar/connection/base.py
peterschutt: Yeh every one of the cached attributes/methods are the same pattern. Before #2736 these attributes were set inside `__init__()` but you can check that PR and related issue for background on moving them.

There was a lot of micro-optimisation work done through these parts, so it could be an artefact of that. This reminds me I should make `ConnectionState` a `__slots__` class.

litestar/constants.py
peterschutt: Yeah - strictly you are right, but the scope state storage is meant to be implementation detail. I'd prefer to ask for forgiveness and put them back in if someone kicks up a stink about it.

If they are relying on these, the fact that the constants are gone will be the least of their worries if this PR goes through.

litestar/datastructures/internal.py
peterschutt: Yep that's probably a good idea!

litestar/utils/scope.py
peterschutt: I had removed them but advanced-alchemy depends on them so I had to add them back. Once I can PR into AA to remove that dependence and it is released we can remove these.

litestar/connection/base.py
guacs: > Yeh every one of the cached attributes/methods are the same pattern. Before #2736 these attributes were set inside `__init__()` but you can check that PR and related issue for background on moving them.
> 
> There was a lot of micro-optimisation work done through these parts, so it could be an artefact of that. This reminds me I should make `ConnectionState` a `__slots__` class.

Yeah I remember that PR. I think that was a good decision. However, now that we have the `ConnectionState`, do we need to maintain this pattern is my question. Instead, could we go for following pattern?

```python
@property
def attribute(self):
    if (attribute := self._connection_state.attribute) is not Empty:
        return attribute
        
    self._connection_state.attribute = attribute = ...
    
    return attribute    
```

litestar/utils/scope.py
peterschutt: Plugins should manage their own state, there's no need to support tighter coupling between plugins and litestar IMO. I'm going to fix advanced-alchemy and I can check any other plugins and PR for them too.

litestar/connection/base.py
peterschutt: The only thing I can think is that your pattern has an extra lookup for every access to get the data after it is cached: self -> connection state -> value. Current pattern only has to go self -> value. I'd be surprised if the difference was significant at all, but I haven't done to work to know if that is the case or not - so I'm siding with status quo.

litestar/utils/scope.py
peterschutt: Lmk any plugins that you are aware of that I should check - I'll make sure to do litestar-asyncpg.

peterschutt: Yeah, after thinking about it, you are right. I'll  deprecate these methods and make them work with the new internal state until 3.0.

litestar/connection/base.py
peterschutt: I'll mark this resolved b/c not strictly related to this PR, but let's look into it. Even if this does result in extra lookups on the read side, there is also an extra write to accomodate that. I'd be surprised if there was a significant difference between your suggestion and the current pattern.

litestar/constants.py
provinzkraut: We treat `utils` like a private namespace. It's always been like that. We recently already had a conversation about this when I removed some unused utils. How about we just make it explicitly private?

litestar/datastructures/internal.py
provinzkraut: Why not move the whole thing to `utils/state`? I feel like the `datastructures` namespace should be reserved for public interfaces

litestar/utils/empty.py
```diff
@@ -0,0 +1,60 @@
...
+def not_empty(value: ValueT | EmptyType, default: EmptyType) -> ValueT:
+    ...
+
+
+@overload
+def not_empty(value: ValueT | EmptyType, default: DefaultT) -> ValueT | DefaultT:
+    ...
+
+
+def not_empty(value: ValueT | EmptyType, default: DefaultT | EmptyType = Empty) -> ValueT | DefaultT:
```
provinzkraut, line 39: I find this function a bit confusing. It feels like it's doing too many things. A function named `not_empty`, I would expect to only ever return a `bool`, indicating if the value I passed in is empty.

Maybe this is a naming issue? Would something like `ensure_not_empty` make more sense?

litestar/utils/scope.py
provinzkraut: > Also, these weren't technically placed in a private namespace. However, they are not documented either.

This is the same conversation as above and that was had before. I really think we should make a decision here and then stick with it :eyes: 

I think I'll open up a PR for this and then everyone can chime in with their thoughts.

litestar/datastructures/internal.py
```diff
@@ -0,0 +1,59 @@
...
+    is_cached: bool | EmptyType = Empty
+    json: Any | EmptyType = Empty
+    log_context: dict[str, Any] = field(default_factory=dict)
+    msgpack: Any | EmptyType = Empty
+    parsed_query: tuple[tuple[str, str], ...] | EmptyType = Empty
+    response_compressed: bool | EmptyType = Empty
+    url: URL | EmptyType = Empty
+
+    @classmethod
+    def from_scope(cls, scope: Scope) -> Self:
```
provinzkraut, line 44: Just for convenience, how about adding a `connection_state` function that calls this?

```python
url = connection_state(scope).url
```

is a bit nicer on the eyes than


```python
url = ConnectionState.from_url(scope).url
```

litestar/datastructures/internal.py
```diff
@@ -0,0 +1,59 @@
...
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+    from litestar.datastructures import URL, Accept, Headers
+    from litestar.types.asgi_types import Scope
+
+
+@dataclass
+class ConnectionState:
```
provinzkraut, line 17: Why the change in naming here from "scope state" to "connection state"? To me "scope state" is a bit more intuitive because it describes more accurately what this is: It takes a `Scope` and stores state in it

provinzkraut (commented):
That's some really nice stuff @peterschutt! The dataclass approach is even better than what I had in mind :)

provinzkraut:
> I had a doubt somewhat related to this. For a lot of this internal stuff we're using a `dataclass`, but since we're already using `msgspec` is there a reason we can't switch to using `Strcut` over dataclasses? If we switch to `Struct`, then we can reap the benefits as stated [here](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs).

I agree with @peterschutt here that consistency is more important. Furthermore, this would also be a breaking change since users might subclass some of these dataclasses. Also, mostly use these for configuration, so performance isn't really an issue there (doesn't hurt though). I think, if anything, this would be a 3.0 thing.

litestar/utils/empty.py
peterschutt: `ensure_not_empty` just feels wordier than it needs to be.. to me, `is_not_empty` is name of the predicate/bool returning function. This receives a value, and promises that it won't return empty.. "not empty".

That's the entirety of my argument so if you still think I should change it, I will.

litestar/datastructures/internal.py
peterschutt: Why treat this different than `Headers.from_scope()`? Or'`URL.from_scope()`? We don't have functions to wrap the constructor methods of these.

litestar/datastructures/internal.py
peterschutt: I'm happy to name it anything

litestar/utils/empty.py
provinzkraut: If it was just a predicate, I'd agree with you, but since it optionally returns a value it feels off.

> is name of the predicate/bool returning function

That's how I would read it, but it doesn't *always* return a bool, sometimes it returns a different value.

How about `not_empty_or`? Less wordy but it doesn't sound as much like just a predicate :eyes:

litestar/datastructures/internal.py
provinzkraut: Fair point. It just looked so verbose to me. But I guess that's just because of the usual access pattern here, which isn't used with the other ones.

litestar/datastructures/internal.py
provinzkraut: That would be a terrible name :upside_down_face: 

Well, I'm +1 on `ScopeState`, but maybe other members have opinions on this as well?
@guacs perhaps? :)

litestar/constants.py
guacs: I think making it explicitly private is a good idea. If we can slowly do that for all private stuff (where we haven't already), then whenever we change stuff there, we don't have to worry about backwards compatibility.

litestar/connection/base.py
guacs: > I'll mark this resolved b/c not strictly related to this PR, but let's look into it. Even if this does result in extra lookups on the read side, there is also an extra write to accomodate that. I'd be surprised if there was a significant difference between your suggestion and the current pattern.

Agreed. I also don't think there would be a real performance degradation unless the user is accessing that value repeatedly in a loop or something.

litestar/datastructures/internal.py
guacs: I'm happy with either name as well. Though I disagree with you, @provinzkraut, that `anything` is a terrible name. It would be a fantastic name :P

guacs:
> > I had a doubt somewhat related to this. For a lot of this internal stuff we're using a `dataclass`, but since we're already using `msgspec` is there a reason we can't switch to using `Strcut` over dataclasses? If we switch to `Struct`, then we can reap the benefits as stated [here](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs).
> 
> I agree with @peterschutt here that consistency is more important. Furthermore, this would also be a breaking change since users might subclass some of these dataclasses. Also, mostly use these for configuration, so performance isn't really an issue there (doesn't hurt though). I think, if anything, this would be a 3.0 thing.

Aah you're right about it being breaking. I hadn't thought of that. I think it's something we could consider seriously for v3 though. It seems like a simple change with nice performance benefits as well.

litestar/datastructures/internal.py
provinzkraut: > Though I disagree with you, @provinzkraut, that `anything` is a terrible name. It would be a fantastic name :P

Reminds me of the `Thing` class in the old Reddit codebase.

Let's leave the naming decision up to @peterschutt then if no one has a strong opinion here :)

litestar/constants.py
peterschutt: This would certainly make it easier to make changes and to know where the line should be between needing to deprecate functionality and being able to remove it.

Of these names in this module, a good chunk of them haven't made it into a release yet - so I will remove them here, and deprecate the others.

litestar/datastructures/headers.py
```diff
@@ -83,10 +86,13 @@ def from_scope(cls, scope: "HeaderScope") -> "Headers":
         Raises:
             ValueError: If the message does not have a ``headers`` key
         """
-        if (headers := scope.get("_headers")) is None:
-            headers = scope["_headers"] = cls(scope["headers"])  # type: ignore[typeddict-unknown-key]
+        if "state" not in scope:
+            return cls(scope["headers"])
```
peterschutt, line 90: We have a test that passes a "http.response.start" message to `from_scope()` here:

https://github.com/litestar-org/litestar/blob/8b621ac13820840f2c48123b603340ec09ff1380/tests/unit/test_datastructures/test_headers.py#L77-L82

However, I cannot see anywhere in the library source where we actually do this - so I wonder if this is just some artifact that can be removed?

@provinzkraut, do you have any insight on this? The behavior that we test would have added a "_headers" key into the message in the old behavior which I'm not sure is what we want, and the new issue is that the asgi messages don't have a "state" key.

litestar/datastructures/internal.py
peterschutt: > Why not move the whole thing to `utils/state`? I feel like the `datastructures` namespace should be reserved for public interfaces

Yes, I'll do this.

litestar/datastructures/internal.py
peterschutt: I'll go with `ScopeState` as that is the existing terminology and otherwise it feels a bit like change for the sake of change.

litestar/utils/empty.py
peterschutt: I'm not using the raising behavior anywhere, so I'm just going to make it `value_or_default()`.

peterschutt:
> > > I had a doubt somewhat related to this. For a lot of this internal stuff we're using a `dataclass`, but since we're already using `msgspec` is there a reason we can't switch to using `Strcut` over dataclasses? If we switch to `Struct`, then we can reap the benefits as stated [here](https://jcristharif.com/msgspec/benchmarks.html#benchmark-structs).
> > 
> > 
> > I agree with @peterschutt here that consistency is more important. Furthermore, this would also be a breaking change since users might subclass some of these dataclasses. Also, mostly use these for configuration, so performance isn't really an issue there (doesn't hurt though). I think, if anything, this would be a 3.0 thing.
> 
> Aah you're right about it being breaking. I hadn't thought of that. I think it's something we could consider seriously for v3 though. It seems like a simple change with nice performance benefits as well.

I'm actually starting to come around on this. Given that this is a private object, and we instantiate it for every request it might be worth considering:

```
 (litestar-3.8) peter@pop-os:~/PycharmProjects/litestar$ python -m timeit -s "from litestar.utils.scope.state import ScopeState" "ScopeState()"
500000 loops, best of 5: 583 nsec per loop
(litestar-3.8) peter@pop-os:~/PycharmProjects/litestar$ python -m timeit -s "from litestar.utils.scope.state import ScopeStateStruct" "ScopeStateStruct()"
2000000 loops, best of 5: 107 nsec per loop
```
The equivalent struct is ~5x faster to instantiate than the current dataclass implementation. Other operations like setting and retrieving values from the object are all in the same ballpark.

provinzkraut:
> The equivalent struct is ~5x faster to instantiate than the current dataclass implementation.

True, but we’re talking about nanoseconds here, not even microseconds. So this is hardly relevant at all I’d say.

litestar/constants.py
```diff
@@ -1,12 +1,14 @@
...
-from typing import Final
+from typing import Any, Final
 
 from msgspec import UnsetType
 
 from litestar.enums import MediaType
 from litestar.types import Empty
+from litestar.utils.deprecation import warn_deprecation
 
+CONNECTION_STATE_KEY: Final = "_connection_state"
```
guacs, line 11: Maybe we can move these internal constants into a `_constants.py` or mark them as private with `_CONNECTION_STATE_KEY`?

litestar/middleware/csrf.py
```diff
@@ -111,17 +110,17 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
             exclude_opt_key=self.config.exclude_from_csrf_key,
             exclude_path_pattern=self.exclude,
         ):
-            token = csrf_cookie or generate_csrf_token(secret=self.config.secret)
-            set_litestar_scope_state(scope=scope, key=SCOPE_STATE_CSRF_TOKEN_KEY, value=token)
+            token = connection_state.csrf_token = csrf_cookie or generate_csrf_token(secret=self.config.secret)
             await self.app(scope, receive, self.create_send_wrapper(send=send, csrf_cookie=csrf_cookie, token=token))
         elif self._csrf_tokens_match(existing_csrf_token, csrf_cookie):
-            set_litestar_scope_state(scope=scope, key=SCOPE_STATE_CSRF_TOKEN_KEY, value=existing_csrf_token)
+            connection_state.csrf_token = existing_csrf_token or ""
```
guacs, line 123: Was this defaulting to an empty string there before as well?

guacs:
> > The equivalent struct is ~5x faster to instantiate than the current dataclass implementation.
> 
> True, but we’re talking about nanoseconds here, not even microseconds. So this is hardly relevant at all I’d say.

I agree. I think this is better left as something we do for v3.

litestar/datastructures/headers.py
provinzkraut: I do. We did the `_`-prefixed thing in the top level scope before we introduces the `__litestar__` scope key. Iirc, when that change was made, only those instances where we explicitly set state (e.g. stuff we did in `set_litestar_scope_state`) were migrated.

Actually, I think it makes *more* sense to store our scope state within a top level `__litestar__` key than within `state`, as `state` is an optional key used by the [lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html), to handle [lifespan state](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#lifespan-state), which has some extra functionality attached to it, if supported by the server, so we're kinda interfering with that; It's more meant for things like our `State`.

peterschutt: To clarify my question, was caching the headers in the `http.response.start` message a particular behavior we were testing for? We don't pass a `http.response.start`, or `websocket.accept` message to `Headers.from_scope()` in our source code anywhere, but we do it in this test.

> Actually, I think it makes more sense to store our scope state within a top level __litestar__ key than within state, as state is an optional key used by the [lifespan protocol](https://asgi.readthedocs.io/en/latest/specs/lifespan.html), to handle [lifespan state](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#lifespan-state), which has some extra functionality attached to it, if supported by the server, so we're kinda interfering with that; It's more meant for things like our State.

That makes sense but it isn't related to this PR, so we can do it in a follow up.

provinzkraut: > To clarify my question, was caching the headers in the `http.response.start` message 

Yeah, no idea either. Doesn’t make any sense really, probably just a bugged test.

> That makes sense but it isn't related to this PR, so we can do it in a follow up.

Yes, let’s do that once this one gets merged :)

litestar/constants.py
peterschutt: We can move it to `utils.scope.state` - I agree, best to keep private things out of here.

litestar/middleware/csrf.py
peterschutt: No, but here's the current `main` implementation for reference:

https://github.com/litestar-org/litestar/blob/7414f7fd7d4782223502895e6a23b77ed635cd2d/litestar/middleware/csrf.py#L87-L127

Look at line 105, we use `dict.get()` to set the value of `existing_csrf_token` so it can be `None` if the header doesn't exist.

At line 123, that block is guarded by `self._csrf_tokens_match()` which will return `False` if it is `None`, so actually `existing_csrf_token` cannot be falsy in this block, its just that its value is not narrowed appropriately.

So, I should probably use a `cast()` or `type: ignore` instead of the `or` here, but even better would be to appropriately narrow the types, which would probably be as easy as using `request.cookies.get(..., "")` and `request.headers.get(..., "")` on lines 104 and 105 respectively, and re-type the `_csrf_tokens_match()` method to only accept strings.

peterschutt: changed to `type: ignore` and opened https://github.com/litestar-org/litestar/issues/2770

peterschutt:
> > > The equivalent struct is ~5x faster to instantiate than the current dataclass implementation.
> > 
> > 
> > True, but we’re talking about nanoseconds here, not even microseconds. So this is hardly relevant at all I’d say.
> 
> I agree. I think this is better left as something we do for v3.

OK reverted.

litestar/datastructures/headers.py
peterschutt: This gives me pause on whether `ScopeState` is the right name for the caching object then. If we are conflating its usage, then surely we are conflating terms also.

guacs (approved):
Great work, @peterschutt!

provinzkraut (approved):

Issue #2739, opened 2023-11-23
Bug: Validation Exception for required headers during CORS pre-flight request


abdulhaq-e:
### Description

CORS pre-flight "OPTIONS" request doesn't include the headers that are attached to the actual request. This causes a validation exception when the litestar handler has a required header.

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Litestar Version

latest

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh Litestar dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2739">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2739/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2739/pledge.svg">
</picture>
</a>

abdulhaq-e:
here is one possible solution that worked for me. Changing the function at https://github.com/litestar-org/litestar/blob/main/litestar/_kwargs/extractors.py#L56 to be:
```python
def _create_param_mappings(
    expected_params: set[ParameterDefinition], connection: ASGIConnection
) -> ParamMappings:
    alias_and_key_tuples = []
    alias_defaults = {}
    alias_to_params: dict[str, ParameterDefinition] = {}
    for param in expected_params:
        alias = param.field_alias
        if param.param_type == ParamType.HEADER:
            alias = alias.lower()

        alias_and_key_tuples.append((alias, param.field_name))

[...]

PR #2755, opened 2023-11-25
fix: skip headers validation for OPTIONS requests

abdulhaq-e:
### Pull Request Checklist

- [X] New code has 100% test coverage

### Description

- The behaviour doesn't happen if the required header is defined on the handler rather than the upper layers.
- Browsers sending the pre-flight CORS `OPTIONS` request by excluding the original headers and including specific ones seems to be part of a [standard](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request).
- The proposed solution skips validation of required headers when the request method is `OPTIONS`.
- One drawback of the proposed solution: developers who have an explicit handler for `OPTIONS` will lose the validation features. If we see this as important, then perhaps the solution can be changed to only skip validation for the [automatically created](https://github.com/litestar-org/litestar/blob/main/litestar/routes/http.py#L235) `OPTIONS` handler.

### Close Issue(s)

#2739 

-

### DIFF
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -53,7 +53,7 @@ class ParamMappings(NamedTuple):
     alias_to_param: dict[str, ParameterDefinition]
 
 
-def _create_param_mappings(expected_params: set[ParameterDefinition], connection: ASGIConnection) -> ParamMappings:
+def _create_param_mappings(expected_params: set[ParameterDefinition]) -> ParamMappings:
     alias_and_key_tuples = []
     alias_defaults = {}
     alias_to_params: dict[str, ParameterDefinition] = {}
@@ -66,8 +66,6 @@ def _create_param_mappings(expected_params: set[ParameterDefinition], connection
 
         if not (param.is_required or param.default is Ellipsis):
             alias_defaults[alias] = param.default
-        elif connection.scope.get("method") == "OPTIONS" and param.param_type == ParamType.HEADER and param.is_required:
-            alias_defaults[alias] = None
 
         alias_to_params[alias] = param
 
@@ -96,8 +94,9 @@ def create_connection_value_extractor(
         An extractor function.
     """
 
+    alias_and_key_tuples, alias_defaults, alias_to_params = _create_param_mappings(expected_params)
+
     def extractor(values: dict[str, Any], connection: ASGIConnection) -> None:
-        alias_and_key_tuples, alias_defaults, alias_to_params = _create_param_mappings(expected_params, connection)
         data = parser(connection, kwargs_model) if parser else getattr(connection, connection_key, {})
 
         try:
@@ -116,8 +115,7 @@ def create_connection_value_extractor(
 
 @lru_cache(1024)
 def create_query_default_dict(
-    parsed_query: tuple[tuple[str, str], ...],
-    sequence_query_parameter_names: tuple[str, ...],
+    parsed_query: tuple[tuple[str, str], ...], sequence_query_parameter_names: tuple[str, ...]
 ) -> defaultdict[str, list[str] | str]:
     """Transform a list of tuples into a default dict. Ensures non-list values are not wrapped in a list.
 
@@ -317,9 +315,7 @@ async def msgpack_extractor(connection: Request[Any, Any, Any]) -> Any:
 
 
 def create_multipart_extractor(
-    field_definition: FieldDefinition,
-    is_data_optional: bool,
-    data_dto: type[AbstractDTO] | None,
+    field_definition: FieldDefinition, is_data_optional: bool, data_dto: type[AbstractDTO] | None
 ) -> Callable[[ASGIConnection[Any, Any, Any, Any]], Coroutine[Any, Any, Any]]:
     """Create a multipart form-data extractor.
 
@@ -369,10 +365,7 @@ def create_multipart_extractor(
 
         return data_dto(connection).decode_builtins(form_values) if data_dto else form_values
 
-    return cast(
-        "Callable[[ASGIConnection[Any, Any, Any, Any]], Coroutine[Any, Any, Any]]",
-        extract_multipart,
-    )
+    return cast("Callable[[ASGIConnection[Any, Any, Any, Any]], Coroutine[Any, Any, Any]]", extract_multipart)
 
 
 def create_url_encoded_data_extractor(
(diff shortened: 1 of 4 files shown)


provinzkraut (changes requested):
The solution to skip the validation entirely for `OPTIONS` requests is way too broad. This can lead to undesired behaviour and it's also not transparent to the user. 

I'm also not exactly sure what header validation you're referencing here, since the generated `OPTIONS` request handlers do not perform regular validation on the headers. It would be helpful if you provide a concrete MCVE for the original issue.

abdulhaq-e:
> The solution to skip the validation entirely for `OPTIONS` requests is way too broad. This can lead to undesired behaviour and it's also not transparent to the user.
> 
> I'm also not exactly sure what header validation you're referencing here, since the generated `OPTIONS` request handlers do not perform regular validation on the headers. It would be helpful if you provide a concrete MCVE for the original issue.

The test I posted fails without the changes. This simulates a browser sending an OPTIONS request to the handler without providing the headers required by the handler.

> since the generated `OPTIONS` request handlers do not perform regular validation on the headers

The validation is performed if the header was defined on the layers above the handlers just like in my example where it is defined on the router.

abdulhaq-e:
This is the line that runs the header validation and causes the failure:

https://github.com/litestar-org/litestar/blob/main/litestar/routes/http.py#L173

provinzkraut:
But that's not specific to `OPTIONS` requests. This will also not generally validate headers. It will only validate headers if you have headers somewhere in the requested dependencies or defined anywhere in the layer above the handler. The `OPTIONS` handler itself does *not* define any headers that will be validated there.

If you *do* specify a header somewhere above the handler, it would actually be a bug if that was just skipped implicitly.

Can you please provide a full MCVE of this?

abdulhaq-e:
> The solution to skip the validation entirely for OPTIONS requests is way too broad. This can lead to undesired behaviour and it's also not transparent to the user.

A possible alternative would be creating a special `BaseRouteHandler` for the auto-generated OPTIONS handler. This will then resolve the layered parameters in a special way such that there are no required headers which will produce a suitable `KwargsModel` [here](https://github.com/litestar-org/litestar/blob/main/litestar/routes/base.py#L126C79-L126C90)

abdulhaq-e:
> If you _do_ specify a header somewhere above the handler, it would actually be a bug if that was just skipped implicitly.
> 


> Can you please provide a full MCVE of this?

The test is literally an MCVE, if the app in the test is served and a GET request is sent from a browser with a different origin, the request will fail on the `OPTIONS` request. Why? Headers don't get sent by the browser in the pre-flight request.

provinzkraut:
> The goal is to skip validation even for headers defined on upper layers when the auto-generated OPTIONS request is being executed because it's a request fully controlled by the browser.

This would disabled behaviour a user specifically enabled. At the very least, this is a major breaking change, and introducing it without any warning could be a significant security risk.


> The test is an MCVE, if the app in the test is served and a GET request is sent from a browser with a different origin, the request will fail on the `OPTIONS` request. Why? Headers don't get sent by the browser in the pre-flight request.

As far as I can see, the app you're using in the example works as I would expect it to. If you want to validate headers conditionally, you'll have to implement that manually, using hooks, guards, dependency injection or middlewares. 

> Headers don't get sent by the browser in the pre-flight request.

But your change doesn't just apply to pre-flight requests. It applies to *all* `OPTIONS` requests. And there's nothing preventing anyone from sending headers with an `OPTIONS` request or using them for things like authentication. In fact, having `OPTIONS` request authenticated isn't uncommon, and this change would make a core functionality that's used in handling authentication in Litestar not available for `OPTIONS` requests anymore at all.

abdulhaq-e:
> As far as I can see, the app you're using in the example works as I would expect it to. If you want to validate headers conditionally, you'll have to implement that manually, using hooks, guards, dependency injection or middlewares.
> 

The header is validated correctly when sending a `GET` request to that endpoint. However, as I mentioned, it fails on the CORS pre-flight request. The other options to make this work without modifying the behaviour of litestar:

1. Make the header optional.
2. Give a default value for the header.
3. Write a custom OPTIONS endpoint for the handler.

As you can see, they're not optimal, it's all to do with the auto-generated OPTIONS handler.

> > Headers don't get sent by the browser in the pre-flight request.
> 
> But your change doesn't just apply to pre-flight requests. It applies to _all_ `OPTIONS` requests. And there's nothing preventing anyone from sending headers with an `OPTIONS` request or using them for things like authentication. In fact, having `OPTIONS` request authenticated isn't uncommon, and this change would make a core functionality that's used in handling authentication in Litestar not available for `OPTIONS` requests anymore at all.

I agree that this solution isn't correct. I will update it now with another implementation that just targets the auto-generated OPTIONS handler.

provinzkraut:
> As you can see, they're not optimal, it's all to do with the auto-generated OPTIONS handler.

They're not optimal in this very specific scenario, but that's not really the issue. You could maybe add a fix for this, but it would be impossible to do so without breaking anything. The thing is, it's simply impossible to have an autogenerated handler that can comply with *all* possible requirements you could throw at it. And it shouldn't have to. I'd say that in this case, you either shouldn't use a generated handler, or change how you define your parameter. As I mentioned before, there are ways of doing this that are more dynamic, which seems to be what you actually want.

abdulhaq-e:
> but it would be impossible to do so without breaking anything. The thing is, it's simply impossible to have an autogenerated handler that can comply with all possible requirements you could throw at it.

That's my point :D, it's an auto-generated handler so who would be relying on validation/dependencies in such a handler, so let's just remove them, I don't see how it can be a breaking change. 

I've made an update to the implementation.

abdulhaq-e:
I'm really sorry about the formatting changes. My editor/config is making them (`black`) but then `pre-commit` is not reversing these as I expected it would.

provinzkraut:
> That's my point :D, it's an auto-generated handler so who would be relying on validation/dependencies in such a handler, so let's just remove them, I don't see how it can be a breaking change.

We can't just assume a feature isn't being used and therefore declare it not breaking.

And you would rely on this because those features being disabled specifically for this one handler type is not expected at all.

Your new changes break things even more. If someone does authentication within a dependency (fairly common), they would have *all* `OPTIONS` routes exposed. You personally might not feel like it's that big of a deal, but we can't just changes how these things work on a whim. People rely on this, and we don't know how they might have set up their applications.

abdulhaq-e:
I've created a thread on discord, lets perhaps continue there.

PR #2756, opened 2023-11-25
feat: Improve ASGI extension interface

provinzkraut:
To improve the handling of [ASGI extensions](https://asgi.readthedocs.io/en/latest/extensions.html), this PR:

- Introduces an `ASGIExtension` enum, holding the names of ASGI extensions
- Adds a `_server_extensions` attribute to `ASGIConnection`
- Adds a `supports_server_push` attribute to `Request`, allowing to check if it is supported before attempting to push. Additionally, a warning is raised when push is not supported. Optionally, this method can now be configured to raise an exception instead of a warning; This behaviour should be made the default in 3.0


JacobCoffee (approved):
very nice

cofin (approved):

Issue #2721, opened 2023-11-20
Bug: Renaming in DTOs with nested fields will rename all occurrences.


simon-lund:
### Description

In the example below, we have two models: 'Chat' and 'Message', each with an 'id' field. We also define a ResponseDTO for the Chat model, using DTOConfig to rename its 'id' field to 'chat_id'. 

The intention is to only rename the 'id' of the Chat model. However, this configuration unexpectedly renames the 'id' field in the nested Message objects as well. As a result, the response output is 

`{"chat_id":"1","messages":[{"chat_id":"1","created_at":"1","user_id":"1"}]}`, 

where both Chat and Message 'id' fields are affected, contrary to the intended behavior.

### MCVE

```python
from typing_extensions import List
from litestar import get, Litestar
from pydantic import BaseModel
from litestar.dto import DTOConfig
from litestar.contrib.pydantic import PydanticDTO

class Message(BaseModel):
    id: str
    created_at: str
    user_id: str

class Chat(BaseModel):
    id: str
    messages: List[Message]

class ResponseDTO(PydanticDTO[Chat]):
    config = DTOConfig(rename_fields={"id": "chat_id"})

@get(return_dto=ResponseDTO)
async def get_chat() -> Chat:
    return Chat(id="1", messages=[Message(id="1", created_at="1", user_id="1")])

app = Litestar([get_chat])
```

### Steps to reproduce

```bash
1. Start Litestar
2. Make API call `curl localhost:8000`
3. Observe :bug:
```

### Litestar Version

2.2.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh Litestar dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2721">
<picture>
[...]

PR #2764, opened 2023-11-26
fix(dto|#2721): nested field renaming

provinzkraut:
Fix an issue with nested field renaming in DTOs that would lead to all fields with a certain name to be renamed in a nested structure.

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -4,7 +4,18 @@ back again, to bytes.
 from __future__ import annotations
 
 from dataclasses import replace
-from typing import TYPE_CHECKING, AbstractSet, Any, Callable, ClassVar, Collection, Final, Mapping, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
+    Any,
+    ClassVar,
+    Collection,
+    Final,
+    Mapping,
+    Protocol,
+    Union,
+    cast,
+)
 
 from msgspec import UNSET, Struct, UnsetType, convert, defstruct, field
 from typing_extensions import get_origin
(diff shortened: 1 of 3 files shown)


cofin (approved):
Thanks for this one!  LGTM

simon-lund:
thank you 😊

litestar/dto/_backend.py
```diff
@@ -365,15 +396,24 @@ def _create_transfer_type(
         field_definition: FieldDefinition,
```
provinzkraut, line 396: @peterschutt 

There's quite a few of those attributes that we keep passing around. Do you think it might be worth introducing some sort of context object here or refactoring this to some kind of builder pattern where these are instance attributes?

peterschutt: I have no doubt there are refactorings that can be made through these parts and I could see how a context object could simplify these signatures and perhaps could also house the logic that modifies the `exclude` and `include` for the level of nesting.

From taking a quick look I can see other things I'd like to tidy up.

Raising `RecursionError` to break out for max nested depth is an interesting choice. Catching a `RecursionError` and continuing makes my spidey sense tingle:

https://github.com/litestar-org/litestar/blob/7414f7fd7d4782223502895e6a23b77ed635cd2d/litestar/dto/_backend.py#L130-L131

And this:

https://github.com/litestar-org/litestar/blob/7414f7fd7d4782223502895e6a23b77ed635cd2d/litestar/dto/_backend.py#L384-L386

Note we call `parse_model()` with `field_definition.annotation`, and then in 2nd line of that method we go right on ahead and create a new one:

https://github.com/litestar-org/litestar/blob/7414f7fd7d4782223502895e6a23b77ed635cd2d/litestar/dto/_backend.py#L110

On that last one, it should be a rule that we never unwrap the annotations from their parsed type object until we absolutely have to.

Issue #2694, opened 2023-11-17
refactor: mypy disallow_any_generics


peterschutt:
This issue was identified in #2689 when we realized that mypy is inferring `Litestar.template_engine` as `Any` due to `AppConfig.template_config` being annotated as `TemplateConfig` (without generic arg) instead of `TemplateConfig[TemplateEngineProtocol]`.

`AppConfig.template_config` is annotated here: https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/litestar/config/app.py#L199

And assigned to the app instance here: https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/litestar/app.py#L412

I really think we should work toward changing https://github.com/litestar-org/litestar/blob/38f3b196d10ac1c3d680d7707d97523679d5f8d3/pyproject.toml#L225 to `true` so these types of errors cannot occur.

_Originally posted by @peterschutt in https://github.com/litestar-org/litestar/pull/2689#discussion_r1396785637_
            

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh Litestar dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2694">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2694/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2694/pledge.svg">
</picture>
</a>

PR #2769, opened 2023-11-26
refactor: Add types to some tests

marcuslimdw:
### Pull Request Checklist

### Description

Add type parameters to some test cases where they were not used; partially fixes https://github.com/litestar-org/litestar/issues/2694.

### DIFF
--- a/litestar/types/builtin_types.py
+++ b/litestar/types/builtin_types.py
@@ -2,10 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Dict, Type, Union
 
-from typing_extensions import (
-    Never,
-    _TypedDictMeta,  # type: ignore
-)
+from typing_extensions import Never, _TypedDictMeta  # type: ignore
 
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
--- a/litestar/types/builtin_types.py
+++ b/litestar/types/builtin_types.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Dict, Type, Union
 
-from typing_extensions import Never, _TypedDictMeta  # type: ignore
+from typing_extensions import Never, _TypedDictMeta  # type: ignore[attr-defined]
 
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
--- a/litestar/types/builtin_types.py
+++ b/litestar/types/builtin_types.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Dict, Type, Union
+from typing import TYPE_CHECKING, Type, Union
 
-from typing_extensions import Never, _TypedDictMeta  # type: ignore[attr-defined]
+from typing_extensions import _TypedDictMeta  # type: ignore[attr-defined]
 
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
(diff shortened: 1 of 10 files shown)


litestar/types/builtin_types.py
```diff
@@ -23,3 +24,4 @@
 
 UnionTypes = {UnionType, Union}
 TypedDictClass: TypeAlias = Type[_TypedDictMeta]
+EmptyDict: TypeAlias = Dict[Never, Never]
```
peterschutt, line 27: Its probably not worth adding a new library type just to type those tests - I'd be happy enough with `dict[Any, Any]` for them.

tests/unit/test_openapi/test_responses.py
```diff
@@ -42,6 +41,8 @@
 from tests.models import DataclassPerson, DataclassPersonFactory
 from tests.unit.test_openapi.utils import PetException
 
+T = TypeVar("T")
+
```
peterschutt, line 45: Any particular reason to prefer local definition over the import?

peterschutt (dismissed):
Thanks!

Do you have specific reasoning for targeting these particular tests? There are still hundreds of cases of these that remain both in the tests and in the library, so I feel like some sort of organized approach to completing this task might be in order.

tests/unit/test_openapi/test_responses.py
marcuslimdw: Two things: `T` is not declared in `litestar.response.base.__all__`, and it seems (a little) weird to me to import an unbounded type variable from a sibling module. I'm not sure what the project convention is though, so happy to revert this if you think it's inappropriate

marcuslimdw:
What I did was set `disallow_any_generics` to `true` locally, and then look for stuff that I was reasonably confident could be changed without requiring in-depth knowledge.

I suggest:

* Generating a list of packages which have `disallow-any-generics` violations
* On the root issue, creating a checklist (not sure what you call those lists of checkboxes) with these packages
* Letting people claim them in comments on the root issue and create PRs to fix those packages, reporting back on the issue when they're done

litestar/types/builtin_types.py
peterschutt: Well, we don't have a documented standard on adding new types to the library, however if we did this use case would fall short of it IMO.

marcuslimdw: Got it, will inline it

tests/unit/test_openapi/test_responses.py
peterschutt: Both of those are valid reasons.

We do often import names that aren't declared in `__all__` into tests b/c we often don't include names in there that we don't want in the public interface yet still want to test.

On point two, if I was writing this module today I wouldn't have imported `T` either. Again, no strict convention, but I would say that the common practice is to not import these.

As a counter argument, minor changes do have a cost in that someone has to review them. So as long as you balance that fact against the changes you propose then I'm happy enough.

JacobCoffee:
Can this be merged (after updates) or should it be closed?

tests/unit/test_openapi/test_responses.py
```diff
@@ -44,6 +43,9 @@
...
+
```
provinzkraut, line 48:
```suggestion
```

provinzkraut (approved):

provinzkraut (approved):
I think this PR has set a new record for longest time between opened and merged :)

PR #2800, opened 2023-11-29
fix: setting props on schema using camelCase names

peterschutt:
### Pull Request Checklist

### Description

We have been setting properties on the schema object using camelCase names when they are declared with snake_case names.

We've been using this mapping:

https://github.com/litestar-org/litestar/blob/9f6ddf43928c6deb033f18fcec5aed21772192ff/litestar/_openapi/schema_generation/schema.py#L79-L98

...for converting values from the `KwargDefinition` object to the `Schema` object, where the keys are the names on `KwargDefinition` and the values are supposed to be the names on the `Schema` object. Which takes place here:

https://github.com/litestar-org/litestar/blob/9f6ddf43928c6deb033f18fcec5aed21772192ff/litestar/_openapi/schema_generation/schema.py#L640-L647

Note that some of the values in that mapping are pointing to camelCase names, but we define the names as snake_case on the schema type.

I'm pretty sure that this means that any of those camelCase schema values woudn't have been getting transferred into the schema correctly, rather they'd have been getting assigned to the schema instance, but not to any of the dataclass field names.

### Close Issue(s)

-

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -77,22 +77,22 @@ if TYPE_CHECKING:
     from litestar.plugins import OpenAPISchemaPluginProtocol
 
 KWARG_DEFINITION_ATTRIBUTE_TO_OPENAPI_PROPERTY_MAP: dict[str, str] = {
-    "content_encoding": "contentEncoding",
+    "content_encoding": "content_encoding",
     "default": "default",
     "description": "description",
     "enum": "enum",
     "examples": "examples",
-    "external_docs": "externalDocs",
+    "external_docs": "external_docs",
     "format": "format",
     "ge": "minimum",
-    "gt": "exclusiveMinimum",
+    "gt": "exclusive_minimum",
     "le": "maximum",
-    "lt": "exclusiveMaximum",
-    "max_items": "maxItems",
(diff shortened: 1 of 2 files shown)


tests/unit/test_openapi/test_schema.py
```diff
@@ -319,8 +319,8 @@ class MyDataclass:
     assert schema.properties["constrained_int"].exclusive_maximum == 10  # type: ignore
     assert schema.properties["constrained_float"].minimum == 1  # type: ignore
     assert schema.properties["constrained_float"].maximum == 10  # type: ignore
-    assert date.fromtimestamp(schema.properties["constrained_date"].exclusive_minimum) == historical_date  # type: ignore
-    assert date.fromtimestamp(schema.properties["constrained_date"].exclusive_maximum) == today  # type: ignore
+    assert schema.properties["constrained_date"].exclusive_minimum == historical_date  # type: ignore
+    assert schema.properties["constrained_date"].exclusive_maximum == today  # type: ignore
```
peterschutt, line 323: I don't get how these were passing before, and what we should be testing here.. why did they need to be converted to date before, and why don't they now. How were they even getting set to the right attribute before this PR?

JacobCoffee (approved):
makes sense

Issue #2804, opened 2023-11-29
bug: 2.4 changes schema key names


ssenior45:
### Summary

There was a change in this PR (https://github.com/litestar-org/litestar/commit/219355c0e322fab41c921c56ca97d356b7ef868b#diff-6cdcc2ab6bdeda3d72588f3b65db0c437f7ba9cad43892829e75c4d701f9e658R659-R661) that updated the way schema keys are generated for the OpenAPI schema.

Where we originally had something like `UserAccount` as the model name, we know have `app_domain_accounts_UserAccount` or something similar.

### Basic Example

MCVE

```python
import json

from litestar import Litestar, get
from litestar.testing import TestClient
from pydantic import BaseModel


class SampleOne(BaseModel):
    property_one: str


class SampleTwo(BaseModel):
    __module__ = ""
    property_one: str


@get("/one")
async def test_handler_one() -> SampleOne:
    return SampleOne(property_one="test")


@get("/two")
async def test_handler_two() -> SampleTwo:
    return SampleTwo(property_one="test")


app = Litestar(route_handlers=[test_handler_one, test_handler_two], debug=True)

with TestClient(app=app) as client:
    response = client.get("/schema/openapi.json")
    print(json.dumps(response.json()["components"]["schemas"], indent=2))
```

This outputs:

```
{
  "main_SampleOne": {
    "properties": {
      "property_one": {
        "type": "string"
      }
    },
    "type": "object",
    "required": [
      "property_one"
    ],
    "title": "SampleOne"
  },
  "SampleTwo": {
    "properties": {
      "property_one": {
        "type": "string"
      }
    },
    "type": "object",
    "required": [
      "property_one"
    ],
    "title": "SampleTwo"
  }
}INFO - 2023-11-29 18:43:59,780 - httpx - _client - HTTP Request: GET http://testserver.local/schema/openapi.json "HTTP/1.1 200 OK"
```

Notice that for the `SampleOne` model the generated schema key name is `main_SampleOne`.

After reviewing the PR as far as I can understand it you are now making use of class metadata to get name of the model, whereas previously you were using the schema title field.

The only workaround I've been able to come up with can be seen in the the `SampleTwo` model where I set `__module__ = ""`. This allows me to get only the class name as the schema key as can be seen in the output.

### Drawbacks and Impact

This is a breaking change for us as we use an OpenAPI typescript generator from the generated `openapi.json`
and the named types are now different than what it expects.

### Unresolved questions

[...]

PR #2805, opened 2023-11-30
refactor: refactor openapi schema generation

peterschutt:
### Pull Request Checklist

### Description

We share a lot of state between each of the different schema element builder functions, so this PR takes a different approach by sharing that state via factory types.

It's a bit of a big PR, apologies.

To summarize, the existing functional approach relied on passing a lot of the same shared state between them to generate the schemas, so this PR takes a different approach through implementing factory types for each of the schema elements that we generate.

The pattern is one factory instance per element generated, i.e., there should be a single OpenAPIFactory instance for the application, one `PathItemFactory` instance per `HTTPRoute`, one `Parameter` factory per handler, and so on.

### Close Issue(s)

precursor work for #2804.


litestar/di.py
```diff
@@ -28,12 +29,14 @@ class Provide:
...
         "has_sync_generator_dependency",
         "has_async_generator_dependency",
+        "parsed_signature",
         "signature_model",
         "sync_to_thread",
         "use_cache",
         "value",
     )
 
+    parsed_signature: ParsedSignature
```
peterschutt, line 39: I've added the parsed sig to the DI interface, as we need to create it to make the signature model, and we've been accessing the parameters from the parsed signature via `signature_model._fields`, when we may as well access them directly from the source.

litestar/handlers/base.py
```diff
@@ -353,13 +353,16 @@ def resolve_dependencies(self) -> dict[str, Provide]:
                         dependencies=self._resolved_dependencies, key=key, provider=provider
                     )
 
+                    if not getattr(provider, "parsed_signature", None):
+                        provider.parsed_signature = ParsedSignature.from_fn(
+                            unwrap_partial(provider.dependency), self.resolve_signature_namespace()
+                        )
```
peterschutt, line 359: as above.

tests/unit/test_cli/test_env_resolution.py
```diff
@@ -75,6 +78,8 @@ def test_env_from_env_autodiscover_from_files(
     dotted_path = _path_to_dotted_path(tmp_file_path.relative_to(Path.cwd()))
 
     assert isinstance(env.app, Litestar)
+
+    print("parent directory content: %s", list(tmp_file_path.parent.iterdir()))  # noqa: T201
```
peterschutt, line 82: I'm trying to see what is in these tmp directories when this test fails.. I don't know any better way to do this unfortunately.

peterschutt:
This PR also addresses one of the remaining cognitive complexity issues left in the code base: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=litestar-org_litestar&open=AYeibLEUxZTizGEJ182r

litestar/_openapi/datastructures.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def add_operation_id(self, operation_id: str) -> None:
+        """Add an operation ID to the context.
+
+        Args:
+            operation_id: Operation ID to add.
+        """
+        if operation_id in self.operation_ids:
+            raise ImproperlyConfiguredException(
+                f"operation_ids must be unique, "
```
provinzkraut, line 46:
```suggestion
                "operation_ids must be unique, "
```

litestar/app.py
```diff
@@ -315,6 +314,9 @@ def __init__(
                 connections.
             experimental_features: An iterable of experimental features to enable
         """
+
+        self._openapi_factory = OpenAPIFactory(self)
```
provinzkraut, line 318: Is this actually needed here? If all it does is hold a reference to the app at this point, why not just initialise the attribute as `None` here (or don't set it at all and just declare it as a class attribute) and instantiate it on the first access to `openapi_schema`? That would allow to skip the "check if initialised" logic inside the factory

litestar/_openapi/parameters.py
```diff
@@ -75,181 +74,146 @@ def list(self) -> list[Parameter]:
...
-    route_handler: BaseRouteHandler,
-    path_parameters: tuple[PathParameterDefinition, ...],
-    schema_creator: SchemaCreator,
-) -> list[Parameter]:
-    """Create and return parameters for a handler.
-
-    If the provided field is not a dependency, a normal parameter is created and returned as a list, otherwise
-    `create_parameter_for_handler()` is called to generate parameters for the dependency.
-    """
+class ParameterFactory:
```
provinzkraut, line 77: I'm not a huge fan of this "class with a single method that's called immediately" pattern. From what I've seen so far, it's only ever created and then has its `create_parameters_for_handler` method called directly; Everything else is only ever called internally.

I guess the intent here was to encapsulate some state, but maybe this class should be private to the module then, and wrapped in a public `create_parameters_for_handler` function then, which sets up the class and calls its `create_parameters_for_handler` method?

litestar/_openapi/path_item.py
```diff
@@ -1,141 +1,124 @@
...
-                security=security,
-            )
-            operation_ids.append(operation_id)
-            setattr(path_item, http_method.lower(), operation)
 
-    return path_item, operation_ids
+__all__ = ("PathItemFactory",)
+
+
+class PathItemFactory:
```
provinzkraut, line 22: Same comment here as for the `ParameterFactory`

litestar/_openapi/request_body.py
```diff
@@ -2,33 +2,65 @@
...
 
 
-def create_request_body(
-    route_handler: BaseRouteHandler, field_definition: FieldDefinition, schema_creator: SchemaCreator
-) -> RequestBody | None:
-    """Create a RequestBody model for the given RouteHandler or return None."""
-    media_type: RequestEncodingType | str = RequestEncodingType.JSON
-    if isinstance(field_definition.kwarg_definition, BodyKwarg) and field_definition.kwarg_definition.media_type:
-        media_type = field_definition.kwarg_definition.media_type
+class RequestBodyFactory:
```
provinzkraut, line 20: Same as above, but even stronger IMO, since it only *has* one method. I guess consistency is nice here, but if the classes were wrapped in a function call as suggested above, that wouldn't be an issue either :)

provinzkraut (approved):
This is some good stuff @peterschutt!

One thing though is that I'm not too keen on the public interface for the factory types. I get why you introduced them, and their inner workings definitely make sense as a class, but the way these are used is like functions, so a functional interface would make more sense I think. Having them be classes that are not shared outside their module and then a public, functional interface that creates the parameter would be the best of both worlds I think

litestar/_openapi/request_body.py
peterschutt: Yeah, consistency with the other types was my reasoning, will prob remove after adding the functional interface as you suggest.

litestar/app.py
peterschutt: Glad you brought it up.

> That would allow to skip the "check if initialised" logic inside the factory

This logic then gets moved outside of the factory, as we call `self._openapi_factory.add_route()` for each http route, which we'd have to guard with `if self._openapi_factory is not None:`.

I'm not really happy with the pattern here, but I don't think that delaying instantiation of the factory type really fixes much.

What I don't like about the way that we have this openapi stuff setup, is that there is special cased openapi schema logic dispersed across different methods of the `Litestar` object - it makes it a bit hard to reason about how it all works.

- In `Litestar.init()` we register the openapi controller 
- In `Litestar.register()` we only call `update_openapi_schema()` if `Litestar._openapi_schema` is not `None`.
- In `Litestar._openapi_schema` we raise if `Litestar.openapi_config` is `None` and if `Litestar._openapi_schema` is `None`, then call `Litestar.openapi_config.to_openapI_schema()` to create the schema, and then call `Litestar.update_openapi_schema()` to update it.
- In `Litestar.update_openapi_schema()` we raise an error if `Litestar._openapi_schema` is `None`, and also if `Litestar._openapi_schema.paths` is `None`. We also raise if there is no `Litestar.openapi_config`. I think at this point, these checks are to do with type narrowing within that method body, and not any functional requirement, but that's non-obvious when reading the code. We then go on to add route handlers to the schema, but only if the path for the route doesn't already exist (thinking out loud, but what if a handler for an existing path is added after the schema has been built?).
- After `Litestar.openapi_schema` has been accessed once, we then call `Litestar.update_openapi_schema()` every time that `Litestar.register()` is called.

It all just feels a bit unsettled to me.

Maybe a plugin would b a better solution, but we don't have a plugin that gets access to routes and handlers as they are added to the application.

So, say we expand plugins protocols to include a `ReceiveRoutePlugin` protocol that has a `receive_route()` method.

We could then build a plugin whose job it is to handle all of these gory details, and it woudn't necessarily need to even be a public plugin, we could just instantiate it and add it to the application's plugin registry if we receive an `OpenAPIConfig` object on initialization. The plugin could:

1. add the openapi controllers to the app route handlers
2. receive the routes as they are added to the application, determine the handlers on the route that should be included in the schema and keep a reference to them and their route.
3. handle deferred building of the openapi schema when it is first accessed. 
4. cache the openapi schema in app state instead of on the app object as currently occurs, removing a special case that doesn't really need to exist.
5. create dependency providers that provide the schema and its serialized versions for the openapi controllers with caching. This could then also handle invalidation of those cached objects better as FWICT that doesn't happen with the openapi controller as it stands, i.e., the json version of the schema is generated and cached and served, but if a route is dynamically added, that schema wouldn't be updated.
6. provide a callable in its namespace that retrieves the openapi_schema from an app instance which we could use wherever we currently access the schema from the app (in `app.openapi_schema` and by the CLI schema method).

We'd end up caching things on app state, using plugins to dynamically add the openapi controller, and have the schemas available via DI, which are all standard ways of doing these things in Litestar, and at the same time we'd remove a lot of special casing that occurs on the app object for the OpenAPI stuff.

Another use for expanding our protocols to receive routes would be the serialization plugins. If these things defined a `receive_route()` method (or maybe even a `receive_route_handler()` method) they'd be able to look at the handlers as they are added to the application, and inspect if they have had a DTO defined or not, and set a dto type for the handler at that point if they support the annotation types. This would save us from having to iterate over the serialization plugins twice for every handler to determine if a plugin should provide the DTO.

peterschutt: Have a look at the latest commit to this branch to get a feel for it: https://github.com/litestar-org/litestar/pull/2805/commits/bbc1d1e059be74fec70ed96ddc0447e923ccdbba

If we were to add the new protocol, that would have to be a separate PR, but I think it tidies things up quite a bit.

litestar/_openapi/plugin.py
```diff
@@ -0,0 +1,102 @@
...
+from litestar.status_codes import HTTP_404_NOT_FOUND
+
+if TYPE_CHECKING:
+    from litestar.app import Litestar
+    from litestar.config.app import AppConfig
+    from litestar.routes import BaseRoute
+
+
+class OpenAPIPlugin(InitPluginProtocol, ReceiveRouteProtocol):
+    def __init__(self, app: Litestar) -> None:
```
peterschutt, line 27: Add slots

litestar/_openapi/plugin.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+        return self._openapi_schema_json
+
+    def provide_openapi_schema_yaml(self) -> bytes:
+        if not self._openapi_schema_yaml:
+            schema_json = self.provide_openapi_schema_json()
+            self._openapi_schema_yaml = dump_yaml(
+                decode_json(schema_json),
+                default_flow_style=False,
+            ).encode("utf-8")
```
peterschutt, line 65: Would rather use `msgspec.yaml.encode` but things broke - will look into it

litestar/_openapi/plugin.py
```diff
@@ -0,0 +1,102 @@
...
+        if not self._openapi_config:
+            raise ImproperlyConfiguredException("OpenAPIConfig not initialized")
+        return self._openapi_config
+
+    def receive_route(self, route: BaseRoute) -> None:
+        if not isinstance(route, HTTPRoute):
+            return
+
+        # Force recompute the schema if a new route is added
+        self._openapi_schema = self._openapi_schema_yaml = self._openapi_schema_json = None
```
peterschutt, line 99: Move this into the `if ...` below

litestar/app.py
```diff
@@ -167,6 +165,7 @@ class Litestar(Router):
         "websocket_class",
         "pdb_on_exception",
         "experimental_features",
+        "_openapi_factory",
```
peterschutt, line 168: Remove

litestar/app.py
```diff
@@ -375,6 +375,9 @@ def __init__(
             websocket_class=websocket_class,
             experimental_features=list(experimental_features or []),
         )
+
+        config.plugins.append(OpenAPIPlugin(self))
```
peterschutt, line 379: Has to be last in case any other init plugins set an openapi config

litestar/di.py
```diff
@@ -28,12 +29,14 @@ class Provide:
         "has_sync_callable",
         "has_sync_generator_dependency",
         "has_async_generator_dependency",
+        "parsed_signature",
```
peterschutt, line 32: `parsed_fn_signature` to be consistent with route handlers

litestar/openapi/controller.py
```diff
@@ -1,31 +1,79 @@
 from __future__ import annotations
```
peterschutt, line 1: Given that this entire controller is documented, I might have gone too hard on the changes. However, using DI and guards etc has really tidied this up (who'd have thought?!😀), and I really just wanted to see what it looked like. Maybe we could have an alternate controller that is registered by the plugin based on feature flag and just leave this one as is.

litestar/openapi/controller.py
```diff
@@ -1,31 +1,79 @@
...
+    """Exception raised when an OpenAPI endpoint is not found."""
+
+    def __init__(self, body: bytes, media_type: MediaType) -> None:
+        self.body = body
+        self.media_type = media_type
+        super().__init__()
+
+
+def openapi_guard(request: ASGIConnection, route_handler: BaseRouteHandler) -> None:
+    if not request.app.openapi_config:  # pragma: no cover
```
peterschutt, line 36: Should access through the plug-in registry

litestar/openapi/controller.py
```diff
@@ -1,31 +1,79 @@
...
+    owner = route_handler.owner
+    if not owner or not isinstance(owner, OpenAPIController):
+        raise RuntimeError("OpenAPIController should be the owner of this route handler")
+
+    asgi_root_path = set(filter(None, request.scope.get("root_path", "").split("/")))
+    full_request_path = set(filter(None, request.url.path.split("/")))
+    request_path = full_request_path.difference(asgi_root_path)
+    root_path = set(filter(None, owner.path.split("/")))
+
+    config = request.app.openapi_config
```
peterschutt, line 51: Same

litestar/plugins/base.py
```diff
@@ -69,6 +71,16 @@ def on_app_init(self, app_config: AppConfig) -> AppConfig:
         return app_config  # pragma: no cover
 
 
+@runtime_checkable
+class ReceiveRouteProtocol(Protocol):
```
peterschutt, line 75: `ReceiveRoutePluginProtocol`

litestar/plugins/base.py
```diff
@@ -219,6 +232,7 @@ def __init__(self, plugins: list[PluginProtocol]) -> None:
         self._plugins = frozenset(plugins)
         self.init = tuple(p for p in plugins if isinstance(p, InitPluginProtocol))
         self.openapi = tuple(p for p in plugins if isinstance(p, OpenAPISchemaPluginProtocol))
+        self.receive_route = tuple(p for p in plugins if isinstance(p, ReceiveRouteProtocol))
```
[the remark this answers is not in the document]
peterschutt: > litestar/plugins/base.py:235: error: Generator has incompatible item type "Union[litestar.plugins.base.<subclass of "SerializationPluginProtocol" and "ReceiveRouteProtocol">1, litestar.plugins.base.<subclass of "InitPluginProtocol" and "ReceiveRouteProtocol">1, litestar.plugins.base.<subclass of "OpenAPISchemaPluginProtocol" and "ReceiveRouteProtocol">1, litestar.plugins.base.<subclass of "OpenAPISchemaPlugin" and "ReceiveRouteProtocol">1, litestar.plugins.base.<subclass of "CLIPluginProtocol" and "ReceiveRouteProtocol">1, litestar.plugins.base.<subclass of "CLIPlugin" and "ReceiveRouteProtocol">1]"; expected "Union[litestar.plugins.base.<subclass of "SerializationPluginProtocol" and "ReceiveRouteProtocol">, litestar.plugins.base.<subclass of "InitPluginProtocol" and "ReceiveRouteProtocol">, litestar.plugins.base.<subclass of "OpenAPISchemaPluginProtocol" and "ReceiveRouteProtocol">, litestar.plugins.base.<subclass of "OpenAPISchemaPlugin" and "ReceiveRouteProtocol">, litestar.plugins.base.<subclass of "CLIPluginProtocol" and "ReceiveRouteProtocol">, litestar.plugins.base.<subclass of "CLIPlugin" and "ReceiveRouteProtocol">]" 

:fearful:

litestar/_openapi/path_item.py
```diff
@@ -1,141 +1,137 @@
...
+
+        return route_handler.operation_class(
+            operation_id=operation_id,
+            tags=route_handler.resolve_tags() or None,
+            summary=route_handler.summary or SEPARATORS_CLEANUP_PATTERN.sub("", route_handler.handler_name.title()),
+            description=self.get_description_for_handler(route_handler),
+            deprecated=route_handler.deprecated,
+            responses=responses,
+            request_body=request_body,
+            parameters=parameters or None,  # type: ignore[arg-type]
```
guacs, line 83: Is this type ignore needed because `mypy` doesn't like `list[Parameter]` for `list[Parameter | Reference]`?

litestar/_openapi/path_item.py
```diff
@@ -1,141 +1,137 @@
...
+        elif callable(route_handler.operation_id):
+            operation_id = route_handler.operation_id(route_handler, http_method, self.route.path_components)
+        else:
+            operation_id = self.context.openapi_config.operation_id_creator(
+                route_handler, http_method, self.route.path_components
+            )
+        self.context.add_operation_id(operation_id)
+        return operation_id
+
+    def get_description_for_handler(self, route_handler: HTTPRouteHandler) -> str | None:
```
guacs, line 110:
```suggestion
    def create_description_for_handler(self, route_handler: HTTPRouteHandler) -> str | None:
```

Just to maintain consistency with the other methods that are named `create_*`.

litestar/_openapi/plugin.py
guacs: Is there a different way to get the `app` instance into the plugin? If we ever want to expose this to the users, they won't be able to instantiate the plugin without the app. Maybe we can consider adding the `Litestar` instance to the `AppConfig`?

litestar/_openapi/path_item.py
peterschutt: I haven't looked into it actually, we have the same in main: https://github.com/litestar-org/litestar/blob/7979771d941ab2cc3be6249f59d31dc48ccc6039/litestar/_openapi/path_item.py#L135

litestar/_openapi/plugin.py
peterschutt: I've just refactored the dependency providers to access the type encoders from `request.app` instead of the `self.app` reference, so the only thing it needs the app for at the moment is to get the openapi schema plugins from the registry.

However, the thing standing most in the way of us making the plugin public interface is that order is important at the moment. B/c any of the init plugins, or on_app_init hooks could configure an `OpenAPIConfig` object, we require that this plugin runs after all of the other init plugins/hooks.

However, If `AppConfig.plugins` was an instance of `PluginRegistry` instead of a list, and we removed `openapi_config` from `Litestar` and `AppConfig` so that using the OpenAPI plugin was more like using the other types of plugins, where the plugin receives the config object directly, then I don't think order would matter. Irrespective of when the OpenAPI plugin ran, it could grab a reference to the plugin registry from the app config object and know that it would have access to all of the openapi schema plugins when it needed them.

I wonder if we could do this in a mostly non-breaking way by making the plugin registry subclass from list?

litestar/plugins/base.py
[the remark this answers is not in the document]
peterschutt: i had to add the type to the `PluginProtocol` union :bulb:

litestar/_openapi/plugin.py
peterschutt: I'm happy to leave it as is - at least for now - and if we find a reason to want to make it public eventually then we can revisit.

peterschutt: > Also, making the registry mutable would be a breaking change IMO.

I didn't think the registry was a documented thing - its not documented on the `Litestar` class, nor in the `plugins` package (at least for API docs).

provinzkraut: > I didn't think the registry was a documented thing 

You're right! Well.. It should be! 
Nevertheless, it's exposed via a non-private attribute on the `Litestar` object. I think we should consider that part of our public API.

guacs (approved):
Fantastic work @peterschutt! 🎉

litestar/handlers/http_handlers/base.py
```diff
@@ -406,6 +410,40 @@ def resolve_include_in_schema(self) -> bool:
...
+
+        Returns:
+            list[str]: The resolved tags property.
+        """
+        if self._resolved_tags is Empty:
+            tag_set = set()
+            for layer in self.ownership_layers:
+                for tag in layer.tags or []:
+                    tag_set.add(tag)
+            self._resolved_tags = sorted(tag_set)
```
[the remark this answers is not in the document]
peterschutt: The logic was moved directly from here:

https://github.com/litestar-org/litestar/blob/5e4a9d182d99e5bc9fb9265e868e33c783538d35/litestar/_openapi/path_item.py#L44-L62

tests/unit/test_contrib/test_repository.py
```diff
@@ -16,7 +16,7 @@
 
 
 def test_app_repository_signature_namespace() -> None:
-    app = Litestar([], on_app_init=[handlers.on_app_init])
+    app = Litestar([], on_app_init=[handlers.on_app_init], openapi_config=None)
```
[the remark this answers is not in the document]
peterschutt: Well, yeh, but this was before I stripped out the work on the new openapi controller from this PR. That work added the `OpenaAPI` object to the signature namespace which would cause this test to fail as the test assumes that nothing else is adding anything to the namespace.

It can be removed.

tests/unit/test_openapi/test_request_body.py
```diff
@@ -18,16 +23,37 @@ class FormData:
     image: UploadFile
 
 
-def test_create_request_body(person_controller: Type[Controller]) -> None:
+ResponseBodyFactory = Callable[[BaseRouteHandler, FieldDefinition], RequestBody]
```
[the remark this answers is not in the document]
peterschutt: yep, ta.

tests/unit/test_openapi/test_responses.py
```diff
@@ -42,20 +42,37 @@
 from tests.models import DataclassPerson, DataclassPersonFactory
 from tests.unit.test_openapi.utils import PetException
 
+CreateFactoryFixture: TypeAlias = "Callable[..., ResponseFactory]"
```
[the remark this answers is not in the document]
peterschutt: The 2nd bool isn't required, it has a default and so this change makes the type checks fail when it is called without the 2nd arg. We could get around this with a protocol, but let's not over think it.

peterschutt:
Thanks heaps for reviews @guacs & @provinzkraut!

PR #2834, opened 2023-12-03
chore: Update PR template

provinzkraut:
Update the PR template, removing the checks and sections, since they enforced by our CI.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,25 +5,7 @@ By submitting this pull request, you agree to:
 - follow the [PSFs's Code of Conduct](https://www.python.org/psf/conduct/)
 -->
 
-### Pull Request Checklist
-
-- [ ] New code has 100% test coverage
-- [ ] (If applicable) The prose documentation has been updated to reflect the changes introduced by this PR
-- [ ] (If applicable) The reference documentation has been updated to reflect the changes introduced by this PR
-- [ ] Pre-Commit Checks were ran and passed
-- [ ] Tests were ran and passed
-
-### Description
-<!--
-Please describe your pull request for new release changelog purposes
--->
-
--
-
-### Close Issue(s)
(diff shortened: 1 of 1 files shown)


provinzkraut:
> I may have misunderstood when you had mentioned this earlier. I thought only the `Pull Request Checklist` was being removed. I think keeping the `Description` and `Close Issue(s)` would still be beneficial.

The thing about the "closes issues" is that it doesn't work. If you put in a list of issues there, it won't be automatically linked by GH. That only works if you write out "fixes <pr number>" or "closes <pr number>", so it doesn't feel all that useful. I left it as a comment on the template to remind people of this, as I felt it was a good tradeoff.

JacobCoffee:
@guacs I think so too.

>  That only works if you write out "fixes " or "closes "

Im not sure what this means? This has always worked for me but GitHub does tell you to include a closing verb.

JacobCoffee (changes requested):
Can we add these back

peterschutt (approved):
I'll throw in my 2c on the templates, while we are here.

Locally, we fix an issue or implement a feature, write a nice commit message, push it up, and GH will use that commit message to populate the PR. 

My experience with the current template is that for every PR I need to cut the auto-populated description from the top of the template and paste it into the description section, and then cut the `closes: #1234` line from the bottom of the description and paste it into the appropriate section below that. Its not much, but it is friction.

I like that this change will mean I'd be able to leave the auto-populated message alone. I'm fine with keeping the checklist, but would prefer it be at the bottom of the PR template, not the top. We could also just include an item in the checklist reminding contributors to include the `closes: #1234` or `for: #2345` directives without it needing to be its own section.

provinzkraut:
I'm in the same boat as @peterschutt really; For my personal workflow it's more of a hindrance, and since most of the PRs here are made by members, I think this is at least worth a try.

If we find that a lot of PRs coming from external contributors are missing things we can always go back :)

PR #2847, opened 2023-12-06
feat(openapi): add ``scalar`` schema documentation

### LINKED ISSUE #2846: Enhancement: Add `Scalar` API documentation
### Summary

- https://scalar.com/swagger-editor
image

### Basic Example

- Swagger
- ReDoc
- RapiDoc
- Stoplight Elements
- **Scalar**

### Drawbacks and Impact

This will need to be refactored into a plugin once we do that for the rest of them.

### Unresolved questions

_No response_

JacobCoffee:
- Add scalar api docs

Closes #2846

### DIFF
--- a/tests/unit/test_openapi/test_controller.py
+++ b/tests/unit/test_openapi/test_controller.py
@@ -12,163 +12,142 @@ from litestar.testing import create_test_client
 root_paths: List[str] = ["", "/part1", "/part1/part2"]
 
 
-def test_default_redoc_cdn_urls(person_controller: Type[Controller], pet_controller: Type[Controller]) -> None:
-    with create_test_client([person_controller, pet_controller]) as client:
-        response = client.get("/schema/redoc")
-        default_redoc_version = "next"
-        default_redoc_js_bundle = (
-            f"https://cdn.jsdelivr.net/npm/redoc@{default_redoc_version}/bundles/redoc.standalone.js"
-        )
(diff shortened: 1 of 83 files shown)


tests/unit/test_openapi/test_controller.py
JacobCoffee, line 1: I did a bunch of squishing here because there were a lot of repetition.. okay to revert if we dont like it

marclave:
> image
> im not sure why this is, but i think its causing the examples to fail loading
> 
> image Maybe @marclave has an idea 😄

ah! this is a bug when the server url isn't set properly.. we should better handle this edge case.

I'll pull and checkout this branch and see if I can fix it up

JacobCoffee:
awesome - thanks!
Usually we would get this merged in pretty quick but we have had a planned OpenAPI refactor for a little bit so it may be a bit before this goes in. I just saw it on Reddit and got excited 😄

marclave:
@JacobCoffee got a fix incoming :) 

https://github.com/scalar/scalar/pull/644

marclave:
@JacobCoffee new version released with the fix!✨

provinzkraut:
@JacobCoffee The pyright failures come from `develop`. We'll need to investigate and fix it there

JacobCoffee:
Closing this in favor of #2906 and #2893

PR #2855, opened 2023-12-08
test: xdist group cli autodiscovery tests

peterschutt:
This test is flaky and best guess is it has to do with executing across multiple cores. This marker ensures they will run on the same node.

See https://discord.com/channels/919193495116337154/1151927578370768958/threads/1179281714653184122


guacs (approved):

JacobCoffee (approved):

tests/examples/test_contrib/test_sqlalchemy/plugins/test_tutorial_example_apps.py
provinzkraut, line 1: These were unrelated and popped up because due to the `xdist_group` added in the CLI tests, they were now scheduled to run in a different order than before, which revealed a previously hidden issue with their test setup

PR #2868, opened 2023-12-10
refactor: use OpenAPISchemaPlugin for all supported complex types

peterschutt:
Instead of having separate methods on the `SchemaCreator` object for types like structs, dataclasses and typeddicts, this PR implements their schema generators as `OpenAPISchemaPlugin` types and adds them to the application's plugins on instantiation.

### DIFF
--- a/litestar/_openapi/schema_generation/plugins/__init__.py
+++ b/litestar/_openapi/schema_generation/plugins/__init__.py
@@ -1,0 +1,16 @@
+from .dataclass import DataclassSchemaPlugin
+from .pagination import PaginationSchemaPlugin
+from .struct import StructSchemaPlugin
+from .typed_dict import TypedDictSchemaPlugin
+
+__all__ = ("openapi_schema_plugins",)
+
+# NOTE: The Pagination type plugin has to come before the Dataclass plugin since the Pagination
+# classes are dataclasses, but we want to handle them differently from how dataclasses are normally
+# handled.
+openapi_schema_plugins = [
+    PaginationSchemaPlugin(),
+    StructSchemaPlugin(),
+    DataclassSchemaPlugin(),
+    TypedDictSchemaPlugin(),
+]
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from collections import deque
 from copy import copy
-from dataclasses import MISSING, fields
 from datetime import date, datetime, time, timedelta
 from decimal import Decimal
 from enum import Enum, EnumMeta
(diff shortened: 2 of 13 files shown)


litestar/_openapi/schema_generation/plugins/__init__.py
```diff
@@ -0,0 +1,16 @@
+from .dataclass import DataclassSchemaPlugin
+from .pagination import PaginationSchemaPlugin
+from .struct import StructSchemaPlugin
+from .typed_dict import TypedDictSchemaPlugin
+
+__all__ = ("openapi_schema_plugins",)
+
+# NOTE: The Pagination type plugin has to come before the Dataclass plugin since the Pagination
```
provinzkraut, line 8: There's a few cases where we rely on implicit ordering (same with middlewares). I think it might be worth it to add an explicit API for this, e.g. with some sort of priority configuration. Wdyt?

litestar/_openapi/schema_generation/schema.py
```diff
@@ -298,8 +290,14 @@ def not_generating_examples(self) -> SchemaCreator:
         return type(self)(generate_examples=False, plugins=self.plugins, prefer_alias=False)
 
     def get_plugin_for(self, field_definition: FieldDefinition) -> OpenAPISchemaPluginProtocol | None:
+        def plugin_supports_field(plugin: OpenAPISchemaPluginProtocol, field: FieldDefinition) -> bool:
```
provinzkraut, line 293: Doesn't seem like this has to be an inline definition, so maybe this should just be a static method?

provinzkraut (approved):
Love the work you've been doing on the OpenAPI stuff recently!

litestar/_openapi/schema_generation/plugins/__init__.py
peterschutt: Yeh I think that makes sense - this would be via the plugin registry I assume?

litestar/_openapi/schema_generation/schema.py
peterschutt: I added it only to simplify the generator expression - I don't mind splitting it out though :+1:

guacs:
@peterschutt I know this is a bit late, but I just had a chance to go through this and this is really fantastic! It makes things so much cleaner and declutters the `SchemaCreator` quite nicely!

peterschutt:
> @peterschutt I know this is a bit late, but I just had a chance to go through this and this is really fantastic! It makes things so much cleaner and declutters the `SchemaCreator` quite nicely!

Thanks @guacs - NGL, always get a bit nervous when a comment pops up on an already-merged-PR! This was nice to hear though, cheers:)

PR #2869, opened 2023-12-10
fix: recursive schema generation

### LINKED ISSUE #2429: Bug: OpenAPI schema generation fails with recursive models
### Description

When using recursive models such as the example below, Litestar gets a recursion error when generating an OpenAPI schema. Litestar should support these schemas either by having an implicit recursion limit when generating the docs or through some other means.

_No response_

### MCVE

```python

class A(BaseModel):
    a: A | None = None
    b: B | None = None

class B(BaseModel):
    a: A | None = None
    b: B | None = None

@get("/")
async def test() -> A:
    return A()

app = Litestar([test])
```

For comparison, here is a FastAPI version which works properly:
```py
from fastapi import FastAPI

class A(BaseModel):
    a: A | None = None
    b: B | None = None

class B(BaseModel):
    a: A | None = None
    b: B | None = None

app = FastAPI()

@app.get("/")
async def test() -> A:
    return A()
```

### Steps to reproduce

```bash
1. Run the litestar API
2. Go to http://127.0.0.1:8000/schema/swagger or a different doc format.
3. See Internal Server Error (or more details if API run with --debug)
4. Refresh, and see an empty schema
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.2.0

### Platform

- [ ] Linux
- [ ] Mac
- [X] Windows

peterschutt:
This PR modifies the way we generate schemas for objects that should be able to be referenced by other schemas.

It adds a new method to `SchemaCreator` called `create_component_schema` that should be used to create a schema that should be able to be referenced by other schemas, and itself.

This method takes care to first create the schema in the registry so that it can be immediately referenced.

In `SchemaCreator.for_field_definition()`, before we generate a schema for any plugin-supported type, we first check whether a reference can be made to an already existing schema and return that if possible.

Closes #2429


provinzkraut (approved):

JacobCoffee (approved):

litestar/plugins/base.py
```diff
@@ -190,10 +190,27 @@ class OpenAPISchemaPlugin(OpenAPISchemaPluginProtocol):
...
+            value: An arbitrary value.
+
+        Returns:
+            A typeguard dictating whether the value is supported by the plugin.
+        """
+        raise NotImplementedError(
+            "One of either is_plugin_supported_type or is_plugin_supported_field should be defined. "
+            "The default implementation of is_plugin_supported_field calls is_plugin_supported_type "
+            "for backwards compatibility. Users should prefer to override is_plugin_supported_field "
+            "as it receives a 'FieldDefinition' instance which is more useful than a raw type."
```
JacobCoffee:
```suggestion
            "One of either 'is_plugin_supported_type' or 'is_plugin_supported_field' should be defined. "
            "The default implementation of 'is_plugin_supported_field' calls 'is_plugin_supported_type' "
            "for backwards compatibility. Users should prefer to override 'is_plugin_supported_field' "
            "as it receives a 'FieldDefinition' instance which is more useful than a raw type."
```
to match `FieldDefinition`?

JacobCoffee (approved):

litestar/plugins/base.py
peterschutt: how is this outdated already?

Anyway, thanks - I remember thinking this when I wrote it but in the moment the satisfaction of having 3 lines with vertically aligned closing `"` was too strong :laughing:

peterschutt: No it was outdated straight after you reviewed.. it wouldn't let me commit the change. I only just make that other commit a few minutes ago.

No worries though I'll fix this before it's merged.

peterschutt:
For some reason, sonar is pinging me for coverage lines and conditions that have nothing to do with this PR :/

JacobCoffee:
> [93.75% Condition Coverage on New Code](https://sonarcloud.io/dashboard?id=litestar-org_litestar&pullRequest=2869) (required ≥ 98%)

could you get coverage up a bit?

peterschutt:
> > [93.75% Condition Coverage on New Code](https://sonarcloud.io/dashboard?id=litestar-org_litestar&pullRequest=2869) (required ≥ 98%)
> 
> 
> 
> could you get coverage up a bit?

image

Issue #2877, opened 2023-12-11
Bug: ServerSentEvent reverts to default event_type message after 1st message sent


euri10:
### Description

Here a small snippet and what i see in my browser

```
    @get("/stream")
    async def stream(self) -> ServerSentEvent:
        async def gen() -> AsyncGenerator[str, None]:
            c = 0
            while True:
                yield f"<div>{c}</div>\n"
                await asyncio.sleep(1)
                c += 1
        return ServerSentEvent(gen(), event_type="my_event")
```

this route in my browser leads to this:

![20231211_0835_1043x229_1702280116](https://github.com/litestar-org/litestar/assets/1104190/56be54c4-3786-4ce8-b24a-f5bdaa105e4a)

so my_event is sent only once then it defaults to the message default

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Litestar Version

2.4.3final0

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2877">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2877/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2877/pledge.svg">
</picture>
</a>

euri10:
after browsing the Discord it seems like this https://discord.com/channels/919193495116337154/1169881627544002590/1169881627544002590 is pretty much the same

PR #2888, opened 2023-12-12
fix: sse event type is only sent once

euri10:
wip, at the moment it fixes the issue below but is breaking existing tests,  though I'm not 100% positive the tests in question are correct, need more time and a clear head, help or pointers appreciated if this goes in a weird direction,

fixes https://github.com/litestar-org/litestar/issues/2877


euri10:
I think the tests that this PR broke and I had to change are false, in particular `test_sse_events_content` but I'm not sure.

I think the current implementation confuses two things, the stream and the sse which is why I renamed ServerSendEvent to ServerSendEventStream (this is what a handler has to return now), and it has to take a  generator that yields ServerSendEvent (now a simple dataclass)

provinzkraut:
> I think the tests that this PR broke and I had to change are false, in particular `test_sse_events_content` but I'm not sure.
> 
> I think the current implementation confuses two things, the stream and the sse which is why I renamed ServerSendEvent to ServerSendEventStream (this is what a handler has to return now), and it has to take a generator that yields ServerSendEvent (now a simple dataclass)

I agree. IMO the behaviour as you implemented it plus the changes to the test cases represent how the SSE *should* behave.

provinzkraut (commented):
This is a good fix, but unfortunately it's also breaking, since users will now have to return `ServerSentEventStream` where they previously could return `ServerSentEvent`. Do you see a way to make this backwards compatible?

euri10:
> This is a good fix, but unfortunately it's also breaking, since users will now have to return `ServerSentEventStream` where they previously could return `ServerSentEvent`. Do you see a way to make this backwards compatible?

what about I name back the `ServerSendEventStream` to `ServerSendEvent` and transform `ServerSendEvent` to `ServerSendEventMessage`

so we would have:

```
@get
async def hander_sse():
	async def generator:
			yield ServerSendEventMessage(event=...., id=.....)
	return ServerSendEvent(generator)
```

provinzkraut:
> > This is a good fix, but unfortunately it's also breaking, since users will now have to return `ServerSentEventStream` where they previously could return `ServerSentEvent`. Do you see a way to make this backwards compatible?
> 
> what about I name back the `ServerSendEventStream` to `ServerSendEvent` and transform `ServerSendEvent` to `ServerSendEventMessage`
> 
> so we would have:
> 
> ```
> @get
> async def hander_sse():
> 	async def generator:
> 			yield ServerSendEventMessage(event=...., id=.....)
> 	return ServerSendEvent(generator)
> ```

Sounds good!

litestar/response/sse.py
```diff
@@ -83,7 +142,7 @@ def __init__(
         encoding: str = "utf-8",
         headers: ResponseHeaders | None = None,
         event_type: str | None = None,
-        event_id: int | None = None,
+        event_id: str | None = None,
```
provinzkraut, line 145: Same as above

litestar/response/sse.py
```diff
@@ -14,21 +16,26 @@
...
+    __slots__ = ("content_async_iterator", "event_id", "event_type", "retry_duration", "comment_message")
 
     content_async_iterator: AsyncIteratorWrapper[bytes | str] | AsyncIterable[str | bytes] | AsyncIterator[str | bytes]
 
     def __init__(
         self,
         content: str | bytes | StreamType[str | bytes],
         event_type: str | None = None,
-        event_id: int | None = None,
+        event_id: str | None = None,
```
provinzkraut, line 31: This will have to be

```suggestion
        event_id: int | str | None = None,
```

at least for the time being. You can keep it as a `str` internally, but the public API needs to be backwards compatible.

provinzkraut (approved):
Looks good, just one minor issue!

provinzkraut:
@euri10 I think the pyright errors are due to the version bump. If you roll back the `pdm.lock` it should be fine. I'll try to get them fixed on `main` nevertheless.

euri10:
> @euri10 I think the pyright errors are due to the version bump. If you roll back the `pdm.lock` it should be fine. I'll try to get them fixed on `main` nevertheless.

done :)

PR #2889, opened 2023-12-13
ci: add new feature notifier

JacobCoffee:
Adds new CI workflow triggeredon comment to request users update their target base

e.g., if `feat:` is not found in the PR title you get this:

image

plus a failed CI task (that will need to be mandatory i guess)

### DIFF
--- a/.github/workflows/pr-target.yml
+++ b/.github/workflows/pr-target.yml
@@ -19,7 +19,6 @@ jobs:
     steps:
       - name: Check PR target
         id: check-pr-target
-        if: ${{ startsWith(github.event.pull_request.title, 'feat') }}
         uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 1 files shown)


JacobCoffee:
[new feature]

JacobCoffee:
[new feature]

JacobCoffee:
[new feature]

JacobCoffee:
[new feature]

JacobCoffee:
asdf

JacobCoffee:
[new feature]

JacobCoffee:
[new feature]

JacobCoffee:
test

JacobCoffee:
[feature]

.github/workflows/pr-target.yml
provinzkraut, line 1: After reading this again, I don't think the logic here makes sense.

The job will only run on `main`, so checking if it targets `main` and then failing if it's a feature PR is redundant. On the other hand, if we run this for all target branches, it will suggest that you target `develop` in any case, because that step doesn't take the target branch into account.

provinzkraut (changes requested):

PR #2893, opened 2023-12-13
refactor: openapi router

### LINKED ISSUE #2541: Enhancement: refactor openapi controller
### Summary

Refactor `OpenAPIController` into a suite of plugins that can be individually added to the application in order to serve a UI of choice for the project.

Currently, every app that configures openapi will have access to redoc, swagger, stoplight elements and rapidoc UIs. This is unnecessary, and the controller is beginning to become bloated.

### Basic Example

```py
app = Litestar(..., plugins=[ReDocUIPlugin()])
```

### Drawbacks and Impact

To implement this in a non-breaking way, we'll need to preserve the current behavior of having all UIs available if there is an OpenAPIConfig object defined and no plugin defined. However, this will give us a path to deprecating current behavior to prefer the plugin-based approach in 3.0.

### Unresolved questions

How should we identify that a plugin for an openapi UI has been registered on the app so as not to register the OpenAPIController (or equivalent)?

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2541">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2541/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
This PR refactors the way that we support multiple UIs for OpenAPI.

We add `litestar.openapi.plugins` where `OpenAPIRenderPlugin` is defined, and implementations of that plugin for all of the frameworks we currently support.

We add `OpenAPIConfig.render_plugins` config option, where a user can explicitly declare a set of plugins for UIs they wish to support.

If a user declares a sub-class of `OpenAPIController` at `OpenAPIConfig.openapi_controller`, then existing behavior is preserved exactly.

However, if no controller is explicitly declared, we invoke the new router-based approach, which should behave identically to the controller based approach (i.e., respect `enabled_endpoints` and `root_schema_site`).

Closes #2541

## TODO

- [x] rewrite docs & remove controller usage docs
- [x] deprecation of controller and related configurations
- [x] tests for plugins
- [x] functional tests that ensure router and controller behave the same
- [x] default UI is first plugin if plugins supplied to openapi config
- [x] API docs
  - [x] should we document the JSON plugin? It is always added so users don't really need to interact with it.
  - [x] should we export the plugins into contrib and document them in that namespace?


litestar/_openapi/plugin.py
```diff
@@ -43,13 +49,52 @@ def _build_openapi_schema(self) -> OpenAPI:
 
     def provide_openapi(self) -> OpenAPI:
         if not self._openapi_schema:
-            self._openapi_schema = self._build_openapi_schema()
+            self._openapi_schema = self._build_openapi()
         return self._openapi_schema
 
+    @staticmethod
+    def provide_openapi_schema(__openapi: Any) -> Dict[str, Any]:  # noqa: UP006
```
peterschutt: The double underscore is to avoid conflicts with any user DI that might be using the `openapi` name.

This causes the dependencies to have to be declared under the name `_OpenAPIPlugin__openapi` due to [name mangling](https://en.wikipedia.org/wiki/Name_mangling), but maybe better to just be explicit about this and name the deps `_openapi_plugin_openapi` or smth.
```python-console
>>> from msgspec import Struct
>>> class MyStruct(Struct):
...   __a: str
... 
>>> MyStruct(__a="abc")
TypeError: Unexpected keyword argument '__a'
>>> MyStruct(_MyStruct__a="abc")
MyStruct(_MyStruct__a='abc')
```

litestar/_openapi/plugin.py
```diff
@@ -43,13 +49,52 @@ def _build_openapi_schema(self) -> OpenAPI:
...
+        return __openapi.to_schema()  # type: ignore[no-any-return]
+
+    def create_openapi_router(self) -> Router:
+        def create_handler(plugin: OpenAPIRenderPlugin) -> HTTPRouteHandler:
+            paths = [plugin.path] if isinstance(plugin.path, str) else list(plugin.path)
+            if any(path.endswith(self.openapi_config.root_schema_site) for path in paths):
+                paths.append("/")
```
peterschutt, line 62: This is how I'm handling honoring the `OpenAPIConfig.root_schema_site` config.. a bit ugly - but in the future we could just make this the first registered render plugin.

litestar/_openapi/plugin.py
```diff
@@ -43,13 +49,52 @@ def _build_openapi_schema(self) -> OpenAPI:
...
+        for plugin in self.openapi_config.render_plugins:
+            plugin.receive_router(router)
+
+        return router
+
     def on_app_init(self, app_config: AppConfig) -> AppConfig:
         if app_config.openapi_config:
             self._openapi_config = app_config.openapi_config
-            app_config.route_handlers.append(self.openapi_config.openapi_controller)
+            if (controller := app_config.openapi_config.openapi_controller) is not None:
```
peterschutt, line 85: If a controller is passed to the config, then a user is sub-classing the `OpenAPIController` and so we register that and don't do anything different. This maintains the backward compat with the controller pattern.

litestar/_openapi/plugin.py
```diff
@@ -43,13 +49,52 @@ def _build_openapi_schema(self) -> OpenAPI:
...
+                app_config.route_handlers.append(controller)
+            else:
```
peterschutt, line 87: Where no controller has been explicitly registered on the config, then we switch to the new router/plugin based approach.

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,552 @@
...
+            <meta charset="utf-8"/>
+            <meta name="viewport" content="width=device-width, initial-scale=1">
+            <script src="{self.js_url}" crossorigin></script>
+            <style>{self.style}</style>
+          </head>
+        """
+
+        body = """
+          <body>
+            <rapi-doc spec-url="openapi.json" />
```
peterschutt, line 201: When we render the UI with a callback to retrieve the JSON, we require that the `openapi.json` endpoint is set in `OpenAPIConfig.enabled_endpoints`, but there is no actual checking or control of that dependency.

This is not new with this plugin-based implementation, but prob worth thinking about all the same.

litestar/_openapi/plugin.py
```diff
@@ -1,22 +1,55 @@
...
     from litestar.openapi.spec import OpenAPI
     from litestar.routes import BaseRoute
 
 
+def handle_not_found(path: str = "/") -> Response:
+    """Handler for returning HTML formatted errors from not-found schema paths.
+
+    This preserves backward compatibility with the Controller-based OpenAPI implementation.
+    """
+    if any(path.endswith(suffix) for suffix in (".json", ".yaml", ".yml")):
```
Alc-Alc, line 34: nit, these can also take a `tuple[str, ...]`
```suggestion
    if path.endswith((".json", ".yaml", ".yml")):
```

JacobCoffee (approved):

peterschutt:
One difference between the two implementation is that the router version doesn't have access to the favicon or style inside the 404 error handler.. I don't know how "breaking" to consider that.. we don't test for it at all.

I would have to move things like that onto OpenAPIConfig and pass them into the plugins to also have them available where the 404 handler is called.

Also, this might be desirable with the plugin-based architecture b/c they would currently have to be set per-plugin.

litestar/_openapi/plugin.py
```diff
@@ -1,22 +1,55 @@
...
     from litestar.config.app import AppConfig
+    from litestar.connection import Request
+    from litestar.handlers import HTTPRouteHandler
     from litestar.openapi.config import OpenAPIConfig
+    from litestar.openapi.plugins import OpenAPIRenderPlugin
     from litestar.openapi.spec import OpenAPI
     from litestar.routes import BaseRoute
 
 
+def handle_not_found(path: str = "/") -> Response:
```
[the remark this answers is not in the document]
peterschutt: I think it would be `not_found_handler()` in that case?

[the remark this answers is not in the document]
JacobCoffee: maybe? I thought it a bit weird since the var is named that. 

any of these? /shrug
- `render_html_not_found_error()`?
- `html_404_error_handler()`?
- `generate_html_error_response()`?

litestar/_openapi/plugin.py
```diff
@@ -43,13 +76,61 @@ def _build_openapi_schema(self) -> OpenAPI:
...
+                not_found_handler,
+            ],
+            include_in_schema=False,
+            dto=None,
+            return_dto=None,
+        )
+        for plugin in self.openapi_config.render_plugins:
+            plugin.receive_router(router)
+
+        return router
```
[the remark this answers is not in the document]
peterschutt: DYT we should offer some way to configure the router via the OpenAPIConfig? I.e., add guards and things? We can also just wait until we are asked I suppose.

litestar/_openapi/plugin.py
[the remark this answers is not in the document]
peterschutt: You pick - I'm happy with anything :)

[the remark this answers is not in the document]
JacobCoffee: oh man im slow, i understand this now to be $verb_not_found instead of what i thought was a typo..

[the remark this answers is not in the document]
peterschutt: yeh, but that misunderstanding is probably a good sign the name can be clearer.. I'm still happy to change it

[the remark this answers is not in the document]
provinzkraut: `handle_schema_not_found`?
IMO that should address all concerns.

litestar/openapi/plugins.py
provinzkraut, line 1: Love this!

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,552 @@
...
+    "SwaggerRenderPlugin",
+    "YamlRenderPlugin",
+)
+
+
+class OpenAPIRenderPlugin:
+    def __init__(
+        self,
+        *,
+        path: str | tuple[str, ...],
```
provinzkraut, line 31: Why `tuple[str, ...]`?

litestar/openapi/plugins.py
provinzkraut: > I think we should just always include the `JsonRenderPlugin` so that `openapi.json` is always available. It is a pretty efficient endpoint, and doesn't require any extra dependencies.

Yeah, JSON should be enabled by default. Everything else is optional.

tests/unit/test_contrib/test_repository.py
```diff
@@ -29,6 +29,7 @@ def test_app_repository_signature_namespace() -> None:
         "NotInSearchFilter": NotInSearchFilter,
         "FilterTypes": FilterTypes,
     }
+    assert all(app.signature_namespace[k] == v for k, v in expected_namespace.items())
```
provinzkraut, line 32: Any reason for this change? Direct comparison gives nicer outputs in pytest because it then includes a diff.

provinzkraut (approved):

litestar/openapi/plugins.py
peterschutt: it was typed as `str | list[str]` which is the same typing as our handler paths, however providing a default value of `=["openapi.yaml", "openapi.yml"]` would be flagged by linters as providing a mutable default value.

I'll type it `Sequence[str]` instead.

litestar/_openapi/plugin.py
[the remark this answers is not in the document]
peterschutt: went with `handle_schema_path_not_found`

tests/unit/test_contrib/test_repository.py
peterschutt: initially, the `OpenAPIPlugin` was adding `OpenAPI` to the signature namespace causing this to fail.

Generally, though, I don't think that tests should assume the signature namespace is empty if they are testing their own interaction with it - because it might not be, especially as we adopt using plugins internally over time.

I can back this out because I've not ended up using DI for the `OpenAPI` object anymore, but even before then I didn't need it because of that whole unconditional type recursion for dataclasses that msgspec does. I was pretty much forced to put either the entire `openapi.spec` namespace into `signature_namespace` to support it wanting to recurse through the subtypes of `OpenAPI`, or type it as `Any` - which is what I did... until I removed it from DI altogether. I think this is a real issue that needs to be addressed.. if a user wants to annotate a handler with a dependency type that they don't control, e.g., comes from a 3rd party, and that 3rd party arbitrarily uses `@dataclass` types and `if TYPE_CHECKING` imports, then it blows up and the user is forced to disable that dependency validation.

Another thing I just thought of is that we should have some internal convention for population of the signature namespace for our internal things, e.g., `from litestar.openapi.spec import OpenAPI as _LS_OpenAPI` or something like that, to reduce the potential for name collisions with user land stuff.

provinzkraut: > Another thing I just thought of is that we should have some internal convention for population of the signature namespace for our internal things, e.g., `from litestar.openapi.spec import OpenAPI as _LS_OpenAPI` or something like that, to reduce the potential for name collisions with user-land stuff.

Sounds like a good idea!

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,584 @@
...
+    from litestar.connection import Request
+    from litestar.router import Router
+
+__all__ = (
+    "OpenAPIRenderPlugin",
+    "RapidocRenderPlugin",
+    "RedocRenderPlugin",
+    "StoplightRenderPlugin",
+    "SwaggerRenderPlugin",
+    "YamlRenderPlugin",
```
JacobCoffee, line 22:
```suggestion
    "OpenAPIRenderPlugin",
    "RapidocRenderPlugin",
    "RedocRenderPlugin",
    "StoplightRenderPlugin",
    "SwaggerRenderPlugin",
    "YamlRenderPlugin",
    "JsonRenderPlugin",
```

docs/usage/openapi/index.rst
```diff
@@ -0,0 +1,22 @@
...
+
+- Automatic `OpenAPI 3.1.0 Schema <https://spec.openapis.org/oas/v3.1.0>`_ generation, which is available as both YAML
+  and JSON.
+- Builtin support for static documentation site generation using several different libraries.
+- Simple configuration using pydantic based classes.
+
+
+Litestar includes a complete implementation of the `latest version of the OpenAPI specification <https://spec.openapis.org/oas/latest.html>`_
+using Python dataclasses. This implementation is used as a basis for generating OpenAPI specs, supporting builtins including
+`dataclasses` and `TypedDict`, as well as Pydantic models and any 3rd party entities for which a plugin is implemented.
```
JacobCoffee, line 14:
```suggestion
``dataclasses`` and ``TypedDict``, as well as Pydantic models and any 3rd party entities for which a plugin is implemented.
```

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,233 @@
...
+- ``**kwargs``: Captures additional arguments to pass to the superclass.
+
+And we construct a url for the RapiDoc JavaScript bundle if it is not provided:
+
+.. literalinclude:: /examples/openapi/plugins/custom_plugin.py
+    :language: python
+    :lines: 30
+
+render()
+~~~~~~~~
```
JacobCoffee, line 203:
```suggestion
``render()``
~~~~~~~~~~
```

litestar/openapi/plugins.py
peterschutt: Intentionally left the json one undocumented because we always add it. Happy to discuss.

docs/usage/openapi/index.rst
peterschutt: Thanks!

docs/usage/openapi/ui_plugins.rst
peterschutt: Thanks!

litestar/openapi/plugins.py
JacobCoffee: I just saw it angry in the tests as undeclared. not pressed about it :D

docs/usage/openapi/schema_generation.rst
```diff
@@ -191,134 +172,6 @@ app instance itself. For example:
...
-
-
-   app = Litestar(
-       route_handlers=[...],
-       openapi_config=OpenAPIConfig(
-           title="My API", version="1.0.0", openapi_controller=MyOpenAPIController
-       ),
-   )
-
-OAuth2 in Swagger UI
```
peterschutt, line 248: Need to include this in the new docs

docs/usage/openapi/schema_generation.rst
```diff
@@ -191,134 +172,6 @@ app instance itself. For example:
...
-   app = Litestar(
-       route_handlers=[...],
-       openapi_config=OpenAPIConfig(
-           title="My API", version="1.0.0", openapi_controller=MyOpenAPIController
-       ),
-   )
-
-
-
-CDN and offline file support
```
peterschutt, line 284: Same for new docs?

docs/usage/openapi/ui_plugins.rst
peterschutt, line 203:
```suggestion
``render()``
~~~~~~~~~~~~
```

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,584 @@
...
+
+    paths: list[str]
+
+    def __init__(
+        self,
+        *,
+        path: str | Sequence[str],
+        media_type: MediaType | OpenAPIMediaType = MediaType.HTML,
+        favicon: str = _default_favicon,
+        style: str = "body { margin: 0; padding: 0 }",
```
peterschutt, line 40: It is inconsistent that we ask for a `<link>` tag for the favicon arg, but not a `<style>` tag for the style arg.

Previously, we only asked for a url for the favicon and constructed the `<link>` tag internally, however the `type=` attribute of the `<link>` tag for the favicon also needs to be specified depending on the the type of the image that is linked. So we modified it to accept the full `<link>`.

`<style>` tags similarly accept attributes so there is utility to accepting a complete `<style>` tag for the style arg beyond consistency.

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,584 @@
...
+    "StoplightRenderPlugin",
+    "SwaggerRenderPlugin",
+    "YamlRenderPlugin",
+)
+
+_favicon_url = "https://raw.githubusercontent.com/litestar-org/branding/main/assets/Branding%20-%20PNG%20-%20Transparent/Badge%20-%20Blue%20and%20Yellow.png"
+_default_favicon = f"<link rel='icon' type='image/png' href='{_favicon_url}'>"
+
+
+class OpenAPIRenderPlugin:
```
peterschutt, line 29: `OpenAPIUIPlugin`? I realize that I've inadvertently used the phrasing "OpenAPI UI Plugin" throughout the docs - I wonder if that naming is more intuitive?

litestar/openapi/config.py
```diff
@@ -88,31 +105,111 @@ class OpenAPIConfig:
...
+    """Plugins for rendering OpenAPI documentation UIs."""
+    openapi_controller: type[OpenAPIController] | None = None
+    """Controller for generating OpenAPI routes.
+
+    Must be subclass of :class:`OpenAPIController <litestar.openapi.controller.OpenAPIController>`.
+    """
+    root_schema_site: Literal["redoc", "swagger", "elements", "rapidoc"] | None = None
+    """The static schema generator to use for the "root" path of `/schema/`."""
+    enabled_endpoints: set[str] | None = None
+    """A set of the enabled documentation sites and schema download endpoints."""
```
peterschutt, line 125: What about something like
```py
    router_kws: RouterKwsDict = field(default_factory=dict)
    """Keyword arguments passed to the OpenAPI router on instantiation."""
```
This would allow users to define things like guards etc for the router. Thoughts?

litestar/_openapi/plugin.py
```diff
@@ -1,22 +1,56 @@
...
+            <title>404 Not found</title>
+            <meta charset="utf-8"/>
+            <meta name="viewport" content="width=device-width, initial-scale=1">
+        </head>
+        <body>
+            <h1>Error 404</h1>
+        </body>
+    </html>
+    """
+    return Response(content, media_type=MediaType.HTML, status_code=HTTP_404_NOT_FOUND)
```
peterschutt, line 51: Consider adding this handler to the openapi config object so that it can be overridden by users

litestar/openapi/plugins.py
peterschutt: Or just `OpenAPIPlugin`?

provinzkraut: `OpenAPIRenderPlugin` is a bit verbose but I think that's okay. `OpenAPIPlugin` is too vague, an `OpenAPI UI Plugin` is not an accurate description, since these are used to not only display UIs, but also JSON and YAML.

litestar/openapi/plugins.py
provinzkraut: +1 for more consistency

litestar/openapi/config.py
provinzkraut: Why not just allow passing in a router directly? We could still `register` things dynamically on it then at that would make passing the config a bit easier.

peterschutt: Added `openapi_router` to the `OpenAPIConfig`.

docs/examples/openapi/plugins/custom_plugin.py
```diff
@@ -0,0 +1,70 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Initialize the OpenAPI UI render plugin.
+
+        Args:
+            version: Rapidoc version to download from the CDN. If js_url is provided, this is ignored.
+            js_url: Download url for the RapiDoc JS bundle. If not provided, the version will be used to construct the
+                url.
+            path: Path to serve the OpenAPI UI at.
+            **kwargs: Additional arguments to pass to the base class.
```
JacobCoffee, line 28: do we want all of the docstring stuff in the examples?

docs/examples/openapi/plugins/custom_plugin.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        self.js_url = js_url or f"https://unpkg.com/rapidoc@{version}/dist/rapidoc-min.js"
+        super().__init__(path=path, **kwargs)
+
+    def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
+        """Render an HTML page for Rapidoc.
+
+        Notes:
+            - override this method to customize the template.
```
JacobCoffee, line 37: same

docs/examples/openapi/plugins/custom_plugin.py
```diff
@@ -0,0 +1,70 @@
...
+            <meta charset="utf-8"/>
+            <meta name="viewport" content="width=device-width, initial-scale=1">
+            <script src="{self.js_url}" crossorigin></script>
+            {self.style}
+          </head>
+        """
+
+        body = """
+          <body>
+            <rapi-doc spec-url="openapi.json" />
```
JacobCoffee, line 60: sorry if this is dumb, but is this an accurate path considering https://github.com/litestar-org/litestar/pull/3196

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+
+In addition to serving a default JSON representation of the OpenAPI specification, built-in UI Plugins are available to
+support a range of popular OpenAPI documentation tools, including:
+
+- `Scalar <https://scalar.com/>`_
+- `RapiDoc <https://rapidocweb.com/>`_
+- `ReDoc <https://redocly.com/>`_
+- `Stoplight Elements <https://stoplight.io/open-source/elements>`_
+- `Swagger UI <https://swagger.io/tools/swagger-ui/yy>`_
+- `YAML <https://yaml.org/>`_ (OK, not a UI framework, but configured the same way!)
```
JacobCoffee, line 16: Missing JSON maybe?

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+
+
+Using OpenAPI UI Plugins
+------------------------
+
+Using OpenAPI UI Plugins is as simple as importing the plugin, instantiating it, and adding it to the OpenAPIConfig.
+
+.. tab-set::
+
+    .. tab-item:: scalar
```
JacobCoffee, line 29: If we add a tab sync here, when they click e.g., `scalar` on this tab set, it will change in all others inside of the respective document.

JacobCoffee: see https://sphinx-design.readthedocs.io/en/latest/tabs.html#synchronised-tabs

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+.. literalinclude:: /examples/openapi/customize_path.py
+    :language: python
+
+This will result in any of the OpenAPI endpoints being served at ``/docs`` instead of ``/schema``, e.g.,
+``/docs/openapi.json``.
+
+Backward Compatibility
+----------------------
+
+OpenAPI UI plugins are a new feature introduced in ``v2.5.0``.
```
JacobCoffee, line 134:
```suggestion
OpenAPI UI plugins are a new feature introduced in ``v2.5.0``.
```

Change version here (?) but also should just use the [`..versionadded: vX.Y.Z` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded)

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+
+Providing a subclass of OpenAPIController
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The previous method of configuring elements such as the root path and styling was to subclass
```
JacobCoffee, line 139: Probably worth using [`..deprecated`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) as well

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Litestar has always supported a ``root_schema_site`` attribute on the :class:`OpenAPIConfig` class. This attribute
+allowed you to elect to serve a UI at the OpenAPI root path, e.g., by default ``redoc`` would be served at both
+``/schema`` and ``/schema/redoc``.
+
+In ``v3.0.0``, the ``root_schema_site`` attribute will be removed, and the first :class:`OpenAPIRenderPlugin` in the
+:attr:`OpenAPIConfig.render_plugins` list will be assigned to the ``/schema`` endpoint.
+
+As of ``v2.5.0``, if you explicitly use the new :attr:`OpenAPIConfig.render_plugins` attribute, you will be
```
JacobCoffee, line 171: Slated for 2.8.0 now

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+
+We that you can preset your clientId or enable PKCE support.
+
+.. literalinclude:: /examples/openapi/plugins/swagger_ui_oauth.py
+    :language: python
+
+Customizing the OpenAPI UI
+--------------------------
+
+Currently, the ``Scalar`` OpenAPI UI plugin supports customization of its style and behavior by
```
JacobCoffee, line 261: Rapidoc does as well, but I havent looked deeply

litestar/openapi/config.py
```diff
@@ -88,31 +106,128 @@ class OpenAPIConfig:
...
+    """Plugins for rendering OpenAPI documentation UIs."""
+    openapi_router: Router | None = None
+    """A router for serving OpenAPI documentation and schema files.
+
+    If provided, ``path`` is ignored.
+
+    Ignored if the deprecated ``openapi_controller`` is provided.
+
+    Not required, however may be passed to customize the configuration of the router used to serve
+    the documentation endpoints. For example, to add middleware or guards to the router.
```
JacobCoffee, line 128:
```suggestion
    """An optional router for serving OpenAPI documentation and schema files.

    If provided, ``path`` is ignored.

    This parameter is also ignored if the deprecated :class:`OpenAPIConfig <.openapi.OpenAPIConfig>` ``openapi_controller`` kwarg is provided.

    The ``openapi_router`` is not required, but it can be passed to customize the configuration of the router used to serve the documentation endpoints. For example, you can add middleware or guards to the router.
```
this was a little confusing to me, feel free to ignore because I am likely bad

litestar/openapi/config.py
```diff
@@ -88,31 +106,128 @@ class OpenAPIConfig:
...
+    Handlers to serve the OpenAPI schema and documentation sites are added to this router according
+    to the ``render_plugins`` attribute, so routes shouldn't be added that conflict with these.
+    """
+    openapi_controller: type[OpenAPIController] | None = None
+    """Controller for generating OpenAPI routes.
+
+    Must be subclass of :class:`OpenAPIController <litestar.openapi.controller.OpenAPIController>`.
+    """
+    root_schema_site: Literal["redoc", "swagger", "elements", "rapidoc"] | None = None
+    """The static schema generator to use for the "root" path of `/schema/`."""
```
JacobCoffee, line 139:
```suggestion
    """The static schema generator to use for the "root" path of ``/schema/``."""
```

litestar/openapi/config.py
```diff
@@ -88,31 +106,128 @@ class OpenAPIConfig:
...
+                        self.default_plugin = plugin
+                        break
+                else:
+                    self.default_plugin = self.render_plugins[0]
+
+    def _issue_deprecations(self) -> None:
+        """Handle deprecated config options."""
+        if self.openapi_controller is not None:
+            warn_deprecation(
+                "v2.7.0", "openapi_controller", "attribute", removal_in="v3.0.0", alternative="render_plugins"
```
JacobCoffee, line 186: Slated for 2.8.0 now

JacobCoffee (approved):

JacobCoffee:
@marclave would you review the Scalar bits? That PR has been merged into this one and we are preparing this one for release when it is up to standard

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+    "OpenAPIRenderPlugin",
+    "RapidocRenderPlugin",
+    "RedocRenderPlugin",
+    "ScalarRenderPlugin",
+    "StoplightRenderPlugin",
+    "SwaggerRenderPlugin",
+    "YamlRenderPlugin",
+)
+
+_favicon_url = "https://raw.githubusercontent.com/litestar-org/branding/main/assets/Branding%20-%20PNG%20-%20Transparent/Badge%20-%20Blue%20and%20Yellow.png"
```
JacobCoffee, line 28: Maybe we point this at a CDN as well? e.g.,
Host it as is, put generate the CDN url for end user redundancy through https://www.jsdelivr.com/github.
`https://cdn.jsdelivr.net/gh/litestar-org/branding@main/assets/Branding%20-%20PNG%20-%20Transparent/Badge%20-%20Blue%20and%20Yellow.png`

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+        )
+
+
+class ScalarRenderPlugin(OpenAPIRenderPlugin):
+    """Plugin to render an OpenAPI schema using Scalar.
+
+    .. versionadded:: 2.5.0
+    """
+
+    _default_litestar_css_url = "https://raw.githubusercontent.com/litestar-org/branding/main/assets/openapi/scalar.css"
```
JacobCoffee, line 325: similar to the above via jsdelivr like we do the other urls (redoc, swagger, etc)
`https://cdn.jsdelivr.net/gh/litestar-org/branding@main/assets/openapi/scalar.css`

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+        """
+        self.js_url = js_url or f"https://unpkg.com/@stoplight/elements@{version}/web-components.min.js"
+        self.css_url = css_url or f"https://unpkg.com/@stoplight/elements@{version}/styles.min.css"
+        super().__init__(path=path, **kwargs)
+
+    def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
+        """Render an HTML page for StopLight Elements.
+
+        Notes:
+            - override this method to customize the template.
```
JacobCoffee, line 459:
```suggestion
       .. note:: Override this method to customize the template.
```
utilize sphinx admonition

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+            **kwargs: Additional arguments to pass to the base class.
+        """
+        self.js_url = js_url or f"https://unpkg.com/rapidoc@{version}/dist/rapidoc-min.js"
+        super().__init__(path=path, **kwargs)
+
+    def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
+        """Render an HTML page for Rapidoc.
+
+        Notes:
+            - override this method to customize the template.
```
JacobCoffee, line 207:
```suggestion
        .. note:: Override this method to customize the template.
```

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+        """
+        self.js_url = js_url or f"https://cdn.jsdelivr.net/npm/redoc@{version}/bundles/redoc.standalone.js"
+        self.google_fonts = google_fonts
+        super().__init__(path=path, **kwargs)
+
+    def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
+        """Render an HTML page for Redoc.
+
+        Notes:
+            - override this method to customize the template.
```
JacobCoffee, line 272:
```suggestion
        .. note:: Override this method to customize the template.
```

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+
+        Args:
+            version: SwaggerUI version to download from the CDN. If js_url is provided, this is ignored.
+            js_url: Download url for the Swagger UI JS bundle. If not provided, the version will be used to construct
+                the url.
+            css_url: Download url for the Swagger UI CSS bundle. If not provided, the version will be used to construct
+                the url.
+            standalone_preset_js_url: Download url for the Swagger Standalone Preset JS bundle. If not provided, the
+                version will be used to construct the url.
+            init_oauth: JSON to initialize Swagger UI OAuth2 by calling the `initOAuth` method.
```
JacobCoffee, line 522:
```suggestion
            init_oauth: JSON to initialize Swagger UI OAuth2 by calling the ``initOAuth`` method.
```

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,694 @@
...
+        router.register(
+            get("/oauth2-redirect.html", media_type=MediaType.HTML, sync_to_thread=False)(self.render_oauth2_redirect),
+        )
+
+    @staticmethod
+    def render_oauth2_redirect() -> bytes:
+        """Render an HTML oauth2-redirect.html page for Swagger-UI.
+
+        Notes:
+            - override this method to customize the template.
```
JacobCoffee, line 611:
```suggestion
        .. note:: Override this method to customize the template.
```

JacobCoffee (commented):
sorry for all the pings :\

JacobCoffee:
> should we export the plugins into contrib and document them in that namespace?

Maybe? Moving `litestar/openapi/plugins.py` into `litestar/plugins/openapi.py` might adhere to principle of least astonishment (having all our plugins in one place)

peterschutt:
> > should we export the plugins into contrib and document them in that namespace?
> 
> 
> 
> Maybe? Moving `litestar/openapi/plugins.py` into `litestar/plugins/openapi.py` might adhere to principle of least astonishment (having all our plugins in one place)

I feel that b/c they aren't plugins in the sense that they can be passed to `Litestar.plugins` that this separation is actually useful.

JacobCoffee:
@peterschutt either sgtm! :)

docs/usage/openapi/ui_plugins.rst
peterschutt: The JSON plugin is non-optional.

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: Yeh not sure - thanks for pointing out. @guacs can you help me out with this one please - all I've done is transfer the logic we had on the controller routes into the individual plugins so a good chance I need to do something differently.

docs/usage/openapi/ui_plugins.rst
peterschutt: I can include it if you want but the way it works is that if a plugin isn't already registered on the that serves at `/openapi.json`, then we add it - so if an OpenAPIConfig is provided for the application `openapi.json` will always be there.

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: related: https://github.com/litestar-org/litestar/pull/2893#discussion_r1437898435

docs/usage/openapi/ui_plugins.rst
peterschutt: related: https://github.com/litestar-org/litestar/pull/2893#discussion_r1437898435

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: Happy for you to push in any changes you want to the docs @JacobCoffee

docs/PYPI_README.md
```diff
@@ -87,6 +87,7 @@ app = Litestar(route_handlers=[hello_world])
   [msgspec](https://github.com/jcrist/msgspec) and [attrs](https://www.attrs.org/en/stable/)
 - Layered parameter declaration
 - [Automatic API documentation with](#redoc-swagger-ui-and-stoplight-elements-api-documentation):
+  - [Scalar](https://scalar.com/)
```
marclave, line 90: we should point to https://github.com/scalar/scalar ✨

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+------------------
+
+OpenAPI UI Plugins are designed to allow easy integration with your OpenAPI UI framework of choice. These plugins
+facilitate the creation of interactive, user-friendly API documentation, making it easier for developers and end-users
+to understand and interact with the API.
+
+In addition to serving a default JSON representation of the OpenAPI specification, built-in UI Plugins are available to
+support a range of popular OpenAPI documentation tools, including:
+
+- `Scalar <https://scalar.com/>`_
```
marclave, line 11: we should point to https://github.com/scalar/scalar ✨

docs/usage/openapi/ui_plugins.rst
```diff
@@ -0,0 +1,268 @@
...
+--------------------------
+
+Currently, the ``Scalar`` OpenAPI UI plugin supports customization of its style and behavior by
+overriding the default ``css_url`` and ``js_url`` attributes on its render plugin class.
+
+.. literalinclude:: /examples/openapi/plugins/scalar_customized.py
+            :language: python
+
+To learn more about customizing the ``Scalar`` UI,
+see the `Scalar documentation <https://docs.scalar.com/>`_.
```
marclave, line 268: we should point to https://github.com/scalar/scalar ✨

marclave (approved):
everything is looking AWESOME! ✨ so stoked on this, the only thing we should do is make all outbound links to scalar go to the repo :) https://github.com/scalar/scalar ✨

I added comments on where it is in the PR, happy to update these myself :)

README.md
```diff
@@ -90,6 +90,7 @@ app = Litestar(route_handlers=[hello_world])
   [msgspec](https://github.com/jcrist/msgspec) and [attrs](https://www.attrs.org/en/stable/)
 - Layered parameter declaration
 - [Automatic API documentation with](#redoc-swagger-ui-and-stoplight-elements-api-documentation):
+  - [Scalar](https://scalar.com/)
```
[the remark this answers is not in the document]
peterschutt: Thanks!

docs/usage/openapi/ui_plugins.rst
JacobCoffee: ahh i remember now, ty

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: removed - also I think I'll change the example case to the scalar plugin since that's the default one.

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: We also have same in the scalar plugin:

```py
        body = f"""
                <noscript>
                    Scalar requires Javascript to function. Please enable it to browse the documentation.
                </noscript>
                <script
                  id="api-reference"
                  data-url="openapi.json">
                </script>
                <script src="{self.js_url}" crossorigin></script>
                ""

File: docs/usage/openapi/ui_plugins.rst (hunk 24 above)
Comment (peterschutt) [reply docs/usage/openapi/ui_plugins.rst:29]: got this thanks

File: docs/usage/openapi/ui_plugins.rst (hunk 25 above)
Comment (peterschutt) [reply docs/usage/openapi/ui_plugins.rst:134]: Maybe at the top of the docs, or in the docstring for the plugins, but as this is a section about backward compatibility it serves as an introduction to the section IMO.

Comment (peterschutt) [reply docs/usage/openapi/ui_plugins.rst:134]: I have added it at the top of the page.

File: litestar/openapi/plugins.py [hunk 43]
```diff
@@ -0,0 +1,691 @@
...
+        """
+        try:
+            with httpx.Client() as client:
+                response = client.get(self._default_litestar_css_url)
+                response.raise_for_status()
+                return response.content
+        except httpx.HTTPError:
+            here = Path(__file__).parent
+            css_path = here / "styles/scalar.css"
+            return css_path.read_bytes()
```
peterschutt, line 424: Is all this necessary @JacobCoffee? We don't do anything like this for any of the other plugins to ensure that the css is available, and I can't see that the `openapi/styles/scalar.css` file exists.

Can we just leave it to hit the cdn for our style sheet like the others?

litestar/openapi/plugins.py
```diff
@@ -0,0 +1,687 @@
...
+
+
+class ScalarRenderPlugin(OpenAPIRenderPlugin):
+    """Plugin to render an OpenAPI schema using Scalar.
+
+    .. versionadded:: 2.8.0
+    """
+
+    _default_litestar_css_url = "https://cdn.jsdelivr.net/gh/litestar-org/branding@main/assets/openapi/scalar.css"
+    _default_scalar_css_url = "/schema/_litestar-internal/scalar.css"
```
peterschutt, line 321: I don't think the `/schema/` part of this can be hard coded as it configurable on `OpenAPIConfig`

litestar/openapi/plugins.py
peterschutt: was it b/c we were using a `githubusercontent` link before?

litestar/_openapi/plugin.py
peterschutt: This can be an improvement if requested.

docs/examples/openapi/plugins/custom_plugin.py
peterschutt: I've found it works if you request `/schema/` (with trailing `/`) but not `/schema` - I suppose I have to work out how to implement something similar to #3196.

peterschutt: OK sorry about the chatter on this one. I've cherry picked #3196 into this branch and then pretty much implemented the same thing in the context of the openapi router/plugins. @guacs would you mind taking a look at https://github.com/litestar-org/litestar/pull/2893/commits/8f1344701b588785ad0639392385116d0cf31852

litestar/constants.py
```diff
@@ -14,6 +17,7 @@
 HTTP_RESPONSE_BODY: Final = "http.response.body"
 HTTP_RESPONSE_START: Final = "http.response.start"
 ONE_MEGABYTE: Final = 1024 * 1024
+OPENAPI_JSON_HANDLER_NAME: Final = f"{uuid4().hex}_litestar_openapi_json"
```
peterschutt, line 20: The UUID is to prevent users being able to depend on the absolute value and also prevent collision to user-space names, but I'm not married to the idea either if anyone thinks it OTT.

litestar/openapi/plugins.py
peterschutt: @JacobCoffee I did this https://github.com/litestar-org/litestar/pull/2893/commits/6cdfe69ab7723b3745adbc6170c6ddf9855c7343 - but we can revert. Mainly I just wanted to see if it would get me across the line with coverage.

litestar/_openapi/plugin.py
```diff
@@ -49,14 +85,106 @@ def _build_openapi_schema(self) -> OpenAPI:
...
+
+            If the plugin is the default plugin, a handler is created for the root path in addition
+            to the plugin's configured paths.
+
+            If the plugin has a path for serving the OpenAPI schema file, the `openapi_json_found`
+            flag is set to `True`, so that we don't create a handler for serving the JSON schema file.
+
+            Args:
+                plugin_: The plugin to create the handler for.
+                handler_name: The name of the handler.
```
peterschutt, line 137:
```suggestion
```

provinzkraut:
:partying_face:

litestar/openapi/config.py
```diff
@@ -90,31 +108,123 @@ class OpenAPIConfig:
...
+
+    Handlers to serve the OpenAPI schema and documentation sites are added to this router according
+    to the ``render_plugins`` attribute, so routes shouldn't be added that conflict with these.
+    """
+    openapi_controller: type[OpenAPIController] | None = None
+    """Controller for generating OpenAPI routes.
+
+    Must be subclass of :class:`OpenAPIController <litestar.openapi.controller.OpenAPIController>`.
+    """
+    root_schema_site: Literal["redoc", "swagger", "elements", "rapidoc"] | None = None
```
JacobCoffee, line 139: Upon testing some production apps against this it seems to be missing `scalar` unless I'm dumb?

peterschutt: In case others come across this, scalar wasn't added to `root_schema_site` literals because that param is deprecated along side the `OpenAPIController` pattern.

To use Scalar in 2.8, one will need to do this:

```py
config = OpenAPIConfig(
    ...
    render_plugins=[ScalarRenderPlugin()]
)
```
Using this pattern implicitly opts the user into "future" mode where `root_schema_site` is ignored and the first plugin named in the `render_plugins` sequence is served at the schema root.

The above pattern (i.e., `ScalarRenderPlugin` as a single render plugin) will become the default behavior in 3.0.

If no `render_plugin` or `openapi_controller` is provided to the config, this enables a compatibility mode, where the future plugin/router based approach is used to emulate the deprecated default `OpenAPIController` behavior.

If an `openapi_controller` is provided, then the user is opted into "classic" mode, which is no change from 2.7.

PR #2894, opened 2023-12-13
fix(channels): Trailing messages after unsubscribes

provinzkraut:
Fix a bug that would allow a channels backend to receive messages from a channel it just unsubscribed from, for a short period of time, due to how the different brokers handle unsubscribes.

```python
await backend.subscribe(["foo", "bar"])  # subscribe to two channels
await backend.publish(b"something", ["foo"])  # publish a message to a channel we're subscribed to

# start the stream after publishing. Depending on the backend
# the previously published message might be in the stream
event_generator = backend.stream_events()

# unsubscribe from the channel we previously published to
await backend.unsubscribe(["foo"])

# this should block, as we expect messages from channels
# we unsubscribed from to not appear in the stream anymore
print(anext(event_generator))
```

### DIFF
--- a/litestar/channels/backends/asyncpg.py
+++ b/litestar/channels/backends/asyncpg.py
@@ -76,8 +76,12 @@ class AsyncPgChannelsBackend(ChannelsBackend):
             raise RuntimeError("Backend not yet initialized. Did you forget to call on_startup?")
 
         while True:
-            yield await self._queue.get()
+            channel, message = await self._queue.get()
             self._queue.task_done()
+            # an UNLISTEN may be in transit while we're getting here, so we double-check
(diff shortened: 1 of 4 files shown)


provinzkraut:
Although this is a fix for a bug present in `main`, I'm targeting `develop` here because it builds upon features only present in `develop`, and the bug isn't that severe.

cofin:
Is the `psycopg` backend not impacted by this?

cofin (approved):

provinzkraut:
> Is the `psycopg` backend not impacted by this?

No, it's not. We `UNLISTEN` directly there, which immediately cancels our `LISTEN`, so we don't have the "pending unsubscribe" state we have with asyncpg.

Issue #2892, opened 2023-12-13
Enhancement: default handlers names for url_for use


euri10:
### Summary

Currently if in a template I want t o do:
`<a href="{{ url_for("login_get_name") }}"></a>`

then in litestar I have to explicitely use the name kwarg in the handler's decorator:
```
@get("/login", name="login_get_name")
async def login_get_func(
    request: Request
) -> Template:
```

I'd like to be able to do `<a href="{{ url_for("login_get_func") }}"></a>` with:
```
@get("/login")
async def login_get_func(
    request: Request
) -> Template:
```
currently a 500 is raised
` litestar.exceptions.http_exceptions.NoRouteMatchFoundException: 500: Route login_get_func can not be found`

peterschutt:
A few things to think about:

- if functions with the same name are declared in different modules and both registered on the app:

````py
# a.py

@get("/a")
def get_handler() -> A:
    ...
```

```py
# b.py

@get("/b")
def get_handler() -> B:
    ...
```

```py
# app.py

from a import get_handler as get_a
from b import get_handler as get_b

app = Litestar([get_a, get_b])
```

- where handlers are registered on a router
- where handlers are registered on a controller

What if we built a name from the handler method/func that included the module path and the `__qualname__` of the handler?

provinzkraut:
> if functions with the same name are declared in different modules and both registered on the app

Also, if a handler is registered multiple times

erhuabushuo:
maybe module.function

provinzkraut:
> What about "module.function"

Unfortunately it's not that easy, because it would mean that

```python
class SomeController(Controller):
  @get()
  def handler(self) -> str:
    ...
```
and 

```python
@get()
def handler() -> str:
  ...
```

would produce the same name.

The pattern @peterschutt suggested in [this comment](https://github.com/litestar-org/litestar/issues/2892#issuecomment-1853508587), using [`__qualname__`](https://docs.python.org/3/glossary.html#term-qualified-name) would get around this, since `__qualname__` is the "full attribute path" to an object, taking into account containing objects such as classes.

PR #2895, opened 2023-12-13
feat: add default names to handlers

euri10:
fixes https://github.com/litestar-org/litestar/issues/2892

draft to propose a potential app with all `func` handlers that covers all the cases mentionned in the issue, hope I didnt forget something,
Once we ok we have all cases, I put in comments the expected default named that `__qualname__` should give us

@peterschutt could you point me where the naming happens ? you mentionned the `__call__` of BaseRouteHandler in the chat but I dont see anything relevant

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,7 +56,7 @@ repos:
       - id: slotscheck
         exclude: "test_*|docs"
   - repo: https://github.com/sphinx-contrib/sphinx-lint
-    rev: "v0.9.0"
+    rev: "v0.9.1"
     hooks:
       - id: sphinx-lint
   - repo: local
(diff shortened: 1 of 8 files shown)


peterschutt:
> @peterschutt could you point me where the naming happens ? you mentionned the `__call__` of BaseRouteHandler in the chat but I dont see anything relevant

`name` is an `__init__()` argument to the handlers:

https://github.com/litestar-org/litestar/blob/a26e1347503c9bc345027cbe02a4cc8adf5ea16d/litestar/handlers/base.py#L87

I mentioned `__call__()` b/c that is the first place we can get our hands on the handler function within the context of the handler itself, however it would probably be better to do it in the `on_registration()` callback, which is called when the application is instantiated, rather than when the handler function is decorated.

https://github.com/litestar-org/litestar/blob/a26e1347503c9bc345027cbe02a4cc8adf5ea16d/litestar/handlers/base.py#L514

PR #2906, opened 2023-12-15
feat: add `scalar` schema ui

JacobCoffee:
- adds scalar docs
- adds test ci for codecov to see how it works (dont merge)
- renames function to less (?) confusing name for the dummies like me who thought it was a typo
- adds deprecation warning for ReDoc starting in 2.5.0 

### TODO

The Scalar Litestar-custom css should be (?) loaded as a string from its home in `openapi/styles/scalar.css`..
do you rely on the StaticFilesConfig here? 

Will have to see if we instantiate one even if a user doesnt pass the config into their Litestar app object.
If a user isnt using it, and we dont, should we generate one so that we can load this based on the static files url?


JacobCoffee:
Sonar issues will need fixing https://sonarcloud.io/project/issues?resolved=false&severities=MAJOR&sinceLeakPeriod=true&pullRequest=2906&id=litestar-org_litestar

JacobCoffee:
image

docs/usage/openapi.rst
peterschutt, line 1: docs for openapi will need a big overhaul in #2893 - so maybe we should even revert these and other docs changes until we know what that looks like.

litestar/openapi/config.py
```diff
@@ -108,7 +108,11 @@ class OpenAPIConfig:
     :class:`Reference <litestar.openapi.spec.reference.Reference>` objects.
     """
     root_schema_site: Literal["redoc", "swagger", "elements", "rapidoc"] = "redoc"
-    """The static schema generator to use for the "root" path of `/schema/`."""
+    """The static schema generator to use for the "root" path of `/schema/`.
+
+    .. deprecated:: 2.5.0
+        ReDoc is no longer the default schema generator. In 3.0, the default will be Scalar.
+    """
```
peterschutt, line 115: We also need to work out the deprecation strategy in #2893

litestar/openapi/plugins.py
```diff
@@ -301,6 +302,79 @@ def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
...
+                If js_url is provided, this is ignored.
+            js_url: Download url for the Scalar JS bundle.
+                If not provided, the version will be used to construct the url.
+            css_url: Download url for the Scalar CSS bundle.
+                If not provided, the Litestar-provided CSS will be used.
+            path: Path to serve the OpenAPI UI at.
+            **kwargs: Additional arguments to pass to the base class.
+        """
+        self.js_url = js_url or f"https://cdn.jsdelivr.net/npm/@scalar/api-reference@{version}"
+        self.css_url = css_url or "_styles/scalar.css"
```
peterschutt, line 333: We can probably leverage the `OpenAPIRenderPlugin.receive_router()` callback to register a handler that serves our custom css if another URL hasn't already been registered.

litestar/openapi/styles/scalar.css
peterschutt, line 1: Do we need to do anything nowadays to ensure this gets included in our package distributions?

peterschutt (commented):
Great stuff!

I think there will be a bit of evolution with the docs and deprecations in #2893.

litestar/openapi/styles/scalar.css
JacobCoffee: I was thinking the sdist include packaged that up as it was under `/litestar` but I may be confused at how that works?

JacobCoffee: Actually yeah im pretty sure its included, we do custom theming with our exception middleware

litestar/openapi/plugins.py
```diff
@@ -301,6 +303,106 @@ def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
...
+        Adds a route to serve the Scalar.com OpenAPI UI CSS.
+
+        Args:
+            router: The router that serves the OpenAPI UI.
+        """
+        router.register(
+            get("/_litestar-internal/scalar.css", media_type=MediaType.CSS, sync_to_thread=False)(
+                self.render_scalar_css
+            ),
+        )
```
peterschutt, line 391: I think we should only register this route if we have to, i.e., if user hasn't provided a custom css url

litestar/openapi/plugins.py
```diff
@@ -301,6 +303,106 @@ def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
...
+        Returns:
+            A rendered html string.
+        """
+        head = f"""
+                  <head>
+                    <title>{openapi_schema["info"]["title"]}</title>
+                    <style>{self.style}</style>
+                    <meta charset="utf-8"/>
+                    <meta name="viewport" content="width=device-width, initial-scale=1">
+                    <link rel="shortcut icon" href="{self.favicon}">
```
peterschutt, line 355:
```suggestion
                    {self.favicon}
```
`self.favicon` is already a html string.

litestar/openapi/plugins.py
JacobCoffee: Done

litestar/openapi/styles/scalar.css
JacobCoffee, line 1: this has made sonar angry, will work on fixing

litestar/openapi/plugins.py
JacobCoffee: [8f80bc4](https://github.com/litestar-org/litestar/pull/2906/commits/8f80bc40ececab879b7b0ec1cdea7d76df8b6ef9)

litestar/openapi/plugins.py
```diff
@@ -387,11 +387,12 @@ def receive_router(self, router: Router) -> None:
         Args:
             router: The router that serves the OpenAPI UI.
         """
-        router.register(
-            get("/_litestar-internal/scalar.css", media_type=MediaType.CSS, sync_to_thread=False)(
-                self.render_scalar_css
-            ),
-        )
+        if self.css_url == "/schema/_litestar-internal/scalar.css":
```
peterschutt, line 390: prob best to define the default url as a class-level constant.

JacobCoffee: how does [06b54c5](https://github.com/litestar-org/litestar/pull/2906/commits/06b54c532d70e30d5d54a38ca7673785d34afa77) look?

peterschutt: :ok_hand:

JacobCoffee:
Seems like the remaining issues revolve around code smells from the base PR and testing (TBD first in #2893)

litestar/openapi/styles/scalar.css
peterschutt, line 1: What about putting this with our other branding materials and set the default URL like we do for the favicon?

JacobCoffee: Wouldn't we hit firewall issues for users that install this? I would personally I think.

peterschutt: Idk, would we? Is that not the same with the favicon then?

peterschutt: Is it an issue specifically with https://raw.githubusercontent.com? And the downloads from https://unpkg.com aren't affected?

JacobCoffee: I guess that makes sense. We can move it

JacobCoffee:
Noticing that RapiDoc has no customizable CSS url yet. May want to add that in to #2893 alongside what was added in [7c6cc9e](https://github.com/litestar-org/litestar/pull/2906/commits/7c6cc9e6e44de60ccbe47b7d1fb2590b80b717fd)

JacobCoffee:
Just waiting on the `openapi-router` work.. maybe this could be merged as is now and worked on alongside the rest of the stuff (@peterschutt ?)

PR #2909, opened 2023-12-15
fix: coverage report uploads

peterschutt:
Change in actions/upload-artifact@v4 broke our workflow by not allowing a 2nd file to be added to an existing artifact.

This PR adds each coverage report to a distinct artifact and pulls down those specific artifacts for combining coverage.

A negative of this approach is we hard code the artifact names with python version in the sonar job.


peterschutt:
This just sucks.

PR #2913, opened 2023-12-17
Align layered annotations

gsakkis:
Make (more) consistent the type annotations of the different app layers (Litestar, Router, Controller, router handler decorators), plus two minor changes for two AppConfig fields


litestar/config/app.py
```diff
@@ -82,7 +83,7 @@ class AppConfig:
     """A sync or async function called after the response has been awaited. It receives the
     :class:`Request <.connection.Request>` object and should not return any values.
     """
-    allowed_hosts: list[str] | AllowedHostsConfig | None = field(default=None)
+    allowed_hosts: AllowedHostsConfig | None = field(default=None)
```
provinzkraut, line 86: This is a breaking change. Users might want to pass a list directly into the config and plugins may receive this config object and expect a `list[str]` here, which they can append to

litestar/config/app.py
```diff
@@ -207,16 +208,7 @@ class AppConfig:
...
-
-    def __post_init__(self) -> None:
-        """Normalize the allowed hosts to be a config or None.
-
-        Returns:
-            Optional config.
-        """
-        if self.allowed_hosts and isinstance(self.allowed_hosts, list):
-            self.allowed_hosts = AllowedHostsConfig(allowed_hosts=self.allowed_hosts)
+    experimental_features: frozenset[ExperimentalFeatures] = frozenset()
```
provinzkraut, line 211: Also a breaking change for the reasons stated above

litestar/controller.py
```diff
@@ -143,7 +143,7 @@ class Controller:
     """A sequence of string tags that will be appended to the schema of all route handlers under the controller."""
     security: Sequence[SecurityRequirement] | None
     """A sequence of dictionaries that to the schema of all route handlers under the controller."""
-    signature_namespace: dict[str, Any]
+    signature_namespace: Mapping[str, Any]
```
provinzkraut, line 146: Breaking as well if users rely on this being a dict

provinzkraut (changes requested):

litestar/config/app.py
gsakkis: Ah I thought `AppConfig` was internal (or at least not supposed to be instantiated by the user). 
Reverted.

litestar/config/app.py
gsakkis: Reverted

litestar/controller.py
gsakkis: Reverted

litestar/config/app.py
provinzkraut: You'll usually instantiate it directly when using `Litestar.from_config`

litestar/controller.py
```diff
@@ -127,7 +127,7 @@ class Controller:
 
     All route handlers under the controller will have the fragment appended to them. If not set it defaults to ``/``.
     """
-    response_class: type[Response] | None
+    response_class: ResponseType | None
```
peterschutt, line 130: I think we should standardize on this in the other direction, that is to prefer `type[Response]`.

Reading `type[Response]` is as clear as you can get, but reading `ResponseType` not so much as one doesn't know if it accepts a union of types, or is optional or anything until inspecting the actual alias definition.

For other composite types where we accept complex unions of things there is a benefit of not having to repeat annotations like `HTTPRouteHandler | WebsocketRouteHandler | ASGIRouteHandler` again and again, but here we are talking about 2 extra characters.

Absolutely not a critique of your work, as I know you are just making this consistent with what we are doing elsewhere, but a good time to discuss.

peterschutt (commented):
This is a nice tidy up, thanks very much!

1 discussion point from me before signing off.

litestar/controller.py
gsakkis: Makes sense, I'm fine with dropping the `ResponseType` alias altogether (fyi it touches 6 more files than the PR as of now).

litestar/config/app.py
peterschutt: Also, `AppConfig` is very public API for plugin authors.

litestar/controller.py
peterschutt: > fyi it touches 6 more files than the PR as of now

I'm OK with that.

I had a quick look through the other type aliases that we have, if there were others that were equally redundant then I'd suggest we do it as a separate effort, but this appears to be the only one.

(there are others that I'd question whether they are complex enough, but they are at least more complex than `ThingType = type[Thing]`)

Thanks!

JacobCoffee (approved):

peterschutt (approved):

provinzkraut:
I think this should go into `develop`, as it's more of a feature than a fix and should be released with the next minor and not as a patch. Aside from that, looks good :)

provinzkraut (approved):

Issue #2862, opened 2023-12-09
Bug: Litestar changing root logger by default


arthurazs:
### Description

Litestar changes the root logger configuration by default.

My current workaround is to set Litestar root logger to an empty dict:

`app = Litestar(logging_config=LoggingConfig(root={}))`

### MCVE

```python
import logging

from litestar import Litestar


logging.basicConfig(level=logging.INFO, format="CRAZY LOG %(message)s")
logger = logging.getLogger(__name__)

logger.info("Logging from lite_app before creating the litestar app")
app = Litestar()
logger.info("Logging from lite_app after creating the litestar app")
```

### Steps to reproduce

Execute above code, expected output:

```bash
CRAZY LOG Logging from lite_app before creating the litestar app
CRAZY LOG Logging from lite_app after creating the litestar app
```

Current output:

```
CRAZY LOG Logging from lite_app before creating the litestar app
INFO - 2023-12-08 20:40:08,708 - __main__ - example - Logging from lite_app after creating the litestar app
```

### Litestar Version

2.3.2

### Platform

- [X] Linux
- [ ] Mac
- [X] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2862">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2862/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/2862/pledge.svg">
</picture>
</a>

EM51641:
I have the exact same problem.

arthurazs:
> I have the exact same problem.

While we wait for the fix, my workaround is `app = Litestar(logging_config=LoggingConfig(root={}))`

provinzkraut:
Closed by #2969.

PR #2919, opened 2023-12-20
fix: unintended error message from `get_logger_placeholder()`

cbscsm:
### Background
- Currently, `Litestar.get_logger` is `get_logger_placeholder` only when the app initialized by keyword argument `logging_config=None`.
  - We using this keyword argument as workaround for an issue like #2862
- `request.logger` is a property that returns `self.app.get_logger()`. It calls `Litestar.get_logger` with no argument. In standard library or picologging, it is allowed and will return root logger.
- code: https://github.com/litestar-org/litestar/blob/d8322b7a588a0a272d8b611d7d0c6ab1d2ccc326/litestar/connection/base.py#L268-L278
- But `get_logger_placeholder` doesn't allow calling it with no argument. So, when you initialize app with no logging_config and try to use `request.logger`, the error is not as intended.
- Actual
```text
TypeError: get_logger_placeholder() missing 1 required positional argument: '_'
```
- Expected
```text
ImproperlyConfiguredException: cannot call '.get_logger' without passing 'logging_config' to the Litestar constructor first
```

### Changes suggested
- So I suggest to add default value `None`.

### DIFF
--- a/tests/unit/test_logging/test_logging_config.py
+++ b/tests/unit/test_logging/test_logging_config.py
@@ -6,6 +6,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from litestar import Request, get
+from litestar.exceptions import ImproperlyConfiguredException
 from litestar.logging.config import LoggingConfig, _get_default_handlers, default_handlers, default_picologging_handlers
 from litestar.logging.picologging import QueueListenerHandler as PicologgingQueueListenerHandler
 from litestar.logging.standard import QueueListenerHandler as StandardQueueListenerHandler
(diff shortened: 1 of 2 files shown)


guacs:
@cbscsm thanks for this! I think this makes sense since it's the same for the standard library `getLogger` function as well. Could you add a test for this as well?

cbscsm:
I added a test case to check if right exception raised when get_logger_placeholder called without args.

tests/unit/test_logging/test_logging_config.py
```diff
@@ -82,6 +83,12 @@ def test_picologging_dictconfig_when_disabled(dict_config_mock: Mock) -> None:
         assert not dict_config_mock.called
 
 
+def test_get_logger_without_logging_config() -> None:
+    with create_test_client(logging_config=None) as client:
+        with pytest.raises(ImproperlyConfiguredException):
```
guacs, line 88: Could you add a `match` to make sure the exception message is the logging one? Just to confirm that the error is from the placeholder function.

guacs (changes requested):

guacs (approved):
Thank you!

guacs:
@all-contributors add @cbscsm for code, test

PR #2921, opened 2023-12-21
fix: message where missing dependency name doesn't match install extra

horo-fox:
PR adds a 3rd argument, `extra: str | None = None` to `MissingDependencyException` to handle case where Litestar's install extra is not named the same as the required package that the extra will install.


peterschutt:
Looking at this test:

https://github.com/litestar-org/litestar/blob/cf25db624d8fb266f4d06699e5a0b36c69c79bc1/tests/unit/test_exceptions.py#L148-L156

... it appears that adding the additional `extra=...` argument doesn't offer any more than is already available to us (even if the naming of the `install_package` isn't totally obvious).

So I think this PR can simply be a 1-liner:

```py
raise MissingDependencyException("jinja2", "jinja")
```

horo-fox:
Well, `pip install jinja` isn't the same as `pip install jinja2`, but the extra is named `jinja`. So the extra name doesn't match the install name, though I assume normally it does match. I guess an alternative to adding an `extra=...` would be to add a new `jinja2` extra?

peterschutt:
> Well, `pip install jinja` isn't the same as `pip install jinja2`, but the extra is named `jinja`. So the extra name doesn't match the install name, though I assume normally it does match. I guess an alternative to adding an `extra=...` would be to add a new `jinja2` extra?

I agree with the problem you have identified. I don't agree that we need to add the `extra` key - this seems to be exactly what `install_package` is intended for.

peterschutt:
Sorry - I see it now, b/c it also uses `install_package` in the `pip install ...` section - sorry you are correct.

litestar/contrib/prometheus/config.py
```diff
@@ -15,7 +15,7 @@
 try:
     import prometheus_client  # noqa: F401
 except ImportError as e:
-    raise MissingDependencyException("prometheus_client") from e
+    raise MissingDependencyException("prometheus_client", extra="prometheus") from e
```
peterschutt, line 18: should this be `"prometheus-client"`?

https://pypi.org/project/prometheus-client/

peterschutt: `"prometheus"` points to https://github.com/slok/prometheus-python which has been archived and no updates for 8 years.

litestar/exceptions/base_exceptions.py
```diff
@@ -41,10 +41,10 @@ class MissingDependencyException(LitestarException, ImportError):
     This exception is raised only when a module depends on a dependency that has not been installed.
     """
 
-    def __init__(self, package: str, install_package: str | None = None) -> None:
+    def __init__(self, package: str, install_package: str | None = None, extra: str | None = None) -> None:
         super().__init__(
             f"Package {package!r} is not installed but required. You can install it by running "
-            f"'pip install litestar[{install_package or package}]' to install litestar with the required extra "
+            f"'pip install litestar[{extra or install_package or package}]' to install litestar with the required extra "
```
peterschutt, line 47: Can you add a test for the new parameter please?

litestar/contrib/prometheus/config.py
horo-fox: The `pip install` bit should use `prometheus_client` (which *should* be an alias to `prometheus-client` IIRC because pypi? might be worth having `install_package="prometheus-client"` though for clarity's sake.) -- `extra` is specifically just for the extra recommendation and doesn't change the package name being recommended.

peterschutt: Oh man - I'm getting all tangled up on this. Agree, `"prometheus"` for the `extra=...` and `"prometheus-client"` for the `install_package=...`.

horo-fox:
Sorry about the delay! This should address everything.

peterschutt (approved):
Thanks!

peterschutt:
@all-contributors add @horo-fox for bug

PR #2923, opened 2023-12-22
refactor: replace unnecessary use of `default_factory` with `default`

wer153:
PR replaces unnecessary use of `default_factory` with `default` for dataclass fields on the `LoggingMiddlewareConfig` where the default value returned by the factory requires no computation and is immutable.

### Description
As far as I understand, By specifying a factory function for a mutable default value, each instance gets its own independent copy of the default value, `preventing unintended sharing of mutable state between instances`. This helps in maintaining encapsulation and avoiding potential side effects that could arise from shared mutable state.

However, in the code, `default_factory was not used for such a purpose`. Therefore, I suggest `using default to avoid misinterpretation of usage`.

### DIFF
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -271,7 +271,7 @@ class LoggingMiddlewareConfig:
     response_log_message: str = field(default="HTTP Response")
     """Log message to prepend when logging a response."""
     request_log_fields: Iterable[RequestExtractorField] = field(
-        default_factory=lambda: (
+        default=(
             "path",
             "method",
             "content_type",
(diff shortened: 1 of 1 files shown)


litestar/middleware/logging.py
```diff
@@ -305,7 +305,7 @@ class LoggingMiddlewareConfig:
             Thus, re-arranging the log-message is as simple as changing the iterable.
         -  To turn off logging of responses, use and empty iterable.
     """
-    middleware_class: type[LoggingMiddleware] = field(default_factory=lambda: LoggingMiddleware)
+    middleware_class: type[LoggingMiddleware] = LoggingMiddleware
```
cbscsm, line 308: How about use `field` function for consistency.
```suggestion
    middleware_class: type[LoggingMiddleware] = field(default=LoggingMiddleware)
```

wer153: Makes sense. Thanks for the suggestion!

peterschutt (approved):
LGTM, thanks!

wer153:
https://github.com/all-contributors add @wer153 for code

wer153:
@all-contributors add @wer153 for code

PR #2927, opened 2023-12-22
refactor: use value_or_default instead of conditional expression

cbscsm:
- There is a raw conditional expression to check `Empty` with default value. And I found an utility function `litestar.utils.empty.value_or_default` that does same thing. So I suggest to replace the conditional expression with the function: https://github.com/litestar-org/litestar/blob/8153d5b8b53d3a703f86dae90e765f56f5f4bac8/litestar/utils/empty.py#L14-L26


peterschutt (approved):
Thanks!

PR #2934, opened 2023-12-30
docs: add a missing code block

jseop-lim:
## Description
Add missing code block wrapping to [docs](https://docs.litestar.dev/2/usage/dependency-injection.html#pre-requisites-and-scope). Among the layers in which dependencies can be defined, only _router dependencies_ is written in plain text, not in code blocks.


docs/usage/dependency-injection.rst
```diff
@@ -80,8 +80,8 @@ The pre-requisites for dependency injection are these:
 
 What is *scope* in this context? Dependencies are **isolated** to the context in which they are declared. Thus, in the
 above example, the ``local_dependency`` can only be accessed within the specific route handler on which it was declared;
-The ``controller_dependency`` is available only for route handlers on that specific controller; And the router
-dependencies are available only to the route handlers registered on that particular router. Only the ``app_dependencies``
+The ``controller_dependency`` is available only for route handlers on that specific controller; And the ``router
+dependencies`` are available only to the route handlers registered on that particular router. Only the ``app_dependencies``
```
Alc-Alc, line 84: If this change is being done, wouldn't it make sense to use

```
``router_dependency`` is available only to the route handlers registered on that particular router. Only the 
```

Considering the other two are variables you can use in the code?

jseop-lim: That's a good point! It's been reflected.

Alc-Alc (approved):
🙂

provinzkraut (approved):

wer153:
Welcome!

PR #2937, opened 2023-12-30
feat(core): Replace `anyio.to_thread.run_sync` with native versions

provinzkraut:
- Replace `anyio.to_thread.run_sync` with native versions of supported libraries, i.e. [`asyncio.loop.run_in_executor`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) and [`trio.to_thread.run_sync`](https://trio.readthedocs.io/en/stable/reference-core.html#trio.to_thread.run_sync)
- Add `set_asyncio_executor` and `get_asyncio_executor` utility functions to configure the executor used when running synchronous functions
- Add `set_trio_capacity_limiter` and `get_trio_capacity_limiter` utility functions to configure the `trio.CapacityLimiter` used when running synchronous functions

<hr>

Reasons for this change are twofold:

1. Performance improvements. Anyio has some [known performance overhead](https://github.com/agronholm/anyio/discussions/543), and its threading solution proved to be slower than both asyncio's and trio's built-in functionalities
2. Configurability of the thread pools used under the hood. Prior to this change the only way a user could influence these was to [set a global default in anyio](https://anyio.readthedocs.io/en/stable/threads.html#adjusting-the-default-maximum-worker-thread-count)

<hr>

Using this very simple benchmark we can see the stark difference in performance between anyio and native asyncio:

```python
import asyncio
import time
from concurrent.futures import ThreadPoolExecutor
from timeit import timeit

from anyio.to_thread import run_sync

from litestar.concurrency import sync_to_thread, set_asyncio_executor

set_asyncio_executor(ThreadPoolExecutor(40))


def func():
    time.sleep(0)


async def run_anyio():
    await run_sync(func)


async def run_native():
    await sync_to_thread(func)


async def run_anyio_batch():
    await asyncio.gather(*[run_sync(func) for _ in range(100)])


async def run_native_batch():
    await asyncio.gather(*[sync_to_thread(func) for _ in range(100)])


def run(fn):
    asyncio.new_event_loop().run_until_complete(fn())


print(timeit("run(run_anyio)", globals=globals(), number=1000))
print(timeit("run(run_native)", globals=globals(), number=1000))

print(timeit("run(run_anyio_batch)", globals=globals(), number=100))
print(timeit("run(run_native_batch)", globals=globals(), number=100))
```

```shell
0.6600718750014494
0.19882075799978338
2.053338443998655
0.7532648650012561
``` 

And testing this in a simple application, using `wrk`:

```python
from litestar import get, Litestar

@get(sync_to_thread=True)
def handler() -> str:
    return "Hello, world!"

app = Litestar(route_handlers=[handler])
```

*Without this PR*

```text
Thread Stats    Avg      Stdev     Max     +/- Stdev
Latency       23.83ms   14.72ms  266.88ms   98.07%
Req/Sec        2.22k    217.67    2.39k     95.00%

22114 requests in 5.01s, 3.10MB read
Requests/sec:   4415.57
Transfer/sec:   633.88KB
```

*With the changes applied*

```text
Thread Stats    Avg      Stdev     Max     +/- Stdev
Latency       17.14ms   13.38ms  254.68ms   96.68%
Req/Sec        3.14k     397.07   3.66k     83.00%

31234 requests in 5.01s, 4.38MB read
Requests/sec:   6236.19
Transfer/sec:   0.87MB
```

### DIFF
--- a/docs/reference/concurrency.rst
+++ b/docs/reference/concurrency.rst
@@ -1,0 +1,5 @@
+cli
+===
+
+.. automodule:: litestar.concurrency
+    :members:
--- a/litestar/concurrency.py
+++ b/litestar/concurrency.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+
+import asyncio
+import contextvars
+from functools import partial
+from typing import TYPE_CHECKING, Callable, TypeVar
+
+import sniffio
+from typing_extensions import ParamSpec
+
+if TYPE_CHECKING:
+    from concurrent.futures import ThreadPoolExecutor
+
+    import trio
+
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+__all__ = (
+    "sync_to_thread",
+    "set_asyncio_executor",
+    "get_asyncio_executor",
+    "set_trio_capacity_limiter",
+    "get_trio_capacity_limiter",
+)
+
+
+class _State:
+    EXECUTOR: ThreadPoolExecutor | None = None
+    LIMITER: trio.CapacityLimiter | None = None
+
+
+async def _run_sync_asyncio(fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T:
+    ctx = contextvars.copy_context()
+    bound_fn = partial(ctx.run, fn, *args, **kwargs)
+    return await asyncio.get_running_loop().run_in_executor(get_asyncio_executor(), bound_fn)  # pyright: ignore
+
+
+async def _run_sync_trio(fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T:
+    import trio
+
+    return await trio.to_thread.run_sync(partial(fn, *args, **kwargs), limiter=get_trio_capacity_limiter())
+
+
+async def sync_to_thread(fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs) -> T:
+    """Run the synchronous callable ``fn`` asynchronously in a worker thread.
+
+    When called from asyncio, uses :meth:`asyncio.AbstractEventLoop.run_in_executor` to
+    run the callable. No executor is specified by default so the current loop's executor
+    is used. A specific executor can be set using
+    :func:`~litestar.concurrency.set_asyncio_executor`. This does not affect the loop's
+    default executor.
+
+    When called from trio, uses :meth:`trio.to_thread.run_sync` to run the callable. No
+    capacity limiter is specified by default, but one can be set using
+    :func:`~litestar.concurrency.set_trio_capacity_limiter`. This does not affect trio's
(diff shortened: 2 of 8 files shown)


cofin (approved):
Nice work.  LGTM

provinzkraut:
Two more things to consider regarding anyio would be file operations (which we're not using that frequently) and `TaskGroup`s, which we *do* use in quite a few places. A quick benchmarks shows the native `asyncio.TaskGroup` to be about twice as performant as the one provided by anyio. This is not an entirely fair comparison though as the anyio's task group does a lot more. It would be possible to provide a backport for the native `TaskGroup` for Python < 3.11 (it's not a lot of code actually) and only fall back on anyio if we're not running on asyncio. 

This poses a general question: Do we want to use anyio for the features it provides, because we want to make use of its structured concurrency approach, or just as a shim to provide asyncio/trio compatibility?

Our trio compatibility is spotty in general and not particularly well tested and dropping it has come up a few times now. On that front it should also be noted that currently the only ASGI server I'm aware of that supports trio is hypercorn, and major libraries like redis or SQLAlchemy don't support it, so the potential audience for it is limited to begin with.

Personally, I'm not too keen to start introducing our own compatibility layer between asyncio/trio, but as it stands, we're not really using anyio for anything more than that aside from a few utilities we could easily implement ourselves. I think if we want to keep anyio, we should actually embrace it and the structured concurrency approach, and if we're not going that route, it makes sense to fall back on pure asyncio to gain some performance, simplify our implementation and remove external dependencies.

@litestar-org/maintainers

FHU-yezi:
I think we can remove `anyio` as welll as our `trio` support in next major version, because the community have poor support in developing API with trio.

euri10 (approved):
the PR lgtm, on the following question:

> This poses a general question: Do we want to use anyio for the features it provides, because we want to make use of its structured concurrency approach, or just as a shim to provide asyncio/trio compatibility?

speaking for myself, I never use trio so not really interested in the asyncio/trio compatibility b but structured concurrency approach is defintely a plus.
would quattro help ? (https://github.com/Tinche/quattro)

JacobCoffee (approved):

provinzkraut:
> > This poses a general question: Do we want to use anyio for the features it provides, because we want to make use of its structured concurrency approach, or just as a shim to provide asyncio/trio compatibility?
> 
> speaking for myself, I never use trio so not really interested in the asyncio/trio compatibility b but structured concurrency approach is defintely a plus. would quattro help ? (https://github.com/Tinche/quattro)

Interesting, hadn't seen that library yet! Looking at it, it seems to be just a few hundred lines of code, half of which is the backport of `asyncio.TaskGroup`. Maybe it's worth considering to just implement what we need ourselves!

PR #2941, opened 2024-01-03
fix: #2867 - Ensure `type_encoders` are passed to exception response

### LINKED ISSUE #2867: Bug: Long cryptic error when excluding a required field from a DTO
### Description

As mentioned on Discord, excluding a required field from a DTO triggers a 500 Internal Server Error at runtime. This is hard to troubleshoot even at debug mode because the traceback is too long and ends with `litestar.exceptions.base_exceptions.SerializationException: Unsupported type: <class '...'>`, which is rather misleading

_No response_

### MCVE

```python
from typing import Annotated

class Parent(BaseModel):
    id: int


class Child(BaseModel):
    name: str
    parent: Parent


ChildDTO1 = PydanticDTO[Annotated[Child, DTOConfig(exclude={"name"})]]


class ChildDTO2(PydanticDTO[Child]):
    config = DTOConfig(exclude={"name"})


@post(path="/child1", dto=ChildDTO1, return_dto=None)
async def post_child1(data: Child) -> Child:
    return data


@post(path="/child2", dto=ChildDTO2, return_dto=None)
async def post_child2(data: Child) -> Child:
    return data


app = Litestar(route_handlers=[post_child1, post_child2])
```

### Steps to reproduce

```bash
1. Start the server
2. Run `curl -X POST http://localhost:8508/child1 -d '{"parent": {"id": 1}}'` (same with `/child2`)
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.4.3

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
Fix a bug that would cause a `SerializationException` when custom types were present within a `ValidationException`, which would be the case if a Pydantic error was passed verbatim. 

The fix is to pass along the `type_encoders` resolved from the route handler of the originating request.

Fixes #2867.

### DIFF
--- a/litestar/middleware/exceptions/_debug_response.py
+++ b/litestar/middleware/exceptions/_debug_response.py
@@ -191,4 +191,9 @@ def create_debug_response(request: Request, exc: Exception) -> Response:
         content = create_plain_text_response_content(exc)
         media_type = MediaType.TEXT
 
-    return Response(content=content, media_type=media_type, status_code=HTTP_500_INTERNAL_SERVER_ERROR)
+    return Response(
+        content=content,
+        media_type=media_type,
+        status_code=HTTP_500_INTERNAL_SERVER_ERROR,
+        type_encoders=request.app.type_encoders,
+    )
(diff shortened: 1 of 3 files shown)


JacobCoffee (approved):

cofin (approved):
LGTM

PR #2942, opened 2024-01-04
fix: ensure PDM is available in the Makefile Path

ashm-tech:
Greetings to you!)

I made small changes to the Makefile by updating the PDM installation function, because without this it did not work normally and calling the Makefile ended with an error calling the newly installed PDM


provinzkraut:
@JacobCoffee

Makefile
```diff
@@ -35,6 +35,7 @@ install-pdm: 										## Install latest version of PDM
 	curl -sSL https://pdm.fming.dev/install-pdm.py.sha256 | shasum -a 256 -c - && \
 	python3 install-pdm.py && \
 	rm install-pdm.py
+	export PATH=/home/$USER/.local/bin:$PATH
```
guacs, line 38: Should we be editing the `PATH` ourselves? Maybe it might be better to just add in the contributing docs that one may need to add `$HOME/.local/bin` to `PATH`.

ashm-tech: So this command only updates the PATH so that make can see that PDM has appeared in the system. In addition, PDM itself recommends executing this command after installation

guacs: Yes, but there too PDM itself doesn't modify `PATH` even though it could as part of its installation script. Users may not want to add it to the path, but rather symlink or something along those lines. Usually most of these don't modify the `PATH` variables, but just document that it should be done.

ashm-tech: PDM changes but does not export the new PATH. And to whom, we add this script only to the part that installs PDM if it is not installed on the machine, so there should be no problems with other names, I think

cofin: Do we actually have to export the variable?

ashm-tech: Yes, because if you do not export the updated PATH variable, the `make install` command will fail after installing PDM

cofin: I don't think this configuration will work for users that install in Windows or OSX.  Is there another way we can get PDM into the path?

ashm-tech: > I don't think this configuration will work for users that install in Windows or OSX. Is there another way we can get PDM into the path?

Yes, I really did not take this fact into account. Now I will try to fix it and update the script

provinzkraut:
@JacobCoffee can you take a look at this one?

Makefile
JacobCoffee: PDM installs into different places depending
```
    $HOME/.local/bin for Unix
    $HOME/Library/Python/<version>/bin for MacOS
    %APPDATA%\Python\Scripts on Windows
```

I think we should just document this as @guacs said as it works ootb for the majority and doing anything else might be undesirable for some and introduce complexity or weirdness trying to shoehorn things into *everyones* environments.

guacs:
@ashm-tech based on what @JacobCoffee said as well, I think we're going to close this one. If you'd like to raise a PR to document that the PDM path might need to be added to PATH, that'd be much appreciated :)

Issue #2151, opened 2023-08-12
Bug: Using `structlog.dev.ConsoleRenderer` causes `TypeError: can only concatenate str (not "bytes") to str`


sykloid:
### Description

Adding an instance of `structlog.dev.ConsoleRenderer` to the `processors` parameter of a `StructLoggingConfig` causes an exception which doesn't appear if said processor is not present (the correct message is logged as JSON to stdout).

I've included one instance of the exception traceback but the exception is thrown multiple times, I believe this is because `log_exceptions="always"` causes a subsequent log of the first exception, which causes another log, etc.

### MCVE

```python
# Almost verbatim from https://docs.litestar.dev/2/usage/logging.html#using-structlog

from litestar import Litestar, Request, get
from litestar.logging import StructLoggingConfig
import structlog


@get("/")
async def my_router_handler(request: Request) -> None:
    request.logger.info("inside a request")
    return None


# Only change to base example, added two parameters to config.
logging_config = StructLoggingConfig(processors=[structlog.dev.ConsoleRenderer()], log_exceptions="always")

app = Litestar(route_handlers=[my_router_handler], logging_config=logging_config)
```

### Steps to reproduce

```bash
1. `litestar run`
2. `GET /`
```

### Logs

```bash
Traceback (most recent call last):
  File "/home/sykloid/src/scratch/litestar_test/.venv/lib/python3.11/site-packages/litestar/middleware/exceptions/middleware.py", line 159, in __call__
    await self.app(scope, receive, send)
  File "/home/sykloid/src/scratch/litestar_test/.venv/lib/python3.11/site-packages/litestar/routes/http.py", line 79, in handle
    response = await self._get_response_for_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sykloid/src/scratch/litestar_test/.venv/lib/python3.11/site-packages/litestar/routes/http.py", line 131, in _get_response_for_request
    response = await self._call_handler_function(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sykloid/src/scratch/litestar_test/.venv/lib/python3.11/site-packages/litestar/routes/http.py", line 160, in _call_handler_function
    response_data, cleanup_group = await self._get_response_data(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #1672, opened 2023-05-13
Bug: debug mode doesn't work with structlog 


doshitan:
### Description

Using the default `StructLoggingConfig` for logging causes the application to fail to start with `litestar run --debug`.

### MCVE

```python
from litestar import Litestar, get
from litestar.logging import StructLoggingConfig


@get("/")
def hello_world() -> dict[str, str]:
    """Keeping the tradition alive with hello world."""
    return {"hello": "world"}

app = Litestar(
    route_handlers=[hello_world],
    logging_config=StructLoggingConfig(),
)
```

### Steps to reproduce

```bash
1. Attempt to run MCVE with `litestar run --debug`
```

### Logs

```bash
litestar run --debug
Traceback (most recent call last):
  File "/nix/store/fwh99risihjp36ypf0hjsf2lj37qiq4s-python3.11-litestar-2.0.0a6/bin/.litestar-wrapped", line 6, in <module>
    from litestar import __main__
  File "/nix/store/fwh99risihjp36ypf0hjsf2lj37qiq4s-python3.11-litestar-2.0.0a6/lib/python3.11/site-packages/litestar/__main__.py", line 3, in <module>
    litestar_group()  # pragma: no cover
    ^^^^^^^^^^^^^^^^
  File "/nix/store/ahk8gvfqrj94qg6ndvg2akn74rhp1xdz-python3.11-click-8.1.3/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/ahk8gvfqrj94qg6ndvg2akn74rhp1xdz-python3.11-click-8.1.3/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/nix/store/ahk8gvfqrj94qg6ndvg2akn74rhp1xdz-python3.11-click-8.1.3/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/ahk8gvfqrj94qg6ndvg2akn74rhp1xdz-python3.11-click-8.1.3/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
[...]

PR #2943, opened 2024-01-04
feat: `structlog` plugin & bug fixes

cofin:
This implements a `StructlogPlugin` class to make it easier to configure logging in a single place.

Fixes a bug when caused when calling `app.debug`.  In some cases, the structlog configuration will only be initalized lazily.  This abstraction does not have the `setLevel` function.

Details:
- Detect of `logger` has `setLevel` before calling
- Set even message name to be init-cap
- Add `set_level` interface to config 
- Allow structlog printer to detect if console is TTY enabled.  If so, use a Struglog color formatter with Rich traceback printer.
- Auto-configure stdlib logger to use the structlog logger
- Adds `StructlogPlugin`


provinzkraut:
@cofin I think you need to rebase the changes onto `develop` :eyes:

litestar/plugins/structlog.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+from litestar.cli._utils import console
```
provinzkraut, line 6: Console is only available when `rich` is installed, not sure this should be used here.

cofin: Isn't Rich included by default now?

provinzkraut: Oh. Yes. Forgot about that :grimacing:

provinzkraut:
@cofin This shoud fix #2151 and #1672, right?

provinzkraut (approved):

PR #2966, opened 2024-01-08
feat: add `SecretString` and `SecretBytes` datastructures

### LINKED ISSUE #1312: Enhancement: Add Secret
Starlite-specific implementation of [Starlette's `datastructures.Secret`](https://github.com/encode/starlette/blob/ada845cc8ad0b531b16749b9bf3a57114c60c679/starlette/datastructures.py#L206-L223).

```py
#  Starlette's implementation of the `Secret`  class in `starlette.datastructures`
class Secret:
    """
    Holds a string value that should not be revealed in tracebacks etc.
    You should cast the value to `str` at the point it is required.
    """

    def __init__(self, value: str):
        self._value = value

    def __repr__(self) -> str:
        class_name = self.__class__.__name__
        return f"{class_name}('**********')"

    def __str__(self) -> str:
        return self._value

    def __bool__(self) -> bool:
        return bool(self._value)
```

**Additional context**

Additionally, with our recent  efforts to generalize away from Pydantic (#1306, etc.) we may want to implement Pydantics `SecretStr` and `SecretBytes` https://docs.pydantic.dev/usage/types/#secret-types

<a href="https://polar.sh/litestar-org/litestar/issues/1312">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/1312/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

sashahx:
This pull request implements `SecretString` and `SecretBytes` data structures to hide sensitive data in tracebacks, etc.

Closes #1312


provinzkraut:
Thanks @sashahx!

I’ll give this a proper review soon but just a bite upfront: 
The reason for having those was so that they could be de/encoded from/to their secret/raw value counterparts. So for them to be useful, they’ll have to be added to the default de/encoders.

sashahx:
Hey there!

No worries on the speedy review—I'll get cracking on adding that functionality you mentioned. I'll shoot you a message once it's good to go, so you can check it out when you have a moment

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
...
+        """Checks if the given object is equal to the current instance.
+
+        Args:
+            other (Any): The object to compare.
+
+        Returns:
+            bool: True if equal, False otherwise.
+        """
+
+        return isinstance(other, self.__class__) and self.get_value() == other._secret_value
```
cbscsm, line 72:
```suggestion
    def __eq__(self, other: object) -> bool:
        """Checks if the given object is equal to the current instance.

        Args:
            other: The object to compare.

        Returns:
            bool: True if equal, False otherwise.
        """

        return isinstance(other, self.__class__) and self.get_value() == other._secret_value
```

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
+from typing import Any, Union
```
cbscsm, line 1:
```suggestion
from typing import Union
```

cbscsm (commented):
I think `object` can replace `Any`. `object` means "any object", but `Any` means "no type constraint for it".

For example:
```python
from typing import Any

def foo(a: Any, b: object):
    a[5]  # Type check pass
    b[5]  # Type check error(No __getitem__ is defined in object)
    if isinstance(b, list):  # object can be narrowed into list because object is the top type
        b[5]  # Type check pass
```

litestar/datastructures/secret_values.py
JacobCoffee, line 1: We can replace with `from __future__ import annotations` to relieve some future maintenance burden here?

And replace subsequent usages of Union e.g., `secret_value: str | bytes`

litestar/datastructures/secret_values.py
sashahx: Done

litestar/datastructures/secret_values.py
sashahx: Done

peterschutt (changes requested):
Clean and well structured code, thanks!

I've made a few suggestions in review. Additionally:
- As mentioned by @provinzkraut, we need to have the default encoders/decoders be able to handle this type.
- It would be nice to see some integration tests that show patterns of expected use, e.g., as a header type, and as a field on a data model, and validate its behavior with logging and inside exception rendering, etc.

I think this will be a valuable feature, thanks for doing the work!

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
+from __future__ import annotations
+
+
+class SecretValue:
```
peterschutt, line 4: I also think this could be better if it was generic on the type of the secret value:

```py
_SecretT = TypeVar("_SecretT", bound="str | bytes")


class SecretValue(Generic[_SecretT]):
    def get_value(self) -> _SecretT: ...
    ...

class SecretString(SecretValue[str]): ...
```
This way, calling `get_value()` and other methods on a sub class won't by typed as returning the union `str | bytes`, instead it will be narrowed to the type that the subtype has been narrowed with.

peterschutt (approved):
LGTM - thanks @sashahx!

peterschutt (changes requested):
Just the issue of configuring the encoders/decoders to handle the type - do you need any assistance with this?

sashahx:
@peterschutt this week, I'll be tackling the default encoders/decoders. Before I begin, I'll research other data structures to get a better grasp of the topic. At the moment, I don't need assistance, but I might have questions later on

Once I've implemented the encoders/decoders, I'll shift focus to developing integration tests

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
...
+    """Represents a secret value that can be of type `str` or `bytes`."""
+
+    def __init__(self, secret_value: str | bytes):
```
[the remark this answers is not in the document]
sashahx: Done

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
...
+        """Hides the secret value.
+
+        Args:
+            value (str | bytes): The secret value to be hidden.
+
+        Returns:
+            str: The hidden representation of the secret value.
+        """
+
+        return "******" if value else ""
```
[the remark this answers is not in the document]
sashahx: Done

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
...
+    def get_value(self) -> str | bytes:
+        """Returns the actual secret value.
+
+        Returns:
+            str | bytes: The secret value.
+        """
+
+        return self._secret_value
+
+    def _get_hidden_value(self) -> str | bytes:
```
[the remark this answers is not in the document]
sashahx: Done

litestar/datastructures/secret_values.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, secret_value: str | bytes):
+        """Initializes a SecretValue object with a secret value of type `str` or `bytes`.
+
+        Args:
+            secret_value (str | bytes): The secret value to be encapsulated.
+        """
+
+        self._secret_value = secret_value
+
+    def get_value(self) -> str | bytes:
```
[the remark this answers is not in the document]
sashahx: Done

peterschutt:
Will also need to rebase this onto the develop branch.

JacobCoffee:
hi @sashahx , any updates here? I think the only remaining thing is the encoders/decoders ask.

sashahx:
hi @JacobCoffee! No updates yet as I've been very busy, but I plan to implement the feature this week, or at least ask some questions. Thank you for your patience

JacobCoffee:
All good, just checking! 😄

PR #2985, opened 2024-01-13
fix: additional `structlog` fixes

cofin:
- fix: correctly render stdlib logs as string instead of bytes:
    This fixes an issue were Structlog raised a warning based on the default configuration
- fix: adds timestamper to standard logging for structlog
- fix: drops the duplicate `color_message` field that is rendered in certain messages


provinzkraut:
@cofin Any chance you can address the coverage? :eyes: 
Could be a separate PR as well if you prefer that, but we have sonar failing on other PRs partly because of the missing struclog coverage.

provinzkraut (approved):
Looks good aside from what @JacobCoffee already mentioned.

provinzkraut (approved):

PR #3007, opened 2024-01-21
ci: Use codecov instead of sonar

provinzkraut:
Use [codecov](https://docs.codecov.com) instead of Sonar for coverage


guacs (approved):
I'm fine with this, but I was just curious as to whether there was a specific reason to switch to codecov?

guacs:
> > I'm fine with this, but I was just curious as to whether there was a specific reason to switch to codecov?
> 
> Sonar has been quite a pain to deal with in the past. It can't reliably detect what "new code" is (i.e. it starts reporting code as decreasing coverage that hasn't been touched) and it's generally not easily configurable.

Oh yeah I've seen that. I'm good with switching to codecov.

JacobCoffee (approved):

PR #3015, opened 2024-01-23
docs: Fix create_async_test_client docstring

cbscsm:
in `create_async_test_client` docstring:
- `TestClient` -> `AsyncTestClient`
- `create_test_client` -> `create_async_test_client`

### DIFF
--- a/litestar/testing/helpers.py
+++ b/litestar/testing/helpers.py
@@ -133,7 +133,7 @@ def create_test_client(
 
             def test_my_handler() -> None:
                 with create_test_client(my_handler) as client:
-                    response == client.get("/some-path")
+                    response = client.get("/some-path")
                     assert response.json() == {"hello": "world"}
 
     Args:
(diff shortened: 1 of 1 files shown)


litestar/testing/helpers.py
```diff
@@ -378,16 +378,16 @@ def create_async_test_client(
...
             @get("/some-path")
             def my_handler() -> dict[str, str]:
                 return {"hello": "world"}
 
 
-            def test_my_handler() -> None:
-                with create_test_client(my_handler) as client:
+            async def test_my_handler() -> None:
+                async with create_async_test_client(my_handler) as client:
                     response == client.get("/some-path")
```
Alc-Alc, line 391:
```suggestion
                    response = await client.get("/some-path")
```

You probably have to remove the `==` in the sync version of the doc as well

https://github.com/litestar-org/litestar/blob/14759558cf231c61049fdd5b65a39eccb8841616/litestar/testing/helpers.py#L136

Alc-Alc (changes requested):

litestar/testing/helpers.py
Alc-Alc: > You probably have to remove the == in the sync version of the doc as well

@cbscsm can you make that change? 🙂

cbscsm: Sorry, I missed that comment. I fixed it.

PR #3016, opened 2024-01-24
fix: simplify conditional logic check for libraries

cofin:
This PR does the following CLI cleanup items:
- use `find_spec` to set "library found" variables
- Rich click's CLI patcher only needs to run once.  From then on we can import the normal click types.  I've removed a few of these conditional checks.
- Only create the beautifier class when executed
- Typo fix


guacs (approved):
Nice!

PR #3017, opened 2024-01-24
docs: Fix incorrect argument name in OpenApi ResponseSpec example

korneevm:
There is an incorrect argument name in the ResponseSpec example in the docs. ResponseSpec has no `model` arg, so when I tried to use this example I got `TypeError: ResponseSpec.__init__() got an unexpected keyword argument 'model'` exception.
The correct arg name seems to be `data_container`, so I've updated the docs accordingly.


Alc-Alc (approved):
🙂

provinzkraut:
Thanks @korneevm!

provinzkraut:
@all-contributors add @korneevm for docs

PR #3027, opened 2024-01-25
feat: allow plugins to be found by dotted path string

cofin:

### DIFF
--- a/litestar/plugins/base.py
+++ b/litestar/plugins/base.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Iterator, Protocol, TypeVar, Union, cast, runtime_checkable
 
+from litestar.utils.module_loader import import_string
+
 if TYPE_CHECKING:
     from click import Group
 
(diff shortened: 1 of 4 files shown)


litestar/plugins/base.py
```diff
@@ -267,15 +269,20 @@ def __init__(self, plugins: list[PluginProtocol]) -> None:
         self.serialization = tuple(p for p in plugins if isinstance(p, SerializationPluginProtocol))
         self.cli = tuple(p for p in plugins if isinstance(p, CLIPluginProtocol))
 
-    def get(self, type_: type[PluginT]) -> PluginT:
+    def get(self, type_: type[PluginT] | str) -> PluginT:
```
provinzkraut, line 272: I don’t think this belongs in the `get` method since it’s conceptually different: Right now the registry is immutable and `get` just returns plugins that were previously registered I straw if loading them.

If we want to allow retrieval via string, it should just look up e.g. the `__name__` attribute of the types of the plugins already registered. If we want to allow *loading* plugins via string, this should happen on app init.

provinzkraut (changes requested):

litestar/plugins/base.py
cofin: Fair point.  Are you ok with me leaving the utils that I've added here?  They were originally in Litestar but were removed at some point.

provinzkraut: > Are you ok with me leaving the utils that I've added here? They were originally in Litestar but were removed at some point.

Sure. We could probably also reuse those in the CLI for the autoimports at some point.

provinzkraut (approved):

PR #3034, opened 2024-01-27
fix: cookie string parser

### LINKED ISSUE #3023: Bug: cookie string parsing incorrectly handles attributes without values
### Description

The `parse_cookie_string()` function incorrectly handles attributes without values and will keep only one due to the dictionary conversion.
For example if the cookie contains the `HttpOnly` and the `Secure` attributes these will both have an empty string as key and eventually only one of these will be returned.
image

https://github.com/litestar-org/litestar/blob/3dfd6fe3de6a246acb358304ae44872f64c2c9f2/litestar/_parsers.py#L50-L67

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

v2.5.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

floxay:
Fixes #3023

### DIFF
--- a/litestar/_parsers.py
+++ b/litestar/_parsers.py
@@ -56,7 +56,7 @@ def parse_cookie_string(cookie_string: str) -> dict[str, str]:
     Returns:
         A string keyed dictionary of values
     """
-    cookies = [cookie.split("=", 1) if "=" in cookie else ("", cookie) for cookie in cookie_string.split(";")]
+    cookies = [cookie.split("=", 1) if "=" in cookie else (cookie, cookie) for cookie in cookie_string.split(";")]
     output: dict[str, str] = {
         k: unquote(unquote_cookie(v))
         for k, v in filter(
--- a/tests/unit/test_parsers.py
+++ b/tests/unit/test_parsers.py
@@ -56,13 +56,24 @@ def test_parse_utf8_form_data() -> None:
         ('foo="bar=123456789&name=moisheZuchmir"', {"foo": "bar=123456789&name=moisheZuchmir"}),
         ("email=%20%22%2c%3b%2f", {"email": ' ",;/'}),
         ("foo=%1;bar=bar", {"foo": "%1", "bar": "bar"}),
-        ("foo=bar;fizz  ; buzz", {"": "buzz", "foo": "bar"}),
-        ("  fizz; foo=  bar", {"": "fizz", "foo": "bar"}),
+        ("foo=bar;fizz  ; buzz", {"buzz": "buzz", "fizz": "fizz", "foo": "bar"}),
+        ("  fizz; foo=  bar", {"fizz": "fizz", "foo": "bar"}),
         ("foo=false;bar=bar;foo=true", {"bar": "bar", "foo": "true"}),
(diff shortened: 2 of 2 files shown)


litestar/_parsers.py
```diff
@@ -56,7 +56,7 @@ def parse_cookie_string(cookie_string: str) -> dict[str, str]:
     Returns:
         A string keyed dictionary of values
     """
-    cookies = [cookie.split("=", 1) if "=" in cookie else ("", cookie) for cookie in cookie_string.split(";")]
+    cookies = [cookie.split("=", 1) if "=" in cookie else (cookie, cookie) for cookie in cookie_string.split(";")]
```
provinzkraut, line 59: I think this should be

```suggestion
    cookies = [cookie.split("=", 1) if "=" in cookie else (cookie, "") for cookie in cookie_string.split(";")]
```

I've checked out other implementations and found that:

- Stdlib refuses to parse this
- Wekzeug defaults to the cookie name and an empty value
- Starlette behaves the same we do 

In addition, Starlette also specifically references this: https://bugzilla.mozilla.org/show_bug.cgi?id=169091.

Personally, I think defaulting to an empty value makes more sense than defaulting to the name as a value, as that's rather arbitrary.

provinzkraut (changes requested):

litestar/_parsers.py
floxay: I agree that an empty string makes way more sense as a value.

The only reason I have went with `(cookie, cookie)` was to not change the output of the data extractors in a way that would potentially break user code (or log DB queries, etc.) in case someone is currently looking for these optional attributes among the cookie values.

provinzkraut: Yeah, I think this is something we'd either have to hide behind a feature flag or wait for 3.0 to release due to the breaking nature.

floxay: Unsure how a feature flag for this would work, my initial idea was to use "experimental features" but the extractor has no access to the app.
Adding an additional parameter such as `future_cookie_string_parsing: bool = False` to `ResponseDataExtractor` to control the behavior of `parse_cookie_string` seems to be the easiest solution. (but would be slightly problematic when it disappears in 3.0)
However after I added this `future_cookie_string_parsing` parameter I noticed another issue; #3040, due to this one I think it might be better to just wait until 3.0 with both of them.

JacobCoffee (approved):
lgtm except for mentioned comments above

litestar/_parsers.py
provinzkraut: > I think it might be better to just wait until 3.0 with both of them

Yes, I agree.

> if so, should the target branch be changed?

You can leave them as-is for now. We'll update the branch once the 3.0 work officially has been started.

peterschutt: PR is pointed at v3.0 branch now, can we merge it there?

JacobCoffee:
@floxay @provinzkraut can we merge this now

JacobCoffee:
v3 branch is available with changes so this could be merged into that  if ci passes

Issue #2998, opened 2024-01-18
Bug: ``uvicorn>=0.26`` introduces a breaking change 


JacobCoffee:
### Description

Due to work done In https://github.com/encode/uvicorn/releases/tag/0.26.0, some users may experience 404s.

We must update our logic, and in the meantime pin to https://github.com/encode/uvicorn/releases/tag/0.25.0

### Litestar Version

v2.5.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows
- [X] Other (Please specify in the description above)

provinzkraut:
This bug does not affect us so there's nothing to do here. I'll add some tests though to ensure we're not hit by changes like these in the future (also for other servers than just uvicorn).

PR #3039, opened 2024-01-28
fix: Handle diverging ASGI `root_path` behaviour

provinzkraut:
As described in #2998, uvicorn change its `root_path` behaviour, which now diverges from most other ASGI servers. To ensure we handle all servers and cases properly I've implemented a fix here and added a few tests against common servers (uvicorn, hypercorn and daphne).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,64 +93,6 @@ sqlalchemy = ["advanced-alchemy>=0.2.2,<1.0.0"]
 standard = ["jinja2", "jsbeautifier", "uvicorn[standard]", "uvloop>=0.18.0; sys_platform != 'win32'", "fast-query-parsers>=1.0.2"]
 structlog = ["structlog"]
 
-[tool.pdm]
-ignore_package_warnings = ["sphinx", "slotscheck"]
-
-[tool.pdm.dev-dependencies]
-dev = [
-  "beanie>=1.21.0",
-  "beautifulsoup4",
-  "fsspec",
-  "greenlet",
-  "hypothesis",
-  "python-dotenv",
-  "starlette",
-  "trio",
-  "aiosqlite",
-  "asyncpg>=0.29.0",
-  "psycopg[pool,binary]>=3.1.10",
-  "psycopg2-binary",
-]
-dev-contrib = ["opentelemetry-sdk", "httpx-sse"]
-docs = [
-  "sphinx>=7.1.2",
-  "sphinx-autobuild>=2021.3.14",
-  "sphinx-copybutton>=0.5.2",
-  "sphinx-toolbox>=3.5.0",
-  "blacken-docs>=1.16.0",
-  "sphinx-design>=0.5.0",
-  "sphinx-click>=4.4.0",
-  "sphinxcontrib-mermaid>=0.9.2",
-  "auto-pytabs[sphinx]>=0.4.0",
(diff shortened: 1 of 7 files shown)


pyproject.toml
provinzkraut, line 1: Not sure why pdm reordered the file like that.. :shrug:

JacobCoffee (approved):

provinzkraut:
Okay so it turns out that there was a different case where we *were* affected by this, as reported by @smithk86 in #3041. I've added a test case for this here and implemented a fix.

I've also included some basic integration tests for a few webservers with different versions for uvicorn. We should probably refactor those into their own test suite, but I didn't want to blow up this PR.

smithk86:
@provinzkraut I tested a few combinations, and I may have found an issue. My MCVE allows for path to be `None`. On Uvicorn 0.25/Litestar 2.5.2 (with no root-path given), I can pass no `path` and get "path: None" as the response payload. On Uvicorn 0.27/Litestar-fix-uvicorn-root-path-branch though, I'm getting a 404 if no path is given.

provinzkraut:
> @provinzkraut I tested a few combinations, and I may have found an issue. My MCVE allows for path to be `None`. On Uvicorn 0.25/Litestar 2.5.2 (**with** root-path given), I can pass no `path` and get "path: None" as the response payload. On Uvicorn 0.27/Litestar-fix-uvicorn-root-path-branch though, I'm getting a 404 if no path is given.
> 
> _edit: The above was originally wrong. Paths are handled correct without a root-path. The null path only breaks when root-path is present._

Good catch, thanks! I've added your MCVE as a test case now and made a new patch. Can you verify that everything works as expected with the newest commit?

provinzkraut:
Tests are currently blocked by #3042

smithk86:
> Good catch, thanks! I've added your MCVE as a test case now and made a new patch. Can you verify that everything works as expected with the newest commit?

@provinzkraut I ran though my sanity checks again, and everything looks great!

peterschutt (approved):
Great idea to test against the different servers!

guacs:
@provinzkraut I had a doubt on this. Shouldn't the `root_path`, if available, be set as the prefix for the path that is created via `route_reverse` and `url_for_static_assets`? That is, which of the assertions below is the expected behavior if the server is ran using `python -m uvicorn app:app --root-path /test`?

```python
@get("/handler", name="handler")
async def handler(request: Request) -> str:
    return request.app.route_reverse("handler")

from typing import List

from litestar.testing.helpers import create_test_client


@get("/handler", name="handler")
async def handler(request: Request) -> str:
    return request.app.route_reverse("handler")


with create_test_client([handler]) as client:
    response = client.get("/handler")

    url = response.json()

    assert url == "/handler"  # current behavior
    assert url == "/test/handler"  # is this the expected behavior?
```

In the [PR](https://github.com/encode/uvicorn/pull/2213) that changed the behavior in uvicorn, the following is stated:

> Uvicorn is handling an ASGI app (e.g. Starlette, FastAPI) that knows of the path /items/42, but doesn't know of the prefix /api/v1, but the app might need to create some url_for in templates or such, so it needs to also know the original path prefix it should use.

provinzkraut:
> @provinzkraut I had a doubt on this. Shouldn't the `root_path`, if available, be set as the prefix for the path that is created via `route_reverse` and `url_for_static_assets`? 

Yes, it should be. It was an oversight, I probably should have double checked our tests for this. I’ll do a follow up PR to fix!

PR #3042, opened 2024-01-28
test: Fix flaky tests

provinzkraut:
Fix flaky tests as seen here: https://github.com/litestar-org/litestar/actions/runs/7687989337/job/20948769774?pr=3039#step:7:159

These flaky tests were caused by re-using modules from example apps between tests.


cofin (approved):
nice work!

provinzkraut:
Blocked by #3043

PR #3048, opened 2024-01-30
fix: ServerSentEvent typing error

euri10:
fixes small typing error:

```
error: Argument 1 to "ServerSentEvent" has incompatible type "AsyncIterable[ServerSentEventMessage]"; expected "str | bytes | Iterable[str | bytes] | Iterator[str | bytes] | AsyncIterable[str | bytes] | AsyncIterator[str | bytes]"  [arg-type]
 ```

inside `test_sse` there was a `Any` I changed to trigger the test then solved it.

### DIFF
--- a/litestar/types/helper_types.py
+++ b/litestar/types/helper_types.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
 
 T = TypeVar("T")
 
-__all__ = ("OptionalSequence", "SyncOrAsyncUnion", "AnyIOBackend", "StreamType", "MaybePartial")
+__all__ = ("OptionalSequence", "SyncOrAsyncUnion", "AnyIOBackend", "StreamType", "MaybePartial", "SSEData")
 
 OptionalSequence: TypeAlias = Optional[Sequence[T]]
 """Types 'T' as union of Sequence[T] and None."""
--- a/litestar/types/helper_types.py
+++ b/litestar/types/helper_types.py
@@ -17,10 +17,10 @@ from typing import (
     Union,
 )
 
-from litestar.response.sse import ServerSentEventMessage
-
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
+
+    from litestar.response.sse import ServerSentEventMessage
 
 
 T = TypeVar("T")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by provinzkraut — CHANGES_REQUESTED** [review]

File: litestar/types/helper_types.py [hunk 1]
```diff
@@ -15,9 +17,12 @@
     Union,
 )
 
+from litestar.response.sse import ServerSentEventMessage
```
euri10, line 20: CI fails for a weird reason to me, mentionning circular imports (https://github.com/litestar-org/litestar/actions/runs/7713504937/job/21023551864?pr=3048) and this is the only instance  can think in the PR could cause this, however at first  put him under TPYE_CHECKING but ruff was complaining

euri10: if I put it under TPYE_CHECKING I get:
litestar/types/helper_types.py:23:39: TCH004 Move import `litestar.response.sse.ServerSentEventMessage` out of type-checking block. Import is used for more than type hinting.

provinzkraut: > Import is used for more than type hinting.

You'll need to put the import in quotes. I left a comment down below!

litestar/types/helper_types.py
```diff
@@ -37,3 +42,6 @@
 
 MaybePartial: TypeAlias = Union[T, partial]
 """A potentially partial callable."""
+
+SSEData: TypeAlias = Union[int, str, bytes, Dict[str, Any], ServerSentEventMessage]
```
provinzkraut, line 46:
```suggestion
SSEData: TypeAlias = Union[int, str, bytes, Dict[str, Any], "ServerSentEventMessage"]
```

litestar/types/helper_types.py
```diff
@@ -15,9 +17,12 @@
     Union,
 )
 
+from litestar.response.sse import ServerSentEventMessage
+
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
 
+
```
provinzkraut, line 25:
```suggestion

if TYPE_CHECKING:
    from typing_extensions import TypeAlias
    from litestar.response.sse import ServerSentEventMessage


```

euri10: done, thanks for bearing with me

provinzkraut (approved):

PR #3064, opened 2024-02-02
feat: only display path in `ValidationException`s

### LINKED ISSUE #3061: Enhancement: only expose the path in `ValidationException`s
### Summary

Currently `ValidationException` exposes the full URL in the error response, leaking internal IP(s) or other similar infra related information.

Relevant code:
https://github.com/litestar-org/litestar/blob/d52caee70c6fa6c6806f7a80833b872118b3fc89/litestar/_signature/model.py#L122 https://github.com/litestar-org/litestar/blob/d52caee70c6fa6c6806f7a80833b872118b3fc89/litestar/_kwargs/extractors.py#L110  `{connection.url}` -> `{connection.url.path}`

### Basic Example

```json
{
    "status_code": 400,
    "detail": "Missing required query parameter 'q' for url http://127.0.0.1:8000/"
}
```
and
```json
{
    "status_code": 400,
    "detail": "Validation failed for GET http://127.0.0.1:8000/?q=s",
    "extra": ...
}
```
would become
```json
{
    "status_code": 400,
    "detail": "Missing required query parameter 'q' for path /"
}
```
```json
{
    "status_code": 400,
    "detail": "Validation failed for GET /",
    "extra": ...
}
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3061">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3061/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

floxay:
Closes #3061

### DIFF
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -107,7 +107,7 @@ def create_connection_value_extractor(
         except KeyError as e:
             param = alias_to_params[e.args[0]]
             raise ValidationException(
-                f"Missing required {param.param_type.value} parameter {param.field_alias!r} for url {connection.url}"
+                f"Missing required {param.param_type.value} parameter {param.field_alias!r} for path {connection.url.path}"
             ) from e
 
     return extractor
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -11,6 +11,7 @@ from litestar._parsers import (
 )
 from litestar.datastructures import Headers
 from litestar.datastructures.upload_file import UploadFile
+from litestar.datastructures.url import URL
 from litestar.enums import ParamType, RequestEncodingType
 from litestar.exceptions import ValidationException
 from litestar.params import BodyKwarg
@@ -106,8 +107,12 @@ def create_connection_value_extractor(
             values.update(connection_mapping)
         except KeyError as e:
             param = alias_to_params[e.args[0]]
+            path = URL.from_components(
+                path=connection.url.path,
+                query=connection.url.query,
+            )
             raise ValidationException(
-                f"Missing required {param.param_type.value} parameter {param.field_alias!r} for path {connection.url.path}"
+                f"Missing required {param.param_type.value} parameter {param.field_alias!r} for path {path}"
             ) from e
 
     return extractor
--- a/litestar/_signature/model.py
+++ b/litestar/_signature/model.py
@@ -32,6 +32,7 @@ from litestar._signature.utils import (
     _validate_signature_dependencies,
 )
 from litestar.datastructures.state import ImmutableState
+from litestar.datastructures.url import URL
 from litestar.dto import AbstractDTO, DTOData
 from litestar.enums import ParamType, ScopeType
 from litestar.exceptions import InternalServerException, ValidationException
@@ -119,9 +120,11 @@ class SignatureModel(Struct):
             for err_message in messages
             if ("key" in err_message and err_message["key"] not in cls._dependency_name_set) or "key" not in err_message
         ]:
-            return ValidationException(
-                detail=f"Validation failed for {method} {connection.url.path}", extra=client_errors
+            path = URL.from_components(
+                path=connection.url.path,
+                query=connection.url.query,
             )
+            return ValidationException(detail=f"Validation failed for {method} {path}", extra=client_errors)
         return InternalServerException()
 
     @classmethod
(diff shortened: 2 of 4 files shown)


provinzkraut (changes requested):
The error should include `path`, `query` and `fragment` so it can easily be used to reproduce and trace errors.

This can be obtained from the `URL` something like this:

```python
URL.from_components(
  path=connection.url.path, 
  query=connection.url.query,
  fragment=connection.url.fragment,
)
```

I also think we should be more consistent about this. For some exceptions we include the path in the message and for some we don't.
@litestar-org/maintainers thoughts on how we should handle this in general?

floxay:
> The error should include `path`, `query` and `fragment` so it can easily be used to reproduce and trace errors.

This logic is a bit strange to me personally.
The server/app ideally would have logging configured to aid with this, while on the other hand the client itself knows the sent values for all of these and the destination.
Including only query params and fragment does not really help to repro (and trace?) validation errors originating from something else than query validation; such as body validation. Server side logging is required there.
/ The argument now could obviously be made that the path in the error itself is also not really useful, as the client knows it and the server/app has access to the whole request/response context which includes the request path, thus can be logged. /

> The error should include

If the final decision from maintainers is to include these (`path`, `query` and `fragment`), where should these be included? In the error `extra` as some sort of context or just in the error `detail` as a string like before?

guacs:
> The error should include `path`, `query` and `fragment` so it can easily be used to reproduce and trace errors.
> 
> This can be obtained from the `URL` something like this:
> 
> ```python
> URL.from_components(
>   path=connection.url.path, 
>   query=connection.url.query,
>   fragment=connection.url.fragment,
> )
> ```
> 
> I also think we should be more consistent about this. For some exceptions we include the path in the message and for some we don't. @litestar-org/maintainers thoughts on how we should handle this in general?

I think defaulting to showing only the part of the path after the domain would be reasonable. This way none of the internal stuff is leaked, but the response will also contain enough information for the API consumer.

floxay:
> This way none of the internal stuff is leaked

Thinking about this, depending on how routing is set-up internally the "none" part might not be true? Sure the IP/domain is not exposed but the path itself can have differences compared to what the user sees from the outside.

> The error should include `path`, `query` and `fragment` so it can easily be used to reproduce and trace errors.

The client knows (or at least should know) where and what it sends, mirroring back its own query and fragment (and a path that may match or not) values does not seem (at least to me) particularly helpful, especially as this only happens matters for a subset of validation errors, for example; body validation errors do not "benefit" from this. Why not just leave this up for logging to deal with?

/ this is just *my opinion* on the matter as a Litestar *user* /

guacs:
> > This way none of the internal stuff is leaked
> 
> Thinking about this, depending on how routing is set-up internally the "none" part might not be true? Sure the IP/domain is not exposed but the path itself can have differences compared to what the user sees from the outside.
> 
> > The error should include `path`, `query` and `fragment` so it can easily be used to reproduce and trace errors.
> 
> The client knows (or at least should know) where and what it sends, mirroring back its own query and fragment (and a path that may match or not) values does not seem (at least to me) particularly helpful, especially as this only happens matters for a subset of validation errors, for example; body validation errors do not "benefit" from this. Why not just leave this up for logging to deal with?
> 
> / this is just _my opinion_ on the matter as a Litestar _user_ / (kinda out of scope at this point considering the original issue and PR)

Yeah, it makes sense to include only the route and not the full path including the query params etc. In fact, I think it may be fine to just not include the path itself since the caller of the API would/should arguably know the path they made the request to.

provinzkraut:
> Yeah, it makes sense to include only the route and not the full path including the query params etc. In fact, I think it may be fine to just not include the path itself since the caller of the API would/should arguably know the path they made the request to.

I disagree. The error message is mostly for debugging / logging; We're not talking about the error detail we return in the actual response here, in that case I'd agree with you. But if you have an error log, it's extremely helpful to know the path.

floxay:
> I disagree. The error message is mostly for debugging / logging; We're not talking about the error detail we return in the actual response here, in that case I'd agree with you. But if you have an error log, it's extremely helpful to know the path.

I'm not sure I understand but as far as I can see currently with the most recent commit to this PR both the path and query is returned to the client. I do not think something that is meant to be used internally for "debugging/logging" should be exposed in the response sent to the client, at least not by default, especially when the "debug flag" is disabled.

Anyway, with 6278966b9ad1a2b0eda753ae114c64397c728fc8 the query parameters now show up in the error and adjusted the test again for this.
The reason `fragment` is not there is that it is entirely client sided, the client itself never sends it to the server, meaning it is is always empty (as far as I can tell). It is also not part of the `Scope` which is used to construct the `URL` instance using `URL.from_scope` which we end up using here from the `connection` to construct the new `URL` with the query parameters.

provinzkraut (approved):

PR #3070, opened 2024-02-04
fix: scope `state` key handling

provinzkraut:
Fix a regression introduced in #2751 that would wrongfully assume the `state` key is always present within the ASGI Scope. This is *only* the case when the `Litestar` root application is invoked first, since we enforce such a key there, but the presence of that key is not actually guaranteed by the ASGI spec and some servers, such as hypercorn, do not provide it.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -537,7 +537,7 @@ class Litestar(Router):
             return
 
         scope["app"] = self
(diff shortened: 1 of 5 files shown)


cofin (approved):
LGTM

PR #3075, opened 2024-02-05
fix(scope): Revert scope state regression

provinzkraut:
Fix a regression introduced in #3070 while fixing another regression.

This now leaves the `scope["state"]` handling functionally untouched, to preserve behaviour. The behaviour is still not entirely correct, but it's not something we can fix without introducing breaking changes.

### DIFF
--- a/litestar/utils/scope/state.py
+++ b/litestar/utils/scope/state.py
@@ -96,9 +96,9 @@ class ScopeState:
         Returns:
             A `ConnectionState` object.
         """
-        base = scope["state"] if "state" in scope else scope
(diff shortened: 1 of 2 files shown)


euri10:
I get the same behavior using this branch than in 2.5.5, last one to work was 2.5.4
I have bad wifi in train will post later with more details hopefully, 
edit : I understood where this comes from, I'm using LifespanManager from asgi_lifespan, the change between 2.5.4 and 2.5.5 impacts this

euri10:
works totally fine for me, I needed a push on the back anyways to use the built-ins Litestar provides :) Time to adapt.

guacs (approved):

PR #3079, opened 2024-02-06
chore(infra): re-add fields for PR templates

JacobCoffee:
## Details

- Adds in this section here

## Closes

- and this section here!

back to the PR templates.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,8 +4,14 @@ By submitting this pull request, you agree to:
 - follow [Litestar's contribution guidelines](https://github.com/litestar-org/.github/blob/main/CONTRIBUTING.md)
 - follow the [PSFs's Code of Conduct](https://www.python.org/psf/conduct/)
 -->
+## Description
+
+-
 
 <!--
 Please add in issue numbers this pull request will close, if applicable
 Examples: Fixes #4321 or Closes #1234
 -->
+## Closes
+
+-


provinzkraut:
I’m okay with adding these back in, but can we do the closing section in a way that works? GitHub only supports the explicitly spelled out closing keywords, and I think it’s super rare to have a PR that closes more than one issue anyway so the list doesn’t really add much value.

guacs:
> I’m okay with adding these back in, but can we do the closing section in a way that works? GitHub only supports the explicitly spelled out closing keywords, and I think it’s super rare to have a PR that closes more than one issue anyway so the list doesn’t really add much value.

We could just include the comment that's there regarding mentioning the issue and remove the `Closes` section.

JacobCoffee:
@guacs Can you show me an example of what you mean? I'm not 100% i follow and if its just the comment as is now people dont read/fill it out :\

guacs:
> @guacs Can you show me an example of what you mean? I'm not 100% i follow and if its just the comment as is now people dont read/fill it out :\

Oh my mistake. I thought the focus was on the `Details` heading to ensure that there is a good description of the PR. I meant something like the following:

```
<!-- guildelines -->
# Details

PR description.

<!-- the existing comment to include the issue that this fixes -->
```
So essentially, just removing the `Closes` heading was what I meant, but if that could result in not including the issue this fixes, then I suppose it may be better to keep it in. If we are keeping it in, then I think it may be better to include the comment that explains how to link the issue this closes under the `Closes` heading (currently, it's above it).

provinzkraut:
> @guacs Can you show me an example of what you mean? I'm not 100% i follow and if its just the comment as is now people dont read/fill it out :\

The issue is just with it being a list. The template could be e.g.

```markdown
**Closes <!-- add issue number here --> **
```

provinzkraut (approved):

PR #2960, opened 2024-01-07
feat: Implement static file serving with regular route handlers

provinzkraut:
Implement static file serving with regular route handlers instead of a specialised ASGI app. At the moment, this is complementary to the usage of `StaticFilesConfig` to keep backwards compatibility.

This achieves a few things:

- Fixes #2629
- Circumvents all of the special casing we have to do in our routing logic for the static files app
- Allows us to drop said special casing when removing the old functionality
- Removes the need for a `static_files_config` on the app
- Removes the need for a special `url_for_static_asset` method on the app since `route_reverse` can be used instead

Additionally:

- Most router options can now be passed to the `create_static_files_router`, allowing further customisation
- A new `resolve_symlinks` flag has been added, defaulting to `True` to keep backwards compatibility

## Usage

Instead of 

```python
app = Litestar(
    static_files_config=[
        StaticFilesConfig(path="/static", directories=["some_dir"])
    ]
)
``` 

You can now simply use

```python
app = Litestar(
  route_handlers=[
    create_static_files_router(path="/static", directories=["some_dir"])
  ]
)
```

This takes the same arguments as the `StaticFilesConfig`, but generates a router internally, with route handlers to serve the static files.

## Todo

- [x] Add docs
- [x] Add deprecation warnings for `StaticFilesConfig` related stuff

PR #3087, opened 2024-02-08
fix: replace deprectated property

cbscsm:
- `Litestar.static_files_config` property is deprecated since 2.6.0 by #2960 
- But `Litestar.__init__` uses the property. I'm not sure that it is intended or not, I can't avoid that deprecated warning from calling `Litestar(...)`.
  ```
  DeprecationWarning: Use of deprecated property 'static_files_config'. Deprecated in litestar 2.6.0. This property will be removed in the next major version. Use create_static_files router instead
  ```
- I suggest replace the property access to internal attribute in internal code.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -457,7 +457,7 @@ class Litestar(Router):
             self.get_logger = self.logging_config.configure()
             self.logger = self.get_logger("litestar")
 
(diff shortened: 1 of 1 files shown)


guacs:
@cbscsm the failing build should be fixed by #3089.

PR #3089, opened 2024-02-08
fix: ensure importing of all click related stuff after running `rich_click_path`

guacs:
Currently, the importing of `LitestarExtensionGroup` occurs before running `rich_click_patch`. This causes an issue due to `rich_click_patch` monkeypatching `click.Group`, `click.Command` etc. For example, consider the following code:

```python
from click import Command, Group
from click.core import Command as CCommand
from click.core import Group as CGroup

from litestar.cli import litestar_group

print(isinstance(litestar_group, (Group, Command)))  # False
print(isinstance(litestar_group, (CGroup, CCommand)))  # True

# These fail.
assert Group is CGroup
assert Command is CCommand
```
The reason for this behavior is that `LitestarExtensionGroup` is not a subclass of `RichGroup`, but a subclass of `click.Group` if it gets imported before `rich_click` gets a chance to do its patching. Then by the time we do the `isinstance` check, `click.Group` is patched to become `rich_click.RichGroup` and so the `isinstance(litestar_group, click.Group` fails.

This causes issues in the documentation generation using `sphinx-click` which does a `isinstance` which is why [this CI](https://github.com/litestar-org/litestar/actions/runs/7824742530/job/21347805990?pr=3087#step:6:48) is failing in #3087.

To fix this, we simply ensure that we run the `rich_click` patching before we any imports from `click`.

### DIFF
--- a/litestar/cli/__init__.py
+++ b/litestar/cli/__init__.py
@@ -1,6 +1,28 @@
 """Litestar CLI."""
 from __future__ import annotations
 
+from importlib.util import find_spec
+
+# Ensure `rich_click` patching occurs before we do any imports from `click`.
+if find_spec("rich_click") is not None:
+    import rich_click as click
+    from rich_click.cli import patch as rich_click_patch
+
+    rich_click_patch()
+    click.rich_click.USE_RICH_MARKUP = True
+    click.rich_click.USE_MARKDOWN = False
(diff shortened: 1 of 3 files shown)


cofin (approved):
Thanks for catching this one.

PR #3095, opened 2024-02-12
fix: cookie naming for short cookies

### LINKED ISSUE #3090: Bug: session middleware cookies always include segment number
### Description

The docstring of [litestar.middleware.session.client_side.CookieBackendConfig](https://github.com/litestar-org/litestar/blob/04a9501d719c26ff4ef8451d216707a1b72a1c77/litestar/middleware/session/client_side.py#L209) suggests that key for the cookie inside the header is of the form `session=<data>`, unless the cookie is over 4KB in which case it's split into segments named e.g. `session-0=<data-0>`, etc.  

However, even when the cookie is under 4KB, the segment number is present in the key, i.e. it's impossible to specify that the key be named `session` -- it will always turn into `session-0`.  The reason for this is that 
 `litestar.middleware.session.client_side.ClientSideSessionBackend._create_session_cookies` returns the result of the list comprehension which includes `[Cookie(..., key=f"{self.config.key}-{i}", ...) for i, datum in enumerate(data)]`.

_No response_

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.6.0

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3090">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3090/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

betaprior:
Previously, cookie names always had a suffix of the form "-{i}" appended to them. With this change, the suffix is omitted if the cookie is short enough (< 4 KB) to not be split into multiple chunks.

If the cookie is so big that it is split into several chunks, the chunk number suffix is appended, as before.

Fixes #3090

### DIFF
--- a/litestar/middleware/session/client_side.py
+++ b/litestar/middleware/session/client_side.py
@@ -108,7 +108,10 @@ class ClientSideSessionBackend(BaseSessionBackend["CookieBackendConfig"]):
         return sorted(key for key in connection.cookies if self.cookie_re.fullmatch(key))
 
     def _create_session_cookies(self, data: list[bytes], cookie_params: dict[str, Any] | None = None) -> list[Cookie]:
-        """Create a list of cookies containing the session data."""
+        """Create a list of cookies containing the session data.
+        If the data is split into multiple cookies, the key will be of the format ``session-{segment number}``,
+        however if only one cookie is needed, the key will be ``session``.
+        """
         if cookie_params is None:
             cookie_params = dict(
                 extract_dataclass_items(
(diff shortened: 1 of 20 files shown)


litestar/middleware/session/client_side.py
```diff
@@ -117,10 +117,18 @@ def _create_session_cookies(self, data: list[bytes], cookie_params: dict[str, An
                     include={f for f in Cookie.__dict__ if f not in ("key", "secret")},
                 )
             )
+
+        def _cookie_chunk_key(i: int) -> str:
+            # If the data is split into multiple cookies, the key will be of the format `session-{segment number}`,
+            # however if only one cookie is needed, the key will be `session`.
+            if len(data) == 1:
+                return self.config.key
+            return f"{self.config.key}-{i}"
```
guacs, line 126: I'm not sure if creating a function here is a good idea. This would result in a new function object being created everytime this method is called.

I think using an if condition directly would be enough.

```python
if len(data) == 1:
    return [Cookie(...)] # correct naming

# the existing list comprehension
...
```

provinzkraut (approved):

provinzkraut:
@all-contributors add @betaprior for code

PR #3096, opened 2024-02-12
feat: support "+json"-suffixed response media types

### LINKED ISSUE #3088: Enhancement: Support "+json" suffixed media types per RFC 6839
### Summary

Json encoded responses in Litestar must have a media type which starts with "application/json", as can be seen https://github.com/litestar-org/litestar/blob/main/litestar/response/base.py#L388
```
if media_type.startswith("application/json"):
    return encode_json(content, enc_hook)
```

I would like this to change so that it supports "application" media types with the "+json" suffix, as described in  [RFC 6839](https://datatracker.ietf.org/doc/html/rfc6839#section-3.1). For example, the lines could change to something like
```
super_type, sub_type = media_type.split("/", 1)
if super_type == "application" and (sub_type.startswith("json") or sub_type.endswith("+json")):
    return encode_json(content, enc_hook)
```
So that I can set the media type to, e.g., "application/problem+json" (as described [here](https://webconcepts.info/concepts/media-type/application/problem+json)), and then json encoding would still work.

### Basic Example

I would like to be able to do something like the below and have the json encoding of the response work. If I run the below app now, the schema look alright, but the endpoint gives a server error because the response encoding fails since the media type does not start with "application/json". (It is clear that this is the problem because if I change the media type to "application/json+problem", it works.)
```python
from typing import Optional
import pydantic
import litestar
import litestar.status_codes

class ProblemDetails(pydantic.BaseModel):
    title: str
    type: str
    status: Optional[int] = None
    detail: Optional[str] = None
    instance: Optional[str] = None


@litestar.get(
    status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT,
    media_type="application/problem+json",
)
async def problem_endpoint() -> ProblemDetails:
    return ProblemDetails(title="broken", type="bad server")

app = litestar.Litestar(route_handlers=[problem_endpoint])
```

### Drawbacks and Impact

I do not see any drawbacks. If you preserve the behavior for media types which start with "application/json", I think it should all be fine. It just extends the media types which can be encoded in responses, which is very unlikely to be problematic since I doubt many servers rely on response encoding failing.

The "+json" suffix for application media types is not uncommon. For example, the "ProblemDetails" specification is quite common. Since there is an RFC for it (RFC 6839) it is clear that this is a standard worthy to support.

### Unresolved questions

_No response_

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3088">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3088/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bunny-therapist:
Giving route handlers a response media type of the form "application/<something>+json" automatically uses json encoding for the response.

Closes #3088

### DIFF
--- a/docs/examples/responses/json_suffix_responses.py
+++ b/docs/examples/responses/json_suffix_responses.py
@@ -1,26 +1,14 @@
-from typing import Optional
-
-from pydantic import BaseModel
-
 import litestar.status_codes
 from litestar import Litestar, get
 
 
-class ProblemDetails(BaseModel):
-    title: str
-    type: str
-    status: Optional[int] = None
-    detail: Optional[str] = None
-    instance: Optional[str] = None
-
-
 @get("/resources", status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT, media_type="application/problem+json")
-async def retrieve_resource() -> ProblemDetails:
-    return ProblemDetails(
-        title="Server thinks it is a teapot",
-        type="Server delusion",
-        status=litestar.status_codes.HTTP_418_IM_A_TEAPOT,
-    )
+async def retrieve_resource() -> dict:
+    return {
+        "title": "Server thinks it is a teapot",
+        "type": "Server delusion",
+        "status": litestar.status_codes.HTTP_418_IM_A_TEAPOT,
+    }
 
 
 app = Litestar(route_handlers=[retrieve_resource])
--- a/docs/examples/responses/json_suffix_responses.py
+++ b/docs/examples/responses/json_suffix_responses.py
@@ -1,11 +1,11 @@
-from typing import Any
+from typing import Any, Dict
 
 import litestar.status_codes
 from litestar import Litestar, get
 
 
 @get("/resources", status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT, media_type="application/problem+json")
-async def retrieve_resource() -> dict[str, Any]:
+async def retrieve_resource() -> Dict[str, Any]:
     return {
         "title": "Server thinks it is a teapot",
         "type": "Server delusion",
--- a/docs/examples/responses/json_suffix_responses.py
+++ b/docs/examples/responses/json_suffix_responses.py
@@ -1,9 +1,11 @@
+from typing import Any
+
 import litestar.status_codes
 from litestar import Litestar, get
 
 
 @get("/resources", status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT, media_type="application/problem+json")
-async def retrieve_resource() -> dict:
+async def retrieve_resource() -> dict[str, Any]:
     return {
         "title": "Server thinks it is a teapot",
         "type": "Server delusion",
--- a/docs/examples/responses/json_suffix_responses.py
+++ b/docs/examples/responses/json_suffix_responses.py
@@ -1,0 +1,26 @@
+from typing import Optional
+
+from pydantic import BaseModel
+
+import litestar.status_codes
+from litestar import Litestar, get
+
+
+class ProblemDetails(BaseModel):
+    title: str
+    type: str
+    status: Optional[int] = None
+    detail: Optional[str] = None
+    instance: Optional[str] = None
+
+
+@get("/resources", status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT, media_type="application/problem+json")
+async def retrieve_resource() -> ProblemDetails:
+    return ProblemDetails(
+        title="Server thinks it is a teapot",
+        type="Server delusion",
+        status=litestar.status_codes.HTTP_418_IM_A_TEAPOT,
+    )
+
+
+app = Litestar(route_handlers=[retrieve_resource])
--- a/litestar/response/base.py
+++ b/litestar/response/base.py
@@ -385,7 +385,9 @@ class Response(Generic[T]):
             if media_type == MediaType.MESSAGEPACK:
                 return encode_msgpack(content, enc_hook)
 
-            if media_type.startswith("application/json"):
+            if media_type.startswith("application/json") or (
+                media_type.startswith("application/") and media_type.endswith("+json")
+            ):
                 return encode_json(content, enc_hook)
 
             raise ImproperlyConfiguredException(f"unsupported media_type {media_type} for content {content!r}")
--- a/litestar/response/base.py
+++ b/litestar/response/base.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import itertools
+import re
 from typing import TYPE_CHECKING, Any, ClassVar, Generic, Iterable, Literal, Mapping, TypeVar, overload
 
 from litestar.datastructures.cookie import Cookie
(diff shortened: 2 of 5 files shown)


bunny-therapist:
The documentation preview does not have my changes. Am I supposed to build it manually somehow?

provinzkraut:
> The documentation preview does not have my changes. Am I supposed to build it manually somehow?

Seems to be there?

Screenshot from 2024-02-12 14-13-26

https://litestar-org.github.io/litestar-docs-preview/3096/usage/responses.html#json-responses

litestar/response/base.py
```diff
@@ -385,7 +385,9 @@ def render(self, content: Any, media_type: str, enc_hook: Serializer = default_s
             if media_type == MediaType.MESSAGEPACK:
                 return encode_msgpack(content, enc_hook)
 
-            if media_type.startswith("application/json"):
+            if media_type.startswith("application/json") or (
+                media_type.startswith("application/") and media_type.endswith("+json")
+            ):
```
provinzkraut, line 390: This could be simplified to a regex

bunny-therapist: Could do `re.match` with `^application/(?:.+\+)?json`

Technically, I suppose I could take into account that the string should either end after json, or have a semicolon, but that would break backwards compatibility with the old `startswith("application/json")`.

provinzkraut: > I do see an issue, now, however, with additional information about encoding/charset in the media type, like "application/json; charset=utf-8". I assume the "startswith" is there because of such information? That would have to be handled as well.

That would be easier to handle with a regex IMO because you don't have to rely on the `+json`/`json` part being the end of the string.

provinzkraut: > Could do `re.match` with `^application/(?:.+\+)?json`

Sounds good.

> Technically, I suppose I could take into account that the string should either end after json, or have a semicolon, but that would break backwards compatibility with the old `startswith("application/json")`.

Yeah, I guess we should do stricter matching here, but I don't think we can before 3.0 because it would be breaking.

bunny-therapist: Done.

bunny-therapist:
"build docs" failed for the commit with
```
/home/runner/work/litestar/litestar/docs/reference/cli.rst:7: ERROR: "<class 'litestar.cli._utils.LitestarExtensionGroup'>" of type "litestar.cli:litestar_group" is not click.Command or click.Group."click.BaseCommand"

Sphinx parallel build error:
tools.sphinx_ext.run_examples.StartupError: App docs/examples/data_transfer_objects/factory/providing_values_for_nested_data.py failed to come online
make: *** [Makefile:152: docs] Error 2
reading sources... [ 89%] usage/databases/piccolo .. usage/metrics/prometheus
Error: Process completed with exit code 2.
```
I can't see how I could have caused this or what I can do about it?

provinzkraut:
> "build docs" failed for the commit with
> 
> ```
> /home/runner/work/litestar/litestar/docs/reference/cli.rst:7: ERROR: "<class 'litestar.cli._utils.LitestarExtensionGroup'>" of type "litestar.cli:litestar_group" is not click.Command or click.Group."click.BaseCommand"
> 
> Sphinx parallel build error:
> tools.sphinx_ext.run_examples.StartupError: App docs/examples/data_transfer_objects/factory/providing_values_for_nested_data.py failed to come online
> make: *** [Makefile:152: docs] Error 2
> reading sources... [ 89%] usage/databases/piccolo .. usage/metrics/prometheus
> Error: Process completed with exit code 2.
> ```
> 
> I can't see how I could have caused this or what I can do about it?

This was fixed in `main` already iirc. Should be good now here too.

bunny-therapist:
> > "build docs" failed for the commit with
> > ```
> > /home/runner/work/litestar/litestar/docs/reference/cli.rst:7: ERROR: "<class 'litestar.cli._utils.LitestarExtensionGroup'>" of type "litestar.cli:litestar_group" is not click.Command or click.Group."click.BaseCommand"
> > 
> > Sphinx parallel build error:
> > tools.sphinx_ext.run_examples.StartupError: App docs/examples/data_transfer_objects/factory/providing_values_for_nested_data.py failed to come online
> > make: *** [Makefile:152: docs] Error 2
> > reading sources... [ 89%] usage/databases/piccolo .. usage/metrics/prometheus
> > Error: Process completed with exit code 2.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I can't see how I could have caused this or what I can do about it?
> 
> This was fixed in `main` already iirc. Should be good now here too.

But this is not based on `main`, but on `develop`, which is 9 commits behind `main`.

provinzkraut:
> But this is not based on `main`, but on `develop`, which is 9 commits behind `main`.

Yeah, that was the issue. `develop` wasn't up to date. The fix was to update `develop` and subsequently your branch so they had the latest changes from `main`. Sorry if my initial explanation was a bit vague :eyes:

bunny-therapist:
Some checks failed with docker compose errors now?

docs/examples/responses/json_suffix_responses.py
provinzkraut, line 1: This shouldn't use Pydantic, as it's not a required dependency of Litestar. To keep the example as simple as possible, I'd just use a list/dict here, which should be enough to demonstrate this functionality.

provinzkraut (changes requested):
I've got one comment about the example, other than that this is good to go in!

docs/examples/responses/json_suffix_responses.py
```diff
@@ -1,26 +1,14 @@
...
-
-
 @get("/resources", status_code=litestar.status_codes.HTTP_418_IM_A_TEAPOT, media_type="application/problem+json")
-async def retrieve_resource() -> ProblemDetails:
-    return ProblemDetails(
-        title="Server thinks it is a teapot",
-        type="Server delusion",
-        status=litestar.status_codes.HTTP_418_IM_A_TEAPOT,
-    )
+async def retrieve_resource() -> dict:
```
provinzkraut, line 6:
```suggestion
async def retrieve_resource() -> dict[str, Any]:
```

Just to keep consistent with the rest of the examples

provinzkraut (approved):

bunny-therapist:
A test failed mysteriously, but I think it is marked as "flaky".

bunny-therapist:
Huh. I guess `dict[str, Any]` only became a thing in python 3.9. I will update.

bunny-therapist:
All checks are passing now. I am not sure what "update branch" strategy is preferred (merge vs rebase) so I will leave that to you.

provinzkraut:
> All checks are passing now. I am not sure what "update branch" strategy is preferred (merge vs rebase) so I will leave that to you.

Do whatever you prefer. Personally, I like rebase more, but it doesn't really make a difference here since we squash merge PRs anyway.

bunny-therapist:
Should I have been added to some "all contributors" file? I see that happening in other PRs but I don't know how this works.

provinzkraut:
Yes, you should have!

You can just request that yourself by making a comment like this:

@all-contributors add @bunny-therapist for code.

PR #3100, opened 2024-02-13
feat: Support `ResponseSpec(..., examples=[...])`

### LINKED ISSUE #3068: Enhancement: Add `ResponseSpec(examples=...)`
### Summary

Support adding custom examples for the response body.

The inbound parameters can be defined via `Parameter`, but the same does not work for the response body. `ResponseSpec` only has `generate_examples`, but it should support also `examples` for custom ones.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

tuukkamustonen:
Fixes #3068

Allows to define custom examples for the responses via `ResponseSpec`.

The examples set this way are always generated locally, for each response:

```json
{
  "paths": {
    "/": {
      "get": {
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {...},
                "examples": ...
```

Examples that go within the schema definition cannot be set by this.

### DIFF
--- a/litestar/_openapi/responses.py
+++ b/litestar/_openapi/responses.py
@@ -9,9 +9,10 @@ from operator import attrgetter
 from typing import TYPE_CHECKING, Any, Iterator
 
 from litestar._openapi.schema_generation import SchemaCreator
+from litestar._openapi.schema_generation.utils import get_formatted_examples
 from litestar.enums import MediaType
 from litestar.exceptions import HTTPException, ValidationException
-from litestar.openapi.spec import Example, OpenAPIResponse
+from litestar.openapi.spec import Example, OpenAPIResponse, Reference
 from litestar.openapi.spec.enums import OpenAPIFormat, OpenAPIType
 from litestar.openapi.spec.header import OpenAPIHeader
 from litestar.openapi.spec.media_type import OpenAPIMediaType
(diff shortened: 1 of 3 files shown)


litestar/_openapi/responses.py
```diff
@@ -240,14 +241,20 @@ def create_additional_responses(self) -> Iterator[tuple[str, OpenAPIResponse]]:
...
             )
-            schema = schema_creator.for_field_definition(
-                FieldDefinition.from_annotation(additional_response.data_container)
-            )
+            field_def = FieldDefinition.from_annotation(additional_response.data_container)
+            schema = schema_creator.for_field_definition(field_def)
+
+            examples: dict[str, Example | Reference] | None
+            if additional_response.examples:
+                examples = dict(get_formatted_examples(field_def, additional_response.examples))
```
tuukkamustonen, line 249: `OpenAPIMediaType.examples` declares `dict[str, ...]` but `get_formatted_examples()` returns `Mapping`, so had to convert to dict.

Alternatively, should I adjust `OpenAPIMediaType.examples` to accept `Mapping` instead?

provinzkraut: > Alternatively, should I adjust OpenAPIMediaType.examples to accept Mapping instead?

I think that's fine, yes

provinzkraut (approved):
Looks good, thanks a lot @tuukkamustonen!

PR #3103, opened 2024-02-13
feat: Allow re-usable `Router` instances

### LINKED ISSUE #3012: Enhancement: Allow re-use of router instances
### Summary

Mounting a router to multiple apps raises:

```
E               litestar.exceptions.http_exceptions.ImproperlyConfiguredException: 500: Router with path / has already been registered
```

There's been some discussion at least in https://github.com/litestar-org/litestar/issues/148 and https://github.com/litestar-org/litestar/issues/1054 and apparently it's not a bug, but intentional. But why does the limitation exist?

The problem is trying to re-use the same router instance when re-creating the application in tests. The first test passes, as the app constructs okay, but when the code is trying to re-build the application for the second test, the error is thrown. This totally kills the ability to create applications with altered configurations in tests.

This is only for the `Router` instances. If you map controllers directly, then there's no issue.

Could the limitation be lifted?

### Basic Example

```py
class MyHandler(Controller): ...

router = Router(route_handlers=[MyHandler])

def create_app():
    return Litestar(route_handlers=[router])

@pytest.fixture(scope="function")
def client():
    return TestClient(app=create_app())

def test_one(client):
    pass

def test_two(client):
    pass
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3012">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3012/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

tuukkamustonen:
Fixes #3012

It was not possible to re-attach a router instance once it was attached. This
makes that possible.

The router instance now gets deecopied when it's registered to another router.

The application startup performance gets a hit here, but the same approach is
already used for controllers and handlers, so this only harmonizes the
implementation.

### DIFF
--- a/litestar/router.py
+++ b/litestar/router.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from copy import copy
+from copy import copy, deepcopy
 from typing import TYPE_CHECKING, Any, Mapping, Sequence, cast
 
 from litestar._layers.utils import narrow_response_cookies, narrow_response_headers
@@ -309,14 +309,12 @@ class Router:
             return value(owner=self).to_handler()  # pyright: ignore
 
         if isinstance(value, Router):
-            if value.owner:
-                raise ImproperlyConfiguredException(f"Router with path {value.path} has already been registered")
-
             if value is self:
                 raise ImproperlyConfiguredException("Cannot register a router on itself")
 
-            value.owner = self
-            return value
+            router_copy = deepcopy(value)
+            router_copy.owner = self
+            return router_copy
 
         if isinstance(value, (ASGIRouteHandler, HTTPRouteHandler, WebsocketRouteHandler)):
             value.owner = self
--- a/tests/e2e/test_router_registration.py
+++ b/tests/e2e/test_router_registration.py
@@ -132,9 +132,7 @@ def test_register_validation_wrong_class() -> None:
 def test_register_already_registered_router() -> None:
     first_router = Router(path="/first", route_handlers=[])
     Router(path="/second", route_handlers=[first_router])
-
-    with pytest.raises(ImproperlyConfiguredException):
-        Router(path="/third", route_handlers=[first_router])
+    Router(path="/third", route_handlers=[first_router])
 
 
 def test_register_router_on_itself() -> None:


provinzkraut (approved):
Looks good, thanks!

In general, I think we should consider moving away from Routers, Controllers and Route Handlers as these transient objects and maybe only treat them as configuration objects (since that what they are).
I'm imagining they could be e.g. a dataclass that hold all the options, and when creating actual instances, we "bind" them explicitly, producing e.g. a `BoundRouter`, which takes in the specific configuration that has been resolved for it. We could even reduce this some more by only ever having `BoundRouteHandler` instances around, which are tied to their actual routes. 

This would also save us some pain regarding a lot of code duplication of parameters and make typing easier (we could create a base "layer config" data object that holds all config available on all layers etc.).

@litestar-org/members @litestar-org/maintainers?

JacobCoffee:
Do you have an example of what this would be like?

provinzkraut:
> Do you have an example of what this would be like?

I'd need to play around with the idea a bit an come up with a poc. But basically the idea is, since the layers are not actually needed at runtime, and are only used once at registration time to resolve the layered configs, we might as well just treat them as config objects in the first place. This would allow us to use a more functional style and to avoid the instance-attribute caching and "not yet initialised" checks and exceptions.

```python
# config all layers have (app, router, controller, handler)

@dataclass
class BaseLayerConfig:
 dependencies: Dependencies | None = None
 guards: Sequence[Guard] | None
 ...


@dataclass(frozen=True)
class HandlerConfig(BaseLayerConfig):
 path: str
 status_code: int = 200
 name: str | None = None
 sync_to_thread: bool | None = None
 http_method: HttpMethod | Sequence[HttpMethod | Method]

 def bind_fn(self, fn: AnyCallable) -> BoundHandlerConfig:
  return BoundHandlerConfig(fn=fn, config=self)

 def __call__(self, fn: AnyCallable) -> BoundHandlerConfig:
  return self.bind_fn(fn)


@dataclass
class get(HandlerConfig):
 http_method: HttpMethod = HttpMethod.GET


@dataclass
class BoundHandlerConfig:
  config: HandlerConfig
  fn: AnyCallable

  def resolve_dependencies(self, owner: Component) -> Dependencies:
   ...

  def to_handler(self, owner: Component) -> RouteHandler:
    # this is called on registration recursively and resolves all
    # the options from the upper layers and returns them as a route handler
    # object
    return RouteHandler(
        owner=owner, 
        fn=self.fn, 
        path=self.config.path,
        dependencies=self.resolve_dependencies(owner),
        ... # extract other options here
    )
```

PR #3106, opened 2024-02-13
fix(3082): Exception handlers extract details from non-litestar exceptions

### LINKED ISSUE #3082: Bug: Exception handler leaks internal exceptions
### Discussed in https://github.com/orgs/litestar-org/discussions/3060

<sup>Originally posted by **floxay** February  1, 2024</sup>
I have posted this [on Discord](https://discord.com/channels/919193495116337154/919193495690936353/1197749178399465542) before but it did not really go anywhere.
As I am unsure whether this should be a bug or enhancement I post it now on Github as a discussion and then can be converted to either if needed. (This is basically the same message as the one on Discord, nothing new has been added just structured slightly differently.)

The [documented behavior of exception handling](https://docs.litestar.dev/latest/usage/exceptions.html#exception-handling) mentions:
> If the errors are ***instances of [HTTPException](https://docs.litestar.dev/latest/reference/exceptions.html#litestar.exceptions.HTTPException)***, the responses will include the appropriate status_code.

Which is incorrect; what actually is happening is that Litestar tries to retrieve the value of the `status_code` field/property, regardless whether the object is an instance of `HTTPException` or not.
Code; https://github.com/litestar-org/litestar/blob/24095740e7c12220a8ca5cc14766cf0ec95764ca/litestar/middleware/exceptions/middleware.py#L125-L127 Notes from the same function above:
https://github.com/litestar-org/litestar/blob/24095740e7c12220a8ca5cc14766cf0ec95764ca/litestar/middleware/exceptions/middleware.py#L113-L116 One of these is definitely incorrect, and I would guess it is the documentation that is incomplete.

Before the documentation just gets changed to reflect what actually is happening I would also like to point out that this behavior is problematic.
I only noticed this issue because I got an unhandled Elasticsearch error back from the backend stringified in the exception details:
image
Related code: https://github.com/elastic/elasticsearch-py/blob/5014ce5337594f66040c81a2610220b1e8c0527e/elasticsearch/exceptions.py#L43-L46
This error also contains other information such as how (e.g. IP:port) the client is connected to elastic and headers which expose versions of used modules.
I do not think this kind of stuff generally should make it out to the client by default, debug mode was also not enabled.

For now I have added a custom exception handler for Elastic errors but on the long term this does not feel right, I would essentially need check out all external modules whether they have a `status_code` filed/property or not and if it would pose a problem.

I am not really sure what the right(?, better?, proper?) way to handle this would be, I've mentioned an exception class whitelist mechanic for the exception handler in the Discord comment as a potential solution but I imagine it would also affect the mounting of other ASGI apps, like Starlette apps and would be *very* breaking.</div>

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3082">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3082/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

provinzkraut:
Closes #3082.

Fix a bug where exception classes that had a `status_code` attribute would be treated as Litestar exceptions and details from them would be extracted and added to the exception response.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -421,6 +421,15 @@ class Litestar(Router):
         if self.pdb_on_exception:
             warn_pdb_on_exception()
 
+        try:
+            from starlette.exceptions import HTTPException as StarletteHTTPException
+
+            from litestar.middleware.exceptions.middleware import _starlette_exception_handler
+
+            config.exception_handlers.setdefault(StarletteHTTPException, _starlette_exception_handler)
+        except ImportError:
+            pass
+
         super().__init__(
             after_request=config.after_request,
             after_response=config.after_response,
--- a/litestar/middleware/exceptions/middleware.py
+++ b/litestar/middleware/exceptions/middleware.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, Any, Type, cast
 
 from litestar.datastructures import Headers
 from litestar.enums import MediaType, ScopeType
-from litestar.exceptions import WebSocketException
+from litestar.exceptions import HTTPException, LitestarException, WebSocketException
 from litestar.middleware.cors import CORSMiddleware
 from litestar.middleware.exceptions._debug_response import _get_type_encoders_for_request, create_debug_response
 from litestar.serialization import encode_json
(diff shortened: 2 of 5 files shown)


JacobCoffee (approved):

provinzkraut:
> Hey, this page does not exist.

Well, it's been almost a year. We don't keep these previews around forever

PR #3111, opened 2024-02-14
fix: close RedisStore client in case the with_client classmethod is used

### LINKED ISSUE #3083: Bug: declaring a RedisStore in the app triggers a 500 error in the 2nd set of parameters of a parametrized test
### Description

If you have a parametrized test that uses a function-scoped fixture that uses itself a session-scoped fixture and you declare a RedisStore in the app, the 2nd test fails with `RuntimeError: Event loop is closed`

The 2nd test passes in those cases:
1. all fixtures are session-based (just comment the client fixture)
2. there is no RedisStore in the app (or replace with a MemoryStore)

_No response_

### MCVE

```python
import msgspec
import pytest

from litestar.middleware.session.server_side import ServerSideSessionConfig
from litestar.stores.redis import RedisStore
from litestar.testing import AsyncTestClient


@get()
async def hget() -> int:
    return 1


class AppSettings(msgspec.Struct):
    debug: bool
    redis_url: str = "redis://localhost:6379"


def get_app(app_settings: AppSettings) -> Litestar:
    # setting up stores
    session_store = RedisStore.with_client(url=app_settings.redis_url)
    app = Litestar(route_handlers=[hget],
                   middleware=[
                       ServerSideSessionConfig().middleware,
                   ],
                   stores={"sessions": session_store, },  # comment this and 2nd test pass
                   debug=app_settings.debug
                   )
    return app


@pytest.fixture(scope="session")
def anyio_backend() -> str:
    return "asyncio"


@pytest.fixture(scope="session")
def app_settings_test():
    return AppSettings(debug=True)


@pytest.fixture(scope="session")
def app_test(app_settings_test: AppSettings):
    app = get_app(app_settings_test)
    yield app


@pytest.fixture  # add scope="session" and 2nd test pass
async def client(app_test: Litestar):
    async with AsyncTestClient(app=app_test) as c:
        yield c


@pytest.mark.anyio
@pytest.mark.parametrize("p1, p2",
                         [(1, 2), (3, 4)])
async def test_param(client: AsyncTestClient, p1: int, p2: int):
    response = await client.get("/")
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.5.5

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3083">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3083/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

euri10:
## Description

WIP, it doesn't pass the test I added from the mvce in the issue (this one could certainly be simplified) so my impleentation has obviously an issue.
I made the `RedisStore` an async context manager, I can see the `_on_shutdon` method being hit, but the `_redis` client is still not closed or it doesnt exists.

from your comment here:
https://discord.com/channels/919193495116337154/1152626430295933060/1207016578823102565

## Closes

Fixes #3083


litestar/stores/redis.py
```diff
@@ -64,6 +68,20 @@ def __init__(self, redis: Redis, namespace: str | None | EmptyType = Empty) -> N
...
+    async def __aenter__(self) -> RedisStore:
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        await self._on_shutdown()
```
provinzkraut, line 83: These should be added to the base store and simply do nothing

litestar/app.py
```diff
@@ -465,6 +465,9 @@ def __init__(
         self.stores: StoreRegistry = (
             config.stores if isinstance(config.stores, StoreRegistry) else StoreRegistry(config.stores)
         )
+        for _, store in self.stores._stores.items():
+            if not store._should_close_store_manually:
+                self._lifespan_managers.append(store)
```
provinzkraut, line 470: If you have implemented the context manager methods on the store base class.
```suggestion
        for store in self.stores._stores.values():
           self._lifespan_managers.append(store)
```

I'd also move the store registration and this handling logic somewhere within the init before the `self._lifespan_managers` are initialised, so you can simply append it to `config.lifespan`

litestar/stores/redis.py
```diff
@@ -30,6 +33,7 @@ def __init__(self, redis: Redis, namespace: str | None | EmptyType = Empty) -> N
                 ``None``. This will make :meth:`.delete_all` unavailable.
         """
         self._redis = redis
+        self._should_close_store_manually = True
```
provinzkraut, line 36: Maybe have this as a parameter in `__init__`, so you don't need to patch this from the outside?

litestar/stores/redis.py
```diff
@@ -68,19 +73,19 @@ def __init__(self, redis: Redis, namespace: str | None | EmptyType = Empty) -> N
...
 
-    async def _on_shutdown(self) -> None:
-        await self._redis.aclose()
+    async def _shutdown(self) -> None:
+        await self._redis.aclose(close_connection_pool=True)
 
-    async def __aenter__(self) -> RedisStore:
-        return self
+    async def __aenter__(self) -> None:
+        ...
```
provinzkraut, line 80:
```suggestion
```

litestar/stores/redis.py
```diff
@@ -18,23 +18,28 @@
...
         """Initialize :class:`RedisStore`
 
         Args:
             redis: An :class:`redis.asyncio.Redis` instance
             namespace: A key prefix to simulate a namespace in redis. If not given,
                 defaults to ``LITESTAR``. Namespacing can be explicitly disabled by passing
                 ``None``. This will make :meth:`.delete_all` unavailable.
+            automatic_store_lifetime: If ``True``, the store lifetime will be handled automatically,
+            works only if you're using the :meth:`.with_client` method.
+            By default it is False and in that case you will have to handle the store lifetime manually.
```
provinzkraut, line 38:
```suggestion
        self, redis: Redis, namespace: str | None | EmptyType = Empty, handle_client_shutdown: bool = False
    ) -> None:
        """Initialize :class:`RedisStore`

        Args:
            redis: An :class:`redis.asyncio.Redis` instance
            namespace: A key prefix to simulate a namespace in redis. If not given,
                defaults to ``LITESTAR``. Namespacing can be explicitly disabled by passing
                ``None``. This will make :meth:`.delete_all` unavailable.
            handle_client_shutdown: If ``True``, handle the shutdown of the `redis` instance automatically during the store's lifespan. Should be set to `True` unless the shutdown is handled externally
```

I think this way is a bit more clear

litestar/stores/redis.py
```diff
@@ -92,6 +97,7 @@ def with_client(
         username: str | None = None,
         password: str | None = None,
         namespace: str | None | EmptyType = Empty,
+        automatic_store_lifetime: bool = True,
```
provinzkraut, line 100:
```suggestion
```

I don't think we need this option here; If we create the client, we should always close it properly

provinzkraut (changes requested):

euri10:
you're reviewing / holding mye hand faster than I can type :)
but there's progress, the test I wrote from the mcve passes,
I'll check later on why it broke 46 existing ones :+1:

litestar/stores/base.py
```diff
@@ -76,6 +78,19 @@ async def expires_in(self, key: str) -> int | None:
...
+        ...
+
+    @abstractmethod
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        ...
```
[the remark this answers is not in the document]
euri10: just had to ignore the B027 to please the linter

provinzkraut:
> you're reviewing / holding mye hand faster than I can type :) 

Sorry :grimacing:

euri10:
> > you're reviewing / holding mye hand faster than I can type :)
> 
> Sorry 😬

no worries that was fun, I almost wondered if you were not a LLM at sone point :rofl:

litestar/app.py
```diff
@@ -388,7 +388,13 @@ def __init__(
 
         self._openapi_schema: OpenAPI | None = None
         self._debug: bool = True
+        self.stores: StoreRegistry = (
+            config.stores if isinstance(config.stores, StoreRegistry) else StoreRegistry(config.stores)
+        )
         self._lifespan_managers = config.lifespan
+        for store in self.stores._stores.values():
+            if store.handle_client_shutdown:
+                self._lifespan_managers.append(store)
```
provinzkraut, line 397: I think my earlier suggestion was bad. 
What I meant was:

- Always add the stores as lifespan managers here
- Handle the `handle_client_shutdown` inside `RedisStore.__aexit__`

That way each store can decide how it wants to handle shutdown if at all

euri10: gotcha, I'll take that back later on, you obviously know in and out the codebase, that's very impressive, the LLM-joke was meant as a compliment by the way !

provinzkraut:
> no worries that was fun, I almost wondered if you were not a LLM at sone point 🤣

I do wonder that too at times. Aren't we all just fancy LLMs at some level? :eyes:

litestar/app.py
euri10: > I think my earlier suggestion was bad. What I meant was:
> 
> * Always add the stores as lifespan managers here
> * Handle the `handle_client_shutdown` inside `RedisStore.__aexit__`
> 
> That way each store can decide how it wants to handle shutdown if at all

done, it's better since it clears those 46 breaks in tests by keeping existing stores the same.

also, the current test is like....complicated, but in the meantime the fact it popped up for me is exactly how we figured the closing wasn't performed, so I'd be tempted to keep it as is, but happ

time for some linter stuff now

euri10:
> > no worries that was fun, I almost wondered if you were not a LLM at sone point 🤣
> 
> I do wonder that too at times. Aren't we all just fancy LLMs at some level? 👀

that kind of would be Nick Bolstrom's theory in "Are you living in a computer simulation?"

euri10:
I might have introduced some flakiness :)
Will take a look tomorrow

euri10:
> I might have introduced some flakiness :) Will take a look tomorrow

ok in fact it's just the test I added doesnt integrate well in the suite, need to find a way to see how to keep it without issues

euri10:
ok I removed my initial test from the mcve that passes with the change made to the lifespan, reason is twofold.

one, as you can notice in the suite, all instances where with_client is used are patched, so if there is a test to be made, one has to follow the same rule or we may end up having unclosed ressources: I therefore think this test is problematic: https://github.com/litestar-org/litestar/blob/323e23e51508cf142074447098b48ddcc2a67ad0/tests/unit/test_stores.py#L242-L245
because it doesnt patch the with_client and will leave things unclosed, the other line is fine

Also, I modified the redis_client fixture: https://github.com/litestar-org/litestar/pull/3111/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128L307-R315

the previous one is problematic because it uses a sync redis instance to flushall and never closes it, one can just use the async one from the fixture and flushall in the finally for the same effect without the unintended consequences,

wdyt ?

provinzkraut:
> the previous one is problematic because it uses a sync redis instance to flushall and never closes it, one can just use the async redis instance from the fixture and flushall in the finally for the same effect without the unintended consequences,
> 
> wdyt ?

I think the proper fix would be to close the sync `Redis` instance. The reason we flush at the beginning and not the end is that if we're flushing at the end, and something goes wrong, we'd be affecting other tests. If we flush at the beginning, every test *starts* with an empty DB. 

Also, I'm surprised that calling `.flush()` after `.close()` does not cause any issues.

euri10:
> > the previous one is problematic because it uses a sync redis instance to flushall and never closes it, one can just use the async redis instance from the fixture and flushall in the finally for the same effect without the unintended consequences,
> > wdyt ?
> 
> I think the proper fix would be to close the sync `Redis` instance. The reason we flush at the beginning and not the end is that if we're flushing at the end, and something goes wrong, we'd be affecting other tests. If we flush at the beginning, every test _starts_ with an empty DB.
> 
> Also, I'm surprised that calling `.flush()` after `.close()` does not cause any issues.

so @provinzkraut I sent a PR fixing the fixture think properly I think here: https://github.com/litestar-org/litestar/pull/3111

questions now on this one:
1. do we deal here with the only test using with_client that doesnt uses mocks.
2. should I add a test checking if the client is closed when using that classmethod, provided there is no is_closed method on the redis client, so I wouldn't know how to test it

tests/conftest.py
```diff
@@ -304,12 +304,7 @@ def get_logger() -> GetLogger:
...
-    yield client
-    try:
-        await client.aclose()  # type: ignore[attr-defined]
-    except RuntimeError:
-        pass
-    finally:
+async def redis_client(docker_ip: str, redis_service: None) -> AsyncRedis:
+    async with AsyncRedis(host=docker_ip, port=6397) as client:
         await client.flushall()
+        return client
```
provinzkraut, line 310: I don't understand why this works. The context manager closes the client immediately before returning so won't the client be closed for all the tests?

euri10: now you mention it this is weird indeed,
https://github.com/litestar-org/litestar/pull/3117 is the change for the redis fixture and all tests pass fine, yielding the client and nothing works,

euri10: given https://docs.python.org/3/reference/compound_stmts.html#async-with

we have then
```
    async with AsyncRedis(host=docker_ip, port=6397) as client:
        await client.flushall()
        return client
```
is in fact (I may have syntax errors here it's just to understand the flow)

```
hit_except = False

try:
    client = AsyncRedis.__aenter__()
    await client.flushall()
    return client
except:
    hit_except = True
    if not await aexit(manager, *sys.exc_info()):
        raise
finally:
    if not hit_except:
        await AsyncRedis.__aexit()
```

litestar/stores/redis.py
```diff
@@ -84,6 +103,7 @@ def with_client(
             username: Redis username to use
             password: Redis password to use
             namespace: Virtual key namespace to use
+            automatic_store_lifetime: If ``True``, the store lifetime will be handled automatically, it is the default with this method.
```
provinzkraut, line 106:
```suggestion
```

provinzkraut (approved):

litestar/stores/redis.py
```diff
@@ -93,14 +112,22 @@ def with_client(
...
+            namespace=namespace,
+            handle_client_shutdown=True,
+        )
 
     def with_namespace(self, namespace: str) -> RedisStore:
         """Return a new :class:`RedisStore` with a nested virtual key namespace.
         The current instances namespace will serve as a prefix for the namespace, so it
         can be considered the parent namespace.
         """
-        return type(self)(redis=self._redis, namespace=f"{self.namespace}_{namespace}" if self.namespace else namespace)
```
euri10, line 103: I didn't add a test on this as I wouldn't know how to do this, the reason it is necessary is that if you are using a namedspace store and with_client then in tests this attribute will get overriden

euri10:
I think it should be ready, I'm not pleased with the fixtures as-is to be fair, so I reverted my changes on this part but that shouldn't prevent the PR to be looked at ;)

PR #3118, opened 2024-02-15
fix: Description not set for UUID based path parameters in OpenAPI

### LINKED ISSUE #2967: Bug: Description not set for UUID based path parameters in OpenAPI
### Description

The value set in `Parameter(description="UUID ID")` is not used for the actual description in the OpenAPI schema if the path parameter is of type `UUID`. I did some digging and found out this is due to the presence of the following code. https://github.com/litestar-org/litestar/blob/44809182b1b62a8da5b7c5bec3b22987581779d5/litestar/_openapi/schema_generation/schema.py#L525-L532

Since description for UUID is already set https://github.com/litestar-org/litestar/blob/44809182b1b62a8da5b7c5bec3b22987581779d5/litestar/_openapi/schema_generation/schema.py#L120 the code above makes it so that description from the user defined `Parameter` is not set again.

#### Proposed Fix 1 (Breaking / debatably breaking?)
Remove "description" in https://github.com/litestar-org/litestar/blob/44809182b1b62a8da5b7c5bec3b22987581779d5/litestar/_openapi/schema_generation/schema.py#L120 such that it appears like so
```suggestion py
UUID: Schema(type=OpenAPIType.STRING, format=OpenAPIFormat.UUID)
```
#### Proposed Fix 2 (Special Casing until next major release then replace with Proposed Fix 1)
Change https://github.com/litestar-org/litestar/blob/44809182b1b62a8da5b7c5bec3b22987581779d5/litestar/_openapi/schema_generation/schema.py#L531 such that it appears like so
```suggestion py
if schema_key == "description" or getattr(schema, schema_key, None) is None:
```

@peterschutt suggested special casing with comments explaining the behavior would be ideal. Thoughts?

_No response_

### MCVE

```python
from uuid import UUID
from typing_extensions import Annotated
from litestar import get

@get("str/{id:str}")
async def str_path(id: Annotated[str, Parameter(description="String ID")]) -> str:
    return id

@get("uuid/{id:uuid}")
async def uuid_path(id: Annotated[UUID, Parameter(description="UUID ID")]) -> UUID:
    return id

with create_test_client([str_path, uuid_path]) as client:
    # correct, currently passes, should pass
    assert client.app.openapi_schema.paths["/str/{id}"].get.parameters[0].description == "String ID"

    # wrong, currently passes, should fail
    assert client.app.openapi_schema.paths["/uuid/{id}"].get.parameters[0].description == "Any UUID string"

    # expected, currently fails, should pass
    assert client.app.openapi_schema.paths["/uuid/{id}"].get.parameters[0].description == "UUID ID"
```

### Steps to reproduce

```bash
1. Save as "mcve.py"
2. Run `python mcve.py`
3. Second assert that currently passes, should fail.
4. Third assert that currently fails, should pass.
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

cc45c1132584210250dd725595612c9c95c4bf68 (main as of issue creation)

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2967">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2967/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Alc-Alc:
## Description

- Resolved the bug where the description was not set for UUID-based path parameters in OpenAPI due to the reason mentioned in the issue.

## Closes
Closes #2967

### DIFF
--- a/tests/unit/test_openapi/test_parameters.py
+++ b/tests/unit/test_openapi/test_parameters.py
@@ -337,7 +337,9 @@ def test_uuid_path_description_generation() -> None:
     async def uuid_path(id: Annotated[UUID, Parameter(description="UUID ID")]) -> UUID:
         return id
 
-    with create_test_client([str_path, uuid_path]) as client:
+    with create_test_client(
+        [str_path, uuid_path], openapi_config=OpenAPIConfig(title="Test API", version="1.0.0")
+    ) as client:
         response = client.get("/schema/openapi.json")
         assert response.json()["paths"]["/str/{id}"]["get"]["parameters"][0]["description"] == "String ID"
         assert response.json()["paths"]["/uuid/{id}"]["get"]["parameters"][0]["description"] == "UUID ID"
--- a/tests/unit/test_openapi/test_parameters.py
+++ b/tests/unit/test_openapi/test_parameters.py
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING, List, Optional, Type, cast
+from uuid import UUID
 
 import pytest
 from typing_extensions import Annotated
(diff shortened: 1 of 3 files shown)


tests/unit/test_openapi/test_parameters.py
```diff
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING, List, Optional, Type, cast
+from uuid import UUID
```
Alc-Alc, line 2: I could import this under `TYPE_CHECKING` but then I would have to import it in runtime (inside the test) and add it using `signature_types`. So I imported it here, lmk if this has to be changed

tests/unit/test_openapi/test_parameters.py
```diff
@@ -324,3 +325,19 @@ async def index(
         assert response.json()["paths"]["/"]["get"]["parameters"][0]["examples"] == {
             "text-example-1": {"summary": "example summary", "value": "example value"}
         }
+
+
+def test_uuid_path_description_generation() -> None:
```
Alc-Alc, line 330: I am also not sure if this is where the test should go 🙃. The test function above what I wrote seems to test something relevant so I added it here

cofin (approved):

PR #3121, opened 2024-02-19
feat: create a rate-limit configuration from a list of values

### LINKED ISSUE #1262: Enhancement: Rate Limit Config
About  `class starlite.middleware.rate_limit.RateLimitConfig`

this， `rate_limit: Tuple[Literal['second', 'minute', 'hour', 'day'], int]`

A tuple containing a time unit (second, minute, hour, day) and quantity, e.g. (“day”, 100) or (“minute”, 5)  could be changed to  [ (“day”, 100) ,(“minute”, 5),("second",1)]?

ch-iv:
## Description

- Adds ability to define a rate-limit configuration using a list of values

### DIFF
--- a/litestar/middleware/rate_limit.py
+++ b/litestar/middleware/rate_limit.py
@@ -55,8 +55,11 @@ class RateLimitMiddleware(AbstractMiddleware):
         )
         self.check_throttle_handler = cast("Callable[[Request], Awaitable[bool]] | None", config.check_throttle_handler)
         self.config = config
-        self.max_requests: int = config.rate_limit[1]
-        self.unit: DurationUnit = config.rate_limit[0]
+
+        self.offsets: list[tuple[int, int]] = [(DURATION_VALUES[unit], n) for unit, n in config.rate_limit]
+        self.offsets.sort(key=lambda x: x[0])
+        self.max_offset = max(self.offsets, key=lambda x: x[0])[0]
+        self.max_requests: int = 1
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         """ASGI callable.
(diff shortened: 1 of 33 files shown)


litestar/middleware/rate_limit.py
```diff
@@ -55,8 +55,14 @@ def __init__(self, app: ASGIApp, config: RateLimitConfig) -> None:
         )
         self.check_throttle_handler = cast("Callable[[Request], Awaitable[bool]] | None", config.check_throttle_handler)
         self.config = config
-        self.max_requests: int = config.rate_limit[1]
-        self.unit: DurationUnit = config.rate_limit[0]
```
provinzkraut, line 59: We'll need to keep these attributes to remain backwards compatible. Otherwise this would be a breaking change.

litestar/middleware/rate_limit.py
```diff
@@ -72,28 +78,47 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+                    limit_most_restrictive=limit_most_restrictive,
+                    requests_remaining=limit_most_restrictive[1] - n - 1,
+                )
 
         await self.app(scope, receive, send)  # pyright: ignore
 
-    def create_send_wrapper(self, send: Send, cache_object: CacheObject) -> Send:
+    def create_send_wrapper(
+        self, send: Send, cache_object: CacheObject, limit_most_restrictive: tuple[int, int], requests_remaining: int
+    ) -> Send:
```
provinzkraut, line 114: This is a breaking change as well

litestar/middleware/rate_limit.py
```diff
@@ -187,26 +216,29 @@ async def should_check_request(self, request: Request[Any, Any, Any]) -> bool:
             return await self.check_throttle_handler(request)
         return True
 
-    def create_response_headers(self, cache_object: CacheObject) -> dict[str, str]:
+    def create_response_headers(
+        self, cache_object: CacheObject, limit_most_restrictive: tuple[int, int], requests_remaining: int
+    ) -> dict[str, str]:
```
provinzkraut, line 221: Also a breaking change

provinzkraut (changes requested):
Hey @ch-iv, thanks for your contribution! Unfortunately, this can't go as it is now, since it would be a breaking change; It alters parts of the public API in a way that's not backwards compatible.

You can either try to make it work in a non-breaking way, or stash this PR for the next major version, where we could do a breaking change. 

A suggestion for how this could be made non-breaking:

Create a new middleware (e.g. `MultiRateLimitingMiddleware`) instead of modifying the existing one. Both could share a base class for the common functionalities, and then they could be used depending on the configuration provided.

JacobCoffee:
Are we able to address concerns and rebase against the latest `develop` branch to finish this one up?
@ch-iv

ch-iv:
I am busy the coming week, but I can address it next week if that's ok. If time is an issue -- someone else can take over.

peterschutt:
What should we aim for here? Non-breaking in a minor, or v3?

ch-iv:
I think it is best kept for v3. Having two rate limit middleware classes is redundant and probably confusing to the user.

JacobCoffee:

Issue #2203, opened 2023-08-23
Enhancement: expose `request_class` to endpoint decorators and `Controller` classes


cofin:
### Summary

Currently, `request_class` is only available on the Litestar application class.  It would be nice if this was exposed to endpoints and Controllers for overriding.

PR #3123, opened 2024-02-22
feat: expose request_class to other layers

kedod:
## Description
- Allow to define `request_class` on router, controller and handler layer.

## Closes [2203](https://github.com/litestar-org/litestar/issues/2203)

TODO:
- docs
- tests


provinzkraut (changes requested):
Looks good, thanks @kedod! Can you add some tests? Then it's good to go in :)

PR #3127, opened 2024-02-22
feat: 3116 enhancement session middleware

### LINKED ISSUE #3116: Enhancement: Session Middleware should create session id right away
### Summary

Follow-up on: https://github.com/orgs/litestar-org/discussions/3112#discussioncomment-8477410

Currently, the `SessionMiddleware` of the `SessionAuth` backend creates the session id in the response wrapper and not when the session is first created, e.g., on user login. This is counter-intuitive since it gives the impression the session is created within the route handler when it's actually not. Furthermore, this makes it difficult or even impossible to add any custom handling of the session id during the login procedure since the session id does not exist yet.

Current state:
```python
@post("/login")
async def login(
        data: UserLogin,
        request: Request,
) -> User:
    user = handle_auth(data)
    request.set_session(
        {"user-id": user.id, "foo": "bar"}
    )
    logger.info(f"Session content: {request.session}")  # This will have content
    logger.info(f"Cookies: {request.cookies}")  # This will be empty. The send wrapper will add a session id here once created
    return user
```

### Basic Example

My first best guess:

On registering the SessionAuth on the app, add a dependency so that it can be accessed as kwarg anywhere (not sure if the reference should be to that objectm, the session backend, or the middleware). On this reference, the user can call a method to pass the session data. This method will:
- create a session id
- store the session data in the store with that key
- add the session data to the connection scope
- potentially also already add it to the cookie? (not sure if this can be done here or should still be done in the send wrapper)
- return the session id

So for the user, it would look something like this:
```python
@post("/login")
async def login(
        data: UserLogin,
        request: Request,
        session_backend: SessionBackend
) -> User:
    user = handle_auth(data)
    session_backend.create_session(request=request, session_data={"user-id": user.id, "foo": bar})
    return user
```

### Drawbacks and Impact

_No response_

### Unresolved questions

Only just started digging through the code to understand the complete flow so might have missed something. Yet, the user perspective should be clear. 

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3116">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3116/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

aranvir:
## Description

- For server side sessions, the session id is now generated before the route handler. Thus, on first visit, a session id will be available inside the route handler's scope instead of afterwards
- A new abstract method `get_session_id` was added to `BaseSessionBackend` since this method will be called for both ClientSideSessions and ServerSideSessions. Only for ServerSideSessions it will return an actual id.
- Using `request.set_session(...)` will return the session id for ServerSideSessions and None for ClientSideSessions
- The session auth MiddlewareWrapper now refers to the Session Middleware via the configured backend, instead of it being hardcoded

Looking forward to your feedback!

## Closes
Closes #3116

### DIFF
--- a/litestar/connection/base.py
+++ b/litestar/connection/base.py
@@ -9,6 +9,7 @@ from litestar.datastructures.state import State
 from litestar.datastructures.url import URL, Address, make_absolute_url
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.types.empty import Empty
+from litestar.utils.empty import value_or_default
 from litestar.utils.scope.state import ScopeState
 
 if TYPE_CHECKING:
@@ -277,22 +278,19 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
         """
         return self.app.get_logger()
 
-    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> str | EmptyType:
+    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> None:
         """Set the session in the connection's ``Scope``.
 
         If the :class:`SessionMiddleware <.middleware.session.base.SessionMiddleware>` is enabled, the session will be added
         to the response as a cookie header.
-        If a :class:`ServerSideSessionConfig <.middleware.session.server_side.ServerSideSessionConfig>` is being used,
-        the corresponding session id will be returned.
 
         Args:
             value: Dictionary or pydantic model instance for the session data.
 
         Returns:
-            Session id string if one exists, else None.
+            None
         """
         self.scope["session"] = value
-        return self.get_session_id()
 
     def clear_session(self) -> None:
         """Remove the session from the connection's ``Scope``.
@@ -306,11 +304,8 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
         self.scope["session"] = Empty
         self._connection_state.session_id = Empty
 
-    def set_session_id(self, value: str) -> None:
-        self._connection_state.session_id = value
-
-    def get_session_id(self) -> str | EmptyType:
-        return self._connection_state.session_id
+    def get_session_id(self) -> str | None:
+        return value_or_default(value=self._connection_state.session_id, default=None)
 
     def url_for(self, name: str, **path_parameters: Any) -> str:
         """Return the url for a given route handler name.
--- a/litestar/connection/base.py
+++ b/litestar/connection/base.py
@@ -282,7 +282,7 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
 
         If the :class:`SessionMiddleware <.middleware.session.base.SessionMiddleware>` is enabled, the session will be added
         to the response as a cookie header.
-        If a :class:`ServerSideSessionConfig <.middleware.session.server_side.ServerSideSessionConfig` is being used,
+        If a :class:`ServerSideSessionConfig <.middleware.session.server_side.ServerSideSessionConfig>` is being used,
         the corresponding session id will be returned.
 
         Args:
--- a/litestar/connection/base.py
+++ b/litestar/connection/base.py
@@ -277,19 +277,22 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
         """
         return self.app.get_logger()
 
-    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> None:
+    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> str | None:
         """Set the session in the connection's ``Scope``.
 
         If the :class:`SessionMiddleware <.middleware.session.base.SessionMiddleware>` is enabled, the session will be added
         to the response as a cookie header.
+        If a :class:`ServerSideSessionConfig <.middleware.session.server_side.ServerSideSessionConfig` is being used,
+        the corresponding session id will be returned.
 
         Args:
             value: Dictionary or pydantic model instance for the session data.
 
         Returns:
-            None.
+            Session id string if one exists, else None.
         """
         self.scope["session"] = value
+        return self.scope.get("session_id", None)
 
     def clear_session(self) -> None:
         """Remove the session from the connection's ``Scope``.
--- a/litestar/connection/base.py
+++ b/litestar/connection/base.py
@@ -277,7 +277,7 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
         """
         return self.app.get_logger()
 
-    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> str | None:
+    def set_session(self, value: dict[str, Any] | DataContainerType | EmptyType) -> str | EmptyType:
         """Set the session in the connection's ``Scope``.
 
         If the :class:`SessionMiddleware <.middleware.session.base.SessionMiddleware>` is enabled, the session will be added
@@ -292,7 +292,7 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
             Session id string if one exists, else None.
         """
         self.scope["session"] = value
-        return self.scope.get("session_id", None)
+        return self.get_session_id()
 
     def clear_session(self) -> None:
         """Remove the session from the connection's ``Scope``.
@@ -304,6 +304,13 @@ class ASGIConnection(Generic[HandlerT, UserT, AuthT, StateT]):
             None.
         """
         self.scope["session"] = Empty
+        self._connection_state.session_id = Empty
+
+    def set_session_id(self, value: str) -> None:
+        self._connection_state.session_id = value
+
+    def get_session_id(self) -> str | EmptyType:
+        return self._connection_state.session_id
 
     def url_for(self, name: str, **path_parameters: Any) -> str:
         """Return the url for a given route handler name.
--- a/litestar/middleware/session/base.py
+++ b/litestar/middleware/session/base.py
@@ -146,6 +146,17 @@ class BaseSessionBackend(ABC, Generic[ConfigT]):
         return cast("dict[str, Any]", decode_json(value=data))
 
     @abstractmethod
+    def get_session_id(self, connection: ASGIConnection) -> str | None:
+        """Fetch or generate session id, if applicable.
+
+        Args:
+            connection: Originating ASGIConnection containing the scope
+
+        Returns:
+            Session id str or None, if not applicable.
+        """
+
+    @abstractmethod
     async def store_in_message(self, scope_session: ScopeSession, message: Message, connection: ASGIConnection) -> None:
         """Store the necessary information in the outgoing ``Message``
 
@@ -241,5 +252,6 @@ class SessionMiddleware(AbstractMiddleware, Generic[BaseSessionBackendT]):
 
         connection = ASGIConnection[Any, Any, Any, Any](scope, receive=receive, send=send)
         scope["session"] = await self.backend.load_from_connection(connection)
+        scope["session_id"] = self.backend.get_session_id(connection)
 
         await self.app(scope, receive, self.create_send_wrapper(connection))
--- a/litestar/middleware/session/base.py
+++ b/litestar/middleware/session/base.py
@@ -146,14 +146,14 @@ class BaseSessionBackend(ABC, Generic[ConfigT]):
         return cast("dict[str, Any]", decode_json(value=data))
 
     @abstractmethod
-    def get_session_id(self, connection: ASGIConnection) -> str | None:
-        """Fetch or generate session id, if applicable.
+    def get_session_id(self, connection: ASGIConnection) -> str:
+        """Try to fetch session id from connection ScopeState. If one does not exist, generate one.
 
         Args:
             connection: Originating ASGIConnection containing the scope
 
         Returns:
-            Session id str or None, if not applicable.
+            Session id str or "null" if the concept of a session id does not apply.
         """
 
     @abstractmethod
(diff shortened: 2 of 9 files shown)


aranvir:
Okay, I see there are quite a few failures. Gonna collect my questions here on how to solve them:

- Pyright flags that "session_id" is not a valid key in Scope. Can I just add it as `session_id: str | None` or should it be of a different type spec?
- Mypy also flags some issues with "session_id" being None when it should be string. Technically, this is only flagged on server side where it should never be None. Should I change the code instead for client side to return "null" so that the type is always `str`?

Will take some more time to work through the failed tests.

guacs:
> Okay, I see there are quite a few failures. Gonna collect my questions here on how to solve them:
> 
> * Pyright flags that "session_id" is not a valid key in Scope. Can I just add it as `session_id: str | None` or should it be of a different type spec?
> * Mypy also flags some issues with "session_id" being None when it should be string. Technically, this is only flagged on server side where it should never be None. Should I change the code instead for client side to return "null" so that the type is always `str`?
> 
> Will take some more time to work through the failed tests.

Regarding `session_id` not being part of the scope, that's my fault. In my example I had shown it being added to the scope because I was being lazy :P The better way to do this is to set this to `ScopeState` (in `litestar.utils.scope.state` which is what we use to maintain all the state that Litestar needs to maintain.

Regarding your second point, while `str | None` is the accurate type hint, I think it may make the life of users far easier if we could just annotate it as `str` so I think your idea of returning `null` for the client side sessions is a good idea.

aranvir:
> Regarding `session_id` not being part of the scope, that's my fault. In my example I had shown it being added to the scope because I was being lazy :P The better way to do this is to set this to `ScopeState` (in `litestar.utils.scope.state` which is what we use to maintain all the state that Litestar needs to maintain.
> 
> Regarding your second point, while `str | None` is the accurate type hint, I think it may make the life of users far easier if we could just annotate it as `str` so I think your idea of returning `null` for the client side sessions is a good idea.

I think I'm gonna need a bit more help here. I thought I'd have to add `session_id` to `litestar.types.asgi_types.BaseScope` or `.HTTPScope` (since we probably dont want to have this for `.WebSocketScope`). You suggest to add it to `ScopeState` instead, but how and where is that used? I see that `litestar.connection.base.ASGIConnection` has both `state` (from `self.scope.get("state")` and `_connection_state` where only the latter is `ScopeState`... but it is build from the scope. Being new to the framework, it is not 100% clear to me how this scope state should be used and how it can be reached throughout the lifetime of a request.

guacs:
@aranvir the reason that we can't add it directly to `HTTPScope` because that's taken directly from [the ASGI specification](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope). The `state` is what's meant to be used for anything specific to the application, or framework.

While we can use `scope["state"]["session_id"]`, the reason I suggest against that is because users are allowed to store stuff in `state` and if someone unkowingly overrides the `session_id`, then it will mess things up. To maintain state internal to Litestar we use the `ScopeState`.

The `Request` class has a `_connection_state` attribute which is an instance of `ScopeState`. You can see examples of how it's used in the various properties like `accept`, `content_type` etc. ([ref](https://github.com/litestar-org/litestar/blob/3e5c179e714bb074bae13e02d10e2f3f51e24d5c/litestar/connection/request.py#L94-L107)).  Also, you can instantiate a new `ScopeState` by using the `from_scope` class method and just pass in the scope. You would need to this in the middleware when setting the `session_id` before we process the request.

NOTE: All the properties on the `Request` class cache the attributes that we retrieve from `ScopeState`, but I don't think that's needed for retrieving the session ID since I doubt that there's going to be any real benefit to caching the `session_id`.

aranvir:
Hi @guacs changed it to storing the session_id in the ScopeState instead of the scope. Is this implementation okay? Also fixed the linting and testing.

Regarding the other failed pipeline jobs:
- Validate PR target: Should I re-open this pull request to point it to the `develop` branch? For sake of keeping these threads I would continue working this pull-request until all checks pass and then I can make a new pull request. Or edit the target - don't know what makes more sense.
- codecov: Never worked with this before so I also don't really understand what it is complaining about. Does this track how much of the code is tested? I barely changed the failing file, and what I added was just so the tests would pass AND mypy would stop complaining about `None` not having the `get_session_id` method.

guacs:
> Hi @guacs changed it to storing the session_id in the ScopeState instead of the scope. Is this implementation okay? Also fixed the linting and testing.
> 
> Regarding the other failed pipeline jobs:
> 
> * Validate PR target: Should I re-open this pull request to point it to the `develop` branch? For sake of keeping these threads I would continue working this pull-request until all checks pass and then I can make a new pull request. Or edit the target - don't know what makes more sense.
> * codecov: Never worked with this before so I also don't really understand what it is complaining about. Does this track how much of the code is tested? I barely changed the failing file, and what I added was just so the tests would pass AND mypy would stop complaining about `None` not having the `get_session_id` method.

I don't think you need to re-open the pull request. You can retarget it to the `develop` branch. As for codecov, you can look at details of the test and you should be able to find a link to the codecov website for this PR where it'll show which lines that aren't being covered in the tests.

I'll take a look at this later today :)

litestar/connection/base.py
```diff
@@ -277,19 +277,22 @@ def logger(self) -> Logger:
...
 
         Args:
             value: Dictionary or pydantic model instance for the session data.
 
         Returns:
-            None.
+            Session id string if one exists, else None.
         """
         self.scope["session"] = value
+        return self.get_session_id()
```
provinzkraut, line 295: I don't think returning this here is the best solution, especially if a public interface already exists. IMO this method should be unchanged

litestar/middleware/session/base.py
```diff
@@ -145,6 +145,17 @@ def deserialize_data(data: Any) -> dict[str, Any]:
         """
         return cast("dict[str, Any]", decode_json(value=data))
 
+    @abstractmethod
+    def get_session_id(self, connection: ASGIConnection) -> str:
```
provinzkraut, line 149:
```suggestion
    def get_session_id(self, connection: ASGIConnection) -> str | None:
```

Makes more sense than an arbitrary `null` string.

litestar/middleware/session/client_side.py
```diff
@@ -206,6 +206,9 @@ async def load_from_connection(self, connection: ASGIConnection) -> dict[str, An
                 return self.load_data(data)
         return {}
 
+    def get_session_id(self, connection: ASGIConnection) -> str:
+        return "null"
+
```
provinzkraut, line 211:
```suggestion
    def get_session_id(self, connection: ASGIConnection) -> str | None:
        return None

```
see above

provinzkraut (changes requested):

litestar/connection/base.py
guacs: I'm fine with this as well.

litestar/connection/base.py
```diff
@@ -301,6 +304,13 @@ def clear_session(self) -> None:
             None.
         """
         self.scope["session"] = Empty
+        self._connection_state.session_id = Empty
+
+    def set_session_id(self, value: str) -> None:
+        self._connection_state.session_id = value
```
guacs, line 310: I don't think this should be added. Users shouldn't be allowed to change the session ID.

litestar/middleware/session/base.py
```diff
@@ -241,5 +252,6 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         connection = ASGIConnection[Any, Any, Any, Any](scope, receive=receive, send=send)
         scope["session"] = await self.backend.load_from_connection(connection)
+        connection.set_session_id(self.backend.get_session_id(connection))
```
guacs, line 255: Related to what I mentioned above. The setting of the session ID should be done manually.

guacs (changes requested):
@aranvir could you add tests as well?

litestar/middleware/session/base.py
aranvir: Okay. I added this method to provide a public interface for setting property in `connection._connection_state`. I expect mypy or pyright to complain when I replace this with `connection._connection_state.session_id`, so I assume it's okay to add a linting exception in that case, right?

litestar/middleware/session/base.py
aranvir: How does `None` and `Empty` work together? All `SessionScope` attributes are `Empty` by default, yet for `session_id` we want to make it `str | None`. Should it then also be that type in `SessionScope`?

Otherwise, mypy will report type conflicts.

aranvir:
> @aranvir could you add tests as well?

@guacs what tests do you have in mind? The only one that I can come up is to modify the [middleware integration test](https://github.com/litestar-org/litestar/blob/main/tests/unit/test_middleware/test_session/test_integration.py) to confirm that the session id is set in the route handler scope - and that it is `not None` for the server side config and `None` for the client side config. But I did not see any assert statements in the route handlers of the other tests, so I wanted to check if this really goes into the right direction.

litestar/middleware/session/base.py
guacs: The type in `SessionScope` should be `str | None | Empty` and then you could use the `litestar.utils.value_or_default` helper function that will get the value if it's not `Empty`, else it'll return a default value.

guacs:
> > @aranvir could you add tests as well?
> 
> @guacs what tests do you have in mind? The only one that I can come up is to modify the [middleware integration test](https://github.com/litestar-org/litestar/blob/main/tests/unit/test_middleware/test_session/test_integration.py) to confirm that the session id is set in the route handler scope - and that it is `not None` for the server side config and `None` for the client side config. But I did not see any assert statements in the route handlers of the other tests, so I wanted to check if this really goes into the right direction.

I think you can add new tests to the `test_middleware.py` and make sure that the session ID that is returned by `get_session_id` is the same as the one set in the cookies. Also, I think a test to ensure that the session ID that is returned is the existing session ID when a session has already been created. That is, we don't want to create a new session ID when one already exists.

litestar/security/session_auth/middleware.py
```diff
@@ -61,7 +60,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                 exception_handlers=litestar_app.exception_handlers or {},  # pyright: ignore
                 debug=None,
             )
-            self.app = SessionMiddleware(
+            self.app = self.config.session_backend_config.middleware.middleware(
```
guacs, line 63: Just for my understanding, why was this change needed?

guacs (approved):
Thanks a bunch @aranvir! This look great :)

guacs:
Also, I just realized that you're targeting the wrong branch. You should be targeting `develop` since this is a new feature.

litestar/security/session_auth/middleware.py
aranvir: Well, before I started working on this I was experimenting with a SessionMiddleware override. But even though, I also overrode the corrsponding `DefineMiddleware` method, my custom version was not loaded. So since it is possible to place a reference to the actual session backend middleware, I thought that was the better way to do it: the session_auth should use whatever session middleware the backend provides, not just hard-code a certain middleware.

provinzkraut (approved):

provinzkraut:
@aranvir Can you address the coverage? Once that's through, this is good to go in :)

litestar/testing/client/base.py
```diff
@@ -162,13 +162,16 @@ def _create_session_cookies(backend: ClientSideSessionBackend, data: dict[str, A
 
     async def _set_session_data(self, data: dict[str, Any]) -> None:
         mutable_headers = MutableScopeHeaders()
+        connection = fake_asgi_connection(
+            app=self.app,
+            cookies=dict(self.cookies),  # type: ignore[arg-type]
+        )
+        session_id = None
+        if self._session_backend is not None:
```
provinzkraut, line 170: You'll need to add a test that hit's this condition, i.e. one where `AsyncClient.set_session_data` / `SyncClient.set_session_data` is called without a session backend being set.

aranvir: Actually, could fix it by replacing it with `self.session_backend`: If the backend is None, an exception is raised, since you should not interatct with a backend that does not exist. Both mypy and tests accept that.

litestar/middleware/session/client_side.py
```diff
@@ -206,6 +206,9 @@ async def load_from_connection(self, connection: ASGIConnection) -> dict[str, An
                 return self.load_data(data)
         return {}
 
+    def get_session_id(self, connection: ASGIConnection) -> str | None:
```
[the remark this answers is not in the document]
aranvir: It implements the abstract method of `BaseSessionBackend` which expects a `str` for server-side backend and `None` for client-side backend. This is why I kept the signature the same, even though it will never return a string.

litestar/middleware/session/server_side.py
```diff
@@ -77,6 +77,26 @@ async def delete(self, session_id: str, store: Store) -> None:
...
+        been returned to the user.
+        Otherwise, a new session must be created.
+
+        Args:
+            connection: Originating ASGIConnection containing the scope
+        Returns:
+            Session id str or None if the concept of a session id does not apply.
+        """
+        session_id = connection.cookies.get(self.config.key)
+        if not session_id or session_id == "null":
```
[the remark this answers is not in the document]
aranvir: Not for me to decide, but I would suggest to make a separate PR for this, if a `null` enum would be useful ;)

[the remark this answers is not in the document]
guacs: @kedod I don't think an enum is needed, however moving that into a constant would be a nice change imo.

Also, I agree with @aranvir, if this change is being made, it should be another PR.

PR #3128, opened 2024-02-23
feat: pass examples given in the `FieldDefinition` to the `OpenAPIMediaType`

### LINKED ISSUE #3076: Bug: Examples not appearing in OpenAPI request
### Description

This could be user error.  Trying to add examples to the request body.
The body has no problem with generating the schema and "example value" based on the structure but I can't seem to get the "examples" working.

Much prefer having the examples decoupled from the model lib, so if I could get this working it would be awesome!

Thanks!

_No response_

### MCVE

```python
@post(path="/signals", tags=["signals"])
async def get_signals(
        data: Annotated[SignalRequest, Body(
            title="Get Signals",
            description="Get signals.",
            examples=[
                Example(
                    summary="Example 1",
                    description="use csv",
                    value={
                        "test": "this"},
                ),
                Example(
                    summary="Example 2",
                    value={
                        "test": "this"},
                )
            ]
        )],
) -> dict:
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.5.5

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

robswc:
## Description

- Passes examples given within the FieldDefinition to the OpenAPIMediaType

```python
@post(media_type=MediaType.TEXT)
def create_person(
    self, data: Annotated[DataclassPerson, Body(examples=[Example(summary="this is an example", value={"key": "value"})])], secret_header: str = Parameter(header="secret")
) -> DataclassPerson:
    return data
```

image

## Fixes

Fixes #3076

### DIFF
--- a/litestar/_openapi/request_body.py
+++ b/litestar/_openapi/request_body.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Mapping
 
 from litestar._openapi.schema_generation import SchemaCreator
+from litestar._openapi.schema_generation.utils import get_formatted_examples
 from litestar.enums import RequestEncodingType
 from litestar.openapi.spec.media_type import OpenAPIMediaType
 from litestar.openapi.spec.request_body import RequestBody
@@ -13,7 +14,7 @@ __all__ = ("create_request_body",)
 if TYPE_CHECKING:
     from litestar._openapi.datastructures import OpenAPIContext
     from litestar.dto import AbstractDTO
-    from litestar.openapi.spec import Example, Reference
+    from litestar.openapi.spec import Example
     from litestar.typing import FieldDefinition
 
 
@@ -48,11 +49,8 @@ def create_request_body(
     else:
         schema = schema_creator.for_field_definition(data_field)
 
-    examples: dict[str, Example | Reference] | None = None
-    if isinstance(data_field.kwarg_definition, BodyKwarg) and data_field.kwarg_definition.examples:
-        examples = {}
-        for example in data_field.kwarg_definition.examples:
-            if isinstance(example.summary, str) and isinstance(example.value, dict):
-                examples[example.summary] = example
+    examples: Mapping[str, Example] | None = None
+    if isinstance(data_field.kwarg_definition, BodyKwarg) and isinstance(data_field.kwarg_definition.examples, list):
+        examples = get_formatted_examples(data_field, data_field.kwarg_definition.examples)
 
     return RequestBody(required=True, content={media_type: OpenAPIMediaType(schema=schema, examples=examples)})
--- a/litestar/_openapi/request_body.py
+++ b/litestar/_openapi/request_body.py
@@ -4,12 +4,12 @@ from typing import TYPE_CHECKING
 
 from litestar._openapi.schema_generation import SchemaCreator
 from litestar.enums import RequestEncodingType
+from litestar.openapi.spec import Example, Reference
 from litestar.openapi.spec.media_type import OpenAPIMediaType
 from litestar.openapi.spec.request_body import RequestBody
 from litestar.params import BodyKwarg
 
 __all__ = ("create_request_body",)
-
 
 if TYPE_CHECKING:
     from litestar._openapi.datastructures import OpenAPIContext
@@ -48,4 +48,10 @@ def create_request_body(
     else:
         schema = schema_creator.for_field_definition(data_field)
 
-    return RequestBody(required=True, content={media_type: OpenAPIMediaType(schema=schema)})
+    examples: dict[str, Example | Reference] | None = None
+    if isinstance(data_field.kwarg_definition, BodyKwarg) and data_field.kwarg_definition.examples:
+        examples = {
+            example.summary: example for example in data_field.kwarg_definition.examples if example.summary is not None
+        }
+
+    return RequestBody(required=True, content={media_type: OpenAPIMediaType(schema=schema, examples=examples)})
--- a/litestar/_openapi/request_body.py
+++ b/litestar/_openapi/request_body.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING
 
 from litestar._openapi.schema_generation import SchemaCreator
 from litestar.enums import RequestEncodingType
-from litestar.openapi.spec import Example, Reference
 from litestar.openapi.spec.media_type import OpenAPIMediaType
 from litestar.openapi.spec.request_body import RequestBody
 from litestar.params import BodyKwarg
@@ -15,6 +14,7 @@ if TYPE_CHECKING:
     from litestar._openapi.datastructures import OpenAPIContext
     from litestar.dto import AbstractDTO
     from litestar.typing import FieldDefinition
+    from litestar.openapi.spec import Example, Reference
 
 
 def create_request_body(
--- a/litestar/_openapi/request_body.py
+++ b/litestar/_openapi/request_body.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from litestar.dto import AbstractDTO
     from litestar.openapi.spec import Example
     from litestar.typing import FieldDefinition
+    from litestar.openapi.spec import Example, Reference
 
 
 def create_request_body(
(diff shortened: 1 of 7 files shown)


guacs:
Thanks for this @robswc! I had a few points/doubts:

1. I think this logic could maybe be moved into `__post_init__` of `OpenAPIMediaType` so that this will work for both request as well as responses. Basically if there are no `examples` and the provided schema does have examples, then do the required transformation.
2. From my understanding, the keys of the dictionaries are used as the name of the example, so it feels a bit confusing to take the `Example.summary` as the name of the example.
3. Was there a reason that only examples in the form of dictionaries are considered? What if the response is just a simple integer, or a string? I think we should be able to pass in any type there and the serialization of the generated schema into JSON will handle that properly.

robswc:
Thanks for the feedback! @guacs 

>From my understanding, the keys of the dictionaries are used as the name of the example, so it feels a bit confusing to take the Example.summary as the name of the example.

Truthfully, I wasn't sure the best way to handle it so I'm on board with whatever is the best (or least worst!) solution.  Any ideas on how to pass the names through?  I suppose I could do it by modifying the `Example` to give it a `name` but wanted to start simple as possible.

>Was there a reason that only examples in the form of dictionaries are considered? What if the response is just a simple integer, or a string? I think we should be able to pass in any type there and the serialization of the generated schema into JSON will handle that properly.

It was a bit of a hail mary as I haven't worked with Sonar before.  Had an unrelated issue with linux permissions which was making linting not work but I moved the project to a new drive so I should be able to generate coverage reports now and make sure that works properly!  IIRC, `value` is an `Any` so that would work best?

Thanks again!

guacs:
> I suppose I could do it by modifying the Example to give it a name but wanted to start simple as possible.

I don't know about this either. Since `Example` is part of the OpenAPI spec but it doesn't include a `name` field as far as I remember.

> value is an Any so that would work best

Seems good to me.

robswc:
>I don't know about this either. Since Example is part of the OpenAPI spec but it doesn't include a name field as far as I remember.

Makes sense.  I like the idea of making as little modifications as possible.

`data: Annotated[SampleClass, Body(examples=[Example(summary="example", value={"name": "John", "age": 30})])]`

What about `Body` in `litestar/params.py` ?

```python
def Body(
    *,
    const: bool | None = None,
    content_encoding: str | None = None,
    default: Any = Empty,
    description: str | None = None,
    examples: list[Example] | None = None,
```

And have examples be `dict[str, Example]` ?  Not sure if that would effect the way things currently work though.

I'm still getting familiar with this amazing framework, so not sure the in's and out's quite yet but feel very close to making this work! :) 

Thanks again for the help!

guacs:
> And have examples be dict[str, Example] ? Not sure if that would effect the way things currently work though.

We will have to allow both `list[Example` as well as `dict[str, Example]`. So if the user gives a `dict[str, Example]`, then the keys should be used as the name and if it's a `list[Example]`, we should generate the example names. I think you can refer `get_formatted_examples` on how to handle the automatic generation of the names.

guacs:
> > And have examples be dict[str, Example] ? Not sure if that would effect the way things currently work though.
> 
> We will have to allow both `list[Example` as well as `dict[str, Example]`. So if the user gives a `dict[str, Example]`, then the keys should be used as the name and if it's a `list[Example]`, we should generate the example names. I think you can refer `get_formatted_examples` on how to handle the automatic generation of the names.

Btw, this is for backwards compatibility.

robswc:
> > And have examples be dict[str, Example] ? Not sure if that would effect the way things currently work though.
> 
> We will have to allow both `list[Example` as well as `dict[str, Example]`. So if the user gives a `dict[str, Example]`, then the keys should be used as the name and if it's a `list[Example]`, we should generate the example names. I think you can refer `get_formatted_examples` on how to handle the automatic generation of the names.

Makes sense!

When I attempt to change the types for "Example" in body, or rather, when I pass a dict instead of list for `Examples` in `Body`, I get this error:

```
>               meta_data = Meta(**meta_kwargs)
E               TypeError: `examples` must be a list, got dict

../../../litestar/_signature/model.py:249: TypeError
```

I think the issue is the name `examples`, there is an attempt to add it to `Meta` in the `model.py` file.

```python
for k, v in kwarg_definition.items():
    if hasattr(Meta, k) and v is not None:
        meta_kwargs[k] = v
    else:
        meta_kwargs["extra"][k] = v
```

Perhaps it would be easier all around to ditch the idea of naming an example and have it generate using the `get_formatted_examples` you mentioned? Personally, naming the examples would be great but for my use case just having them there would be priceless!

guacs:
> > > And have examples be dict[str, Example] ? Not sure if that would effect the way things currently work though.
> > 
> > 
> > We will have to allow both `list[Example` as well as `dict[str, Example]`. So if the user gives a `dict[str, Example]`, then the keys should be used as the name and if it's a `list[Example]`, we should generate the example names. I think you can refer `get_formatted_examples` on how to handle the automatic generation of the names.
> 
> Makes sense!
> 
> When I attempt to change the types for "Example" in body, or rather, when I pass a dict instead of list for `Examples` in `Body`, I get this error:
> 
> ```
> >               meta_data = Meta(**meta_kwargs)
> E               TypeError: `examples` must be a list, got dict
> 
> ../../../litestar/_signature/model.py:249: TypeError
> ```
> 
> I think the issue is the name `examples`, there is an attempt to add it to `Meta` in the `model.py` file.
> 
> ```python
> for k, v in kwarg_definition.items():
>     if hasattr(Meta, k) and v is not None:
>         meta_kwargs[k] = v
>     else:
>         meta_kwargs["extra"][k] = v
> ```
> 
> Perhaps it would be easier all around to ditch the idea of naming an example and have it generate using the `get_formatted_examples` you mentioned? Personally, naming the examples would be great but for my use case just having them there would be priceless!

For now, I think we can stick to using `get_formatted_examples` like you said. Allowing the explicit naming of examples can be dealt with seperately if there's enough demand for it.

robswc:
> For now, I think we can stick to using `get_formatted_examples` like you said. Allowing the explicit naming of examples can be dealt with seperately if there's enough demand for it.

Sounds good to me! @guacs - the title was changed to `feat` (which makes sense, initially I thought it was a bug/fix), I got some CI errors, looked like it wanted me to point to "develop" branch, which I have done.  Is there anything else needed on my end?

Thanks again!

provinzkraut:
> Sounds good to me! @guacs - the title was changed to `feat` (which makes sense, initially I thought it was a bug/fix), I got some CI errors, looked like it wanted me to point to "develop" branch, which I have done. Is there anything else needed on my end?

All good on that front, just needs a review to go in

litestar/_openapi/request_body.py
```diff
@@ -48,4 +49,8 @@ def create_request_body(
     else:
         schema = schema_creator.for_field_definition(data_field)
 
-    return RequestBody(required=True, content={media_type: OpenAPIMediaType(schema=schema)})
+    examples: Mapping[str, Example] | None = None
+    if isinstance(data_field.kwarg_definition, BodyKwarg) and isinstance(data_field.kwarg_definition.examples, list):
```
provinzkraut, line 53:
```suggestion
    if isinstance(data_field.kwarg_definition, BodyKwarg) and data_field.kwarg_definition.examples is not None:
```

This check is a bit more resilient to refactorings than `isinstance(..., list)` (allow-list vs. block-list). With the suggested approach we could easily change the type of `KwargDefinition.examples` to allow any compatible type without having to adapt this check here.

litestar/openapi/spec/media_type.py
```diff
@@ -32,7 +32,7 @@ class OpenAPIMediaType(BaseSchemaObject):
     example provided by the schema.
     """
 
-    examples: dict[str, Example | Reference] | None = None
+    examples: dict[str, Example | Reference] | Mapping[str, Example] | None = None
```
provinzkraut, line 35: While I think this is a good change, it's unfortunately breaking; If someone is using this attribute and referring to dict-only methods on it, passing a different type of mapping here would break their code and also their type checking. 

Iiuc having this be a `Mapping` isn't really necessary for the proposed feature, right?

provinzkraut (changes requested):
Thanks for your contribution @robswc! I've left two comments that would need addressing before this can go in. Otherwise it looks good to me :)

JacobCoffee:
Are we able to rebase against the latest `develop` branch and get this merged?
@robswc

robswc:
> Are we able to rebase against the latest `develop` branch and get this merged? @robswc

I can look into it!  Been super busy lately and IIRC, changing some of the types was the only way I could get it to work with mypy.  I'll give it another shot though!

PR #3131, opened 2024-02-23
fix: pass cache control header for static files

### LINKED ISSUE #3129: Bug: CacheControlHeader not applying on create_static_files_router
### Description

When applying a `CacheControlHeader` to the [static files router](https://docs.litestar.dev/latest/reference/static_files.html#litestar.static_files.create_static_files_router), the cache headers are not applied to requests served from this router.

`CacheControlHeader` works correctly on regular routers.

NA

### MCVE

```python
from litestar.response import File
from litestar.static_files import create_static_files_router
from litestar.datastructures.headers import CacheControlHeader
import uvicorn

def create_static_router() -> Router:
    return create_static_files_router(
        path="/static",
        directories=["./static"],
        name="static",
        cache_control=CacheControlHeader(max_age=3600)
    )

def create_direct_route_handler() -> Router:
    @get("/example.txt")
    async def example_txt() -> File:
        return File("static/example.txt", content_disposition_type="inline", media_type="text/plain", filename="example.txt")
    
    return Router(
        path="/static",
        cache_control=CacheControlHeader(max_age=3600),
        route_handlers=[example_txt],
    )

app = Litestar(route_handlers=[create_static_router()]) # <- this does not work as expected
#app = Litestar(route_handlers=[create_direct_route_handler()]) # <- this works as expected

if __name__ == "__main__":
    uvicorn.run(
        app,
    )

# Expects a "static" folder in the same directory as this file and an example.txt file in the "static" folder
```

### Steps to reproduce

```bash
1. Put the MVCE beside a folder called static with example.txt inside it
2. Run the server and make a request to /static/example.txt
3. Observe that there is no cache header applied to the request
4. Change the code to the create_direct_route_handler() version and repeat
5. Observe the cache header applied as expected
```

### Screenshots

```bash
NA
```

### Logs

_No response_

### Litestar Version

2.6.1

### Platform

- [ ] Linux
- [ ] Mac
- [X] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3129">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3129/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

guacs:
## Description

The `cache_control` that is set on a router created by `create_static_files_router` isn't considered because the return type of the handler function is an async callable so the resolved headers are ignored [here](https://github.com/litestar-org/litestar/blob/3e5c179e714bb074bae13e02d10e2f3f51e24d5c/litestar/handlers/http_handlers/base.py#L489-L492). As a workaround, the cache control header is resolved and passed directly to the `ASGIFileResponse`.

It may have been better for the `StaticFiles.handle` to return a `File` response so that this would be handled as per our normal flow and there would be no need for special casing here. This would also allow the users to provide any headers/cookies they might want to attach to the router. However, this would be a breaking change and this could be a change considered for the next major version.

## Closes
Fixes #3129

### DIFF
--- a/litestar/static_files/base.py
+++ b/litestar/static_files/base.py
@@ -14,6 +14,7 @@ __all__ = ("StaticFiles",)
 
 
 if TYPE_CHECKING:
+    from litestar.datastructures.headers import CacheControlHeader
     from litestar.types import Receive, Scope, Send
     from litestar.types.composite_types import PathType
     from litestar.types.file_types import FileInfo, FileSystemProtocol
@@ -22,7 +23,7 @@ if TYPE_CHECKING:
 class StaticFiles:
     """ASGI App that handles file sending."""
 
-    __slots__ = ("is_html_mode", "directories", "adapter", "send_as_attachment")
+    __slots__ = ("is_html_mode", "directories", "adapter", "send_as_attachment", "headers")
 
     def __init__(
         self,
@@ -31,6 +32,7 @@ class StaticFiles:
         file_system: FileSystemProtocol,
         send_as_attachment: bool = False,
         resolve_symlinks: bool = True,
+        cache_control: CacheControlHeader | None = None,
     ) -> None:
         """Initialize the Application.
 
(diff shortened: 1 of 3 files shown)


provinzkraut:
> It may have been better for the StaticFiles.handle to return a File response so that this would be handled as per our normal flow and there would be no need for special casing here. This would also allow the users to provide any headers/cookies they might want to attach to the router. However, this would be a breaking change and this could be a change considered for the next major version.

Yeah, the whole `StaticFiles` app thing isn't really needed anymore and just kept around for backwards compatibility.

litestar/static_files/base.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         file_system: FileSystemProtocol,
         send_as_attachment: bool = False,
         resolve_symlinks: bool = True,
+        cache_control: CacheControlHeader | None = None,
```
provinzkraut, line 35: I think accepting `headers` here directly and resolving the `CacheControlHeader` on the layer above if present makes more sense

provinzkraut (changes requested):

litestar/static_files/base.py
guacs: Yeah, that's much cleaner.

provinzkraut (approved):

PR #3143, opened 2024-02-26
docs: Fix testing code examples

hugovk:
## Description

- At https://docs.litestar.dev/2/usage/testing.html, the examples after "We would then be able to rewrite our test like so:" don't follow. They contain a full-working example (production code, fixture and test case: [sync](https://github.com/litestar-org/litestar/blob/main/docs/examples/testing/test_health_check_sync.py) and [async](https://github.com/litestar-org/litestar/blob/main/docs/examples/testing/test_health_check_async.py)). We had just moved the fixture to `tests/conftest.py`, and the production code is still in `my_app/main.py`, so we just need the test case in `tests/test_health_check.py`.
- And fix some typos.


provinzkraut:
Thanks for the explanation @hugovk, makes sense.

I think the idea here was that example files should be self-contained, which is why all the code was copied into one file. Of course that doesn't make sense here given the text. I see two ways to fix this: Either stick to "one self-contained file" and remove the `conftest.py` or adjust the tests accordingly to use the fixtures from `conftest.py`.

I'm actually leaning towards option 1 slightly, because the `conftest.py` isn't really needed for what this example is trying to demonstrate. So the example with the fixture should just look like 

```python
import pytest

from litestar.status_codes import HTTP_200_OK
from typing import Generator

from my_app.main import app


@pytest.fixture()
def client() -> Generator[TestClient, None, None]:
        yield client


def test_health_check(client: TestClient):
    response = client.get("/health-check")
    assert response.text == "healthy"
```

and the text be adjusted accordingly.

guacs (approved):

provinzkraut:
@guacs I think this still needs the changes discussed 👀

guacs:
> @guacs I think this still needs the changes discussed 👀

Oh I think I misunderstood then :P I thought the changes were fine since it showed that you can use the test client as a fixture, but now looking at it again I think you may be right. 

I think you wanted to make the change so that we remove the reference to `conftest.py` right?

provinzkraut:
> I think you wanted to make the change so that we remove the reference to conftest.py right?

That's what I thought @kedod was agreeing to here:

> Option 1 sounds fine. Like you say, the main point is you can put the client in a fixture for re-use, and you can put that where you like (same file or conftest.py).

It also still has the in-line code, which we don't test.

guacs:
> > I think you wanted to make the change so that we remove the reference to conftest.py right?
> 
> That's what I thought @kedod was agreeing to here:
> 
> > Option 1 sounds fine. Like you say, the main point is you can put the client in a fixture for re-use, and you can put that where you like (same file or conftest.py).
> 
> It also still has the in-line code, which we don't test.

You're right. Also, did you mean @hugovk?

PR #3153, opened 2024-03-02
feat: Add `type_decoders` Router and route handlers

kedod:
## Description

- added `type_decoders` to `__init__` method for handler, routers and decorators to keep consistency with `type_encoders` parameter
- added test  for resolving type decoders


provinzkraut (changes requested):
Looks good, you seem to be on a roll here :slightly_smiling_face: 

To be consistent though, the parameter should be added to the `websocket_listener` as well, since that also has the `type_encoders`

provinzkraut (approved):

PR #3162, opened 2024-03-03
feat: Pass `type_decoders` in WebsocketListenerRouteHandler

kedod:
## Description
- pass `type_decoders` to parent's `__init__` in `WebsocketListenerRouteHandler` init, otherwise `type_decoders` will be `None` 
- add tests and `baby` refactor request and response class resolution tests
- replace params order in docs, `__init__` (`decoders` before `encoders`)

### DIFF
--- a/tests/unit/test_response_class_resolution.py
+++ b/tests/unit/test_response_class_resolution.py
@@ -1,8 +1,9 @@
-from typing import Type, Union
+from typing import Optional, Type
 
 import pytest
 
 from litestar import Controller, HttpMethod, Litestar, Response, Router, get
+from litestar.handlers.http_handlers.base import HTTPRouteHandler
 
 RouterResponse: Type[Response] = type("RouterResponse", (Response,), {})
 ControllerResponse: Type[Response] = type("ControllerResponse", (Response,), {})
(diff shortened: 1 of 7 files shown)


tests/unit/test_response_class_resolution.py
```diff
@@ -52,10 +52,10 @@ def handler(self) -> None:
...
 
-    route_handler, _ = app.routes[0].route_handler_map[HttpMethod.GET]  # type: ignore
+    route_handler = app.route_handler_method_map["/"][HttpMethod.GET]
 
     if handler_response_class:
-        route_handler.response_class = handler_response_class
+        route_handler.response_class = handler_response_class  # type: ignore
 
-    response_class = route_handler.resolve_response_class()
+    response_class = route_handler.resolve_response_class()  # type: ignore
```
provinzkraut, line 60: We try not to use blanket `type: ignore`s. Can you add the specific error codes?

provinzkraut (approved):
Thanks :)

tests/unit/test_response_class_resolution.py
provinzkraut: Actually, we should enforce this. There's a mypy setting `enable_error_code = "ignore-without-code"` that does this. Would you be interested in creating a PR for this? :)

PR #3163, opened 2024-03-03
chore: Enforce mypy error definition for ignored lines.

kedod:
## Description

- Added `enable_error_code = "ignore-without-code"` to mypy config

### DIFF
--- a/docs/examples/middleware/session/cookie_backend.py
+++ b/docs/examples/middleware/session/cookie_backend.py
@@ -3,6 +3,6 @@ from os import urandom
 from litestar import Litestar
 from litestar.middleware.session.client_side import CookieBackendConfig
 
-session_config = CookieBackendConfig(secret=urandom(16))  # type: ignore[arg-type]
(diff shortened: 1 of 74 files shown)


JacobCoffee (approved):

docs/examples/middleware/session/cookie_backend.py
```diff
@@ -3,6 +3,6 @@
 from litestar import Litestar
 from litestar.middleware.session.client_side import CookieBackendConfig
 
-session_config = CookieBackendConfig(secret=urandom(16))  # type: ignore[arg-type]
+session_config = CookieBackendConfig(secret=urandom(16))  # type: ignore
```
JacobCoffee, line 6: idk how i missed this, how did this pass the CI if we disallow generic ignores?

JacobCoffee: Also Ruff's `PGH003` should be catching too. Weird.

PR #3171, opened 2024-03-05
feat: Add support for v2 model_dump and v1 dict to pydantic plugin

### LINKED ISSUE #3147: Enhancement: Support all kwargs for `model_dump` in PydanticPlugin
### Summary

`litestar.contrib.pydantic.PydanticPlugin` has the argument `prefer_alias` which sets the `by_alias` argument of the created Pydantic type encoders. The type encoders are based on `pydantic.BaseModel.model_dump` which also has the arguments `exclude_none` and `exclude_defaults`. This enhancement suggests to add similar arguments to `PydanticPlugin` to configure the type encoders.

### Basic Example

```python
import pydantic
import typing
import litestar
import litestar.contrib.pydantic

class MyModel(pydantic.BaseModel):
    some_number: int = pydantic.Field(alias="someNumber")
    title: str = "a number"
    comment: typing.Optional[str] = None

@litestar.get("/")
def get_stuff() -> MyModel:
    return MyModel(some_number=42)

app = litestar.Litestar(
    route_handlers=[get_stuff],
    plugins=[
        litestar.contrib.pydantic.PydanticPlugin(
            prefer_alias=True,
            # NOTE: The argument names are just provisional
            exclude_none=True,  # new
            exclude_defaults=True,  # new
        )
    ],
)

```
A GET request to `/` would return `{"someNumber":42}` because of the plugin configuration. Without it, it would return `{"some_number":42,"title":"a number","comment":null}`.

### Drawbacks and Impact

You can currently achieve the same result without this feature by adding a custom type encoder for `pydantic.BaseModel`. However, it must be combined with `PydanticPlugin` for the generated schema to use aliases.
```python
import pydantic
import typing
import litestar
import litestar.contrib.pydantic

class MyModel(pydantic.BaseModel):
    some_number: int = pydantic.Field(alias="someNumber")
    title: str = "a number"
    comment: typing.Optional[str] = None

@litestar.get("/")
def get_stuff() -> MyModel:
    return MyModel(some_number=42)


app = litestar.Litestar(
    route_handlers=[get_stuff],
    plugins=[litestar.contrib.pydantic.PydanticPlugin(prefer_alias=True)],
    type_encoders={
        pydantic.BaseModel: lambda x: x.model_dump(mode="json", by_alias=True, exclude_none=True, exclude_default=True),
    },
)

```

Therefore, the only purpose of this enhancement is to make this use case (which I believe to be a common one) simpler to express in code by making the plugin able to do more, reducing the need for type encoders. The community must decide if this is desirable or not.

As long as the default values of the new arguments match the old behavior, this should not cause any issues with backwards compatibility. 

### Unresolved questions

I can see that there may be concerns about having to support a growing list of `pydantic.BaseModel.model_dump` arguments (for example `exlude_unset`). I believe that concern to be valid, but you could always draw a line and say "the plugin supports these extremely common arguments, if you want more, use a custom type encoder". Still, the concern is valid and I believe it should be part of the discussion.

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3147">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3147/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

kedod:
## Description

- Add support for additional V1 `model_dump` and V2 `dict` parameters in `PydanticPlugin`


guacs:
What if instead of taking in the arguments explicitly, we take a dictionary of the keyword arguments that we'll pass to each function? My reasoning is that, this way if `pydantic` adds a new kwarg or changes a default value or deprecates a kwarg, then we don't have to worry about it. The user can make the change accordingly in their code and there would be no issues. Otherwise, we would have to keep in sync with `pydantic` which can become difficult if the kwargs differ based on the pydantic version.

A drawback to this approach is that users lose the type checking, but I feel the flexibility it provides is nice.


Also, I was thinking that along with this we could take arguments for `model_validate`/`pase_obj` as well though I don't know if `parse_obj` does take any arguments other than the value as a dictionary.

provinzkraut:
@guacs I agree that it would be easier, but I'd say we should side with the more type safe option here, since that's what we're doing everywhere else.

There is a way to type something like this btw, but it's not pretty:

```python
from typing import Callable, Any, Generic, ParamSpec

P = ParamSpec("P")

class _Foo(Generic[P]):
    def __init__(self, **kwargs: P.kwargs) -> None:
        pass


def _make_foo(f: Callable[P, Any]) -> Type[_Foo[P]]:
    return _Foo[P]


def func(param: str) -> None:
    pass


Foo = _make_foo(func)

Foo(param=1)
```

This actually type-checks correctly, but it's quite ugly. Not sure if there's a better way to do something like this; If there is, I'm not aware. @sobolevn maybe you have an idea here? :)

provinzkraut:
@kedod Have you checked what schema Pydantic generates in those cases?

provinzkraut:
No I meant what Pydantic itself generates, not what we generate: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_json_schema

provinzkraut:
Interesting. It seems like Pydantic itself simply doesn't take these particularities into account. There is the `by_alias` option for `model_json_schema`, but that's about it.

provinzkraut:
How about we simply make them not required if any of the `exclude` options is set that could apply to a certain field? E.g. a field that can be `None` would be optional if `exclude_none=True`?

guacs:
> @guacs I agree that it would be easier, but I'd say we should side with the more type safe option here, since that's what we're doing everywhere else.
> 
> There is a way to type something like this btw, but it's not pretty:
> 
> ```python
> from typing import Callable, Any, Generic, ParamSpec
> 
> P = ParamSpec("P")
> 
> class _Foo(Generic[P]):
>     def __init__(self, **kwargs: P.kwargs) -> None:
>         pass
> 
> 
> def _make_foo(f: Callable[P, Any]) -> type[_Foo[P]]:
>     return _Foo[P]
> 
> 
> def func(param: str) -> None:
>     pass
> 
> 
> Foo = _make_foo(func)
> 
> Foo(param=1)
> ```
> 
> This actually type-checks correctly, but it's quite ugly. Not sure if there's a better way to do something like this; If there is, I'm not aware. @sobolevn maybe you have an idea here? :)

I didn't understand how this black magic will allow us to type hint the arguments automatically.

guacs:
@kedod could you take a look at the conflicts and rebase from develop?

Also, I think explicit arguments is better as well.

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -322,3 +371,34 @@ def for_pydantic_model(cls, field_definition: FieldDefinition, schema_creator: S
...
+        include = include or []
+
+        for prop in property_fields.values():
+            name = prop.name
+
+            if any(
+                [
+                    prop.name in exclude,
+                    exclude_none and (prop.is_optional or prop.is_none_type),
+                    (exclude_defaults or exclude_unset) and prop.has_default,
```
guacs, line 396: I have a doubt here. What if the user only wants to ensure that pydantic doesn't include the fields that have a default value (or any of the other cases), when pydantic serializes a model into JSON, but they still want those fields to be documented?

litestar/typing.py
```diff
@@ -327,7 +327,7 @@ def is_required(self) -> bool:
         if isinstance(self.kwarg_definition, ParameterKwarg) and self.kwarg_definition.required is not None:
             return self.kwarg_definition.required
 
-        return not self.is_optional and not self.is_any and (not self.has_default or self.default is None)
+        return not self.is_optional and not self.is_any and (not self.has_default)
```
guacs, line 330: Do you know why we were checking for `self.default is None` here?

guacs (commented):
Thanks @kedod! I had a few doubts that I've mentioned as review comments.

litestar/_openapi/schema_generation/schema.py
```diff
@@ -246,7 +246,17 @@ def create_schema_for_annotation(annotation: Any) -> Schema:
...
+    __slots__ = (
+        "exclude",
+        "exclude_defaults",
+        "exclude_none",
+        "exclude_unset",
+        "include",
+        "generate_examples",
+        "plugins",
+        "prefer_alias",
+        "schema_registry",
```
guacs, line 258: I think removing from this was missed.

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -234,21 +255,43 @@ def to_openapi_schema(self, field_definition: FieldDefinition, schema_creator: S
...
+                exclude_defaults=self.exclude_defaults,
+                exclude_none=self.exclude_none,
+                exclude_unset=self.exclude_unset,
+                include=self.include,
+            )
         return PYDANTIC_TYPE_MAP[field_definition.annotation]  # pragma: no cover
 
     @classmethod
-    def for_pydantic_model(cls, field_definition: FieldDefinition, schema_creator: SchemaCreator) -> Schema:  # pyright: ignore
+    def for_pydantic_model(
```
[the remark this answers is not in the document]
guacs: I know I suggested this, but would this be considered breaking?

litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -322,3 +355,34 @@ def for_pydantic_model(cls, field_definition: FieldDefinition, schema_creator: S
...
+        include = include or []
+
+        for prop in property_fields.values():
+            name = prop.name
+
+            if any(
+                [
+                    name in exclude,
+                    exclude_none and (prop.is_optional or prop.is_none_type),
+                    (exclude_defaults or exclude_unset) and prop.has_default,
```
guacs, line 380: I'm still a bit hesitant regarding this. This makes sense for the schema we generate for responses, but it doesn't make sense for the schema we generate for request bodies. Since we don't differentiate between whether the model is used as a request or response body, this may be problematic.

litestar/typing.py
guacs: Yeah I felt like that was a bug we had initially as well.

litestar/contrib/pydantic/pydantic_schema_plugin.py
guacs: I'm still a bit hesitant regarding this. This makes sense for the schema we generate for responses, but it doesn't make sense for the schema we generate for request bodies. Since we don't differentiate between whether the model is used as a request or response body, this may be problematic.

JacobCoffee (approved):

litestar/contrib/pydantic/pydantic_schema_plugin.py
guacs: I don't think we can since we create the same schema for both request and response bodies. Changing that would be possible, but that could cause issues such as with the naming of the schemas. We would need to name them differently based on whether they were the request or response bodies. I'm not sure if adding that complexity and these changes are worth the benefit.

guacs: Personally, I think adding it only for `PydanticInitPlugin` would be enough. We can just document that while these flags effect the responses, they're not reflected in the generated OpenAPI schema.

provinzkraut: > Personally, I think adding it only for `PydanticInitPlugin` would be enough. We can just document that while these flags effect the responses, they're not reflected in the generated OpenAPI schema.

I think that's going to lead to confusion. The core issue here is, and has always been, that you can set conflicting values in various places and it's not immediately clear which ones take precedence. However, I also think that's okay to some degree, as long as we make sure that the defaults and patterns we promote are sensible and the rest of the behaviour is documented well.  

As for this particular problem: If we were to only expose the init plugin for Pydantic, we could ensure internal consistency here, right? So why don't we accept that users will be able to create inconsistent schemas for now if they really want to, and change this in 3.0?

JacobCoffee:

PR #3176, opened 2024-03-07
fix: sending empty data in sse in js client 

euri10:
## Description

- from our discussion here https://discord.com/channels/919193495116337154/1213242966592856164 it appears that javascript clients dont "see" an event that has no data, while we explicitly test for this and that httpx-sse receives the event fine in that case.
- the spec https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream is not that clear to me as to whether or not an event without data is ok.
- Considering the EventSource "client" is not ok with it, and that it's so easy DX-wise to make the mistake not explicitely sending it we fix it by defaulting to the empty-string

 
The added app is just to manually test the fix works, I dont know how we could incorporate it in our test suite given it involves running a server and checkin if the html page yields the event fine, maybe in e2e testing ?

### DIFF
--- a/litestar/response/sse.py
+++ b/litestar/response/sse.py
@@ -91,7 +91,7 @@ class _ServerSentEventIterator(AsyncIteratorWrapper[bytes]):
 
 @dataclass
 class ServerSentEventMessage:
-    data: str | int | bytes | None = None
+    data: str | int | bytes | None = ""
     event: str | None = None
     id: int | str | None = None
     retry: int | None = None
--- a/test_apps/sse/sse.html
+++ b/test_apps/sse/sse.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html lang="en">
+<body>
+<h2>Server-Sent Events</h2>
+<div id="output"></div>
+<script>
+    if (typeof EventSource !== "undefined") {
+        var eventSource = new EventSource("http://localhost:8000/test_sse_empty");
+        eventSource.onmessage = function (event) {
+            console.log(event.data);
+            document.getElementById("output").innerHTML += event.data + "<br>";
+        };
+        eventSource.addEventListener("empty", (ev) => {
+            console.log("empty arrived");
+            document.getElementById("output").innerHTML += "endx arrived" + "<br>";
+            eventSource.close();
+        });
+
+        eventSource.addEventListener("error", (err) => {
+            console.error(err)
+            document.getElementById("output").innerHTML += "error: " + err + "<br>";
+        })
+    } else {
+        document.getElementById("output").innerHTML =
+            // If browser does not support SSE
+            "Sorry, the browser does not support SSE";
+    }
+</script>
+</body>
+</html>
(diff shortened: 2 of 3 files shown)


JacobCoffee (approved):

provinzkraut (approved):
Just a thought, something like this would be good to test end-to-end with a headless browser. @euri10 wdyt?

JacobCoffee:
@provinzkraut how would we do this?

provinzkraut:
Probably with [playwright](https://playwright.dev/python/docs/intro) or something.

euri10:
I'm all for it but it seems sse is not supported: https://github.com/microsoft/playwright/issues/7466
this is the best I came upwith and it sends error:

```
import asyncio
from contextlib import asynccontextmanager
from pathlib import Path
from typing import AsyncIterator

from playwright.async_api import Page, async_playwright

from litestar.config.cors import CORSConfig
from litestar.response import ServerSentEvent, ServerSentEventMessage
from litestar.types import SSEData


@get("/test_sse_empty")
async def handler() -> ServerSentEvent:
    async def generate() -> AsyncIterator[SSEData]:
        event = ServerSentEventMessage(event="empty")
        yield event

    return ServerSentEvent(generate())



html="""<!DOCTYPE html>
<html lang="en">
<body>
<h2>Server-Sent Events</h2>
<div id="output"></div>
<script>
    if (typeof EventSource !== "undefined") {
        var eventSource = new EventSource("http://localhost:8000/test_sse_empty");
        eventSource.onmessage = function (event) {
            // console.log(event.data);
            document.getElementById("output").innerHTML += event.data + "<br>";
        };
        eventSource.addEventListener("empty", (ev) => {
            // console.log("empty arrived");
            document.getElementById("output").innerHTML += "endx arrived" + "<br>";
            eventSource.close();
        });

        eventSource.addEventListener("error", (err) => {
            // console.error(err)
            document.getElementById("output").innerHTML += "error: " + err + "<br>";
        })
    } else {
        document.getElementById("output").innerHTML =
            // If browser does not support SSE
            "Sorry, the browser does not support SSE";
    }
</script>
</body>
</html>"""

from uvicorn.config import Config
from uvicorn.server import Server

@asynccontextmanager
async def run_server(config: Config, sockets=None):
    server = Server(config=config)
    task = asyncio.create_task(server.serve(sockets=sockets))
    await asyncio.sleep(0.1)
    try:
        yield server
    finally:
        await server.shutdown()
        task.cancel()

async def test_get_started_link(page: Page, tmp_path: Path ) -> None:
    app = Litestar(route_handlers=[handler], cors_config=CORSConfig(allow_origins=["*"]))
    index = tmp_path / "index.html"
    index.write_text(html)
    async with run_server(config=Config(app=app, port=8765, lifespan="off")):
        async with async_playwright() as p:
            browser = await p.chromium.launch()
            page = await browser.new_page()
            await page.goto(f"file://{index.as_posix()}")
            # Click the get started link.#
            output_div_content = await page.query_selector('div#output')
            print(await output_div_content.inner_text())
            await browser.close()


```

PR #3177, opened 2024-03-07
fix: close the underlying connection on psycopg channels backend

euri10:
## Description

- the underlying connection is not closed which might (or not) be a reason for those flaky runs

## Closes

nothing, trying to help on flaky tests, like this one https://github.com/litestar-org/litestar/actions/runs/8186487504/job/22384971272

### DIFF
--- a/litestar/channels/backends/psycopg.py
+++ b/litestar/channels/backends/psycopg.py
@@ -25,6 +25,7 @@ class PsycoPgChannelsBackend(ChannelsBackend):
         await self._exit_stack.enter_async_context(self._listener_conn)
 
     async def on_shutdown(self) -> None:
+        await self._listener_conn.close()
         await self._exit_stack.aclose()
 
     async def publish(self, data: bytes, channels: Iterable[str]) -> None:


guacs (approved):

litestar/channels/backends/psycopg.py
```diff
@@ -25,6 +25,7 @@ async def on_startup(self) -> None:
         await self._exit_stack.enter_async_context(self._listener_conn)
 
     async def on_shutdown(self) -> None:
+        await self._listener_conn.close()
```
provinzkraut, line 28: Hm. I think this shouldn't be needed. We add `self._listener_conn` to `self._exit_stack.enter_context` in `on_startup` and close the exit stack in `on_shutdown`. That should already close the connection.

provinzkraut (changes requested):
See comment

litestar/channels/backends/psycopg.py
euri10: yes you're right, my bad, loked at it too fast
just fr my understanding, any readon why you'r using a stack ?
seems like it works alright with just :
```
class PsycoPgChannelsBackend(ChannelsBackend):
    _listener_conn: psycopg.AsyncConnection

    def __init__(self, pg_dsn: str) -> None:
        self._pg_dsn = pg_dsn
        self._subscribed_channels: set[str] = set()
        # self._exit_stack = AsyncExitStack()

    async def on_startup(self) -> None:
        self._listener_conn = await psycopg.AsyncConnection.connect(self._pg_dsn, autocommit=True)
        # await self._exit_stack.enter_async_context(self._listener_conn)

    async def on_shutdown(self) -> None:
        await self._listener_conn.close()
        # await self._exit_stack.aclose()
```

provinzkraut: I think when I started out I had more than one contextmanager to close. Other than that, no particular reason :shrug:

PR #3180, opened 2024-03-07
feat: allow for console output to be silenced

cofin:
## Description

In some cases, I've wanted to change the name of the "Application" to show something else instead of "Litestar".  This is usually to make a CLI feel more cohesive and part of a larger application.  In the same vein, I've found cases where I wanted to complete suppress the initial `from_env` or the Rich info table at startup.

This PR allows each of these to be configured with optional environment variables. 

-


provinzkraut (commented):
I think we could just do `console.quiet = True` if we want to silence it.

cofin:
> I think we could just do `console.quiet = True` if we want to silence it.

I didn't use something like this because i thought it might silence some valid commands that we'd want to print out from the other CLI commands or plugins.  The idea was to just get some of the noise out of the way.  

Maybe I've misunderstood your idea though?

litestar/cli/_utils.py
```diff
@@ -107,10 +107,13 @@ def from_env(cls, app_path: str | None, app_dir: Path | None = None) -> Litestar
...
             dotenv.load_dotenv()
         app_path = app_path or getenv("LITESTAR_APP")
+        app_name = getenv("LITESTAR_APP_NAME") or "Litestar"
+        quiet_console = getenv("LITESTAR_QUIET_CONSOLE") or False
         if app_path and getenv("LITESTAR_APP") is None:
             os.environ["LITESTAR_APP"] = app_path
         if app_path:
-            console.print(f"Using Litestar app from env: [bright_blue]{app_path!r}")
+            if not quiet_console:
+                console.print(f"Using {app_name} from env: [bright_blue]{app_path!r}")
```
JacobCoffee, line 116:
```suggestion
                console.print(f"Using {app_name} app from env: [bright_blue]{app_path!r}")
```

litestar/cli/commands/core.py
```diff
@@ -189,7 +189,7 @@ def run_command(
 
     if pdb:
         os.environ["LITESTAR_PDB"] = "1"
-
+    quiet_console = os.getenv("LITESTAR_QUIET_CONSOLE") or False
```
JacobCoffee, line 192: Can we expose this from `_utils.py` instead of having another call to `getenv()` like we do the other below?

tests/unit/test_cli/test_core_commands.py
```diff
@@ -278,6 +280,58 @@ def test_run_command_debug(
...
+    console = Console(file=io.StringIO())
+    monkeypatch.setattr(_utils, "console", console)
+
+    path = create_app_file("_create_app_with_path.py", content=CREATE_APP_FILE_CONTENT)
+    app_path = f"{path.stem}:create_app"
+    monkeypatch.delenv("LITESTAR_QUIET_CONSOLE", raising=False)
+    result = runner.invoke(cli_command, ["--app", app_path, "run"])
+    assert result.exit_code == 0
+    normal_output = console.file.getvalue()  # type: ignore[attr-defined]
+    assert "Using Litestar from env:" in normal_output
```
JacobCoffee, line 296:
```suggestion
    assert "Using Litestar app from env:" in normal_output
```
if changed in the above `_utils.py`

tests/unit/test_cli/test_core_commands.py
```diff
@@ -278,6 +280,58 @@ def test_run_command_debug(
...
+    del result
+    console = Console(file=io.StringIO())
+    monkeypatch.setattr(_utils, "console", console)
+    monkeypatch.setenv("LITESTAR_QUIET_CONSOLE", "1")
+    assert os.getenv("LITESTAR_QUIET_CONSOLE") == "1"
+    result = runner.invoke(cli_command, ["--app", app_path, "run"])
+    assert result.exit_code == 0
+    quiet_output = console.file.getvalue()  # type: ignore[attr-defined]
+    assert "Starting server process" not in result.stdout
+    assert "Using Litestar from env:" not in quiet_output
```
JacobCoffee, line 307:
```suggestion
    assert "Using Litestar from env:" not in quiet_output
```
if changed in the above `_utils.py`

tests/unit/test_cli/test_core_commands.py
```diff
@@ -278,6 +280,58 @@ def test_run_command_debug(
...
+    _output = console.file.getvalue()  # type: ignore[attr-defined]
+    assert "Using Litestar from env:" in _output
+    console = Console(file=io.StringIO())
+    monkeypatch.setattr(_utils, "console", console)
+    monkeypatch.setenv("LITESTAR_APP_NAME", "My Stuff")
+    assert os.getenv("LITESTAR_APP_NAME") == "My Stuff"
+    result = runner.invoke(cli_command, ["--app", app_path, "run"])
+    assert result.exit_code == 0
+    _output = console.file.getvalue()  # type: ignore[attr-defined]
+    assert "Using My Stuff from env:" in _output
```
JacobCoffee, line 332:
```suggestion
    assert "Using My Stuff app from env:" in _output
```
if changed in the above `_utils.py`

JacobCoffee (approved):

JacobCoffee:
Needs rebase looks like?

litestar/cli/_utils.py
cofin: I purposely removed this word to make the language read a bit more clearlty.

PR #3185, opened 2024-03-09
fix(logging): queue_listener handler for Python >= 3.12

### LINKED ISSUE #2954: Bug: Logs and Stack traces are not visible OOTB in Python3.12
### Description

Logging (on routes) and also stack trace (when using `litestar.testing.create_test_client` in debug mode) don't appear to be printed on the terminal when executed in Python 3.12. Running the minimal example in Python 3.8,  I can see contents of the "Logs" section in my terminal; both a `WARNING` and an `ERROR` log, along with a stack trace. However, neither is visible when run on Python 3.12. There is one workaround (that I am aware of) shared by @cofin in the Discord server but (IMO) these functionalities should simply "just work".

> @cofin's example modified to suit the example code (the actual MCVE is under the MCVE section)
This at least makes one of the log ("WARNING" log) show up on 3.12, but no "ERROR" log and stack trace.

```python
import logging
from litestar import get, Request
from litestar.logging.config import LoggingConfig


@get()
async def evil_route(request: Request) -> str:
    request.logger.warning("you can't see me")
    1 / 0
    return "so, you will never catch me"


with create_test_client(
    [evil_route],
    logging_config=LoggingConfig(
        root={"level": logging.getLevelName(logging.INFO), "handlers": ["console"]},
        formatters={
            "standard": {
                "format": "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
            }
        },
    ),
) as client:
    response = client.get("")
    assert response.status_code == 500
```

PS: This may probably be a duplicate issue, so feel free to close as one.

_No response_

### MCVE

```python
from litestar import get, Request

@get()
async def evil_route(request: Request) -> str:
    request.logger.warning("you can't see me")
    1 / 0
    return "so, you will never catch me"

with create_test_client([evil_route]) as client:
    response = client.get("")
    assert response.status_code == 500
```

### Steps to reproduce

```bash
1. Save as "some_python_file.py"
2. Run `python some_python_file.py`
4. See error (actually you can't :p that is the error)
```

### Screenshots

```bash
```

### Logs

```bash
WARNING - ,-555869153 - root - http - you can't see me
ERROR - ,-555869152 - litestar - middleware - exception raised on http connection to route /

  File "litestar/litestar/middleware/exceptions/middleware.py", line 192, in __call__
  File "litestar/litestar/routes/http.py", line 82, in handle
  File "litestar/litestar/routes/http.py", line 134, in _get_response_for_request
  File "litestar/litestar/routes/http.py", line 154, in _call_handler_function
    response_data, cleanup_group = await self._get_response_data(
  File "litestar/litestar/routes/http.py", line 207, in _get_response_data
    data = await route_handler.fn(**parsed_kwargs)
  File "litestar/app.py", line 7, in evil_route
    1 / 0
ZeroDivisionError: division by zero
  File "litestar/litestar/middleware/exceptions/middleware.py", line 192, in __call__
  File "litestar/litestar/routes/http.py", line 82, in handle
  File "litestar/litestar/routes/http.py", line 134, in _get_response_for_request
  File "litestar/litestar/routes/http.py", line 154, in _call_handler_function
    response_data, cleanup_group = await self._get_response_data(
  File "litestar/litestar/routes/http.py", line 207, in _get_response_data
    data = await route_handler.fn(**parsed_kwargs)
  File "litestar/app.py", line 7, in evil_route
    1 / 0
ZeroDivisionError: division by zero
```

### Litestar Version

804f0151fc9b03e81f6cb6cbb38a9dfd395ead79 (main as of issue creation)

### Platform

- [] Linux
- [X] Mac (sorry, I forgot I had to check this)
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/2954">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2954/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

jderrien:
## Description

- Fix the `queue_listener` handler for Python 3.12

Python 3.12 introduced a new way to configure `QueueHandler` and `QueueListener` via `logging.config.dictConfig()`. This described [here](https://docs.python.org/3/library/logging.config.html#configuring-queuehandler-and-queuelistener).

The listener still needs to be started & stopped, as previously. To do so, I've introduced `LoggingQueueListener`.

And as stated in the doc:

> Any custom queue handler and listener classes will need to be defined with the same initialization signatures as [QueueHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.QueueHandler) and [QueueListener](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.QueueListener).

## Closes

Closes #2954

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -2,13 +2,12 @@ from __future__ import annotations
 
 import sys
 from abc import ABC, abstractmethod
-from dataclasses import asdict, dataclass, field
+from dataclasses import asdict, dataclass, field, fields
 from importlib.util import find_spec
 from logging import INFO
 from typing import TYPE_CHECKING, Any, Callable, Literal, cast
 
 from litestar.exceptions import ImproperlyConfiguredException, MissingDependencyException
-from litestar.serialization import encode_json
 from litestar.serialization.msgspec_hooks import _msgspec_json_encoder
 from litestar.utils.deprecation import deprecated
 
@@ -49,7 +48,20 @@ default_handlers: dict[str, dict[str, Any]] = {
 }
 
 if sys.version_info >= (3, 12, 0):
-    default_handlers["queue_listener"]["handlers"] = ["console"]
+    default_handlers["queue_listener"].update(
+        {
+            "class": "logging.handlers.QueueHandler",
+            "queue": {
+                "()": "queue.Queue",
+                "maxsize": -1,
+            },
+            "listener": "litestar.logging.standard.LoggingQueueListener",
+            "handlers": ["console"],
+        }
+    )
+
+    # do not format twice, the console handler will do the job
+    del default_handlers["queue_listener"]["formatter"]
 
 
 default_picologging_handlers: dict[str, dict[str, Any]] = {
@@ -228,21 +240,25 @@ class LoggingConfig(BaseLoggingConfig):
             A 'logging.getLogger' like function.
         """
 
-        if "picologging" in str(encode_json(self.handlers)):
+        excluded_fields: tuple[str, ...]
+        if "picologging" in " ".join([handler["class"] for handler in self.handlers.values()]):
             try:
                 from picologging import config, getLogger
             except ImportError as e:
                 raise MissingDependencyException("picologging") from e
 
-            values = {
-                k: v
-                for k, v in asdict(self).items()
-                if v is not None and k not in ("incremental", "configure_root_logger")
-            }
+            excluded_fields = ("incremental", "configure_root_logger")
         else:
             from logging import config, getLogger  # type: ignore[no-redef, assignment]
 
-            values = {k: v for k, v in asdict(self).items() if v is not None and k not in ("configure_root_logger",)}
+            excluded_fields = ("configure_root_logger",)
+
+        values = {
+            _field.name: getattr(self, _field.name)
+            for _field in fields(self)
+            if getattr(self, _field.name) is not None and _field.name not in excluded_fields
+        }
+
         if not self.configure_root_logger:
             values.pop("root")
         config.dictConfig(values)
--- a/litestar/logging/standard.py
+++ b/litestar/logging/standard.py
@@ -41,7 +41,7 @@ else:
             Args:
                 queue: The queue to send messages to
                 *handlers: A list of handlers which will handle entries placed on the queue
-                respect_handler_level: If respect_handler_level is True, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
+                respect_handler_level: If ``respect_handler_level`` is ``True``, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
             """
             super().__init__(queue, *handlers, respect_handler_level=respect_handler_level)
             self.start()
--- a/litestar/logging/standard.py
+++ b/litestar/logging/standard.py
@@ -36,7 +36,7 @@ else:
         """Custom ``QueueListener`` which starts and stops the listening process."""
 
         def __init__(self, queue: Queue[LogRecord], *handlers: Handler, respect_handler_level: bool = False) -> None:
-            """Initialize `LoggingQueueListener`.
+            """Initialize ``LoggingQueueListener``.
 
             Args:
                 queue: The queue to send messages to
--- a/litestar/logging/standard.py
+++ b/litestar/logging/standard.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import atexit
-import sys
 from logging import Handler, LogRecord, StreamHandler
 from logging.handlers import QueueHandler, QueueListener
 from queue import Queue
@@ -9,40 +8,40 @@ from typing import Any
 
 from litestar.logging._utils import resolve_handlers
 
-__all__ = ("QueueListenerHandler",)
+__all__ = ("LoggingQueueListener", "QueueListenerHandler")
 
 
-if sys.version_info < (3, 12):
+class LoggingQueueListener(QueueListener):
+    """Custom ``QueueListener`` which starts and stops the listening process."""
 
-    class QueueListenerHandler(QueueHandler):
-        """Configure queue listener and handler to support non-blocking logging configuration."""
+    def __init__(self, queue: Queue[LogRecord], *handlers: Handler, respect_handler_level: bool = False) -> None:
+        """Initialize ``LoggingQueueListener``.
 
-        def __init__(self, handlers: list[Any] | None = None) -> None:
-            """Initialize `QueueListenerHandler`.
+        Args:
+            queue: The queue to send messages to
+            *handlers: A list of handlers which will handle entries placed on the queue
+            respect_handler_level: If ``respect_handler_level`` is ``True``, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
+        """
+        super().__init__(queue, *handlers, respect_handler_level=respect_handler_level)
+        self.start()
+        atexit.register(self.stop)
 
-            Args:
-                handlers: Optional 'ConvertingList'
-            """
-            super().__init__(Queue(-1))
-            handlers = resolve_handlers(handlers) if handlers else [StreamHandler()]
-            self.listener = QueueListener(self.queue, *handlers)
-            self.listener.start()
 
-            atexit.register(self.listener.stop)
+class QueueListenerHandler(QueueHandler):
+    """Configure queue listener and handler to support non-blocking logging configuration.
 
-else:
+    .. caution::
 
-    class LoggingQueueListener(QueueListener):
-        """Custom ``QueueListener`` which starts and stops the listening process."""
+        This handler doesn't work with Python >= 3.12 and ``logging.config.dictConfig``. It might
+        be deprecated in the future. Please use ``logging.QueueHandler`` instead.
+    """
 
-        def __init__(self, queue: Queue[LogRecord], *handlers: Handler, respect_handler_level: bool = False) -> None:
-            """Initialize ``LoggingQueueListener``.
+    def __init__(self, handlers: list[Any] | None = None) -> None:
+        """Initialize ``QueueListenerHandler``.
 
-            Args:
-                queue: The queue to send messages to
-                *handlers: A list of handlers which will handle entries placed on the queue
-                respect_handler_level: If ``respect_handler_level`` is ``True``, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
-            """
-            super().__init__(queue, *handlers, respect_handler_level=respect_handler_level)
-            self.start()
-            atexit.register(self.stop)
+        Args:
+            handlers: Optional 'ConvertingList'
+        """
+        super().__init__(Queue(-1))
+        handlers = resolve_handlers(handlers) if handlers else [StreamHandler()]
+        self.listener = LoggingQueueListener(self.queue, *handlers)  # type: ignore[arg-type]
--- a/litestar/logging/standard.py
+++ b/litestar/logging/standard.py
@@ -33,7 +33,7 @@ if sys.version_info < (3, 12):
 else:
 
     class LoggingQueueListener(QueueListener):
-        """Custom `QueueListener` which starts and stops the listening process."""
+        """Custom ``QueueListener`` which starts and stops the listening process."""
 
         def __init__(self, queue: Queue[LogRecord], *handlers: Handler, respect_handler_level: bool = False) -> None:
             """Initialize `LoggingQueueListener`.
--- a/litestar/logging/standard.py
+++ b/litestar/logging/standard.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import atexit
 import sys
-from logging import StreamHandler
+from logging import Handler, LogRecord, StreamHandler
 from logging.handlers import QueueHandler, QueueListener
 from queue import Queue
 from typing import Any
@@ -18,7 +18,7 @@ if sys.version_info < (3, 12):
         """Configure queue listener and handler to support non-blocking logging configuration."""
 
         def __init__(self, handlers: list[Any] | None = None) -> None:
-            """Initialize `?QueueListenerHandler`.
+            """Initialize `QueueListenerHandler`.
 
             Args:
                 handlers: Optional 'ConvertingList'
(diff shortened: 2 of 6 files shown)


provinzkraut (approved):
@JacobCoffee Can you take a look at this? Looks good from my side.

provinzkraut:
@jderrien Thanks for the PR :)
Can you expand the PR description a bit, e.g. how this changes fixes the referenced issue?

jderrien:
> @jderrien Thanks for the PR :) Can you expand the PR description a bit, e.g. how this changes fixes the referenced issue?

I've updated the description, don't hesitate to comment the PR if you need more explainations. :)

Also, please wait before merging this. It seems some test jobs hung on on GitHub Actions. It runs well locally since I added `cleanup_logging()`, I'm not sure what's going on. :/

jderrien:
> It seems some test jobs hung on on GitHub Actions

https://github.com/litestar-org/litestar/actions/runs/8238535303
https://github.com/litestar-org/litestar/actions/runs/8250579528

The test suite on Python 3.12 takes ~10 min and gets killed.

While I didn't have this issue on macOS, I'll try to run the test suite in a Linux container tomorrow. Hopefully I'll be able to reproduce the issue.

peterschutt:
I did some testing yesterday and found it to be flaky, sometimes it would hang and sometimes not. I re-requested the 3.12 tests in CI and they passed that time which confirms it. Unfortunately, when it did hang I'd get different behaviors after hitting `^C`, sometimes the tests would continue and pass, sometimes I'd get errors out of pytest plugins, and sometimes errors out of low level threading stuff, but nothing that would actually point to any of our code :/

litestar/logging/config.py
```diff
@@ -49,7 +49,17 @@
 }
 
 if sys.version_info >= (3, 12, 0):
-    default_handlers["queue_listener"]["handlers"] = ["console"]
+    default_handlers["queue_listener"].update(
+        {
+            "class": "logging.handlers.QueueHandler",
+            "queue": Queue(-1),
```
jderrien, line 55: Instantiating a `queue.Queue` object in the handler config might be the source of the "test suite hanging" problem.

peterschutt (changes requested):
Flaky test.

Just adding this so we don't accidentally merge.

litestar/logging/config.py
peterschutt: Perhaps - maybe try marking all of those logging tests with `pytest.mark.xdist("something")` so they all run in the same process and see if that helps?

jderrien: In fact, I guess `litestar/logging/config.py` and so `default_handlers` might be loaded in different tests and context. Instanciating a `Queue` object directly in the config seems to be a bad idea in most cases.

Instead, as mentioned in the Python doc, I can also use a factory. From what I've seen locally, it fixes the problem. I'm going to push the change, but not sure how I can run the test suite multiple times on Github Actions to check if everything is fine... I can just amend the commit without any change to refresh the commit id. But maybe you'd have a better idea?

provinzkraut: > I can just amend the commit without any change to refresh the commit id. But maybe you'd have a better idea?

You can also just push an empty commit if you want to trigger a rerun

jderrien:
> Flaky test.
> 
> Just adding this so we don't accidentally merge.

@guacs advised me to use Github Actions directly on my fork, so I did.

I worked on [a specific branch](https://github.com/jderrien/litestar/tree/fix-logging-queue-listener-handler-testing), the only change is [this commit](https://github.com/jderrien/litestar/commit/c1cee6521728cd57d1707640ffcf8a0d8b73b32f).

On [these 25 runs](https://github.com/jderrien/litestar/actions/workflows/ci.yml) (26 but the first one failed due to a config mistake), everything looks good, at least on Python 3.12.

There were 3 failures on Python 3.11 on `tests/unit/test_channels/test_plugin.py::test_create_ws_route_handlers_arbitrary_channels_allowed` (which is already marked as flaky).

The problem seems to be solved.

litestar/logging/standard.py
```diff
@@ -18,7 +18,7 @@ class QueueListenerHandler(QueueHandler):
         """Configure queue listener and handler to support non-blocking logging configuration."""
 
         def __init__(self, handlers: list[Any] | None = None) -> None:
-            """Initialize `?QueueListenerHandler`.
+            """Initialize `QueueListenerHandler`.
```
jderrien, line 21: Was this a typo?

[The doc preview](https://litestar-org.github.io/litestar-docs-preview/3185/reference/logging/standard.html) seems to be broken, compared to the [official one](https://docs.litestar.dev/2/reference/logging/standard.html).

JacobCoffee: Im looking into this

JacobCoffee: @provinzkraut im not sure what makes it not load inside the preview, or if on next docs deploy the production docs would also stop loading?

I cant get them to load in any sort of configuration of Python version venvs and other means...
https://paste.pythondiscord.com/FKRQ is what I have for adding some notes to this, but i cant get our custom `QueueListenerHandler` to document itself in any way. Maybe it will be fine

litestar/logging/standard.py
```diff
@@ -31,4 +31,18 @@ def __init__(self, handlers: list[Any] | None = None) -> None:
             atexit.register(self.listener.stop)
 
 else:
-    QueueListenerHandler = QueueHandler
+
+    class LoggingQueueListener(QueueListener):
+        """Custom `QueueListener` which starts and stops the listening process."""
```
JacobCoffee, line 36:
```suggestion
        """Custom ``QueueListener`` which starts and stops the listening process."""
```

litestar/logging/standard.py
```diff
@@ -31,4 +31,18 @@ def __init__(self, handlers: list[Any] | None = None) -> None:
...
+
+        def __init__(self, queue: Queue[LogRecord], *handlers: Handler, respect_handler_level: bool = False) -> None:
+            """Initialize `LoggingQueueListener`.
```
JacobCoffee, line 39:
```suggestion
            """Initialize ``LoggingQueueListener``.
```

litestar/logging/standard.py
```diff
@@ -31,4 +31,18 @@ def __init__(self, handlers: list[Any] | None = None) -> None:
...
+
+            Args:
+                queue: The queue to send messages to
+                *handlers: A list of handlers which will handle entries placed on the queue
+                respect_handler_level: If respect_handler_level is True, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
```
JacobCoffee, line 44:
```suggestion
                respect_handler_level: If ``respect_handler_level`` is ``True``, a handler's level is respected (compared with the level for the message) when deciding whether to pass messages to that handler
```

JacobCoffee (approved):
lgtm aside from [weird page load](https://litestar-org.github.io/litestar-docs-preview/3185/reference/logging/standard.html)

peterschutt:
> > Flaky test.
> > Just adding this so we don't accidentally merge.
> 
> @guacs advised me to use Github Actions directly on my fork, so I did.
> 
> I worked on [a specific branch](https://github.com/jderrien/litestar/tree/fix-logging-queue-listener-handler-testing), the only change is [this commit](https://github.com/jderrien/litestar/commit/c1cee6521728cd57d1707640ffcf8a0d8b73b32f).
> 
> On [these 25 runs](https://github.com/jderrien/litestar/actions/workflows/ci.yml) (26 but the first one failed due to a config mistake), everything looks good, at least on Python 3.12.
> 
> There were 3 failures on Python 3.11 on `tests/unit/test_channels/test_plugin.py::test_create_ws_route_handlers_arbitrary_channels_allowed` (which is already marked as flaky).
> 
> The problem seems to be solved.

Fantastic! Thanks very much!

peterschutt (approved):

litestar/logging/standard.py
provinzkraut: It's because this is hidden in an if block 🤷

jderrien: @provinzkraut the `if` block is already on the `main` branch and the doc is correct? 🤔

provinzkraut: But the docs build on `3.12`. For `3.12`, the `QueueListenerHandler` isn't defined. This is a general issue and a breaking change by the way. The previous version had `QueueListenerHandler` defined in both branches.

jderrien: @provinzkraut I guess I understand better.

But it means [the official doc](https://docs.litestar.dev/latest/reference/logging/standard.html) has been generated with Python < 3.12 since it refers to the former `QueueListenerHandler` class (not to the aliased one), right?

Or does it mean the official doc is not up-to-date (not generated/published recently)?

While I understand this is a breaking change at the API level, I'm not sure what I should do next. Any advice?

provinzkraut: I'm not 100% sure, but I think the issue is that the `LoggingQueueListener` isn't in `__all__` and what *is* in `__all__` (`QueueListenerHandler`), isn't available on < 3.12. My guess would be that if you rename ``LoggingQueueListener` > `QueueListenerHandler`, it should work. If not, then I'm wrong and the issue is something else

jderrien: You're right about `__all__`.

That's said, renaiming `LoggingQueueListener` to `QueueListenerHandler` is not a good solution IMO, because this is a `QueueListener` object and not a `logging.Handler` one. I think naming is important. Also, the constructor args are not the same and I don't think I can change that... So it would be confusing, and the API doc would be incorrect.

I could keep `QueueListenerHandler = QueueHandler`, but it won't help to get `QueueListenerHandler` fully documented (for Python < 3.12).

I could remove the `if` condition, it shouldn't be a problem (technically). Even if it cannot be used with Python >= 3.12, it can be documented. Or I can keep the `if` condition and redefine `QueueListenerHandler` with the same constructor signature, and raise an exception (or a warning) stating it cannot be used with Python > 3.12.

I'm not sure how to manage this at the project level.

I tried to take a second look to check if we could keep `QueueListenerHandler` around. 👀 And from what I've understood, the signature has to be changed so it would break the API anyway. This is because the handler is instanciated with a single argument, which is the queue (and not `handlers: list[Any] | None`): https://github.com/python/cpython/blob/v3.12.2/Lib/logging/config.py#L744 -- `_configure_queue_handler` has been introduced in Python 3.12.

provinzkraut: I think for now we'll have to just keep the `QueueListenerHandler = QueueHandler` around somehow. `QueueListenerHandler` *and* `LoggingQueueListener` need to always be defined, no matter what Python version is being used. They should also be both in `_all__`. 

> I could remove the if condition, it shouldn't be a problem (technically). Even if it cannot be used with Python >= 3.12, it can be documented.

I think that's the best approach for now.

jderrien: Sorry if all of this is confusing, but removing the `if` statement AND keeping `QueueListenerHandler = QueueHandler` won't work. :)

To clarify (and if I'm not mistaken):

Currently (`main` branch):
⛔️ The default `queue_listener` handler is broken on Python 3.12
⛔️ Generating the API doc shows `QueueListenerHandler` as an alias of `logging.QueueHandler`, which is not the case for Python < 3.12
⚠️ Official API doc still shows the previous implementation (Python < 3.12) for an unknown reason (?)
⛔️ When upgrading to Python 3.12, the constructor signature of`QueueListenerHandler` is not the same anymore, which implies a breaking change for any "custom" usage of `QueueListenerHandler`

Keeping `QueueListenerHandler = QueueHandler` won't help to get a correct API doc for Python < 3.12.

I guess removing the `if` statement seems better:
✅ The default `queue_listener` handler will work out of the box
✅ API doc will show the informations for `QueueListenerHandler` (and `LoggingQueueListener`)
✅ The fact it doesn't work with Python >= 3.12 can be documented
✅ `QueueListenerHandler` will be able to be deprecated later (maybe we could/should add a warning right now?)
⛔️ Any usage relying on `QueueListenerHandler = QueueHandler` will break (that's said, as of today, the official API doc doesn't mention this aliasing on Python >= 3.12) -- I guess it might be a problem mostly if some custom configs have been put in place because it was not working out of the box with Python 3.12.
⛔️ Usage of `QueueListenerHandler` will break when upgrading to Python 3.12 for any custom implementation using it. But it can be documented. And using `QueueListenerHandler = QueueHandler` wouldn't help, because the constructor signatures are not the same (as stated before).

In all cases, there is a breaking change we can't avoid due to Python 3.12. So the best we can do is making it clear. Also, I'm not sure anyone is using `QueueListenerHandler` outside of the `default_handlers` provided, so the change might be minimal (that's said, I'm new to Litestar).

litestar/logging/standard.py
```diff
@@ -1,34 +1,45 @@
...
+
+    def __init__(self, handlers: list[Any] | None = None) -> None:
+        """Initialize `QueueListenerHandler`.
+
+        Args:
+            handlers: Optional 'ConvertingList'
+        """
+        super().__init__(Queue(-1))
+        handlers = resolve_handlers(handlers) if handlers else [StreamHandler()]
+        self.listener = LoggingQueueListener(self.queue, *handlers)  # type: ignore[arg-type]
```
jderrien, line 45: mypy was returning:
```
litestar/logging/standard.py:45: error: Argument 1 to "LoggingQueueListener" has incompatible type "_QueueLike[Any]"; expected "Queue[LogRecord]"  [arg-type]
```

I don't know if I could do something better here. It seems I cannot import `_QueueLike`.

litestar/logging/standard.py
jderrien: @provinzkraut I updated the PR, preview doc is available [here](https://litestar-org.github.io/litestar-docs-preview/3185/reference/logging/standard.html).

Any recommendation?

provinzkraut: Thanks for the great summary @jderrien! The current implementation seems like the best solution to me. Let's just keep it like that.

provinzkraut (approved):

litestar/logging/standard.py
provinzkraut: I think this is fine. It seems that `QueueHandler.queue` isn't typed properly as `Queue[LogRecord]` internally, so nothing we can fix here.

provinzkraut:
@JacobCoffee Good to go from your end?

litestar/logging/standard.py
jderrien: @provinzkraut I guess I could type `LoggingQueueListener.queue` as `Any` and remove `type: ignore[arg-type]` here. But not sure it would be better.

provinzkraut: I think that'd be more confusing actually

JacobCoffee (approved):

PR #3194, opened 2024-03-13
fix: replace TestClient.__enter__ return type with Self

cbscsm:
## Description
`TestClient.__enter__` and `AsyncTestClient.__enter__` return `self`. If you inherit `TestClient`, `__enter__` method should return derived class's instance unless override the method. So I think `typing_extensions.Self` could be more flexible return type.


peterschutt (approved):
Makes sense, thanks @cbscsm!

PR #3204, opened 2024-03-14
feat: Support `schema_extra` in `Parameter` and `Body`

### LINKED ISSUE #3022: Enhancement: Allow customizing/extending `Parameter` schema
### Summary

Pydantic provides `Field(json_schema_extra=...)` which allows to do something like:

```py
def forbid_internal_labels(value):
    if re.match(r"^reserved-for-internal-use[0-9]+:.*$", value):
        raise ValueError(...)

class Foo(BaseModel):
    label: Annotated[str, AfterValidator(forbid_internal_labels), Field(json_schema_extra={
        "not": {
            "anyOf": [
                {
                    "type": "string",
                    "pattern": "^reserved-for-internal-use1:.*$"
                },
                {
                    "type": "string",
                    "pattern": "^reserved-for-internal-use2:.*$"
                },
            ]
        }
    })]
```

This would generate:

```json
                "type": "string",
                "not": {
                    "anyOf": [
                        {
                            "type": "string",
                            "pattern": "^reserved-for-internal-use1$"
                        },
                        {
                            "type": "string",
                            "pattern": "^reserved-for-internal-use2$"
                        },
                    ]
                }
```

Similar mechanism would needed for `Parameter`, to inject discriminators into the OpenAPI spec (although the validation is decoupled from the schema declaration).

The builtin generation uses `oneOf` for union types (IIRC Pydantic uses `anyOf` for those). But I don't think it's possible to declare something like above.

Workarounds at this point would be:

- To adjust via custom `Operation` class (did not try but I assume it could work)
- To adjust the OpenAPI generation on a more global level (by extending `OpenAPIPlugin`?)

Both are quite ugly hacks, while something like this should be supported out-of-the-box.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

tuukkamustonen:
## Description

This adds sort of a backdoor for modifying the generated OpenAPI spec.

The value is given as `dict[str, Any]` where the key must match with the
keyword parameter name in `Schema`. The values are used to override items in
the generated `Schema` object, so they must be in correct types (ie. not in
dictionary/json format).

The values are added at main level, without recursive merging (because we're
adjusting `Schema` object and not a dictionary). Recursive merge would be much
more work.

Chose not to implement the same for `ResponseSpec` because response models are
generated as schema components, while `ResponseSpec` can be locally different.
Handling the logic of creating new components when `schema_extra` is passed in
`ResponseSpec` would be extra effort, and isn't probably as important as being
able to adjust the inbound parameters, which are actually validated (and for
which the documentation is even more important, than for the response).

## Closes

#3022

### DIFF
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -116,7 +116,7 @@ class KwargDefinition:
     """Extensions to the generated schema.
 
     If set, will overwrite the matching fields in the generated schema.
-    
+
     .. versionadded:: 2.8.0
     """
 
@@ -237,6 +237,8 @@ def Parameter(
         title: String value used in the title section of the OpenAPI schema for the given parameter.
         schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
             schema.
+
+            .. versionadded:: 2.8.0
     """
     return ParameterKwarg(
         annotation=annotation,
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -116,6 +116,8 @@ class KwargDefinition:
     """Extensions to the generated schema.
 
     If set, will overwrite the matching fields in the generated schema.
+    
+    .. versionadded:: 2.8.0
     """
 
     @property
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -345,6 +345,8 @@ def Body(
         title: String value used in the title section of the OpenAPI schema for the given parameter.
         schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
             schema.
+
+            .. versionadded:: 2.8.0
     """
     return BodyKwarg(
         media_type=media_type,
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -112,6 +112,11 @@ class KwargDefinition:
     """A sequence of valid values."""
     read_only: bool | None = field(default=None)
     """A boolean flag dictating whether this parameter is read only."""
+    schema_extra: dict[str, Any] | None = field(default=None)
+    """Extensions to the generated schema.
+
+    If set, will overwrite the matching fields in the generated schema.
+    """
 
     @property
     def is_constrained(self) -> bool:
@@ -187,6 +192,7 @@ def Parameter(
     query: str | None = None,
     required: bool | None = None,
     title: str | None = None,
+    schema_extra: dict[str, Any] | None = None,
 ) -> Any:
     """Create an extended parameter kwarg definition.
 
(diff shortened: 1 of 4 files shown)


tuukkamustonen:
Uh, `JSONDecodeError` on Python 3.11... what what...

JacobCoffee:
I think this needs a rebase :)

tuukkamustonen:
> I think this needs a rebase :)

Done.

Is it common convention for you to force-push to `develop` branch, though? Makes all PRs outdated.

tuukkamustonen:
`make docs` seems to fail, is it broken also in the `develop`...?

peterschutt:
> `make docs` seems to fail, is it broken also in the `develop`...?

Yep, it's being worked on in
https://github.com/litestar-org/litestar/pull/3220

JacobCoffee:
It should not fail for you locally but it will give warnings

litestar/params.py
```diff
@@ -112,6 +112,11 @@ class KwargDefinition:
     """A sequence of valid values."""
     read_only: bool | None = field(default=None)
     """A boolean flag dictating whether this parameter is read only."""
+    schema_extra: dict[str, Any] | None = field(default=None)
+    """Extensions to the generated schema.
+
+    If set, will overwrite the matching fields in the generated schema.
```
JacobCoffee, line 118:
```suggestion
    If set, will overwrite the matching fields in the generated schema.
    
    .. versionadded:: 2.8.0
```

litestar/params.py
```diff
@@ -331,6 +341,8 @@ def Body(
         pattern: A string representing a regex against which the given string will be matched.
             Equivalent to pattern in the OpenAPI specification.
         title: String value used in the title section of the OpenAPI schema for the given parameter.
+        schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
+            schema.
```
JacobCoffee, line 345:
```suggestion
        schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
            schema.

            .. versionadded:: 2.8.0
```

litestar/params.py
```diff
@@ -227,6 +233,8 @@ def Parameter(
         required: A boolean flag dictating whether this parameter is required.
             If set to False, None values will be allowed. Defaults to True.
         title: String value used in the title section of the OpenAPI schema for the given parameter.
+        schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
+            schema.
```
JacobCoffee, line 237:
```suggestion
        schema_extra: Extensions to the generated schema. If set, will overwrite the matching fields in the generated
            schema.

            .. versionadded:: 2.8.0
```

JacobCoffee (approved):

tuukkamustonen:
@provinzkraut You seem to have a convention to force push to `develop`, which generates conflicts in the PRs. Is it intentional or...?

Should I rebase this or...?

cofin (approved):
Thanks for this!  LGTM

PR #3205, opened 2024-03-15
feat: Use memoized `request_class` and `response_class` values

kedod:
## Description

- Use memoized `request_class` and `response_class` values

### DIFF
--- a/litestar/handlers/http_handlers/base.py
+++ b/litestar/handlers/http_handlers/base.py
@@ -78,6 +78,7 @@ class HTTPRouteHandler(BaseRouteHandler):
         "_resolved_before_request",
         "_response_handler_mapping",
         "_resolved_include_in_schema",
+        "_resolved_request_class",
         "_resolved_tags",
         "_resolved_security",
         "after_request",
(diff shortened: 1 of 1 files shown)


cofin (approved):
LGTM

cofin:
@all-contributors add @kedod for code

peterschutt (changes requested):
Can we do `resolve_response_class()` at the same time? Its documented as being memoized just like this one, but isn't.

peterschutt:
For interests sake, the memoization of `resolve_response_class()` was removed in #357.

That PR added `resolve_response_handler()` which is itself cached and appears to be the only place that we call `resolve_response_class()` internally - so I can understand why its caching was removed in that case.

However, it has since always been documented as being cached, even though it isn't, and we offer it as public API, so if users are calling it more than once they'd benefit from the caching as documented and it wouldn't really cost us anything. So, now that we are caching `resolve_request_class()` I _think_ its it probably better to memoize it too so the similar methods are consistent.

In that PR we also removed caching for `resolve_response_cookies()` and `resolve_response_headers()`, but the docstrings of those methods were updated to reflect that.

provinzkraut (approved):

peterschutt (approved):

PR #3207, opened 2024-03-15
revert: "feat: allow for console output to be silenced"

cofin:
Reverts litestar-org/litestar#3180

Branch needed to be rebased before merging


cofin:
This is not needed.

PR #3208, opened 2024-03-15
docs: fix included line range from example

hugovk:
## Description

Now, https://docs.litestar.dev/dev/tutorials/sqlalchemy/2-serialization-plugin.html#compare-handlers-before-and-after-serialization-plugin shows the "Before" as lines 79-100:

This is https://github.com/litestar-org/litestar/blob/ef9dbb7a0cc4c7031dbf9a7bff02719d8a65a008/docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py#L79-L100

We need 10 more lines, 69-100:

https://github.com/litestar-org/litestar/blob/ef9dbb7a0cc4c7031dbf9a7bff02719d8a65a008/docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_no_plugins.py#L69-L100

PR #3209, opened 2024-03-15
docs: fix included line range from example

hugovk:
## Description

Similar to https://github.com/litestar-org/litestar/pull/3208 but for another page.

Fix https://docs.litestar.dev/dev/tutorials/sqlalchemy/4-final-touches-and-recap.html so the first example highlights this line:

https://github.com/litestar-org/litestar/blob/243733a004057ba86ea9c4bc94cbf51f58f3e0dc/docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_plugin.py#L83

And so the last example is:

https://github.com/litestar-org/litestar/blob/243733a004057ba86ea9c4bc94cbf51f58f3e0dc/docs/examples/contrib/sqlalchemy/plugins/tutorial/full_app_with_plugin.py#L80-L84


JacobCoffee (approved):
thank you again :)

PR #3215, opened 2024-03-16
feat(DTO): Enable codegen backend by default

provinzkraut:
Enable the codegen backend for DTOs introduced in #2388 by default. We have initially hidden this behind a feature flag to see how it works out in production. It's been smooth sailing for about 6 months now so I think we can enable it by default. This gives us enough time to decide if we want to remove the regular backend entirely in 3.0.

### DIFF
--- a/docs/usage/dto/0-basic-use.rst
+++ b/docs/usage/dto/0-basic-use.rst
@@ -78,11 +78,12 @@ DTOs can similarly be defined on :class:`Routers <litestar.router.Router>` and
 Improving performance with the codegen backend
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. admonition:: Experimental feature
-    :class: danger
+.. note::
 
-    This is an experimental feature and should be approached with caution. It may
-    behave unexpectedly, contain bugs and may disappear again in a future version.
+    This feature was introduced in ``2.2.0`` and hidden behind the ``DTO_CODEGEN``
+    feature flag. As of ``2.8.0`` it is considered stable and enabled by default. It can
+    still be disabled selectively by using the
+    ``DTOConfig(experimental_codegen_backend=True)`` override.
 
 
 The DTO backend is the part that does the heavy lifting for all the DTO features. It
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import inspect
 import logging
 import os
+import warnings
 from contextlib import (
     AbstractAsyncContextManager,
     AsyncExitStack,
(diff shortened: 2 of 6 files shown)


JacobCoffee (approved):
i wish we had some good metrics on usage here to see how many people are trying this out.

peterschutt (approved):

cofin:
I agree with this approach.  I've been using it for some time with no issues.

cofin (approved):

Alc-Alc:
> i wish we had some good metrics on usage here to see how many people are trying this out.

While we have two users (maintainers 🙂) using and approving this feature, I have to say I am kinda of "weary" (if this comes off as a strong word, my apologies) of this approach. I feel we should "promote" this experimental feature more. What I mean is, probably have a lot of links (backlinks?) that lead to https://docs.litestar.dev/latest/usage/dto/0-basic-use.html#enabling-the-backend.

Or even have a poll (or just a normal question) in the discord to get these two data points.
1. How many are aware and use this
2. How stable has it been

My reason for this comment is, *I* (not talking about other maintainers / members 🙂 ) do not think many are aware of this. How did I arrive at that conclusion? I have not seen a significant number of users ask about this in discord. To me that could mean either of the following
1. Anyone referring / getting inspired / copy pasting / carefully handpicking things from full stack have been using this and had zero issues. This is the "good ending" ✅  and we can go ahead.
2. Not many have used it and we don't have a wide spread opinion on this. This is the "bad ending" ❌ and maybe make some promotion or market this as a mandatory 3.0 alpha feature.

peterschutt:
> > i wish we had some good metrics on usage here to see how many people are trying this out.
> 
> While we have two users (maintainers 🙂) using and approving this feature, I have to say I am kinda of "weary" (if this comes off as a strong word, my apologies) of this approach. I feel we should "promote" this experimental feature more. What I mean is, probably have a lot of links (backlinks?) that lead to https://docs.litestar.dev/latest/usage/dto/0-basic-use.html#enabling-the-backend.
> 
> Or even have a poll (or just a normal question) in the discord to get these two data points.
> 
>     1. How many are aware and use this
> 
>     2. How stable has it been
> 
> 
> My reason for this comment is, _I_ (not talking about other maintainers / members 🙂 ) do not think many are aware of this. How did I arrive at that conclusion? I have not seen a significant number of users ask about this in discord. To me that could mean either of the following
> 
>     1. Anyone referring / getting inspired / copy pasting / carefully handpicking things from full stack have been using this and had zero issues. This is the "good ending" ✅  and we can go ahead.
> 
>     2. Not many have used it and we don't have a wide spread opinion on this. This is the "bad ending" ❌ and maybe make some promotion or market this as a mandatory 3.0 alpha feature.

I see your point @Alc-Alc  but the backend is supposed to be backward compatible - so as long as we do this with the guarantee that _any_ change in behavior should be considered a bug then this shouldn't be different to rolling out any other backward compatible optimization. Also, this PR doesn't remove the current backend, so there is a pretty easy escape hatch users can apply while we address any issues.

provinzkraut:
Definitely need to adjust the docs though, thanks for the reminder @Alc-Alc

Alc-Alc (approved):
I was here

PR #3220, opened 2024-03-17
docs: add missing `alembic` class references

cofin:
## Description

This PR adds  `alembic` to the intersphinx map  and adds `advanced_alchemy` entries into the nitpick ignore list


cofin:
This should address the failure we see in the #3216 test run

docs/conf.py
```diff
@@ -156,16 +156,19 @@
     (PY_CLASS, "litestar.typing.ParsedType"),
     (PY_METH, "litestar.dto.factory.DTOData.create_instance"),
     (PY_METH, "litestar.dto.interface.DTOInterface.data_to_encodable_type"),
+    (PY_CLASS, "MetaData"),
     (PY_CLASS, "advanced_alchemy.repository.typing.ModelT"),
     (PY_OBJ, "advanced_alchemy.config.common.SessionMakerT"),
     (PY_OBJ, "advanced_alchemy.config.common.ConnectionT"),
     (PY_CLASS, "advanced_alchemy.extensions.litestar.plugins._slots_base.SlotsBase"),
     (PY_CLASS, "advanced_alchemy.config.EngineConfig"),
+    (PY_CLASS, "advanced_alchemy.config.common.GenericAlembicConfig"),
```
peterschutt, line 165: I think this class isn't being documented in the advanced-alchemy docs b/c it isn't referenced in `__all__` in the `advanced_alchemy.config.common` module. Is that intentional? If that was added there I _think_ that would make the class available via intersphinx and this ignore not necessary. (a lot of "i think"'s here because often what I think would fix these docs builds doesn't)..

peterschutt (changes requested):
We don't need to ignore the alembic one as we can just bring in the alembic intersphinx like we do for other libraries.

Up to you if you want to act on my point about trying to fix the `GenericAlembicConfig` one upstream as I see we are already ignoring a stack of warnings from AA so I won't nitpick on this one.

docs/conf.py
JacobCoffee: I would like this over not documenting if possible

cofin: Well, the user never usies the GenericAlembicConfig.  This is the base class for the Sync/Async alembic configs.

peterschutt: For the same reason that `GenericSQLAlchemyConfig` and `GenericSessionConfig` is documented, I think so. It means that users can click on the links to the base class and see the docs for fields defined on the base.

cofin: I've opened up a PR on AA to add this to `__all__`.  Let me know if you want to wait until that's addressed before merging.  

I figure we could hit some of the documentation ignores en masse on a separate PR.

peterschutt: OK sounds good.

peterschutt (approved):

Issue #2849, opened 2023-12-06
Enhancement: Support OpenAPI examples defined in Structs, Pydantic models and dataclasses


bogdanteleaga:
### Description

Examples provided on json parameter fields do not get rendered in the Swagger UI.

The `/schema/openapi.json` does get generated similarly to https://github.com/litestar-org/litestar/pull/2660, but it seems this does not result in the examples being actually rendered in the Swagger UI.

I included examples for defining the classes with any of pydantic, msgspec and dataclass.

### MCVE

```python
from litestar import post, Litestar
from typing import Annotated
from msgspec import Struct, Meta


class Lookup(Struct):
    id: Annotated[str, Meta(examples=["example-1", "example-2"])]

#class Lookup(BaseModel):
#    id: Annotated[str, Field(json_schema_extra={"example": "example-1"})

#@dataclass
#class Lookup:
#    id: Annotated[str, Parameter(examples=[Example(value="example-1"), Example(value="example-2")])


@post(path="/lookup")
async def lookup(lookup: Lookup) -> Lookup:
    return Lookup(id=lookup.id)


app = Litestar(route_handlers=[lookup])
```

### Steps to reproduce

```bash
1. Go to `http://127.0.0.1:8000/schema/swagger`
2. Click on `/lookup`
3. Scroll down to examples
4. They are missing
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Litestar Version

2.4.2

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh Litestar dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/2849">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/2849/pledge.svg?darkmode=1">
[...]

PR #3224, opened 2024-03-18
fix: JSON schema `examples` were OpenAPI formatted

tuukkamustonen:
## Description

The generated `examples` in *JSON schema* objects were formatted as:

```json
    "examples": {
      "some-id": {
        "description": "Lorem ipsum",
        "value": "the real beef"
      }
   }
```

However, above is OpenAPI example format, and must not be used in JSON schema
objects. Schema objects follow different formatting:

```json
    "examples": [
      "the real beef"
   ]
```

This is explained in
https://medium.com/apis-you-wont-hate/openapi-v3-1-and-json-schema-2019-09-6862cf3db959

Schema objects spec in
https://spec.openapis.org/oas/v3.1.0#schema-object.

OpenAPI example format spec in
https://spec.openapis.org/oas/v3.1.0#example-object. This is referenced at
least from parameters, media types and components.

The technical change here is to define `Schema.examples` as `list[Any]` instead
of `list[Example]`. Examples can and must still be defined as `list[Example]`
for OpenAPI objects (e.g. `Parameter`, `Body`) but for JSON schema `examples`
the code now internally generates/converts `list[Any]` format instead.

Extra confusion here comes from the OpenAPI 3.0 vs OpenAPI 3.1 difference.
OpenAPI 3.0 only allowed `example` (singular) field in schema objects.
OpenAPI 3.1 supports the full JSON schema 2020-12 spec and so `examples` array
in schema objects.

Both `example` and `examples` seem to be supported, though the former is marked
as deprecated in the latest specs.

This can be tested over at https://editor-next.swagger.io by loading up the
OpenAPI 3.1 Pet store example. Then add `examples` in `components.schemas.Pet`
using the both ways and see the Swagger UI only render the example once it's 
properly formatted (it ignores is otherwise).

## Closes

n/a

Partially fixes https://github.com/litestar-org/litestar/issues/2849 (the OpenAPI UIs don't render badly formatted examples in the spec).

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -44,7 +44,7 @@ from litestar._openapi.schema_generation.utils import (
     _should_create_enum_schema,
     _should_create_literal_schema,
     _type_or_first_not_none_inner_type,
-    get_json_schema_examples,
+    get_json_schema_formatted_examples,
 )
 from litestar.datastructures import UploadFile
 from litestar.exceptions import ImproperlyConfiguredException
@@ -558,7 +558,7 @@ class SchemaCreator:
                     not isinstance(value, Hashable) or not self.is_undefined(value)
                 ):
                     if schema_key == "examples":
-                        value = get_json_schema_examples(cast("list[Example]", value))
+                        value = get_json_schema_formatted_examples(cast("list[Example]", value))
 
                     # we only want to transfer values from the `KwargDefinition` to `Schema` if the schema object
                     # doesn't already have a value for that property. For example, if a field is a constrained date,
@@ -572,7 +572,7 @@ class SchemaCreator:
         if not schema.examples and self.generate_examples:
             from litestar._openapi.schema_generation.examples import create_examples_for_field
 
-            schema.examples = get_json_schema_examples(create_examples_for_field(field))
+            schema.examples = get_json_schema_formatted_examples(create_examples_for_field(field))
 
         if schema.title and schema.type == OpenAPIType.OBJECT:
             key = _get_normalized_schema_key(field.annotation)
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -44,7 +44,7 @@ from litestar._openapi.schema_generation.utils import (
     _should_create_enum_schema,
     _should_create_literal_schema,
     _type_or_first_not_none_inner_type,
-    get_formatted_examples,
+    get_json_schema_examples,
 )
 from litestar.datastructures import UploadFile
 from litestar.exceptions import ImproperlyConfiguredException
@@ -558,7 +558,7 @@ class SchemaCreator:
                     not isinstance(value, Hashable) or not self.is_undefined(value)
                 ):
                     if schema_key == "examples":
-                        value = get_formatted_examples(field, cast("list[Example]", value))
+                        value = get_json_schema_examples(field, cast("list[Example]", value))
 
                     # we only want to transfer values from the `KwargDefinition` to `Schema` if the schema object
                     # doesn't already have a value for that property. For example, if a field is a constrained date,
@@ -572,7 +572,7 @@ class SchemaCreator:
         if not schema.examples and self.generate_examples:
             from litestar._openapi.schema_generation.examples import create_examples_for_field
 
-            schema.examples = get_formatted_examples(field, create_examples_for_field(field))
+            schema.examples = get_json_schema_examples(field, create_examples_for_field(field))
 
         if schema.title and schema.type == OpenAPIType.OBJECT:
             key = _get_normalized_schema_key(field.annotation)
@@ -587,7 +587,7 @@ class SchemaCreator:
         property_fields: Mapping[str, FieldDefinition],
         openapi_type: OpenAPIType = OpenAPIType.OBJECT,
         title: str | None = None,
-        examples: Mapping[str, Example] | None = None,
+        examples: list[Any] | None = None,
     ) -> Schema:
         """Create a schema for the components/schemas section of the OpenAPI spec.
 
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -558,7 +558,7 @@ class SchemaCreator:
                     not isinstance(value, Hashable) or not self.is_undefined(value)
                 ):
                     if schema_key == "examples":
-                        value = get_json_schema_examples(field, cast("list[Example]", value))
+                        value = get_json_schema_examples(cast("list[Example]", value))
 
                     # we only want to transfer values from the `KwargDefinition` to `Schema` if the schema object
                     # doesn't already have a value for that property. For example, if a field is a constrained date,
@@ -572,7 +572,7 @@ class SchemaCreator:
         if not schema.examples and self.generate_examples:
             from litestar._openapi.schema_generation.examples import create_examples_for_field
 
-            schema.examples = get_json_schema_examples(field, create_examples_for_field(field))
+            schema.examples = get_json_schema_examples(create_examples_for_field(field))
 
         if schema.title and schema.type == OpenAPIType.OBJECT:
             key = _get_normalized_schema_key(field.annotation)
--- a/litestar/_openapi/schema_generation/utils.py
+++ b/litestar/_openapi/schema_generation/utils.py
@@ -3,12 +3,12 @@ from __future__ import annotations
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Mapping, _GenericAlias  # type: ignore[attr-defined]
 
+from litestar.openapi.spec import Example
 from litestar.utils.helpers import get_name
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
-    from litestar.openapi.spec import Example
     from litestar.typing import FieldDefinition
 
 __all__ = (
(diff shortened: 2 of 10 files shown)


guacs:
@tuukkamustonen thanks for this! This is something I'd noticed and was meaning to fix, but hadn't gotten the time to actually investigate.

tuukkamustonen:
Let's see if I can be type linters pass... interesting errors...

provinzkraut:
Is this a regression @guacs? I vaguely remember a PR that fixed a similar (or actually the same) bug?

guacs:
> Is this a regression @guacs? I vaguely remember a PR that fixed a similar (or actually the same) bug?

If I remember correctly, that PR was for the rendering of examples of things like path parameters, query parameters etc. This is for the examples of the schemas we generate.

tuukkamustonen:
I'm getting these errors even on latest `main`:

```
=> Running mypy
tests/e2e/test_pydantic.py:11: error: Function is missing a type annotation  [no-untyped-def]
tests/unit/test_contrib/test_pydantic/test_schema_plugin.py:73: error: Function is missing a type annotation  [no-untyped-def]
tests/unit/test_openapi/test_responses.py:1: error: Cannot find implementation or library stub for module named "litestar.openapi.plugins"  [import-not-found]
tests/unit/test_openapi/test_responses.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
tests/unit/test_security/test_security.py:1: error: Cannot find implementation or library stub for module named "litestar.openapi.plugins"  [import-not-found]
Found 4 errors in 4 files (checked 691 source files)
```

In `develop`:

```
=> Running mypy
tests/e2e/test_pydantic.py:11: error: Function is missing a type annotation  [no-untyped-def]
tests/unit/test_contrib/test_pydantic/test_schema_plugin.py:73: error: Function is missing a type annotation  [no-untyped-def]
Found 2 errors in 2 files (checked 697 source files)
```

tuukkamustonen:
Worked after rebase on latest `main` now, okay...

peterschutt:
> Is this a regression @guacs? I vaguely remember a PR that fixed a similar (or actually the same) bug?

I think this #2660 which was a fix for #2272

litestar/typing.py
```diff
@@ -81,10 +92,11 @@ def _parse_metadata(value: Any, is_sequence_container: bool, extra: dict[str, An
         **cast("dict[str, Any]", extra or getattr(value, "extra", None) or {}),
         **(getattr(value, "json_schema_extra", None) or {}),
     }
+    example_list: list[Any] | None
     if example := extra.pop("example", None):
-        example_list = [Example(value=example)]
+        example_list = [example]
     elif examples := getattr(value, "examples", None):
-        example_list = [Example(value=example) for example in cast("list[str]", examples)]
+        example_list = examples
```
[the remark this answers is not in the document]
peterschutt: Looking at where _we_ use it, changing the type of `KwargDefinition.examples` likely isn't that big a deal, but it is publicly documented as it is, and I don't know that changing its type to `Any` is very good.

I think this also means that `get_formatted_examples()` could receive `Sequence[Any]` now, is that right?

https://github.com/litestar-org/litestar/blob/024dccff1e6bf433d260c7fd2add7694eb05f63f/litestar/_openapi/schema_generation/utils.py#L103

litestar/_openapi/schema_generation/schema.py
```diff
@@ -558,7 +558,7 @@ def process_schema_result(self, field: FieldDefinition, schema: Schema) -> Schem
                     not isinstance(value, Hashable) or not self.is_undefined(value)
                 ):
                     if schema_key == "examples":
-                        value = get_formatted_examples(field, cast("list[Example]", value))
+                        value = get_json_schema_examples(field, cast("list[Example]", value))
```
peterschutt, line 561: The `cast` here would be `list[Any]` assuming we changed the type of `KwargDefinition.examples`

litestar/_openapi/schema_generation/utils.py
```diff
@@ -107,3 +107,14 @@ def get_formatted_examples(field_definition: FieldDefinition, examples: Sequence
     name = name.lower()
 
     return {f"{name}-example-{i}": example for i, example in enumerate(examples, 1)}
+
+
+def get_json_schema_examples(field_definition: FieldDefinition, examples: Sequence[Example | str]) -> list[Any]:
```
peterschutt, line 112:
```suggestion
def get_json_schema_examples(examples: Sequence[Example | str]) -> list[Any]:
```
Wonder why linters didn't get this.

peterschutt:
This was the MCVE from #2272:

```py
from typing_extensions import Annotated


class TestBody(Struct):
    field1: Annotated[str, Meta(examples=["the answer"])]
    field2: Annotated[int, Meta(examples=[42])]


@post(sync_to_thread=False)
def test(data: TestBody) -> None:
    ...


if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app=Litestar(route_handlers=[test]), port=8080)
```

If I run that app on `main` we get:
image

Which doesn't include the example values for those fields, so either we've had a regression or that wasn't fixed.

Same app on this branch:

image

So LGTM on that front.

guacs:
> This was the MCVE from #2272:
> 
> ```python
> from typing_extensions import Annotated
> from msgspec import Struct, Meta
> from litestar import Litestar, post
> 
> 
> class TestBody(Struct):
>     field1: Annotated[str, Meta(examples=["the answer"])]
>     field2: Annotated[int, Meta(examples=[42])]
> 
> 
> @post(sync_to_thread=False)
> def test(data: TestBody) -> None:
>     ...
> 
> 
> if __name__ == "__main__":
>     import uvicorn
>     uvicorn.run(app=Litestar(route_handlers=[test]), port=8080)
> ```
> 
> If I run that app on `main` we get: image
> 
> Which doesn't include the example values for those fields, so either we've had a regression or that wasn't fixed.
> 
> Same app on this branch:
> 
> image
> 
> So LGTM on that front.

@peterschutt how were you able to verify that things are working with the fix introduced in the linked PR? I tried running the app in this comment after checking out to `680c2df00`, but I still was getting this incorrect example. I think I may be doing something wrong.

litestar/_openapi/schema_generation/schema.py
tuukkamustonen: The `cast` here is for the `KwargDefinition.examples` field, which still is declared as `list[Example] | None` and passed into `get_json_schema_examples()`. It _should_ be correct afaik.

The return value is `list[Any]` which is for JSON schema (and not OpenAPI) examples. Also that _should_ be correct.

litestar/_openapi/schema_generation/utils.py
tuukkamustonen: Ha, indeed...

litestar/typing.py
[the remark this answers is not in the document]
tuukkamustonen: The method here picks up stuff from the field declaration, and puts into that dict what is relevant / needed to create... at least `Schema` entry <-- I think my understanding/assumption was that it's only used for `Schema` creation, but if it's also used for something else(?) then there may be something wrong here.

But that's different than the flow that you described.

I'm not changing `KwargDefinition.examples` type and must not change it. It will remain as `list[Example]`, as that's the syntax/format how _OpenAPI_ examples are declared currently and that's fine. But JSON schema `examples` need that `list[Any]` instead (as they're just arrays of values).

I guess the question here is exactly what this `_parse_metadata()` is used for - where is that return value used... I'll check again.

> I find this corner of the library really hard to follow

Yes, I would even describe it as _super_ complex and kludgy 😵‍💫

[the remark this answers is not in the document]
tuukkamustonen: Yeah, I believe you're righting. I reverted that in a fixup commit.

Funnily, changing it one way or around doesn't break anything. For sure there's _some_ impact (in the OpenAPI output), but I guess it's not unit test covered.

tuukkamustonen:
@peterschutt Fixed as suggested in the fixup commit. Great that you spotted that! Take another look.

@guacs I also tried the MVCE there, and it does work for me 🤷🏻‍♂️

peterschutt:
> @peterschutt how were you able to verify that things were working with the fix introduced in the linked PR? I tried running the app in this comment after checking out to `680c2df00`, but I still was getting this incorrect example. I think I may be doing something wrong.

My verification was that it was incorrect on `main` and `correct` on this branch - given that you get the incorrect behavior on `680c2df00`, and I get same on `main` and `680c2df00` was merged into main, I think we are saying the same thing. Unless I'm misunderstanding (very possible :D)

litestar/typing.py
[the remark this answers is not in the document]
peterschutt: > but if it's also used for something else(?)

Its used on any occasion that we parse an annotation that has _any object_ that defines those openapi/schema constraint attributes in its metadata, e.g., `query_value: Annotated[int, Parameter(example=...)]` on a handler, and also `field_on_a_model: Annotated[int, Meta(example=...)]` on some domain model. Both of these cases get end up with the example value getting shoehorned into `Example`, which - as your recent work has brought to light - is not a correct abstraction for the case where the example should exist on a json schema document.

However imperfect it is that we are using `Example` to carry around that json schema example value internally here between model parsing and schema generation - I guess "it is what it is" for now.

> Yes, I would even describe it as super complex and kludgy 😵‍💫

`FieldDefinition` used to be called `ParsedType` - and its only purpose was to represent a type annotation. It would handle extraction of meta values out of `Annotated`, but not make any assumptions about how that metadata should be used - just a container for type information.

I don't think we need to do all of this metadata parsing/merging/transforming here because its got nothing to do with parsing type annotations and everything to do with schema generation, so should probably exist in `litestar._openapi`. Problem there is that at the same time `ParsedType` turned into `FieldDefinition`, `FieldDefinition.kwarg_definition` was added to the public api, which is the result of all of this metadata wrangling. So its probably going to have to be a 3.0 thing to unwind.

litestar/_openapi/schema_generation/schema.py
peterschutt: My understanding is that it is correct now given the rollback on changes to metadata parsing - my point was that we were parsing the data that feeds `KwargDefinition.examples` with `list[Any]`.. anyhow I think its moot now.

peterschutt (approved):
Thanks @tuukkamustonen!

winstxnhdw:
Hey, is there a reason why this doesn't work when using class-based controllers?

PR #3225, opened 2024-03-18
feat: Support class level `json_schema_extra` for Pydantic models

tuukkamustonen:
## Description

`json_schema_extra` was already supported for individual fields (in
`litestar.typing._parse_metadata`), just not for the whole model.

This will allow to specify examples (and other) within Pydantic models, e.g.:

```py
class MyModel(BaseModel):
    model_config = ConfigDict(
        json_schema_extra={
            "examples": ["Alice", "Bob"]
        }
    )

    name: str
```

This change also drops support for looking up an example from
`ConfigDict(example=...)` because Pydantic doesn't define such a field:

https://docs.pydantic.dev/1.10/usage/model_config/
https://docs.pydantic.dev/latest/api/config/

The technical change is addition of `Schema.extra`, which can store the
`json_schema_extra` (and similar) value. The data is merged into the dictified
schema in a non-recursive fashion. JSON schema does not define `extra` field,
so this is a local extension/hack.

## Closes

N/A

Partially fixes https://github.com/litestar-org/litestar/issues/2849 (adds support for _JSON schema_ examples defined per Pydantic class).


litestar/contrib/pydantic/pydantic_schema_plugin.py
```diff
@@ -309,16 +308,15 @@ def for_pydantic_model(cls, field_definition: FieldDefinition, schema_creator: S
...
-                if example is None
-                else get_formatted_examples(
-                    field_definition, [Example(description=f"Example {field_definition.name} value", value=example)]
-                )
-            ),
         )
+
+        if json_schema_extra:
+            for key, value in json_schema_extra.items():
+                setattr(schema, key, value)
```
tuukkamustonen, line 320: Actually, I think I need to re-work this. `json_schema_extra` should overwrite the dict-formatted schema, not Litestar's `Schema` object...

litestar/_openapi/schema_generation/schema.py
```diff
@@ -603,6 +604,7 @@ def create_component_schema(
             openapi_type: The OpenAPI type, defaults to ``OpenAPIType.OBJECT``.
             title: The schema title, generated if not provided.
             examples: A mapping of example names to ``Example`` instances, not required.
+            extra: Extra kwargs to add/overwrite in the generated schema.
```
JacobCoffee, line 607: Lets add a `.. versionadded::` directive here for the docs

litestar/openapi/spec/schema.py
```diff
@@ -643,6 +643,14 @@ class Schema(BaseSchemaObject):
     discouraged, and later versions of this specification may remove it.
     """
 
+    extra: Mapping[str, Any] | None = None
+    """Container for extensions to the generated JSON schema. This can be used to store local extensions/customizations,
+    e.g. `json_schema_extra` from Pydantic modes and fields.
+
+    The values are applied to the generated JSON schema after dumping it into mapping type. Values are merged over in a
+    non-recursive manner.
+    """
```
JacobCoffee, line 652: same

litestar/openapi/spec/schema.py
```diff
@@ -643,6 +643,14 @@ class Schema(BaseSchemaObject):
     discouraged, and later versions of this specification may remove it.
     """
 
+    extra: Mapping[str, Any] | None = None
+    """Container for extensions to the generated JSON schema. This can be used to store local extensions/customizations,
+    e.g. `json_schema_extra` from Pydantic modes and fields.
```
JacobCoffee, line 648:
```suggestion
    e.g. ``json_schema_extra`` from Pydantic modes and fields.
```

JacobCoffee (changes requested):

litestar/_openapi/schema_generation/schema.py
tuukkamustonen: Added, but I'm not sure of the exact syntax.

Also, aren't these internal APIs (in `litestar._openapi` package) so is that really needed here?

litestar/openapi/spec/schema.py
```diff
@@ -643,6 +643,14 @@ class Schema(BaseSchemaObject):
     discouraged, and later versions of this specification may remove it.
     """
 
+    extra: Mapping[str, Any] | None = None
```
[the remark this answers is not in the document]
tuukkamustonen: Polished a bit and added more explicit explanation.

tuukkamustonen:
@JacobCoffee @guacs Just reminding to re-review once you have time! 😃

litestar/_openapi/schema_generation/schema.py
JacobCoffee: sorry you're right.. not sure what I was thinking

peterschutt:
> This change also drops support for looking up an example from ConfigDict(example=...) because Pydantic doesn't define such a field:

But does pydantic throw an error if extra things are defined in `ModelConfig` or `ConfigDict` (I guess it would throw a static type error from `ConfigDict` as the key isn't defined, but not at runtime)? IDK if it was a deliberate decision to ingest the "example" name or not, but we _have_ been ingesting it and who knows if people have been relying on our use of it or not.

> The technical change is addition of Schema.extra, which can store the json_schema_extra (and similar) value.

Any examples for "and similar" here? Is there another use-case for this other than `json_schema_extra` right now?

> JSON schema does not define extra field, so this is a local extension/hack.

+1

> it's a bit ugly having to add this field to the Schema object, which is supposed to represent JSON schema spec and not hold extras like that.

+1

If these changes are only to support pydantic's `json_schema_extra` I think we should try to confine modifications to within the pydantic schema plugin if possible.

guacs:
@peterschutt I also had this concern regarding adding `extra` to `Schema`. As @tuukkamustonen mentioned, what should be done is to take the `json_schema_extra` and merge it into the schema object we create rather than via the `extra` field.

It's not as convenient, but I think we should go for that rather than the direction taken in this PR currently.

peterschutt:
> what should be done is to take the `json_schema_extra` and merge it into the schema object we create rather than via the `extra` field.

This is inside the plugin right? I.e., the schema object that we return from `PydanticSchemaPlugin.to_openapi_schema()`?

guacs:
> > what should be done is to take the `json_schema_extra` and merge it into the schema object we create rather than via the `extra` field.
> 
> This is inside the plugin right? I.e., the schema object that we return from `PydanticSchemaPlugin.to_openapi_schema()`?
> 
> 

Yes. It makes sense for it to be the plugin's responsibility to handle all of that.

tuukkamustonen:
@guacs @peterschutt @JacobCoffee Hey, sorry for taking so long here.

I'm looking into it again, and I don't see how converting the `json_schema_extra: dict` into `Schema` object could happen conveniently:

- Can't just use `setattr()` because many `Schema` object fields are also custom types (e.g. `Schema.any_of` and friends)
- Can't easily merge nested structures, because I'd have to first convert them to right type
- _Recursively_ merging (not sure how Pydantic does that, but I assume recursively) would require extra logic

Properly supporting it would add a lot of (unjustified) bloat. I don't think it makes sense. (All that to just get converted back to `dict` again.)

Do you have better suggestions? Maybe passing `extra` as separate argument, if it cannot live in `Schema`, or sub-classing `Schema` for this special case?

---

The scope of this topic is actually wider. Pydantic also supports:

1. [Defining `json_schema_extra` as a callable](https://docs.pydantic.dev/latest/concepts/json_schema/#using-json_schema_extra-with-a-callable)
2. [WithJsonSchema](https://docs.pydantic.dev/latest/concepts/json_schema/#using-json_schema_extra-with-a-callable)

I had an idea here, to only support the _callable_ format and do something like:

```py
    class MyBody(pydantic_v2.BaseModel):
        def _add_examples(self, schema):
            schema.examples = [{"foo": "hello"}]
        model_config = pydantic_v2.ConfigDict(json_schema_extra=_add_examples)
        foo: str
```

...but Pydantic defines `json_schema_extra` as:

```py
JsonSchemaExtraCallable: TypeAlias = Union[
    Callable[[JsonDict], None],
    Callable[[JsonDict, Type[Any]], None],
]

class ConfigDict(TypedDict):
    json_schema_extra: JsonDict | JsonSchemaExtraCallable | None
```

So we would be a bit breaking that API as in this case typing would be for `Callable[[Schema], None]` (and not `JsonDict`)... maybe not too bad, but still.

Finally, that `WithJsonSchema` is something Litestar doesn't support at all now (AFAIK).

I'm still wondering about the rationale to re-write the whole OpenAPI spec parsing when you could just call [BaseModel.model_json_schema](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_json_schema) and let Pydantic (and MsgSpec, et al) generate it. It would automatically support the bells and whistles seen here, plus any that they might add/alter in the future. And you'd save a lot of extra code, rather than trying to play the catch up game?

JacobCoffee:

peterschutt:
@tuukkamustonen - needs to be rebased on main and the PR reopened. I tried but I can't force push to your branch. Sorry about this.

tuukkamustonen:
@JacobCoffee @peterschutt Sure, I can do that (why though? Isn't 'develop' the target for new features anymore?)

Anyway, please check the writeup above in https://github.com/litestar-org/litestar/pull/3225#issuecomment-2036828214 - I'm not sure how to proceed here, at least not so that it would be convenient... any ideas?

JacobCoffee:
Re: the develop thing, you made a good point on the pain points of force pushing a lot. We are trying a new method of just developing everything against main (sans v3 branch) and using tags to cherry pick the desired things into patch or minor releases

peterschutt:
> Anyway, please check the writeup above in [#3225 (comment)](https://github.com/litestar-org/litestar/pull/3225#issuecomment-2036828214) - I'm not sure how to proceed here, at least not so that it would be convenient... any ideas?

Someone will devote time - just a matter of bandwidth. Copy that over to the new PR to make sure it gets eyes on it.

tuukkamustonen:
Okay, I'll open a new PR, with that comment ~copied over there.

Totally understand the lack of bandwidth. Thanks for your efforts!

PR #3236, opened 2024-03-22
Merge main into develop

JacobCoffee:
## Description

- when installing the package via git there seems to be some discrepancies causing runtime errors, this was made to help find out why


JacobCoffee:
ended up being https://github.com/cofin/litestar-granian/issues/9

PR #3255, opened 2024-03-26
feat: problem details plugin

### LINKED ISSUE #3199: Enhancement: Support for RFC 7807 / 9457
### Summary

A Boolean option to enable [RFC7807](https://datatracker.ietf.org/doc/html/rfc7807#section-6.1) support might be a good way to enable this otherwise breaking change.

ie: ` app = Litestar(rfc7807=True)`

Perhaps some of the common exceptions (like ValueError) could have:
- **type** URI enpoint / page provided by Litestar (plugin?)
- **title** field set automatically
- **status** field set automatically (use same value determination as current)
- **detail** field set automatically

A new Exception class perhaps `JsonProblem(HTTPException)` which auto-includes "application/problem+json" in header and has fields for: type, title, status, etc.

This would be nice to have and complement OpenAPI / Swagger support with the emerging standard for error responses.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3199">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3199/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

guacs:
## Description

- A plugin to enable usage of problem details as the response.

## Closes

Closes #3199.


guacs:
The way this works is by injecting an exception handler into the app level exception handlers to convert `ProblemDetailsException` into a response following the specification as per RFC 9457.

Users can pass in a mapping of exception types to callables that will convert those exception types into `ProblemDetailsException` for handling specific exceptions such as pydantic's `ValidationError`. That converted `ProblemDetailsException` will then be used to create the response. This should allow for flexibility when needed.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,111 @@
...
+        type_: str = "about:blank",
+        status_code: int | None = None,
+        title: str | None = None,
+        instance: str | None = None,
+        detail: str | dict[str, Any] | list[Any] | None = None,
+        extra: dict[str, Any] | None = None,
+        headers: dict[str, Any] | None = None,
+    ) -> None:
+        self.type_ = type_
+        self.status_code = status_code or self.status_code
```
skewty, line 50: I think what is happening here isn't immediately clear (it falls back to ClassVar status_code). Additionally there is no call to `super()`. I would personally prefer to see a call to super and only show the new instance level attributes in `__init__`. It would solve both of my concerns at the same time and likely improve code clarity IMO.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,111 @@
...
+
+@dataclass
+class ProblemDetailsConfig:
+    exception_handler: ProblemDetailsExceptionHandlerType = _problem_details_exception_handler
+    """The exception handler used for ``ProblemdetailsException.``"""
+
+    enable_for_all_http_exceptions: bool = False
+    """Flag indicating whether to convert all exceptions into ``ProblemDetailsException.``"""
+
+    exception_to_problem_detail_map: ExceptionToProblemDetailMapType | None = None
```
skewty, line 87: Is there any reason or value in this being `None`? I am thinking it is better to make the attribute non-optional and fallback to empty dict `{}`. It will also make static type checkers happier as it avoids the need for `None` checks before each read or write to it.

litestar/plugins/problem_details.py
bunny-therapist: I reacted to the same thing. If it inherits from HTTPException, it should call super init.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.instance is not None:
+            problem_details["instance"] = self.instance
+        if self.detail is not None:
+            problem_details["detail"] = self.detail
+        if self.extra is not None:
+            problem_details.update(self.extra)
+
+        content = encode_json(problem_details, request.route_handler.default_serializer)
+        headers = self.headers or {}
+        headers["content-type"] = self._PROBLEM_DETAIL_HEADER
```
bunny-therapist, line 70: Wouldn't it be enough to just have the media type of the underlying HTTPException be application/problem+json? I am a bit surprised to see this setting of content-type and encoding inside the new exception class, since there is already code for all of that.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return Response(content, headers=headers)
+
+    @classmethod
+    def from_http_exception(cls, exc: HTTPException) -> ProblemDetailsException:
+        return ProblemDetailsException(status_code=exc.status_code)
```
bunny-therapist, line 76: I don't quite see the point of converting HTTPException to ProblemDetailsException? I feel like what we should have is
1) ProblemDetailsException
2) Exception handler that catches ProblemDetailsException and makes it into an application/problem+json response
3) Exception handler that catches HTTPException and makes it into an application/problem+json response

1 and 2 work on their own, whereas 3 is a change to default behavior

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,111 @@
...
+
+        return problem_details_exc.to_response(req)
+
+    return _exception_handler
+
+
+class ProblemDetailsException(HTTPException):
+    _PROBLEM_DETAIL_HEADER = "application/problem+json"
+
+    def __init__(
```
skewty, line 39: I don't see a benefit in making these parameters `| None` here:

```python
title: str = '',
instance: str = '',
detail: str | dict[str, Any] | list[Any] = '',
```
Also notice that you changed the default value of `detail` from `''` to `None`. I don't see any benefit to it. Is there any?

The other parameters are best to follow what happens in super-class though so totally understand and would like to see them `| None`.

litestar/plugins/problem_details.py
bunny-therapist: Also, isn't there more that can be salvaged from an HTTPException than just the status code? Don't the standard litestar error responses already have a "detail" entry for example?

litestar/plugins/problem_details.py
skewty: This is part of the (current) RFC spec I believe and the main purpose of this plugin so the overwritten / forced content-type  makes sense to me.

Is there a better place to do the content-type assignment?

bunny-therapist: a `Response` with `media_type="application/problem+json"` will be automatically json-encoded and get the content-type header `"application/problem+json"`.

litestar/plugins/problem_details.py
guacs: Aah I didn't really put that much thought into this when I wrote this, but I agree it should be similar to the super class.

litestar/plugins/problem_details.py
guacs: It definitely should. Like I'd mentioned in the other comment, I hadn't put that much thought into this. I just wanted to get a basic PoC up and see whether the approach was fine. There's a bit of cleanup to do here and add tests as well.

litestar/plugins/problem_details.py
guacs: @bunny-therapist you're absolutely right. I forgot about that PR that added support for this. We should just be setting the `media_type` and allow the serialization to happen as per the normal flow.

litestar/plugins/problem_details.py
guacs: I'd kept it there in the beginning, but I would probably move that into a separate function rather than as a classmethod. There's no real benefit since users overriding it provide no benefit either.

guacs: > Also, isn't there more that can be salvaged from an HTTPException than just the status code? Don't the standard litestar error responses already have a "detail" entry for example?

Yes, there should be more that we can extract from the `HTTPException` such as `detail` and/or `extra`.

litestar/plugins/problem_details.py
guacs: Oh you're thinking of instantiating the config and then adding to the map? I'd just naively assumed that the usage would something like the following:

```python

config = ProblemDetailsConfig(exception_to_problem_detail_map=my_custom_dictionary_instance)
```

I can see that people may prefer to modify the attribute on the config object directly, so I'm fine with defaulting to an empty dictionary. I'd just kept it as `None` as the default to semantically indicate that the value is optional.

litestar/plugins/problem_details.py
guacs: > Edit: Is it just to enable the exception_to_problem_detail_map feature? Is that worth it though? It will only handle conversion from one type of exception (barring some aggressive use of isinstance) anyway, right?

I didn't quite understand. It's for cases where you want to handle a specific exception in a specific manner. For example, you might want to convert all `ValidationError` exceptions into a specific format whereas you might want to handle another exception differently. If this wasn't provided, then users have no control over the transformation of the exception into a problem detail. It's the same way litestar takes in the custom exception handlers, but instead of returning a `Response`, you return a `ProblemDetailException.`

Also, by providing this, users can override the handling of `HTTPException` as well by just providing a callable that does the transformation.

tests/unit/test_cli/test_core_commands.py
```diff
@@ -410,6 +410,115 @@ def test_run_command_arguments_precedence(
...
+            ("LITESTAR_DEBUG", "1"),
+        ),
+        (
+            ("--pdb", True),
+            ("LITESTAR_PDB", False),
+            ("LITESTAR_PDB", "1"),
+        ),
+    ),
+)
+def test_run_command_arguments_precedence(
```
bunny-therapist, line 463: Is this supposed to appear here? It does not appear to be part of the PR, rather the result of some rebasing?

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,120 @@
...
+
+@dataclass
+class ProblemDetailsConfig:
+    """The configuration object for ``ProblemDetailsPlugin.``"""
+
+    exception_handler: ProblemDetailsExceptionHandlerType = _problem_details_exception_handler
+    """The exception handler used for ``ProblemdetailsException.``"""
+
+    enable_for_all_http_exceptions: bool = False
+    """Flag indicating whether to convert all exceptions into ``ProblemDetailsException.``"""
```
bunny-therapist, line 93: All exceptions or all HTTPExceptions?

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,120 @@
...
+
+    exception_to_problem_detail_map: ExceptionToProblemDetailMapType = field(default_factory=dict)
+    """A mapping to convert exceptions into ``ProblemDetailsException.``
+
+    All exceptions provided in this will get a custom exception handler where these exceptions
+    are converted into ``ProblemDetailException`` before handling them.
+    """
```
bunny-therapist, line 100: Adding HTTPException to the map currently takes precedence over `enable_for_all_http_exceptions` - should that be documented?

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,120 @@
...
+        self.title = title
+        self.instance = instance
+
+    def to_response(self, request: Request) -> Response:
+        """Convert the problem details exception into a ``Response.``"""
+
+        problem_details = {"type": self.type_, "status": self.status_code}
+        for attr in ("title", "instance", "detail", "extra"):
+            if (value := getattr(self, attr, None)) is not None:
+                problem_details[attr] = value
```
bunny-therapist, line 80: Should `"extra"` maybe be added with one level removed? I.e., instead of adding "extra" as a key in ProblemDetails dict, add the keys of "extra" to ProblemDetails dict directly?

I am just thinking of what the problem details in the rfc looked like. They suggested adding new keys as needed to the problem details object. We are forcing all of those things to be wrapped by "extra", which I think is mostly a convention from the regular litestar error responses, right?

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _http_exception_to_problem_detail_exception(exc: HTTPException) -> ProblemDetailsException:
+    return ProblemDetailsException(status_code=exc.status_code, detail=exc.detail, extra=exc.extra, headers=exc.headers)
+
+
+class ProblemDetailsException(HTTPException):
+    """A problem details exception as per RFC 9457."""
+
+    _PROBLEM_DETAIL_MEDIA_TYPE = "application/problem+json"
```
bunny-therapist, line 42: Inconsistent: ProblemDetails vs PROBLEM_DETAIL, suggest changing to _PROBLEM_DETAILS_MEDIA_TYPE

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def to_response(self, request: Request) -> Response:
+        """Convert the problem details exception into a ``Response.``"""
+
+        problem_details = {"type": self.type_, "status": self.status_code}
+        for attr in ("title", "instance", "detail", "extra"):
+            if (value := getattr(self, attr, None)) is not None:
+                problem_details[attr] = value
+
+        return Response(problem_details, headers=self.headers, media_type=self._PROBLEM_DETAIL_MEDIA_TYPE)
```
bunny-therapist, line 82: This Response should get the status code.

JacobCoffee:

litestar/plugins/problem_details.py
guacs: I'm commenting here, even though the PR is now at #3323. I think Litestar by default should only handle `HTTPExceptions`. If the user wants to handle `Exception` in general, then they can add an conversion callable.

litestar/plugins/problem_details.py
guacs: Oh good point. Yeah, I did that intentionally so that they could override the `HTTPException` conversion into problem details if needed.

litestar/plugins/problem_details.py
guacs: I did think of this initially, but that would only be possible if `extra` is a `Mapping` which it doesn't have to be. It could also be a list. Maybe we should consider checking for what type `extra` is and change the response accordingly?

litestar/plugins/problem_details.py
guacs: Good catch!

guacs:
@bunny-therapist if you'd like to review it again at #3323, that'd be great! I can't tag you there unfortunately.

litestar/plugins/problem_details.py
bunny-therapist: I understand, but the docstring says "all exceptions"

litestar/plugins/problem_details.py
bunny-therapist: I would be fine with that

litestar/plugins/problem_details.py
bunny-therapist: Is it possible to add/expand a test to catch this?

bunny-therapist:
Thanks. I will have a look tomorrow.

litestar/plugins/problem_details.py
guacs: Oh I totally missed that! Thanks!

litestar/plugins/problem_details.py
guacs: I've modified the tests to check for the status on the responses as well.

PR #3212, opened 2024-03-16
docs: update links, formatting, grammar, and remove contractions

JacobCoffee:
## Description

- Link more things
  - Adds `sphinx-paramlinks` to help!
  - Usage of `term` and external intersphinx links more
- Correct more spelling
- Moar consistency of what is `wrapped` vs. not
- No contractions
- Shorten line lengths
- Cleanup of the verbose calls e.g., `MyClass <litestar.path.to.MyClass>` vs `~path.to.MyClass`
- Probably some more shit I didn't think about

## Closes

- Closes #3260, 
- Fixes #3259

PR #3256, opened 2024-03-26
docs(csrf): Add usage example

Alc-Alc:
## Description
- Adds client usage example for CSRF middleware.
- Slightly modifies the existing server example to make it more copy pastable. 
- References to external links for some related concepts.

Doc link: https://litestar-org.github.io/litestar-docs-preview/3256/usage/middleware/builtin-middleware.html#csrf

### DIFF
--- a/docs/usage/middleware/builtin-middleware.rst
+++ b/docs/usage/middleware/builtin-middleware.rst
@@ -76,7 +76,6 @@ The following snippet demonstrates how to change the cookie name to "some-cookie
 
 
 A CSRF protected route can be accessed by any client that can make a request with either the header or form-data key.
-Both of which can be configured in :class:`CSRFConfig <.config.csrf.CSRFConfig>` as show above.
 
 
 .. note::
--- a/docs/usage/middleware/builtin-middleware.rst
+++ b/docs/usage/middleware/builtin-middleware.rst
@@ -105,6 +105,11 @@ The following is an example using ``httpx.Client``.
         post_response_using_form_data = client.post("http://localhost:8000/1", data={"_csrf_token": csrf})
         assert post_response_using_form_data.status_code == 201
 
+        # despite the header being passed, this request will fail as it does not have a cookie in its session
+        # note the usage of ``httpx.post`` instead of ``client.post``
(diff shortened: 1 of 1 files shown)


docs/usage/middleware/builtin-middleware.rst
```diff
@@ -42,29 +42,74 @@ This middleware prevents CSRF attacks by doing the following:
...
+.. code-block:: python
 
+    import httpx
+
+
+    with httpx.Client() as client:
+        get_response = client.get("http://localhost:8000/")
+
+        # "csrftoken" is the default cookie name
+        csrf = get_response.cookies["csrftoken"]
```
euri10, line 98: would it makes sense to add a post with a wrong header / cookie just to show the protection ?

Alc-Alc: would this be ok?
```py
    # the following is a request that will fail because it does not have a cookie in its session
    post_response_with_no_persisted_cookie = httpx.post("http://localhost:8000/1", headers={"x-csrftoken": csrf})
    assert post_response_with_no_persisted_cookie.status_code == 403
    assert "CSRF token verification failed" in post_response_with_no_persisted_cookie.text
    ```

euri10: yep, it shows well the protection, now I'm not sure this is not too detailed but imho it makes sense to show it working.

Alc-Alc: I understand but I am unsure how else to approach this tbh, I feel if I make it more detailed it would just parrot the tests we already have, one can argue normal users won't look at tests, so I am open for improvements here 🙂

euri10 (approved):

docs/usage/middleware/builtin-middleware.rst
guacs: > it makes sense to show it working

Would it? As a user I would assume that, by default, sending the wrong cookie/header wouldn't work.

guacs (approved):

JacobCoffee:
automerge screws me yet again

JacobCoffee:
I can just make these changes in https://github.com/litestar-org/litestar/pull/3212 sorry to spam

PR #3258, opened 2024-03-26
docs: Update test_client fixture documentation

sherbang:
## Description

- Make it clearer that the test_client needs to be used as a contextmanager.

I didn't follow this example closely enough, and forgot the `with` statement in my tests.  I only realized it was already there after I made all of these changes and was reviewing everything to make sure it was all complete.  Since it's currently technically correct, these changes aren't as necessary as I thought, but I still think this is clearer than what's currently here.

This version also makes it easier to use since each test doesn't have to have a separate `with` block.

While I was at it, I fixed the "partially unknown" typing.

### DIFF
--- a/docs/examples/testing/test_health_check_async.py
+++ b/docs/examples/testing/test_health_check_async.py
@@ -1,3 +1,5 @@
+from typing import AsyncIterator
+
 import pytest
 
 from litestar import Litestar, MediaType, get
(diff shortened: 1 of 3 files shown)


peterschutt (approved):
Thanks for this!

I agree with the pattern.

I wish there was a nicer way to annotate these fixtures, e.g., `AsyncIterator[AsyncTestClient[Litestar]]` is pretty tough on the eye, and its also a pain to have to use `AsyncTestClient[Litestar]]` as an annotation right throughout your test suite wherever the client is injected. I personally create a type alias for the client and use that, e.g.,:

```py
LitestarClient: TypeAlias = AsyncTestClient[Litestar]

@pytest.fixture()
async def test_client() -> AsyncIterator[LitestarClient]: ...

async def test_blah(test_client: LitestarClient) -> None: ...
```
IDK if its worth introducing a pattern like that here, or not.. @litestar-org/maintainers any opinion on this?

Alc-Alc:
@all-contributors add @sherbang for docs

PR #3263, opened 2024-03-27
feat(v3): remove deprecated checks and calls

JacobCoffee:
## Description

- Initial work of removing the deprecated things in v3


JacobCoffee:
Think i might need help on the new openapi stuff, i wish i had the docs rendered :P

JacobCoffee:
#3264 should give us the rendered docs

litestar/response/base.py
```diff
@@ -88,11 +87,6 @@ def __init__(
         status_code = status_code or HTTP_200_OK
         self.headers = MutableScopeHeaders()
 
-        if encoded_headers is not None:
```
provinzkraut, line 91: I think all of these changes ought to remove the deprecated parameter as well, not just the deprecation warning :P

litestar/static_files/config.py
provinzkraut, line 1: I think this should be a separate PR because it means we can remove all the special casing for the static files app in our routing logic as well

JacobCoffee:
make it smaller :)

PR #3273, opened 2024-03-27
fix: address `__slots__` and `__all__` misconfigurations

### LINKED ISSUE #3251: Bug: slotscheck pre-commit misconfiguration
### Description

👋 Hi there @Goldziher. I'm happy to see you're still using slotscheck, but I noticed that the pre-commit `exclude` option is configured so that it isn't run on most files:

```
$ pre-commit run slotscheck --all-files -v
All OK!
stats:
  modules:     8
    checked:   8
    excluded:  0
    skipped:   5

  classes:     1
    has slots: 1
    no slots:  0
    n/a:       0
```

8 modules scanned is way too few, so I had a look.  It took me a while to realize that the pre-commit option `exclude: "test_*|docs|.github"`  regex matches `test` and `litestar` contains the string `test` 😬.

It also looks like the `flake8-dunder-all` pre-commit would have the same issue 👀. Might be worth scanning any other tooling you have set up to ignore things with "test" in the name 😄 

I'll have a look now and, if the change is small, submit a PR shortly.

_No response_

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

main

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3251">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3251/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

ariebovenberg:
Fixes #3251 

Remarks:
- For the 'ensure dunder all' check, I was unsure what the exact intention is. Running it resulted in many `__all__` definitions, which I left unchanged.
- To properly run slotscheck it needs to import the code, which requires all dependencies. Keeping them up-to-date will probably be a chore. The alternative is to use `language: system`, but this negates the benefit of pre-commit environment isolation.
- I wasn't able to get `__slots__` working on `AbstractMiddleware` without making breaking changes. This is due to the fact that the slot names are used both as instance and class variables. To prevent confusion, I've removed the (ineffective) slots from its subclasses
- A python bug (github.com/python/cpython/pull/106771) renders slots on event emitters ineffective. Since this will roll out in Python 3.13, I've kept the slots


.pre-commit-config.yaml
```diff
@@ -39,13 +39,55 @@ repos:
...
+          - prometheus-client
+          - pydantic
+          - email-validator
+          - pydantic-extra-types
+          - redis[hiredis]>=4.4.4
+          - advanced-alchemy>=0.2.2,<1.0.0
+          - structlog
+          - asyncpg
+          - psycopg
+          - docutils
```
peterschutt, line 90: We recently moved mypy and pyright out of pre-commit partially because of this duplicated environment management - can slotscheck be run as a standalone tool outside of pre-commit?

litestar/channels/subscriber.py
```diff
@@ -6,6 +6,12 @@
 from contextlib import AsyncExitStack, asynccontextmanager, suppress
 from typing import TYPE_CHECKING, Any, AsyncGenerator, Awaitable, Callable, Generic, Literal, TypeVar
 
+__all__ = (
+    "AsyncDeque",
+    "Subscriber",
+)
+
+
```
peterschutt, line 14: Personally, not a fan that `flake8-dunder-all` injects this in between runtime imports and type-checking imports.

I'm also not sure whether blindly adding any non-underscore names into an `__all__` is desirable. For instance in this case, I don't believe `AsyncDeque` is intended to be part of the modules public interface.

peterschutt (commented):
Thanks very much for identifying this issue @ariebovenberg, and for submitting the patch!

I think slotscheck is an important tool for us, and by the looks of this we had a couple of configuration issues (i.e., the exclude regex and improper environment setup). Is it possible for the tool to be run outside of pre-commit? We recently removed mypy and pyright from pre-commit due to having to maintain multiple lists of dependencies which would frequently be forgotten about as the library dependencies changed.

I have particular concern with `flake8-dunder-all`, firstly the way it injects the `__all__` in between the runtime and typing imports, and secondly I feel that the names added to `__all__` needs a bit more consideration than writing any non-underscore name in there, especially in our public namespaces.

@litestar-org/maintainers any thoughts on `flake8-dunder-all`?

.pre-commit-config.yaml
peterschutt: nvm I've just been schooled on the history of slotscheck/pre-commit and litestar (starlite) :grimacing:

litestar/stores/redis.py
```diff
@@ -21,7 +21,7 @@
 class RedisStore(NamespacedStore):
     """Redis based, thread and process safe asynchronous key/value store."""
 
-    __slots__ = ("_redis",)
+    __slots__ = ("_redis", "handle_client_shutdown", "_get_and_renew_script", "_delete_all_script")
```
JacobCoffee, line 24: probably a dumb question but why do we expose private methods here

litestar/utils/warnings.py
```diff
@@ -4,6 +4,13 @@
 from litestar.exceptions import LitestarWarning
 from litestar.types import AnyCallable, AnyGenerator
 
+__all__ = (
+    "warn_implicit_sync_to_thread",
+    "warn_pdb_on_exception",
+    "warn_sync_to_thread_with_async_callable",
+    "warn_sync_to_thread_with_generator",
+)
```
JacobCoffee, line 12: idk if we want these exposed in documentation

.pre-commit-config.yaml
JacobCoffee: i did the same thing 🙃

litestar/channels/subscriber.py
JacobCoffee: i can think of a few instances we havent privatized a thing and want it undocumented, but im more tilted about how it doesnt place them under `TYPE_CHECKING` imports as well :D

.pre-commit-config.yaml
ariebovenberg: As the author of slotscheck, I'd actually _recommend_ running slotscheck as standalone—for exactly the reasons we're encountering here 👍

litestar/stores/redis.py
ariebovenberg: note that `__slots__` isn't about exposing things like `__all__` is. `__slots__` merely tells Python to only allow certain attributes on the class in order to optimize memory. Private attributes are no exception.

litestar/utils/warnings.py
ariebovenberg: I decided to leave the flake8 dunder additions unmodified so those more experienced with the library can decide what to exclude 👍

peterschutt:
I made a PR to see if we can get f8-d-a to play nice with the type check imports - don't really know what I'm doing though, so lets see what happens: https://github.com/python-formate/flake8-dunder-all/pull/65

litestar/stores/redis.py
JacobCoffee: ahh yep.. i thought this was __all__ i think i got lost in the buzz.

peterschutt:
> I made a PR to see if we can get f8-d-a to play nice with the type check imports - don't really know what I'm doing though, so lets see what happens: [python-formate/flake8-dunder-all#65](https://github.com/python-formate/flake8-dunder-all/pull/65)

They released f8-d-a 0.4 which has support for an `if TYPE_CHECKING` block however there's an issue with how it handles `if TYPE_CHECKING` blocks that have stuff other than simple imports in it which I've opened an issue for. Either way this PR will be superseded, or we change the way we use the tool, or stop using it altogether.

WRT changing how we use the tool, it might be better to use it as a lint without the auto-fix, so it forces us to have the `__all__` defined, but not populate it for us.

PR #3279, opened 2024-03-28
fix: openapi generation with pydantic examples

### LINKED ISSUE #3277: Bug: schema generation crashes in 2.7.1
### Description

Using 2.7.0, I code view the "/schema" page of my app without problem, but with 2.7.1, I get internal server error 500 back. I ran it with debug and got this:

`'str' object has no attribute 'value'`

and a trace pointing to this line: https://github.com/litestar-org/litestar/blob/main/litestar/_openapi/schema_generation/utils.py#L114
as called from [litestar/_openapi/schema_generation/schema.py in SchemaCreator.process_schema_result at line 561](https://github.com/litestar-org/litestar/blob/main/litestar/_openapi/schema_generation/schema.py#L561)

I believe the error comes from this PR: https://github.com/litestar-org/litestar/pull/3224/files

The problem is triggered by _pydantic models with examples_ in the `json_schema_extra`. See MCVE. The examples being strings trigger this problem, but I am sure that other types would trigger similar errors.

_No response_

### MCVE

```python

import litestar
import pydantic  # v2


class Foo(pydantic.BaseModel):
    bar: str = pydantic.Field(
        json_schema_extra={
            "examples": ["this-is-a-string"]
        }
    )


@litestar.get()
async def get_foo() -> Foo:
    ...


with create_test_client([get_foo]) as client:
    resp = client.get("/schema/openapi.json")
    print(resp.status_code)  # 500

```

### Steps to reproduce

Run the above MCVE. It prints 500. Remove the examples from `json_schema_extra` and it prints 200.

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.7.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3277">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3277/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

bunny-therapist:
## Description

Resolves an issue with OpenAPI schema generation when using ``Pydantic`` examples

## Closes
- Closes #3277

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -67,7 +67,7 @@ from litestar.utils.typing import (
 
 if TYPE_CHECKING:
     from litestar._openapi.datastructures import OpenAPIContext
-    from litestar.openapi.spec import Example, Reference
+    from litestar.openapi.spec import Reference
     from litestar.plugins import OpenAPISchemaPluginProtocol
 
 KWARG_DEFINITION_ATTRIBUTE_TO_OPENAPI_PROPERTY_MAP: dict[str, str] = {
@@ -558,7 +558,7 @@ class SchemaCreator:
                     not isinstance(value, Hashable) or not self.is_undefined(value)
                 ):
                     if schema_key == "examples":
-                        value = get_json_schema_formatted_examples(cast("list[Example]", value))
+                        value = get_json_schema_formatted_examples(cast("list[Any]", value))
 
                     # we only want to transfer values from the `KwargDefinition` to `Schema` if the schema object
                     # doesn't already have a value for that property. For example, if a field is a constrained date,
--- a/litestar/_openapi/schema_generation/utils.py
+++ b/litestar/_openapi/schema_generation/utils.py
@@ -3,12 +3,12 @@ from __future__ import annotations
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Mapping, _GenericAlias  # type: ignore[attr-defined]
 
+from litestar.openapi.spec import Example
 from litestar.utils.helpers import get_name
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
-    from litestar.openapi.spec import Example
     from litestar.typing import FieldDefinition
 
 __all__ = (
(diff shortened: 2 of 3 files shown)


bunny-therapist:
I am not entirely sure about treating the examples as `Example` objects or just include them raw, but then again, this is arguably more strict than casting something to `list[Example]` that apparently was sometimes not that all, but e.g. `list[str]`. I don't quite understand the (apparently faulty) logic of the casting; I just made this work by not crashing if the elements are not `Example` objects (which is probably better than crashing).

peterschutt:
The issue is really in: 

https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/typing.py#L81

... where the schema properties such as examples, are unpacked into an instance of [`KwargDefinition`](https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/params.py#L28).

Here, the `json_schema_extra` is unpacked into a dict: https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/typing.py#L92-L95

Then the list of `Example` object is constructed: https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/typing.py#L96-L102

However, `extra` is only checked to see if it contains `"example"`, not `"examples"`, so `example_list` is `None` and `"examples"` remains in the `extra` dict - and at this point examples could contain _anything_.

Then, that `extra` dict is unpacked into the dict returned from that function: https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/typing.py#L127

Eventually the dict returned which now contains `"examples": ["example"]` is used to instantiate the `KwargDefinition` here: https://github.com/litestar-org/litestar/blob/d2cb891fcc5264620674aa0010f0a0aab682aa85/litestar/typing.py#L176

And that is the problem, `KwargDefinition.examples` is typed `list[Example]` - not `list[Any]` but we are instantiating it with whatever `json_schema_extra["examples"]` is set to because we don't pop them from that `extra` dict as pointed out above.

We actually spoke about this in #3224 (that is carrying around jsonschema examples in openapi spec example objects. Not ideal but its a legacy thing). 

This is why that `list[Example]` typing exists that you question above.

This metadata handling is an absolute labyrinth so no judgement about the fix you applied in this PR, but I think resolving this per above is preferable.

peterschutt:
Superseded by #3281 - copied your test over with attribution and applied the fix @bunny-therapist - cheers!

PR #3280, opened 2024-03-29
fix(openapi): set default on schema from `FieldDefinition`

### LINKED ISSUE #3278: Bug: default not generated for query parameter in openapi spec.
### Description

Hi,

The default of a query parameter is not properly generated when creating the openapi specification. See MVCE down below.

The generated specification looks like this:

```json
...
"paths": {
    "/query_default": {
      "get": {
        "summary": "QueryDefault",
        "operationId": "QueryDefaultQueryDefault",
        "parameters": [
          {
            "name": "foo",
            "in": "query",
            "schema": {
              "type": "integer"
            },
            "required": false,
            "deprecated": false,
            "allowEmptyValue": false,
            "allowReserved": false
          }
        ],
...
```
where the default is missing in the schema property.

_No response_

### MCVE

```python
@get(path="query_default")
def query_default(foo: int = 12) -> None:
    return

app = Litestar(route_handlers=[query_default])
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

Version 2.7.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3278">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3278/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

guacs:
## Description

Consider the following:

```python
    def get_foo(foo_id: int = 10) -> None:
        ...
```
In such cases, no `KwargDefinition` is created since there is no metadata provided via `Annotated`. The default is still parsed, and set on the generated `FieldDefinition`, however the `SchemaCreator` currently only considers defaults that are set on `KwargDefinition`.

So in such cases, we should fallback to the default set on the `FieldDefinition` if there is a valid default value.

## Closes
Fixes #3278.

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -569,6 +569,9 @@ class SchemaCreator:
                     if getattr(schema, schema_key, None) is None:
                         setattr(schema, schema_key, value)
 
+        if schema.default is None and field.default is not Empty:
+            schema.default = field.default
+
         if not schema.examples and self.generate_examples:
             from litestar._openapi.schema_generation.examples import create_examples_for_field
 
--- a/tests/unit/test_openapi/test_schema.py
+++ b/tests/unit/test_openapi/test_schema.py
@@ -37,7 +37,7 @@ from litestar.openapi.spec import ExternalDocumentation, OpenAPIType, Reference
 from litestar.openapi.spec.example import Example
 from litestar.openapi.spec.schema import Schema
 from litestar.pagination import ClassicPagination, CursorPagination, OffsetPagination
-from litestar.params import Parameter, ParameterKwarg
+from litestar.params import KwargDefinition, Parameter, ParameterKwarg
 from litestar.testing import create_test_client
 from litestar.types.builtin_types import NoneType
 from litestar.typing import FieldDefinition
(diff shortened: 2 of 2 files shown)


provinzkraut (approved):

guacs:
This does raise an interesting condition that we don't currently handle.

```python

def get_foo(foo: Annotated[int, Parameter(default=20)] = 12) -> None:
    ....
```

In this case, the default in the generated schema would be `20`. If there's such a mismatch between the values for default in `FieldDefinition` and it's associated `KwargDefinition`, then an error should be raised during startup itself. I didn't include it as part of this PR because I think that's best left as another PR, but also because I wasn't sure if that would be considered a breaking change.

peterschutt (approved):

provinzkraut:
> This does raise an interesting condition that we don't currently handle.
> 
> ```python
> def get_foo(foo: Annotated[int, Parameter(default=20)] = 12) -> None:
>     ....
> ```
> 
> In this case, the default in the generated schema would be `20`. If there's such a mismatch between the values for default in `FieldDefinition` and it's associated `KwargDefinition`, then an error should be raised during startup itself. I didn't include it as part of this PR because I think that's best left as another PR, but also because I wasn't sure if that would be considered a breaking change.

Good catch! Yeah, that should be an error on startup. I'd go even further as to now allow specifying defaults in two places to begin with. A default in `Parameter` should *only* be allowed if it's being used with the old `param: int = Parameter(default=1)` style, but not `param: Annotated[int, Parameter(default=1)]`.

PR #3281, opened 2024-03-29
fix: pydantic `json_schema_extra` examples.

### LINKED ISSUE #3277: Bug: schema generation crashes in 2.7.1

peterschutt:
Fixes a regression introduced in 2.7 where an example for a field provided in pydantic's `Field.json_schema_extra` would cause an error.

Closes #3277

## Description

-

### DIFF
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -96,7 +96,7 @@ def _parse_metadata(value: Any, is_sequence_container: bool, extra: dict[str, An
     example_list: list[Any] | None
     if example := extra.pop("example", None):
         example_list = [Example(value=example)]
-    elif examples := getattr(value, "examples", None):
+    elif examples := (extra.pop("examples", None) or getattr(value, "examples", None)):
         example_list = [Example(value=example) for example in cast("list[str]", examples)]
     else:
         example_list = None
--- a/tests/unit/test_contrib/test_pydantic/test_openapi.py
+++ b/tests/unit/test_contrib/test_pydantic/test_openapi.py
@@ -553,6 +553,26 @@ def test_create_schema_for_field_v2() -> None:
     assert value.examples == ["example"]
 
 
+def test_create_schema_for_field_v2__examples() -> None:
+    class Model(pydantic_v2.BaseModel):
+        value: str = pydantic_v2.Field(
+            title="title", description="description", max_length=16, json_schema_extra={"examples": ["example"]}
+        )
+
+    schema = get_schema_for_field_definition(
+        FieldDefinition.from_kwarg(name="Model", annotation=Model), plugins=[PydanticSchemaPlugin()]
+    )
+
+    assert schema.properties
+
+    value = schema.properties["value"]
+
+    assert isinstance(value, Schema)
+    assert value.description == "description"
+    assert value.title == "title"
+    assert value.examples == ["example"]
+
+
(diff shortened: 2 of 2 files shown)


litestar/typing.py
```diff
@@ -96,7 +96,7 @@ def _parse_metadata(value: Any, is_sequence_container: bool, extra: dict[str, An
     example_list: list[Any] | None
     if example := extra.pop("example", None):
         example_list = [Example(value=example)]
-    elif examples := getattr(value, "examples", None):
+    elif examples := (extra.pop("examples", None) or getattr(value, "examples", None)):
```
peterschutt, line 99: Gives examples in json_schema_extra precedence over `Field(examples=...)`.

bunny-therapist: Nice! I understood that my fix was a kind of desperate thing in the wrong place (because the typing in the surrounding code was already incorrect) but I would have never found this.

For the sake of making sense of the types, may I ask why there is a cast to `list[str]` in the next line? I understand that we are now converting the raw examples to `list[Example]`, but I don't understand the assumption (built into the cast) that all examples are `str`? Couldn't they be anything?
```
example_list = [Example(value=example) for example in cast("list[str]", examples)]
```

peterschutt: I can't really shed any light on why that is - but it seems unnecessary and arbitrary. `Example.value` is typed `Any` so the `cast()` to `str` for the values seems redundant, and `examples` in that context would be `Any` also due to `dict.pop()` and `getattr()` to assign its value, so the cast to list wouldn't be necessary either I don't think...

PR #3287, opened 2024-03-29
docs(channels): Fix subsriber examples

provinzkraut:
Fix outdated usage of the `channels.subscribe` method; The `channel.start_subscription` method should be used instead.


provinzkraut:
Probably. This would also have been caught by mypy, which we stopped using a while ago because the error suppression needed polluted the examples quite a bit.

Now I'm thinking though that it would be relatively straightforward to strip the `# type: ignore` comments from the examples. Maybe we could make sure the examples are nicely typed and then implement the stripping on top to get rid of the remaining things.

cofin (approved):

PR #3271, opened 2024-03-27
refactor: removes openapi controller and related

peterschutt:
## Description

This PR removes all deprecated elements of OpenAPIConfig and the OpenAPIController, removes any obsolete tests and refactors tests that were parametrized to test both OpenAPIController and the router-based approach.

This also sets Scalar as our default plugin.

PR #3297, opened 2024-03-31
refactor: removes deprecated OpenAPIController

peterschutt:
## Description

This PR removes all deprecated elements of OpenAPIConfig and the OpenAPIController, removes any obsolete tests and refactors tests that were parametrized to test both OpenAPIController and the router-based approach.

Supersedes #3271


docs/usage/openapi/ui_plugins.rst
```diff
@@ -139,57 +139,6 @@ In the following example, we configure the OpenAPI root path to be ``/docs``:
 This will result in any of the OpenAPI endpoints being served at ``/docs`` instead of ``/schema``, e.g.,
 ``/docs/openapi.json``.
 
-Backward Compatibility
```
provinzkraut, line 142: I think we should keep this around and add it to a "Migrating to 3.0" doc. Every breaking change should be documented there, and ideally, we do it as we go, so we don't have to do it all at once like last time with 2.0 :eyes:

provinzkraut (changes requested):

docs/release-notes/whats-new-3.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Imports
+-------
+
++----------------------------------------------------+------------------------------------------------------------------------+
+| ``2.x``                                            | ``3.x``                                                                |
++====================================================+========================================================================+
+| **OpenAPI**                                                                                                                 |
++----------------------------------------------------+------------------------------------------------------------------------+
+| ``litestar.openapi.OpenAPIController``             | :mod:`.openapi.plugins`                                                |
```
peterschutt, line 20: Not sure if this is really an appropriate import redirection, but I mainly just wanted to keep the structure of the table in place so its there to fill out for other things.

docs/conf.py
```diff
@@ -244,7 +244,7 @@
 ]
 
 auto_pytabs_min_version = (3, 8)
-auto_pytabs_max_version = (3, 11)
+auto_pytabs_max_version = (3, 12)
```
peterschutt, line 247: I assume we want 3.12 here?

docs/release-notes/index.rst
```diff
@@ -6,6 +6,8 @@ Release notes
 .. toctree::
     :titlesonly:
 
+    whats-new-3
     whats-new-2
+    3.x Changelog <https://litestar.dev/3-dev/release-notes/changelog.html>
```
peterschutt, line 11: Will we be starting a new changelog document for the 3.x series, or just continue on with the existing changelog and add the 3.x stuff at the top? At the moment, this points to the 2.x changelog document?

docs/conf.py
```diff
@@ -302,7 +302,9 @@
             },
         },
         "Release notes": {
+            "What's new in 3.0": "release-notes/whats-new-3",
             "What's new in 2.0": "release-notes/whats-new-2",
+            "3.x Changelog": "https://docs.litestar.dev/3-dev/release-notes/changelog.html",
```
peterschutt, line 307: Same here, will this be a totally new doc?

docs/release-notes/index.rst
JacobCoffee: New one for 3.0 preferably

peterschutt: And does the 2.x changelog still exist in the 3.0 branch, or do I just delete everything from the existing one and change the document title?

docs/release-notes/whats-new-3.rst
JacobCoffee, line 20: not sure if this is a good idea or not but what if we add, in v3 branch only, an intersphinx link to Litestar v2 docs so was can directly link things like `litestar.openapi.OpenAPIController`? Idk if it would even work.. just a thought

PR #3299, opened 2024-04-01
feat(v3)!: remove abstract repository

cofin:
## Description

This PR removes the abstract repository from Litestar.  This is not really being used and has it's own implementation in advanced alchemy.


cofin:
i'll revisit this as we get a close to 3.0

PR #3307, opened 2024-04-03
refactor: remove CacheControlHeader dependency on AbstractDTO

peterschutt:
`CacheControlHeader` is using `AbstractDTO` to access the inner type of an optional union in order to get at the instantiable type within.

This PR does that without needing an import from `litestar.dto` and helps to fix a circular import issue I've bumped into elsewhere.

## Description

-

### DIFF
--- a/litestar/datastructures/headers.py
+++ b/litestar/datastructures/headers.py
@@ -2,9 +2,10 @@ import re
 from abc import ABC, abstractmethod
 from contextlib import suppress
 from copy import copy
-from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


cofin (approved):

peterschutt:
Would you mind taking another look @cofin - the implementation has materially changed since your last review. I've ended up using msgspec to convert from the kwargs parsed out of the header which means we don't have to roll our own type coercion solution.

provinzkraut (approved):
That is much nicer in general :)

PR #3308, opened 2024-04-03
refactor: metadata parsing

peterschutt:
As spoken about in discord recently, the objective of the `litestar.typing` module was to have a simple interface to represent types. However, a lot of logic to do with schema constraint parsing has made its way in there, which has been the source of many bugs and hard to follow logic. This bit me again today as the root of a circular import problem.

This PR attempts to move metadata parsing to where it is needed, and where it has the context of the library/framework from where the metadata originates.

WIP - not sure if feasible, but lets see.

## Description

-


provinzkraut:
Superseded by #3721

PR #3314, opened 2024-04-04
feat: Add `path` parameter to Litestar application class

kedod:
## Description

- Expose `path` parameter at `Litestar` application class level

At this moment if we want use prefixed path for all routes in the application we have to create 
dedicated router. E.g:
```python


@get("/")
async def index() -> str:
    return "Hello, world!"


router = Router(path="/path", route_handlers=[index])


app = Litestar(route_handlers=[router])
```

This PR allows to set `path` on the `Litestar` application level.
Thanks to this we don't need additional `Router` class.

```python


@get("/")
async def index() -> str:
    return "Hello, world!"


app = Litestar(path="/path", route_handlers=[index])
```


cofin (approved):
LGTM

litestar/app.py
```diff
@@ -277,6 +278,8 @@ def __init__(
                 :class:`ASGI Scope <.types.Scope>`.
             parameters: A mapping of :class:`Parameter <.params.Parameter>` definitions available to all application
                 paths.
+            path: A path fragment that is prefixed to all route handlers, controllers and routers associated
+                with the application instance.
```
JacobCoffee, line 282:
```suggestion
            path: A path fragment that is prefixed to all route handlers, controllers and routers associated
                with the application instance.
                
                .. versionadded:: 2.8.0
```

litestar/testing/helpers.py
```diff
@@ -201,6 +202,8 @@ def test_my_handler() -> None:
             :class:`ASGI Scope <.types.Scope>`.
         parameters: A mapping of :class:`Parameter <.params.Parameter>` definitions available to all application
             paths.
+        path: A path fragment that is prefixed to all route handlers, controllers and routers associated
+            with the application instance.
```
JacobCoffee, line 206:
```suggestion
        path: A path fragment that is prefixed to all route handlers, controllers and routers associated
            with the application instance.
            
            .. versionadded:: 2.8.0
```

litestar/testing/helpers.py
```diff
@@ -456,6 +461,8 @@ async def test_my_handler() -> None:
...
```
JacobCoffee, line 465:
```suggestion
        path: A path fragment that is prefixed to all route handlers, controllers and routers associated
            with the application instance.
            
            .. versionadded:: 2.8.0
```

litestar/config/app.py
```diff
@@ -151,6 +151,7 @@ class AppConfig:
     """
     parameters: ParametersMap = field(default_factory=dict)
     """A mapping of :class:`Parameter <.params.Parameter>` definitions available to all application paths."""
+    path: str = field(default="")
```
JacobCoffee, line 154:
```suggestion
    path: str = field(default="")
    """A base path that prefixed to all route handlers, controllers and routers associated with the 
    application instance.
    
    .. versionadded:: 2.8.0
    """
```

JacobCoffee (changes requested):

JacobCoffee:
seems good :) thx

PR #3315, opened 2024-04-05
fix: asgi lifespan msg after lifespan context exception

peterschutt:
An exception raised within an asgi lifespan context manager would result in a "lifespan.startup.failed" message being sent after we've already sent a "lifespan.startup.complete" message. This would cause uvicorn to raise a `STATE_TRANSITION_ERROR` assertion error due to their [check for that condition][1], if asgi lifespan is forced (i.e., with `$ uvicorn test_apps.test_app:app --lifespan on`).

E.g.,

```python-traceback

  File "/home/peter/.local/share/pdm/venvs/litestar-dj-FOhMr-3.8/lib/python3.8/site-packages/uvicorn/lifespan/on.py", line 86, in main
    await app(scope, self.receive, self.send)
  File "/home/peter/.local/share/pdm/venvs/litestar-dj-FOhMr-3.8/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
  File "/home/peter/PycharmProjects/litestar/litestar/app.py", line 568, in __call__
    await self.asgi_router.lifespan(receive=receive, send=send)  # type: ignore[arg-type]
  File "/home/peter/PycharmProjects/litestar/litestar/_asgi/asgi_router.py", line 180, in lifespan
    await send(failure_message)
  File "/home/peter/.local/share/pdm/venvs/litestar-dj-FOhMr-3.8/lib/python3.8/site-packages/uvicorn/lifespan/on.py", line 116, in send
    assert not self.startup_event.is_set(), STATE_TRANSITION_ERROR
AssertionError: Got invalid state transition on lifespan protocol.
```

This PR modifies `ASGIRouter.lifespan()` so that it sends a shutdown failure message if we've already confirmed startup. This is consistent with [starlette's behavior][2] under the same conditions.

[1]: https://github.com/encode/uvicorn/blob/a2219eb2ed2bbda4143a0fb18c4b0578881b1ae8/uvicorn/lifespan/on.py#L115-L117
[2]: https://github.com/encode/starlette/blob/4e453ce91940cc7c995e6c728e3fdf341c039056/starlette/routing.py#L744-L745

## Description

-

### DIFF
--- a/litestar/_asgi/asgi_router.py
+++ b/litestar/_asgi/asgi_router.py
@@ -158,27 +158,28 @@ class ASGIRouter:
         Returns:
             None.
         """
-
-        message = await receive()
         shutdown_event: LifeSpanShutdownCompleteEvent = {"type": "lifespan.shutdown.complete"}
         startup_event: LifeSpanStartupCompleteEvent = {"type": "lifespan.startup.complete"}
 
+        await receive()
+
+        started = False
         try:
             async with self.app.lifespan():
                 await send(startup_event)
-                message = await receive()
+                started = True
+                await receive()
 
         except BaseException as e:
             formatted_exception = format_exc()
             failure_message: LifeSpanStartupFailedEvent | LifeSpanShutdownFailedEvent
 
-            if message["type"] == "lifespan.startup":
-                failure_message = {"type": "lifespan.startup.failed", "message": formatted_exception}
-            else:
+            if started:
                 failure_message = {"type": "lifespan.shutdown.failed", "message": formatted_exception}
+            else:
+                failure_message = {"type": "lifespan.startup.failed", "message": formatted_exception}
 
             await send(failure_message)
-
             raise e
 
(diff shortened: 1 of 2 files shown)


litestar/_asgi/asgi_router.py
```diff
@@ -158,27 +158,28 @@ async def lifespan(self, receive: LifeSpanReceive, send: LifeSpanSend) -> None:
...
 
+        await receive()
+
+        started = False
         try:
             async with self.app.lifespan():
                 await send(startup_event)
-                message = await receive()
+                started = True
+                await receive()
```
peterschutt, line 171: The issue being that if the exception comes from the lifespan context before a shutdown message is received but after we've confirmed the startup, then `message` is still the startup one causing us to send the startup failure in the exception handler.

provinzkraut (approved):

PR #3321, opened 2024-04-05
chore: rebase v3 onto `main@2.8`

JacobCoffee:
Rebase v3 on latest main


JacobCoffee:
Feel like this is the wrong way of dealing with conflicts @litestar-org/maintainers

PR #3323, opened 2024-04-05
feat: problem details plugin

### LINKED ISSUE #3199: Enhancement: Support for RFC 7807 / 9457

peterschutt:
## Description

A plugin to enable usage of problem details as the response.

The way this works is by injecting an exception handler into the app level exception handlers to convert `ProblemDetailsException` into a response following the specification as per RFC 9457.

Users can pass in a mapping of exception types to callables that will convert those exception types into `ProblemDetailsException` for handling specific exceptions such as pydantic's `ValidationError`. That converted `ProblemDetailsException` will then be used to create the response. This should allow for flexibility when needed.

## Closes

Closes #3199.


cofin (approved):
Nice. This looks useful.  LGTM

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,135 @@
...
+
+@dataclass
+class ProblemDetailsConfig:
+    """The configuration object for ``ProblemDetailsPlugin.``"""
+
+    exception_handler: ProblemDetailsExceptionHandlerType = _problem_details_exception_handler
+    """The exception handler used for ``ProblemdetailsException.``"""
+
+    enable_for_all_http_exceptions: bool = False
+    """Flag indicating whether to convert all ``HTTPException`` into ``ProblemDetailsException.``"""
```
JacobCoffee, line 107:
```suggestion
    """Flag indicating whether to convert all :exc:`HTTPException` into ``ProblemDetailsException.``"""
```

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,130 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
bunny-therapist: Is it a problem that the current docs use application/problem+json as an example for json encoding when there is a plugin for it? https://docs.litestar.dev/2/usage/responses.html#json-responses

It might be fine? Technically there is nothing incorrect, it just might give someone the impression that that is the way to handle problem details, so they won't find the plugin.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,135 @@
...
+
+        problem_details = {"type": self.type_, "status": self.status_code}
+        for attr in ("title", "instance", "detail"):
+            if (value := getattr(self, attr, None)) is not None:
+                problem_details[attr] = value
+
+        if isinstance(self.extra, Mapping):
+            problem_details.update(self.extra)
+        elif self.extra is not None:
+            problem_details["extra"] = self.extra
```
bunny-therapist, line 89: What about the case when "extra" contains title, type, instance, status or detail? It will currently overwrite the preexisting ones. If we swapped the order of assignments, then those fields in extra would be overridden instead. I think I prefer that option, since especially overriding status seems a bit weird.

Then again, it is kind of a user error to put stuff like that in the "extra" dict, so I don't think it is a big deal.

I assume there is no way to make a TypedDict with forbidden keys, right?

bunny-therapist:
What is the status on this?

peterschutt:
> What is the status on this?

At risk of stating the obvious, it looks like its waiting on some comments to be resolved and a rebase.

guacs:
> What is the status on this?

I'll get it fixed in a few days. I've been a bit busy.

litestar/plugins/problem_details.py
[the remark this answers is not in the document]
guacs: Yupp, we'll link to the plugin from there instead.

litestar/plugins/problem_details.py
guacs: I was thinking it's better to leave an option for the user to be able to override those keys. If we always end up overriding `extra`, then the users would have no way of overriding it. This isn't a concern if they're raising the `ProblemDetailsException` directly, but would be helpful for other custom errors they raise that then get converted into problem details via this plugin.

We should document this behavior explictily so that users can be aware of it.

litestar/plugins/problem_details.py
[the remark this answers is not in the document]
bunny-therapist: Does that make sense? The example isn't about how to use problem details, but how to use "+json" suffixed media types. Would it make more sense to change the example to use some arbitrary "+json" suffixed media type?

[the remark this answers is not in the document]
guacs: Good point. Is there another common usecase of the suffix? If not, I was thinking we could keep that, but then also note that there is a builtin plugin for this specific case.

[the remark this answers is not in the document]
guacs: Okay, I just used a generic example for this instead of problem details.

litestar/plugins/problem_details.py
```diff
@@ -75,14 +77,22 @@ def __init__(
...
                 problem_details[attr] = value
 
+        if isinstance(self.extra, Mapping):
+            problem_details.update(self.extra)
+        elif (extra := self.extra) is not None:
+            if isinstance(extra, Mapping):
+                problem_details.update(extra)
+            else:
+                problem_details["extra"] = extra
+
```
bunny-therapist, line 95: Isn't there some strange redundancy in these conditionals?
Both L88 and L91 check if self.extra is a Mapping, meaning that if it is a Mapping, L92 will never happen. Right?

To me, it looks like L91-93 can be removed and L94 can be unindented one step to replace the whole interior conditional.

bunny-therapist:
This is tagged for 2.9.0 but 2.9.0 was just released. I think the milestone needs to be updated.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,140 @@
...
+            media_type=self._PROBLEM_DETAILS_MEDIA_TYPE,
+            status_code=self.status_code,
+        )
+
+
+@dataclass
+class ProblemDetailsConfig:
+    """The configuration object for ``ProblemDetailsPlugin.``"""
+
+    exception_handler: ProblemDetailsExceptionHandlerType = _problem_details_exception_handler
```
Anu-cool-007, line 108: Can `_problem_details_exception_handler` accept an optional callable from the user?
This will help add custom side effects while handling these exceptions.
One use case can be to send some analytics for certain exceptions.

bunny-therapist:
This is tagged for v2.10.0 but it just came out and it did not contain this at all.

JacobCoffee:
dang.. sorry about this @bunny-therapist - yeah it needed to be merged to be included - perhaps @Alc-Alc had a reason for not merging ready PRs before?

@litestar-org/maintainers

Alc-Alc:
> perhaps @Alc-Alc had a reason for not merging ready PRs before?

it was on oversight on my side , I just forgot🙃

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,140 @@
...
+    _PROBLEM_DETAILS_MEDIA_TYPE = "application/problem+json"
+
+    def __init__(
+        self,
+        *args: Any,
+        detail: str = "",
+        status_code: int | None = None,
+        headers: dict[str, str] | None = None,
+        extra: dict[str, Any] | list[Any] | None = None,
+        type_: str = "about:blank",
```
agronholm, line 57: Couldn't this be `None` by default? The spec says that this value is assumed anyway if it's absent.

litestar/plugins/problem_details.py
```diff
@@ -0,0 +1,140 @@
...
+    def _exception_handler(req: Request, exc: exc_type) -> Response:  # type: ignore[valid-type]
+        problem_details_exc = exc_to_problem_details_exc_fn(exc)
+
+        return problem_details_exc.to_response(req)
+
+    return _exception_handler
+
+
+def _http_exception_to_problem_detail_exception(exc: HTTPException) -> ProblemDetailsException:
+    return ProblemDetailsException(status_code=exc.status_code, detail=exc.detail, extra=exc.extra, headers=exc.headers)
```
agronholm, line 42: Perhaps `HTTPException.detail` should rather map to to the `title` field? The `detail` field in `HTTPException` typically contains just a terse HTTP status name, e.g. `Bad Request`.

PR #3331, opened 2024-04-07
refactor(routing): Move kwargs model creation to handler

provinzkraut:
Move the creation of the `KwargsModel` from the routes to the route handlers. Since this operates on the handlers almost exclusively, it makes more sense to have it in there.

### DIFF
--- a/litestar/routes/base.py
+++ b/litestar/routes/base.py
@@ -138,7 +138,9 @@ class BaseRoute(ABC):
             )
 
     @classmethod
-    def _parse_path(cls, path: str) -> tuple[str, str, list[str | PathParameterDefinition], dict[str, PathParameterDefinition]]:
+    def _parse_path(
+        cls, path: str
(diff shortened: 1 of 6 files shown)


litestar/routes/base.py
```diff
@@ -123,23 +118,6 @@ async def handle(self, scope: Scope, receive: Receive, send: Send) -> None:
...
 
-    def create_handler_kwargs_model(self, route_handler: BaseRouteHandler) -> KwargsModel:
-        """Create a `KwargsModel` for a given route handler."""
-
-        path_parameters = set()
-        for param in self.path_parameters:
-            if param.name in path_parameters:
-                raise ImproperlyConfiguredException(f"Duplicate parameter '{param.name}' detected in '{self.path}'.")
-            path_parameters.add(param.name)
-
```
provinzkraut, line 134: No need to do this every time we create the kwargs model; For routes with multiple handlers, we'd do this multiple times

peterschutt (approved):
Looks good!

PR #3322, opened 2024-04-05
feat: add `SecretString` and `SecretBytes` datastructures

peterschutt:
## Description

This pull request implements `SecretString` and `SecretBytes` data structures to hide sensitive data in tracebacks, etc.

## Closes

Closes #1312
Closes #3248

PR #3332, opened 2024-04-07
ci:  add labeler for prs

JacobCoffee:
## Description

- adds an automatic labeler for PRs for project and release automations

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -271,9 +271,8 @@ labels:
   - label: 'area/private-api'
     sync: true
     matcher:
-    files:
-      any: ['litestar/_*.py', 'litestar/*/_*.py', 'litestar/_*/**/*.py']
-
+      files:
+        any: ['litestar/_*.py', 'litestar/*/_*.py', 'litestar/_*/**/*.py']
 
   # -- Size Based Labels -------------------------------------------------------
   - label: 'size: small'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,337 @@
+version: v1
+
+labels:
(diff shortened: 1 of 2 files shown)


peterschutt (approved):
Looks good!

What is the objective of the size labels, and did those numbers come from anywhere in particular? I am not against having them there as I can see the benefit of knowing where to look for a quick PR to review if there is not much time up my sleeve, but I also can see the opposite, where PRs will be ignored based on having a label of "large". For example, in my head #3322 wasn't a big PR, it only touched 4 files inside the library, but after adding usage docs, api docs, examples and tests it ended up touching 17 files.

I'm still happy to try them out though and we can tweak the numbers as we go if necessary.

JacobCoffee:
@peterschutt
> What is the objective of the size labels, and did those numbers come from anywhere in particular?

thin air / the GH action. I was more pleased with the diff +/- but this action didnt have an option for that only file count 

image

But the others i looked at didnt have the ability to label based on branch... so :\

.github/labeler.yml
```diff
@@ -0,0 +1,306 @@
...
+        count:
+          gte: 5
+          lte: 10
+
+  - label: 'size: large'
+    sync: true
+    matcher:
+      files:
+        count:
+          gte: 11
```
JacobCoffee, line 299:
```suggestion
  - label: 'size: small'
    sync: true
    matcher:
      files:
        count:
          gte: 1
          lte: 10

  - label: 'size: medium'
    sync: true
    matcher:
      files:
        count:
          gte: 10
          lte: 25

  - label: 'size: large'
    sync: true
    matcher:
      files:
        count:
          gte: 26
```

PR #3303, opened 2024-04-02
docs: update `docs/usage/security/*`

JacobCoffee:
fix:  update incorrect statement in middleware/authentication

## Description

- Updates security usage docs as part of #3212 work

PR #3344, opened 2024-04-09
docs: update ``docs/usage/security/*``

JacobCoffee:
## Description

- Updates security usage docs as part of #3212 work
- fix:  update incorrect statement in middleware/authentication
- Follow up of https://github.com/litestar-org/litestar/pull/3303


docs/usage/security/abstract-authentication-middleware.rst
```diff
@@ -1,83 +1,91 @@
...
 
+        class User(Base):
+            id: uuid.UUID | None = Column(
+                UUID(as_uuid=True), default=uuid.uuid4, primary_key=True
+            )
+            # ... other fields follow, but we only require id for this example
 
 We will also need some utility methods to encode and decode tokens. To this end we will use
-the `python-jose <https://github.com/mpdavis/python-jose>`_ library. We will also create a pydantic model representing a
+the `python-jose <https://github.com/mpdavis/python-jose>`_ library. We will also create a Pydantic model representing a
```
JacobCoffee, line 73: Maybe we should step away from external lib usage like this and use dataclass instead

JacobCoffee:
ready minus one comment that would be converted to an issue if its something we want to do

docs/usage/security/abstract-authentication-middleware.rst
```diff
@@ -1,83 +1,91 @@
...
-    class User(Base):
-        id: uuid.UUID | None = Column(
-            UUID(as_uuid=True), default=uuid.uuid4, primary_key=True
-        )
-        # ... other fields follow, but we only require id for this example
+        Base = declarative_base()
 
 
+        class User(Base):
+            id: uuid.UUID | None = Column(
```
Alc-Alc, line 67: should we use SQLA 2.x ``mapped_column`` so that we are consistent with the other examples where we use SQLA models?

docs/usage/security/abstract-authentication-middleware.rst
```diff
@@ -1,83 +1,91 @@
...
+        from sqlalchemy import Column
+        from sqlalchemy.dialects.postgresql import UUID
+        from sqlalchemy.orm import declarative_base
 
-    class User(Base):
-        id: uuid.UUID | None = Column(
-            UUID(as_uuid=True), default=uuid.uuid4, primary_key=True
-        )
-        # ... other fields follow, but we only require id for this example
+        Base = declarative_base()
```
Alc-Alc, line 63: same here, 2.x way?

docs/usage/security/abstract-authentication-middleware.rst
JacobCoffee: yes

docs/usage/security/abstract-authentication-middleware.rst
peterschutt: We'd have to handle the type coercion that case, but I think better to use either msgspec and `convert()` if we don't want to roll the type parsing ourselves, or dataclasses.

peterschutt (approved):

docs/usage/security/abstract-authentication-middleware.rst
```diff
@@ -1,83 +1,91 @@
...
-Litestar exports :class:`AbstractAuthenticationMiddleware <.middleware.authentication.AbstractAuthenticationMiddleware>`,
-which is an Abstract Base Class (ABC) that implements the :class:`MiddlewareProtocol <.middleware.base.MiddlewareProtocol>`.
+Litestar exports :class:`~.middleware.authentication.AbstractAuthenticationMiddleware`, which is an
+:term:`abstract base class` (ABC) that implements the :class:`~.middleware.base.MiddlewareProtocol`.
 To add authentication to your app using this class as a basis, subclass it and implement the abstract method
-:meth:`authenticate_request <.middleware.authentication.AbstractAuthenticationMiddleware.authenticate_request>`:
+:meth:`~.middleware.authentication.AbstractAuthenticationMiddleware.authenticate_request`:
 
-.. code-block:: python
+.. dropdown:: Click to see an example
```
provinzkraut, line 10: Not a fan of these dropdowns everywhere tbh :eyes: 
When I'm reading docs, I *want* to see the code. This just makes it harder to skim and see what's actually going on. I'd be pretty annoyed if I'd have to expand every code example like this. Also, since the text below this is referring to the content of the dropdown, having it hidden by default feels a bit odd. 

> As you can see [...]

But I *can't* see :grimacing: 

IMO dropdowns should be used if there's some additional information, not on the main content.

PR #3309, opened 2024-04-04
docs: update `usage/caching`

JacobCoffee:
## Description

- updates `usage/caching`  as part of #3212 work

PR #3345, opened 2024-04-09
docs: update ``usage/caching``

JacobCoffee:
## Description

- updates `usage/caching`  as part of #3212 work
- followup of https://github.com/litestar-org/litestar/pull/3309


JacobCoffee:
Good to go minus https://github.com/litestar-org/litestar/pull/3309#discussion_r1550874080

docs/usage/caching.rst
```diff
@@ -15,59 +15,63 @@ expected. Litestar comes with a simple mechanism for caching:
...
-any :class:`Store <.stores.base.Store>`, for example :class:`RedisStore <.stores.redis.RedisStore>`:
+By default, caching will use the :class:`~.stores.memory.MemoryStore`, but it can be configured with
+any :class:`~.stores.base.Store`, for example :class:`~.stores.redis.RedisStore`:
 
 .. code-block:: python
+    :caption: Using Redis as the cache store.
 
-   from litestar.config.cache import ResponseCacheConfig
-   from litestar.stores.redis import RedisStore
+    from litestar.config.cache import ResponseCacheConfig
```
Alc-Alc, line 69: I think this is a remnant of an old pattern. IIUC the snippet in the docs would no longer work.

The following is an MCVE that shows how to use a custom store using the current interfaces.

```py
import asyncio
from litestar.config.response_cache import ResponseCacheConfig
from litestar.stores.redis import RedisStore


@get(cache=10)
async def something() -> str:
    await asyncio.sleep(1)
    return "something"


redis_store = RedisStore.with_client(url="redis://localhost", port=6379, db=0)
cache_config = ResponseCacheConfig(store="redis_backed_store")
app = Litestar(
    [something],
    stores={"redis_backed_store": redis_store},
    response_cache_config=cache_config,
)
```

Some may argue this is not good DX, I personally do not find this any difficult.

JacobCoffee: This is more evidence towards us needing to move any examples into `examples/` and have them tested with each PR so they dont go stale.

Alc-Alc (approved):
One could say the example docs were... *cached* 🙃

PR #3347, opened 2024-04-09
ci: add pydantic v1 test

peterschutt:
## Description

-


.github/workflows/ci.yml
```diff
@@ -275,3 +275,61 @@ jobs:
           path: |
             docs/_build/html
             .pr_number
+
```
JacobCoffee, line 278: this is a good move imo

.github/workflows/ci.yml
```diff
@@ -275,3 +275,61 @@ jobs:
...
+  test_minimal_app:
+    name: Test Minimal Application with Base Dependencies
```
JacobCoffee, line 280: Do we want to run these if `test` flow fails in general? maybe add a `requires: test` or something to these so that they arent wastefully ran if other issues are present? idk

JacobCoffee (approved):

PR #3350, opened 2024-04-09
docs: update ``usage/cli``

JacobCoffee:
## Description

- updates ``usage/cli`` docs as part of #3212 work


docs/usage/cli.rst
```diff
@@ -38,21 +43,21 @@ The autodiscovery follows these lookup locations in order:
...
-4. A callable named ``create_app``
-5. A callable annotated to return an instance of :class:`Litestar <.app.Litestar>`
+1. An :term:`object` named ``app`` that is an instance of :class:`Litestar <.app.Litestar>`
+2. An :term:`object` named ``application`` that is an instance of :class:`Litestar <.app.Litestar>`
+3. Any :term:`object` that is an instance of :class:`Litestar <.app.Litestar>`
+4. A :term:`callable` named ``create_app``
+5. A :term:`callable` annotated to return an instance of :class:`Litestar <.app.Litestar>`
 
 Commands
 --------
```
JacobCoffee, line 53: Okay hear me out - We have some unique things in here but with a big note to maintainability and scalability, they are not terribly useful compared to what the Click directive gives use (see https://docs.litestar.dev/latest/reference/cli.html)

@litestar-org/maintainers wdyt about removing this and just moving the reference of the commands up some.

My only concern is things like https://docs.litestar.dev/latest/usage/cli.html#ssl.

I was thinking we could enhance the `help=` for those commands but that file would become garagantuan.. I'm not sure of a better way. We could have a few sections I guess for special things like this - but I feel like they would become stale without us seeing it. The "Help" for those seems already good enough.. e.g., `litestar run --help` -> look at what the subcommands are.

Thoughts?

JacobCoffee: This too coupled with this massive ass table.

We could move it to a csv or similar sphinx table structure but I don't think it solves my concerns.

JacobCoffee: No pressure on just keeping it, too.

peterschutt (approved):

provinzkraut (approved):
I still think the tables are more comprehensive, but I'm with @peterschutt on the maintainability aspect

PR #3377, opened 2024-04-11
test: Add tests for `defining_dtos_on_layers.py` docs example

kedod:
## Description

- Add tests for `defining_dtos_on_layers.py` docs example


docs/examples/data_transfer_objects/defining_dtos_on_layers.py
```diff
@@ -1,31 +1,45 @@
...
-from .models import User, UserDTO, UserReturnDTO
+
+@dataclass
+class User:
+    id: UUID
+    name: str
+
+
+UserDTO = DataclassDTO[User]
+UserReturnDTO = DataclassDTO[User]
```
peterschutt, line 17: There isn't really a point in having a distinct data and return DTO if they are both the same.

There's two things we could do:

1. change the example to only use `UserDTO` and not define `return_dto` on the router (in that case `UserDTO` will apply to both data and return). We would need to change the surrounding documentation along with this.
2. configure `UserReturnDTO` so that it is different, e.g., add a `secret` field to the model and exclude it from the return data.

I think I'd prefer 2.

peterschutt (approved):
Looks good!

docs/examples/data_transfer_objects/defining_dtos_on_layers.py
peterschutt: perfect!

docs/examples/data_transfer_objects/defining_dtos_on_layers.py
```diff
@@ -1,31 +1,53 @@
...
     def update_user(self, data: User) -> User:
         return data
 
-    @delete("/{user_id:uuid}", return_dto=None)
+    @delete("/{user_id:uuid}", return_dto=None, sync_to_thread=False)
     def delete_user(self, user_id: UUID) -> None:
         return None
+
+
+app = Litestar(UserController)
```
Alc-Alc, line 53: shouldn't this be?
```suggestion
app = Litestar([UserController])
```

Alc-Alc: seems its related to concurrency issues, https://github.com/litestar-org/litestar/pull/3382 disabled concurrency and the tests ran fine

JacobCoffee (approved):

PR #3386, opened 2024-04-13
refactor(routing)!: Refactor routes and route handlers

provinzkraut:
Refactor routes and route handlers in an attempt to simplify configuration.

The two general goals here are:

1. Routes should not be concerned with any logic pertaining to handlers
2. Routing layers (e.g. app/router instances) should not perform any manual handler setup

In general I'd like to restrict the routes to an internal part of the ASGI routing, only taking care of mapping a path > route handler(s), and establishing the connection. Everything else should be taken care of by the handlers.

What this currently does:

- Remove the `handler_names` property from `HTTPRoute`
- Move the `OPTIONS` handler creation from the `HTTPRoute` to the route registration process
- Remove the `methods` attribute from `BaseRoute`; It does not make sense on the asgi/websocket routes
- Move the creation of the `KwargsModel` from the routes to the handlers and remove the outside mutation of the handlers
- Move the mapping of route > handler > kwargsmodel to the handler itself
- Move the actual route handling to the handler
- Establish `ASGIConnection`/`WebSocket`/`Request` in the routes and pass them to handlers instead of raw ASGI objects (`scope`, `receive`, `send`)
- Some minor refactorings that were enabled by these things

### DIFF
--- a/litestar/_asgi/routing_trie/mapping.py
+++ b/litestar/_asgi/routing_trie/mapping.py
@@ -143,8 +143,7 @@ def configure_node(
         node.path_parameters = {}
 
     if isinstance(route, HTTPRoute):
-        for method, handler_mapping in route.route_handler_map.items():
-            handler, _ = handler_mapping
+        for method, handler in route.route_handler_map.items():
             node.asgi_handlers[method] = ASGIHandlerTuple(
                 asgi_app=build_route_middleware_stack(app=app, route=route, route_handler=handler),
                 handler=handler,
--- a/litestar/handlers/asgi_handlers.py
+++ b/litestar/handlers/asgi_handlers.py
@@ -85,14 +85,14 @@ class ASGIRouteHandler(BaseRouteHandler):
 
     async def handle(self, connection: ASGIConnection[ASGIRouteHandler, Any, Any, Any]) -> None:
         """ASGI app that authorizes the connection and then awaits the handler function.
-    
-    .. versionadded: 3.0
+
+        .. versionadded: 3.0
 
         Args:
-            connection: The ASGI connection
+                connection: The ASGI connection
 
         Returns:
-            None
+                None
         """
 
         if self.resolve_guards():
--- a/litestar/handlers/asgi_handlers.py
+++ b/litestar/handlers/asgi_handlers.py
@@ -85,6 +85,8 @@ class ASGIRouteHandler(BaseRouteHandler):
 
     async def handle(self, connection: ASGIConnection[ASGIRouteHandler, Any, Any, Any]) -> None:
         """ASGI app that authorizes the connection and then awaits the handler function.
+    
+    .. versionadded: 3.0
 
         Args:
             connection: The ASGI connection
--- a/litestar/handlers/asgi_handlers.py
+++ b/litestar/handlers/asgi_handlers.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Mapping, Sequence
 
+from litestar.connection import ASGIConnection
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.handlers.base import BaseRouteHandler
 from litestar.types.builtin_types import NoneType
@@ -15,6 +16,9 @@ if TYPE_CHECKING:
         ExceptionHandlersMap,
         Guard,
         MaybePartial,  # noqa: F401
+        Receive,
+        Scope,
+        Send,
     )
 
 
@@ -81,6 +85,24 @@ class ASGIRouteHandler(BaseRouteHandler):
             )
         if not is_async_callable(self.fn):
             raise ImproperlyConfiguredException("Functions decorated with 'asgi' must be async functions")
+
+    async def handle(self, scope: Scope, receive: Receive, send: Send) -> None:
+        """ASGI app that authorizes the connection and then awaits the handler function.
+
+        Args:
+            scope: The ASGI connection scope.
+            receive: The ASGI receive function.
+            send: The ASGI send function.
+
+        Returns:
+            None
+        """
+
+        if self.resolve_guards():
+            connection = ASGIConnection["ASGIRouteHandler", Any, Any, Any](scope=scope, receive=receive)
+            await self.authorize_connection(connection=connection)
+
+        await self.fn(scope=scope, receive=receive, send=send)
 
 
 asgi = ASGIRouteHandler
--- a/litestar/handlers/asgi_handlers.py
+++ b/litestar/handlers/asgi_handlers.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Mapping, Sequence
 
-from litestar.connection import ASGIConnection
 from litestar.exceptions import ImproperlyConfiguredException
 from litestar.handlers.base import BaseRouteHandler
 from litestar.types.builtin_types import NoneType
@@ -12,13 +11,11 @@ __all__ = ("ASGIRouteHandler", "asgi")
 
 
 if TYPE_CHECKING:
+    from litestar.connection import ASGIConnection
     from litestar.types import (
         ExceptionHandlersMap,
         Guard,
         MaybePartial,  # noqa: F401
-        Receive,
-        Scope,
-        Send,
     )
 
 
@@ -86,23 +83,20 @@ class ASGIRouteHandler(BaseRouteHandler):
         if not is_async_callable(self.fn):
             raise ImproperlyConfiguredException("Functions decorated with 'asgi' must be async functions")
 
-    async def handle(self, scope: Scope, receive: Receive, send: Send) -> None:
+    async def handle(self, connection: ASGIConnection[ASGIRouteHandler, Any, Any, Any]) -> None:
         """ASGI app that authorizes the connection and then awaits the handler function.
 
         Args:
-            scope: The ASGI connection scope.
-            receive: The ASGI receive function.
-            send: The ASGI send function.
+            connection: The ASGI connection
 
         Returns:
             None
         """
 
         if self.resolve_guards():
-            connection = ASGIConnection["ASGIRouteHandler", Any, Any, Any](scope=scope, receive=receive)
             await self.authorize_connection(connection=connection)
 
-        await self.fn(scope=scope, receive=receive, send=send)
+        await self.fn(scope=connection.scope, receive=connection.receive, send=connection.send)
 
 
 asgi = ASGIRouteHandler
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -62,7 +62,6 @@ class BaseRouteHandler:
         "_resolved_type_decoders",
         "_resolved_type_encoders",
         "_signature_model",
-        "_kwargs_model",
         "dependencies",
         "dto",
         "exception_handlers",
@@ -133,7 +132,6 @@ class BaseRouteHandler:
         self._resolved_type_decoders: TypeDecodersSequence | EmptyType = Empty
         self._resolved_type_encoders: TypeEncodersMap | EmptyType = Empty
         self._signature_model: type[SignatureModel] | EmptyType = Empty
-        self._kwargs_model: KwargsModel | EmptyType = Empty
 
         self.dependencies = dependencies
         self.dto = dto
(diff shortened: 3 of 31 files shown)


litestar/handlers/base.py
```diff
@@ -526,7 +527,7 @@ def _validate_dependency_is_unique(dependencies: dict[str, Provide], key: str, p
                     f"If you wish to override a provider, it must have the same key."
                 )
 
-    def on_registration(self, app: Litestar) -> None:
+    def on_registration(self, app: Litestar, route: BaseRoute) -> None:
```
provinzkraut, line 530: Not too psyched about having to pass in the route here, but then again we also pass it to the `ReceiveRoutePlugin` :woman_shrugging:

litestar/handlers/http_handlers/base.py
```diff
@@ -566,6 +570,14 @@ def on_registration(self, app: Litestar) -> None:
             self._fn = ensure_async_callable(self.fn)
             self.has_sync_callable = False
 
+        self._get_kwargs_model_for_route(route)
+
+    def _get_kwargs_model_for_route(self, route: BaseRoute) -> KwargsModel:
```
provinzkraut, line 575: This is needed because the same handler can have different `KwargsModel`s  for different routes, because their `path_parameters` might differ. Logically though, the `KwargsModel` belongs to the handlers, and the only thing we do with it on the routes is pass it back to the handler, which also doesn't feel too great.

This isn't the final step yet and I'm hoping to get to a point where we don't need to do this memo step here at all

litestar/routes/http.py
```diff
@@ -73,255 +48,20 @@ async def handle(self, scope: HTTPScope, receive: Receive, send: Send) -> None:
...
-
-        await response(scope, receive, send)
-
-        if after_response_handler := route_handler.resolve_after_response():
-            await after_response_handler(request)
-
-        if form_data := scope.get("_form", {}):
-            await self._cleanup_temporary_files(form_data=cast("dict[str, Any]", form_data))
+        route_handler = self.route_handler_map[scope["method"]]
+        connection = route_handler.resolve_request_class()(scope=scope, receive=receive, send=send)
```
provinzkraut, line 52: Ideally this also wouldn't happen here, so the route would simply take in a `connection_class`, but that would require further config resolution in case of the HTTP handlers. I'm going to leave this up to another PR

litestar/routes/http.py
```diff
@@ -36,32 +23,20 @@ def __init__(
...
-            options_handler = self.create_options_handler(path)
-            options_handler.owner = route_handlers[0].owner
-            route_handlers.append(options_handler)
-
-        self.route_handlers = route_handlers
-        self.route_handler_map: dict[Method, tuple[HTTPRouteHandler, KwargsModel]] = {}
+        super().__init__(path=path)
+        self.route_handler_map: dict[Method, HTTPRouteHandler] = self.create_handler_map(route_handlers)
+        self.route_handlers = tuple(self.route_handler_map.values())
+        self.methods = tuple(self.route_handler_map)
```
provinzkraut, line 37: These might as well be properties, but that requires implementing a `__deepcopy__` method, and I hope we can get rid of the copying entirely, so I'm going to leave this as-is for now.

litestar/_asgi/routing_trie/mapping.py
```diff
@@ -143,8 +143,7 @@ def configure_node(
         node.path_parameters = {}
 
     if isinstance(route, HTTPRoute):
-        for method, handler_mapping in route.route_handler_map.items():
-            handler, _ = handler_mapping
+        for method, handler in route.route_handler_map.items():
```
provinzkraut, line 146: I think we can get rid of the references to the handlers within the asgi router entirely, and have the resolution back to a handler happen at a later point via the routes. I'm gonna leave this as-is for now though and maybe later handle this change in a separate PR

litestar/handlers/base.py
peterschutt: Why?

`ReceiveRoutePlugin` was intentionally left undocumented. It was added to support the openapi generation refactor a while back, but the fact that is there doesn't bind us to iterating over the routes in this `Litestar.register()`.

litestar/handlers/asgi_handlers.py
```diff
@@ -82,5 +83,20 @@ def _validate_handler_function(self) -> None:
         if not is_async_callable(self.fn):
             raise ImproperlyConfiguredException("Functions decorated with 'asgi' must be async functions")
 
+    async def handle(self, connection: ASGIConnection[ASGIRouteHandler, Any, Any, Any]) -> None:
```
peterschutt, line 86: Should we make `BaseRouteHandler` abstract on this?

litestar/handlers/asgi_handlers.py
```diff
@@ -82,5 +83,20 @@ def _validate_handler_function(self) -> None:
...
+        """ASGI app that authorizes the connection and then awaits the handler function.
```
peterschutt, line 87: Not really an asgi app anymore :)

litestar/handlers/http_handlers/base.py
```diff
@@ -598,5 +619,139 @@ def _validate_handler_function(self) -> None:
         if "data" in self.parsed_fn_signature.parameters and "GET" in self.http_methods:
             raise ImproperlyConfiguredException("'data' kwarg is unsupported for 'GET' request handlers")
 
+    async def handle(self, connection: Request[Any, Any, Any]) -> None:
+        """ASGI app that creates a Request from the passed in args, determines which handler function to call and then
```
peterschutt, line 623: same

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -97,5 +103,46 @@ def _validate_handler_function(self) -> None:
         if not is_async_callable(self.fn):
             raise ImproperlyConfiguredException("Functions decorated with 'websocket' must be async functions")
 
+    def on_registration(self, app: Litestar, route: BaseRoute) -> None:
+        super().on_registration(app=app, route=route)
+        self._kwargs_model = self._create_kwargs_model(path_parameters=route.path_parameters)
+
+    async def handle(self, connection: WebSocket[Any, Any, Any]) -> None:
+        """ASGI app that creates a WebSocket from the passed in args, and then awaits the handler function.
```
peterschutt, line 111: same

litestar/router.py
```diff
@@ -259,17 +260,15 @@ def register(self, value: ControllerRouterHandler) -> list[BaseRoute]:
 
     @property
     def route_handler_method_map(self) -> dict[str, RouteHandlerMapItem]:
-        """Map route paths to :class:`RouteHandlerMapItem <litestar.types.internal_typ es.RouteHandlerMapItem>`
+        """Map route paths to :class:`RouteHandlerMapItem <litestar.types.internal_types.RouteHandlerMapItem>`
```
peterschutt, line 263:
```suggestion
        """Map route paths to :class:`~litestar.types.internal_types.RouteHandlerMapItem`
```
I've noticed @JacobCoffee (aka docs-god) doing this lately, it only renders in the docs as `RouterHandlerMapItem`.

litestar/routes/http.py
peterschutt: Why can't it happen in the handler's `handle()` method? IOW, why can't the handler's `handle()` method be an asgi app and create the connection object for themselves?

peterschutt (approved):

litestar/routes/http.py
provinzkraut: Ideally the split would be that the routes handle the ASGI stuff and the handlers shouldn't need to be concerned with this. I know we're not there at all, but I felt like this was a good step if we want to move in that direction.
wdyt?

peterschutt: Can you elaborate on what makes drawing the line for asgi interfaces there ideal? Not arguing at all - you've put a lot of thought into this so I just want to fully understand your PoV.

> the handlers shouldn't need to be concerned with this

It feels to me like a pretty thin line between handing a connection object and the individual asgi components to the handler, yet it would remove the need to book-keep the appropriate connection class type for the handler on the route side.

litestar/handlers/base.py
provinzkraut: Hmm. We only need this information for the `HTTPRoute` to create the kwarg models for the different routes. In theory we could pass the path params, as that's the only thing that changes, but I thought that would be an even more leaky abstraction.

Since this PR is more of an intermediary step with the configuration changes, and the end goal is to get rid of the need for the `on_registration` callback entirely, I felt like this was an ok trade off to make for now as it's not *that* bad if we leave it in in the end. wdyt?

litestar/routes/http.py
provinzkraut: You are right, this is rather arbitrary but here's my thought process so far, (keep in mind that this PR is intended to serve as a first step towards a larger refactoring goal):

In the end, I'd like routes to not be concerned with the route handler business logic at all. They simply map one or more async callables to a path (and in the case of HTTP a few methods), establish the connection and pass it to the callable.
This would also allow us to easily split off that part into a separate package if we want to do so in the future; It seems natural to me that routes are seen as a part of the routing logic, whereas handlers represent the framework side.

As for the connection vs. ASGI components, I'd also like to separate things more there, ideally only having the `ASGIResponse` classes dealing with the ASGI stuff, while everything else only handles higher-level abstractions.

I had actually already started moving the connection class config to the routes, but then I noticed it would require other changes and felt like that would be better kept as a separate PR. I could also make a PR into this branch for that, but tbh I'd like to keep these ones short lived to make them easier to keep up to date and review.

peterschutt: > It seems natural to me that routes are seen as a part of the routing logic

Hard to argue with that :)

> As for the connection vs. ASGI components, I'd also like to separate things more there, ideally only having the `ASGIResponse` classes dealing with the ASGI stuff, while everything else only handles higher-level abstractions.

Let me see if I'm following correctly. Currently, we receive a "http.request" message and routing, dependency resolution, and response generation all happens within the scope of nested ASGI apps (i.e., `ExceptionHandlerMiddleware` and `route.handle`).

In this future scenario, we receive the "http.request" message and route it, then call into a non-asgi handler framework that returns a `Response` object that is naive to the server protocol (i.e., it shouldn't know or care that we are ASGI or not). We receive the protocol-naive response and use that to the protocol-aware response and call it?

> I had actually already started moving the connection class config to the routes

Still to be configurable per-handler?

litestar/handlers/base.py
peterschutt: Sounds good.

litestar/routes/http.py
provinzkraut: > We receive the protocol-naive response and use that to create the protocol-aware response and call it?

That about sums it up, yes!

> Still to be configurable per-handler?

Of course! Basically, we would just do e.g. `WebSocketRoute(..., connection_class=handler.resolve_websocket_class()`.

peterschutt: > Of course! Basically, we would just do e.g. WebSocketRoute(..., connection_class=handler.resolve_websocket_class().

And something a little more complex for http given there could be multiple connection class types per route?

No blockers from me!

litestar/handlers/asgi_handlers.py
JacobCoffee, line 87:
```suggestion
        """ASGI app that authorizes the connection and then awaits the handler function.
    
    .. versionadded: 3.0
```

litestar/handlers/http_handlers/base.py
```diff
@@ -598,5 +619,139 @@ def _validate_handler_function(self) -> None:
...
+        handles the call.
```
JacobCoffee, line 624:
```suggestion
        handles the call.
    
    .. versionadded: 3.0
```

litestar/handlers/http_handlers/base.py
JacobCoffee, line 623:
```suggestion
        """ASGI app that creates a :class:`~.connection.Request` from the passed in args, determines which handler function to call and then
```

litestar/routes/http.py
provinzkraut: @peterschutt I've actually changed my mind and reverted this for now. I still think it makes sense to do, but it doesn't really belong into the scope of this PR. Leaving that out makes it a bit easier to handle things down the road, and the change is easy enough to apply at some other point. I'm aiming to do this now when I'm tackling the config stuff, where it makes more sense, as it's not an intermediate step there.

PR #3389, opened 2024-04-13
refactor: exception handler middleware optimizations

peterschutt:
Optimize use of exception handler middleware by:

- Only wrap `route.handle` with the exception handler middleware if there is other middleware defined for the route.
- Storing the resolved exception handlers for the route on the connection scope, and retrieve them from within the exception middleware.

We need to wrap the `route.handle` method if there is other middleware defined because that middleware may wrap the `send()` coroutine, and that send wrapper needs an exception handler middleware between itself and the handler in order for the wrapper to be called if the handler raises an exception. If there is no middleware between the route handler and the outer exception handler middleware, this is unnecessary.

Storing resolved exception handlers in the ASGI scope removes the requirement for the exception handler middleware that wraps the resolved middleware stack, as the outer most exception middleware can have access to the exception handlers that have been determined after route resolution.

In this PR we also deprecate `ExceptionHandlerMiddleware` from public API. It was not truly documented, but also not in a private namespace. It is deprecated because our use of the type is not customizable by the end user.

Alongside that deprecation, we move some utility functions out of the exception handler middleware namespace, into a public namespace, as these are generally more useful (currently used in `litestar-fullstack`). These are:

- `middleware.exceptions.middleware.create_exception_response` moved to `exceptions.responses.create_exception_response`
- `middleware.exceptions.middleware._debug_response.create_debug_response` moved to `exceptions.responses.create_debug_response`
- `middleware.exceptions.middleware.ExceptionResponseContent` moved to `exceptions.responses.ExceptionResponseContent`

## Description

-


provinzkraut (approved):
Nice!

JacobCoffee (approved):

litestar/_asgi/routing_trie/mapping.py
```diff
@@ -189,33 +189,49 @@ def build_route_middleware_stack(
...
+            asgi_handler = AllowedHostsMiddleware(app=asgi_handler, config=app.allowed_hosts)
+
+        for middleware in route_handler.resolve_middleware():
+            if hasattr(middleware, "__iter__"):
+                handler, kwargs = cast("tuple[Any, dict[str, Any]]", middleware)
+                asgi_handler = handler(app=asgi_handler, **kwargs)
+            else:
+                asgi_handler = middleware(app=asgi_handler)  # type: ignore[call-arg]
+
+    if exception_handlers:
```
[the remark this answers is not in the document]
peterschutt: Actually, we can go without this intermediate wrapper in all cases if we can put the exception handlers for the request into the connection scope when the handler is resolved for routing.

litestar/_asgi/asgi_router.py
```diff
@@ -83,9 +86,15 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
-        )
+        try:
+            asgi_app, scope["route_handler"], scope["path"], scope["path_params"] = self.handle_routing(
+                path=normalized_path, method=scope.get("method")
+            )
+        except Exception:
+            ScopeState.from_scope(scope).exception_handlers = self._app_exception_handlers
+            raise
+        else:
+            ScopeState.from_scope(scope).exception_handlers = scope["route_handler"].resolve_exception_handlers()
```
provinzkraut, line 97:
```suggestion
            asgi_app, route_handler, scope["path"], scope["path_params"] = self.handle_routing(
                path=normalized_path, method=scope.get("method")
            )
            scope["route_handler"] = route_handler
            ScopeState.from_scope(scope).exception_handlers = route_handler.resolve_exception_handlers()
        except Exception:
            ScopeState.from_scope(scope).exception_handlers = self._app_exception_handlers
            raise          
```

Avoiding one dictionary loopkup :P

litestar/_asgi/routing_trie/mapping.py
```diff
@@ -189,33 +189,36 @@ def build_route_middleware_stack(
...
+    handler_middleware = route_handler.resolve_middleware()
+    has_cached_route = isinstance(route, HTTPRoute) and any(r.cache for r in route.route_handlers)
+    has_middleware = (
+        app.csrf_config or app.compression_config or has_cached_route or app.allowed_hosts or handler_middleware
     )
 
-    if app.csrf_config:
-        asgi_handler = CSRFMiddleware(app=asgi_handler, config=app.csrf_config)
+    if has_middleware:
+        # if there is middleware, they may wrap the `send()` coro, so we need to wrap the handler in
```
provinzkraut, line 200:
```suggestion
        # if there is middleware, they may wrap the `send()` callable, so we need to wrap the handler in
```

Nit but it's not really a coroutine just yet but a coroutine function :eyes:

litestar/middleware/exceptions/middleware.py
```diff
@@ -183,16 +187,31 @@ def __init__(self, app: ASGIApp, debug: bool | None, exception_handlers: Excepti
 
         .. deprecated:: 2.0.0
             The ``debug`` parameter is deprecated. It will be inferred from the request scope
+
+        .. deprecated:: 2.9.0
+            The ``exception_handlers`` parameter is deprecated. It will be inferred from the application or the
```
provinzkraut, line 192: If we only ever infer the exception handlers, this makes this middleware truly internal, as it's not really useful for an end user. Maybe we should deprecate this thing entirely then from the public interface?

peterschutt: Is it currently public interface? TBH I added this b/c we'd done the same for the debug parameter, but the type isn't actually documented anywhere. Although I concede that not being in a private namespace muddies the water.

That aside, yeah, I agree we should deprecate the whole thing like CORS, b/c it requires special handling internally, and our use of it is hard coded. Unless there's a use case for adding extra EHM in the ASGI stack that I'm not seeing. The only thing I can think of is for catching errors from an interim middleware to covert it to a response before the next middleware. However, for that case I think it should be up to that middleware to convert expected errors into a response, and let unexpected errors propagate. Lo

litestar/_asgi/routing_trie/mapping.py
peterschutt: This whole comment was pretty poorly written, so I rewrote it and included this adjustement.

litestar/middleware/exceptions/middleware.py
provinzkraut: > However, for that case I think it should be up to that middleware to convert expected errors into a response, and let unexpected errors propagate.

Yeah, that sounds like the way to go

peterschutt: There's a good example of this in our own code.

In using `SessionAuth`, we add `session_auth.on_app_init` to the `on_app_init` application parameter, which inserts `SessionAuth.middleware` as the first middleware on the app config object.

`SessionAuth.middleware` is actually an instance of `DefineMiddleware` that in turn instantiates an instance of `MiddlewareWrapper`.

On first call, `MiddlewareWrapper` builds a new middleware stack around the asgi app it has received on instantiation. It wraps `SessionAuthMiddleware` with `ExceptionHandlerMiddleware`, and then wraps that with 
`SessionMiddleware`.

Due to the fact that `SessionMiddleware` wraps `send()`, and we use raising `NotAuthorizedException` or `PermissionDeniedException` to communicate from `AbstractAuthenticationMiddleware.authenticate_request()` that a request has failed auth'n, we then need to _immediately_ wrap that middleware in the exception handler middleware so that the failed requests are turned into a response and we call `send()` _before_ the `SessionMiddleware` in the asgi stack has been bypassed.

So by only adding `SessionAuth` to an application, prior to this PR a connection would pass through:

`EHM-1 -> EHM-2 -> SM -> EHM-3 -> SAM -> EHM-4 -> Route.handle`

Where:

- EHM-1: outermost exception handler middleware, wrapping routing to capture 404 etc
- EHM-2: exception handler middleware that wraps the routed asgi middleware stack
- SM: session middleware
- EHM-3: exception handler middleware to catch auth exceptions and convert to a response/call `send()` before SM
- SAM: session auth middleware
- EHM-4: exception handler middleware to catch handler exceptions
- Route.handle: the inner most asgi app that calls the handler

If instead of raising `NotAuthorizedException`/`PermissionDeniedException`, `AbstractAuthenticationMiddleware.authenticate_request()` returned some other signal, e.g., ` -> AuthenticationResult | NotAuthorized | PermissionDenied`, and the middleware itself immediately constructed and awaited the response, there would be no need to wrap `SessionAuthMiddleware` in `ExceptionHandlerMiddleware`.

If that was the case, and after this PR, we'd have reduced the 4 layers of EHM, down to 2:

`EHM-1 -> SM -> SAM -> EHM-2 -> Route.handle`

litestar/middleware/_internal/exceptions/middleware.py
```diff
@@ -0,0 +1,336 @@
...
+                await hook(e, scope)
+
+            if litestar_app.pdb_on_exception:
+                pdb.post_mortem()
+
+            if scope["type"] == ScopeType.HTTP:
+                await self.handle_request_exception(
+                    litestar_app=litestar_app, scope=scope, receive=receive, send=send, exc=e
+                )
+            else:
```
provinzkraut, line 255: What about exceptions in the asgi scope?

peterschutt: DYM something like `elif ScopeType == ScopeType.ASGI: ...`?

Actually, news to me that we use `ScopeType.ASGI` internally, as its not part of the spec I hadn't anticipated it.

I _think_ its OK, b/c the only place the IDE shows me that we read that value is when instantiating `ASGIRoute`, where we pass it to the `scope_type` parameter of `BaseRoute` - it doesn't look like we ever overwrite the value of "type" within the scope.

Nonetheless, I'll add an e2e test to ensure that exceptions raised out of a mounted asgi app are handled.

Side note: this code hasn't changed in this PR, only moved.

peterschutt: This commit adds that test: https://github.com/litestar-org/litestar/pull/3389/commits/1dddfe05d5051d304a1479c99eec8ad71ab51df8

LMK if that's what you meant.

provinzkraut: > LMK if that's what you meant.

Yup, exactly!

provinzkraut (approved):

PR #3392, opened 2024-04-14
docs: move ``.. code-block::`` directives into ``docs/examples/``

### LINKED ISSUE #3362: Docs: Move `.. code-block::` directives into `docs/examples/`
## Description

During docs updates we have run into a few places were we have stale code because they are not tested. 

For example we have some references to Pydantic in the core lib even though it was removed, Caching methods no longer available, etc.

We should move all of these out into `docs/examples` so that they can be tested inside of our CI to catch stale entries.

## From Comment:

This is more evidence towards us needing to move any examples into `examples/` and have them tested with each PR so they dont go stale.

_Originally posted by @JacobCoffee in https://github.com/litestar-org/litestar/pull/3345#discussion_r1556865895_

Kumzy:
## Description

Moving all code-block to files

## Closes

#3362


Kumzy:
Just to be sure, do I have to move also sh and json files?

JacobCoffee:
@Kumzy actually I don't think so

docs/examples/middleware/builtin/csrf_httpx.py
```diff
@@ -0,0 +1,21 @@
+import httpx
+
+with httpx.Client() as client:
+    get_response = client.get("http://localhost:8000/")
+
+    # "csrftoken" is the default cookie name
+    csrf = get_response.cookies["csrftoken"]
+
+    # "x-csrftoken" is the default header name
+    post_response_using_header = client.post("http://localhost:8000/", headers={"x-csrftoken": csrf})
```
Alc-Alc, line 10: this is a mistake I made when I wrote this, its not on you, but can you change this ? 🙂 
```suggestion
    post_response_using_header = client.post("http://localhost:8000/1", headers={"x-csrftoken": csrf})
```

Kumzy: Done

JacobCoffee:
I'm thinking maybe we shouldn't change the changelog ones here as special cases. They were "correct at that point in time of the changelog" and once we build v3 docs as a primary source, some things just wont exist and we will be trying to import from the main library@v3 and we will get a bunch of errors.

JacobCoffee:
I tried to push to your branch bunch its a fork and things are weird - so it created its own branch. It preserves all of your contributions/commit authorship so im going to close this in favor of that.
Also small nit not to create PRs off of your fork's `main` branch
(notes on that: https://hynek.me/articles/pull-requests-branch/)

PR #3395, opened 2024-04-16
refactor: CORS handling

peterschutt:
## Description

Applies three refactors to our CORS handling:

1. Move pre-flight response generation from the generated OPTIONS handlers to the CORS middleware.
2. Make CORS outermost middleware.
3. Deprecate CORS middleware from public API.

Fixes case where pre-flight response not applied to mounted ASGI apps. Previously, we would only generate OPTIONS handlers for HTTP routes, not mounts, so pre-flight requests would be passed through to the mounted ASGI app.

Fixes the case where a user defined OPTIONS handler may not include CORS logic. In fact, this is they only way we can be sure that we honor CORS pre-flight if the config object is provided.

Remove `ExceptionHandlerMiddleware`'s dependence on `CORSMiddleware` by making `CORSMiddleware` the outermost middleware in the stack if the app has CORS configured. This makes responses sent via the exception middleware pass through the CORS middleware send wrapper if CORS is configured, which is behaviourally equal to the current behavior of wrapping the `send` coro inside the exception handler middleware with the CORS middleware send wrapper.


litestar/middleware/cors.py
```diff
@@ -68,15 +78,55 @@ async def wrapped_send(message: Message) -> None:
...
-                allow_methods = headers.get("Access-Control-Allow-Methods")
-                if not allow_methods and self.config.allow_methods:
-                    headers["Access-Control-Allow-Methods"] = ", ".join(sorted(set(self.config.allow_methods)))
+                headers["Access-Control-Allow-Methods"] = ", ".join(sorted(set(self.config.allow_methods)))
 
             await send(message)
 
         return wrapped_send
+
+    def create_preflight_response(self, origin: str, request_headers: Headers) -> Response[str | None]:
```
JacobCoffee, line 89: is this meant to be private?

peterschutt: It's an interesting question, because while this class is public interface, there isn't really any way for users to customize the way it works. So IMO this whole class should be private.

Prior to this PR, there needed to be a `CORSConfig` passed to the application object in order for pre-flight responses to be generated:

https://github.com/litestar-org/litestar/blob/a0d11092dae300af4025d619d8ac818790dede74/litestar/routes/http.py#L258-L262

There is no way to only supply the `CORSMiddleware` to a subset of routes via a `Router.middleware` or similar, because of the fact that the pre-flight response generation is conditional upon the application having access to the `CORSConfig` object. In other words, if a user didn't supply the `CORSConfig` object to the application, but did supply `DefineMiddleware(CORSMiddleware, config=CORSConfig(...))` to a router, then the pre-flight responses for that router wouldn't have worked anyway because there's no `CORSConfig` on the app. And if the `CORSConfig` is added to the app, then there is no point adding the middleware to a sub-router, because we'd then apply it to the whole application anyway.

Also, the `CORSMiddleware` class is hard coded as a wrapper around the `ASGIRouter`:

https://github.com/litestar-org/litestar/blob/a0d11092dae300af4025d619d8ac818790dede74/litestar/app.py#L842-L844

There is no way for a user to sub-class the middleware to alter its behavior, as there is no way to pass that sub-classed middleware to us.

So IMO, `CORSMiddleware` is implementation detail - the whole thing should be undocumented and we should probably deprecate it from public use.

peterschutt: @litestar-org/maintainers, can I deprecate `CORSMiddleware` from our public interface based on :arrow_up:?

peterschutt:
Splitting this up.

PR #3400, opened 2024-04-18
refactor: Make CORS outermost middleware

peterschutt:
Remove `ExceptionHandlerMiddleware`'s dependence on `CORSMiddleware` by making `CORSMiddleware` the outermost middleware in the stack if the app has CORS configured.

This makes responses sent via the exception middleware pass through the CORS middleware send wrapper if CORS is configured, which is behaviourally equal to the current behavior of wrapping the `send` coro inside the exception handler middleware with the CORS middleware send wrapper.

## Description

-

### DIFF
--- a/litestar/middleware/exceptions/middleware.py
+++ b/litestar/middleware/exceptions/middleware.py
@@ -7,17 +7,12 @@ from sys import exc_info
 from traceback import format_exception
 from typing import TYPE_CHECKING, Any, Type, cast
 
-from litestar.datastructures import Headers
 from litestar.enums import MediaType, ScopeType
 from litestar.exceptions import HTTPException, LitestarException, WebSocketException
-from litestar.middleware.cors import CORSMiddleware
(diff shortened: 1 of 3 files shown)


litestar/middleware/exceptions/middleware.py
```diff
@@ -37,6 +32,8 @@
     )
     from litestar.types.asgi_types import WebSocketCloseEvent
 
+__all__ = ("ExceptionHandlerMiddleware", "ExceptionResponseContent", "create_exception_response")
+
 
 def get_exception_handler(exception_handlers: ExceptionHandlersMap, exc: Exception) -> ExceptionHandler | None:
     """Given a dictionary that maps exceptions and status codes to handler functions, and an exception, returns the
```
JacobCoffee, line 39: do we add a `.. versionchanged::` directive?

peterschutt: I don't _think_ so. AFICT, exception handler middleware is undocumented, and this shouldn't change the behavior of the application at all.

provinzkraut (approved):

PR #3420, opened 2024-04-23
fix: flash messages were not displayed on Redirect

### LINKED ISSUE #3325: bug: Plugin "Flash Messages" not working with "Redirect" Response
### Litestar (2.8.0)

```python
from litestar import Litestar, Request, get, post
from litestar.contrib.jinja import JinjaTemplateEngine
from litestar.plugins.flash import FlashConfig, FlashPlugin, flash
from litestar.response import Template, Redirect
from litestar.template.config import TemplateConfig

template_config = TemplateConfig(engine=JinjaTemplateEngine, directory="templates")
flash_plugin = FlashPlugin(config=FlashConfig(template_config=template_config))


@get('/')
async def index() -> Redirect:
    return Redirect('/login')


@get('/login')
async def login(request: Request) -> Template:
    flash(request, "Flash Test!", category="info")
    return Template(template_str='{% for f in get_flashes() %} <p>{{ f.message }}</p>{% endfor %}'
                                 '<form action="/check" method="post"><button type="submit">login</button></form>')


@post('/check')
async def check(request: Request) -> Redirect:
    flash(request, "User not Found!", category="warning")
    return Redirect('/login')


app = Litestar(plugins=[flash_plugin], route_handlers=[index,login, check], template_config=template_config)

if __name__ == '__main__':
    import uvicorn

    uvicorn.run('1:app')

```

```
INFO:     Started server process [13056]
INFO:     127.0.0.1:50961 - "GET / HTTP/1.1" 302 Found
INFO:     127.0.0.1:50961 - "GET /login HTTP/1.1" 200 OK
INFO:     127.0.0.1:50961 - "GET /login HTTP/1.1" 200 OK
INFO:     127.0.0.1:50961 - "POST /check HTTP/1.1" 302 Found
INFO:     127.0.0.1:50961 - "GET /login HTTP/1.1" 200 OK
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [13056]

Process finished with exit code 0

```

litestar

### Problem: Flash Message from "POST /check" is missing at final page (screenshot above)

---

### Flask (3.0.2)

```python
from flask import Flask, flash, render_template_string, redirect

app = Flask(__name__)
app.secret_key = b'123'


@app.get('/')
def index() -> redirect:
    return redirect('/login')


@app.get('/login')
def login():
    flash('Flash Test!')
    return render_template_string('{% for m in get_flashed_messages() %} <p>{{ m }}</p>{% endfor %}'
                                  '<form action="/check" method="post"><button type="submit">login</button></form>')


@app.post('/check')
def check():
    flash("User not Found!")
    return redirect('/login')


if __name__ == '__main__':
    app.run()

```

```
 * Serving Flask app '2'
 * Debug mode: off
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:5000
Press CTRL+C to quit
127.0.0.1 - - [06/Apr/2024 14:50:47] "GET / HTTP/1.1" 302 -
127.0.0.1 - - [06/Apr/2024 14:50:47] "GET /login HTTP/1.1" 200 -
127.0.0.1 - - [06/Apr/2024 14:50:48] "POST /check HTTP/1.1" 302 -
127.0.0.1 - - [06/Apr/2024 14:50:48] "GET /login HTTP/1.1" 200 -

```

flask

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3325">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3325/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

euri10:
## Description

- fixes #3325


litestar/plugins/flash.py
```diff
@@ -57,26 +61,19 @@ def on_app_init(self, app_config: AppConfig) -> AppConfig:
...
-
-    Args:
-        context: The context dictionary.
-
-    Returns:
-        The flash messages, if any.
-    """
-    scope_state = ScopeState.from_scope(_get_request_from_context(context).scope)
-    return scope_state.flash_messages
+    return (
```
Alc-Alc, line 75: just a nit, can this be simplified to `_get_request_from_context(context).session.pop("_messages", [])`?

litestar/plugins/flash.py
```diff
@@ -46,6 +48,8 @@ def on_app_init(self, app_config: AppConfig) -> AppConfig:
         Returns:
             The application configuration with the message callable registered.
         """
+        if self.config.session_config is None:
+            raise litestar.exceptions.ImproperlyConfiguredException("Flash messages require a session middleware.")
```
peterschutt, line 52:
```suggestion
        for mw in app_config.middleware:
            if isinstance(mw, SessionMiddleware):
                break
        else:
            raise litestar.exceptions.ImproperlyConfiguredException("Flash messages require a session middleware.")
```
My thinking here is that we won't need the session config object attached to the flash config.

Is it important that the messages are stored in a backend session?

litestar/plugins/flash.py
```diff
@@ -57,26 +61,19 @@ def on_app_init(self, app_config: AppConfig) -> AppConfig:
...
-    scope_state = ScopeState.from_scope(connection.scope)
-    scope_state.flash_messages.append({"message": message, "category": category})
+def flash(
+    request: Request,
+    message: Any,
+    category: str,
+) -> None:
+    if "_messages" not in request.session:
+        request.session["_messages"] = []
+    request.session["_messages"].append({"message": message, "category": category})
```
peterschutt, line 71:
```suggestion
    request.session.setdefault("_messages", []).append({"message": message, "category": category})
```
Just another nit.

peterschutt (commented):
Is the flash config's reference to the session config necessary? I.e., can we check for a session middleware instead?

Also, I understand that current behavior is broken WRT to redirects, but should we try to make this change in a more backward compatible manner? I.e., if a session middleware is not found, fall back to the current implementation and include a deprecation warning that a session middleware will be required in a future version.

litestar/plugins/flash.py
euri10: doing this change breaks the tests, the ImproperlyConfiguredException is always raised because I think by the time the plugin initializes the app doesnt know about middlewares yet. is that possible ?

euri10: I think it is important messages are stored in whatever survives the orginal request scope, so maybe a backend session is only one of the options I dont know, I tried a little with context vars but while this passes the test it's because it keeps adding on the messages so the tests are not that good if not paying attention (https://github.com/litestar-org/litestar/pull/3420/commits/fd55270dd9a80857c2f8f9d62571a8dca069734b)

peterschutt: > doing this change breaks the tests, the ImproperlyConfiguredException is always raised because I think by the time the plugin initializes the app doesnt know about middlewares yet. is that possible ?

I should have checked, the session config adds an instance of `DefineMiddleware`, so it would need to be `if isinstance(mw, DefineMiddleware) and isinstance(mw.middleware, SessionMiddleware):`

peterschutt: > I think it is important messages are stored in whatever survives the orginal request scope, so maybe a backend session is only one of the options I dont know, I tried a little with context vars but while this passes the test it's because it keeps adding on the messages so the tests are not that good if not paying attention (https://github.com/litestar-org/litestar/pull/3420/commits/fd55270dd9a80857c2f8f9d62571a8dca069734b)

Sorry, my question should have been more clear. Is there a reason that a client side session is unsuitable? By checking for presence of a session middleware, it could be configured either with client side or server side storage.

euri10: > if isinstance(mw, DefineMiddleware) and isinstance(mw.middleware, S

ok this is weird, the tests fail with this:
` isinstance(mw, DefineMiddleware) ` is True 
but ` isinstance(mw.middleware, SessionMiddleware)` is False

euri10: > > I think it is important messages are stored in whatever survives the orginal request scope, so maybe a backend session is only one of the options I dont know, I tried a little with context vars but while this passes the test it's because it keeps adding on the messages so the tests are not that good if not paying attention ([fd55270](https://github.com/litestar-org/litestar/commit/fd55270dd9a80857c2f8f9d62571a8dca069734b))
> 
> Sorry, my question should have been more clear. Is there a reason that a client side session is unsuitable? By checking for presence of a session middleware, it could be configured either with client side or server side storage.

I dont think there is any importance

euri10: would `issubclass(mw.middleware, SessionMiddleware)` be ok you think @peterschutt ?

peterschutt: Sorry, yeah my bad:

`if isinstance(mw, DefineMiddleware) and issubclass(mw.middleware, SessionMiddleware):`

euri10: this pass tests with it but I truly dont get how to fix that mypy error

peterschutt: B/c `DefineMiddleware.middleware` is typed as `Callable[..., ASGIApp]`, it doesn't have to be a class, but `issubclass()` will error if the thing we pass it isn't a class. So we have to check for that first. For this we have `litestar.utils.is_class_and_subclass()`.

I've just pushed a commit.

peterschutt: I've found there is one other case where session middleware can exist in the middleware stack without us detecting here: if a user has added `SessionAuth` to the application. Adding that results in a type of `security.session_auth.middleware.MiddlewareWrapper` middleware added to the stack, inside of `DefineMiddleware`. And inside `MiddlewareWrapper` it creates its own little asgi middleware stack that include `SessionMiddleware` :face_with_spiral_eyes: 

So, i think we have to do:
```py
if isinstance(mw, DefineMiddleware) and issubclass(mw.middleware, (SessionMiddleware, MiddlewareWrapper)):
```
... to actually ascertain if `SessionMiddleware` is in the stack.

I think `SessionAuth` should behave exactly like you are implementing here, that is, not add session middleware to the stack, but require it to be there already.

I created a development thread for this in discord, here: https://discord.com/channels/919193495116337154/1232602408119570473

euri10: I just added a test to up the coverage which dropped, this should be ok now I think

peterschutt: > I've found there is one other case where session middleware can exist in the middleware stack without us detecting here: if a user has added `SessionAuth` to the application.
> 
> That results in `security.session_auth.middleware.MiddlewareWrapper` added to the middleware stack, itself wrapped inside of `DefineMiddleware`.
> 
> Inside `MiddlewareWrapper` it creates its own little asgi middleware stack that includes `SessionMiddleware` 😵‍💫
> 
> So, i think we have to do:
> 
> ```python
> if isinstance(mw, DefineMiddleware) and issubclass(mw.middleware, (SessionMiddleware, MiddlewareWrapper)):
> ```
> 
> ... to actually ascertain if `SessionMiddleware` is in the stack.
> 
> I think `SessionAuth` should behave exactly like you are implementing here, that is, not add session middleware to the stack, but require it to be there already.
> 
> I created a development thread for this in discord, here: https://discord.com/channels/919193495116337154/1232602408119570473

Added a commit for this - we should be good to go now!

peterschutt (approved):

PR #3425, opened 2024-04-24
fix: logging to file with structlog

peterschutt:
## Description

PR fixes issue with converting `StructLoggingConfig` to dict during call to `configure()` when the config object has a custom logger factory that references a `TextIO` object, which cannot be pickled.

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import sys
 from abc import ABC, abstractmethod
-from dataclasses import asdict, dataclass, field, fields
+from dataclasses import dataclass, field, fields
 from importlib.util import find_spec
 from logging import INFO
 from typing import TYPE_CHECKING, Any, Callable, Literal, cast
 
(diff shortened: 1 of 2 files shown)


JacobCoffee (approved):

litestar/logging/config.py
```diff
@@ -313,7 +314,9 @@ def stdlib_json_serializer(value: EventDict, **_: Any) -> str:  # pragma: no cov
     return _msgspec_json_encoder.encode(value).decode("utf-8")
 
 
-def default_structlog_processors(as_json: bool = True) -> list[Processor]:  # pyright: ignore
+def default_structlog_processors(
+    as_json: bool = True, json_serializer: Callable[[Any], Any] = default_json_serializer
+) -> list[Processor]:  # pyright: ignore
```
peterschutt, line 319: To use the `WriteLoggerFactory` with a file, the processor chain needs to return a string. This just gives me a way to pass the json encoder in without needing to build another processor chain.

PR #3430, opened 2024-04-26
fix: regular handler under mounted app

### LINKED ISSUE #3429: Bug: regular handler under asgi mount path conflicts with routing to asgi app
### Description

If we have a regular handler mounted under an asgi mounted path, and the path of the regular handler is a non-prefix sub-string of the a request path the request will not get mounted to the asgi app.

I.e., if we have an asgi mounted on `"/"`, and a regular handler at `"/path"`, a request to `"/some/path"` does not get routed to the asgi app.

_No response_

### MCVE

```python

from litestar import Litestar, asgi, get

if TYPE_CHECKING:
    from litestar.types.asgi_types import Receive, Scope, Send


async def asgi_app(scope: Scope, receive: Receive, send: Send) -> None:
    assert scope["type"] == "http"
    await send({
        "type": "http.response.start",
        "status": 200,
        "headers": [
            (b"content-type", b"text/plain"),
            (b"content-length", b"%d" % len(scope["raw_path"])),
        ],
    })
    await send({
        "type": "http.response.body",
        "body": scope["raw_path"],
    })

asgi_handler = asgi("/", is_mount=True)(asgi_app)


@get("/path")
def get_handler() -> str:
    return "Hello, world!"


def test_regular_handler_under_mounted_asgi_app() -> None:
    app = Litestar(
        route_handlers=[asgi("/", is_mount=True)(asgi_app), get_handler],
        openapi_config=None,
        debug=True,
    )

        resp = client.get("/some/path")
        assert resp.content == b"/some/path"
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

main

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

peterschutt:
Fix an issue where a regular handler under a mounted asgi app would prevent a request from routing through the mounted application if the request path contained the path of the regular handler as a substring.

Closes #3429

## Description

-

### DIFF
--- a/tests/e2e/test_regular_handler_under_asgi_mount_path.py
+++ b/tests/e2e/test_regular_handler_under_asgi_mount_path.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from litestar import Litestar, asgi, get
+from litestar.testing import TestClient
+
+if TYPE_CHECKING:
+    from litestar.types.asgi_types import Receive, Scope, Send
+
+
+async def asgi_app(scope: Scope, receive: Receive, send: Send) -> None:
+    assert scope["type"] == "http"
+    await send(
+        {
+            "type": "http.response.start",
+            "status": 200,
+            "headers": [
+                (b"content-type", b"text/plain"),
+                (b"content-length", b"%d" % len(scope["raw_path"])),
+            ],
+        }
+    )
+    await send(
+        {
+            "type": "http.response.body",
+            "body": scope["raw_path"],
+        }
+    )
+
+
+asgi_handler = asgi("/", is_mount=True)(asgi_app)
+
+
+@get("/path")
+async def get_handler() -> str:
+    return "Hello, world!"
+
+
+def test_regular_handler_under_mounted_asgi_app() -> None:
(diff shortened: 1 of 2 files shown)


tests/e2e/test_regular_handler_under_asgi_mount_path.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    await send(
+        {
+            "type": "http.response.body",
+            "body": scope["raw_path"],
+        }
+    )
+
+
+asgi_handler = asgi("/", is_mount=True)(asgi_app)
```
Alc-Alc, line 32: was this meant to be added at line 42?

peterschutt: ahh yeah thanks, at one point I was considering adding multiple tests to the module.

Alc-Alc (approved):

Bobronium:
Just spend some time debugging this issue and almost finished filling a bug report, but when trying to reference the line I found in sources, discovered it has already been fixed. Nice!

And thank you.

PR #3436, opened 2024-04-27
feat!: Make route handlers functional decorators

provinzkraut:
Make route handler decorators functional decorators. 

This gets rids of the `__call__` method having to mutate internal state and lets us clean up things a bit.

It also makes for a nicer syntax and improved type/runtime checking of the handler classes, because they can now receive a `fn` as a required keyword argument, which gets rid of the `HandlerClass("/")(handler_fn)` pattern.


litestar/handlers/asgi_handlers.py
```diff
@@ -42,14 +43,15 @@ def __init__(
         """Initialize ``ASGIRouteHandler``.
 
         Args:
+            path: A path fragment for the route handler function or a list of path fragments. If not given defaults to
+                ``/``
+            fn: The handler function
```
JacobCoffee, line 48:
```suggestion
        Args:
            path: A path fragment for the route handler function or a list of path fragments. If not given defaults to
                ``/``
            fn: The handler function
            
                .. versionadded:: 3.0
```

litestar/handlers/asgi_handlers.py
```diff
@@ -101,4 +104,28 @@ async def handle(self, connection: ASGIConnection[ASGIRouteHandler, Any, Any, An
         await self.fn(scope=connection.scope, receive=connection.receive, send=connection.send)
 
 
-asgi = ASGIRouteHandler
+def asgi(
```
JacobCoffee, line 107: is this public/documented?

litestar/handlers/base.py
```diff
@@ -99,6 +100,7 @@ def __init__(
         Args:
             path: A path fragment for the route handler function or a sequence of path fragments. If not given defaults
                 to ``/``
+            fn: The handler function
```
JacobCoffee, line 103:
```suggestion
            fn: The handler function

                .. versionadded:: 3.0
```

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -44,6 +45,7 @@ def __init__(
         Args:
             path: A path fragment for the route handler function or a sequence of path fragments. If not given defaults
                 to ``/``
+            fn: The handler function
```
JacobCoffee, line 48:
```suggestion
            fn: The handler function

                .. versionadded:: 3.0
```

JacobCoffee (approved):

litestar/handlers/http_handlers/base.py
```diff
@@ -619,6 +620,16 @@ def _validate_handler_function(self) -> None:
...
             raise ImproperlyConfiguredException("'data' kwarg is unsupported for 'GET' request handlers")
 
+        if self.http_methods == {HttpMethod.HEAD}:
+            # we allow here File and File because these have special setting for head responses
+            return_annotation = self.parsed_fn_signature.return_type.annotation
+            if not (
+                return_annotation in {NoneType, None}
+                or is_class_and_subclass(return_annotation, File)
+                or is_class_and_subclass(return_annotation, ASGIFileResponse)
+            ):
```
peterschutt, line 630:
```suggestion
        if self.http_methods == {HttpMethod.HEAD} and not self.parsed_fn_signature.return_type.is_subclass_of(
            (NoneType, File, ASGIFileResponse)
        ):
```

litestar/handlers/http_handlers/base.py
```diff
@@ -285,7 +302,7 @@ def __init__(
         self.response_cookies: Sequence[Cookie] | None = narrow_response_cookies(response_cookies)
         self.response_headers: Sequence[ResponseHeader] | None = narrow_response_headers(response_headers)
 
-        self.sync_to_thread = sync_to_thread
+        self.has_sync_callable = has_sync_callable
```
peterschutt, line 305: We can remove the class-level annotation for this on line 133 now that we assign it here.

litestar/handlers/websocket_handlers/listener.py
```diff
@@ -248,10 +251,8 @@ def __call__(self, fn: AnyCallable) -> Self:
             },
         )
 
-        return super().__call__(
-            ListenerHandler(
-                listener=self, fn=fn, parsed_signature=parsed_signature, namespace=self.resolve_signature_namespace()
-            )
+        return ListenerHandler(
+            listener=self, fn=fn, parsed_signature=parsed_signature, namespace=self.resolve_signature_namespace()
```
peterschutt, line 255: Its been a while since I've had my head in this part of the code, so I could very well be forgetting something.

However, when this was first rolled out, I recall we had a context object that allowed us to delay resolution of things like the signature namespace until the handler was actually registered on the application. FWICT, this is called as part of `BaseRouteHandler.__init__()` flow, which will be called when a function is decorated, and at that point it doesn't have any reference to its owner. This means that calling any layer resolution methods is redundant, right?

I can see that this is just transferring logic from the `__call__()` method in the previous implementation, but how does the layer resolution work here?

peterschutt: For reference, this is what he had in place at that time:

https://github.com/litestar-org/litestar/blob/fb0e98a01b8e2980b7e32df957e3d56f3bdbf802/litestar/handlers/websocket_handlers/listener.py#L152-L153

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -145,4 +148,32 @@ async def handle(self, connection: WebSocket[Any, Any, Any]) -> None:
...
+    dependencies: Dependencies | None = None,
+    exception_handlers: dict[int | type[Exception], ExceptionHandler] | None = None,
+    guards: list[Guard] | None = None,
+    middleware: list[Middleware] | None = None,
+    name: str | None = None,
+    opt: dict[str, Any] | None = None,
+    signature_namespace: Mapping[str, Any] | None = None,
+    websocket_class: type[WebSocket] | None = None,
+    **kwargs: Any,
+) -> Callable[[AsyncAnyCallable], WebsocketRouteHandler]:
```
peterschutt, line 163: This need a docstring?

peterschutt (approved):
Love it :heart_eyes: 

I do have a question about how we are handling layer resolution in the websocket listener `_prepare_fn()` method. But I'm pretty sure that this, and the refactoring suggestions I've made, are for things that you've just moved around in the PR, not things you've changed.

Overall though, feels way more pythonic, and I like the simplified wording you've used in the docs in places.

litestar/handlers/websocket_handlers/listener.py
provinzkraut: > I can see that this is just transferring logic from the __call__() method in the previous implementation, but how does the layer resolution work here?

I think it doesn't.. I'll have to check this later, but this might be a bug.

peterschutt: > one that would also exist in the current implementation

Yep, absolutely not introduced here, I just happen to be reading this code now.

litestar/handlers/asgi_handlers.py
JacobCoffee: and a `.. versionadded::` directive :D

provinzkraut: Actually, I'm not so sure about the `versionadded` with these. The names aren't newly introduced, and their usage remains largely the same. The fact that they're now implemented as decorator functions instead of callable class instance is an implementation detail. I think it's fair to say this is merely a change and not addition of a new feature.

wdyt @JacobCoffee @peterschutt?

PR #3445, opened 2024-04-27
Kumzy/main

### LINKED ISSUE #3362: Docs: Move `.. code-block::` directives into `docs/examples/`

JacobCoffee:
## Description

- Moving all code-block to files

## Closes

#3362


JacobCoffee:
Still remaining @Kumzy we just need to fix any pre-commit failures 
```
pre-commit install
pre-commit run --all-files
```
and fix the ones in the new `docs/examples` dirs.

Basically most of them look like missing imports like
```
docs/examples/templating/template_file.py:1:2: F821 Undefined name `get`
```

JacobCoffee:
@litestar-org/maintainers when we have examples like this, how do we want to handle the moving of them? Should we *create* `my_app`? or do we ignore and allow the test to not run (and then go stale, and so then why have we moved all of these code blocks 🙃)

```
 /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
docs/examples/testing/test_request_factory_3.py:2: in <module>
    from my_app.guards import secret_token_guard
E   ModuleNotFoundError: No module named 'my_app'
```
```py
from os import environ

from my_app.guards import secret_token_guard

from litestar import get


@get(path="/secret", guards=[secret_token_guard], opt={"secret": environ.get("SECRET")})
def secret_endpoint() -> None: ...
```

provinzkraut:
> @litestar-org/maintainers when we have examples like this, how do we want to handle the moving of them? Should we _create_ `my_app`? or do we ignore and allow the test to not run (and then go stale, and so then why have we moved all of these code blocks 🙃)
> 
> ```
>  /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/importlib/__init__.py:126: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> docs/examples/testing/test_request_factory_3.py:2: in <module>
>     from my_app.guards import secret_token_guard
> E   ModuleNotFoundError: No module named 'my_app'
> ```
> 
> ```python
> from os import environ
> 
> from my_app.guards import secret_token_guard
> 
> from litestar import get
> 
> 
> @get(path="/secret", guards=[secret_token_guard], opt={"secret": environ.get("SECRET")})
> def secret_endpoint() -> None: ...
> ```

Examples should all be self-contained. There are rare exceptions (like templates / static files), but this isn't one of them; The guard could very well be defined in the same file.

Having an app for every code example also give the benefit of us being able to auto-run them which, even if the example isn't tested explicitly, at least implicitly ensures that we don't have broken examples.

Kumzy:
Had some question regarding latest changes to add import:
1. for channels, is it correct to fix the following with ``from litestar.channels import ChannelsPlugin as Channels`` and replace ``channels`` by ``Channels``
```python
subscriber = await channels.subscribe(["foo", "bar"])
try:
    ...  # do some stuff here
finally:
    await channels.unsubscribe(subscriber)
```

2. How should I fix the following, I have to create ctx_a ctx_b and hook_a and hook_b ?
```python
from litestar import Litestar

app = Litestar(lifespan=[ctx_a, ctx_b], on_shutdown=[hook_a, hook_b])
```

provinzkraut:
@JacobCoffee I don't think this can be salvaged anymore.. Feel free to re-open for v3, I think we should be able to cherry-pick most of the changes from here

euri10:
one suggestion maybe, I never used it but maybe using https://github.com/sphinx-contrib/sphinxcontrib-run might help ?

PR #3454, opened 2024-04-30
docs: improve sse

### LINKED ISSUE #3011: Docs: Document SSE
### Summary

The SSE documentation is currently lacking:

- Docs for `ServerSentEventMessage`
- Sending messages a dicts

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3011">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3011/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

euri10:
## Description

- fixes #3011

### DIFF
--- a/docs/examples/responses/sse_responses.py
+++ b/docs/examples/responses/sse_responses.py
@@ -1,11 +1,12 @@
 from asyncio import sleep
-from typing import Any, AsyncGenerator
+from typing import AsyncGenerator
 
 from litestar import Litestar, get
 from litestar.response import ServerSentEvent, ServerSentEventMessage
+from litestar.types import SSEData
 
 
-async def my_generator() -> AsyncGenerator[Any, None]:
+async def my_generator() -> AsyncGenerator[SSEData, None]:
     count = 0
     while count < 10:
         await sleep(0.01)
--- a/docs/examples/responses/sse_responses.py
+++ b/docs/examples/responses/sse_responses.py
@@ -2,15 +2,27 @@ from asyncio import sleep
 from typing import AsyncGenerator
 
 from litestar import Litestar, get
-from litestar.response import ServerSentEvent
+from litestar.response import ServerSentEvent, ServerSentEventMessage
 
 
-async def my_generator() -> AsyncGenerator[bytes, None]:
+async def my_generator() -> AsyncGenerator[ServerSentEventMessage, None]:
     count = 0
     while count < 10:
         await sleep(0.01)
         count += 1
+        # In the generator you can yield integers, strings, bytes, dictionaries, or ServerSentEventMessage objects
+        # dicts can have the following keys: data, event, id, retry, comment
+
+        # here we yield an integer
+        yield count
+        # here a string
         yield str(count)
+        # here bytes
+        yield str(count).encode("utf-8")
+        # here a dictionary
+        yield {"data": 2 * count, "event": "event2", "retry": 10}
+        # here a ServerSentEventMessage object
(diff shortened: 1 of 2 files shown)


Alc-Alc (approved):

docs/examples/responses/sse_responses.py
```diff
@@ -2,15 +2,27 @@
 from typing import AsyncGenerator
 
 from litestar import Litestar, get
-from litestar.response import ServerSentEvent
+from litestar.response import ServerSentEvent, ServerSentEventMessage
 
 
-async def my_generator() -> AsyncGenerator[bytes, None]:
+async def my_generator() -> AsyncGenerator[ServerSentEventMessage, None]:
```
peterschutt, line 8:
```suggestion
async def my_generator() -> AsyncGenerator[Any, None]:
```
The generator is yielding many different types in the example, so perhaps this makes sense?

peterschutt (approved):
Thanks @euri10!

docs/examples/responses/sse_responses.py
Alc-Alc: what about `ServerSentEvent`?

peterschutt: Is that a public type?

Alc-Alc: don't see `_`s so yes?

euri10: yep, forgot about this one thanks

peterschutt: I don't see it in the api docs though 🤷‍♂️

euri10: it's weird, it's not in the docs indeed but in https://github.com/litestar-org/litestar/blob/b9e501fba6ff2ab06ab3afb53d9071a0a8699c55/litestar/types/helper_types.py#L28

is there an issue with the api docs generation ?

Alc-Alc: I guess this has to be manually added (not automatic) in https://github.com/litestar-org/litestar/blob/b9e501fba6ff2ab06ab3afb53d9071a0a8699c55/docs/reference/types.rst?

peterschutt: Ref docs only include names listed in `__all__` - so I assume that's why this is.. and if this name isn't in `__all__` then that would mean it's not a public name.

We can add it though - but for the case of this example, is it really that helpful? A reader isn't going to know what the `SSEData` data type represents anyway.

Maybe we should define that type alias in the example itself, wdyt?

peterschutt: But it is in `__all__` 🫠

peterschutt: It seems we explicitly name things for the types namespace, and it's just not explicitly named:

https://github.com/litestar-org/litestar/blob/main/docs/reference/types.rst#L141-L150

peterschutt: > I guess this has to be manually added (not automatic) in https://github.com/litestar-org/litestar/blob/b9e501fba6ff2ab06ab3afb53d9071a0a8699c55/docs/reference/types.rst?

Oops missed this, yeah, this.

euri10: ok, so should I add it here or open another issue / pr to do that ? or revert to Any ?
I'm ok with both :)

Alc-Alc: I would probably do it in a new PR so more eyes are on it :)

JacobCoffee (approved):

PR #3456, opened 2024-04-30
docs: Refactor and add test for `paginated_return_data.py`

kedod:
## Description

- Refactor and add test for `paginated_return_data` doc example

Visible code shows the simpliest usage of the funcionality.
The `Full code` dropdown shows the full working example code with highlighted lines.

image
image


docs/usage/dto/1-abstract-dto.rst
```diff
@@ -352,9 +352,14 @@ Working with Litestar's Pagination Types
 Litestar offers paginated response wrapper types, and DTO Factory types can handle this out of the box.
 
 .. literalinclude:: /examples/data_transfer_objects/factory/paginated_return_data.py
-    :caption: Paginated Return Data
     :language: python
-    :linenos:
+    :lines: 9-11,26-40
```
peterschutt, line 356:
```suggestion
    :lines: 9-11,25-39
```

docs/usage/dto/1-abstract-dto.rst
```diff
@@ -352,9 +352,14 @@ Working with Litestar's Pagination Types
...
+
+.. dropdown:: Full code
```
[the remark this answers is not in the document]
peterschutt: I'm fine with Full Code - we should just be consistent everywhere. Have we done this anywhere else yet?

docs/usage/dto/1-abstract-dto.rst
```diff
@@ -352,9 +352,14 @@ Working with Litestar's Pagination Types
...
+
+    .. literalinclude:: /examples/data_transfer_objects/factory/paginated_return_data.py
+        :language: python
+        :emphasize-lines: 9,26-40
```
peterschutt, line 362:
```suggestion
        :emphasize-lines: 9,25-39
```

peterschutt (commented):
Thanks! I think the auto format might have thrown out the line numbers by one.

Do you think we should we use literal includes for the remaining code blocks in that section?

docs/examples/data_transfer_objects/factory/paginated_return_data.py
```diff
@@ -1,13 +1,15 @@
...
 
 from litestar import Litestar, get
+from litestar.contrib.sqlalchemy.base import CommonTableAttributes, UUIDPrimaryKey
 from litestar.contrib.sqlalchemy.dto import SQLAlchemyDTO
 from litestar.dto import DTOConfig
 from litestar.pagination import ClassicPagination
 
-from .my_lib import Base
+
+class Base(CommonTableAttributes, UUIDPrimaryKey, DeclarativeBase): ...
```
Alc-Alc, line 12: if you are using stuff from advanced alchemy, you might as well directly use `UUIDBase` and make things simpler

provinzkraut:
> im struggling with this because now we split into two places and increase maintenance burden around `:linenos:` upkeep but on the flip side I really hate scrolling through giant walls of code to get to the next section.
> 
> Aesthetically I would prefer the usage of `sphinx-togglebutton`. I think it looks a lot cleaner and is customizable. (https://shibuya.lepture.com/extensions/sphinx-togglebutton/)
> 
> Maybe we could standardize on just always showing full blocks of code, but if those are over a certain line count e.g., wont fit in half of the viewport or maybe the full length of the viewport then we chunk it into a `` .. dropdown::`` directive with ``:class: dropdown` to style it?

I have a quite strong opinions on the dropdowns in that I think they should almost always be avoided in reference docs when they're hiding code. The code is what these docs are talking about, so hiding it is simply obscuring one of the main pieces of information.

@JacobCoffee If you don't like scrolling, we should improve the section layout / side nav. But hiding code is the worst option IMO.

cofin (approved):

provinzkraut:
Closing as stale

PR #3467, opened 2024-05-04
feat: add pydantic's error dict to ValidationException

### LINKED ISSUE #3466: Enhancement: Add Pydantic's error dictionary to ValidationException's extra dict
### Summary

To send a custom message for Pydantic errors, we require the error `type`. Pydantic's error metadata is lost while building the error message in `SignatureModel._build_error_message`. If we add the `exc` dict to this message, it will be propagated to exception handlers

### Basic Example

```
"SignatureModel"
@classmethod
    def _build_error_message(
        cls,
        keys: Sequence[str],
        exc_msg: str,
        connection: ASGIConnection,
        exc: Optional[Dict[str, Any]] = None
    ) -> ErrorMessage:
    ...
    if exc:
        message["exc"] = exc
    ...
```
Then in an exception handler, Pydantic's error dict can be accessed by:
`validation_exception["extra"][0]["exc"]`

### Drawbacks and Impact

_No response_

### Unresolved questions

Is there a better way to propagate Pydantic's error object to `ValidationException` received by the handlers?

Anu-cool-007:
## Description

- Updated `SignatureModel`'s `_build_error_message` method to accept an optional error dict and add it to the message.
- This can be used to propagate the original error object to the error handlers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ __pypackages__/
 .tool-versions
 .python-version
 /.dmypy.json
+
+# test certificates
+certs/
--- a/docs/usage/cli.rst
+++ b/docs/usage/cli.rst
@@ -100,28 +100,33 @@ The ``run`` command executes a Litestar application using `uvicorn <https://www.
 Options
 ~~~~~~~
 
-+---------------------------------------+---------------------------------+----------------------------------------------------------------+
-|                 Flag                  |      Environment variable       |                          Description                           |
-+=======================================+=================================+================================================================+
-| ``-r``\ , ``--reload``                | ``LITESTAR_RELOAD``             | Reload the application when files in its directory are changed |
-+---------------------------------------+---------------------------------+----------------------------------------------------------------+
-| ``-R``\ ,``--reload-dir``             | ``LITESTAR_RELOAD_DIRS``        | Specify directories to watch for reload.                       |
-+---------------------------------------+---------------------------------+----------------------------------------------------------------+
-| ``-p``\ , ``--port``                  | ``LITESTAR_PORT``               | Bind the server to this port [default: 8000]                   |
-+---------------------------------------+---------------------------------+----------------------------------------------------------------+
-| ``--wc``\ , ``--web-concurrency``     | ``WEB_CONCURRENCY``             | The number of concurrent web workers to start [default: 1]     |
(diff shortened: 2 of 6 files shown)


litestar/_signature/model.py
```diff
@@ -61,6 +61,7 @@ class ErrorMessage(TypedDict):
     key: NotRequired[str]
     message: str
     source: NotRequired[Literal["body"] | ParamType]
+    exc: NotRequired[Dict[str, Any]]
```
guacs, line 64: I think a different name like `errors` or something may be better since `exc` is usually used to mean an exception. Wdyt?

Anu-cool-007: Good point. This will contain data of a single error and will be accessed with `validation_exception["extra"][0]["exc"]`.
How about `raw`, `detail` or `error_detail`?

Anu-cool-007:
> I'm not enthusiastic about adding more pydantic specific handling in our core, and wonder if there's not a way we can better manage this type of thing with our plugin architecture.
> 
> We have this `ExtendedMsgSpecValidationError` raised by the pydantic type decoders to communicate to our core error handling that the exception has extra detail that can be included in the error response. I wonder if it would be more generic to have the `ExtendedMsgSpecValidationError` type accept an `original_exception` parameter instead of the `errors` mapping, and let the pydantic plugin register an exception handler that can access the original pydantic exception and add the extra detail to the exception response. This way, other serialization plugins could leverage the same behavior.

This approach feels more accessible. I'll try to implement this.

peterschutt:
> > I'm not enthusiastic about adding more pydantic specific handling in our core, and wonder if there's not a way we can better manage this type of thing with our plugin architecture.
> 
> > 
> 
> > We have this `ExtendedMsgSpecValidationError` raised by the pydantic type decoders to communicate to our core error handling that the exception has extra detail that can be included in the error response. I wonder if it would be more generic to have the `ExtendedMsgSpecValidationError` type accept an `original_exception` parameter instead of the `errors` mapping, and let the pydantic plugin register an exception handler that can access the original pydantic exception and add the extra detail to the exception response. This way, other serialization plugins could leverage the same behavior.
> 
> 
> 
> This approach feels more accessible. I'll try to implement this.

Cool, feel free to share early thoughts and if you have any questions I'll do my best to try and answer them. Perhaps best to do this in another branch in case we hit a dead end and end up back here?

PR #3478, opened 2024-05-08
fix: logging middleware with multi-body response

### LINKED ISSUE #3477: Bug: Multi-body response incompatible with LoggingMiddleware
### Description

When using `ServerSentEvent` Responses with `StructlogPlugin`, the application errors. 

Preliminary research led me to https://github.com/litestar-org/litestar/blob/main/litestar/middleware/logging.py#L180, where `scope.state._ls_connection_state.log_context` does not have any values during second push of an `ServerSentEventMessage`. 

_No response_

### MCVE

```python
from asyncio import sleep
from collections.abc import AsyncGenerator

from litestar.plugins.structlog import StructlogPlugin
from litestar.response import ServerSentEvent, ServerSentEventMessage
from litestar.types import SSEData


async def my_generator() -> AsyncGenerator[SSEData, None]:
    count = 0
    while count < 10:
        await sleep(0.01)
        count += 1
        yield ServerSentEventMessage(event="something-with-comment", retry=1000, comment="some comment")


@get(path="/count", sync_to_thread=False)
def sse_handler() -> ServerSentEvent:
    return ServerSentEvent(my_generator())


app = Litestar(route_handlers=[sse_handler],
               plugins=[StructlogPlugin()])
```

### Steps to reproduce

1. Run the server above

```bash
litestar --app main:app run
```

2. Send a request

```bash
 curl localhost:8000/count
```

3. Observe error

### Screenshots

_No response_

### Logs

```bash
File "/.venv/lib/python3.11/site-packages/litestar/middleware/logging.py", line 194, in extract_response_data
     connection_state.log_context.pop(HTTP_RESPONSE_START),
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'http.response.start'
```

### Litestar Version

2.8.3

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3477">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3477/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

peterschutt:
Prevent logging middleware from failing with a KeyError when a response sends multiple "http.response.body" messages.

Closes #3477

## Description

-

### DIFF
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -191,8 +191,8 @@ class LoggingMiddleware(AbstractMiddleware):
         connection_state = ScopeState.from_scope(scope)
         extracted_data = self.response_extractor(
             messages=(
-                connection_state.log_context.pop(HTTP_RESPONSE_START),
-                connection_state.log_context.pop(HTTP_RESPONSE_BODY),
+                connection_state.log_context[HTTP_RESPONSE_START],
+                connection_state.log_context[HTTP_RESPONSE_BODY],
             ),
         )
         response_body_compressed = value_or_default(connection_state.response_compressed, False)
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -191,8 +191,10 @@ class LoggingMiddleware(AbstractMiddleware):
         connection_state = ScopeState.from_scope(scope)
         extracted_data = self.response_extractor(
             messages=(
+                # NOTE: we don't pop the start message from the logging context in case
+                #   there are multiple body messages to be logged
                 connection_state.log_context[HTTP_RESPONSE_START],
-                connection_state.log_context[HTTP_RESPONSE_BODY],
+                connection_state.log_context.pop(HTTP_RESPONSE_BODY),
             ),
         )
         response_body_compressed = value_or_default(connection_state.response_compressed, False)
@@ -224,6 +226,10 @@ class LoggingMiddleware(AbstractMiddleware):
             elif message["type"] == HTTP_RESPONSE_BODY:
                 connection_state.log_context[HTTP_RESPONSE_BODY] = message
                 self.log_response(scope=scope)
+
+                if not message["more_body"]:
+                    connection_state.log_context.clear()
+
             await send(message)
 
         return send_wrapper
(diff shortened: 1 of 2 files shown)


litestar/middleware/logging.py
```diff
@@ -191,8 +191,8 @@ def extract_response_data(self, scope: Scope) -> dict[str, Any]:
         connection_state = ScopeState.from_scope(scope)
         extracted_data = self.response_extractor(
             messages=(
-                connection_state.log_context.pop(HTTP_RESPONSE_START),
-                connection_state.log_context.pop(HTTP_RESPONSE_BODY),
+                connection_state.log_context[HTTP_RESPONSE_START],
+                connection_state.log_context[HTTP_RESPONSE_BODY],
```
peterschutt, line 195: I don't think we have a real need to clear this context out. Firstly, we need the "http.response.start" message to live for as long as we are logging body messages to prevent #3477. And secondly, for every new body message sent, we overwrite the "http.response.body" message in the log context.

The scope will only live as long as the connection, and I'm pretty sure there's no way we could introduce a reference cycle by storing these messages, so I think safe enough to just leave them there.

provinzkraut: > and I'm pretty sure there's no way we could introduce a reference cycle by storing these messages

Seems reasonable, but do you see an easy way to test this, just to be sure?

peterschutt: I'll have a think about it..

The ASGI message would have to have a reference to the scope, or the ScopeState instance, or the log context dict itself to introduce a reference cycle, right?

peterschutt: Probably something we should always be wary of with ScopeState. I.e., if we put something in ScopeState that references the scope where the scope state itself is stored, it would be bad for business.

provinzkraut: > The ASGI message would have to have a reference to the scope, or the ScopeState instance, or the log context dict itself to introduce a reference cycle, right?

Yeah. I'm just not 100% sure this can't happen somewhere. I'll give it a think as well

JacobCoffee (approved):

PR #3484, opened 2024-05-10
chore: upgrade deps and tools

peterschutt:
- Addresses violations of new "safe-super" mypy check
- Addresses rich-click deprecation for importing `patch` from `rich_click.cli`
- Fixes test failure caused by `isinstance(command, Group)` no longer holding true after update to rich-click 1.8

## Description

-

### DIFF
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -252,8 +252,8 @@ def _inject_args(func: Callable[P, T]) -> Callable[P, T]:
 
 def _wrap_commands(commands: Iterable[Command]) -> None:
     for command in commands:
-        if isinstance(command, Group):
-            _wrap_commands(command.commands.values())
+        if hasattr(command, "commands"):
+            _wrap_commands(command.commands.values())  # pyright: ignore[reportGeneralTypeIssues]
         elif command.callback:
             command.callback = _inject_args(command.callback)
 
--- a/litestar/contrib/minijinja.py
+++ b/litestar/contrib/minijinja.py
@@ -105,7 +105,6 @@ class MiniJinjaTemplateEngine(TemplateEngineProtocol["MiniJinjaTemplate", StateP
             directory: Direct path or list of directory paths from which to serve templates.
             engine_instance: A Minijinja Environment instance.
         """
-        super().__init__(directory, engine_instance)
         if directory and engine_instance:
             raise ImproperlyConfiguredException(
                 "You must provide either a directory or a minijinja Environment instance."
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: unasyncd
         additional_dependencies: ["ruff"]
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.3.5"
+    rev: "v0.4.4"
     hooks:
       - id: ruff
         args: ["--fix"]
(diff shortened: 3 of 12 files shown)


litestar/cli/_utils.py
```diff
@@ -252,7 +252,7 @@ def wrapped(ctx: Context, /, *args: P.args, **kwargs: P.kwargs) -> T:
 
 def _wrap_commands(commands: Iterable[Command]) -> None:
     for command in commands:
-        if isinstance(command, Group):
```
peterschutt, line 255: This test failed after upgrading rich-click: https://github.com/litestar-org/litestar/blob/56b7395fa75a304dfea5711fe62646c469ce8616/tests/unit/test_cli/test_cli.py#L68

Issue is that, even though we import `Group` from `click`, it is actually `rich_click.patch._PatchedRichGroup`, and where `command` is an instance of `rich_click.rich_command.RichGroup`, `isinstance(command, Group)` returns `False`, meaning that the inner commands of groups weren't being wrapped to inject args.

This change seems to work, but I'm not familiar enough with click to know if there is a better way. @litestar-org/maintainers

provinzkraut: Seems like an okay fix 👍

litestar/contrib/minijinja.py
```diff
@@ -105,7 +105,6 @@ def __init__(self, directory: Path | list[Path] | None = None, engine_instance:
             directory: Direct path or list of directory paths from which to serve templates.
             engine_instance: A Minijinja Environment instance.
         """
-        super().__init__(directory, engine_instance)
```
provinzkraut, line 108: Why are these removed? Iiuc this could cause issues when using these classes in multiple inheritance

provinzkraut (approved):
Approved with a question

litestar/contrib/minijinja.py
peterschutt: Mypy says it's unsafe: https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-call-to-an-abstract-method-via-super-is-valid-safe-super

> Iiuc this could cause issues when using these classes in multiple inheritance

Can you demo this for me please?

peterschutt: Wouldn't the protocol `__init__` methods need to call super for that scenario too?

Alc-Alc: I *think* these only happen when they are used as "mixins" in which case I would assume the protocol has to delegate to the super  `__init__`, we can maybe mark is as `final` indicating this should not be inherited?

peterschutt: Shouldn't `MiniJinjaTemplateEngine` inherit from the protocol in this scenario?

Alc-Alc: thank you, edited, so

>  the protocol __init__ methods need to call super for that scenario too?
seems like it should

provinzkraut: > Wouldn't the protocol `__init__` methods need to call super for that scenario too?

Ah yeah, good point. Uh. I guess the question is how multiple inheritance works with protocols? Actually I'm not sure how they behave at runtime.

provinzkraut: Okay, I checked. They don't seem to call `super().__init__`, so we'd have to add that explicitly to make them work with multi-inheritance anyway. In that case I'm fine with this change as-is.

JacobCoffee (approved):

JacobCoffee:
> safe-super

PR #3491, opened 2024-05-11
chore: unpin Advanced Alchemy

cofin:
## Description

This PR unpins Advanced Alchemy.

This allows version `0.9.3` to be installed with teh `litestar[sqlalchemy]` command group.

This enables the following:
- A numebr of repository and service optimizations and fixes
- SQLAlchemyQueryRepository from advanced alchemy
- `to_schema` from the `Service` used there (see the fullstack repository for usage details)

However, this AA release also had to integrate it's own copies of the Litesatr `OffsetPagination` dataclass.  Litestar will attempt to use this version of the dataclass when a supported version is installed.  This is implemented similar to how the filters in `filters.py` work currently.

### DIFF
--- a/litestar/plugins/sqlalchemy.py
+++ b/litestar/plugins/sqlalchemy.py
@@ -17,7 +17,7 @@ def __getattr__(attr_name: str) -> object:
     }:
         warn_deprecation(
             deprecated_name=f"litestar.plugins.sqlalchemy.{attr_name}",
-            version="2.8.4",
+            version="2.9.0",
             kind="import",
             removal_in="3.0",
             info=f"importing {attr_name} from 'litestar.plugins.sqlalchemy' is deprecated, please"
--- a/litestar/plugins/sqlalchemy.py
+++ b/litestar/plugins/sqlalchemy.py
@@ -1,54 +1,31 @@
-from advanced_alchemy import filters, types
-from advanced_alchemy.base import (
-    AuditColumns,
-    BigIntAuditBase,
-    BigIntBase,
-    BigIntPrimaryKey,
-    CommonTableAttributes,
-    UUIDAuditBase,
-    UUIDBase,
-    UUIDPrimaryKey,
-    orm_registry,
-)
-from advanced_alchemy.extensions.litestar import (
-    AlembicAsyncConfig,
-    AlembicCommands,
-    AlembicSyncConfig,
-    AsyncSessionConfig,
-    EngineConfig,
-    SQLAlchemyAsyncConfig,
-    SQLAlchemyDTO,
-    SQLAlchemyDTOConfig,
-    SQLAlchemyInitPlugin,
-    SQLAlchemyPlugin,
-    SQLAlchemySerializationPlugin,
-    SQLAlchemySyncConfig,
-    SyncSessionConfig,
-)
+from litestar.utils import warn_deprecation
 
-__all__ = (
-    "orm_registry",
-    "filters",
-    "types",
-    "AuditColumns",
-    "BigIntAuditBase",
-    "BigIntBase",
-    "UUIDAuditBase",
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -112,7 +112,16 @@ include = [
...
-ignore_package_warnings = ["sphinx", "slotscheck"]
+ignore_package_warnings = [
+  "alabaster",
+  "sphinxcontrib-*",
+  "sphinx-*",
+  "sphinx",
+  "pre-commit",
+  "pydata-sphinx-*",
+  "slotscheck",
+  "autobahn","accessible-pygments"
```
cofin, line 123: My console is a bit noisy without these suppressions

tests/unit/test_repository/models_bigint.py
```diff
@@ -5,8 +5,8 @@
 from datetime import date, datetime
 from typing import List
 
-from advanced_alchemy import SQLAlchemyAsyncRepository, SQLAlchemySyncRepository
 from advanced_alchemy.base import BigIntAuditBase, BigIntBase
+from advanced_alchemy.repository import SQLAlchemyAsyncRepository, SQLAlchemySyncRepository
```
cofin, line 9: in the latest AA, we no longer export everything at the top level of the package.

peterschutt:
For the 3.8 failure, you'd either have to update the AA pagination class to use the typing generic collection types (ie `List` not `list`), or restrict AA to a lower version for python 3.8 installs.

cofin:
Thanks - Just saw this as I came to the same conclusions.

Resolved in https://github.com/litestar-org/advanced-alchemy/pull/191

pyproject.toml
peterschutt: After running `make lint`?

litestar/plugins/sqlalchemy.py
```diff
@@ -1,54 +1,31 @@
...
+        "BigIntPrimaryKey",
+        "CommonTableAttributes",
+        "UUIDAuditBase",
+        "UUIDBase",
+        "UUIDPrimaryKey",
+        "orm_registry",
+    }:
+        warn_deprecation(
+            deprecated_name=f"litestar.plugins.sqlalchemy.{attr_name}",
+            version="2.8.4",
```
peterschutt, line 20: We've been holding back new deprecations for minor releases - there was at least one PR that we kept out of 2.8.3 that was only a refactor, but included a new depreciation, so we held it back for 2.9 - this would fall in the same category.

litestar/plugins/sqlalchemy.py
```diff
@@ -1,54 +1,31 @@
...
+            kind="import",
+            removal_in="3.0",
+            info=f"importing {attr_name} from 'llitestar.plugins.sqlalchemy' is deprecated, please"
```
peterschutt, line 23:
```suggestion
            info=f"importing {attr_name} from 'litestar.plugins.sqlalchemy' is deprecated, please"
```

pyproject.toml
```diff
@@ -112,7 +112,17 @@ include = [
...
+  "alabaster",
+  "sphinxcontrib-*",
+  "sphinx-*",
+  "sphinx",
+  "pre-commit",
+  "pydata-sphinx-*",
+  "slotscheck",
+  "autobahn",
+  "accessible-pygments"
+]
```
peterschutt, line 125: What command do you run that produces these warnings?

cofin: just the normal `make install`.  I get a warning about these libraries not having 3.8 support.

litestar/plugins/sqlalchemy.py
cofin: Thanks.  Updated.

peterschutt (approved):

PR #3506, opened 2024-05-17
fix: print warning if incorrect annotations for return_dto

### LINKED ISSUE #3485: Bug: `return_dto` is silently ignored if return data type does not match DTO definition
### Description

Below is a minimal example that showcases the issue.

The context in which this came up for me is that I'm trying to port an existing app which uses FastAPI/SQLModel to Litestar. As illustrated in the SQLModel docs (for example [here](https://sqlmodel.tiangolo.com/tutorial/fastapi/multiple-models/#multiple-models-with-inheritance)), I have several pydantic models for read/write/db operations - most of which I would like to ditch in favour of Litestar's DTOs. I accidentally defined my return DTOs using the wrong pydantic model and things weren't working as expected - it would simply ignore the DTO and return the full, unfiltered data.

Since I'm new to the concept of DTOs it took me a while to figure out what the cause of the issue was. For a new user like me it would be very helpful to have Litestar print a prominent warning (if not abort with an error).

_No response_

### MCVE

```python

class MyMessage(BaseModel):
    msg: str
    priority: int


class MyMessage2(BaseModel):
    msg: str
    priority: int


class ReadDTO(PydanticDTO[MyMessage]):
    config = DTOConfig(exclude={"priority"})


@get("/", return_dto=ReadDTO)
async def hello_world() -> MyMessage:
    return MyMessage(msg="Hello world", priority=1)


```

### Steps to reproduce

```bash
1. Save the above MCVE as `app.py` and run it via `litestar run`
2. `curl http://127.0.0.1:8008` returns the expected output: `{"msg":"Hello world"}` (note that the `priority` field is correctly stripped by the `return_dto`).
3. In the definition of `ReadDTO`, change `PydanticDTO[MyMessage]` to `PydanticDTO[MyMessage2]`.
4. `curl http://127.0.0.1:8008` now returns the _full_ dictionary including the `priority` field: `{"msg": "Hello world", "priority": 1}`.

This is of course due to the fact that the `hello_world` handler constructs an instance of `MyMessage` while the return DTO is parameterised on `MyMessage2`.

However, there is no warning or any indication that there is a mismatch. Instead, the return DTO is just silently ignored and the full data is returned.
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

```
$ litestar version
2.8.3final0
```

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

tibor-reiss:
## Description
Print a warning during initialization if types do not match.

Due to the caching I think this is the only possibility where it's possible to do it.

`self.app.logger` is not available yet, so I used warnings.warn. For the latter, stacklevel is not the best, another option would be to use warnings.warn_explicit.

## Closes
#3485

### DIFF
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -23,6 +23,7 @@ from litestar.typing import FieldDefinition
 from litestar.utils import ensure_async_callable, get_name, normalize_path
 from litestar.utils.helpers import unwrap_partial
 from litestar.utils.signature import ParsedSignature, add_types_to_signature_namespace
+from litestar.utils.warnings import warn_return_dto_type_mismatch
 
 if TYPE_CHECKING:
     from typing_extensions import Self
@@ -500,12 +501,16 @@ class BaseRouteHandler:
             else:
                 return_dto = self.resolve_data_dto()
 
-            if return_dto and return_dto.is_supported_model_type_field(self.parsed_return_field):
-                return_dto.create_for_field_definition(
-                    field_definition=self.parsed_return_field,
-                    handler_id=self.handler_id,
-                )
-                self._resolved_return_dto = return_dto
+            if return_dto:
+                if return_dto.is_supported_model_type_field(self.parsed_return_field):
+                    return_dto.create_for_field_definition(
+                        field_definition=self.parsed_return_field,
+                        handler_id=self.handler_id,
+                    )
+                    self._resolved_return_dto = return_dto
+                else:
+                    warn_return_dto_type_mismatch(return_dto, self.parsed_return_field.raw)
+                    self._resolved_return_dto = None
             else:
                 self._resolved_return_dto = None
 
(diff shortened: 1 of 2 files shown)


peterschutt:
This isn't a bug, as DTOs can be defined on layers, and it would be quite restrictive to only allow routes beneath a layer that has a dto defined, to only return that DTO supported type. For example:

```py
    return_dto = SomeDTO[SomeType]

    @get("/{id:int}")
    def get_handler(self) -> SomeType: ...

    @get("/")
    def list_handler(self) -> list[SomeType]: ...

    @delete("/{id:int}")
    def delete_handler(self) -> None: ...
```
In that example, the delete handler not returning a model instance shouldn't force the user to have to define the dto individually on all of the other handlers.

I can see some utility to raising an error if a dto is defined directly on a handler and that handler doesn't utilise it, but that implementation will look different to the one you have defined here. Also, it should consider the data parameter.

PR #3511, opened 2024-05-18
fix: use re.match instead of re.search for mounted app path (#3501)

### LINKED ISSUE #3501: Bug: mounted app path interferes with regular paths
### Description

According to ["Mounting ASGI apps"](https://docs.litestar.dev/latest/usage/routing/overview.html#mounting-asgi-apps) documentation section, Litestar can mount ASGI apps in **sub-paths**. So it is expected that if ASGI app is mounted with `path='/magic'`, every route _starting with `/magic`_ will be handled by the ASGI app, and any other route will be handled by other handlers. However, it is not true.

Imagine this setup:

```python
@asgi("/magic", is_mount=True)
async def xxx(...):
    print('Mounted')
    ...

@get("/{number:int}/magic/")
async def yyy() -> str:
    print('Parametrized')

@get("/static/magic/")
async def zzz() -> str:
    print('Static')
```

Here's "expectations VS reality" table:
| Request path  | Expected output | Real output |
| -------------------- | ------------------------ | ------------------|
| `/magic`           | `Mounted`           | `Mounted`     |
|`/123/magic/`   | `Parametrized`   | `Mounted`     |
|`/static/magic/`| `Static`                  | `Static`    |
|`/non-existent/magic/` | 404 error | `Mounted` |

## Why this happens?

`litestar/_asgi/routing_trie/traversal.py:parse_path_to_route` method has [this line](https://github.com/litestar-org/litestar/blob/main/litestar/_asgi/routing_trie/traversal.py#L139):
```
if mount_paths_regex and (match := mount_paths_regex.search(path)):
```
So instead of **matching** `/magic` to `path`, `re.search` is used which searches for occurrence of `/magic` anywhere in `path`, thus resulting in "false positives" for strings such as `/123/magic/`, `/non-existent/magic/` and `/non-existent/magic/something`.

## Possible solution

This cannot be solved by simply using regex:
```python
@asgi("^/magic", is_mount=True)
```
since `mount_paths_regex` becomes `re.compile('/^/magic')`, so it not only doesn't solve the problem, but the `/magic` endpoint itself stops working.

I believe it may be solved by replacing `mount_paths_regex.search(path)` with `mount_paths_regex.match(path)` - I did manual tests and it solved the problem completely, but ofc such a change requires tests to ensure nothing else is broken.

I am ready to create a full-fledged pull request with tests once the issue is approved :)

_No response_

### MCVE

```python
from typing import Any

from litestar import Litestar, asgi, get
from litestar.response.base import ASGIResponse


@asgi("/magic", is_mount=True)
async def mounted_handler(scope: Any, receive: Any, send: Any) -> None:
    body = 'mounted!'
    response = ASGIResponse(body=body.encode("utf-8"))
    await response(scope, receive, send)


@get("/{number:int}/magic/")
async def parametrized_handler() -> str:
    return 'parametrized'


@get("/static/magic/")
async def static_handler() -> str:
    return 'static'


app = Litestar(route_handlers=[
    mounted_handler,
    parametrized_handler,
    static_handler,
])
```

### Steps to reproduce

```bash
1. Use the source code from MCVE to run Litestar app
2. Run curl to see wrong handler invoked for parametrized path:
  > curl http://127.0.0.1:8000/123/magic
  mounted!     
3. Run curl to see wrong handler invoked for non-existent path:
  > curl http://127.0.0.1:8000/whatever/magic
  mounted!
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.8.3

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

0xE111:
## Description
See #3501 for detailed description.

TL;DR: When mounting an app, path resolution uses `re.search` instead or `re.match`, thus mounted app matches any path which contains mount path.

Original code used `re.search` and had explicit "sub-path match" (`path[match.start() : match.end()]`) which makes me think there was some reasoning behind `re.search` but tbh I don't get it:

```python
if mount_paths_regex and (match := mount_paths_regex.search(path)):
    mount_path = path[match.start() : match.end()]
```

## Closes
Fixes #3501


peterschutt:
> Original code used `re.search` and had explicit "sub-path match" (`path[match.start() : match.end()]`) which makes me think there was some reasoning behind `re.search` but tbh I don't get it:

I'm trying to figure this out for myself - its almost as though it is designed to match against a sub-path and then extract the mount path component.. but why.. I'm not sure.. I'll spend more time trying to figure it out. On the plus side, your implementation doesn't break any tests - so maybe it is just a weirdness that exists due to code evolution or whatever, but I'd like to take a bit of time to think it through.

peterschutt:
> Original code used `re.search` and had explicit "sub-path match" (`path[match.start() : match.end()]`) which makes me think there was some reasoning behind `re.search` but tbh I don't get it:
> 
> ```python
> if mount_paths_regex and (match := mount_paths_regex.search(path)):
>     mount_path = path[match.start() : match.end()]
> ```

@provinzkraut My best guess with this is that it was required before we started handling `root_path` in https://github.com/litestar-org/litestar/pull/3039. That is, before that PR, the path we matched the mount path against, may still have had the root path prepended to it, and so indexing both ends of the slice would have effectively stripped the root path from the path we pass through to the mounted app.. this is my best guess.. any thoughts?

peterschutt (approved):
Thanks @0xE111 - I'm just waiting on another set of eyes on that path slicing business before we merge.

provinzkraut (approved):

provinzkraut:
> @provinzkraut My best guess with this is that it was required before we started handling `root_path` in #3039. That is, before that PR, the path we matched the mount path against, may still have had the root path prepended to it, and so indexing both ends of the slice would have effectively stripped the root path from the path we pass through to the mounted app.. this is my best guess.. any thoughts?

Can't come up with a better explanation, so I'd say it's okay to merge this as is!

provinzkraut:
@all-contributors add @0xE111 for code

PR #3526, opened 2024-05-25
feat: allow creating parent directories for a file store

Murtagy:
## Description

Allow mkdir True when creating a file store.

Use case is rather small - initially deploying an application with FileStores used.


Murtagy:
@all-contributors please add @Murtagy

litestar/stores/file.py
```diff
@@ -88,7 +88,7 @@ async def set(self, key: str, value: str | bytes, expires_in: int | timedelta |
             ``None``
         """
 
-        await self.path.mkdir(exist_ok=True)
+        await self.path.mkdir(exist_ok=True, parents=True)
```
Alc-Alc, line 91: if this change is being introduced, would it be better to make it configurable by a bool in `FileStore.__init__`?

provinzkraut: Yes. This absolutely needs to be configurable and set to `False` by default. Otherwise it's not only a breaking change, but can also lead to nasty side effects

provinzkraut (changes requested):
Thanks @Murtagy for the PR :)

In general we can add this but:

1. It should be configurable
2. It should happen once during store startup, in the store's `__aenter__`

litestar/stores/file.py
Murtagy: Why creating 1 folder is fine, but creating 2 has to be configurable. The "allow_create_folders" flag makes sense to me. "allow_create_more_than_1_folder" makes no sense to me.
Or I misunderstand the flag's purpose?

Murtagy:
> Thanks @Murtagy for the PR :)
> 
> In general we can add this but:
> 
> 1. It should be configurable
> 2. It should happen once during store startup, in the store's `__aenter__`

1. Why creating 1 folder is fine, but creating 2 has to be configurable. The "allow_create_folders" flag makes sense to me. "allow_create_more_than_1_folder" makes no sense to me.
Or I misunderstand the flag's purpose?
2. I am a little slow in understanding the idea of scopes and the control flow of the framework, so done as a knee-jerk reaction to the exception, in the place where exception is made. Will look into it...

Alc-Alc:
> > Thanks @Murtagy for the PR :)
> > In general we can add this but:
> > 
> > 1. It should be configurable
> > 2. It should happen once during store startup, in the store's `__aenter__`
> 
> 1. Why creating 1 folder is fine, but creating 2 has to be configurable. The "allow_create_folders" flag makes sense to me. "allow_create_more_than_1_folder" makes no sense to me.
>    Or I misunderstand the flag's purpose?
>    UPD: don't mind me, I for some reason thought that 1 dir is made by default, which is not the case
> 2. I am a little slow in understanding the idea of scopes and the control flow of the framework, so done as a knee-jerk reaction to the exception, in the place where exception is made. Will look into it...

You just need one flag `create_folders: bool = False`. You can do `await self.path.mkdir(exist_ok=True, parents= self.create_folders)`. The choice of naming is up in the air, I am impartial as long as it's self descriptive.

litestar/stores/file.py
cofin: Please follow the guidance given here.  It is definitely a breaking change and has security implications.

Murtagy:
@provinzkraut (and/or others 😄), did this change in correct direction?

litestar/stores/file.py
```diff
@@ -28,15 +28,22 @@ def _safe_file_name(name: str) -> str:
 class FileStore(NamespacedStore):
     """File based, thread and process safe, asynchronous key/value store."""
 
-    __slots__ = {"path": "file path"}
+    __slots__ = {"path": "file path", "_create_folders": "flag to create folders of path"}
```
JacobCoffee, line 31: is this meant to be `create_folders` instead?

litestar/stores/file.py
```diff
@@ -28,15 +28,22 @@ def _safe_file_name(name: str) -> str:
...
 
-    def __init__(self, path: PathLike[str]) -> None:
+    def __init__(self, path: PathLike[str], *, create_folders: bool = False) -> None:
         """Initialize ``FileStorage``.
 
         Args:
             path: Path to store data under
+            create_folders: bool, if ``True``, create the folders in ``path`` if they don't exist
```
JacobCoffee, line 38:
```suggestion
            create_folders: Create the folders in ``path`` if they don't exist
                Default: ``False``

                .. versionadded:: 2.9.0
```

provinzkraut (changes requested):
Looks good @Murtagy, just one more thing I'm afraid :)
I'd prefer if we change the naming from "folders" to "directories". This is the more commonly used term and is also used throughout Litstar.

litestar/stores/file.py
Murtagy: Actually I see the dictionary in slots for the first time ever, so I don't really know. I used private attribute notation, and expected the slot entry to match the attribute name `self._create_folders`

Murtagy: Doesn't this duplicate the docstr overall?

provinzkraut (approved):

Murtagy:
`codecov` was correctly highlighting that the line I added did not run - the store was not going through context manager. The test succeeded, because the directory was created by other directories, so the parent folder already existed. 
Fixed that, but overall not sure if the test harness I made is good, I did not work with pytest before

tests/conftest.py
```diff
@@ -92,6 +93,18 @@ def file_store(tmp_path: Path) -> FileStore:
     return FileStore(path=tmp_path)
 
 
+@pytest.fixture()
+def file_store_create_directories(tmp_path: Path) -> FileStore:
+    shutil.rmtree(tmp_path, ignore_errors=True)  # in case the path was already created by different tests - we clean it
+    return FileStore(path=tmp_path.joinpath("subdir"), create_directories=True)
```
provinzkraut, line 99: I don't think this is the correct way to do this. Instead, you should create a path under `tmp_path` that has at least to levels, e.g. `tmp_path / "subdir_1" / "subdir_2". That should have the same effect, without you having to mess with pytest's `tmp_path`

litestar/stores/file.py
```diff
@@ -28,15 +28,25 @@ def _safe_file_name(name: str) -> str:
 class FileStore(NamespacedStore):
     """File based, thread and process safe, asynchronous key/value store."""
 
-    __slots__ = {"path": "file path"}
+    __slots__ = {"path": "file path", "create_directories": "flag to create directories of path"}
```
euri10, line 31:
```suggestion
    __slots__ = {"path": "file path", "create_directories": "flag to create directories in path"}
```

litestar/stores/file.py
```diff
@@ -28,15 +28,25 @@ def _safe_file_name(name: str) -> str:
...
+                Default: ``False``
+
+                .. versionadded:: 2.9.0
         """
         self.path = Path(path)
+        self.create_directories = create_directories
+
+    async def __aenter__(self) -> None:
+        if self.create_directories:
+            await self.path.mkdir(exist_ok=True, parents=True)
```
euri10, line 48: is it just me or it's the 1st time I see an await in front of pathlib mkdir ?

Murtagy:
image
How do I get these running?

Murtagy:
I think I have addressed all the comments and successfully restarted tests via merging master. 
Are we good to go?

PR #3530, opened 2024-05-27
docs: updated example for flash messages

cofin:
## Description

The flash messages plugin requires a valid session configuration.  This PR adds a basic configuration to the examples.  Previously, an exception would be raised when executing the examples.


Alc-Alc (approved):
`docs: ` not `chore(docs)` 😛

PR #3538, opened 2024-06-02
fix: add OPTIONS to the default safe methods for CSRFConfig

wer153:
## Description

- make OPTIONS the default safe method for CSRFConfig
- related [doc](https://developer.mozilla.org/en-US/docs/Glossary/Safe)
- related PR #340

### DIFF
--- a/litestar/config/csrf.py
+++ b/litestar/config/csrf.py
@@ -34,7 +34,7 @@ class CSRFConfig:
     """The value to set in the ``SameSite`` attribute of the cookie."""
     cookie_domain: str | None = field(default=None)
     """Specifies which hosts can receive the cookie."""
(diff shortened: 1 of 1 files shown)


Alc-Alc (approved):
The PR linked by @wer153 in the description had `OPTIONS` in the `safe_methods` set, at least until some point before merging. I do not understand why it was removed in this ["Address PR comments" commit](https://github.com/litestar-org/litestar/pull/340/commits/56499c3a735d914dbfdb59b254546f0e1dd02a89) as I don't see any comments asking to remove.

With that context and the mdn docs linked, I am approving this PR. I will let other maintainers also get an eye. Thanks 🙂

cofin (approved):

PR #3541, opened 2024-06-03
feat(logging): always log exceptions by default

jderrien:
## Description

Better to always show exceptions by default.

In development, it will be easier for new users, even if using the debug mode will shows the exceptions.
In production, it's better to show the error if something goes wrong.

This would increase the log size.

While I think it's better to always log server errors (5xx errors), there is no real value to log expcetions for (some) client errors, at least for 404 errors.

So we might want to update the default `exception_logging_handler` at the same time to avoid that.

Happy to discuss more before merging this.

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -218,7 +218,7 @@ class LoggingConfig(BaseLoggingConfig):
     """
     configure_root_logger: bool = field(default=True)
     """Should the root logger be configured, defaults to True for ease of configuration."""
-    log_exceptions: Literal["always", "debug", "never"] = field(default="debug")
+    log_exceptions: Literal["always", "debug", "never"] = field(default="always")
     """Should exceptions be logged, defaults to log exceptions when 'app.debug == True'"""
     traceback_line_limit: int = field(default=-1)
     """Max number of lines to print for exception traceback.
@@ -438,7 +438,7 @@ class StructLoggingConfig(BaseLoggingConfig):
     """Logger factory to use."""
     cache_logger_on_first_use: bool = field(default=True)
     """Whether to cache the logger configuration and reuse."""
-    log_exceptions: Literal["always", "debug", "never"] = field(default="debug")
+    log_exceptions: Literal["always", "debug", "never"] = field(default="always")
     """Should exceptions be logged, defaults to log exceptions when 'app.debug == True'"""
     traceback_line_limit: int = field(default=-1)
     """Max number of lines to print for exception traceback.


JacobCoffee (commented):
this would be considering breaking i think? @litestar-org/maintainers

jderrien:
> this would be considering breaking i think?

It doesn't break the API per se, but it's a major behavior change.
That's said, I don't believe people live without logging exceptions even in production? I might be naive. 👀

In all cases, it would be fine to reduce the log size.

* Print exceptions on 5xx errors
* For 4xx error:
  * Do not print the exception, an error log would be enough, or
  * Reduce stacktrace size

---

About reducting the stacktrace size

For exampe, the exception is raised twice for a 404 error. It could be raised once, it would be less verbose. But even doing so, the stacktrace is not very interesting for the debug):
```
INFO:     127.0.0.1:55018 - "GET /notfound HTTP/1.1" 404 Not Found
ERROR - 2024-06-03 17:04:39,776 - litestar - config - Uncaught exception (connection_type=http, path=/notfound):
  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/_asgi/routing_trie/traversal.py", line 157, in parse_path_to_route
    node, path_parameters, path = traverse_route_map(
                                  ^^^^^^^^^^^^^^^^^^^
  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/_asgi/routing_trie/traversal.py", line 54, in traverse_route_map
    raise NotFoundException()
litestar.exceptions.http_exceptions.NotFoundException: 404: Not Found

  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/middleware/_internal/exceptions/middleware.py", line 158, in __call__
  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/_asgi/asgi_router.py", line 90, in __call__
    asgi_app, route_handler, scope["path"], scope["path_params"] = self.handle_routing(
                                                                   ^^^^^^^^^^^^^^^^^^^^
  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/_asgi/asgi_router.py", line 112, in handle_routing
    return parse_path_to_route(
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/jderrien/projects/litestar-minimal/.venv/lib/python3.12/site-packages/litestar/_asgi/routing_trie/traversal.py", line 174, in parse_path_to_route
    raise NotFoundException() from e
litestar.exceptions.http_exceptions.NotFoundException: 404: Not Found
```

provinzkraut:
> this would be considering breaking i think? @litestar-org/maintainers

I would consider this a breaking change. @jderrien you're right that it doesn't break an API, but behavioural changes can be breaking as well. I think we should shelve this for 3.0

jderrien:
@provinzkraut I can understand.

Instead, I'll send another PR for 2.0 to update the doc (to explain the current behavior and advise to set `log_exceptions=True`). This should be fine?

provinzkraut:
> @provinzkraut I can understand.
> 
> Instead, I'll send another PR for 2.0 to update the doc (to explain the current behavior and advise to set `log_exceptions=True`). This should be fine?

Sounds good!
 
> Do you want me to close this PR? Re-open one on the `v3.0` branch?

Feel free to reopen on v3.0. But seeing how we might drop a lot of the logging functionality for that, it's possible the changes won't even make it to the final release.

PR #3543, opened 2024-06-03
refactor(logging): improve `LoggingConfig` & deprecate `LoggingConfig.propagate`

jderrien:
## Description

I started by refactoring the tests. It might be easier to review this PR commit by commit.

This PR will:
* Fix minor bugs like the one described in https://github.com/orgs/litestar-org/discussions/3482
    > What's a bit strange is that I have to provide a "standard" formatter otherwise I get a KeyError:
* Only pass valid values to the `dictConfig` classes
* Deprecate `propagate` at the root `dictConfig` level (this value is only supported at the `Logger` level, it was passed to `dictConfig()` but has no effect AFAIK)

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -200,7 +200,11 @@ class LoggingConfig(BaseLoggingConfig):
     Filter instance.
     """
     propagate: bool = field(default=True)
-    """If messages must propagate to handlers higher up the logger hierarchy from this logger."""
+    """If messages must propagate to handlers higher up the logger hierarchy from this logger.
+
+    .. deprecated:: 2.10.0
+        This parameter is deprecated. It will be removed in a future release. Use `propagate` at the logger level.
+    """
     formatters: dict[str, dict[str, Any]] = field(default_factory=_get_default_formatters)
     """A dict in which each key is a formatter and each value is a dict describing how to configure the corresponding
     Formatter instance. A `standard` formatter is provided.
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -196,26 +196,34 @@ class LoggingConfig(BaseLoggingConfig):
     disable_existing_loggers: bool = field(default=False)
     """Whether any existing non-root loggers are to be disabled."""
     filters: dict[str, dict[str, Any]] | None = field(default=None)
-    """A dict in which each key is a filter id and each value is a dict describing how to configure the corresponding
-    Filter instance.
+    """A dict in which each key is a filter id and each value is a dict describing how to configure the
+    corresponding Filter_ instance.
+
+    .. _Filter: https://docs.python.org/3/library/logging.html#filter-objects
     """
     propagate: bool = field(default=True)
     """If messages must propagate to handlers higher up the logger hierarchy from this logger.
 
     .. deprecated:: 2.10.0
-        This parameter is deprecated. It will be removed in a future release. Use `propagate` at the logger level.
+        This parameter is deprecated. It will be removed in a future release. Use ``propagate`` at the logger level.
     """
     formatters: dict[str, dict[str, Any]] = field(default_factory=_get_default_formatters)
-    """A dict in which each key is a formatter and each value is a dict describing how to configure the corresponding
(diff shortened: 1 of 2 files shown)


jderrien:
Pyright reports issues like this one:
```
  /home/runner/work/litestar/litestar/tests/unit/test_logging/test_logging_config.py:155:22 - error: Cannot access member "handlers" for type "Logger"
```

If I add `type: ignore[attr-defined]`, then mypy reports issues like `error: Unused "type: ignore" comment  [unused-ignore]`.

So I end up with `type: ignore[attr-defined,unused-ignore]` to get both mypy & pyright happy.

If anyone has a better idea, let me know. Thank 🙏

litestar/logging/config.py
```diff
@@ -188,23 +200,22 @@ class LoggingConfig(BaseLoggingConfig):
...
     """
-    loggers: dict[str, dict[str, Any]] = field(
-        default_factory=lambda: {
-            "litestar": {"level": "INFO", "handlers": ["queue_listener"], "propagate": False},
-        }
-    )
+    loggers: dict[str, dict[str, Any]] = field(default_factory=_get_default_loggers)
     """A dict in which each key is a logger name and each value is a dict describing how to configure the corresponding
-    Logger instance.
+    Logger instance. A 'litestar' logger is mandatory and will be configured as required.
```
JacobCoffee, line 218:
```suggestion
    Logger instance. A ``litestar`` logger is mandatory and will be configured as required.
```

litestar/logging/config.py
```diff
@@ -188,23 +200,22 @@ class LoggingConfig(BaseLoggingConfig):
...
-            "standard": {"format": "%(levelname)s - %(asctime)s - %(name)s - %(module)s - %(message)s"}
-        }
-    )
+    """If messages must propagate to handlers higher up the logger hierarchy from this logger.
+
+    .. deprecated:: 2.10.0
+        This parameter is deprecated. It will be removed in a future release. Use `propagate` at the logger level.
+    """
+    formatters: dict[str, dict[str, Any]] = field(default_factory=_get_default_formatters)
+    """A dict in which each key is a formatter and each value is a dict describing how to configure the corresponding
```
JacobCoffee, line 209:
```suggestion
    """A :term:`dict` in which each key is a formatter and each value is a dict describing how to configure the corresponding
```

litestar/logging/config.py
```diff
@@ -188,23 +200,22 @@ class LoggingConfig(BaseLoggingConfig):
...
+    Formatter instance. A `standard` formatter is provided.
```
JacobCoffee, line 210:
```suggestion
    Formatter instance. A ``standard`` formatter is provided.
```

JacobCoffee: for the noun `Formatter` could we link to https://docs.python.org/3/library/logging.html#formatter-objects?

jderrien:
> So I end up with `type: ignore[attr-defined,unused-ignore]` to get both mypy & pyright happy.

Finally updated to `pyright: ignore[reportGeneralTypeIssues]`

litestar/logging/config.py
jderrien: Sure, I also added links to Logger, Handler and Filter

litestar/logging/config.py
jderrien: I'm reverting these ones (if you agree?) because it would link to https://litestar-org.github.io/litestar-docs-preview/3543/reference/datastructures/index.html#litestar.datastructures.Cookie.dict

Also, the link to dict is just above, in the type signature.

JacobCoffee (approved):

provinzkraut:
@jderrien Can you update the branch so this can be merged? :)

provinzkraut (approved):

jderrien:
@provinzkraut I rebased the branch and pushed it, but the auto-merge got disabled in the process. Could you merge it please?

What should I have done to keep the auto-merge working?

provinzkraut:
> What should I have done to keep the auto-merge working? Do it via the GitHub UI?

GitHub tells me that this only works if you have given maintainers write access to the branch in your fork.

PR #3549, opened 2024-06-06
chore(refactor): python-jose is removed from project.optional-dependency.

Squidtyper:
in security, the tokens' encode and decode now uses pyjwt.

## Description

-

### DIFF
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -87,10 +87,7 @@ class Token:
             NotAuthorizedException: If the token is invalid.
         """
         try:
-            payload = jwt.decode(jwt=encoded_token,
-                key=secret,
-                algorithms=[algorithm],
-                options={"verify_aud": False})
+            payload = jwt.decode(jwt=encoded_token, key=secret, algorithms=[algorithm], options={"verify_aud": False})
             exp = datetime.fromtimestamp(payload.pop("exp"), tz=timezone.utc)
             iat = datetime.fromtimestamp(payload.pop("iat"), tz=timezone.utc)
             field_names = {f.name for f in dataclasses.fields(Token)}
@@ -117,8 +114,7 @@ class Token:
         """
         try:
             return jwt.encode(
-                payload={k: v for k, v in asdict(self).items() if v is not None},
-                key=secret,
-                algorithm=algorithm)
+                payload={k: v for k, v in asdict(self).items() if v is not None}, key=secret, algorithm=algorithm
+            )
         except (PyJWTError, InvalidTokenError) as e:
             raise ImproperlyConfiguredException("Failed to encode token") from e
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -5,7 +5,8 @@ from dataclasses import asdict, dataclass, field
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 
-from jose import JWSError, JWTError, jwt
+import jwt
+from jwt import InvalidTokenError, PyJWTError
 
 from litestar.exceptions import ImproperlyConfiguredException, NotAuthorizedException
 
@@ -86,7 +87,10 @@ class Token:
             NotAuthorizedException: If the token is invalid.
         """
         try:
-            payload = jwt.decode(token=encoded_token, key=secret, algorithms=[algorithm], options={"verify_aud": False})
+            payload = jwt.decode(jwt=encoded_token,
+                key=secret,
+                algorithms=[algorithm],
+                options={"verify_aud": False})
             exp = datetime.fromtimestamp(payload.pop("exp"), tz=timezone.utc)
             iat = datetime.fromtimestamp(payload.pop("iat"), tz=timezone.utc)
             field_names = {f.name for f in dataclasses.fields(Token)}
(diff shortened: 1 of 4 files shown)


provinzkraut:
Thanks @Squidtyper, this is a welcome change!
Am I correct to assume this is motivated by the maintenance status of python-jose?

Squidtyper:
Yes. We are eliminating the use of python-jose in our own code and we think litestar is better off with pyjwt as well. It's my first time contributing. I see there are some errors.

cofin:
before merging this, I would like to evaluate [this library ](https://github.com/authlib/joserfc) instead of `pyjwt`.  Both `python-jose` and `pyjwt` have gone through periods of inactivity in their history.

pyproject.toml
```diff
@@ -31,19 +31,20 @@ classifiers = [
...
+    "importlib-resources>=5.12.0; python_version < \"3.9\"",
+    "msgspec>=0.18.2",
+    "multidict>=6.0.2",
+    "polyfactory>=2.6.3",
+    "pyyaml",
+    "typing-extensions",
+    "click",
+    "rich>=13.0.0",
+    "rich-click",
+    "pyjwt>=2.8.0",
```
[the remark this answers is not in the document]
cofin: Yes, it should remain optional.

sherbang:
> before merging this, I would like to evaluate [this library ](https://github.com/authlib/joserfc) instead of `pyjwt`. Both `python-jose` and `pyjwt` have gone through periods of inactivity in their history.

joserfc and the authlib project as a whole are certainly interesting, but it seems to me that pyjwt is the better option right now.

We can start with Snyk Advisor scores of [79 for joserfc](https://snyk.io/advisor/python/joserfc) vs [91 for pyjwt](https://snyk.io/advisor/python/pyjwt) which also shows more consistent commit activity for the last 2 years for pyjwt.

Pyjwt is simply a much more popular package right now, which suggests it's more likely to have some longevity, it has far more contributors, and thus far more eyes on it for future security issues.

The authlib project is interesting, in that they've built a small company around the auth code they've built, but looking at the activity graph of [their github](https://github.com/authlib) suggests that they haven't been very active in this for a while, so it may be a dying effort.

However, either of these packages would be an improvement over `python-jose`, which has unresolved security issues.  I suggest that the priority now should be removing `python-jose`, and not delaying over considering which replacement library is marginally better.

Ref: https://github.com/jpadilla/pyjwt/issues/942

provinzkraut:
> joserfc and the authlib project as a whole are certainly interesting, but it seems to me that pyjwt is the better option right now.

I agree. `pyjwt` is the de-facto standard and does everything we need here. It being the most popular JWT library also has the benefit that if users are doing something else with JWTs outside what Litestar provides, it's most likely this is the library they'll be using.

cofin (approved):

cofin:
> However, either of these packages would be an improvement over `python-jose`, which has unresolved security issues. I suggest that the priority now should be removing `python-jose`, and not delaying over considering which replacement library is marginally better.

Agree with this.  

So that we don't get bogged down without a fix, I'm inclined to go with this change.  It can be revisited should there be a need.

JacobCoffee (approved):

JacobCoffee:
im having trouble getting docs build to work

litestar/security/jwt/token.py
```diff
@@ -85,8 +90,35 @@ def decode(cls, encoded_token: str, secret: str | dict[str, str], algorithm: str
         Raises:
             NotAuthorizedException: If the token is invalid.
         """
+
+        def base64url_decode(code: str) -> bytes:
+            padding = "=" * (4 - (len(code) % 4))
+            return base64.urlsafe_b64decode(code + padding)
```
provinzkraut, line 96: Why are we doing custom b64 decoding here?

litestar/security/jwt/token.py
```diff
@@ -85,8 +90,35 @@ def decode(cls, encoded_token: str, secret: str | dict[str, str], algorithm: str
...
+                y = int.from_bytes(base64url_decode(secret["y"]), byteorder="big")
+                converted_secret = ec.EllipticCurvePublicNumbers(x, y, ec.SECP256R1()).public_key(default_backend())
+            elif secret["kty"] == "OKP" and secret["crv"] == "Ed25519":
+                x = base64url_decode(secret["x"])
+                converted_secret = ed25519.Ed25519PublicKey.from_public_bytes(x)
+            elif secret["kty"] == "OKP" and secret["crv"] == "Ed448":
+                x = base64url_decode(secret["x"])
+                converted_secret = ed448.Ed448PublicKey.from_public_bytes(x)
+            else:
+                raise TypeError("The secret is not a form of allowed public key.")
```
provinzkraut, line 116: These changes seem to be unrelated to the switch to pyjwt. If this is a feature you want to add, this should be done in a separate PR.

provinzkraut (changes requested):
@Squidtyper Sorry for the long turnaround time on this. After the discussion we had about this, I agree with @cofin that this is okay to go in. 

I've left two comments to address, once that's done, this can be merged.

provinzkraut:
> im having trouble getting docs build to work

@JacobCoffee There were some accidental commits in here. I've cleaned the branch and now everything's good. There seems to be however an issue with the `pdm.lock` / `pyproject.toml`.

litestar/security/jwt/token.py
cofin: @provinzkraut Some details from the author would be good, but I do believe this is park of the JWK/backend implementations here: https://github.com/mpdavis/python-jose/tree/master/jose

litestar/security/jwt/token.py
cofin: This also looks like functionality from: https://github.com/mpdavis/python-jose/blob/master/jose/utils.py#L65

provinzkraut:
Closing this in favour of #3684, so we can get a release out.

@Squidtyper If you want to complete the work on the features you've started here, feel free to open another PR for this! :)

Squidtyper:
Thank you a lot!

Issue #3554, opened 2024-06-09
Unexpected behavior from `module_to_os_path`


ghferrari:
### Description

The file [litestar/utils/module_loader.py](https://github.com/litestar-org/litestar/blob/84f51c8afc3203cd4914922b2ec3c1e92d5d40ba/litestar/utils/module_loader.py#L21) contains a function definition for `module_to_os_path`.

Assuming I've understood the code comments, the purpose of this code is to return the path to a **directory**, which is either the base directory of the project or (when supplied with the name of a module) the base directory of the module.

Unfortunately, if you define your Litestar object in a file named `app.py`, and there is no other module named `app`, then this function returns the path to that file. This is already an error given that the intention is to return a path to a directory.

I noticed this problem while attempting to set up my own starting configuration following the `litestar-fullstack` repository. This repository defines a `BASE_DIR` property at https://github.com/litestar-org/litestar-fullstack/blob/8e6edb90a401778741062a8383ff6e6f354b44dd/src/app/config/base.py#L23 using `module_to_os_path` which is then used to define subdirectories for alembic. In my case (using a file named `app.py`, no module named `app`), this produces invalid directory pathnames which cannot then be created. For example, when attempting to run as per the MCVE below, I see the error:
```
  File "/home/ghf/projects/dappel-litestar/.venv/lib/python3.11/site-packages/alembic/command.py", line 99, in init
    script._generate_template(
  File "/home/ghf/projects/dappel-litestar/.venv/lib/python3.11/site-packages/alembic/script/base.py", line 593, in _generate_template
    util.template_to_file(src, dest, self.output_encoding, **kw)
  File "/home/ghf/projects/dappel-litestar/.venv/lib/python3.11/site-packages/alembic/util/pyfiles.py", line 41, in template_to_file
    with open(dest, "wb") as f:
         ^^^^^^^^^^^^^^^^
NotADirectoryError: [Errno 20] Not a directory: '/home/ghf/projects/dappel-litestar/src/app.py/db/migrations/alembic.ini'
```

### URL to code causing the issue

https://github.com/litestar-org/litestar/blob/84f51c8afc3203cd4914922b2ec3c1e92d5d40ba/litestar/utils/module_loader.py#L21

### MCVE

```python
from __future__ import annotations
from typing import TYPE_CHECKING
from advanced_alchemy.extensions.litestar import (
    AlembicAsyncConfig,
    AsyncSessionConfig,
    SQLAlchemyPlugin,
[...]

PR #3565, opened 2024-06-12
fix(#3554): correct ``module_to_os_path`` to return directory paths

supercoder-dev:
To resolve the issue, we need to update the `module_to_os_path` function to check if the found path is a file. If it is, we should return the parent directory instead. This can be done by using `Path(src.origin).parent` if `src.origin` is a file.

### DIFF
--- a/litestar/utils/module_loader.py
+++ b/litestar/utils/module_loader.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


cofin (approved):

cofin:
@all-contributors, add @supercoder-dev for bugs, code

PR #3567, opened 2024-06-13
fix(testing): `.websocket_connect` does not respect `base_url`

provinzkraut:
Fix a bug that caused the `.websocket_connect` methods on `TestClient` and `AsyncTestClient` to not respect the `base_url` set in the client's constructor, and instead would use the static `ws://testerver` URL as a base.

Also removes most of the test client code as it was unneeded and in the way of this fix :)

<hr>

Explanation for the last part: All the extra code we had was just proxying method calls to the `httpx.Client` / `httpx.AsyncClient`, while altering the base URL. Since we already set the base URL on the httpx Client's superclass instance, which in turn does this merging internally, this step isn't needed at all.


cofin (approved):
LGTM

Issue #3548, opened 2024-06-06
Enhancement: Add class and funcion name to `ImproperlyConfiguredException`s


rafalkrupinski:
### Summary

When you forget to add a type annotation in a route handler function, Litestar raises ImproperlyConfiguredException but neither the stack trace, nor the message shows which function or method is in error.

If the annotation were checked in `@get` decorators, the stack trace would let you find the line with the problem. If that's impossible, the exception should at least show function name, and the class in case of methods.

I volunteer to add function name in `_validate_handler_function()`.

euri10:
I definitely +1 this : I found that when switching from another framework, sometimes you make big refactors, forget some stuff and get hit by that and it's hard to find where you made the mistakes as you just changed too many things, this would definitly make the process smoother.
There are obviouly other cases for that, but this one comes to mind :)

rafalkrupinski:
> when switching from another framework

That's exactly where I'm coming from

euri10:
I should have listed them, but there were other cases where this `ImproperlyConfiguredException` popped up in case of big refactors, and wasn't too useful because of the lack of more granular info. Would that be a good time to list where they're used and maybe investigate if providing more info would make sense ?
Anyways a goo 1st step would be this, thanks for raising it.

provinzkraut:
>  Would that be a good time to list where they're used and maybe investigate if providing more info would make sense ?

Yeah, absolutely. I think we can generally improve a lot of our error messages with more details to make them more useful, but it's going to be a longer process I think. Incremental PRs are definitely welcome there! :)

provinzkraut:
Implemented in #3575.

PR #3575, opened 2024-06-16
feat: add handler name to exceptions in handler validation.

rafalkrupinski:
Related: #

## Description

- Add handler name to exceptions raise by _validate_handler_function.

## Closes
 
addresses #3548

### DIFF
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -255,6 +255,16 @@ class BaseRouteHandler:
         return get_name(unwrap_partial(self.fn))
 
     @property
+    def handler_fullname(self) -> str:
+        """Get the qualified name of the handler function.
+
+        Returns:
+            Qualified name of the handler function
+        """
+        handler_fn = unwrap_partial(self.fn)
+        return handler_fn.__module__ + "." + handler_fn.__qualname__
+
+    @property
     def dependency_name_set(self) -> set[str]:
         """Set of all dependency names provided in the handler's ownership layers."""
         layered_dependencies = (layer.dependencies or {} for layer in self.ownership_layers)
--- a/litestar/handlers/base.py
+++ b/litestar/handlers/base.py
@@ -255,16 +255,6 @@ class BaseRouteHandler:
         return get_name(unwrap_partial(self.fn))
 
     @property
-    def handler_fullname(self) -> str:
-        """Get the qualified name of the handler function.
-
-        Returns:
-            Qualified name of the handler function
-        """
-        handler_fn = unwrap_partial(self.fn)
-        return handler_fn.__module__ + "." + handler_fn.__qualname__
-
-    @property
     def dependency_name_set(self) -> set[str]:
         """Set of all dependency names provided in the handler's ownership layers."""
         layered_dependencies = (layer.dependencies or {} for layer in self.ownership_layers)
--- a/litestar/handlers/http_handlers/decorators.py
+++ b/litestar/handlers/http_handlers/decorators.py
@@ -597,7 +597,7 @@ class head(HTTPRouteHandler):
             or is_class_and_subclass(return_annotation, ASGIFileResponse)
         ):
             raise ImproperlyConfiguredException(
-                f"{self.handler_fullname}: Handlers for 'HEAD' requests must not return a value. Either return 'None' or a response type without a body."
+                f"{self}: Handlers for 'HEAD' requests must not return a value. Either return 'None' or a response type without a body."
             )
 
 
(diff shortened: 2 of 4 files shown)


litestar/handlers/base.py
```diff
@@ -254,6 +254,16 @@ def handler_name(self) -> str:
...
 
+    @property
+    def handler_fullname(self) -> str:
+        """Get the qualified name of the handler function.
+
+        Returns:
+            Qualified name of the handler function
+        """
+        handler_fn = unwrap_partial(self.fn)
+        return handler_fn.__module__ + "." + handler_fn.__qualname__
```
provinzkraut, line 265: I think the logic here should be 

```suggestion
        return self.name or self.handler_name
```
so custom handler names show up as well, which means the handler name shown in the exception matches e.g. the one you'd put into `route_reverse`.

litestar/handlers/http_handlers/decorators.py
```diff
@@ -596,7 +596,9 @@ def _validate_handler_function(self) -> None:
             or is_class_and_subclass(return_annotation, File)
             or is_class_and_subclass(return_annotation, ASGIFileResponse)
         ):
-            raise ImproperlyConfiguredException("A response to a head request should not have a body")
+            raise ImproperlyConfiguredException(
+                f"A route handler function '{self.handler_fullname}' that handles HEAD request should not have a body"
```
provinzkraut, line 600:
```suggestion
                f"{self.handler_fullname}: Handlers for 'HEAD' requests must not return a value. Either return 'None' or a response type without a body."
```

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -85,17 +85,25 @@ def _validate_handler_function(self) -> None:
         super()._validate_handler_function()
 
         if not self.parsed_fn_signature.return_type.is_subclass_of(NoneType):
-            raise ImproperlyConfiguredException("Websocket handler functions should return 'None'")
+            raise ImproperlyConfiguredException(
+                f"Websocket handler function '{self.handler_fullname}' should return 'None'"
```
provinzkraut, line 89:
```suggestion
                f"{self.handler_fullname}: WebSocket handlers must return 'None'"
```

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -85,17 +85,25 @@ def _validate_handler_function(self) -> None:
...
+            )
 
         if "socket" not in self.parsed_fn_signature.parameters:
-            raise ImproperlyConfiguredException("Websocket handlers must set a 'socket' kwarg")
+            raise ImproperlyConfiguredException(
+                f"Websocket handler '{self.handler_fullname}' must set a 'socket' kwarg"
```
provinzkraut, line 94:
```suggestion
                f"{self.handler_fullname}: WebSocket handlers must define a 'socket' parameter"
```

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -85,17 +85,25 @@ def _validate_handler_function(self) -> None:
...
+            )
 
         for param in ("request", "body", "data"):
             if param in self.parsed_fn_signature.parameters:
-                raise ImproperlyConfiguredException(f"The {param} kwarg is not supported with websocket handlers")
+                raise ImproperlyConfiguredException(
+                    f"{self.handler_fullname}: the {param} kwarg is not supported with websocket handlers"
```
provinzkraut, line 100:
```suggestion
                    f"{self.handler_fullname}: The {param} kwarg is not supported with websocket handlers"
```

litestar/handlers/websocket_handlers/route_handler.py
```diff
@@ -85,17 +85,25 @@ def _validate_handler_function(self) -> None:
...
+                )
 
         if not is_async_callable(self.fn):
-            raise ImproperlyConfiguredException("Functions decorated with 'websocket' must be async functions")
+            raise ImproperlyConfiguredException(
+                f"Function '{self.handler_fullname}' decorated with 'websocket' must be async functions"
```
provinzkraut, line 105:
```suggestion
                f"{self.handler_fullname}: WebSocket handler functions must be asynchronous"
```

provinzkraut (changes requested):
Thanks @rafalkrupinski!

I've left a few suggestions for consistency

litestar/handlers/base.py
rafalkrupinski: But this way we're loosing package and class name...

provinzkraut: How about just using `str(handler.fn)`? I'm hesitant to add yet another handler identifier that does something slightly different; We already have `handler_id`, `name` and `handler_name`.

Maybe we could change `handler_name` in v3 to use the fully qualified name instead.

rafalkrupinski: Actually `str(self)` does exactly that (module+qualname)

provinzkraut (approved):

PR #3578, opened 2024-06-17
feat(logging): add param `logging_module` to `LoggingConfig`

### LINKED ISSUE #3536: Enhancement: add an option `logging_module` to `LoggingConfig`
### Summary

It would permit to switch from `picologging` to `logging` without having to hack around or uninstall `picologging`.
It would be useful in tests but also in production (vs having to update a requirements file and rebuild an artifact).

This option would be optional. The default behavior would be the same as today (use `picologging` if the lib is installed).

### Basic Example

```
std_logging_config = LoggingConfig(logging_module="logging")
picologging_logging_config = LoggingConfig(logging_module="picologging")
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

jderrien:
## Description

`logging_module` provides a way to switch easily from `logging` to `picologging`.

## Closes

Closes #3536

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -90,13 +90,19 @@ def get_logger_placeholder(_: str | None = None) -> NoReturn:
     )
 
 
-def _get_default_handlers() -> dict[str, dict[str, Any]]:
+def _get_default_logging_module() -> str:
+    if find_spec("picologging"):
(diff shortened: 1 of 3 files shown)


cofin (approved):
LGTM

PR #3580, opened 2024-06-17
fix(OpenAPI): Correctly handle `typing.NewType` 

provinzkraut:
When encountering a `typing.NewType` during OpenAPI schema generation, we currently treat it as an opaque type. This PR changes the behaviour such that `typing.NewType`s are always unwrapped during schema generation.


cofin (approved):
LGTM.  Sorry this took so long.

JacobCoffee (approved):

PR #3585, opened 2024-06-20
fix: encode response content object returned from an exception handler.

rafalkrupinski:
## Description

When an handler raises an exception and exception handler returns a Response with a model (e.g. pydantic) object, encode that object with the app type_encoders.

### DIFF
--- a/litestar/middleware/_internal/exceptions/middleware.py
+++ b/litestar/middleware/_internal/exceptions/middleware.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
     from litestar import Response
     from litestar.app import Litestar
     from litestar.connection import Request
+    from litestar.handlers import BaseRouteHandler
     from litestar.logging import BaseLoggingConfig
     from litestar.types import (
         ASGIApp,
@@ -202,7 +203,11 @@ class ExceptionHandlerMiddleware:
         exception_handler = get_exception_handler(exception_handlers, exc) or self.default_http_exception_handler
         request: Request[Any, Any, Any] = litestar_app.request_class(scope=scope, receive=receive, send=send)
         response = exception_handler(request, exc)
-        await response.to_asgi_response(app=None, request=request)(scope=scope, receive=receive, send=send)
+        route_handler: BaseRouteHandler | None = scope.get("route_handler")
+        type_encoders = route_handler.resolve_type_encoders() if route_handler else litestar_app.type_encoders
+        await response.to_asgi_response(app=None, request=request, type_encoders=type_encoders)(
+            scope=scope, receive=receive, send=send
+        )
 
     @staticmethod
     async def handle_websocket_exception(send: Send, exc: Exception) -> None:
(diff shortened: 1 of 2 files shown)


rafalkrupinski:
discussion: https://discord.com/channels/919193495116337154/1240949774631440394

litestar/middleware/_internal/exceptions/middleware.py
```diff
@@ -202,7 +202,9 @@ async def handle_request_exception(
         exception_handler = get_exception_handler(exception_handlers, exc) or self.default_http_exception_handler
         request: Request[Any, Any, Any] = litestar_app.request_class(scope=scope, receive=receive, send=send)
         response = exception_handler(request, exc)
-        await response.to_asgi_response(app=None, request=request)(scope=scope, receive=receive, send=send)
+        await response.to_asgi_response(app=None, request=request, type_encoders=litestar_app.type_encoders)(
```
provinzkraut, line 205:
```suggestion
        await response.to_asgi_response(app=None, request=request, type_encoders=litestar_app.type_encoders)(
```

This should try to retrieve the type encoders from the route handler, and only if one isn't found (e.g. during a `404`), fall back to the app's type encoders

provinzkraut (changes requested):

provinzkraut:
Thanks @rafalkrupinski! I've left one suggestion

litestar/middleware/_internal/exceptions/middleware.py
rafalkrupinski: I don't see any changes

rafalkrupinski: > This should try to retrieve the type encoders from the route handler, and only if one isn't found (e.g. during a 404), fall back to the app's type encoders

Isn't that the default behaviour anyway?
https://github.com/litestar-org/litestar/blob/3de1b1dabd53b85e17eae351e4e73eddf71ce284/litestar/response/base.py#L262

provinzkraut: > I don't see any changes

Sorry, this wasn't meant to be a code suggestion, just a comment.

> Isn't that the default behaviour anyway?

No. The response isn't tied to the app/route handler's type encoders. You need to call `handler.resolve_type_encoders()` to do that.

rafalkrupinski: Ah, gotcha, there three levels of type encoders - app, handler and response and I've missed the handler

provinzkraut: > Ah, gotcha, there three levels of type encoders - app, handler and response and I've missed the handler

Actually, there's an unknown number of type encoders, since type encoders are supported on all layers (app, routers, controllers, handlers). The `handlers.resolve_type_encoders()` will merge those and give you the result back. In the happy path, this is not an issue since we'll always have access to a handler when creating the response, so we can just pass those in, but during exception handling, we have to do a bit of legwork :)

rafalkrupinski: yeah, I had a peek in that method. Thanks for the explanation!

rafalkrupinski:
Thanks for merging! Can't wait for the release ;)

Issue #3597, opened 2024-06-25
Enhancement: add after_startup and before_shutdown hooks


betaboon:
### Summary

I need to run a task after the app has been fully setup.
thus `on_startup` is to early.

Talking to @cofin on discord revealed the following:
- [Litestar.lifespan](https://github.com/litestar-org/litestar/blob/de8f4a77ff1ef2095e2dc2657b77b9b866322d3f/litestar/app.py#L600) seems to be the right location to add this
- `after_startup` and `before_shutdown` should be added to the app/appconfig objects

cofin:
We should consider renaming this to be Prioritized startup handlers?

mixartemev:
I need to run a task after the ChannelsPlugin loaded in app.
`on_startup` it's not loaded yet,
before `yield` in lifespan it's not loaded yet also.
How should I run this task?

provinzkraut:
> I need to run a task after the ChannelsPlugin loaded in app. `on_startup` it's not loaded yet, before `yield` in lifespan it's not loaded yet also. How should I run this task?

This is tricky, and so are lifespan hook priorities unfortunately.

Could you describe a bit more what you need to do after the channels plugin is loaded? Maybe there's a different way to achieve this.

mixartemev:
> Could you describe a bit more what you need to do after the channels plugin is loaded? Maybe there's a different way to achieve this.


<img width="1176" alt="Image" src="https://github.com/user-attachments/assets/c5d10c93-f8db-47c5-a87d-947b0be98c4a" />

I need to runnig listener of users in db throughout the app lifespan.
But at this line (before yield) plugin AsyncPgChannelsBackend is not started up yet, so its `_listener_conn` is not inited yet.

euri10:
not tested but that's on top of my head how I'd do it
```
@asynccontextmanager
async def lifespan(app: Litestar):
    cp = next((p for p in app.plugins.init if isinstance(p, ChannelsPlugin)), None)
    asyncio.create_task(user_create(cp))
    yield


app = Litestar(
    debug=True,
    plugins=[ChannelsPlugin(AsyncPgChannelsBackend(dsn), arbitrary_channels_allowed=True),],
    lifespan=[lifespan]
)
```

PR #3598, opened 2024-06-25
feat: adds before_shutdown/after_startup lifespan hooks

cofin:
## Description

-
Closes: #3597

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,13 +24,13 @@ repos:
       - id: unasyncd
         additional_dependencies: ["ruff"]
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.4.4"
+    rev: "v0.4.10"
     hooks:
       - id: ruff
         args: ["--fix"]
       - id: ruff-format
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.6
+    rev: v2.3.0
     hooks:
       - id: codespell
         exclude: "tests/openapi/typescript_converter/test_converter|README.md"
--- a/docs/usage/responses.rst
+++ b/docs/usage/responses.rst
@@ -497,7 +497,7 @@ in:
    Set-Cookie: router-cookie=router value; Path=/; SameSite=lax
    Set-Cookie: app-cookie=app value; Path=/; SameSite=lax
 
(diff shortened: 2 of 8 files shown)


JacobCoffee:
am I thinking of something else or did we have these before?

provinzkraut:
@JacobCoffee you're right. #1663.

@cofin I'm not sure if this is a good solution or pattern we should support / promote.

This would effectively be the same then as adding something to the front of `on_shutdown`. 

If the goal is to provide a way to say "run this before all `on_shutdown` hooks", it would just move the problem further down the line, because then, how do you say "run this before all the other `before_shutdown` hooks"?

IMO having hooks that depend on other hooks is an anti pattern. If your hooks involve state, you should use the lifespan context manager. And if they don't involve state, then the order they're called in shouldn't matter. If the order they are called in *does* matter, combine them together in a single hook.

(same applies to the startup hooks as well)

cofin:
> @JacobCoffee you're right, see #1663.
> 
> @cofin I'm not sure if this is a good solution or pattern we should support / promote.
> 
> This would effectively be the same then as adding something to the front of `on_shutdown`.
> 
> If the goal is to provide a way to say "run this before all `on_shutdown` hooks", it would just move the problem further down the line, because then, how do you say "run this before all the other `before_shutdown` hooks"?
> 
> IMO having hooks that depend on other hooks is an anti pattern. If your hooks involve state, you should use the lifespan context manager. And if they don't involve state, then the order they're called in shouldn't matter. If the order they are called in _does_ matter, combine them together in a single hook.
> 
> (same applies to the startup hooks as well)
> 
> If we _do_ want to support giving priorities to some hooks, then we should implement a priority configuration

The specific need was `after_startup`.  I addd in `before_shutdown` since it seemed to make logical sense at the time.  I don't think we really need the `before_shutdown`.

While implementing these changes, I remembered us having this support previously, but I don't remember exactly why we removed though.  

I do think there's a valid case for having a set of actions that can run after specific startup sequences have loaded, so maybe, we should think about a priority/ordering mechanism?

provinzkraut:
> I do think there's a valid case for having a set of actions that can run after specific startup sequences have loaded, so maybe, we should think about a priority/ordering mechanism?

Definitely. IMO implementing priorities would be the way to go about this.

PR #3603, opened 2024-06-27
docs: Document custom types

stewit:
## Description
- Add documentation that explains the usage of `type_decoders` and `type_encoders` in order to allow litestar to en/decode custom types via msgspec
- See question #3584


docs/usage/custom-types.rst
```diff
@@ -0,0 +1,25 @@
+Custom types
+=============
```
JacobCoffee, line 2:
```suggestion
Custom types
============
```

docs/usage/custom-types.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Data serialization / deserialization (encoding / decoding) and validation are important parts of any API framework.
+
+Among others, litestar supports Python's builtin dataclasses, Pydantic and msgspec for these tasks, defaulting to msgspec with json as serialization protocol.
```
JacobCoffee, line 6:
```suggestion
Among others, Litestar supports Python's built-in dataclasses, Pydantic, and msgspec for these tasks, defaulting to msgspec with JSON as the serialization protocol.
```

provinzkraut, line 6: I don't think "defaulting to msgspec" is the correct phrasing here. Litestar only uses msgspec for serialisation to JSON. It *can* integrate with Pydantic and others, but doesn't use those to encode anything to JSON directly; It all goes through msgspec eventually.

I'm not sure msgspec needs to be mentioned in this part of the documentation in general, as it's more of an implementation detail. We *should* mention that that's how it works, but perhaps we can find a better place for that

provinzkraut (changes requested):

cofin:
@all-contributors, please add @stewit for docs

cofin (approved):

PR #3632, opened 2024-07-16
fix: allow non-Mapping Iterable parameters

### LINKED ISSUE #3631: Enhancement: Allow parameter of type `Iterable` to be processed as  `list`
### Description

Litestar doesn't think it should process `Iterable` parameter as an array.

The original [discussion](https://discord.com/channels/919193495116337154/1260559977081471058) was about default parameter value, but the problem is actually with how litestar handles `Iterable` itself

https://github.com/litestar-org/litestar/blob/8c4c15bb501879dabaecfbf0af541ac571c08cf3/litestar/_kwargs/parameter_definition.py#L67C9-L67C60

### MCVE

```python
@get('/sequence')
async def sequence(foo: Sequence[str]) -> Sequence[str]:
    return foo

@get('/iterable')
async def iterable(foo: Iterable[str]) -> Iterable[str]:
    return foo

app = Litestar(
    route_handlers=(
        sequence,
        iterable,
    )
)
```

### Steps to reproduce

```bash
$ curl  'localhost:8000/sequence?foo=s1&foo=s2'
["s1","s2"]

$ curl 'localhost:8000/iterable?foo=s1&foo=s2'
s2
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.9.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3631">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3631/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

rafalkrupinski:
## Description

Allow `Iterable`s other than `str` or `Mapping`s to be used as `Sequence` parameters

## Closes

Closes #3631

### DIFF
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -228,7 +228,7 @@ def is_non_string_non_mapping_iterable(annotation: Any) -> TypeGuard[Iterable[An
         return False
     typ = origin or annotation
     try:
-        return issubclass(origin or annotation, Iterable) and not (
+        return issubclass(typ, Iterable) and not (
             issubclass(typ, (str, bytes)) or issubclass(typ, Mapping)
         )
     except TypeError:  # pragma: no cover
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -55,6 +55,7 @@ __all__ = (
     "is_generic",
     "is_mapping",
     "is_non_string_iterable",
+    "is_non_string_non_mapping_iterable",
     "is_non_string_sequence",
     "is_optional_union",
     "is_undefined_sentinel",
@@ -207,6 +208,28 @@ def is_non_string_sequence(annotation: Any) -> TypeGuard[Sequence[Any]]:
                 set,
                 frozenset,
             ),
+        )
+    except TypeError:  # pragma: no cover
+        return False
+
+
+def is_non_string_non_mapping_iterable(annotation: Any) -> TypeGuard[Iterable[Any]]:
+    """Given a type annotation determine if the annotation is a sequence.
+
+    Args:
+    annotation: A type.
+
+    Returns:
+        A typeguard determining whether the type can be cast as :class`Iterable <typing.Iterable>`
+         that is not a string or Mapping.
+    """
+    origin = get_origin_or_inner_type(annotation)
+    if not origin and not isclass(annotation):
+        return False
+    typ = origin or annotation
+    try:
+        return issubclass(origin or annotation, Iterable) and not (
+            issubclass(typ, (str, bytes)) or issubclass(typ, Mapping)
         )
     except TypeError:  # pragma: no cover
         return False
--- a/litestar/_kwargs/parameter_definition.py
+++ b/litestar/_kwargs/parameter_definition.py
@@ -64,7 +64,7 @@ def create_parameter_definition(
         and default is None
         and not field_definition.is_optional
         and not field_definition.is_any,
-        is_sequence=field_definition.is_non_string_sequence,
+        is_sequence=field_definition.is_non_string_non_mapping_iterable,
     )
 
 
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -30,6 +30,7 @@ from litestar.utils.predicates import (
     is_class_and_subclass,
     is_generic,
     is_non_string_iterable,
+    is_non_string_non_mapping_iterable,
     is_non_string_sequence,
     is_union,
 )
@@ -159,7 +160,7 @@ def _traverse_metadata(
 def _create_metadata_from_type(
     metadata: Sequence[Any], model: type[T], annotation: Any, extra: dict[str, Any] | None
 ) -> tuple[T | None, dict[str, Any]]:
-    is_sequence_container = is_non_string_sequence(annotation)
+    is_sequence_container = is_non_string_non_mapping_iterable(annotation)
     result = _traverse_metadata(metadata=metadata, is_sequence_container=is_sequence_container, extra=extra)
 
     constraints = {k: v for k, v in result.items() if k in dir(model)}
@@ -289,6 +290,17 @@ class FieldDefinition:
         if self.is_optional:
             annotation = make_non_optional_union(annotation)
         return is_non_string_sequence(annotation)
+
+    @property
+    def is_non_string_non_mapping_iterable(self) -> bool:
+        """Check if the field type is an Iterable other than string or mapping.
+
+        If ``self.annotation`` is an optional union, only the non-optional members of the union are evaluated.
+        """
+        annotation = self.annotation
+        if self.is_optional:
+            annotation = make_non_optional_union(annotation)
+        return is_non_string_non_mapping_iterable(annotation)
 
     @property
     def is_any(self) -> bool:
--- a/tests/unit/test_utils/test_predicates.py
+++ b/tests/unit/test_utils/test_predicates.py
@@ -38,6 +38,7 @@ from litestar.utils.predicates import (
     is_generic,
     is_mapping,
     is_non_string_iterable,
+    is_non_string_non_mapping_iterable,
     is_non_string_sequence,
     is_undefined_sentinel,
 )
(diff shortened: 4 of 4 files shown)


rafalkrupinski:
While we're at it, is_non_string_sequence seems buggy - it claims to return True for types that can be cast to `Sequence`, but it allows `Set`s.

It's also smelly - it tests against a whole bunch of types, while `issubclass(..., Sequence)` should suffice, unless I'm missing something.

Also, `Collections.abc.Sequence` type differs from its supertypes by adding `reverse` and `__gettem__`. Reverse is never used and index access is only used once in `_extract_multipart`.
I suspect the problem is with python type system lacking a type suitable for sequential processing
`Sequence` doesn't include sets, and `Iterable` includes `Mapping` and `str`.

Perhaps we could introduce a new type, lets call it `ArrayLike` for now, as it can be represented as JSON arrays and list parameters in HTTP.

litestar/utils/predicates.py
```diff
@@ -212,6 +213,28 @@ def is_non_string_sequence(annotation: Any) -> TypeGuard[Sequence[Any]]:
...
+    Returns:
+        A typeguard determining whether the type can be cast as :class`Iterable <typing.Iterable>`
+         that is not a string or Mapping.
+    """
+    origin = get_origin_or_inner_type(annotation)
+    if not origin and not isclass(annotation):
+        return False
+    typ = origin or annotation
+    try:
+        return issubclass(origin or annotation, Iterable) and not (
```
provinzkraut, line 231:
```suggestion
        return issubclass(typ, Iterable) and not (
```

litestar/utils/predicates.py
```diff
@@ -212,6 +213,28 @@ def is_non_string_sequence(annotation: Any) -> TypeGuard[Sequence[Any]]:
         return False
 
 
+def is_non_string_non_mapping_iterable(annotation: Any) -> TypeGuard[Iterable[Any]]:
```
provinzkraut, line 216: Why does this specifically exclude `Mapping`? Wouldn't it make more sense to adapt the existing `is_non_string_sequence` and create a `is_non_string_iterable`? `Mapping` is an iterable after all, and if we're going to be broad in what we accept here, why specifically exclude it?

provinzkraut (changes requested):

litestar/utils/predicates.py
rafalkrupinski: from collections.abc import Mapping, Iterable
issubclass(Mapping, Iterable)
True

rafalkrupinski: It's been two weeks, I'll need to take a look.

rafalkrupinski: It's because `Object` (python type `Mapping`) is encoded differently. The idea was to support python types that are equivalent of `Array`. If we widen accepted types from `Sequence` to `Iterable` to support `Array` we must exclude `Mapping`.

Also HTTP allows Arrays and Objects with only Scalar values, and python `Mapping` element is complex type `tuple`.

provinzkraut: > It's been two weeks, I'll need to take a look.

Don't have much time outside work at the moment, sorry :)



> It's because `Object` (python type `Mapping`) is encoded differently. The idea was to support python types that are equivalent of `Array`. If we widen accepted types from `Sequence` to `Iterable` to support `Array` we must exclude `Mapping`.
> 
> Also HTTP allows Arrays and Objects with only Scalar values, and python `Mapping` element is complex type `tuple`.

Not sure I follow your reasoning here. If we allow `Mappings` where we allow `Iterable`, I wouldn't expect any issues, since you *can* iterate over a mapping, hence making it an iterable. As you pointed out in [this](https://github.com/litestar-org/litestar/issues/3631#issuecomment-2231645694) discussion, things aren't as strict as they should be anyway, so the line to draw is kinda arbitrary. Explicitly excluding `Mapping` would just move the arbitrary line somewhere else, and IMO it's more intuitive to just treat all non-string iterables the same here.

In any case, why can't we reuse `is_non_string_iterable` for this?

rafalkrupinski: > Don't have much time outside work at the moment, sorry :)

No problem, that's life 🤷

> I wouldn't expect any issues, since you can iterate over a mapping, hence making it an iterable

It's not about Markov principle, it's about how these types are mapped to OpenAPI and which parses are used on incoming requests.
It's clear to me that `Iterable[Scalar]` should be represented in OpenAPI as `array` and should make `ParameterDefinition.is_sequence` be `True`.
But `Mapping` is really `Iterable[tuple]` - what would it map to? Specifications don't support arrays of non-scalar types.

> As you pointed out in https://github.com/litestar-org/litestar/issues/3631#issuecomment-2231645694 discussion, things aren't as strict as they should be anyway, so the line to draw is kinda arbitrary

I said things aren't as intuitive as they should be :)

If you mean https://github.com/litestar-org/litestar/pull/3632#issuecomment-2230898096, that's completely different story. My thinking was going even further than this PR.

I thought of introducing a concept of `ArrayLike`, which would encompass python types that correspond to HTTP Array.
`is_non_string_non_mapping_iterable` could be renamed to `is_array_like`, and would replace `is_non_string_sequence` in most (perhaps all) places, since this is what its implementation suggest it's trying to do.

rafalkrupinski: > It's not about Markov principle

OK, it is all about Markov principle :) only we're handling a dual type system - python and HTTP/OpenAPI.

rafalkrupinski:
Can someone either accept it or drop it, before the changes go completely outdated? Thanks!

provinzkraut (commented):
Alright, I've taken another look at this and I think the change as such is good. The implementation can be simplified though to the following:

- Add an `is_iterable` property to `FieldDefinition`
- Set `is_sequence=field_definition.is_sequence and not field_definition.is_mapping`

The rest can be removed.

Also, this should include a test case that reproduces the actual bug it's fixing.

rafalkrupinski:
> Alright, I've taken another look at this and I think the change as such is good. The implementation can be simplified though to the following:
> 
>     * Add an `is_iterable` property to `FieldDefinition`
> 
>     * Set `is_sequence=field_definition.is_sequence and not field_definition.is_mapping`
> 
> 
> The rest can be removed.

Makes sense

> Also, this should include a test case that reproduces the actual bug it's fixing.

I'd rather copy and adjust an existing test, not to break existing patterns, but I couldn't find anything similar.

provinzkraut:
> I'd rather copy and adjust an existing test, not to break existing patterns, but I couldn't find anything similar.

You could basically just copy the example you gave in #3631, and create a parametrized test case that shows it works as expected with a bunch of iterable types that should be supported.

rafalkrupinski:
I wrote a nice matrix parametrized tests with Iterable, Sequece, List x str, int and it fails for Iterable
The patch as it is makes litestar generate correct openapi, but [msgspec officially doesn't support Iterable](https://jcristharif.com/msgspec/supported-types.html).

I think the list of supported types could be clearer and link to https://jcristharif.com/msgspec/supported-types.html

provinzkraut:
> I wrote a nice matrix parametrized tests with Iterable, Sequece, List x str, int and it fails for Iterable The patch as it is makes litestar generate correct openapi, but [msgspec officially doesn't support Iterable](https://jcristharif.com/msgspec/supported-types.html).
> 
> I think the list of supported types could be clearer and link to https://jcristharif.com/msgspec/supported-types.html

Well that was kind of my point :) 
It doesn't really make sense to decode *into* an `Iterable`, as you'll always have to decode into a concrete type, in this case a `list`. If we *were* to support encoding into `Iterable`, we'd have to override its actual type for the conversion to a `list` at some point.

IMO it's easier (and more correct) to just use a list to begin with.

Maybe your original problem can be addressed in a different way? Iiuc it's about setting defaults for mutable types. Maybe supporting a `default_factory` would help?

PR #3635, opened 2024-07-17
fix(OpenAPI): handle invalid schema keys

### LINKED ISSUE #3630: Bug: Invalid OpenAPI schema keys generated for generic types
### Description

According to [OpenAPI Specification](https://spec.openapis.org/oas/latest.html#fixed-fields-5), schemas MUST use keys that match the regular expression: `^[a-zA-Z0-9\.\-_]+$`.

When using generics, the generated schema key will contain square brackets.

This is a problem because some tools won't allow invalid OpenAPI schemas to be uploaded.

_No response_

### MCVE

```python

T = TypeVar("T", bound=BaseModel)


class Wrapper(BaseModel, Generic[T]):
    data: T


class Foo(BaseModel):
    bar: str


@get(path="/")
async def foobar() -> Wrapper[Foo]:
    return Wrapper(data=Foo(bar="bar"))


app = Litestar([foobar])
```

### Steps to reproduce

```bash
1. Run the provided MCE
2. Visit `/schema/openapi.json`
3. Notice the schema key generated for the response is `Wrapper[Foo]`


{
  "info": {
    "title": "Litestar API",
    "version": "1.0.0"
  },
  "openapi": "3.1.0",
  "servers": [
    {
      "url": "/"
    }
  ],
  "paths": {
    "/": {
      "get": {
        "summary": "Foobar",
        "operationId": "Foobar",
        "responses": {
          "200": {
            "headers": {

            },
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Wrapper[Foo]"
                }
              }
            }
          }
        },
        "deprecated": false
      }
    }
  },
  "components": {
    "schemas": {
      "Foo": {
        "properties": {
          "bar": {
            "type": "string"
          }
        },
        "type": "object",
        "required": [
          "bar"
        ],
        "title": "Foo"
      },
      "Wrapper[Foo]": {
        "properties": {
          "data": {
            "$ref": "#/components/schemas/Foo"
          }
        },
        "type": "object",
        "required": [
          "data"
        ],
        "title": "Wrapper[Foo]"
      }
    }
  }
}
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.8.2

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3630">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3630/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

thiagozf:
## Description
This PR changes OpenAPI schema key generation function [`_get_normalized_schema_key`](https://github.com/litestar-org/litestar/blob/8c4c15bb501879dabaecfbf0af541ac571c08cf3/litestar/_openapi/schema_generation/utils.py#L86) to use valid characters according to the [OpenAPI specification](https://spec.openapis.org/oas/latest.html#fixed-fields-5), section `4.8.7.1`.

The strategy adopted by this PR was to replace sequences of invalid characters with underscores. Examples:
- `Foo[Bar]` -> `Foo_Bar`
- `dict[str, int]` -> `dict_str_int_`

Some tests had to be changed, because they were expecting the use of invalid characters (like square brackets, spaces and colons).

## Closes
Closes #3630

### DIFF
--- a/litestar/_openapi/datastructures.py
+++ b/litestar/_openapi/datastructures.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from collections import defaultdict
 from typing import TYPE_CHECKING, Iterator, Sequence, _GenericAlias  # type: ignore[attr-defined]
 
@@ -11,6 +12,9 @@ if TYPE_CHECKING:
     from litestar.openapi import OpenAPIConfig
     from litestar.plugins import OpenAPISchemaPluginProtocol
     from litestar.typing import FieldDefinition
+
+
+INVALID_KEY_CHARACTER_PATTERN = re.compile(r"[^a-zA-Z0-9._-]+")
 
 
 def _longest_common_prefix(tuples_: list[tuple[str, ...]]) -> tuple[str, ...]:
@@ -61,7 +65,7 @@ def _get_normalized_schema_key(field_definition: FieldDefinition) -> tuple[str, 
     module = getattr(annotation, "__module__", "")
     name = str(annotation)[len(module) + 1 :] if isinstance(annotation, _GenericAlias) else annotation.__qualname__
     name = name.replace(".<locals>.", ".")
-    return *module.split("."), name
+    return *module.split("."), re.sub(INVALID_KEY_CHARACTER_PATTERN, "_", name)
 
 
 class RegisteredSchema:
--- a/tests/unit/test_dto/test_factory/test_integration.py
+++ b/tests/unit/test_dto/test_factory/test_integration.py
@@ -999,7 +999,7 @@ app = Litestar(route_handlers=[get_users])
 """
     )
     openapi = cast("Litestar", module.app).openapi_schema
-    schema = openapi.components.schemas["WithCount[litestar.dto._backend.GetUsersUserResponseBody]"]
+    schema = openapi.components.schemas["WithCount_litestar.dto._backend.GetUsersUserResponseBody_"]
     assert not_none(schema.properties).keys() == {"count", "data"}
     model_schema = openapi.components.schemas["GetUsersUserResponseBody"]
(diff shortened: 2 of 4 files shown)


provinzkraut:
Thanks for the PR @thiagozf!

I'm not sure if we can accept this into a 2.x release though, as it will break the schema for users. While it's technically a bug fix and makes Litestar adhere stricter to the spec, a lot of the existing OpenAPI tooling does work with these forbidden characters, and if a user relied on them being present (i.e. performing a key lookup somewhere), it would break their use case.

I'm leaning towards targeting 3.0 with this, but let's hear what the other @litestar-org/maintainers think.

provinzkraut (changes requested):
Thanks for the PR @thiagozf!

I'm not sure if we can accept this into a 2.x release though, as it will break the schema for users. While it's technically a bug fix and makes Litestar adhere stricter to the spec, a lot of the existing OpenAPI tooling does work with these forbidden characters, and if a user relied on them being present (i.e. performing a key lookup somewhere), it would break their use case.

I'm leaning towards targeting 3.0 with this, but let's hear what the other @litestar-org/maintainers think.

provinzkraut:
@thiagozf How about we merge this into the v3.0 branch, and for 2.x, we instead issue warning that an invalid key is generated? The user could then override the key themselves. Not a perfect solution, but a workaround.

provinzkraut (approved):
Alright, after thinking about it for a bit, I think considering this a bug is more appropriate, and since the changes are within the references, this shouldn't really cause a breaking change in the schema, as, if you follow the references, things still work as expected.

PR #3639, opened 2024-07-21
fix(connection): Fix creation of FormMultiDict in Request.form to properly handle multi-keys

### LINKED ISSUE #3627: Bug: FormMultiDict is incorrectly created
### Description

When a multipart encoded message with multiple values for a key is posted, the resulting `FormMultiDict` from `request.form()` does not have multiple values per key, but just a list of the values as a single value.

https://github.com/litestar-org/litestar/blob/f9b85f9a362c15873921e7368e20313c4402a00c/litestar/_multipart.py#L96-L101
returns a dict where the values are either single values or list of values.

https://github.com/litestar-org/litestar/blob/f9b85f9a362c15873921e7368e20313c4402a00c/litestar/connection/request.py#L213-L217
https://github.com/litestar-org/litestar/blob/f9b85f9a362c15873921e7368e20313c4402a00c/litestar/connection/request.py#L227
This creates the multidict from the parsed dict, but according to https://multidict.aio-libs.org/en/stable/multidict/#multidict when multiple values for the same key are present the dict must be constructed with a list of `(key, value)` tuples.

I did this with the `data` parameter. 

At least the `request.form()` function should be changed. But it might be better to also change `parse_multipart_form` to return a list of tuples, so a `FormMultiDict` can be created directly from the `data` parameter.

_No response_

### MCVE

```python
from typing import Annotated, Any
from litestar import Litestar, post, Request
from litestar.datastructures import FormMultiDict

@post("/")
async def test_handler(
    data: Annotated[Any, Body(media_type=RequestEncodingType.MULTI_PART)],
    request: Request
) -> None:
    items = []
    for k,v in data.items():
        if isinstance(v, list):
            items.extend((k, i) for i in v)
        else:
            items.append((k, v))
    print("multidict items", items)
    data = FormMultiDict(items)
    print("FormMultiDict items", list(data.items()))
    print("First item", data["key1"], "All items", data.getall("key1"))

    data = await request.form()
    print("request.form items", list(data.items()))
    print("First item", data["key1"], "All items", data.getall("key1"))


```

### Steps to reproduce

```bash
1. Run `LITESTAR_APP=multidict_bug:app .venv/bin/litestar run --debug`
2. Run `curl -X POST http://127.0.0.1:8000/ -F key1=value1 -F key1=value2`
3. See log output
```

### Screenshots

_No response_

### Logs

```bash
multidict items [('key1', 'value1'), ('key1', 'value2')]
FormMultiDict items [('key1', 'value1'), ('key1', 'value2')]
First item value1 All items ['value1', 'value2']
request.form items [('key1', ['value1', 'value2'])]
First item ['value1', 'value2'] All items [['value1', 'value2']]
INFO:     127.0.0.1:35256 - "POST / HTTP/1.1" 201 Created
```

### Litestar Version

2.9.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
Fix #3627 by properly handling the creation of `FormMultiDict` where multiple values are given for a single key, to make `Request.form()` match the behaviour of receiving form data via the `data` kwarg.

<hr>

**Before**

```python
@post("/")
async def handler(request: Request) -> Any:
    return (await request.form()).getall("foo")

    print(client.post("/", data={"foo": ["1", "2"]}).json()) # [["1", "2"]]
```

**After**

```python
@post("/")
async def handler(request: Request) -> Any:
    return (await request.form()).getall("foo")

    print(client.post("/", data={"foo": ["1", "2"]}).json()) # ["1", "2"]
```

### DIFF
--- a/litestar/connection/request.py
+++ b/litestar/connection/request.py
@@ -74,7 +74,7 @@ class Request(Generic[UserT, AuthT, StateT], ASGIConnection["HTTPRouteHandler", 
         super().__init__(scope, receive, send)
         self.is_connected: bool = True
         self._body: bytes | EmptyType = Empty
-        self._form: dict[str, str | list[str]] | EmptyType = Empty
+        self._form: FormMultiDict | EmptyType = Empty
         self._json: Any = Empty
         self._msgpack: Any = Empty
         self._content_type: tuple[str, dict[str, str]] | EmptyType = Empty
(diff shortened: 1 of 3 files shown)


cofin (approved):
LGTM

PR #3661, opened 2024-08-12
fix: Small docs update.

Zimzozaur:
Starting from version 2, Pydantic uses `model_validate()` instead of `parse_obj()`. 
https://docs.pydantic.dev/2.8/migration/#changes-to-pydanticbasemodel

I have updated 2 places in the docs where the old deprecated method is used.


JacobCoffee (approved):
thanks :)

JacobCoffee:
@all-contributors add @Zimzozaur docs

PR #3669, opened 2024-08-15
chore: update dependencies

provinzkraut:
Update package dependencies :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ default_language_version:
   python: "3.8"
 repos:
   - repo: https://github.com/compilerla/conventional-pre-commit
-    rev: v3.2.0
+    rev: v3.4.0
     hooks:
       - id: conventional-pre-commit
         stages: [commit-msg]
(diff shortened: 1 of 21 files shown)


Alc-Alc (approved):

provinzkraut:
@JacobCoffee Same sphinx related issues here as well. I have no idea where this is coming from. The sphinx version hasn't been updated, and I've manually tried downgrading every sphinx-related package we have..

provinzkraut:
@cofin AA PR https://github.com/litestar-org/advanced-alchemy/pull/229 seems to break some test, so I've pinned added a constraint for `<=0.17.1`. We should investigate. I believe it has something to do with this in particular: https://github.com/litestar-org/advanced-alchemy/blame/e39430e0a9cd2f3d86eb604cd305e8b58ae8d2b8/advanced_alchemy/extensions/litestar/plugins/init/config/sync.py#L163

provinzkraut:
I've also added a constraing for [`psycopg<3.2`](https://www.psycopg.org/psycopg3/docs/news.html#psycopg-3-2), as it's breaking something in the channels backend (unsubscribes seem to hang). We'll need to investigate there.

PR #3671, opened 2024-08-16
docs: Fix misplaced comment in dependency-injection.rst for better clarity

Zimzozaur:
The comment "on the router" was incorrectly placed below the `my_route_handler()`, which could cause confusion. I've moved the comment to the appropriate location above the router configuration for better clarity. 

No functional changes were made.


provinzkraut (approved):
Thanks @Zimzozaur

PR #3685, opened 2024-08-22
fix(DTO): Inconsistent use of strict mode 

provinzkraut:
Fix inconsistent usage of msgspec's `strict` mode in the base DTO backend.

`strict=False` was being used when transferring from builtins, while `strict=True` was used transferring from raw data, causing an unwanted discrepancy in behaviour,

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -232,9 +232,9 @@ class DTOBackend:
         type_decoders = asgi_connection.route_handler.resolve_type_decoders()
 
         if request_encoding == RequestEncodingType.MESSAGEPACK:
-            result = decode_msgpack(value=raw, target_type=self.annotation, type_decoders=type_decoders)
+            result = decode_msgpack(value=raw, target_type=self.annotation, type_decoders=type_decoders, strict=False)
         else:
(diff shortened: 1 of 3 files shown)


cofin (approved):
LGTM

gsakkis:
After upgrading from 2.10 to 2.11 I got an error probably from this change, `msgspec.ValidationError: Object contains unknown field "id"`. The `id` field is defined in a superclass of the model I annotate with `DTOData`. Does `DTOData[Foo]` support intentionally only the fields defined in `Foo` and not its parent(s)?

provinzkraut:
> After upgrading from 2.10 to 2.11 I got an error probably from this change, `msgspec.ValidationError: Object contains unknown field "id"`. The `id` field is defined in a superclass of the model I annotate with `DTOData`. Does `DTOData[Foo]` support intentionally only the fields defined in `Foo` and not its parent(s)?

Definitely not intentional, but I'm also not sure it's is related to this PR. Can you provide a simple reproducer?

gsakkis:
Ah nevermind, I misdiagnosed the problem: `id` is excluded from the respective `DTO` class but before 2.11  it was silently ignored when being passed in; in 2.11 (most likely due to this PR) it results in error, as it should.

PR #3691, opened 2024-08-24
feat(DTO): Support `extra="forbid"` model config for `PydanticDTO`

provinzkraut:
Set `forbid_unkown_fields=True` for `PydanticDTOs` where the Pydantic model has an [`extra="forbid"`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra) config.

<hr>

- Add a new `get_config_for_model_type` method to `AbstractDTO`, that allows to customise the base config defined on the DTO factory for a specific model type
- Use `get_config_for_model_type` to set `forbid_unkown_fields=True` for Pydantic models that use the `extra="forbid" config

<hr>

Depends on #3690. Don't merge before that one :)

### DIFF
--- a/litestar/contrib/pydantic/pydantic_dto_factory.py
+++ b/litestar/contrib/pydantic/pydantic_dto_factory.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
+import dataclasses
 from dataclasses import replace
 from typing import TYPE_CHECKING, Any, Collection, Generic, TypeVar
 from warnings import warn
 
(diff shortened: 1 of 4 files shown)


provinzkraut:
@peterschutt to keep you in the loop a bit. This isn't *really* merging configurations, but it's a step in the right direction and at least allows per-model dynamic configs. I think this should help to handle a few of the use cases of config merging already.

peterschutt:
> I think this should help to handle a few of the use cases of config merging already.

Sounds great! Thanks for the ping:)

PR #3696, opened 2024-08-24
fix: inconsistent behavior between `signature_namespace` and `signature_types`

### LINKED ISSUE #3681: bug: inconsistent behavior between `signature_namespace` and `signature_types`
It's currently difficult to use the simpler `signature_types` syntax in plugins due to the potential of there being an exception raised during startup. 

We should consider removing this entire section to keep the behavior consistent between `signature_namespace` and `signature_types`.

https://github.com/litestar-org/litestar/blob/238d26bab30afa1f852ef0d88830077aab77ba35/tests/unit/test_utils/test_signature.py#L159-L168

---
> [!NOTE]  
>

<a href="https://polar.sh/litestar-org/litestar/issues/3681">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3681/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

cofin:
Fixes #3681

## Description

-

Closes #3681

### DIFF
--- a/litestar/utils/signature.py
+++ b/litestar/utils/signature.py
@@ -11,7 +11,6 @@ from typing import TYPE_CHECKING, Any, Union
 from typing_extensions import Annotated, Self, get_args, get_origin, get_type_hints
 
 from litestar import connection, datastructures, types
-from litestar.exceptions import ImproperlyConfiguredException
 from litestar.types import Empty
 from litestar.typing import FieldDefinition
 from litestar.utils.typing import expand_type_var_in_type_hint, unwrap_annotation
@@ -20,6 +19,7 @@ if TYPE_CHECKING:
     from typing import Sequence
 
     from litestar.types import AnyCallable
+    from litestar.types.protocols import Logger
 
 if sys.version_info < (3, 11):
     from typing import _get_defaults  # type: ignore[attr-defined]
@@ -244,7 +244,7 @@ class ParsedSignature:
 
 
 def add_types_to_signature_namespace(
-    signature_types: Sequence[Any], signature_namespace: dict[str, Any]
+    signature_types: Sequence[Any], signature_namespace: dict[str, Any], logger: Logger | None = None
 ) -> dict[str, Any]:
     """Add types to ith signature namespace mapping.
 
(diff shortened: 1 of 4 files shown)


provinzkraut:
So, this wouldn't warn by default? 
Is there a reason you went with this logger approach instead of the standard warning stuff we usually do?

litestar/utils/signature.py
```diff
@@ -253,16 +254,46 @@ def add_types_to_signature_namespace(
...
-            raise ImproperlyConfiguredException(f"Type '{name}' is already defined in the signature namespace")
-        signature_namespace[name] = typ
+    return merge_signature_namespaces(
+        signature_namespace=signature_namespace,
+        additional_signature_namespace={signature_type.__name__: signature_type for signature_type in signature_types},
+        logger=logger,
+    )
+
+
+def merge_signature_namespaces(
```
provinzkraut, line 272: Should use this in `BaseRouteHandler.resolve_signature_namespace` as well for consistency :)

cofin:
> So, this wouldn't warn by default? Is there a reason you went with this logger approach instead of the standard warning stuff we usually do?

Yes.  I really don't think this should be a warning at all.  It should just be something the user is aware of.

For instance, if a plugin sets something automatically and then the user sets their own.  I think we should just notify the user that there could be a problem.  I wouldn't consider this an actual warning.

provinzkraut:
>  I think we should just notify the user that there could be a problem. I wouldn't consider this an actual warning.

I'm with you there, but we aren't doing that, because we don't pass a logger by default, right? 
I feel like I'm missing something here :grimacing:

tests/unit/test_utils/test_signature.py
```diff
@@ -156,16 +159,29 @@ def test_add_types_to_signature_namespace() -> None:
     assert ns == {"int": int, "str": str}
 
 
-def test_add_types_to_signature_namespace_with_existing_types() -> None:
+def test_add_types_to_signature_namespace_no_warn() -> None:
     """Test add_types_to_signature_namespace with existing types."""
-    ns = add_types_to_signature_namespace([str], {"int": int})
-    assert ns == {"int": int, "str": str}
+    if os.environ.get("LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE", None):
+        del os.environ["LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE"]
```
provinzkraut, line 165:
```suggestion
    monkeypatch.delenv("LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE", raising=False)
```

So it's resets things properly. Easier to reason about IMO

tests/unit/test_utils/test_signature.py
```diff
@@ -156,16 +159,29 @@ def test_add_types_to_signature_namespace() -> None:
...
+        add_types_to_signature_namespace([int], {"int": int})
 
 
-def test_add_types_to_signature_namespace_with_existing_types_raises() -> None:
+def test_add_types_to_signature_namespace_with_existing_types_warn() -> None:
     """Test add_types_to_signature_namespace with existing types raises."""
-    with pytest.raises(ImproperlyConfiguredException):
-        add_types_to_signature_namespace([int], {"int": int})
+    if os.environ.get("LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE", None):
+        del os.environ["LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE"]
```
provinzkraut, line 174: Same here

tests/unit/test_utils/test_signature.py
```diff
@@ -156,16 +159,29 @@ def test_add_types_to_signature_namespace() -> None:
...
+    with pytest.warns(LitestarWarning):
+        add_types_to_signature_namespace([int], {"int": str})
+
+
+def test_add_types_to_signature_namespace_warn_disabled() -> None:
+    """Test add_types_to_signature_namespace with existing types."""
+    os.environ["LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE"] = "0"
```
provinzkraut, line 181:
```suggestion
    monkeypatch.setenv("LITESTAR_WARN_SIGNATURE_NAMESPACE_OVERRIDE")
```

provinzkraut (approved):
Great! Just one more comment regarding the env test setup

Issue #3688, opened 2024-08-23
Enhancement: no way to disable csrf and session middleware on some router


trim21:
### Summary

I have a simple router try to serve a simple badge to all users, and I hope CDN can cache it, so I need to remove all set-cookies header on this router.

But I have session middleware and csrf middleware, I try `exclude_from_csrf_key` and `exclude`, non of them work for this case.

Is there a way I can disable key generation on some routers?

### Basic Example

```python
import asyncio
from pprint import pprint
from typing import Any

import httpx
import litestar
from litestar.config.csrf import CSRFConfig
from litestar.connection import ASGIConnection
from litestar.middleware import AuthenticationResult
from litestar.middleware.session.server_side import ServerSideSessionConfig
from litestar.security.session_auth import SessionAuth, SessionAuthMiddleware


@litestar.get("/badge.svg", sync_to_thread=False, exclude_from_csrf=True)
def handler() -> str:
    return "hello world"


class MyAuthenticationMiddleware(SessionAuthMiddleware):
    async def authenticate_request(
        self, connection: ASGIConnection[Any, Any, Any, Any]
    ) -> AuthenticationResult:
        if not connection.session or connection.scope["session"] is Empty:
            # the assignment of 'Empty' forces the session middleware to clear session data.
            return AuthenticationResult(user=None, auth=None)

        user = {"user": 1}

        return AuthenticationResult(user=user, auth=user)


session_auth_config = SessionAuth[Any, Any](
    retrieve_user_handler=lambda x: {"user": 1},
    session_backend_config=ServerSideSessionConfig(),
    authentication_middleware_class=MyAuthenticationMiddleware,
    exclude=[r"/badge\.svg"]
)
app = litestar.Litestar(
    [handler],
    csrf_config=CSRFConfig(secret="secret-key", exclude=[r"/badge\.svg"]),
    middleware=[session_auth_config.middleware],
)


async def main():
    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=app)) as c:
        res = await c.get("http://example.com/badge.svg")
        pprint(dict(res.headers))  # with 'set-cookie'


asyncio.run(main())

```

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
[...]

PR #3698, opened 2024-08-25
fix: csrf middleware excluding router

trim21:
## Description

- csrf middleware should skip router entirely on match exclude config

close https://github.com/litestar-org/litestar/issues/3688

### DIFF
--- a/tests/unit/test_middleware/test_csrf_middleware.py
+++ b/tests/unit/test_middleware/test_csrf_middleware.py
@@ -236,10 +236,12 @@ def test_csrf_middleware_exclude_from_check() -> None:
         data = {"field": "value"}
         response = client.post("/protected-handler", data=data)
         assert response.status_code == HTTP_403_FORBIDDEN
+        assert "set-cookie" in response.headers
 
         response = client.post("/unprotected-handler", data=data)
         assert response.status_code == HTTP_201_CREATED
         assert response.json() == data
+        assert "set-cookie" not in response.headers
 
 
 def test_csrf_middleware_configure_name_for_exclude_from_check_via_opts() -> None:
--- a/tests/unit/test_middleware/test_csrf_middleware.py
+++ b/tests/unit/test_middleware/test_csrf_middleware.py
@@ -244,7 +244,6 @@ def test_csrf_middleware_exclude_from_check() -> None:
         data = {"field": "value"}
         response = client.post("/protected-handler", data=data)
         assert response.status_code == HTTP_403_FORBIDDEN
-        assert "set-cookie" in response.headers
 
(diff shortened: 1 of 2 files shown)


trim21:
> Hey @trim21, thanks for your contribution! For a bugfix, please add a test that triggers the bug you are fixing. This prevents regression, clearly indicates the change in behaviour, and makes it easier to reproduce.

Is new commits fine?

tests/unit/test_middleware/test_csrf_middleware.py
```diff
@@ -241,6 +249,18 @@ def post_handler2(data: dict = Body(media_type=RequestEncodingType.URL_ENCODED))
         assert response.status_code == HTTP_201_CREATED
         assert response.json() == data
 
+    with create_test_client(
```
provinzkraut, line 252: This should be a separate test, with a descriptive name. You can also add a link to the GH issue as a comment there. That makes it easier to track why things are being tested

provinzkraut (changes requested):

provinzkraut:
> Is new commits fine?

Yes. We squash merge PRs anyway, so commit as you like :)

cofin:
Looks good to me!  We can merge after the comments are addressed.

@all-contributors add @trim21 for code and tests

cofin (approved):

trim21:
done

tests/unit/test_middleware/test_csrf_middleware.py
```diff
@@ -242,6 +242,28 @@ def post_handler2(data: dict = Body(media_type=RequestEncodingType.URL_ENCODED))
         assert response.json() == data
 
 
+def test_csrf_middleware_exclude_from_set_cookies() -> None:
+    @get("/protected-handler")
```
provinzkraut, line 246:
```suggestion
def test_csrf_middleware_exclude_from_set_cookies() -> None:
    #  https://github.com/litestar-org/litestar/issues/3688
    #  middleware should be bypassed completely when excluded, so no cookies should be set
    
    @get("/protected-handler")
```

For future reference

provinzkraut (approved):

trim21:
does litestar have a release cycle? when this PR may be release?

provinzkraut:
> does litestar have a release cycle? when this PR may be release?

We make a minor release roughly about once a month, and patch releases as needed.
A minor release will happen within the next few days, so this will be included there

trim21:
> > does litestar have a release cycle? when this PR may be release?
> 
> We make a minor release roughly about once a month, and patch releases as needed. A minor release will happen within the next few days, so this will be included there

thanks

PR #3712, opened 2024-08-29
fix(warnings): the exclude pattern matching warned too much

euri10:
## Description

- I added a test that doesnt pass with the current implementation, that is wen you exclude `^/$`

### DIFF
--- a/litestar/middleware/_utils.py
+++ b/litestar/middleware/_utils.py
@@ -33,7 +33,8 @@ def build_exclude_path_pattern(
 
     try:
         pattern = re.compile("|".join(exclude)) if isinstance(exclude, list) else re.compile(exclude)
-        if pattern.match("/") and exclude != "^/$":
+        if pattern.match("/") and pattern.match("/982c7064-6ac7-44b7-9be5-07a2ff6d8a92"):
+        # match a UUID to ensure that it matches paths greedily and not just a literal /
             warn_middleware_excluded_on_all_routes(pattern, middleware_cls=middleware_cls)
         return pattern
 
--- a/litestar/middleware/_utils.py
+++ b/litestar/middleware/_utils.py
@@ -32,11 +32,10 @@ def build_exclude_path_pattern(
         return None
 
     try:
-        exclude_list = [exclude] if isinstance(exclude, str) else exclude
-        for el in exclude_list:
-            if el == "/":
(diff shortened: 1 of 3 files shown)


litestar/middleware/_utils.py
```diff
@@ -32,10 +32,11 @@ def build_exclude_path_pattern(
         return None
 
     try:
-        pattern = re.compile("|".join(exclude)) if isinstance(exclude, list) else re.compile(exclude)
-        if pattern.match("/"):
-            warn_middleware_excluded_on_all_routes(pattern, middleware_cls=middleware_cls)
-        return pattern
+        exclude_list = [exclude] if isinstance(exclude, str) else exclude
+        for el in exclude_list:
+            if el == "/":
```
provinzkraut, line 37: Hm. I think this has the opposite problem of being too lax, as it would e.g. not warn for `/*`. Maybe we extend the original pattern test with another check to verify it's actually greedy, i.e. matches `/` *and* other paths with a leading `/`?

euri10: yes you're right
I reverted and kept the tests, but cant find now a good way to match / and other paths.
I just addded a weird test on the pattern being not ^/$ but finding greedy patterns is context dependant so I'm not sure now how to handle this properly of if it even possible.

cofin (approved):

litestar/middleware/_utils.py
provinzkraut: Yeah, I don't think it's possible to do this 100% reliably. But it is just a warning after all, so maybe that's okay. How about testing if the pattern matches `/` *and* `/` followed by something randomly unique? That way we could be somewhat certain that it would greedily match?

litestar/middleware/_utils.py
```diff
@@ -33,9 +33,10 @@ def build_exclude_path_pattern(
 
     try:
         pattern = re.compile("|".join(exclude)) if isinstance(exclude, list) else re.compile(exclude)
-        if pattern.match("/"):
+        if pattern.match("/") and exclude != "^/$":
```
provinzkraut, line 36:
```suggestion
        if pattern.match("/") and pattern.match("/982c7064-6ac7-44b7-9be5-07a2ff6d8a92"):
        # match a UUID to ensure that it matches paths greedily and not just a literal /
```

provinzkraut (approved):

PR #3738, opened 2024-09-14
feat: Allow customizing schema component keys

provinzkraut:
Allow customizing the schema key used for a component in the OpenAPI schema. The motivation is to give the user an escape hatch when we inevitably generate some less than pretty looking names for deeply nested types for the sake of uniqueness.

These supplied key are enforced to be unique, and it is checked that they won't be reused across different types.

<hr>

- Add a new `schema_component_key` to `KwargDefinition` and the `Body` and `Parameter` functions
- Add a friendly error message when keys are not unique / reused across different types

### DIFF
--- a/litestar/_openapi/datastructures.py
+++ b/litestar/_openapi/datastructures.py
@@ -96,7 +96,7 @@ class SchemaRegistry:
         self._schema_key_map: dict[tuple[str, ...], RegisteredSchema] = {}
         self._schema_reference_map: dict[int, RegisteredSchema] = {}
         self._model_name_groups: defaultdict[str, list[RegisteredSchema]] = defaultdict(list)
-        self._component_type_map: dict[str, FieldDefinition] = {}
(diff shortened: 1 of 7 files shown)


litestar/_openapi/datastructures.py
```diff
@@ -1,14 +1,67 @@
...
 from litestar.openapi.spec import Reference, Schema
+from litestar.params import KwargDefinition
 
 if TYPE_CHECKING:
     from litestar.openapi import OpenAPIConfig
     from litestar.plugins import OpenAPISchemaPluginProtocol
+    from litestar.typing import FieldDefinition
+
+
+def _longest_common_prefix(tuples_: list[tuple[str, ...]]) -> tuple[str, ...]:
```
provinzkraut, line 16: Lifted this out so we don't define it inline anymore

Alc-Alc (approved):

PR #3740, opened 2024-09-15
feat: Raise exception when body parameter is annotated with non-bytes type

provinzkraut:
Add an informative error message to help users avoid the common mistake of attempting to use the `body` parameter to receive validated / structured data, by annotating it with a type such as `list[str]`, instead of `bytes`. In such a case, the user will now be instructed to use the `data` parameter instead.

### DIFF
--- a/tests/unit/test_handlers/test_http_handlers/test_validations.py
+++ b/tests/unit/test_handlers/test_http_handlers/test_validations.py
@@ -1,12 +1,13 @@
 from pathlib import Path
 from types import ModuleType
-from typing import Callable, Dict
+from typing import Annotated, Any, Callable, Dict
 
 import pytest
 
-from litestar import HttpMethod, Litestar, WebSocket, delete, get, route
+from litestar import HttpMethod, Litestar, WebSocket, delete, get, patch, post, put, route
 from litestar.exceptions import ImproperlyConfiguredException, ValidationException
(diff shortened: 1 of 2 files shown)


tests/unit/test_handlers/test_http_handlers/test_validations.py
```diff
@@ -144,3 +146,18 @@ def no_response_handler() -> {return_annotation}:
...
+def test_body_param_with_non_bytes_annotation_raises(decorator: Callable[..., Any]) -> None:
+    def handler_fn(body: List[str]) -> None:
+        pass
+
+    with pytest.raises(ImproperlyConfiguredException, match="Invalid type annotation for 'body' parameter"):
+        Litestar([decorator()(handler_fn)])
+
+
+@pytest.mark.parametrize("decorator", [post, put, patch])
+def test_body_param_with_metadata_allowed(decorator: Callable[..., Any]) -> None:
```
Alc-Alc, line 161: why is `decorator` paremeterized but not used? what is the behavior this is testing?

provinzkraut: Looks like I forgot to commit something here :)

provinzkraut: Seems like I forgot to commit something :)

Alc-Alc (approved):

PR #3748, opened 2024-09-21
feat: allow before_request and after_request handlers to accept a parent argument

charles-dyfis-net:
…to wrap or override the handler from an enclosing scope.

At present, this is a proposal / discussion piece rather than a contribution submitted as ready-to-merge.

## Description

Presently, hooks at an inner layer replace those at outer layers. In many use cases, this is unsuitable: If the router handling all `/tenants/{tenant:uuid}` requests performs a lookup and authentication, then a router mounted under it will still want that same authentication even if it adds additional common code as well.

This PR proposes a change to allow child handlers to _wrap_ parent handlers. Where previously one might write:

```python
def tenant_router_pre_request(request: litestar.Request):
    if (tenant := request.path_params.get("tenant")) is not None:
        structlog.contextvars.bind_contextvars(tenant=str(tenant))
    # author needs to know that the parent app's pre_request was this specific other function
    return await root_app_pre_request(request)
```

...we can instead write:

```python
def tenant_router_pre_request(request: litestar.Request, /, *, parent: Optional[BeforeRequestHookHandler]):
    if (tenant := request.path_params.get("tenant")) is not None:
        structlog.contextvars.bind_contextvars(tenant=str(tenant))
    return (await parent(request)) if parent is not None else None
```

...and let litestar do the work of populating `parent` as/when appropriate.


charles-dyfis-net:
With test coverage added and passing, removing from draft.

(Having discussed with @provinzkraut, I understand that this isn't likely to be merged in its current form -- we can do better in 3.0 by building a list of hooks as we do guards, and pre-3.0 this change is considered to break compatibility guarantees -- but I did want to get this into a less-broken state before being done with it, even as a conversation piece).

provinzkraut:
Closing this as a different solution has been merged into 3.0, and we're planning to re-visit hooks entirely

PR #3753, opened 2024-09-21
chore: Fix PyPI readme generation on Windows

floxay:
Currently `tools/pypi_readme.py` (and with it pre-commit too) is failing on Windows because Python uses CP1252 by default to open text files (and will continue to do so until Python 3.15(?); [PEP 686 – Make UTF-8 mode default](https://peps.python.org/pep-0686/)), this PR sets the file encoding for the README files in `tools/pypi_readme.py` to UTF-8.


Alc-Alc (approved):
Thank you 🙂 

PS: ~~Python3.15~~ Python3.26

PR #3754, opened 2024-09-21
feat: OpenAPI plugins send CSRF request header

floxay:
Supported OpenAPI UI clients will extract the CSRF cookie value and attach it to the request headers if CSRF is enabled on the application.

- [x] Swagger
- [x] RapiDoc
- [ ] Scalar
- [ ] Redoc
- [ ] Stoplight Elements

Scalar (https://github.com/scalar/scalar/discussions/2810), Redoc and (Stoplight) Elements does not seem to support interceptors or some other mechanism to achieve this.

Currently it also sends the header on "safe" methods but for both currently supported UI clients (Swagger and RapiDoc) it's possible to limit this to "unsafe" methods only, if required.

### DIFF
--- a/litestar/openapi/plugins.py
+++ b/litestar/openapi/plugins.py
@@ -12,6 +12,7 @@ from litestar.handlers import get
 from litestar.serialization import encode_json, get_serializer
 
 if TYPE_CHECKING:
+    from litestar.config.csrf import CSRFConfig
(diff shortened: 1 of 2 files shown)


cofin (approved):
Thanks for these changes!

cofin:
@all-contributors add @floxay for code and tests

PR #3776, opened 2024-10-03
fix: sign bug in rate limit middelware

pogopaule:
## Description
When using the rate limit middleware, the response header fields `RateLimit-Remaining` and `RateLimit-Reset` have negative values. This branch fixes the issue.

This might be a breaking change for consumers of an API implemented with litestar.

I split the PR into four commits which might make it easier to review this PR.

Side note: Initially my commits didn't follow the conventional commits guidelines, but pre-commit didn't complain. See point 6 in https://github.com/litestar-org/litestar/blob/main/CONTRIBUTING.rst#workflow
Is this a bug in the pre-commit setup?

### DIFF
--- a/tests/unit/test_middleware/test_rate_limit_middleware.py
+++ b/tests/unit/test_middleware/test_rate_limit_middleware.py
@@ -44,8 +44,11 @@ async def test_rate_limiting(unit: DurationUnit) -> None:
         assert response.headers.get(config.rate_limit_policy_header_key) == f"1; w={DURATION_VALUES[unit]}"
         assert response.headers.get(config.rate_limit_limit_header_key) == "1"
         assert response.headers.get(config.rate_limit_remaining_header_key) == "0"
-        assert response.headers.get(config.rate_limit_reset_header_key) == str(int(time()) - cache_object.reset)
+        # Since the time is frozen, no time has passed.
+        # Hence the remaining seconds for the current quota window should be the same as the whole window length.
+        assert response.headers.get(config.rate_limit_reset_header_key) == str(DURATION_VALUES[unit])
 
+        # Move time one second before the end of the quota window
         frozen_time.shift(DURATION_VALUES[unit] - 1)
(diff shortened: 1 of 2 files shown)


pogopaule:
@all-contributors please add @pogopaule for code

tests/unit/test_middleware/test_rate_limit_middleware.py
```diff
@@ -28,33 +28,46 @@ async def test_rate_limiting(unit: DurationUnit) -> None:
...
     store = app.stores.get("rate_limit")
 
     with travel(datetime.utcnow, tick=False) as frozen_time, TestClient(app=app) as client:
         response = client.get("/")
-        assert response.status_code == HTTP_200_OK
+
         cached_value = await store.get(cache_key)
         assert cached_value
         cache_object = CacheObject(**decode_json(value=cached_value))
         assert len(cache_object.history) == 1
```
pogopaule, line 42: IMO, this block code is unnecessary because it tests implementation details. WDYT?

provinzkraut: While it does test implementation details, that's intentional here, since it ensure e.g. that we don't accidentally save more things in the history than we intend to

provinzkraut (approved):

provinzkraut:
> Side note: Initially my commits didn't follow the conventional commits guidelines, but pre-commit didn't complain. See point 6 in https://github.com/litestar-org/litestar/blob/main/CONTRIBUTING.rst#workflow Is this a bug in the pre-commit setup?

Oh, that's fine. We squash merge everything, and only check that these commits follow conventional commit guidelines. Within a PR branch you can do whatever you want

provinzkraut:
Thanks @pogopaule, nice seeing you here :)

pogopaule:
> Thanks @pogopaule, nice seeing you here :)

👋🙂

PR #3784, opened 2024-10-11
fix(prometheus): fix path template for routes without path parameters

mohammedbabelly20:
## Description

When `PrometheusConfig.group_path=True` the metrics exporter response content ignores all paths with no path parameters.

- Example for path `/test/litestar` when `group_path=True`

    > litestar_request_duration_seconds_bucket{app_name="litestar",le="0.075",method="GET",path="",status_code="200"} 1.0

- Example for path `/test/litestar` when `group_path=False`

    > litestar_request_duration_seconds_bucket{app_name="litestar",le="2.5",method="GET",path="/test/litestar",status_code="200"} 1.0

So, when the path has no params return the actual path as template so Prometheus can use it when `group_path=True`.

### QA
Added more test cases to cover this case.

### DIFF
--- a/tests/examples/test_contrib/prometheus/test_prometheus_exporter_example.py
+++ b/tests/examples/test_contrib/prometheus/test_prometheus_exporter_example.py
@@ -20,7 +20,6 @@ def clear_collectors() -> None:
 @pytest.mark.parametrize(
     "group_path, route_path, route_template, expected_path",
     [
-        (True, "/test", "test", "/test"),
         (True, "/test/litestar", "test/litestar", "/test/litestar"),
         (False, "/test/litestar", "test/litestar", "/test/litestar"),
         (True, "/test/litestar", "test/{name:str}", "/test/{name}"),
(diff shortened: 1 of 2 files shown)


tests/examples/test_contrib/prometheus/test_prometheus_exporter_example.py
```diff
@@ -20,7 +20,8 @@ def clear_collectors() -> None:
 @pytest.mark.parametrize(
     "group_path, route_path, route_template, expected_path",
     [
-        (True, "/test/litestar", "test/{name:str}", "/test/{name}"),
```
provinzkraut, line 23: The removed test case should still pass, otherwise we'd have to consider this a breaking change

mohammedbabelly20: It does pass.
It was a duplicate test case, see https://github.com/litestar-org/litestar/blob/36d3faa8109d90de2a714e9cc0b21cf0334b04d9/tests/examples/test_contrib/prometheus/test_prometheus_exporter_example.py#L25

provinzkraut (approved):

provinzkraut:
@all-contributors add @mohammedbabelly20 for code

PR #3803, opened 2024-10-15
docs: mention `id` field in `1-abstract-dto.rst`

sobolevn:
It caught my attention that it was never explained: why do we have an extra `id` field in the response. Since this part is dedicated to explaining how request / response fields are marked, I think that it is important to mention that `id` is special:
Снимок экрана 2024-10-15 в 18 50 23

I had to grep what `.my_lib` is and how `class _Base(CommonTableAttributes, UUIDPrimaryKey, DeclarativeBase):` is constructed. Only finding `UUIDPrimaryKey` and its `id: Mapped[UUID] = mapped_column(default=uuid4, primary_key=True)` has anwsered my question :)


provinzkraut:
I'm not really sure "special" is the right wording here, since it's simply an inherited attribute. It only happens to be the only one that's inherited from the base class, but there's nothing special about the `id` attribute itself

sobolevn:
@provinzkraut can we say that "id is readonly because it is a primary key"?

provinzkraut:
> @provinzkraut can we say that "id is readonly because it is a primary key"?

I think it should say something like "we mark `id` read only since it represents a primary key", to emphasize that this isn't done automagically, and could apply to other arbitrary fields as well

docs/usage/dto/1-abstract-dto.rst
```diff
@@ -50,6 +50,8 @@ Fields marked as ``"private"`` or ``"read-only"`` will not be parsed from client
     :emphasize-lines: 6,14,15
     :linenos:
 
+Note that ``id`` field is handled specially, since it is a primary key.
```
cofin, line 53:
```suggestion
Note that ``id`` field is the primary key and is handled specially by the defined SQLAlchemy base.
```

cofin (approved):

cofin:
I'm going to go ahead and merge with the provided updates.  We can open a followup PR to dial in the verbiage if needed.

Thanks @sobolevn

PR #3808, opened 2024-10-15
docs: fix highlighted lines in DTO usage

sobolevn:
Link: https://docs.litestar.dev/latest/usage/dto/1-abstract-dto.html#dto-data

It used to be like this:
Снимок экрана 2024-10-15 в 21 52 44

I moved all highlighted lines one line up.


sobolevn:
There are other problems (like mentioning `Person` model that does not exist) in this part, but they are unrelated to this change and will take me some time to fix.

provinzkraut (approved):

PR #3823, opened 2024-10-17
docs: refactor exception usage

sobolevn:
I made quite a lot of changes:
1. I replaced `:doc:` with `:ref:` since it is more reliable
2. I replaced things like

```rst
:class:`HTTPException <.exceptions.http_exceptions.HTTPException>`
```

to be just

```rst
:class:`~litestar.exceptions.http_exceptions.HTTPException`
```

Because it the same thing

3. I changed a lot of wordings to be more readable (for my taste, please, suggest better ones)
4. I changed `.. :currentmodule:: litestar.exceptions.http_exceptions` in two ways: removed `py`, because it is the default role, and reset it after the table. It can be very hard to find why something does not work in the future, when you manipulate current module without reseting it properly
5. I moved exception in the table to be ordered by error code, it used to be:
Снимок экрана 2024-10-17 в 10 29 41
6. I moved the paragraph about handling exceptions in different layers of the app after we introduce the layers concept :)


docs/usage/exceptions.rst
```diff
@@ -49,26 +49,31 @@ Litestar also offers several pre-configured exception subclasses with pre-set er
...
 +----------------------------------------+-------------+------------------------------------------+
 | :class:`ServiceUnavailableException`   | 503         | HTTP status code 503                     |
 +----------------------------------------+-------------+------------------------------------------+
 
-When a value fails ``pydantic`` validation, the result will be a :class:`ValidationException` with the ``extra`` key set to the
-pydantic validation errors. Thus, this data will be made available for the API consumers by default.
+.. :currentmodule:: None
+
+When a value fails ``pydantic`` validation, the result will be a :class:`~litestar.exceptions.http_exceptions.ValidationException` with the ``extra`` key set to the
+``pydantic`` validation error message.
```
provinzkraut, line 66:
```suggestion
When a value fails validation, the result will be a :class:`~litestar.exceptions.http_exceptions.ValidationException` with the ``extra`` key set to the validation error message.
```

docs/usage/exceptions.rst
```diff
@@ -116,3 +113,11 @@ As a result of the above structure, the exceptions raised by the ASGI Router its
 and ``405 Method Not Allowed`` are handled only by exception handlers defined on the app layer. Thus, if you want to affect
 these exceptions, you will need to pass the exception handlers for them to the Litestar constructor and cannot use other
 layers for this purpose.
+
+While it does not make much sense to have different functions with a top-level exception handling,
```
provinzkraut, line 117: I think if you're saying something doesn't make sense but can still be done, you also should explain:

- Why it doesn't make sense
- Why it can still be done
- If / When this should be used

provinzkraut (changes requested):

docs/usage/exceptions.rst
sobolevn: I removed this phrase. For this does make perfect sense: you customize the error handling for different routes.

docs/usage/exceptions.rst
sobolevn: Done!

provinzkraut (approved):

PR #3824, opened 2024-10-17
chore(mypy): enable `truthy-bool` error code

sobolevn:
Docs: https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-expression-is-not-implicitly-true-in-boolean-context-truthy-bool

I am using `# type: ignore[truthy-bool]` for `.components` check, because I see that it is treated specially for some reason (like `pragma: no cover`). Is this correct?

I also removed configs that are part of the `--strict` flag from settings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -247,6 +247,7 @@ xfail_strict = true
 packages = ["litestar", "tests"]
 plugins = ["pydantic.mypy"]
 enable_error_code = [
+  "truthy-bool",
   "truthy-iterable",
   "unused-awaitable",
   "ignore-without-code",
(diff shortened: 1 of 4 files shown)


sobolevn:
Please, take a look at these lines:
Снимок экрана 2024-10-17 в 11 43 46

Looks like `else` is never actually reached. Should I remove this branch?

pyproject.toml
```diff
@@ -253,21 +254,19 @@ enable_error_code = [
...
 disallow_any_generics = false
-disallow_untyped_decorators = true
-implicit_reexport = false
 show_error_codes = true
 strict = true
-warn_redundant_casts = true
-warn_return_any = true
 warn_unreachable = true
-warn_unused_configs = true
-warn_unused_ignores = true
```
provinzkraut, line 264: Why have these been removed?

provinzkraut:
> Looks like `else` is never actually reached. Should I remove this branch?

Seems to be a missing test case?

pyproject.toml
Alc-Alc: from PR description
> I also removed configs that are part of the --strict flag from settings.

provinzkraut:
>     * Extra `else` branch in this `if` ?

Isn't it the `elif` branch that's extra then, and should be an `else` instead?

provinzkraut (approved):

PR #3825, opened 2024-10-17
docs: rewrite htmx guide

sobolevn:
- I added a short intro: what is HTMX? Because not people are familiar with it (I was not for example)
- I also explained what `if request.htmx` means
- Always spell `HTMX` as `HTMX` and not `htmx` or anything else
- Always use `~` instead of duplicating class names
- Use special syntax for notes: `.. note::`

### DIFF
--- a/docs/usage/htmx.rst
+++ b/docs/usage/htmx.rst
@@ -1,7 +1,12 @@
 HTMX
 ====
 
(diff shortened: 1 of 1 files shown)


sobolevn:
@euri10 thank you for the quick review! 👍

PR #3837, opened 2024-10-20
feat: implement `HTMX` plugin using `litestar-htmx`

cofin:
## Description

This plugin migrates the HTMX integration to `litestar.plugins.htmx`.  

This logic has been moved to it's own repository named `litestar-htmx`

This is in preparation for 3.0 release where `contrib` will be deprecated.

### DIFF
--- a/litestar/contrib/htmx/_utils.py
+++ b/litestar/contrib/htmx/_utils.py
@@ -1,12 +1,22 @@
 from __future__ import annotations
 
-from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, cast
-from urllib.parse import quote
+from typing import TYPE_CHECKING
 
-from litestar.exceptions import ImproperlyConfiguredException
-from litestar.serialization import encode_json
+from litestar.utils import warn_deprecation
 
+if TYPE_CHECKING:
+    from litestar_htmx._utils import (  # noqa: TCH004
+        HTMXHeaders,
+        get_headers,
+        get_location_headers,
+        get_push_url_header,
+        get_redirect_header,
+        get_refresh_header,
+        get_replace_url_header,
+        get_reswap_header,
+        get_retarget_header,
+        get_trigger_event_headers,
+    )
 __all__ = (
     "HTMXHeaders",
     "get_headers",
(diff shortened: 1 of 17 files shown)


litestar/contrib/htmx/_utils.py
```diff
@@ -1,12 +1,22 @@
...
-from typing import TYPE_CHECKING, Any, Callable, cast
-from urllib.parse import quote
+from typing import TYPE_CHECKING
 
-from litestar.exceptions import ImproperlyConfiguredException
-from litestar.serialization import encode_json
+from litestar.utils import warn_deprecation
 
+if TYPE_CHECKING:
+    from litestar_htmx._utils import (  # noqa: TCH004
```
Alc-Alc, line 8: While this is something we control, is it good practice to import a third party "private" module inside Litestar's private module?

pyproject.toml
```diff
@@ -44,6 +44,8 @@ dependencies = [
     "click",
     "rich>=13.0.0",
     "rich-click",
+    # default litestar plugins
```
Alc-Alc, line 47: Is it possible in PDM to move this to a different group called "litestar-plugins" (not bikeshedded),that is still mandatory but just for cosmetic reasons so its easier to read?

litestar/contrib/htmx/_utils.py
cofin: No, but I figured these methods may be used enough for us to have to have it re-exported.

tests/unit/test_template/test_template.py
provinzkraut, line 1: These changes seem unrelated?

tests/unit/test_testing/test_test_client.py
provinzkraut, line 1: More unrelated changes?

provinzkraut (approved):

provinzkraut (changes requested):
Right now this is a breaking change though, due to the fact that this line here sets the request class of the application to `HTMXRequest`.

https://github.com/litestar-org/litestar-htmx/blob/4f5fc871ee5011bcb4b3b422c476db09770bf574/litestar_htmx/plugin.py#L42

tests/unit/test_testing/test_test_client.py
cofin: We have some issues after upgrading all of the latest dependencies.  

There's an issue in the latest version of AnyIO that makes the trio implementation fail on this test.

cofin:
> Right now this is a breaking change though, due to the fact that this line here sets the request class of the application to `HTMXRequest`.
> 
> https://github.com/litestar-org/litestar-htmx/blob/4f5fc871ee5011bcb4b3b422c476db09770bf574/litestar_htmx/plugin.py#L42

There is no `HTMXPlugin` currently.  This is actually new functionality that I've implemented with the move from `contrib`.

tests/unit/test_template/test_template.py
cofin: This is the same as the issue below.  The tests no longer pass after upgrading the project dependencies.

provinzkraut (approved):

PR #3843, opened 2024-10-29
feat(Pydantic): honor hide_input_in_errors in throwing validation exceptions

charles-dyfis-net:
## Description

Pydantic's `BaseModel` supports configuration to hide data values when throwing exceptions, via setting `hide_input_in_errors` -- see https://docs.pydantic.dev/2.0/api/config/#pydantic.config.ConfigDict.hide_input_in_errors and https://docs.pydantic.dev/latest/usage/model_config/#hide-input-in-errors (the value is also in 2.0, but not linkable in the non-`latest` version of the documentation build).

At present, Litestar does not honor this configuration when generating exceptions for validation errors, which can lead to sensitive data leaking into logs.


cofin (approved):
LGTM

cofin:
@all-contributors add @charles-dyfis-net for code

PR #3849, opened 2024-11-07
docs: fix the description of when to use the synchronous function

vvanglro:
## Description

-

### DIFF
--- a/docs/topics/sync-vs-async.rst
+++ b/docs/topics/sync-vs-async.rst
@@ -101,7 +101,7 @@ When to use a synchronous function
 ----------------------------------
 
 As an inverse of the previous paragraph, it follows that synchronous functions should
-be used for non-io, intensive tasks. The synchronous execution
+be used for non-io intensive tasks. The synchronous execution
 model allows for the smallest amount of overhead and should therefore be preferred in
 such situations where no asynchronous functionality is made use of.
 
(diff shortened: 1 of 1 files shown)


docs/topics/sync-vs-async.rst
```diff
@@ -101,7 +101,7 @@ When to use a synchronous function
 ----------------------------------
 
 As an inverse of the previous paragraph, it follows that synchronous functions should
-be used for non-blocking, non-computationally intensive tasks. The synchronous execution
+be used for non-io, intensive tasks. The synchronous execution
```
sobolevn, line 104:
```suggestion
be used for non-io intensive tasks. The synchronous execution
```

sobolevn (approved):
I think that this change makes the whole phrase easier. Thank you!

cofin (approved):

docs/topics/sync-vs-async.rst
```diff
@@ -101,7 +101,7 @@ When to use a synchronous function
 ----------------------------------
 
 As an inverse of the previous paragraph, it follows that synchronous functions should
-be used for non-blocking, non-computationally intensive tasks. The synchronous execution
+be used for non-io intensive tasks. The synchronous execution
```
provinzkraut, line 104: This makes it sound like you should use synchronous functions for lightweight IO tasks, which is the opposite of what the paragraph is trying to say

PR #3850, opened 2024-11-09
feat: add support for Python 3.13

cofin:
## Description

This PR implements support for Python 3.13.

**Note** 
- Python 3.8 continues to be supported with the existing `msgspec` implementation, while all other versions prefer the pre-built litestar wheels.
- There are no Python 3.13 prebuilt wheels for `psycopg[binary]`.  If you rely on this for development, you'll need to have the postgres development libraries installed
- `picologging` does not currently support Python 3.13.  These tests are skipped automatically.
- uses an updated `pytest-asyncio`.  Once the branch has been merged upstream, we can remove this test dependency.
- Running `make install` will still install using Python 3.12.  This is done to ensure we don't get additional bug reports due to build dependencies missing.  If you want to use python 3.13, run `uv sync --all-extras --dev --python 3.13`


pyproject.toml
```diff
@@ -96,7 +101,7 @@ opentelemetry = ["opentelemetry-instrumentation-asgi"]
 piccolo = ["piccolo"]
 picologging = ["picologging"]
 prometheus = ["prometheus-client"]
-pydantic = ["pydantic", "email-validator", "pydantic-extra-types"]
+pydantic = ["pydantic", "email-validator", "pydantic-extra-types<=2.9.0; python_version < \"3.9\"", "pydantic-extra-types>=2.9.0; python_version >= \"3.9\""]
```
cofin, line 104: There's an open build issue re: [this](https://github.com/pydantic/pydantic-extra-types/issues/248).

pyproject.toml
```diff
@@ -130,13 +135,14 @@ dev = [
   "trio",
   "aiosqlite",
   "asyncpg>=0.29.0",
-  "psycopg[pool,binary]>=3.1.10,<3.2",
+  "psycopg[pool,binary]>=3.1.10,<3.2; python_version < \"3.13\"",
+  "psycopg[pool,c]; python_version >= \"3.13\"",
```
cofin, line 139: Package needs to be built for 3.13

tests/examples/test_contrib/test_sqlalchemy/test_sqlalchemy_examples.py
```diff
@@ -1,14 +1,32 @@
...
 
+from litestar.plugins.sqlalchemy import AsyncSessionConfig, SQLAlchemyAsyncConfig
 from litestar.testing import TestClient
 
 pytestmark = pytest.mark.xdist_group("sqlalchemy_examples")
 
 
-def test_sqlalchemy_declarative_models() -> None:
-    from docs.examples.contrib.sqlalchemy.sqlalchemy_declarative_models import app
+async def test_sqlalchemy_declarative_models(tmp_path: Path, monkeypatch: MonkeyPatch) -> None:
```
cofin, line 14: Without this additional monkey-patching, there were intermittent failures on this job.  This is the only way I could make the tests consistently pass on all python versions.

tests/helpers.py
```diff
@@ -90,9 +92,9 @@ def cleanup_logging_impl() -> Generator:
         # Don't interfere with PyTest handler config
         if not isinstance(std_handler, (_LiveLoggingNullHandler, LogCaptureHandler)):
             std_root_logger.removeHandler(std_handler)
-
+    picologging = pytest.importorskip("picologging")
```
cofin, line 95: TIL about this.  It will auto-skip if this import fails

tests/unit/test_kwargs/test_validations.py
```diff
@@ -46,19 +46,19 @@ def handler(my_key: str = Parameter(**{param_field: "my_key"})) -> None:  # type
 @pytest.mark.parametrize("reserved_kwarg", sorted(RESERVED_KWARGS))
 def test_raises_when_reserved_kwargs_are_misused(reserved_kwarg: str) -> None:
     decorator = post if reserved_kwarg != "socket" else websocket
-
-    exec(f"async def test_fn({reserved_kwarg}: int) -> None: pass")
-    handler_with_path_param = decorator("/{" + reserved_kwarg + ":int}")(locals()["test_fn"])
+    local = dict(locals(), **globals())
```
cofin, line 49: This also caused some weird issues with the scope of the `locals()` output.  This was the only way I could consistently make sure the call to `locals` contained the `test_fn` key

pyproject.toml
```diff
@@ -31,22 +32,25 @@ classifiers = [
...
-    "rich>=13.0.0",
-    "rich-click",
-    "multipart>=1.2.0",
-    # default litestar plugins
-    "litestar-htmx>=0.4.0"
+  "anyio>=3 ; python_version < \"3.9\"",
+  "anyio>=4 ; python_version >= \"3.9\"",
+  "httpx>=0.22",
+  "exceptiongroup; python_version < \"3.11\"",
+  "eval_type_backport; python_version <= \"3.9\"",
```
provinzkraut, line 39: Curious: Why do we need this?

cofin: I don't think we do.  Removing.

cofin:
Breaking off a few smaller PRs that are unrelated but caused errors based on linting and environment updates.

#3918  #3917  #3916

tests/unit/test_middleware/test_middleware_handling.py
```diff
@@ -102,13 +103,22 @@ class JSONRequest:
...
+            assert "test logging" in caplog.text
+    else:
         client = create_test_client(
             route_handlers=[post_handler], middleware=[MiddlewareProtocolRequestLoggingMiddleware]
         )
         response = client.post("/", json={"name": "moishe zuchmir", "age": 40, "programmer": True})
         assert response.status_code == 201
-        assert "test logging" in caplog.text
+        log = capsys.readouterr()
+        assert "test logging" in log.err
```
cofin, line 121: I've tried everything to get these to behave more consistently.  For now, this is working, but I think it should be revisited.  However, let's not hold up because of this.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: "3"
+  python: "3.12"
```
cofin, line 2: We want to leave the version at 3.12 to avoid having to have the odd dependency setup for 3.13.  Once psycopg ships with a binary, we can change this to 3.13

pyproject.toml
```diff
@@ -94,9 +98,14 @@ mako = ["mako>=1.2.4"]
...
 opentelemetry = ["opentelemetry-instrumentation-asgi"]
 piccolo = ["piccolo"]
-picologging = ["picologging"]
+picologging = ["picologging; python_version < \"3.13\""]
 prometheus = ["prometheus-client"]
-pydantic = ["pydantic", "email-validator", "pydantic-extra-types"]
+pydantic = [
+  "pydantic",
+  "email-validator",
+  "pydantic-extra-types!=2.9.0; python_version < \"3.9\"",
```
cofin, line 106: It looks like only the 2.9.0 version is impacted.  Since it wasn't pulled from pypi, i'm just ignoring this single release.

cofin:
For reference on the `pytest-asyncio` issues, please see [this](https://github.com/pytest-dev/pytest-asyncio/pull/1034) PR

pyproject.toml
```diff
@@ -31,12 +32,14 @@ classifiers = [
   "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
-  "anyio>=3",
+  "anyio>=3 ; python_version < \"3.9\"",
+  "anyio>=4 ; python_version >= \"3.9\"",
```
provinzkraut, line 36: I think this needs to be 

```suggestion
  "anyio>=3 ; python_version < \"3.13\"",
  "anyio>=4 ; python_version >= \"3.13\"",
```

cofin: It's been reverted to just `anyio>=3`

pyproject.toml
```diff
@@ -167,7 +175,8 @@ linting = [
 test = [
   "covdefaults",
   "pytest",
-  "pytest-asyncio",
+  "pytest-asyncio; python_version < \"3.13\"",
+  "pytest-asyncio @ git+https://github.com/provinzkraut/pytest-asyncio.git@shutdown-asyncgens; python_version >= \"3.13\"",
```
cofin, line 179:
```suggestion
  "pytest-asyncio",
```
Revert this once the PR has been merged.

pyproject.toml
```diff
@@ -175,8 +175,7 @@ linting = [
 test = [
   "covdefaults",
   "pytest",
-  "pytest-asyncio; python_version < \"3.13\"",
-  "pytest-asyncio @ git+https://github.com/provinzkraut/pytest-asyncio.git@shutdown-asyncgens; python_version >= \"3.13\"",
+  "pytest-asyncio",
```
provinzkraut, line 178: Should probably add a lower bound for the latest release here?

cofin: It looks like they removed support for python 3.8, so I don't think we can pin it to the latest version that was just released for all pythons.  Here's the lock file entry:
```py
    { name = "pytest-asyncio", version = "0.24.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.9'" },
    { name = "pytest-asyncio", version = "0.25.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.9'" },
    ```

provinzkraut: Could we pin for >=3.13 only at least? I just want to avoid some accidental downgrades in the future

provinzkraut (approved):

pyproject.toml
cofin: pinned the newer version for 3.9+ and pinned the last supported version for 3.8

PR #3852, opened 2024-11-09
feat: deprecate `litestar.contrib.pydantic`

cofin:
## Description

This PR continues preparation for 3.0 by deprecating `litestar.contrib.pydantic` in favor of `litestar.plugins.pydantic`


uv.lock
provinzkraut, line 1: This was committed by accident I think?

pdm.lock
provinzkraut, line 1: These updates should probably happen in a separate PR

provinzkraut (approved):

PR #3863, opened 2024-11-17
feat: move `litestar.contrib.prometheus` to `litestar.plugins.prometheus`

cofin:
## Description

feat: deprecate `litestar.contrib.prometheus` in favor of the new location of `litestar.plugins.prometheus`

### DIFF
--- a/tests/unit/test_contrib/test_prometheus.py
+++ b/tests/unit/test_contrib/test_prometheus.py
@@ -1,217 +1,63 @@
-import re
-import time
-from http.client import HTTPException
+# ruff: noqa: TCH004, F401
+from __future__ import annotations
+
+import importlib
+import sys
+from importlib.util import cache_from_source
 from pathlib import Path
-from typing import Any
 
 import pytest
-from _pytest.monkeypatch import MonkeyPatch
-from prometheus_client import REGISTRY
-from pytest_mock import MockerFixture
-
-from litestar import get, post, websocket_listener
-from litestar.contrib.prometheus import PrometheusConfig, PrometheusController, PrometheusMiddleware
-from litestar.status_codes import HTTP_200_OK
-from litestar.testing import create_test_client
 
 
-def create_config(**kwargs: Any) -> PrometheusConfig:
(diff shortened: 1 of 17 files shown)


tests/unit/test_contrib/test_prometheus.py
```diff
@@ -1,217 +1,63 @@
...
-    monkeypatch.setenv(env_var, str(proc_dir))
-    config = create_config()
-    mock_registry = mocker.patch("litestar.contrib.prometheus.controller.CollectorRegistry")
-    mock_collector = mocker.patch("litestar.contrib.prometheus.controller.multiprocess.MultiProcessCollector")
 
-    with create_test_client([PrometheusController], middleware=[config.middleware]) as client:
-        client.get("/metrics")
 
-    mock_collector.assert_called_once_with(mock_registry.return_value)
+def purge_module(module_names: list[str], path: str | Path) -> None:
```
Alc-Alc, line 12: can we move this to `conftest.py` as a fixture? I noticed we seem to be using this for all of these "contrib to plugin" PRs.

cofin: Maybe we can do this in a followup PR, and move all of the references to it?

JacobCoffee (commented):
sorry maybe a dumb q - what is the benefit here of moving from `contrib` to `plugins` aside from renaming it?

also i feel like this one is a good candidate to move outside of the stdlib into its own package and then we can assign it as an extra

cofin:
> sorry maybe a dumb q - what is the benefit here of moving from `contrib` to `plugins` aside from renaming it?
> 
> also i feel like this one is a good candidate to move outside of the stdlib into its own package and then we can assign it as an extra

The goal is to get any non-essential core code into a plugin style structure.  Right now, we have both `litestar.plugins` and `litestar.contrib` and it's contributing to code sprawl and additional maintenance overhead.

Re: the move to an external library I also agree.  In fact, I think most (maybe all) of the plugins can ultimately be treated that way.  We can do that in steps though.

provinzkraut (approved):

PR #3869, opened 2024-11-21
feat: add cache store parameter on route level basis

### LINKED ISSUE #3854: Enhancement: Allow the cache to specify the namespace on a route level
### Summary

Today, every cached response is cached inside the namespace "response_cache" or any namespace you defined in a root level. However, I would like certain route to be cached on different namespace to manage the cache invalidation differently according to the response.

### Basic Example

```
@get(cache=True, cache_namespace='another_response_cache', sync_to_thread=False)
def cached_handler() -> str:
    # this will use app.stores.get("response_cache")
    return "Hello, world!"
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

dylandoamaral:
## Description

Allow the user to define on a route level basis what would be the used namespace by overriding the default one. It allows, as an example, to delete some routes easily solely based on the namespace.

This is my first commit in this repository, I don't know if I did it right. Sorry for the inconvenience ! By the way, after reading the documentation for contributing, I don't know when and if I have to run the all-contributors myself.

I will also have to update the documentation, but first I prefer someone to tell me if this feature should be merged or ignored :).

## Closes

Closes #3854


provinzkraut (changes requested):
Hey @dylandoamaral, thanks for the PR!

The change is okay in general, but due to the change in the `get_store_from_app` signature, it would be considered breaking. You could either try to find a way to make this not breaking, or target version 3 by pointing your PR at the `v3.0` branch

dylandoamaral:
Thanks for your response. It is still breaking even with an optional defaulted value ?

provinzkraut:
> Thanks for your response. It is still breaking even with an optional defaulted value ?

Yes, it would.

For example, if someone has overwritten this method in a subclass, it would break for them, because their subclass does not know of this parameter.

dylandoamaral:
Supersed by https://github.com/litestar-org/litestar/pull/3871

PR #3870, opened 2024-11-21
fix: remove optional group for `litestar-htmx`

cofin:
## Description

This PR removes the optional `htmx` group for the project dependencies.  HTMX is a built in feature and is not optional.


provinzkraut (approved):

charles-dyfis-net:
Insofar as the impetus for raising this duplication was a bug encountered in poetry2nix (infinite recursion at evaluation time when Litestar 2.13.0 is in use) -- with this PR merged the bug still is reachable but has a much easier workaround: adding `litestar-htmx` as a direct dependency of the project also using litestar suffices to address it.

PR #3872, opened 2024-11-23
feat(core): Streaming multipart parser

provinzkraut:
Add a streaming multipart parser via the [multipart](https://github.com/defnull/multipart) lib.

This gives us:

- Ability to stream large / larger-than-memory file uploads
- Better / more correct edge case handling
- Still good performance

### DIFF
--- a/litestar/_multipart.py
+++ b/litestar/_multipart.py
@@ -1,41 +1,22 @@
-"""The contents of this file were adapted from sanic.
-
-MIT License
-
-Copyright (c) 2016-present Sanic Community
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
-"""
-
 from __future__ import annotations
 
 import re
 from collections import defaultdict
-from email.utils import decode_rfc2231
-from typing import TYPE_CHECKING, Any
-from urllib.parse import unquote
+from typing import TYPE_CHECKING, Any, AsyncGenerator
+
+from multipart import (  # type: ignore[import-untyped]
+    MultipartSegment,
+    ParserError,
+    ParserLimitReached,
+    PushMultipartParser,
+)
 
 from litestar.datastructures.upload_file import UploadFile
-from litestar.exceptions import ValidationException
+from litestar.exceptions import ClientException
 
-__all__ = ("parse_body", "parse_content_header", "parse_multipart_form")
+__all__ = ("parse_content_header", "parse_multipart_form")
 
+from litestar.utils.compat import async_next
 
 if TYPE_CHECKING:
     from litestar.types import TypeDecodersSequence
@@ -67,34 +48,8 @@ def parse_content_header(value: str) -> tuple[str, dict[str, str]]:
     return value.strip().lower(), options
 
 
-def parse_body(body: bytes, boundary: bytes, multipart_form_part_limit: int) -> list[bytes]:
-    """Split the body using the boundary
-        and validate the number of form parts is within the allowed limit.
-
-    Args:
-        body: The form body.
-        boundary: The boundary used to separate form components.
-        multipart_form_part_limit: The limit of allowed form components
-
-    Returns:
-        A list of form components.
-    """
-    if not (body and boundary):
-        return []
-
-    form_parts = body.split(boundary, multipart_form_part_limit + 3)[1:-1]
-
-    if len(form_parts) > multipart_form_part_limit:
-        raise ValidationException(
-            f"number of multipart components exceeds the allowed limit of {multipart_form_part_limit}, "
-            f"this potentially indicates a DoS attack"
-        )
-
-    return form_parts
-
-
-def parse_multipart_form(
-    body: bytes,
+async def parse_multipart_form(  # noqa: C901
+    stream: AsyncGenerator[bytes, None],
     boundary: bytes,
     multipart_form_part_limit: int = 1000,
     type_decoders: TypeDecodersSequence | None = None,
@@ -102,7 +57,7 @@ def parse_multipart_form(
     """Parse multipart form data.
 
     Args:
-        body: Body of the request.
+        stream: Body of the request.
         boundary: Boundary of the multipart message.
         multipart_form_part_limit: Limit of the number of parts allowed.
         type_decoders: A sequence of type decoders to use.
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -44,8 +44,9 @@ dependencies = [
     "click",
     "rich>=13.0.0",
     "rich-click",
+    "multipart@git+https://github.com/defnull/multipart.git",
```
provinzkraut, line 47: Pinned to `master` for now until a new release is out. See https://github.com/defnull/multipart/issues/68

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -292,6 +296,7 @@ def test_multipart_request_without_charset_for_filename() -> None:
         }
 
 
+@pytest.mark.xfail
```
provinzkraut, line 299: Figure out why this is failing. I suspect `filename*` might not be (fully) supported

defnull: Yes, `filename*` is not supported because it is deprecated (actually forbidden) according to RFC-7578 section 4.2)

> NOTE: The encoding method described in [RFC5987], which would add a "filename*" parameter to the Content-Disposition header field, MUST NOT be used.

litestar/_multipart.py
```diff
@@ -113,51 +68,55 @@ def parse_multipart_form(
...
+                                headers=dict(segment.headerlist),
+                            )
+                    elif form_part:
+                        if isinstance(data, UploadFile):
+                            await data.write(form_part)
+                        else:
+                            data += form_part
+                    else:
+                        # end of part
+                        if segment is None:
```
defnull, line 96: This cannot happen, `segment` will always have a value at this point. If not, then that would be a bug in PushMultipartParser. The contract is as follows: For each part of a valid multipart stream, the parser will emit exactly one `MultipartSegment` instance, followed by zero or more non-empty content chunks, followed by `None`. A segment cannot end before its header section is complete, that would be a `ParserError` exception.

litestar/_multipart.py
```diff
@@ -113,51 +68,55 @@ def parse_multipart_form(
...
+                            data = UploadFile(
+                                content_type=segment.content_type or "text/plain",
+                                filename=segment.filename,
+                                headers=dict(segment.headerlist),
+                            )
+                    elif form_part:
+                        if isinstance(data, UploadFile):
+                            await data.write(form_part)
+                        else:
+                            data += form_part
```
defnull, line 93: Appending to a `bytearray` is significantly faster than appending to a `bytes` instance. Change `data` to a bytearray?

defnull: Don't forget a size check here or an attacker could upload arbitrary large text fields and fill up memory. Maybe also limit the total number of parts, because both text fields (byte strings) and `UploadFile` instances that are smaller than the disk spool limit, can fill up memory if the attacker uploads thousands of fields.

litestar/_multipart.py
```diff
@@ -113,51 +68,55 @@ def parse_multipart_form(
...
+                            # received a complete header segment
+                            raise ClientException("Unexpected eof in multipart/form-data")
+                        if segment.name:
+                            if isinstance(data, UploadFile):
+                                await data.seek(0)
+                                fields[segment.name].append(data)
+                            elif data:
+                                fields[segment.name].append(data.decode(segment.charset or "utf-8"))
+                            else:
+                                fields[segment.name].append(None)
```
defnull, line 107: Shouldn't this be an empty string instead? A text form field without content is empty, but not missing or invalid.

defnull (changes requested):
I couldn't wait till tomorrow with the review ;)

litestar/_multipart.py
provinzkraut: I know, this check mostly exists for the benefit of the type checker. One could argue that it should be an `assert` then, but since it's just an `is None` check I felt that there's not that much benefit to it (and we usually don't use the `assert` pattern elsewhere)

litestar/_multipart.py
```diff
@@ -113,51 +68,55 @@ def parse_multipart_form(
...
+                            await data.write(form_part)
+                        else:
+                            data += form_part
+                    else:
+                        # end of part
+                        if segment is None:
+                            # we have reached the end of a segment before we have
+                            # received a complete header segment
+                            raise ClientException("Unexpected eof in multipart/form-data")
+                        if segment.name:
```
defnull, line 100: Segments always have a name

tests/unit/test_kwargs/test_multipart_data.py
provinzkraut: Interesting, I was not aware of this. Hmm. We have previously supported this, so removing support for it now would constitute a breaking change. Do you think there's a reasonable workaround for us here while still using `multipart`?

litestar/_multipart.py
provinzkraut: > Don't forget a size check here or an attacker could upload arbitrary large text fields and fill up memory.

Well, we have introduced the feature to limit request body size recently, I think that should cover this case as well.

> Maybe also limit the total number of parts

Isn't that what the `max_segment_count` parameter does? I'm setting this here: 
https://github.com/litestar-org/litestar/blob/3d967c9d06108a5a1c7a9b99d121b0d3a93247a2/litestar/_multipart.py#L76

tests/unit/test_kwargs/test_multipart_data.py
defnull: I'm open for adding optional features if there is an actual real-world use case for it, but `filename*` is a `MUST NOT` and I do not know any browser or HTTP client that actually does this. I'd argue that following the RFC and WHATWG HTML5 spec is a bugfix, not a breaking change ;) Breaking broken clients is okay.

litestar/_multipart.py
provinzkraut: We use `None` here for the benefit of the validation libraries consuming this data, which do make an effort to distinguish between `None`, `""` and not being provided a value. In this case, *we* provide the field with a value, but want to signal that no value was provided to us.

litestar/_multipart.py
defnull: Yes, sorry, I'm tired :) The segment count is indeed limited by `PushMultipartParser`, but memory consumption is not because it does not deal with buffering field content. An app may want to benefit from the new streaming parser and increase the request body size to, say, 4GBs to accept ISO image uploads. That's fine if the upload is buffered to disk by `UploadFile` but a real issue if it's not and ends up in a byte(array) because it did not provide a `filename` option. The non-async `MultipartParser` has two additional limits for this reason, one for on-disk buffers and one for in-memory buffers.

defnull: Fun fact: Werkzeug and Starlette had the exact same issue with unbounded text field in-memory buffers and it was fixed just recently in both frameworks. 
https://nvd.nist.gov/vuln/detail/CVE-2024-49767 and https://nvd.nist.gov/vuln/detail/CVE-2024-47874

tests/unit/test_kwargs/test_multipart_data.py
euri10: > Breaking broken clients should be allowed.

I agree on this.

litestar/_multipart.py
provinzkraut: > but memory consumption is not because the parser does not deal with buffering field content. An app may want to benefit from the new streaming parser and increase the request body size limit to, say, 4GBs to accept ISO image uploads. That's fine if the upload is buffered to disk by `UploadFile` but a real issue if it's not and ends up in a byte(array) because the segment did not have a `filename`

So, we could introduce another limit here for the max size of an individual form part, but I'd prefer not to do that, since it's another potentially breaking change. I do see the issue with what you're talking about, but I'm inclined to say that within the "proper usage" category. Maybe instead of arbitrarily adding a limit here, we could, in a follow-up PR, introduce a new parameter that allows to add size limits to form parts, and a section in the documentation / docstrings about the potential issues from not setting them?

@litestar-org/maintainers

pyproject.toml
```diff
@@ -95,7 +96,7 @@ opentelemetry = ["opentelemetry-instrumentation-asgi"]
 piccolo = ["piccolo"]
 picologging = ["picologging"]
 prometheus = ["prometheus-client"]
-pydantic = ["pydantic", "email-validator", "pydantic-extra-types"]
+pydantic = ["pydantic<2.10", "email-validator", "pydantic-extra-types"]
```
euri10, line 99: you can relax that since you fixed it elsewhere

provinzkraut: Working on it at least 😅

pyproject.toml
euri10: good switch, I didn't know that library and it looks very well thought, I'm always impressed by people talking bytes.

euri10 (approved):
amazing addition

defnull:
Multipart v1.2.0 is released now :)

PR #3879, opened 2024-11-28
fix(devcontainers): Replace PDM with UV

Alc-Alc:
~~Obviously next is Poetry 2 (@Kumzy )~~

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,10 @@
 # [Choice] Python version (use -bookworm or -bullseye variants on local arm64/Apple Silicon): 3, 3.13, 3.12, 3.11, 3.10, 3.9, 3.8, 3-bookworm, 3.13-bookworm, 3.12-bookworm, 3.11-bookworm, 3.10-bookworm, 3.9-bookworm, 3.8-bookworm, 3-bullseye, 3.11-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3-buster, 3.11-buster, 3.10-buster, 3.9-buster, 3.8-buster
 ARG VARIANT=3.12-bookworm
-FROM python:${VARIANT}
+FROM ghcr.io/astral-sh/uv:python${VARIANT}
+ENV UV_LOCKED=1
+ENV UV_PYTHON=/usr/local/bin/python
 
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get purge -y fish
 
-RUN python3 -m pip install --upgrade setuptools cython pip pdm
+RUN python3 -m pip install --upgrade setuptools cython pip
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,8 +1,10 @@
 # [Choice] Python version (use -bookworm or -bullseye variants on local arm64/Apple Silicon): 3, 3.13, 3.12, 3.11, 3.10, 3.9, 3.8, 3-bookworm, 3.13-bookworm, 3.12-bookworm, 3.11-bookworm, 3.10-bookworm, 3.9-bookworm, 3.8-bookworm, 3-bullseye, 3.11-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3-buster, 3.11-buster, 3.10-buster, 3.9-buster, 3.8-buster
 ARG VARIANT=3.12-bookworm
-FROM python:${VARIANT}
+FROM ghcr.io/astral-sh/uv:python${VARIANT}
```
sobolevn, line 3: I advocate for using the raw `python` image and download `uv` as a binary exeuctable, it is easier and we would still use the most basic image.

Alc-Alc: I will make this change in the upcoming commit 🙂

Alc-Alc: @sobolevn would you prefer we installed from [PyPi](https://pypi.org/project/uv/) (like we did for pdm) or would you still advocate for installing it as a binary (from the installation script)?

If we are installing as a binary I am unsure if all the base images will have curl (or wget) to download the installation script.

sobolevn: `pip` installation can cause problems in theory. I had several when doing so with `poetry`

Alc-Alc: a746281fbee48b1d57907143446b296439f15a60

.devcontainer/Dockerfile
```diff
@@ -1,10 +1,18 @@
...
 
+FROM python:${VERSION}${VARIANT}
+
+ENV UV_LOCKED=1 UV_PYTHON=${VERSION} PATH="/root/.local/bin/:$PATH"
+
+# https://docs.astral.sh/uv/guides/integration/docker/#installing-uv
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get purge -y fish
+    && apt-get purge -y fish \
+    && apt-get install -y --no-install-recommends curl ca-certificates
```
sobolevn, line 12: Don't forget to clean everything :)

```
  && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
  && apt-get clean -y && rm -rf /var/lib/apt/lists/*
```

.devcontainer/Dockerfile
```diff
@@ -1,10 +1,18 @@
...
+
+ADD https://astral.sh/uv/install.sh /uv-installer.sh
+
+RUN sh /uv-installer.sh && rm /uv-installer.sh
```
sobolevn, line 16: Maybe better to use `curl` and `rm` in a single run? Now `/uv-installer.sh` will be in `ADD` layer.

Alc-Alc: @sobolevn what about copying as shown [here](https://docs.astral.sh/uv/guides/integration/docker/#installing-uv)? It should avoid the need to download curl and also cleanup?

```
FROM python:3.12-slim-bookworm
COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```

sobolevn: Great idea!

JacobCoffee (approved):

sobolevn (approved):

PR #3885, opened 2024-11-30
fix: typo in exception message for ImproperlyConfiguredException

Zimzozaur:
Corrected the exception message from "Provider dependency must a callable value" to "Provider dependency must be a callable value". This ensures clarity and proper grammar in error messages.


sobolevn (approved):
Thank you!

Issue #3889, opened 2024-12-05
Bug: OpenAPI schema generation for handler <...> detected multiple parameters named <...> with different types


Molozey:
### Description

Hello everyone! I encountered a behavior, and I'm not sure if it's correct.

The reproduction steps are attached in the form of a repository (the following description will refer to it).

As a demonstration, I wanted to create BasicViewProvider, which is a simple key-value pair and used as dependency at Controller
```
class BasicViewProvider(ABC):
    """
    Abstract View Provider for demo key-value
    """

    @abstractmethod
    def put(self, key: str, value: str) -> None: ...

    @abstractmethod
    def get(self, key: str) -> str: ...

    @abstractmethod
    def delete(self, key: str) -> bool: ...
 ```
```
class ExampleController(Controller):
    """
    Example Controller
    """

    path = "/example"
    tags = ["Example"]

    # If we remove dependencies examples: all is OK!
    dependencies = {"view_repo": Provide(view_provider_factory)}

    @post(path="/put")  # If we remove dependencies examples: all is OK!
    async def put(
        self,
        idx: str,
        obj_id: str,
        view_repo: BasicViewProvider
    ) -> None:
        view_repo.put(idx, obj_id)
      ...
```

At the same time, my path group requires additional parameters (in the real task, to provide object access models via middleware). So, I register the group as

```
example_route = Router(
    path="/objects",
    route_handlers=[ExampleController],
    parameters={
        # For simple parameter we can provide examples to solve problem
        "team_id": Parameter(
            str,
            description="Affected team id",
            required=True,
            # examples=[
            #     Example(
            #         value="mocked-team-id-uuid",
            #         summary="summary",
            #         description="descr",
            #         external_value="Test-team-id",
            #     )
            # ],
        ),

        # For cookie parameter examples will not help
        "x-session-token": Parameter(
            str,
            description="Session JWT token",
            cookie="x-session-token",
            required=False,
            # examples=[]
        ),
    },
    # middleware=[auth_middleware()],     # Assume that we need parameters to auth and permissions
)
```

The handler calls work correctly and do what they should, however, after I enter Swagger, I get an error:
```
500: OpenAPI schema generation for handler `app.controllers.example.ExampleController.hidden` detected multiple parameters named 'x-session-token' with different types.
```

It's important to note that when the application starts, I add example creation in the OpenAPI configuration.

If I turn off example generation, the error disappears.
```
def app():

    return Litestar(
        debug=True,
[...]

PR #3890, opened 2024-12-05
chore: eq for Schema

Molozey:
Closes: https://github.com/litestar-org/litestar/issues/3889

## Description

-


provinzkraut:
This does not really fix the underlying bug, it just hides the symptom. I haven't investigated yet, but it seems like what's happening is that we're not re-using the schemas properly when generating examples. @Alc-Alc?

PR #3892, opened 2024-12-07
feat: add Valkey as a native store

ftsartek:
## Description

- Adds Valkey as a native store alternative to Redis.
- Adds an optional Litestar install dependency: `litestar[valkey]` which installs `valkey` with `libvalkey` as an optimisation layer.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,6 +63,7 @@ intersphinx_mapping = {
     "trio": ("https://trio.readthedocs.io/en/stable/", None),
     "pydantic": ("https://docs.pydantic.dev/latest/", None),
     "typing_extensions": ("https://typing-extensions.readthedocs.io/en/stable/", None),
+    "valkey": ("https://valkey-py.readthedocs.io/en/latest/", None),
 }
 
 napoleon_google_docstring = True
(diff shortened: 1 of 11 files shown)


ftsartek:
While this is very much unfinished (where are the tests??), I'm creating this now as somewhere to discuss this.

Valkey is _currently_ effectively API-compatible with Redis, so the only functional change here is the dependencies installed and the imports used, plus we call `valkey` instead of `redis` in the registered scripts (though even this is, I believe, compatible for now). As of Redis 8.x my understanding is that these will diverge a bit and lose their direct compatibility.

What I want to understand before I finish this off is largely how you as the maintainers would prefer this approached:
- Should the Redis and Valkey controllers be abstracted on top of an abstract KV store class? I considered this, but changing `self._redis` to `self._client` (or similar) has the potential of being a breaking change, or if it's overridden in the controllers themselves then all the methods would have to be overridden anyway.
- What's the preferred approach for tests with a strongly-shared basis of code like this? Do you prefer that the `test_stores` just runs twice, once with Redis and once with Valkey, or do we just get that coverage to 100% with some extra tests?

ftsartek:
Well, I got a bit bored and went through and basically replicated the existing Redis tests for Valkey. They're just new tests, though I'm open to a less-duplicative approach if anyone can point me in the right direction (I guess parametrise all the currently-redis-specific tests and then add checks for type to handle `client._redis` vs `client._valkey`, etc?).

The only thing that came up is that Valkey uses `server.call` instead of the `valkey.call` analogue you might expect for Redis' `redis.call`, but otherwise the scripts are identical.

I'm going to leave this as it is now pending any feedback.

cofin (approved):
LGTM.  I think this is fine for the initial implementation.  There is probably some larger restructuring we'll do on a few things for 3.0, and this may be part of it.  However,  i think it's better to get this implemented in smaller chunks.

cofin:
@all-contributors add @ftsartek for docs, tests and code

PR #3894, opened 2024-12-07
feat: WebSocket send stream

provinzkraut:
Add a new `websocket_stream` route handler that supports streaming data *to* a WebSocket via an async generator.

```python
@websocket_stream("/")
async def handler() -> AsyncGenerator[str, None]:
    yield str(time.time())
    await asyncio.sleep(.1)
```

This is roughly equivalent to (with some edge case handling omitted):
```python
@websocket("/")
async def handler(socket: WebSocket) -> None:
  await socket.accept()

  try:
    async with anyio.task_group() as tg:
      # 'receive' in the background to catch client disconnects
      tg.start_soon(socket.receive)
  
      while True:
        socket.send_text(str(time.time()))
        await asyncio.sleep(.1)
  finally:
    await socket.close()
```

Just like the WebSocket listeners, it also supports dependency injection and serialization:

```python
@dataclass
class Event:
    time: float
    data: str


async def provide_client_info(socket: WebSocket) -> str:
    return f"{socket.client.host}:{socket.client.port}"
  

@websocket_stream("/", dependencies={"client_info": provide_client_info})
async def handler(client_info: str) -> AsyncGenerator[Event, None]:
    yield Event(time=time.time(), data="hello, world!")
    await asyncio.sleep(.1)
```


cofin (approved):
LGTM

Issue #3891, opened 2024-12-07
Enhancement: Add query params to litestar.response.redirect.Redirect


marcuslimdw:
### Summary

I'd like to be able to specify query parameters in `Redirect.__init__`, rather than having to encode and append them to the redirect path manually.

### Basic Example

Before:
```
from urllib.parse import urlencode

from litestar.response import Redirect

params = {"key_1": "value_1", "key_2": ["value_2a", "value_2b"]}
query_string = urlencode(params)
response = Redirect(f"/my_path?{query_string}")
```

After:
```
from litestar.response import Redirect

response = Redirect("/my_path", params=params)
```

### Drawbacks and Impact

It would be easier to provide query params when redirecting to a URL.

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/3891">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3891/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/3891/pledge.svg">
</picture>
</a>

PR #3901, opened 2024-12-14
feat: add query params to redirect response

marcuslimdw:
## Description

Closes https://github.com/litestar-org/litestar/issues/3891.

### DIFF
--- a/litestar/response/redirect.py
+++ b/litestar/response/redirect.py
@@ -123,7 +123,7 @@ class Redirect(Response[Any]):
             self.url = path
         elif isinstance(query_params, MultiDict):
             # We can't use MultiDictMixin.dict() because it's not deterministic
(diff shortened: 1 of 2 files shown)


litestar/response/redirect.py
```diff
@@ -94,6 +97,7 @@ def __init__(
         media_type: str | MediaType | None = None,
         status_code: RedirectStatusType | None = None,
         type_encoders: TypeEncodersMap | None = None,
+        query_params: Mapping[str, str | Sequence[str]] | None = None,
```
provinzkraut, line 100: This should support `datastructures.MultiDict` instances, as that's what we're returning from `URL.query_params`

provinzkraut (commented):
Thanks @marcuslimdw!

provinzkraut (approved):

PR #3916, opened 2024-12-30
fix(channels): use `SQL` function for `psycopg`

cofin:
## Description

This PR updates the channels backend to use the native psycopg `SQL` API. 

Additionally, it preemptively fixes some new warnings from upcoming pyright and ruff changes


provinzkraut (approved):

cofin:
@provinzkraut Any ideas why these may be hanging now?  I wouldn't have expected this change to use the built-in query helpers to have impacted this, but maybe so.

provinzkraut:
> @provinzkraut Any ideas why these may be hanging now? I wouldn't have expected this change to use the built-in query helpers to have impacted this, but maybe so.

Unfortunately no. Have you tried debugging the response from psycopg? Maybe it errors out weirdly and that causes our tests to get stuck waiting for something from a listener?

PR #3920, opened 2025-01-03
feat: use "path" as an error message source

### LINKED ISSUE #3919: Enhancement: Use "path" as an error message source in litestar._signature.SignatureModel._build_error_message
### Summary

I'd like the validation error message for a path parameter to have "path" as the source instead of "query." This makes sense since the error is happening w.r.t. a path parameter and not a query parameter. It will also match the parameter location described in the generated OpenAPI schema.

### Basic Example

Using the "[Adding extra validation and documentation to a path parameter](https://docs.litestar.dev/2/usage/routing/parameters.html#id6)" example from the Parameters docs, the generated OpenAPI schema shows that `version` is a path parameter:
```
[...]
  "paths": {
    "/versions/{version}": {
      "get": {
        "summary": "GetProductVersion",
        "operationId": "VersionsGetProductVersion",
        "parameters": [
          {
            "name": "version",
            "in": "path",
[...]
```
But a validation error has "query" as its source:
```sh
$ curl localhost:8000/versions/11
{"status_code":400,"detail":"Validation failed for GET /versions/11","extra":[{"message":"Expected `int` <= 10","key":"version","source":"query"}]}
```
With this enhancement, "path" will be the source:
```sh
$ curl localhost:8000/versions/11
{"status_code":400,"detail":"Validation failed for GET /versions/11","extra":[{"message":"Expected `int` <= 10","key":"version","source":"path"}]}
```

### Drawbacks and Impact

The error message "source" property will be more accurate.

### Unresolved questions

_No response_

250MHz:
## Description

- Use "path" as the "source" property for an error message if the key is a path parameter

## Closes

Closes #3919

### DIFF
--- a/litestar/_signature/model.py
+++ b/litestar/_signature/model.py
@@ -156,6 +156,8 @@ class SignatureModel(Struct):
                 message["source"] = "body"
             elif key in connection.query_params:
                 message["source"] = ParamType.QUERY
+            elif key in connection.path_params:
+                message["source"] = ParamType.PATH
 
             elif key in cls._fields and isinstance(cls._fields[key].kwarg_definition, ParameterKwarg):
                 if cast(ParameterKwarg, cls._fields[key].kwarg_definition).cookie:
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):
Thanks @250MHz!

PR #3922, opened 2025-01-04
fix: ignore platform/version specific `trio` failures

cofin:
## Description

This PR temporarily ignores failures related to `trio` and specific `anyio` versions on Windows platforms.

[Here](https://github.com/litestar-org/litestar/actions/runs/12613952703/job/35152354906#step:7:313) is an example of the related error


cofin:
I think this is no longer needed after the `pytest` upgrade here #3925

PR #3928, opened 2025-01-06
Remove pytest asyncio > anyio + fixx some unclosed ressources

euri10:
wanted to give it a try again with https://github.com/litestar-org/litestar/commit/2fb6f6b118f1247421db027fafb14507b3f8f84d done

it fails on 4 tests left as is as they dont impact the rest, and on 3 that are impacting the rest so I commented them, that should be the focus

the 4 tests that have no impact on the rest so I left them failing are:
* tests.unit.test_concurrency.test_get_set_asyncio_executor 
* tests.unit.test_concurrency.test_get_set_trio_capacity_limiter

I'm not too sure about those 2 are failing to be fair

* tests.unit.test_cli.test_session_commands.test_delete_session
* tests.unit.test_cli.test_session_commands.test_clear_sessions

those 2 are failing with  RuntimeError('Already running asyncio in this thread') which seems normal to me and it looks like the test itself should probably needs to be rewritten, 

The 3 commented which impacts the rest are:
*  test_create_ws_route_handlers
*  test_ws_route_handlers_receive_arbitrary_message
*  test_create_ws_route_handlers_arbitrary_channels_allowed

I'm not too sure in that case that we shouldnt handle the Plugin shutdown manually or soething, just a guess.

I'll leave some notes on the changes if someone wanna pick that or try fix the rest, that should be a good start,


litestar/channels/backends/redis.py
```diff
@@ -102,7 +102,7 @@ async def on_startup(self) -> None:
         pass
 
     async def on_shutdown(self) -> None:
-        await self._pub_sub.reset()
+        await self._pub_sub.aclose()
```
euri10, line 105: reset is deprecated

tests/conftest.py
```diff
@@ -81,13 +84,13 @@ async def asgi_app(scope: Scope, receive: Receive, send: Send) -> None: ...
 
 
 @pytest.fixture()
-def redis_store(redis_client: AsyncRedis) -> RedisStore:
-    return RedisStore(redis=redis_client)
+def redis_store(redis_client: Redis) -> RedisStore:
+    return RedisStore(redis=redis_client, handle_client_shutdown=True)
```
euri10, line 88: we need to explicitely handle shutdown in test or we're left with dangling ressources

tests/conftest.py
```diff
@@ -327,31 +330,17 @@ def get_logger() -> GetLogger:
...
-    client: AsyncRedis = AsyncRedis(host=docker_ip, port=6397)
-    yield client
-    try:
-        await client.aclose()  # type: ignore[attr-defined]
-    except RuntimeError:
-        pass
+async def redis_client(redis_service: None, docker_ip: str):
+    async with Redis(host=docker_ip, port=6397) as client:
+        yield client
+        await client.flushall()
```
euri10, line 336: overall this feels more natural and the try / except of the original fixture is just a way to mask unclosed ressources

tests/unit/test_channels/test_plugin.py
```diff
@@ -77,7 +77,6 @@ def test_plugin_dependency_signature_namespace(memory_backend: MemoryChannelsBac
     assert app.signature_namespace["ChannelsPlugin"] is ChannelsPlugin
 
 
-@pytest.mark.flaky(reruns=5)
```
euri10, line 80: I removed all the flaky instances : it will definitely play bad with unclosed ressources and makes it hard to track down things

tests/unit/test_channels/test_plugin.py
```diff
@@ -113,68 +110,65 @@ async def test_pub_sub_run_in_background(channels_backend: ChannelsBackend, asyn
...
+#     with TestClient(app) as client:
+#         with client.websocket_connect("/ws/foo") as ws:
+#             channels_plugin.publish("something", "foo")
+#             assert ws.receive_text(timeout=2) == "something"
+#
+#         time.sleep(0.1)
+#
+#         with client.websocket_connect("/ws/bar") as ws:
+#             channels_plugin.publish("something else", "bar")
+#             assert ws.receive_text(timeout=2) == "something else"
```
euri10, line 171: those 2 tests are the only one where it feels like we have an issue closing things, I didnt manage to understand the issue but uncommenting those 2 and you'll have dangling sockets in subsequent tests

euri10:
closing as it's rather old and leading nowhere, the asynctestclient fix also showed small discrepencies between pytest-asyncio and anyio ;0

provinzkraut:
That being said, I think we should add some basic setup to test the test client with both `pytest-asyncio` and `anyio`.

euri10:
well i dont see how feasible that is since you have to choose one or the other no ?

provinzkraut:
> well i dont see how feasible that is since you have to choose one or the other no ?

Mmh. I think they might be able to co-exist if we were to mark all tests explicitly and turn off the `auto` mode for `pytest-asyncio`. Otherwise we'd have to do separate test suites, like we do for the Pydantic compat or server integration tests.

euri10:
ok I added a issue to track this

PR #3942, opened 2025-01-10
docs(v3): small fixes

JacobCoffee:
## Description

- fixes wording on table
- fixes code colorway (sidestream changes also in litestar-sphinx-theme v3 branch)
- (needs review?) add missing example files


docs/examples/static_files/upgrade_from_static_1.py
JacobCoffee, line 1: y are these even needed though?

euri10: I dont think so since this is deprecated in v2

pyproject.toml
```diff
@@ -105,17 +105,19 @@ structlog = ["structlog"]
...
 ]
 
-
 [tool.uv]
 default-groups = ["dev", "linting", "test", "docs"]
 
+#[tool.uv.sources]
+# Only needed when working on the theme; also need to uncomment docs package
+#litestar-sphinx-theme = { path = "../litestar-sphinx-theme", editable = true }
+
```
euri10, line 120: it feels weird to have to comment / uncomment to work on docs, or there's something I dont get, isn't there a better way ?

euri10 (approved):

pyproject.toml
cofin: Yes, I think you can use something like [this](https://github.com/litestar-org/litestar-htmx/blob/1be9b70db5208aedc9e46c521f55f2416ed2dd29/pyproject.toml#L64-L65)

JacobCoffee: Would this work? The issue I am fighting and that this solves currently is that I can have a locally checked out copy of `litestar-sphinx-theme` and can uncomment this section and the one in the `docs = [...]` group so that changes can be seen live across the project without having to push upstream or something.

cofin: Yep.  I think this should handle this case.

PR #3945, opened 2025-01-11
fix(ASGI mounts): Prevent accidental scope overrides by mounted ASGI apps

### LINKED ISSUE #3934: Bug: ASGIApp's with custom exception handlers result in Litestar throwing `AttributeError` when an exception occurs within the ASGI APP
### Description

If an ASGI App mounted using `@asgi("/", is_mount=True)` defines a custom exception handler, litestar will throw an `AttributeError` from the `litestar/middleware/_internal/exceptions/middleware.py` file when said ASGI App raises an exception.

It looks like the file takes `app: ASGIApp` but the middleware exception handling attempts to possibly treat it like `app: Litestar`. One way I have found which resolves this (and possibly introduces unintended side effects) is as follows:
```py
...

litestar_app = scope["app"]
from litestar import Litestar
if not isinstance(litestar_app, Litestar):
    # Unknown ASGI App, cannot handle automatically
    raise

if litestar_app.logging_config and (logger := litestar_app.logger):
    self.handle_exception_logging(logger=logger, logging_config=litestar_app.logging_config, scope=scope)

...
```

_No response_

### MCVE

```python
from litestar import Litestar, asgi
from litestar.types import Receive, Scope, Send
from starlette.exceptions import HTTPException
from fastapi import FastAPI

async def log_error(_, exc: HTTPException):
    raise exc

@asgi("/", is_mount=True)
async def admin(scope: Scope, receive: Receive, send: Send) -> None:
    app = FastAPI(exception_handlers={500: log_error})
    @app.get("/")
    async def route() -> None:
        raise Exception("sub app")
    await app(scope, receive, send)

app = Litestar(route_handlers=[admin])
```

### Steps to reproduce

```bash
1. Run the MCVE
2. Navigate to the site and observe the `AttributeError` being thrown by Litestar
```

### Screenshots

_No response_

### Logs

```bash

  File "/home/skelmis/.cache/pypoetry/virtualenvs/zero_website-Nzy3tQZa-py3.11/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
  File "/home/skelmis/.cache/pypoetry/virtualenvs/zero_website-Nzy3tQZa-py3.11/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
  File "/home/skelmis/.cache/pypoetry/virtualenvs/zero_website-Nzy3tQZa-py3.11/lib/python3.11/site-packages/litestar/app.py", line 615, in __call__
  File "/home/skelmis/.cache/pypoetry/virtualenvs/zero_website-Nzy3tQZa-py3.11/lib/python3.11/site-packages/litestar/middleware/_internal/exceptions/middleware.py", line 166, in __call__
    if litestar_app.logging_config and (logger := litestar_app.logger):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'FastAPI' object has no attribute 'logging_config'
```

### Litestar Version

```
poetry show litestar
 name         : litestar                                                                        
 version      : 2.14.0
```

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
When mounting ASGI apps, there's no guarantee they won't overwrite some key in the `scope` that we rely on, e.g. `scope["app"]`, which is what caused #3934.

To prevent this, I've implemented two things:

1. Do not store the Litestar instance under the generic `app` key, but the more specific `litestar_app` key. I've also added a `Litestar.from_scope` method, which can be used to safely access the current app from the sope
2. Added a new parameter `copy_scope` to the ASGI route handler, which, when set to `True` will copy the scope before calling into the mounted ASGI app. This should make things behave more as expected, since it truly give the called app its own environment without causing any side-effects. Since this change might break some things, I've left it with a default of `None`, which does not copy the scope, but will issue a warning if the mounted app modified it, enabling users to decide how to deal with that situation

Fixes #3934

### DIFF
--- a/docs/usage/applications.rst
+++ b/docs/usage/applications.rst
@@ -110,10 +110,10 @@ is accessible.
 :ref:`reserved keyword arguments <usage/routing/handlers:"reserved" keyword arguments>`.
 
 It is important to understand in this context that the application instance is injected into the ASGI ``scope`` mapping
-for each connection (i.e. request or websocket connection) as ``scope["app"]``. This makes the application
-accessible wherever the scope mapping is available, e.g. in middleware, on :class:`~.connection.request.Request` and
-:class:`~.connection.websocket.WebSocket` instances (accessible as ``request.app`` / ``socket.app``), and many
-other places.
+for each connection (i.e. request or websocket connection) as ``scope["litestar_app"]``, and can be retrieved using
+:meth:`~litestar.Litestar.from_scope`. This makes the application accessible wherever the scope mapping is available,
+e.g. in middleware, on :class:`~.connection.request.Request` and :class:`~.connection.websocket.WebSocket` instances
+(accessible as ``request.app`` / ``socket.app``), and many other places.
 
 Therefore, :paramref:`~.app.Litestar.state` offers an easy way to share contextual data between disparate parts
 of the application, as seen below:
--- a/docs/usage/applications.rst
+++ b/docs/usage/applications.rst
@@ -111,7 +111,7 @@ is accessible.
 
 It is important to understand in this context that the application instance is injected into the ASGI ``scope`` mapping
 for each connection (i.e. request or websocket connection) as ``scope["litestar_app"]``, and can be retrieved using
-:meth:`~litestar.Litestar.from_scope`. This makes the application accessible wherever the scope mapping is available,
+:meth:`~.Litestar.from_scope`. This makes the application accessible wherever the scope mapping is available,
 e.g. in middleware, on :class:`~.connection.request.Request` and :class:`~.connection.websocket.WebSocket` instances
 (accessible as ``request.app`` / ``socket.app``), and many other places.
 
--- a/litestar/testing/client/base.py
+++ b/litestar/testing/client/base.py
@@ -54,6 +54,7 @@ def fake_asgi_connection(app: ASGIApp, cookies: dict[str, str]) -> ASGIConnectio
         "http_version": "1.1",
         "extensions": {"http.response.template": {}},
         "app": app,  # type: ignore[typeddict-item]
+        "litestar_app": app,
         "state": {},
         "path_params": {},
         "route_handler": None,
--- a/docs/examples/application_hooks/after_exception_hook.py
+++ b/docs/examples/application_hooks/after_exception_hook.py
@@ -19,7 +19,7 @@ def my_handler() -> None:
 
 async def after_exception_handler(exc: Exception, scope: "Scope") -> None:
     """Hook function that will be invoked after each exception."""
-    state = scope["app"].state
+    state = Litestar.from_scope(scope).state
     if not hasattr(state, "error_count"):
         state.error_count = 1
     else:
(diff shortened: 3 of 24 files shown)


euri10:
dumb question maybe but what would happen if you mount a litestar app in a litestar app, won't you end up with the same issue you're fixing for other asgi apps ?

provinzkraut:
> dumb question maybe but what would happen if you mount a litestar app in a litestar app, won't you end up with the same issue you're fixing for other asgi apps ?

True, didn't think of that case as it's not something we really advertise and I can't immediately think of a use case. 

`copy_scope=True` would fix this though. Not sure if we want to special-case `app` in the scope too much, since mounted apps might modify other things as well, which is why I added the copy option, so issues with this can be prevented outright.

If we want to handle this one case in particular, we could just "restore" the original `app` key after calling the mounted ASGI app?

euri10:
The use case to is pretty much the same as this one: you mount something you have no control on ?

Realistically we'll have more Litestar users mounting FastAPI for the time being but you never know ;)

This was just a small remark, every asgi app is a squatter of the same namespace, but I'm all for squatts generally speaking so probably leaving this as is is a good option if ths copy_scope let the user solve the issue, it gives Litestar another edge other probably don't have.

litestar/testing/client/base.py
```diff
@@ -54,6 +54,7 @@ def fake_asgi_connection(app: ASGIApp, cookies: dict[str, str]) -> ASGIConnectio
         "http_version": "1.1",
         "extensions": {"http.response.template": {}},
         "app": app,  # type: ignore[typeddict-item]
```
euri10, line 56: so we need to keep the "app" key here ?

euri10 (approved):
excellent :)

docs/usage/applications.rst
```diff
@@ -110,10 +110,10 @@ is accessible.
 :ref:`reserved keyword arguments <usage/routing/handlers:"reserved" keyword arguments>`.
 
 It is important to understand in this context that the application instance is injected into the ASGI ``scope`` mapping
-for each connection (i.e. request or websocket connection) as ``scope["app"]``. This makes the application
-accessible wherever the scope mapping is available, e.g. in middleware, on :class:`~.connection.request.Request` and
-:class:`~.connection.websocket.WebSocket` instances (accessible as ``request.app`` / ``socket.app``), and many
-other places.
+for each connection (i.e. request or websocket connection) as ``scope["litestar_app"]``, and can be retrieved using
+:meth:`~litestar.Litestar.from_scope`. This makes the application accessible wherever the scope mapping is available,
```
JacobCoffee, line 114:
```suggestion
:meth:`~.Litestar.from_scope`. This makes the application accessible wherever the scope mapping is available,
```

JacobCoffee (approved):

litestar/testing/client/base.py
provinzkraut: Yeah, since we don't want to break a bunch of stuff that may rely on it. We're explicitly documenting this usage in our docs.

PR #3952, opened 2025-01-15
feat(routing): simplify route creation

provinzkraut:
Addition to #3900.

Remove unnecessary complexity and also add a nice performance boost for app startup :)

<hr>

There's a bunch of other stuff that we can remove due to the cleanup in #3900, but I'd rather add this in a separate PR as it doesn't strictly belong here.

### DIFF
--- a/litestar/app.py
+++ b/litestar/app.py
@@ -945,18 +945,6 @@ class Litestar(Router):
 
         return join_paths(output)
 
-    @property
-    def route_handler_method_view(self) -> dict[str, list[str]]:
-        """Map route handlers to paths.
-
-        Returns:
-            A dictionary of router handlers and lists of paths as strings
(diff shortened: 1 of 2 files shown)


litestar/app.py
```diff
@@ -971,18 +947,6 @@ def get_membership_details(group_id: int, user_id: int) -> None:
 
         return join_paths(output)
 
-    @property
-    def route_handler_method_view(self) -> dict[str, list[str]]:
```
provinzkraut, line 975: This was only used to construct the routes and isn't *really* necessary, as you can achieve the same with the `route_handler_method_map` attribute. 

Need to add a deprecation warning for this in 2.x

cofin (approved):
LGTM

PR #3953, opened 2025-01-15
chore: pin to Python 3.12 for docs commands only; fix contributing docs 

sloria:
## Description

- fix some typos in the contributing docs


sloria:
@all-contributors add @sloria for docs

sloria:
hm, i hope i did things right 😬

CONTRIBUTING.rst
```diff
@@ -40,12 +40,12 @@ the environment will bootstrap itself automatically. The steps below are for loc
...
 #. Run ``make install`` to create a `virtual environment <https://docs.python.org/3/tutorial/venv.html>`_
-   and install the required development dependencies or run the uv sync command manually:
+   and install the required development dependencies or run the ``uv sync`` command manually:
 
    .. code-block:: shell
-        :caption: Installing the documentation dependencies
+        :caption: Installing the development dependencies
 
-        uv install
+        uv sync --python 3.12
```
provinzkraut, line 48: Selecting a python version isn't really necessary here, and if you do select one, it should be the lowest supported version

provinzkraut (changes requested):

CONTRIBUTING.rst
sloria: this was to align with the Makefile here: https://github.com/litestar-org/litestar/blob/bb6aa49deb968da086a4b4d8089b16c4b2aede7e/Makefile#L32

should I change both to use 3.8?

fwiw, `uv sync` won't work its own doesn't work if you have python 3.13 installed. building the docs raises an error because `picologging` (which isn't python 3.13 compatible) isn't installed (it gets [conditionally installed on python<3.13](https://github.com/litestar-org/litestar/blob/518af0dda4e1fea75755147fee2a90dee778e121/pyproject.toml#L99)).

provinzkraut: Hmm. For `docs` we should probably install `--python=3.12` then and omit a specific version for everything else.

sloria: done in the latest commit

Makefile
```diff
@@ -29,7 +29,7 @@ upgrade:       										## Upgrade all dependencies to the latest stable versio
 
 .PHONY: install
 install:
-	@uv sync --python 3.12
+	@uv sync
```
sloria, line 32: might want to add `--frozen` so people don't inadvertently update the lockfile, but i'm not 100% sure so left that alone

provinzkraut (approved):

PR #3955, opened 2025-01-15
chore: remove unnecessary flags from install command in Makefile

sloria:
## Description

- Remove `--all-extras --dev`, as they're redundant with the default-groups defined here https://github.com/litestar-org/litestar/blob/518af0dda4e1fea75755147fee2a90dee778e121/pyproject.toml#L125


provinzkraut (approved):

provinzkraut:
Thanks for the housekeeping PRs @sloria! :)

PR #3957, opened 2025-01-15
chore: don't rebuild docs from scratch when using make docs-serve

sloria:
## Description

The docs take quite a long time to build from scratch (>1m on my old pre-Apple Silicon macbook air). After the initial build, they build quickly, so long as the `_build` dir isn't removed. I tend to kill then re-run sphinx-autobuild pretty often, so this ends up saving a lot of time

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ docs-clean: 										## Dump the existing built docs
 	@rm -rf docs/_build
 	@echo "=> Removed existing documentation build assets"
 
-docs-serve: docs-clean 								## Serve the docs locally
+docs-serve:  								## Serve the docs locally
 	@echo "=> Serving documentation"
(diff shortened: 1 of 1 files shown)


sloria:
also, sorry for the barrage of tiny PRs 😬  some people are pretty opinionated about these little DX things, so i opted for independent PRs so they could be merged separately

provinzkraut:
Mmh. Not sure this is ideal. There's a reason we do the clean and rebuild, because some things don't always rebuild correctly, so I'm a bit hesitant to merge this one. @JacobCoffee?

sloria:
yes, there are rare cases where stale build files end up in `docs/_build`, but imo the small cost of having to run `make docs-clean` manually for those cases doesn't outweigh the much faster builds for the common case.

sloria:
also note that `make docs-serve` is only used for local development, so having clean `docs/_build` isn't as critical. i left the `make docs` command alone, since it does get used in CI so having a clean build dir makes more sense there

Issue #3968, opened 2025-01-23
Enhancement: Provide resolved handler routes for testing


aedify-swi:
### Summary

Hej,

as I find it tedious and error-prone to enter the correct (layered) paths for my routes in tests, I created a little function to resolve the full path from my handler function object. I wasn't able to find anything on this in the docs. 

Would you be open to accept a PR adding this to the e.g. the AsyncTestClient to make developer ergonomics a little nicer?

### Basic Example

Definition: 

```python
def get_route_path(
    client: AsyncTestClient[Litestar],
    route_handler: HTTPRouteHandler,
    **path_parameters: Any,  # noqa: ANN401
) -> str:
    app = cast(Litestar, client.app)
    name = f"{route_handler._fn.__module__}.{route_handler._fn.__qualname__}"  # noqa: SLF001

    return app.route_reverse(name=name, **path_parameters)
```
Usage: 

```python
manual_path = f"neighbourhood/house/things/view/{uuid4().hex}" # ugh, tedious

new_and_improved_path = get_route_path(
      test_client, ThingController.get_this_thing, id=uuid4()
  ) # passing the function object enables better IDE support

response = await test_client.get(new_and_improved_path)
```

Proposed new method on AsyncTestClient:

```python
class AsyncTestClient(AsyncClient, BaseTestClient, Generic[T]):  # type: ignore[misc]
    ...
    url_for_handler( # or something like that
        self,
        route_handler: HTTPRouteHandler,
        **path_parameters: Any,  # noqa: ANN401
) -> str:
    name = f"{route_handler._fn.__module__}.{route_handler._fn.__qualname__}"  # noqa: SLF001
    return self.app.route_reverse(name=name, **path_parameters)
```

### Drawbacks and Impact

Drawbacks: A slightly larger API surface to maintain.
Impact: Reduced chance of typing related errors in tests.

### Unresolved questions

1. Is there a better approach? Accessing a private field (_fn) like that always feels icky.
2. Somewhat related: Would you be open to allowing UUIDs to be passed as strings into route_reverse? (Works fine locally when adding UUID to `allow_str_instead = {datetime, date, time, timedelta, float, Path, UUID}` in `route_reverse`)

euri10:
just to be sure I follow you

if you have
```python
@get("/a_given_path",)
async def a_handler_name()
```
[...]

PR #3972, opened 2025-01-24
feat: allow route_reverse params of type uuid to be passed as str

aedify-swi:
## Description

This allows params of type UUID to be passed as strings (e.g. their hex representation) into route_reverse (just like datetime, Path etc.). I don't know if the test is really necessary, but that is for the maintainers to decide.

Originally I intended to allow int too, but this broke about 20 tests I did not feel confident to fix them all in a reasonable amount of time.

Discussed in [#3968](https://github.com/litestar-org/litestar/issues/3968#issuecomment-2612178106)


provinzkraut (approved):
Thanks @aedify-swi!

PR #3973, opened 2025-01-25
docs: gunicorn with uvicorn workers

YorkSu:
## Description

- Add a page in the Deployments section for using Uvicorn workers with Gunicorn to deploy ASGI applications.

**NOTE**

The Gunicorn+Uvicorn pattern is considered legacy for ASGI deployments since [Uvicorn 0.30.0+](https://github.com/encode/uvicorn/releases/tag/0.30.0) includes native worker management. 

References:

- https://github.com/encode/uvicorn/releases/tag/0.30.0
- https://github.com/encode/uvicorn/pull/2183
- FastAPI old docs: https://github.com/fastapi/fastapi/blob/705659bb2277633e75269de9bac9d84cc3c15ae4/docs/en/docs/deployment/server-workers.md
- https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker/tree/master

## Closes

Part of #2609


provinzkraut (changes requested):
Don't think we should, as it's not really something I'd recommend anyway, and neither does uvicorn, since they've improved their own proc management. 

@litestar-org/members @litestar-org/maintainers if you feel differently, please weigh in

euri10:
I wouldnt recommend it either, but then we need to update the issue linked so that efforts like this aren't made for nothing.

provinzkraut:
> I wouldnt recommend it either, but then we need to update the issue linked so that efforts like this aren't made for nothing.

Good point. Done.

YorkSu:
I wouldnt recommend it either.

Considering that users might search for the keyword Gunicorn, would you consider adding a brief description to the `manually-with-asgi-server` section?

provinzkraut:
> I wouldn't recommend it either.
> 
> Considering that users might search for the keyword `Gunicorn`, would you consider adding a brief description to the `manually-with-asgi-server` section?
> 
> Since some developers might not be aware that the latest Uvicorn versions no longer need to deploy with Gunicorn for multi-workers, providing it could be helpful.

I think it should be *just* that notice

provinzkraut (approved):

PR #3963, opened 2025-01-21
fix(docs): improve text color for dark theme in mermaid diagrams

oek1ng:
## Description

- Changed mermaid diagrams text color for dark theme to black

## Closes
#3950

PR #3974, opened 2025-01-26
docs: use transparent bg for mermaid

sobolevn:
This will remove white backgroud from graphs.

Before:
изображение

Refs https://github.com/litestar-org/litestar/pull/3963


sobolevn:
I will post "after" when preview will be ready :)

sobolevn:
It is way more complex than that. I will get back to it later.

Issue #3979, opened 2025-01-28
Enhancement: Support properties in DTOs


sloria:
### Summary

DataclassDTO (and potentially other DTO types--I only tested with a Dataclass) don't seem to support properties.

### Basic Example

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "litestar",
#     "uvicorn",
# ]
# ///
from dataclasses import dataclass

import uvicorn
from litestar import Litestar, get
from litestar.dto import DataclassDTO, DTOConfig


@dataclass
class User:
    name: str

    @property
    def name_upper(self) -> str:
        return self.name.upper()


class UserDTO(DataclassDTO[User]):
    config = DTOConfig(
        include={"name_upper"},  # Doesn't work
    )


@get("/", dto=UserDTO)
async def get_user() -> User:
    return User(name="Alice")


app = Litestar(route_handlers=[get_user])

if __name__ == "__main__":
    uvicorn.run(app)
```

```console-session
$ uv run repro.py
$ curl http://127.0.0.1:8000
```

Expected:

```
{"name_upper": "ALICE"}
```

Actual:

```
{}
```

provinzkraut:
Thanks for bringing this up @sloria, this should definitely be supported! :)

PR #3981, opened 2025-02-01
feat(DTOs): Support `@property` fields for msgspec and dataclass

provinzkraut:
Implements #3979

### DIFF
--- a/tests/unit/test_dto/test_factory/test_dataclass_dto.py
+++ b/tests/unit/test_dto/test_factory/test_dataclass_dto.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import sys
 from dataclasses import dataclass, field, replace
 from typing import ClassVar, List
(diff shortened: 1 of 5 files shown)


tests/unit/test_dto/test_factory/test_dataclass_dto.py
```diff
@@ -139,3 +170,22 @@ class Model:
...
     assert fields[1].dto_field == DTOField("read-only")
+
+
+def test_property_underscore_exclude() -> None:
+    @dataclass
+    class Model:
+        one: str
+
+        @property
+        def _computed(self) -> int:
```
Alc-Alc, line 181: for my understanding, why is this not covering the test case that codecov is complaining about?

provinzkraut: Because it's complaining about this for the msgspec DTO :)

Alc-Alc (approved):
🔥

Issue #3912, opened 2024-12-23
Bug: "type" keyword doesn't behave the same as TypeAlias for Dependency


Rubikoid:
### Description

Dependency's `skip_validation=True` parameter don't work for `type` keyword 

### MCVE

```python
from typing import Annotated, TypeAlias

from litestar import Litestar, get
from litestar.di import Provide
from litestar.params import Dependency


async def dep_eq() -> int:
    return 0


async def dep_type() -> int:
    return 1


async def dep_alias() -> int:
    return 2


ARG_EQ = Annotated[str, Dependency(skip_validation=True)]
type ARG_TYPE = Annotated[str, Dependency(skip_validation=True)]
ARG_ALIAS: TypeAlias = Annotated[str, Dependency(skip_validation=True)]  # noqa: UP040


@get(
    "/broken",
    dependencies={
        "arg_eq": Provide(dep_eq),
        "arg_type": Provide(dep_type),
        "arg_alias": Provide(dep_alias),
    },
)
async def broken(
    arg_eq: ARG_EQ,
    arg_type: ARG_TYPE,
    arg_alias: ARG_ALIAS,
) -> str:
    return (
        f"{arg_eq = } ({type(arg_eq) = }), "
        f"{arg_type = } ({type(arg_type) = }), "
        f"{arg_alias = } ({type(arg_alias) = })"
    )


@get(
    "/working",
    dependencies={
        "arg_eq": Provide(dep_eq),
        "arg_alias": Provide(dep_alias),
    },
)
async def working(
    arg_eq: ARG_EQ,
    arg_alias: ARG_ALIAS,
) -> str:
    return (
        f"{arg_eq = } ({type(arg_eq) = }), "  #
        f"{arg_alias = } ({type(arg_alias) = })"
    )


app = Litestar(
    debug=True,
    route_handlers=[
        broken,
        working,
    ],
)
```

### Steps to reproduce

```bash
1. Go to `/docs`
2. Execute `/broken`
3. See error

Expected output:

arg_eq = 0 (type(arg_eq) = <class 'int'>), arg_type = 1 (type(arg_type) = <class 'int'>), arg_alias = 2 (type(arg_alias) = <class 'int'>)
```
```


### Screenshots

_No response_

### Logs

```bash
ERROR - 2024-12-23 19:36:57,329 - litestar - config - Uncaught exception (connection_type=http, path=/broken):
Traceback (most recent call last):
[...]

PR #3982, opened 2025-02-01
fix(typing): always unwrap `TypeAliasType`s

provinzkraut:
While fixing #3912 I noticed we only unwrap these partially. However, I am not sure if unwrapping them *at all* is proper, because of a particular edge case: They can be self-referential. This is used e.g. in `pydantic.JsonType`.

`SelfReferential = TypeAliasType("Outer", Union[Annotated[int, "meta"], list["Outer"]])`

This type can be understood by e.g. mypy, but is hard to parse for our typing system since we want to extract the metadata from the `Annotated`, but need to keep the outer `TypeAliasType`, so the self-referencing works in later stages (e.g. Pydantic).

I'm not entirely sure how we should best handle this, as our typing system has fundamentally different requirements than that of the 3rd party tools we rely on. 

Divergences in how self-referential types are treated by these tools further complicates things.

For example, this is fine in Pydantic:

```python
import pydantic

type Foo = int | list[Foo]

class SomeModel(pydantic.BaseModel):
    a: Foo

SomeModel.model_validate_json('{"a": 1}')
```

But msgspec raises a `RecursionError`:

```python
import msgspec

type Foo = int | list[Foo]

class SomeStruct(msgspec.Struct):
    a: Foo

msgspec.json.decode('{"a": 1}', type=SomeStruct)
```

<hr>

Not really sure yet how to proceed, I'll have to think about this a bit. I'd be happy about some input though @litestar-org/members

### DIFF
--- a/litestar/utils/typing.py
+++ b/litestar/utils/typing.py
@@ -37,7 +37,16 @@ from typing import (
     cast,
 )
 
-from typing_extensions import Annotated, NewType, NotRequired, Required, get_args, get_origin, get_type_hints
+from typing_extensions import (
+    Annotated,
+    NewType,
+    NotRequired,
+    Required,
+    TypeAliasType,
+    get_args,
+    get_origin,
+    get_type_hints,
+)
 
 from litestar.types.builtin_types import NoneType, UnionTypes
 
@@ -128,7 +137,7 @@ This is necessary because occasionally we want to rebuild a generic outer type w
 ``collections.abc.Mapping``, are not valid generic types in Python 3.8.
 """
 
-wrapper_type_set = {Annotated, Required, NotRequired}
+wrapper_type_set = {Annotated, Required, NotRequired, TypeAliasType}
 """Types that always contain a wrapped type annotation as their first arg."""
 
 
@@ -151,7 +160,7 @@ def make_non_optional_union(annotation: UnionT | None) -> UnionT:
 
 
 def unwrap_annotation(annotation: Any) -> tuple[Any, tuple[Any, ...], set[Any]]:
-    """Remove "wrapper" annotation types, such as ``Annotated``, ``Required``, and ``NotRequired``.
+    """Remove "wrapper" annotation types, such as ``Annotated``, ``Required``, ``NotRequired`` or ``TypeAliasType``.
 
     Note:
         ``annotation`` should have been retrieved from :func:`get_type_hints()` with ``include_extras=True``. This
@@ -163,14 +172,30 @@ def unwrap_annotation(annotation: Any) -> tuple[Any, tuple[Any, ...], set[Any]]:
     Returns:
         A tuple of the unwrapped annotation and any ``Annotated`` metadata, and a set of any wrapper types encountered.
     """
-    origin = get_origin(annotation)
-    wrappers = set()
+
     metadata = []
-    while origin in wrapper_type_set:
-        wrappers.add(origin)
-        annotation, *meta = get_args(annotation)
-        metadata.extend(meta)
-        origin = get_origin(annotation)
+    wrappers = set()
+
+    stack = [annotation]
+
+    while stack:
+        ann = stack.pop()
+
+        if isinstance(ann, TypeAliasType):
+            wrappers.add(TypeAliasType)
+            stack.append(ann.__value__)
+            continue
+
+        origin = get_origin(ann)
+        if origin in wrapper_type_set:
+            ann, *meta = get_args(ann)
+            metadata.extend(meta)
+            wrappers.add(origin)
+            stack.append(ann)
+            continue
+
+        return ann, tuple(metadata), wrappers
+
     return annotation, tuple(metadata), wrappers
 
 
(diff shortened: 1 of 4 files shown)


litestar/utils/typing.py
```diff
@@ -163,14 +172,30 @@ def unwrap_annotation(annotation: Any) -> tuple[Any, tuple[Any, ...], set[Any]]:
...
+    wrappers = set()
+
+    stack = [annotation]
+
+    while stack:
+        ann = stack.pop()
+
+        if isinstance(ann, TypeAliasType):
+            wrappers.add(TypeAliasType)
+            stack.append(ann.__value__)
```
sobolevn, line 186: This line can raise all kinds of exceptions. Do we have cases like `NameError` and `TypeError` covered in tests?

tests/unit/test_typing.py
```diff
@@ -474,3 +476,34 @@ def test_unwrap_type_alias_type_keyword() -> None:
...
+        (TypeAliasType("SomeAlias", Annotated[int, "meta"]), ("meta",), int),
+        (TypeAliasType("SomeAlias", Annotated[TypeAliasType("InnerAlias", int), "meta"]), ("meta",), int),
+        (Annotated[TypeAliasType("SomeAlias", Annotated[int, "inner meta"]), "meta"], ("meta", "inner meta"), int),
+        (NestedAlias, ("meta",), NestedAlias),
+    ],
+)
+def test_unwrap_annotation_type_alias_type_nested(
+    annotation: Any, expected_meta: tuple[str, ...], expected_type: Any
+) -> None:
+    annotation, metadata, wrappers = unwrap_annotation(annotation)
```
sobolevn, line 506: Question: is it fine that `unwrap_annotation` is not tested on `type` keyword aliases?

sobolevn (commented):
I also don't quite know on how to proceed. I think that probably documenting different corner cases is a must here. For example: how different serializer frameworks handle recursive aliases and when to use them. Maybe some recipies. 

Thanks for raising this! I am really interested in this topic.

tests/unit/test_typing.py
provinzkraut: Hmm. I was under the assumption that `type` always creates a `TypeAliasType`. At least that's how I understand the docs: https://docs.python.org/3.13/library/typing.html#typing.TypeAliasType

litestar/utils/typing.py
provinzkraut: Could you elaborate what kind of exception scenarios you're thinking of here? My thinking was that, if this does raise an exception, it's okay to let it bubble up since we can't properly deal with it

tests/unit/test_typing.py
sobolevn: Yes, it does. But, `TypeVarType` can be constructed manually or by the compiler. There were bugs that showed differences.

litestar/utils/typing.py
sobolevn: `type A = Undefined`

Yeah, I agree that it is fine, we just need to test these cases.

provinzkraut:
I think this might be related: https://github.com/python/mypy/issues/18252

Mypy also doesn't seem to handle these cyclic definitions

litestar/utils/typing.py
```diff
@@ -163,14 +172,30 @@ def unwrap_annotation(annotation: Any) -> tuple[Any, tuple[Any, ...], set[Any]]:
...
-        metadata.extend(meta)
-        origin = get_origin(annotation)
+    wrappers = set()
+
+    stack = [annotation]
+
+    while stack:
+        ann = stack.pop()
+
+        if isinstance(ann, TypeAliasType):
```
Viicos, line 184: Just a heads up: `typing-extensions` 4.13.0 (yet to be released) has a backported `TypeAliasType` version, meaning:

```python
from typing import TypeAliasType
from typing_extensions import TypeAliasType as ExtensionsTypeAliasType

Test = ExtensionsTypeAliasType('Test', int)
isinstance(Test, TypeAliasType)
#> False
```

Also on Python 3.10 a bug leads to the following:

```python
T = TypeVar('T')

Test = TypeAliasType('Test', list[T], type_params=(T,))
isinstance(Test[int], TypeAliasType)
#> True, but should be false as Test[int] is a `types.GenericAlias` instance`. In your case, your are going to loose the parametrization and `__value__` will be list[T]
```

[This is how we do it in Pydantic.](https://github.com/pydantic/pydantic/blob/61cfdad7c80dafb3094620fd8b7610aab291f8ac/pydantic/_internal/_typing_extra.py#L406-L425)

Overall it seems that you may face similar issues in the future if typing-extensions starts exposing new versions of the `typing.*` constructs. One example is also in this function: you extract type qualifiers such as `Required`, and latter in the code base a `Required in type_wrappers` is performed. Currently, `typing_extensions` reexports `typing.(Not)Required` on 3.11, but that might change in the future.

provinzkraut: Thanks for the heads up @Viicos! Seems like another indicator that what was discussed in https://github.com/beartype/beartype/discussions/479#discussioncomment-11811247 would be quite useful

PR #3984, opened 2025-02-02
feat(core): Cancel handler functions / responses when client disconnects

provinzkraut:
poc for listening for early client disconnects and cancelling handler functions / responses in progress.

<hr>

I currently consider this not viable, since it imposes a 50% performance penalty when looking at throughput of a simple plaintext benchmark. This is due to the overhead introduced by running the response cycle in an `anyio.TaskGroup`, which is considerable. There is a pure asyncio solution, which is a bit more performant, but overhead is still in the same ballpark.

<hr>

This also surfaces some flaws in how we're handling static files, which is what's causing most of the test failures, and would need to be addressed.

<hr>

### Open questions

- Should background tasks be cancelled?
- Should background tasks run if the response hasn't finished due to the client disconnecting?

### DIFF
--- a/litestar/routes/http.py
+++ b/litestar/routes/http.py
@@ -1,19 +1,24 @@
 from __future__ import annotations
 
+from functools import partial
 from itertools import chain
 from typing import TYPE_CHECKING, Any
 
+import anyio
 from msgspec.msgpack import decode as _decode_msgpack_plain
 
 from litestar.datastructures.multi_dicts import FormMultiDict
 from litestar.enums import HttpMethod, MediaType, ScopeType
 from litestar.exceptions import ClientException, ImproperlyConfiguredException, SerializationException
+from litestar.exceptions.base_exceptions import ClientDisconnect
 from litestar.handlers.http_handlers import HTTPRouteHandler
 from litestar.response import Response
 from litestar.routes.base import BaseRoute
 from litestar.status_codes import HTTP_204_NO_CONTENT
 from litestar.types.empty import Empty
 from litestar.utils.scope.state import ScopeState
+from litestar.utils.sync import AsyncCallable
+from exceptiongroup import ExceptionGroup
 
 if TYPE_CHECKING:
     from litestar._kwargs import KwargsModel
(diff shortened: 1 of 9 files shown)


litestar/routes/http.py
```diff
@@ -78,14 +113,37 @@ async def handle(self, scope: HTTPScope, receive: Receive, send: Send) -> None:
...
+                            scope=scope,
+                            send=send,
+                            receive=receive,
+                            request=request,
+                            route_handler=route_handler,
+                            parameter_model=parameter_model,
+                            cancel_scope=tg.cancel_scope,
+                        ),
+                    )
+                    tg.start_soon(self._listen_for_disconnect, request, tg.cancel_scope)
```
provinzkraut, line 141: This is a more performant, but we can't really use it for reasons of anyio/trio compatibility.

```suggestion
tasks = [
    asyncio.create_task(
        self._handle_response_cycle(
            scope=scope,
            send=send,
            receive=receive,
            request=request,
            route_handler=route_handler,
            parameter_model=parameter_model,
        )
    ),
    asyncio.create_task(request._listen_for_disconnect()),
]
done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
done.pop().result()
if done:
    done.pop().result()

if pending:
    pending = pending.pop()
    pending.cancel()
    await pending
    pending.result()
```

provinzkraut:
What I could see is maybe having this be an opt-in feature. That would introduce additional complexity, but leave it up to the user to decide whether the overhead is worth it. I'd suspect though, that in most cases, a simple timeout on the request would suffice

FHU-yezi:
> What I could see is maybe having this be an opt-in feature. That would introduce additional complexity, but leave it up to the user to decide whether the overhead is worth it. I'd suspect though, that in most cases, a simple timeout on the request would suffice.

Make sense. For most of use cases, the action is not so heavy that we should cancel the handler when use disconnected. User disconnected should be considered as an exception, not a common situation.

The idempotency of API should be handled by backend logic, according to the HTTP method.

Kumzy:
@provinzkraut is this still something we want to do?

provinzkraut:
I think so, yes. I just don't have bandwidth to see it through at the moment. I'll see if I can get it done for v3

PR #3987, opened 2025-02-08
docs: dto rationale

euri10:
## Description

- add a topic on dto rationale

### DIFF
--- a/docs/topics/dto-rationale.rst
+++ b/docs/topics/dto-rationale.rst
@@ -1,0 +1,41 @@
+DTO rationale
+=============
+
+.. admonition:: What particular problem does DTO solve in Litestar ?
+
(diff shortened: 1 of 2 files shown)


docs/topics/dto-rationale.rst
```diff
@@ -0,0 +1,41 @@
+DTO rationale
+=============
+
+.. admonition:: What particular problem does DTO solve in Litestar ?
```
sobolevn, line 4: Maybe make this a section in the "Basic Use" doc of DTO? Maybe at the bottom with a link somewhere here:
Снимок экрана 2025-02-08 в 22 24 22

euri10: I agree it's good to know the why even if most people want the how !

I was looking at your suggestion, whcih I like, but I find  `0-basic-use.html` is already quite long and focued on the how, what about linking this dto-rationale just above in the index just before the content ?

20250209_0826_1081x661_1739085972

to something like :

20250209_0831_1041x539_1739086275

PR #3988, opened 2025-02-10
docs: Updated the Pydantic model configuration in Requests example with multiple files

Ada-lave:
## Description

- Updated the Pydantic model configuration in Requests example with multiple files


euri10 (approved):
lgtm thanks !

PR #3990, opened 2025-02-10
docs: Added information about the JSON-like logging format for the standard library

### LINKED ISSUE #3827: Docs: provide an example of using JSON-based logging
### Summary

https://docs.litestar.dev/latest/usage/logging.html features a lot of useful examples, but it does not have one with logging JSON data. I think that it is the de-facto standard these days and should be included.

RenameMe1:
## Description

- I added json format string example in `Using Python standard library` section
- I added information about format string style for `LogginConfig`

## Closes

#3827

### DIFF
--- a/docs/usage/logging.rst
+++ b/docs/usage/logging.rst
@@ -127,7 +127,7 @@ The above example is the same as using logging without the litestar ``LoggingCon
     )
 
 Formatter like JSON string
-----------------
+--------------------------
 
 * printf-style String Formatting
 
--- a/docs/usage/logging.rst
+++ b/docs/usage/logging.rst
@@ -80,7 +80,7 @@ below:
         logging_config=logging_config,
     )
 
-If you want to use ``LoggingConfig`` with another format string style, you need add a key "style" 
+If you want to use ``LoggingConfig`` with another format string style, you need add a key "style"
 
 .. code-block:: python
 
@@ -134,7 +134,7 @@ Formatter like JSON string
 .. code-block:: python
 
    format = ('{"asctime": "%(asctime)s", "name": "%(name)s", "level": "%(levelname)s", "message": "%(message)s"}')
-   
+
    ch.setFormatter(logging.Formatter(format))
 
 * str.format() String Formatting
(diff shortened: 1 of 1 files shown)


RenameMe1:
A friendly ping @provinzkraut. Could you help me get a review for this PR?

provinzkraut:
Not sure if this belongs in the Litestar logging tutorial, as it strictly deals with stdlib logging configuration. Maybe it would be better to link to the stdlib logging docs?

RenameMe1:
> Not sure if this belongs in the Litestar logging tutorial, as it strictly deals with stdlib logging configuration. Maybe it would be better to link to the stdlib logging docs?

There are no such examples in the stdlib logging documentation. They are also not listed in the stdlib cookbook. 

Maybe I can create a for `litestar.logging.config.py `the 'JsonFormat` class that will return a string in json format? Usage example:

``` python
from litestar.logging import LoggingConfig, JsonFormat


@get("/")
def my_router_handler(request: Request) -> None:
    request.logger.info("inside a request")
    return None


logging_config = LoggingConfig(
    root={"level": "INFO", "handlers": ["queue_listener"]},
    formatters={
        "standard": {"format": JsonFormat(["%(asctime)s", "%(name)s", "%(levelname)s", "%(message)s"])} # {'asctime': '%(asctime)s', 'name': ' %(name)s', 'levelname': ' %(levelname)s', 'message': ' %(message)s'}
    },
    log_exceptions="always",
)

app = Litestar(route_handlers=[my_router_handler], logging_config=logging_config)
```

I think this is a useful feature.

provinzkraut:
> There are no such examples in the stdlib logging documentation. They are also not listed in the stdlib cookbook.

Sounds to me like you should make a PR to the stdlib docs then :)
If this is a generally good thing to know, and not specific to Litestar, others might benefit from it as well!

RenameMe1:
Thanks for feedback. It's was very usefull

RenameMe1:
Thanks for feedback. It's was very usefull

PR #3992, opened 2025-02-12
Improve few documentation pages

eandersons:
## Description

The main goal was to update code examples in tutorial to include imports as well, otherwise it can be quite annoying to scroll between text and full code to see from where a specific class or function is imported if one is writing code from scratch following the tutorial.

Other part of changes are related to minor typo, grammar, word placement, formatting (both text and code) and similar fixes, as well as minor example code improvements.

Some changes might be a bit too opinionated, but I could not resit to make them, and in some diffs they might bury the actually needed changes.

### DIFF
--- a/docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
+++ b/docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
@@ -19,42 +19,43 @@ from litestar.plugins.sqlalchemy import (
     AsyncSessionConfig,
     SQLAlchemyAsyncConfig,
     SQLAlchemyInitPlugin,
-    base,
-    filters,
-    repository,
 )
+from litestar.plugins.sqlalchemy.base import UUIDAuditBase, UUIDBase
+from litestar.plugins.sqlalchemy.filters import LimitOffset
+from litestar.plugins.sqlalchemy.repository import SQLAlchemyAsyncRepository
 
 if TYPE_CHECKING:
     from sqlalchemy.ext.asyncio import AsyncSession
 
 
 class BaseModel(_BaseModel):
-    """Extend Pydantic's BaseModel to enable ORM mode"""
+    """Extended Pydantic's `BaseModel` to enable the ORM mode."""
 
     model_config = {"from_attributes": True}
 
 
-# The SQLAlchemy base includes a declarative model for you to use in your models.
-# The `UUIDBase` class includes a `UUID` based primary key (`id`)
-class AuthorModel(base.UUIDBase):
-    # we can optionally provide the table name instead of auto-generating it
+# The SQLAlchemy base includes a declarative model for you to use in your
+# models. The `UUIDBase` class includes an `UUID` based primary key (`id`).
+class AuthorModel(UUIDBase):
+    # We can optionally provide the table name instead of auto-generating it.
     __tablename__ = "author"  #  type: ignore[assignment]
     name: Mapped[str]
     dob: Mapped[date | None]
     books: Mapped[list[BookModel]] = relationship(back_populates="author", lazy="noload")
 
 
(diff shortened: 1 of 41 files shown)


docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
```diff
@@ -19,42 +19,43 @@
...
 
     model_config = {"from_attributes": True}
 
 
-# The SQLAlchemy base includes a declarative model for you to use in your models.
-# The `UUIDBase` class includes a `UUID` based primary key (`id`)
-class AuthorModel(base.UUIDBase):
-    # we can optionally provide the table name instead of auto-generating it
+# The SQLAlchemy base includes a declarative model for you to use in your
+# models. The `UUIDBase` class includes an `UUID` based primary key (`id`).
```
provinzkraut, line 38: *a* is actually correct here. "n before a vowel" refers to the vowel *sound*, not whether the word starts with a letter that's a vowel. This applies to most of the changes made in this PR.

https://www.scribbr.com/commonly-confused-words/a-vs-an/

docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
```diff
@@ -19,42 +19,43 @@
...
+from litestar.plugins.sqlalchemy.filters import LimitOffset
+from litestar.plugins.sqlalchemy.repository import SQLAlchemyAsyncRepository
 
 if TYPE_CHECKING:
     from sqlalchemy.ext.asyncio import AsyncSession
 
 
 class BaseModel(_BaseModel):
-    """Extend Pydantic's BaseModel to enable ORM mode"""
+    """Extended Pydantic's `BaseModel` to enable the ORM mode."""
```
provinzkraut, line 32: The phrasing here is intentional, as it's meant to be instructive rather than descriptive. Although it should probably go in a regular comment not a docstring

docs/examples/security/jwt/using_jwt_auth.py
```diff
@@ -22,22 +22,26 @@ class User(BaseModel):
...
 
-# JWTAuth requires a retrieve handler callable that receives the JWT token model and the ASGI connection
-# and returns the 'User' instance correlating to it.
+# JWTAuth requires a retrieve handler callable that receives the JWT token
+# model and the ASGI connection and returns the `User` instance correlating to
+# it.
 #
 # Notes:
-# - 'User' can be any arbitrary value you decide upon.
+# - `User` can be any arbitrary value you decide upon.
```
provinzkraut, line 30: The quotes here are intentional as well. Since this isn't rendered anywhere, backticks don't have a special representation, so using standard quotes is better for readability

provinzkraut:
Thanks for the effort @eandersons, but I'm afraid this PR is a bit too broad, and a lot of the changes would need to be rolled back. I didn't go through all of them because it's a bit much, but I think you get the gist.

If you still wish to contribute to the documentation, I suggest you start with a smaller, more focused PR.

docs/examples/contrib/sqlalchemy/sqlalchemy_async_repository.py
eandersons: Yes, you are right. I did not pay attention that this is inside a documentation string, so it was quite fine as it were.

docs/examples/security/jwt/using_jwt_auth.py
eandersons: Well, the term "readability" sometimes can be quite subjective.
As I see it, quotes and double quotes are used for grammatical constructs, so words or phrases wrapped in backticks (be they single or double) indicates that the value is related to code (in this case it is the name of a class). It would be less debatable if instead of just 'User' it would be "The class name 'User' [..]" (or "Name of the class 'User' [..]"; I am not too sure which would be more correct).

But anyways, your project - your rules.

PR #3994, opened 2025-02-13
docs: Add missing port to url

pech2:
## Description

The URL is missing the default port so it doesn't load correctly. I added it like all the other URLs to 127.0.0.1 on the page.


provinzkraut (approved):

provinzkraut:
Thanks @pech2!

PR #3996, opened 2025-02-13
feat(core): new base middleware

provinzkraut:
Add a new base middleware class to facilitate easier configuration and middleware dispatching.

The new `ASGIMiddleware` features the same functionality as the `AbstractMiddleware`, but makes it easier to pass configuration directly to middleware classes without a separate configuration object, allowing to get rid of the `DefineMiddleware` pattern entirely.

<hr> 

If we are to add this, I propose we do so in v2, and migrate existing middlewares to this pattern in v3.

### DIFF
--- a/litestar/middleware/base.py
+++ b/litestar/middleware/base.py
@@ -169,6 +169,7 @@ class ASGIMiddleware(abc.ABC):
         An instance of the individual middleware's will be created *once* and used to
         build up the internal middleware stack. As such, middlewares should *not* be
         stateful, as this state will be shared across all requests.
+        Any connection-specific state should be scoped to the `handle` implementation. Not doing so would typically lead to conflicting variable reads/writes across requests, and most likely bugs.
 
     .. example::
 
--- a/litestar/middleware/base.py
+++ b/litestar/middleware/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Callable, Protocol, runtime_checkable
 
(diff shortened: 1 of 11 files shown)


litestar/middleware/base.py
```diff
@@ -149,3 +155,83 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+    to dynamically skip the middleware based on ASGI ``scope["type"]``, handler ``opt``
+    keys or path patterns and a simple way to pass configuration to middlewares.
+
+    This base class does not implement an ``__init__`` method, so subclasses are free
+    to use it to customize the middleware's configuration.
+
+    .. important::
+        An instance of the individual middleware's will be created *once* and used to
+        build up the internal middleware stack. As such, middlewares should *not* be
+        stateful, as this state will be shared across all requests.
```
euri10, line 171:
```suggestion
        stateful, as this state will be shared across all requests.
        Any connection-specific state should be scoped to the `handle` implementation. Not doing so would typically lead to conflicting variable reads/writes across requests, and most likely bugs.
```

euri10: Adapted from the Starlette docs here: https://www.starlette.io/middleware/#asgi-middleware-should-be-stateless 
The do/dont could be soething interesting to add to our docs too

euri10 (approved):
lgtm
maybe add an example of transition from previous to new implementation ?

litestar/middleware/base.py
provinzkraut: > The do/dont could be soething interesting to add to our docs too

Yeah, good idea

provinzkraut:
> lgtm maybe add an example of transition from previous to new implementation ?

Yeah, I want to expand the docs in general a bit

euri10 (approved):

PR #4010, opened 2025-02-16
feat(core)!: drop Python 3.8 support

provinzkraut:
Drop support for Python 3.8. 

**How to review this?**

This is a rather big PR, however, almost all changes have been made automatically by ruff. I've put those into separate commits, so I suggest reviewing the commits *not* consisting of ruff changes individually.


tools/prepare_release.py
```diff
@@ -10,12 +10,15 @@
...
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Generator
+from typing import TYPE_CHECKING
 
 import click
 import httpx
 import msgspec
 
+if TYPE_CHECKING:
```
sobolevn, line 19: Sorry, why is this import under `TYPE_CHECKING`?

sobolevn (commented):
Great to see old code removed! 🎉

tools/prepare_release.py
provinzkraut: No idea. That's something ruff did automatically :shrug:

sobolevn: this does not have any runtime benefits, because we already import `typing` and `typing` imports `collections.abc`, so I guess it is safe to simplify the code. (moreover, we don't really care about import-time speed in a `tools/` file)

provinzkraut: Yeah, I know it doesn't really make sense. But I don't want to fight with ruff. I prefer just letting it do what it feels like doing, so I don't have to think about it :)

sobolevn: sounds good to me!

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_async_serialization_dto.py
```diff
@@ -8,7 +8,7 @@
 from litestar.plugins.sqlalchemy import SQLAlchemyDTO
 
 if TYPE_CHECKING:
-    from typing import List
+    pass
```
cofin, line 11: I think we can just remove the entire block

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_async_serialization_plugin.py
```diff
@@ -8,7 +8,7 @@
 from litestar.plugins.sqlalchemy import SQLAlchemySerializationPlugin
 
 if TYPE_CHECKING:
-    from typing import List
+    pass
```
cofin, line 12: I think we can remove the TYPE_CHECKING import

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_async_serialization_plugin_marking_fields.py
```diff
@@ -9,7 +9,7 @@
 from litestar.plugins.sqlalchemy import SQLAlchemySerializationPlugin
 
 if TYPE_CHECKING:
-    from typing import List
+    pass
```
cofin, line 12: Same

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_sync_serialization_plugin.py
```diff
@@ -8,7 +8,7 @@
 from litestar.plugins.sqlalchemy import SQLAlchemySerializationPlugin
 
 if TYPE_CHECKING:
-    from typing import List
+    pass
```
cofin, line 11: Same

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_sync_serialization_plugin_marking_fields.py
```diff
@@ -9,7 +9,7 @@
 from litestar.plugins.sqlalchemy import SQLAlchemySerializationPlugin
 
 if TYPE_CHECKING:
-    from typing import List
+    pass
```
cofin, line 12: And another

litestar/handlers/websocket_handlers/listener.py
```diff
@@ -281,8 +277,8 @@ def _validate_handler_function(self) -> None:
     async def default_connection_lifespan(
         self,
         socket: WebSocket,
-        on_accept_dependencies: Optional[Dict[str, Any]] = None,  # noqa: UP006, UP007
-        on_disconnect_dependencies: Optional[Dict[str, Any]] = None,  # noqa: UP006, UP007
+        on_accept_dependencies: Optional[dict[str, Any]] = None,  # noqa: UP007
```
cofin, line 280: Should we remove `UP007` and the `Optional` as well?

litestar/plugins/pydantic/utils.py
```diff
@@ -337,7 +335,7 @@ def _create_field_definition_v1(  # noqa: C901
             )
             field_definition_kwargs["raw"] = field_annotation
             # on < 3.9, these builtins are not generic
-            origin = get_safe_generic_origin(None, field_annotation.__origin__)
+            origin = get_safe_generic_origin(field_annotation.__origin__, field_annotation.__origin__)
```
cofin, line 338: Is this correct?

litestar/handlers/websocket_handlers/listener.py
FHU-yezi: Sure because Python 3.8 also have support of `|` operator as `Union`, we can use `SomeType | None`.

tests/helpers.py
```diff
@@ -25,16 +26,6 @@
 RANDOM = random.Random(b"bA\xcd\x00\xa9$\xa7\x17\x1c\x10")
 
 
-# TODO: Remove when dropping 3.9
```
euri10, line 28: not sure why it is removed given the comment

tests/unit/test_dto/test_factory/test_dataclass_dto.py
```diff
@@ -31,67 +29,6 @@ def fx_dto_type() -> type[DataclassDTO[Model]]:
     return DataclassDTO[Model]
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 9), reason="generic builtin collection")
```
euri10, line 34: same it is removed but we drop 3.8 only no ?

tests/unit/test_utils/test_typing.py
```diff
@@ -3,10 +3,9 @@
 from __future__ import annotations
 
 from sys import version_info
-from typing import Any, Dict, Generic, List, Optional, TypeVar, Union
+from typing import Annotated, Any, Dict, Generic, List, Optional, TypeVar, Union  # noqa: UP035
```
euri10, line 6: i dont get the UP035

docs/examples/contrib/sqlalchemy/plugins/sqlalchemy_async_serialization_dto.py
provinzkraut: Yeah, no idea why ruff doesn't do these automatically?

litestar/plugins/pydantic/utils.py
provinzkraut: Yeah, it just means that in case the annotation isn't defined in our "safe generic origins" map, we'll return the original annotation

tests/helpers.py
provinzkraut: The comment was wrong. If you look at the condition, it checks if the version is *less than* 3.9, so the `if` branch was only being taken on 3.8. `randbytes` was added in 3.9 actually

tests/unit/test_dto/test_factory/test_dataclass_dto.py
provinzkraut: The test will run on 3.8 only, since it's being skipped for 3.9 and above. It's checking the generic builtins backwards compatibility for 3.8 specifically. 

We should probably to these checks like `> (3, 8)`, so it's a bit more clear where the boundary is supposed to be

tests/unit/test_utils/test_typing.py
provinzkraut: We test these annotations here (`Dict`, `List`, etc.). And while they are not required anymore on > 3.8, they are still available and we want to support them, so it's important that we test for those. Ruff though wants to upgrade them away, so we prevent it from doing so

euri10 (approved):

cofin (approved):
Changes look good aside from the changes mentioned.  I can take another look if there are any significant changes before merging.

litestar/handlers/websocket_handlers/listener.py
provinzkraut: Done

provinzkraut:
> Changes look good aside from the changes mentioned. I can take another look if there are any significant changes before merging.

Just implemented the thing you mentioned, everything else seems good to go :)

PR #4015, opened 2025-02-18
fix: check for default headers before appending to cors header failures

saska:
## Description

- Some clients sometimes apparently (AFAIK needlessly) check OPTIONS even if they only have CORS-safelisted headers to ask about; I ran into this with [Dio](https://pub.dev/packages/dio). I see no reason to allow I guess?

### DIFF
--- a/litestar/middleware/_internal/cors.py
+++ b/litestar/middleware/_internal/cors.py
@@ -50,7 +50,11 @@ class CORSMiddleware(AbstractMiddleware):
             )
             await asgi_response(scope, receive, send)
         elif origin:
-            await self.app(scope, receive, self.send_wrapper(send=send, origin=origin, has_cookie="cookie" in headers))
+            await self.app(
+                scope,
+                receive,
+                self.send_wrapper(send=send, origin=origin, has_cookie="cookie" in headers),
+            )
         else:
             await self.app(scope, receive, send)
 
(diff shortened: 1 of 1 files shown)


saska:
i.e. dio sent this `OPTIONS`
image

litestar/middleware/_internal/cors.py
```diff
@@ -50,7 +50,11 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             )
             await asgi_response(scope, receive, send)
         elif origin:
-            await self.app(scope, receive, self.send_wrapper(send=send, origin=origin, has_cookie="cookie" in headers))
+            await self.app(
+                scope,
+                receive,
+                self.send_wrapper(send=send, origin=origin, has_cookie="cookie" in headers),
+            )
```
provinzkraut, line 57: This change seems unrelated?

provinzkraut (commented):
Looks good, except for the unrelated change. Can you remove that?

litestar/middleware/_internal/cors.py
saska: Oh yeah that's probably just `ruff`, dunno if I had a weird config or something, I'll get rid

provinzkraut: You've added a trailing comma. Ruff then will always put the args on separate lines

saska: Oh, then I think what must have happened was my editor auto formatted with black on save which added the comma and then I ran ruff on top of that :D

litestar/middleware/_internal/cors.py
```diff
@@ -113,7 +113,13 @@ def _create_preflight_response(self, origin: str, request_headers: Headers) -> R
                 response_headers["Access-Control-Allow-Headers"] = ", ".join(
                     sorted(set(pre_flight_requested_headers) | DEFAULT_ALLOWED_CORS_HEADERS)  # pyright: ignore
                 )
-            elif any(header.lower() not in self.config.allow_headers for header in pre_flight_requested_headers):
+            elif any(
+                header.lower()
+                not in set(self.config.allow_headers).union(
+                    default_header.lower() for default_header in DEFAULT_ALLOWED_CORS_HEADERS
```
provinzkraut, line 119: The construction of this set should be moved out of the loop. Otherwise it would be re-created on every iteration. Probably make an `else` branch out of the `elif` so its construction is still conditional.

provinzkraut (approved):

PR #4018, opened 2025-02-19
feat!: security requirements override

### LINKED ISSUE #4016: Bug: routes and routers cannot override OpenAPI security requirement
## Context

Litestar provides a way to declare global security components, for example by the using of a global security configuration like `JWTAuth` or by building your own (Example implementation in `AbstractSecurityConfig.on_app_init`).

This creates both a global security component and requirement in your openapi:

Image

According to the [openapi v3 security documentation](https://swagger.io/docs/specification/v3_0/authentication/#describing-security) you can override at the security keyword at the operation level:

> When used on the root level, security applies the specified security schemes globally to all API operations, **unless overridden on the operation level**.

## The problem

The problem is that you cannot override the security requirement neither at the route or router level with an empty list.

```py
@get(
    "/login",
    # security=[],  # This should override the global security, but is being ignored
)
async def login() -> str:
    return "Sample"
```

Although Litestar provides an API to do so, it does not respect the provided in the `security` because of how it internally handles empty security requirements.

Note that passing a non empty list like `security=[{"SAMPLE": []}]` would actually made it to the openapi security requirements of the route.

**Expected result:**

Image

**Actual:**

Image

## Technical details of the bug and proposed fix

The bug is effectively here:

https://github.com/litestar-org/litestar/blob/6768845e1cbc5d09dc54a8f093f5e554ad187b66/litestar/_openapi/path_item.py#L86

The `or None` replaces any empty resolved security requirement with  None, removing it.

The problem is that simply removing `or None` would make all routes have an overridden security keyword, which is clearly not the desirable effect.

My proposal is to change the default from something which represents as it being not set and handle it properly in the [resolve_security](https://github.com/litestar-org/litestar/blob/6768845e1cbc5d09dc54a8f093f5e554ad187b66/litestar/handlers/http_handlers/base.py#L449-L464) method.

Should I attempt a fix?

_No response_

### MCVE

```python
from typing import Any

from litestar.security.jwt.auth import JWTAuth

from litestar.connection import ASGIConnection


@get(
    "/login",
    security=[],  # This should override the global security, but is being ignored
)
async def login() -> str:
    return "Sample"


@get("/hello-world")
async def hello() -> str:
    return "Hello world"


async def retrieve_user_handler(
    session: dict[str, Any],
    connection: ASGIConnection[Any, Any, Any, Any],
) -> str:
    return "ignored"


jwt_auth = JWTAuth(
    token_secret="secret",
    retrieve_user_handler=retrieve_user_handler,
    exclude=[
        "/login",
        "/schema",
    ],
)


app = Litestar(
    route_handlers=[login, hello],
    on_app_init=[jwt_auth.on_app_init],
)

# How to run this:
# Instal dependencies with pip install litestar[full] uvicorn
# Start the server with `python -m uvicorn app:app --port 8000 --reload``
```

### Steps to reproduce

```bash
1. Run the above code
2. See that the route `/login` has the security in the swagger UI `http://localhost:8000/schema/swagger/` instead of having been overridden by the `security` route argument at line `11`.
```

### Screenshots

```bash
(cannot upload images directly in this field because of improperly configured `BUG.yml`, see Description for screenshots)
```

### Logs

```bash
(not related)
```

### Litestar Version

2.14.0

### Platform

- [x] Linux
- [x] Mac
- [x] Windows

tofran:
Closes #4016
This started as a fix to the issue above, but after thorough consideration this needs a feature which I called `security_override`.

With `security_override` you can override OpenAPI's security requirements passed via `security` in any layer of your application (app, router, route).

Refer to the included docstrings for more information.


provinzkraut:
@tofran There seem to be some test failures. Can you address those?

tofran:
@provinzkraut all ready to review now.

tofran:
@provinzkraut just changed the base and fixed the conflicts. I haven't implemented the proper attribute merging/override and I might need help doing so.
This new attribute merging seems more meta and harder for me to understand (as an external contributor).

Tests should be with 100% coverage and no edge cases left to cover, so we just need to:

- [ ] Implement the two TODO's in `_get_merge_opts`.
- [ ] Finish the documentation changelog

I'm available on discord.

litestar/handlers/http_handlers/base.py
```diff
@@ -372,9 +387,22 @@ def _get_merge_opts(self, others: tuple[Router, ...]) -> dict[str, Any]:
...
             merge_opts["tags"] = (*other.tags, *merge_opts.get("tags", ()))
 
+            current_security = merge_opts.get("security", None)
+
+            # TODO(tofran): properly implement layered security
+            if other.security is not None:
+                if current_security is None:
+                    merge_opts["security"] = other.security
+                else:
+                    merge_opts["security"] = (*current_security, *other.security)
```
provinzkraut, line 399: You don't need this here, as it's already in the base handler `_get_merge_opts`

litestar/handlers/base.py
```diff
@@ -203,6 +203,14 @@ def _get_merge_opts(self, others: tuple[Router, ...]) -> dict[str, Any]:
                 merge_opts.get("signature_namespace", {}), other.signature_namespace
             )
 
+            # TODO(tofran): properly implement layered security
+            current_security = merge_opts.get("security")
+            if other.security is not None:
+                if current_security is None:
+                    merge_opts["security"] = other.security
+                else:
+                    merge_opts["security"] = (*current_security, *other.security)
```
provinzkraut, line 212: This can be simplified like the other merge opts, e.g. `middleware`:
```suggestion
            merge_opts["security"] = (*other.security, *merge_opts.get("security", ()))
```

tofran:
@provinzkraut thank you for your suggestions.

I've made the necessary adjustments to support overriding and still this new merge_opts is not behaving as expected.

The first fail currently is:

```python
assert paths["/router_with_security/route_with_security_override"][method]["security"] == ({"MyRouteToken": []},)
# AssertionError: assert ({'MyRouteToken': []}, {'MyRouterToken': []}) == ({'MyRouteToken': []},)
```

<details><summary>And the odd part is that I created a test case that only has this one route and router, it passes:</summary>

sample POC `tests/unit/test_openapi/test_debug_security_schemes.py`:

```python
from litestar.openapi.spec import Components
from litestar.openapi.spec.security_scheme import SecurityScheme


def test_debug() -> None:
    openapi_config_with_optional_security_scheme: OpenAPIConfig = OpenAPIConfig(
        title="test app",
        version="0.0.1",
        components=Components(
            security_schemes={
                "MyGlobalToken": SecurityScheme(
                    type="http",
                    scheme="bearer",
                ),
                "MyRouterToken": SecurityScheme(
                    type="http",
                    scheme="bearer",
                ),
                "MyRouteToken": SecurityScheme(
                    type="http",
                    scheme="bearer",
                ),
            },
        ),
    )

    @get("/route_with_security_override", security_override=[{"MyRouteToken": []}])
    def route_with_security_override() -> None: ...

    sample_handlers = [
        Router(
            "/router_with_security",
            route_handlers=[route_with_security_override],
        ),
    ]

    app = Litestar(
        openapi_config=openapi_config_with_optional_security_scheme,
        route_handlers=sample_handlers,
    )

    schema = app.openapi_schema

    method = "get"
    schema_dict = schema.to_schema()
    paths = schema_dict["paths"]

    assert paths["/router_with_security/route_with_security_override"][method]["security"] == ({"MyRouteToken": []},)

```
</details>

There's for sure some pollution between multiple routes/routers - either in my code or in this new merge options strategy.

I need your help understanding why and if possible how to fix it.
Thank you in advance.

tofran:
@provinzkraut thank you for your suggestions.

(I deleted the previous comment as I managed to solve the issue)

Everything should be working, apart from the `build-docs` which stalls on my machine (maybe broken on v3 already?).

tofran:
> However, it's also breaking existing behaviour.

Thinking twice, I don't think this (or the previous) implementation breaks existing behaviour. It just adds a new parameter that modifies the behaviour. If you keep using only `security` all the additive functionality is kept the same. Maybe you would like to change that?

cofin:
@tofran It looks like there are many unrelated changes here (perhaps linting related).   Maybe it's just a product of targeting the v3 branch and there's still some work being done.  However, if not, can we limit this PR to specifically the changes required for the security override?

euri10:
> @tofran It looks like there are many unrelated changes here (perhaps linting related). Maybe it's just a product of targeting the v3 branch and there's still some work being done. However, if not, can we limit this PR to specifically the changes required for the security override?

I think it's v3 related and rerere @provinzkraut magic can solve this, this is alien to me

provinzkraut:
> I think it's v3 related and rerere @provinzkraut magic can solve this, this is alien to me

Fixed

tofran:
Thank you @provinzkraut, for some reason v3 keeps getting force pushes making it very hard to have a stable PR.

Also, this PR is ready, I think the actions fail because the upstream is also failing

provinzkraut:
Thanks @tofran this looks good. Could you add the reasoning (in the PR or a comment) for why we specifically need a `security_override`, which diverges from the usual pattern of layered parameters, where the ones set on lower layers override the ones set above those layers?

tofran:
@provinzkraut, thank you. Here's a short summary.

In OpenAPI, overriding global security requirement at the route level is possible according go the specification. While it doesn’t enforce authentication, it documents the security expectations for API consumers using OpenAPI.

Refer to the [openapi v3 security documentation](https://swagger.io/docs/specification/v3_0/authentication/#describing-security):

> When used on the root level, security applies the specified security schemes globally to all API operations, unless overridden on the operation level.

This is useful to specify public endpoints where the API consumer does not need **and should not** provide an authentication parameter, breaking the traditional Litestar layered architecture.


Example `openapi.yaml`:

```yaml
openapi: 3.0.3
info:
  title: Example API with Security Override
  version: 1.0.0

components:
  securitySchemes:
    GobalBearerAuth:
      type: http
      scheme: bearer
    RouteSepcificApiKey:
      type: apiKey
      in: header
      name: X-API-Key

security:
  - GobalBearerAuth: []

paths:
  /login:
    get:
      summary: Public login endpoint
      description: This endpoint is public and does not require authentication. (Override)
      security: []
      responses:
        '200':
          description: Successful login

  /user:
    get:
      summary: User endpoint with security requirement
      description: This endpoint inherits the global security requirement.
      responses:
        '200':
          description: User data returned
        '401':
          description: Unauthorized

  /admin:
    get:
      summary: Admin-only endpoint
      description: This endpoint adds a route specific API key, but wishes to preserve the GobalBearerAuth, meaning it has to be specified too.
      security:
        - GobalBearerAuth: []
        - RouteSepcificApiKey: []
      responses:
        '200':
          description: Admin access granted
        '401':
          description: Unauthorized


```

provinzkraut:
@tofran Maybe I was being imprecise with my question. I'm not asking why we want to be able to override security schemes at the route level. I'm asking why we need `security_overrides` to do so.

Usually, with layered parameters in Litestar, you can override the ones set on a higher level by defining an override on a lower level. What is the reasoning for diverging from this pattern in this case?

In your issue, you've stated that this should work but doesn't:
```python
@get(
    "/login",
    security=[],  # This should override the global security, but is being ignored
)
async def login() -> str:
    return "Sample"
```

Why can't this be made to work and instead needs a new parameter?

tofran:
I see, that's a good question.

Initially I started by just fixing the bug. To do this I thought one could simply remove those empty defaults / if conditions, but in reality **the current implementation is additive**, meaning each layer adds to the previous one instead of replacing it.

As from the tests and implementation this seemed an intended behaviour and I did not want to break existing functionality, therefore I decided the only way was adding a new parameter to override the security requirements.

But after all this we are pointing to v3.0 and effectively can make breaking changes. So if you wish to remove the additive functionality we are still on time.

As an observation: You can see in my example ([previous comment](https://github.com/litestar-org/litestar/pull/4018#issuecomment-2746131371)) the `/admin` route must specify again the `GobalBearerAuth` otherwise it would just have the overridden requirement: `RouteSepcificApiKey`. This means that OpenAPI does not have this layered append by default system.

provinzkraut:
> But after all this we are pointing to v3.0 and effectively can make breaking changes. So if you wish to remove the additive functionality we are still on time.

If we could find a way to make it work without introducing a new parameter, I think that would be best. In other places for example we're allowing both `None` and `EmptyType` to be passed, with the distinction being that `Empty` is the default, so you can explicitly pass `None` to say "this should have a value of `None`".

That would at least solve the "this route should have no security scheme" case. Not sure about the other cases. There's limitations to the current design in this PR as well, e.g. you can only override *all* security schemes, but you can't selectively replace them. 

If that's the goal, maybe a different solution would be appropriate here. We could, for example, introduce an `Override` wrapper for parameters (could literally just be a `(T) -> T` that somehow marks it as an override).

So this would be additive:

```python
@get(
    "/login",
    security=[],
)
async def login() -> str:
    return "Sample"
```
while this would be an override:

```python
@get(
    "/login",
    security=Override([]),
)
async def login() -> str:
    return "Sample"
```

tofran:
> you can only override all security schemes, but you can't selectively replace them.

Yeah I was trying to keep this simple without too much complexity. Do you think that is a requirement? Can't users just pass the same at the route level plus whatever they passed in the previous layer?

> We could, for example, introduce an Override
> we could find a way to make it work without introducing a new parameter

I do really like this signature actually. Maybe `SecurityOverride`, or do you plan to use it elsewhere?


If you want I can try to do this.


My question still persists. Is there a need for an additive behaviour? Was it intentional?

provinzkraut:
> Yeah I was trying to keep this simple without too much complexity. Do you think that is a requirement?

Not necessarily. It's just one more divergence in the parameter design

> I do really like this signature actually. Maybe `SecurityOverride`, or do you plan to use it elsewhere?

If we introduce it, it could work for other layered parameters as well in the future, so I'd say we keep it generic for now.


> On the other hand, my question still persists: Is there a need for an additive behaviour? Was it intentional?

Yes. It's the default behaviour of layered parameters in Litestar that can take a sequence, e.g. `middleware`, `tags`, `guards`, etc.

tofran:
Thank your foe the feedback.

Just pushed a cleanup, removing the `security_override` param, to the branch [tofran:security-requirements-override](https://github.com/litestar-org/litestar/compare/v3.0...tofran:litestar:security-requirements-override?expand=1).

I changed the tests slightly and think they are still valid, they should work after implementing the feature.

TODO:

- [ ] Actually create the `Override` symbol and move it out of `litestar/openapi/spec/security_requirement.py` (updating imports)
- [ ] Implement the layered param merging in `litestar/handlers/base.py:251`

@provinzkraut Can you help me with this to close this PR once and for all?

PR #4022, opened 2025-02-23
ci: Add aggregate test step to allow for dynamic test matrix jobs to be requried checks

provinzkraut:
Add an 'aggregate' step that fails if one of the matrix jobs from the `test` job fail, so we can target this aggregate job as "required check" instead of the individual matrix jobs. 

This allows us to e.g. have a `3.8` matrix job on one branch but not on the other, while still making the run fail if any of the matrix jobs fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,19 @@ jobs:
       coverage: ${{ (matrix.python-version == '3.12' || matrix.python-version == '3.8') }}
       python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
There's a special github action for this: https://github.com/re-actors/alls-green

We use it in CPython: https://github.com/python/cpython/blob/25a7ddf2efeaf77bcf94dbfca28ba3a6fe9ab57e/.github/workflows/build.yml#L607-L676

cofin (approved):

PR #4025, opened 2025-02-25
feat: Add `SerializationPlugin` and `InitPlugin` to replace their respective protocols 

provinzkraut:
- Add `SerializationPlugin` to replace `SerializationPluginProcotol`
- Add `InitPlugin` to replace `InitPluginProtocol`

These are the only ones for which the protocols don't have a non-protocol equivalent. 

Following the same approach as for other plugins, they inherit their respective protocol for now, to keep type / `isinstance` checks compatible.


provinzkraut:
@cofin Once merged, we can migrate AA to these

PR #4034, opened 2025-02-27
feat!: Remove deprecated `body` param from streaming and file responses

provinzkraut:
Remove the deprecated `body` param of `ASGIStreamingResponse` and `ASGIFileResopnse`. The parameter previous did nothing.

### DIFF
--- a/litestar/response/file.py
+++ b/litestar/response/file.py
@@ -124,7 +124,6 @@ class ASGIFileResponse(ASGIStreamingResponse):
         self,
         *,
         background: BackgroundTask | BackgroundTasks | None = None,
-        body: bytes | str = b"",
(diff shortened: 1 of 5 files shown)


litestar/response/file.py
```diff
@@ -370,7 +367,6 @@ def to_asgi_response(
         """Create an :class:`ASGIFileResponse <litestar.response.file.ASGIFileResponse>` instance.
 
         Args:
-            app: The :class:`Litestar <.app.Litestar>` application instance.
```
euri10, line 373: is this related ?

provinzkraut: No, I just noticed this was left in there. The param doesn't exist anymore 🤷

euri10 (approved):

PR #4042, opened 2025-03-03
docs: Use the article "an" for words beginning with SQL

eandersons:
By submitting this pull request, I agree to:

- [x] follow Litestar's [Code of Conduct](https://github.com/litestar-org/.github/blob/main/CODE_OF_CONDUCT.md);
- [x] follow Litestar's [contribution guidelines](https://github.com/litestar-org/.github/blob/main/CONTRIBUTING.md);
- [x] follow the PSFs's [Code of Conduct](https://www.python.org/psf/conduct/).

## Description

Article "a" was replaced with "an" for words beginning with "SQL" as the pronunciation of SQL starts with a vowel sound: /ˌɛsˌkjuˈɛl/ (S-Q-L). Previously both "a SQL*" and "an SQL*" were used.
That is, if project authors do not prefer the alternative pronunciation /ˈsiːkwəl/ ("sequel"), in which case the existing occurrences of "an SQL*" should be replaced with "a SQL*".

For reference:

- <https://www.scribbr.com/commonly-confused-words/a-vs-an/#acronyms>
- <https://en.wikipedia.org/wiki/SQL>

## Closes

Not applicable.

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -3638,7 +3638,7 @@
         Fix the behaviour of DTOs such that they will no longer access fields that have
         been included. This behaviour would previously cause issues when these
         attributes were either costly or impossible to access (e.g. lazy loaded
-        relationships of a SQLAlchemy model).
+        relationships of an SQLAlchemy model).
 
     .. change:: DTO | Regression: ``DTOData.create_instance`` ignores renaming
         :type: bugfix
@@ -4176,7 +4176,7 @@
         :type: feature
         :pr: 1852
 
-        Replace the nondescript :exc:`KeyError` raised when a SQLAlchemy DTO is
+        Replace the nondescript :exc:`KeyError` raised when an SQLAlchemy DTO is
         constructed from a model that is missing a type annotation for an included
         column with an :exc:`ImproperlyConfiguredException`, including an explicit error
         message, pointing at the potential cause.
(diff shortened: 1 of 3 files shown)


eandersons:
> I'd rather debate on vim/emacs but well you mentionned [/ˈsiːkwəl/](https://en.wikipedia.org/wiki/Help:IPA/English) "sequel" so I'd rather see it the other way around, but no hard feelings on this, I'll let maintainers maintain which pronouciation they think is appropriate :)

I am not trying to start a flame war. I noticed that both "a" and "an" were used with "SQL*" and I subjectively objectively chose the right one.
This pull request is not about which pronunciation of "SQL" is the correct one, but to point to inconsistencies with article used with "SQL*". In the end it is up to project authors/maintainers to decide which article should be used with "SQL*" in project's documentation.

euri10:
> > I'd rather debate on vim/emacs but well you mentionned [/ˈsiːkwəl/](https://en.wikipedia.org/wiki/Help:IPA/English) "sequel" so I'd rather see it the other way around, but no hard feelings on this, I'll let maintainers maintain which pronouciation they think is appropriate :)
> 
> I am not trying to start a flame war. I noticed that both "a" and "an" were used with "SQL*" and I subjectively objectively chose the right one. This pull request is not about which pronunciation of "SQL" is the correct one, but to point to inconsistencies with article used with "SQL*". In the end it is up to project authors/maintainers to decide which article should be used with "SQL*" in project's documentation.

I was joking @eandersons ! And I agree with you with the consistency part while I disagree on the pronouciation ;)

cofin:
I would prefer to use the `a` article for SQL.  I generally pronounce the SQLAlchemy library similarly to "sequel alchemy".

In fact, I almost never treat "SQL" as an initialism.  It's always used as an acronym.

I definitely agree that it should be consistent though.

Issue #4044, opened 2025-03-05
Bug: internal server error when raising exception from listeners sync callable


Harshal6927:
### Description

When passing sync functions to listeners which can raise an error put the Litestar into unrecoverable state. (https://discord.com/channels/919193495116337154/1342406039013949440)

This is happening because of this line https://github.com/litestar-org/litestar/blob/main/litestar/events/listener.py#L68, specifically the `fn.__name__`

When wrapping `sync` function to make it `async` with the help of `AsyncCallable` class, it loses the metadata of the original function (it uses `partial` from `functools`) so the new wrapped `async` callable does not have `__name__` along with other metadata.

### MCVE

```python
from litestar import Litestar, Request, get
from litestar.events import listener


@listener("raise_exception")
def raise_exception_if_odd(value) -> None:
    if value is not None and value % 2 != 0:
        raise ValueError(f"{value} is odd")


@get("/check-value/{value:int}")
async def check_value(request: Request, value: int) -> None:
    request.app.emit("raise_exception", value)


app = Litestar([check_value], listeners=[raise_exception_if_odd])


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app=app, port=9283)
```

### Logs

```text
INFO:     Started server process [66070]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:9283 (Press CTRL+C to quit)
INFO:     127.0.0.1:45100 - "GET /check-value/11 HTTP/1.1" 200 OK
ERROR:    Traceback (most recent call last):
  File "/home/shadowmonarch/harshal/litestar/litestar/app.py", line 660, in lifespan
    yield
  File "/home/shadowmonarch/harshal/litestar/litestar/_asgi/asgi_router.py", line 184, in lifespan
    await receive()
  File "/home/shadowmonarch/harshal/litestar/.venv/lib/python3.12/site-packages/uvicorn/lifespan/on.py", line 137, in receive
    return await self.receive_queue.get()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/queues.py", line 158, in get
    await getter
asyncio.exceptions.CancelledError: Cancelled by cancel scope 7f74ae3a7410

During handling of the above exception, another exception occurred:

  + Exception Group Traceback (most recent call last):
[...]

PR #4045, opened 2025-03-05
fix(events): ensure callable metadata is replicated

Harshal6927:
## Description

- When passing sync functions to listeners which can raise an error put the Litestar into unrecoverable state. (https://discord.com/channels/919193495116337154/1342406039013949440)

## Closes
https://github.com/litestar-org/litestar/issues/4044


cofin (approved):
LGTM

PR #4051, opened 2025-03-07
feat: adds a `granian` optional installation group

cofin:
Implements a `granian` installation group similar to the `uvicorn` or `standard` group.  This group auto-installs `litestar-granian` and necessary performance enhancements.

Additionally, the latest linting rules were applied after upgrading the project depedencies


pyproject.toml
```diff
@@ -88,6 +88,7 @@ full = [
   "litestar[annotated-types,attrs,brotli,cli,cryptography,jinja,jwt,mako,minijinja,opentelemetry,piccolo,picologging,prometheus,pydantic,redis,sqlalchemy,standard,structlog,valkey]; python_version < \"3.13\"",
   "litestar[annotated-types,attrs,brotli,cli,cryptography,jinja,jwt,mako,minijinja,opentelemetry,piccolo,prometheus,pydantic,redis,sqlalchemy,standard,structlog,valkey]; python_version >= \"3.13\"",
 ]
+granian = ["litestar-granian[uvloop]", "python-dotenv", "httptools", "websockets"]
```
Kumzy, line 91: Why add httptools there ? (already satisfied in litestar-granian)
What's the use of websockets and python-dotenv also there?
Finally why uvloop as extra dependencies of litestar-granian? as uvloop/rloop are optional extras of granian

cofin: Currently, if you install `litestar[cli]` or `litestar[standard]` you get uvicorn installed.  This allows you to get an exact equivalent of `litestar[cli]/litestar[standard]`, but with granian instaed of uvicorn.

I'm not sure I'm following the question re: the `uvloop` question, but the intent of all of this is to give you the optimal/"most performant" default set of packages without installing any extras that come with the other packages.

cofin: It does looks like `httptools` can be removed though.  Same for `websockets` if it's also in the `litestar-granian` dependency list.

i'll take a look at that tomorrow (feel free to open a PR to `litestar-granian` or modify this one as well)

Kumzy: Just removed the ``httptools`` as already in litestar-granian.

Kumzy:
@cofin do we still want to do that or let `litestar-granian` as a plugin?

PR #4053, opened 2025-03-11
Add Dockerfile for Litestar

ImMedved:
## Description

-


sobolevn (commented):
Sorry, it is not clear for me what this dockerfile does and why it is needed. Can you please clarify?

JacobCoffee:
Also we have a Dockerfile for the dev containers in .devcontainer if that's what you are looking to add

PR #4058, opened 2025-03-18
refactor: improve import formatting and type hinting consistency across modules

cofin:
This PR also updates the dependency and tests against the latest versions of ruff and pyright.

There are also a few documentations related updates for Advanced Alchemy 1.0


sobolevn (approved):
Good call, thank you!

pyproject.toml
```diff
@@ -103,7 +103,7 @@ pydantic = [
   "pydantic-extra-types; python_version >= \"3.9\"",
 ]
 redis = ["redis[hiredis]>=4.4.4"]
-sqlalchemy = ["advanced-alchemy>=0.2.2"]
+sqlalchemy = ["advanced-alchemy"]
```
provinzkraut, line 106: Why drop the lower bound?

litestar/_openapi/schema_generation/constrained_fields.py
```diff
@@ -80,7 +80,7 @@ def create_string_constrained_field_schema(
         schema.max_length = kwarg_definition.max_length
     if kwarg_definition.pattern:
         schema.pattern = (
-            kwarg_definition.pattern.pattern  # type: ignore[attr-defined]
+            kwarg_definition.pattern.pattern  # type: ignore[attr-defined,unused-ignore]
```
provinzkraut, line 83:
```suggestion
            kwarg_definition.pattern.pattern
```

litestar/plugins/pydantic/plugins/di.py
```diff
@@ -22,5 +22,5 @@ def get_typed_init(self, type_: Any) -> tuple[Signature, dict[str, Any]]:
             inspect.Parameter(name=field_name, kind=inspect.Parameter.KEYWORD_ONLY, annotation=Any)
             for field_name in model_fields
         ]
-        type_hints = {field_name: Any for field_name in model_fields}
+        type_hints = dict.fromkeys(model_fields, Any)
```
provinzkraut, line 25: I personally find the other method more readable as it's more explicit

litestar/contrib/minijinja.py
```diff
@@ -207,7 +207,7 @@ def from_environment(cls, minijinja_environment: Environment) -> MiniJinjaTempla
 @pass_state
 def _minijinja_from_state(func: Callable, state: StateProtocol, *args: Any, **kwargs: Any) -> str:  # pragma: no cover
     template_context = {"request": state.lookup("request"), "csrf_input": state.lookup("csrf_input")}
-    return cast(str, func(template_context, *args, **kwargs))
+    return cast("str", func(template_context, *args, **kwargs))
```
provinzkraut, line 210: Did you do all of these by hand or is it a ruff rule? I couldn't find any change in the rules that would do this.

IMO unless we have a rule that enforces this, we shouldn't be changing it

provinzkraut (changes requested):

litestar/contrib/minijinja.py
adhtruong: Looks like https://docs.astral.sh/ruff/rules/runtime-cast-value/ is the rule and this was stabilised in https://github.com/astral-sh/ruff/releases/tag/0.10.0

provinzkraut: Ah! Good to know!

litestar/plugins/pydantic/plugins/di.py
provinzkraut: Can we ignore it?

litestar/_openapi/schema_generation/constrained_fields.py
cofin: For some reason this is raising linter errors locally but not in CI.  I'm wondering if it's a difference between pylance/pyright?

cofin:
Closing — this work was already merged to main via commit b42aeb916 (2025-03-18). The import formatting, type hinting consistency improvements, and advanced-alchemy 1.x update are all on main.

PR #4063, opened 2025-03-21
feat: v3 - Move `litestar-htmx` to a package extra

provinzkraut:
Move `litestar-htmx` from a default dependency to the `litestar[htmx]` extra.

PR #4065, opened 2025-03-22
docs(v3): fix build error

provinzkraut:
Finally found the source of the mysterious

```
docstring of builtins.dict:3: ERROR: Unexpected indentation.
docstring of builtins.dict:4: WARNING: Block quote ends without a blank line; unexpected unindent.
docstring of builtins.dict:7: ERROR: Unexpected indentation.
docstring of builtins.dict:9: WARNING: Inline strong start-string without end-string.
```

Still no idea *why* that suddenly is an issue but :woman_shrugging:

### DIFF
--- a/litestar/openapi/spec/security_requirement.py
+++ b/litestar/openapi/spec/security_requirement.py
@@ -1,4 +1,4 @@
-from typing import TypeAlias
+from typing_extensions import TypeAlias
 
 SecurityRequirement: TypeAlias = "dict[str, list[str]]"
 """Lists the required security schemes to execute this operation. The name used for each property MUST correspond to a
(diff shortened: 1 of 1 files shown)


euri10 (approved):
bravo, 
how did you manage to find that is interesting,

provinzkraut:
> bravo, how did you manage to find that is interesting,

You set me on the right path with [this comment of yours](https://github.com/litestar-org/litestar/pull/4063#issuecomment-2745116877). I basically did a `git bisect` on the docs building. Started deleting 50% of them and see if that eliminated the error, if it did, restore 50% of that, see if it pops up again, etc. Only took me about 10 rounds or so to narrow it down to a single file, and then I could easily find a commit that matched :)

euri10:
worth a report upstream the error is cryptic, kudos on the method !

PR #4070, opened 2025-03-25
Remove excluded pydantic field  from open api response

StickKing:
## Description

If we use a Pedantic model to type the response of a request, the fields marked as 'exclude' are displayed in the Open API schema, even though they won't be present after serialization.

Example:
```python
from pprint import pprint


class Model(BaseModel):
    """Model."""

    name: str
    age: int
    exclude_field: str = Field(exclude=True)


@get("/")
def test() -> Model:
    return Model(name="David", age=25, exclude_field="man")


pprint(app.openapi_schema.to_schema())
```

Result:
```
{'components': {'schemas': {'Model': {'properties': {'age': {'type': 'integer'},
                                                     'exclude_field': {'type': 'string'},
                                                     'name': {'type': 'string'}},
                                      'required': ['age',
                                                   'exclude_field',
                                                   'name'],
                                      'title': 'Model',
                                      'type': 'object'}}},
 'info': {'title': 'Litestar API', 'version': '1.0.0'},
 'openapi': '3.1.0',
 'paths': {'/': {'get': {'deprecated': False,
                         'operationId': 'Test',
                         'responses': {'200': {'content': {'application/json': {'schema': {'$ref': '#/components/schemas/Model'}}},
                                               'description': 'Request '
                                                              'fulfilled, '
                                                              'document '
                                                              'follows',
                                               'headers': {}}},
                         'summary': 'Test'}}},
 'servers': [{'url': '/'}]}
```

To prevent this, I created a attribute called `schema_context` on the `SchemaCreator` object, in which I pass the request context to determine if the current schema is a response; if so, we filter out the fields by the `exclude` attribute.

### DIFF
--- a/litestar/_openapi/datastructures.py
+++ b/litestar/_openapi/datastructures.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import re
 from collections import defaultdict
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Iterator, Sequence, _GenericAlias  # type: ignore[attr-defined]
 
 from litestar.exceptions import ImproperlyConfiguredException
@@ -246,3 +247,10 @@ class OpenAPIContext:
                 f"please ensure the value of 'operation_id' is either not set or unique for {operation_id}"
             )
         self.operation_ids.add(operation_id)
+
+
+@dataclass
+class SchemaContext:
+    """Context data for schema creator."""
+
+    is_response: bool = field(default=False)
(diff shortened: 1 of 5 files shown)


provinzkraut:
Why do we need the `SchemaContext` object for this?
Iiuc, the `exclude` isn't context sensitive. It will always be excluded by default for the model the field was define on.

StickKing:
> Why do we need the `SchemaContext` object for this? Iiuc, the `exclude` isn't context sensitive. It will always be excluded by default for the model the field was define on.

@provinzkraut 
So that the Pydantic model used as data in a POST request does not exclude the field from OpenAPI.

Issue #3606, opened 2024-07-01
Bug: Duplication in request URL path when configuring OpenAPI Servers


mohammedbabelly20:
### Description

All requests URLs in documentation have duplicate paths when configuring the Litestar app with a custom path and configure the OpenAPI with servers.
So the final request URL has the full server url from OpenAPIConfig and the path added from the Litestar app.
`http://localhost:8000/api/v3/api/v3/...` instead of `http://localhost:8000/api/v3/...`

### MCVE

```python
return Litestar(path="api/v3", ...)


```py
    return OpenAPIConfig(
        path="/schema",
        servers=[Server(url="http://localhost:8000/api/v3")],
        render_plugins=[StoplightRenderPlugin(version="latest")],
    )
```
```


### Steps to reproduce

```bash
1. Create an OpenAPIConfig that has at least one server e.g. `http://localhost:8000/api/v3`
2. Create a new Litestar app with `api/v3` path and use the OpenAPIConfig from above
3. Run the application and access the schema
4. See that all requests in schema has full server URL from OpenAPI and also the app path which is `api/path` so the final URL is incorrect `http://localhost:8000/api/v3/api/v3/...`
```

### Litestar Version

`2.9.1`

### Platform

- [X] Linux
- [X] Mac
- [X] Windows
- [ ] Other (Please specify in the description above)

PR #4072, opened 2025-03-26
fix: duplication in request URL path when configuring OpenAPI #3306

StickKing:
## Description

I would like to propose a solution for issue [#3306.](https://github.com/litestar-org/litestar/issues/3606) I have created two methods in `OpenAPIPlugin`:  
1. `_normalize_route_path` is designed to remove redundant parts from the URL.  Here, I assume that the URLs specified in servers have higher priority when building the Open API schema.
2. `_normalize_default_server` is needed in cases where the user sets `app.path` but does not define `servers` in `OpenApiConfig`

### DIFF
--- a/litestar/_openapi/plugin.py
+++ b/litestar/_openapi/plugin.py
@@ -86,7 +86,7 @@ class OpenAPIPlugin(InitPlugin, ReceiveRoutePlugin):
         """
         path = route.path_format or "/"
         if self.app.path != "/":
-            return path.replace(self.app.path, "") or "/"
+            return path[len(self.app.path) :] or "/"
         return path
 
     def _build_openapi(self) -> OpenAPI:
--- a/litestar/_openapi/plugin.py
+++ b/litestar/_openapi/plugin.py
@@ -67,6 +67,28 @@ class OpenAPIPlugin(InitPlugin, ReceiveRoutePlugin):
         self._openapi: OpenAPI | None = None
         self._openapi_schema: dict[str, object] | None = None
 
+    def _normalize_default_server(self) -> None:
+        """If the default server is used along with app.path, it changes the server URL."""
+        if self.app.path == "/":
+            return
+        for server in self.openapi_config.servers:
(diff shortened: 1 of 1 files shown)


litestar/_openapi/plugin.py
```diff
@@ -67,6 +67,28 @@ def __init__(self, app: Litestar) -> None:
...
+
+        Args:
+            route: http route object
+
+        Returns:
+            path: normalized path
+        """
+        path = route.path_format or "/"
+        if self.app.path != "/":
+            return path.replace(self.app.path, "") or "/"
```
provinzkraut, line 89: This can easily lead to bugs, if `self.app.path` is contained somewhere within `path`. *If* we are to do this, it should use `.removeprefix`.

StickKing: Maybe use `return path[len(self.app.path):] or "/"` to support Python 3.8. ?

litestar/_openapi/plugin.py
```diff
@@ -67,6 +67,28 @@ def __init__(self, app: Litestar) -> None:
         self._openapi: OpenAPI | None = None
         self._openapi_schema: dict[str, object] | None = None
 
+    def _normalize_default_server(self) -> None:
```
provinzkraut, line 70: If this alters something in the instance state, it should happen in `__post_init__`.

provinzkraut (changes requested):

litestar/_openapi/plugin.py
StickKing: @provinzkraut This instance does not have access to the object from which the information is obtained (`app.path`), therefore `__post_init__` cannot be used.

It may be better to define this method on the Server class and pass `app.path` to it.

provinzkraut: > It may be better to define this method on the Server class and pass app.path to it.

Sounds like a reasonable approach, yes

PR #4078, opened 2025-03-29
refactor: use typos

adhtruong:
## Description

- Migrate from codespell to typos

### DIFF
--- a/litestar/cli/commands/core.py
+++ b/litestar/cli/commands/core.py
@@ -108,7 +108,7 @@ def _run_uvicorn_in_subprocess(
     )
 
 
-class CommaSplittedPath(click.Path):
+class CommaSplitPath(click.Path):
     """A Click Path that splits the input string by commas.
 
     .. versionadded:: 2.8.0
@@ -139,7 +139,7 @@ def info_command(app: Litestar) -> None:
     "-R",
     "--reload-dir",
     help="Directories to watch for file changes",
-    type=CommaSplittedPath(),
+    type=CommaSplitPath(),
     multiple=True,
     envvar="LITESTAR_RELOAD_DIRS",
 )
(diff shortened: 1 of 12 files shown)


litestar/cli/commands/core.py
```diff
@@ -108,7 +108,7 @@ def _run_uvicorn_in_subprocess(
     )
 
 
-class CommaSplittedPath(click.Path):
+class CommaSplitPath(click.Path):
```
provinzkraut, line 111: Can't rename this in v2 since it's a public class. We could add an alias and deprecate it, or do this move in v3

typos.toml
```diff
@@ -0,0 +1,12 @@
+[default]
+extend-ignore-re = ["(?Rm)^.*(#|//)\\s*codespell:ignore\\s*$"]
```
adhtruong, line 2: Note this is arbitrary the way it's defined. Matching previous codespell tool

provinzkraut (changes requested):

typos.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[default.extend-words]
+selectin = 'selectin'
+odf = 'odf'
```
adhtruong, line 6: Required for https://github.com/adhtruong/litestar/blob/refactor-use-typos/tests/examples/test_request_data.py#L77. Feels like an identifier so adding here rather than inlining

litestar/cli/commands/core.py
adhtruong: Aiming at v3 currently but should I add a deprecation too?

adhtruong: Would it be better to aim at `main` instead?

adhtruong:
Closing in favour of targeting main here https://github.com/litestar-org/litestar/pull/4080

PR #4085, opened 2025-04-03
fix: normalize openapi server url

StickKing:
## Description
If we specify a server URL in `OpenAPIConfig` without a leading `/`, the schema generates incorrect paths.
Example:
```python
from litestar.openapi.spec import Server

@get("/")
async def test() -> dict:
    return {}

@get("/test")
async def test1() -> dict:
    return {}

app = Litestar(
    route_handlers=[test, test1],
    openapi_config=OpenAPIConfig(
        title="Litestar",
        version="1.0.0",
        servers=[Server(url="api/v3")],
    )
)
```
image

As a solution, I decided to use `from litestar.utils import normalize_path` to normalize the server URLs.

### DIFF
--- a/litestar/openapi/spec/server.py
+++ b/litestar/openapi/spec/server.py
@@ -36,4 +36,5 @@ class Server(BaseSchemaObject):
 
     def __post_init__(self) -> None:
         """Normalize the url."""
-        self.url = normalize_path(self.url)
+        if self.url.startswith("http") is False:
+            self.url = normalize_path(self.url)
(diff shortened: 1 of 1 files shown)


provinzkraut:
This is not a bug, it's simply how URLs join/resolve. If you join an absolute to a relative URL, the latter will take precedence:

```python
urllib.parse.urljoin("api/v3", "/test")  # '/test'
```

You should just declare your path as `/api/v3` instead.

PR #4089, opened 2025-04-06
fix(typing): Correctly handle `typing_extensions.TypeAliasType` on `typing-extensions>4.13.0` (fix #4088)

### LINKED ISSUE #4088: Bug: tests/unit/test_typing.py::test_unwrap_type_alias_type_keyword fails with typing-extensions 4.13.0
### Description

Within a recent rebuild of package in Debian the litestar test suite is failing on the test `tests/unit/test_typing.py::test_unwrap_type_alias_type_keyword` since typing-extensions 4.13.0 did reach the Debian archive. With typing-extensions << 4.13.0 the tests did succeed.

The relevant output of the failing test is
```
=================================== FAILURES ===================================
_____________________ test_unwrap_type_alias_type_keyword ______________________

    @pytest.mark.skipif(sys.version_info < (3, 12), reason="type keyword not available before 3.12")
    def test_unwrap_type_alias_type_keyword() -> None:
        ctx: dict[str, Any] = {}
        exec("type IntAlias = int", ctx, None)
        annotation = ctx["IntAlias"]
        field_definition = FieldDefinition.from_annotation(annotation)
>       assert field_definition.is_type_alias_type
E       AssertionError: assert False
E        +  where False = FieldDefinition(raw=IntAlias, annotation=IntAlias, type_wrappers=set(), origin=None, args=(), metadata=(), instantiabl...one, safe_generic_origin=None, inner_types=(), default=<_EmptyEnum.EMPTY: 0>, extra={}, kwarg_definition=None, name='').is_type_alias_type
```
The full log is visible on https://bugs.debian.org/1101855

polyfactory did had a similar issue that got fixed by https://github.com/litestar-org/polyfactory/pull/684

I've seen there is also https://github.com/litestar-org/litestar/pull/3982 but I don't know enough if this PR is targeted to fix the issue from above.

_No response_

### MCVE

```python

```

### Steps to reproduce

Should be visible if the tests are run with typing-extensions 4.13.0 is around. I haven't run any tests within a venv, only within chroot while a rebuild of the Debian package.

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.15.1

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
- Fix #4088 by handling the diverging `TypeAliasType` introduced in typing-extensions `4.13.0`; This type is no longer backwards compatible, as it is a distinct new type from `typing.TypeAliasType`. We need to check for both types now
- Add a new test job to the CI to test against known diverging versions of `typing-extensions`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,30 @@ jobs:
       - name: Report success or fail
         run: exit ${{ needs.test.result == 'success' && '0' || '1' }}
 
+  test_typing_extensions:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        typing-extensions: ["4.12.1", "4.13.1"]
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up pytho
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.13
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "0.6.12"
+          enable-cache: true
+
+      - name: Test
+        run: uv run --with="typing-extensions==${{ matrix.typing-extensions }}" -- python -m pytest tests/unit/test_typing.py
+
+
   test_integration:
     name: Test server integration
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,7 +124,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v4
 
-      - name: Set up pytho
+      - name: Set up python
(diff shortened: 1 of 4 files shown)


sobolevn (approved):
Thanks!

optional: I would also test `type` keyword, it also creates a `TypeAliasType` object, but just to be sure.

provinzkraut:
> Thanks!
> 
> optional: I would also test `type` keyword, it also creates a `TypeAliasType` object, but just to be sure.

We have a separate test case for that! :)

https://github.com/litestar-org/litestar/blob/bf86f7ebf47068139fa82789920579bceffeecea/tests/unit/test_typing.py#L491-L497

PR #4090, opened 2025-04-06
ci: Test against typing-extension pre-releases

provinzkraut:
To ensure compatibility, run against typing-extension pre-releases as an early warning

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,11 +139,11 @@ jobs:
         run: uv sync
 
       - name: Install typing-extensions
-        if: typing-extensions != 'latest'
-        run: uv pip install typing-extensions==${{ matrix.typing-extensions }}"
+        if: ${{ matrix.typing-extensions != 'latest' }}
+        run: uv pip install typing-extensions=="${{ matrix.typing-extensions }}"
 
       - name: Install typing-extensions
-        if: typing-extensions == 'latest'
+        if: ${{ matrix.typing-extensions == 'latest' }}
         run: uv pip install typing-extensions --upgrade --prerelease=allow
 
       - name: Test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        typing-extensions: ["4.12.1", "4.13.1"]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -135,8 +135,19 @@ jobs:
...
+        run: uv sync
+
+      - name: Install typing-extensions
+        if: typing-extensions != 'latest'
+        run: uv pip install typing-extensions==${{ matrix.typing-extensions }}"
+
+      - name: Install typing-extensions
+        if: typing-extensions == 'latest'
+        run: uv pip install typing-extensions --upgrade --prerelease=allow
+
```
sobolevn, line 148:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -135,8 +135,19 @@ jobs:
...
       - name: Test
-        run: uv run --with="typing-extensions==${{ matrix.typing-extensions }}" -- python -m pytest tests/unit/test_typing.py
+        run: uv run --no-sync -- python -m pytest tests/unit/test_typing.py
```
sobolevn, line 150:
```suggestion
        run: |
          if [[ "${{ matrix.typing-extensions }}" == 'latest' ]]; then
            uv run --with=typing-extensions -- python -m pytest tests/unit/test_typing.py
          else
             uv run --with="typing-extensions==${{ matrix.typing-extensions }}" -- python -m pytest tests/unit/test_typing.py
          fi
```

sobolevn (commented):
How about a `bash`-level `if`?

.github/workflows/ci.yml
provinzkraut: Mh. Not sure if I find this more readable?

sobolevn: Both are fine :)

sobolevn: It can even be simplier with `bash`

provinzkraut: That won't work tho as ` uv run --with="typing-extensions=="` isn't valid. You need the different cmds in this case; Unfortunately there's no way (that I know of) to use `--with` to install and upgrade to the latest version

sobolevn: hm, I solved `==` issue, but I am not sure about the "update" part :)
anyways, let's not spend more time on this. if your version works - let's use it :_

cofin (approved):

Issue #2778, opened 2023-11-27
Docs: `server_lifespan` usage


provinzkraut:
### Summary

The `server_lifespan` hook introduced in #2658 is missing usage docs

euri10:
Closed by #4097

PR #4097, opened 2025-04-09
docs: add usage of server_lifespan hook and motivations

euri10:
## Description

-

## Closes

https://github.com/litestar-org/litestar/issues/2778

### DIFF
--- a/docs/usage/cli.rst
+++ b/docs/usage/cli.rst
@@ -144,9 +144,9 @@ You can achieve this by adding the special ``app`` parameter to your CLI functio
 Using the `server_lifespan` hook
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Server lifespan hooks are a powerful way to run code before and after the server starts and stops.
-It is useful when running the underlying asgi server with multiple workers, as it allows you to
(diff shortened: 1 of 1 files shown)


docs/usage/cli.rst
```diff
@@ -141,6 +141,38 @@ You can achieve this by adding the special ``app`` parameter to your CLI functio
     @click.command()
     def my_command(app: Litestar) -> None: ...
 
+Using the `server_lifespan` hook
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Server lifespan hooks are a powerful way to run code before and after the server starts and stops.
+It is useful when running the underlying asgi server with multiple workers, as it allows you to
+run code before the server starts and after it stops, regardless of the number of workers.
```
provinzkraut, line 149:
```suggestion
Server lifespan hooks provide a way to run code before and after the *server* starts and stops. In contrast to the regular `lifespan` hooks, they only run once, even when a server starts multiple workers, whereas `lifespan` hooks would run for each individual worker.

This makes them suitable for tasks that should happen exactly once, like initializing a database.
```

provinzkraut (approved):

PR #4098, opened 2025-04-09
feat: sse timeout events

### LINKED ISSUE #4082: Enhancement: recurring ping in sse responce
### Summary

problem: response failed due to timeout if no events come for a long time;
proposal: add ping message every x seconds in sse responce.

### Basic Example

like in [sse-starlette](https://github.com/sysid/sse-starlette/blob/main/sse_starlette/sse.py#L122)

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

RenameMe1:
## Description

I would like to propose a solution for issue #4082. 

- I'm added optional `timeout` keyword in [ServerSentEvent](https://github.com/litestar-org/litestar/commit/6f46fe29bd00c9621c94ff8cd7d006749329e7d6), [ASGIStreamingResponse and Stream](https://github.com/litestar-org/litestar/commit/377d93f8bd389e0003ef30377d7632ed2e51f097)

-  A way to send a timeout signal has also been [added](https://github.com/litestar-org/litestar/commit/a2bd3a63abbb59ffdfeebf437debac6a271d8deb).

## Closes

#4082


RenameMe1:
@sobolevn can you help me please with litestar/response/streaming.py:154: error:  in mypy. 

I don't follow why this error occur. 

- AsyncIteratorWrapper has __anext__
    https://github.com/litestar-org/litestar/blob/027ce12a64711f90c9d13b79d753be81779e6db0/litestar/utils/sync.py#L48
    
- all content wraps by _ServerSentEventIterator(AsyncIteratorWrapper[bytes]) in ServerSentEvent
https://github.com/litestar-org/litestar/blob/027ce12a64711f90c9d13b79d753be81779e6db0/litestar/response/sse.py#L130C7-L130C22

- ...
https://github.com/litestar-org/litestar/blob/027ce12a64711f90c9d13b79d753be81779e6db0/litestar/response/streaming.py#L90

provinzkraut:
This should not be named timeout as it is not a timeout.

litestar/response/streaming.py
```diff
@@ -119,15 +127,56 @@ async def _stream(self, send: Send) -> None:
...
+
+        Args:
+            send: The ASGI Send function.
+
+        Returns:
+            None
+        """
+        while True:
+            start_time = time.perf_counter()
+            chunk = self.iterator.__anext__()
```
provinzkraut, line 154: You shouldn't call dunder methods directly like this. Instead you should use `anext` here

litestar/response/streaming.py
```diff
@@ -119,15 +127,56 @@ async def _stream(self, send: Send) -> None:
...
+            task = asyncio.create_task(chunk)
```
provinzkraut, line 155: You cannot assume asyncio. Use an anyio task group instea

provinzkraut (changes requested):
I would suggest a different design. Take a look at how the stream handles client disconnects. Ideally, the two background tasks can share the same task group. Otherwise, there's unnecessary overhead involved, and cancellations become quite complex

litestar/response/sse.py
```diff
@@ -141,6 +141,7 @@ def __init__(
         retry_duration: int | None = None,
         comment_message: str | None = None,
         status_code: int | None = None,
+        ping_interval: int | None = None,
```
euri10, line 144: let's make it a float

litestar/response/streaming.py
```diff
@@ -64,6 +68,7 @@ def __init__(
             iterator: An async iterator or iterable.
             media_type: The response media type.
             status_code: The response status code.
+            ping_interval: Interval for sending message with "ping" body.
```
euri10, line 71:
```suggestion
            ping_interval: The interval in seconds between "ping" messages.
```

litestar/response/streaming.py
```diff
@@ -110,6 +116,25 @@ async def _listen_for_disconnect(self, cancel_scope: CancelScope, receive: Recei
             else:
                 await self._listen_for_disconnect(cancel_scope=cancel_scope, receive=receive)
 
+    async def _send_ping_event(self, send: Send) -> None:
+        """Send ping events via ping_interval time.
```
euri10, line 120:
```suggestion
        """Send ping events every `ping_interval` second.
```

litestar/response/sse.py
```diff
@@ -159,6 +160,7 @@ def __init__(
             event_id: The event ID. This sets the event source's 'last event id'.
             retry_duration: Retry duration in milliseconds.
             comment_message: A comment message. This value is ignored by clients and is used mostly for pinging.
+            ping_interval: Interval for sending message with "ping" body.
```
euri10, line 163:
```suggestion
            ping_interval: The interval in seconds between "ping" messages.
```

euri10 (changes requested):
I added some minor comments.
We'd need tests for the missing coverage and to test the functionnality too but this looks good overall!

RenameMe1:
> I added some minor comments. We'd need tests for the missing coverage and to test the functionnality too but this looks good overall!

I have read all the SSE tests and don't understand how I can test it. Can you help me, please?

I thought I would create an example with ping_interval and test it.  

- https://github.com/litestar-org/litestar/blob/main/docs/examples/responses/sse_responses.py

- https://github.com/litestar-org/litestar/blob/main/tests/examples/test_responses/test_sse_responses.py

euri10:
> I thought I would create an example with ping_interval and test it.
> 

sounds like a good idea to me, the only thing you need to know is that for the client side you want to use connect_sse or aconnect_sse from the httpx-sse package, we should have it installed as it's been used in test_sse.py in our suite.

as for the test I'd run it for 1s and count ping, so 10 expected if you wait 0.1s or somehting, in order to do that you'll have to have a tiemout param in the generator the sse using, see an example here:

https://github.com/euri10/litestar_saq_htmx/blob/main/tests/test_saq_htmx.py

obvisouly if you think of something else this is not the single option

euri10 (approved):
appart from updating the docs (yeah I know it's the boring part) it looks good to me.
I'll defer to others as to whether or not we want it here on a seperate follow-up PR

docs/examples/responses/sse_responses.py
```diff
@@ -26,9 +26,22 @@ async def my_generator() -> AsyncGenerator[SSEData, None]:
...
+        yield ServerSentEventMessage(data="content", event="message")
+
+
 @get(path="/count", sync_to_thread=False)
 def sse_handler() -> ServerSentEvent:
     return ServerSentEvent(my_generator())
 
 
-app = Litestar(route_handlers=[sse_handler])
+@get(path="/with_ping", sync_to_thread=False)
```
provinzkraut, line 42: This is missing a test case. Should be added here: https://github.com/litestar-org/litestar/blob/66d7dfa3851486376908deb32787d7228ce723da/tests/examples/test_responses/test_sse_responses.py

provinzkraut (commented):
Looks all good except for the missing docs. @euri10 IMO they should be added right away

docs/examples/responses/sse_responses.py
RenameMe1: Do you mean this test case? https://github.com/litestar-org/litestar/pull/4098/commits/7018a69fd283ae9af0a5e1f42a2255b6bda45c7a

provinzkraut: Sorry, seems that I've gotten confused about the location of the test cases :)

In this case, it's missing unit tests. The documentation test cases are only to ensure that the examples provided work as described in the docs, not to test the actual functionality (they do that as well, but that's mainly a side effect). For all features, there should exist distinct unit test.

The ping tests should go [here](https://github.com/litestar-org/litestar/blob/66d7dfa3851486376908deb32787d7228ce723da/tests/unit/test_response/test_sse.py). In the examples given in the documentation, we won't test edge cases for example, while unit tests should do exactly that.

RenameMe1: I added the simple unit test case. Maybe something else needs to be added to it?

litestar/response/streaming.py
```diff
@@ -110,6 +116,25 @@ async def _listen_for_disconnect(self, cancel_scope: CancelScope, receive: Recei
             else:
                 await self._listen_for_disconnect(cancel_scope=cancel_scope, receive=receive)
 
+    async def _send_ping_event(self, send: Send) -> None:
```
provinzkraut, line 119: This shouldn't go into the base `ASGIStreamingResponse`, as pings don't really have a meaning there. This belongs strictly to SSE.

Also, there's an issue with this design, in that the ping task should gracefully stop if the SSE event generator is exhausted

provinzkraut (changes requested):

RenameMe1:
@provinzkraut, friendly ping. Can you review, please?

I'm moved out work on ping messages to another class `ASGIStreamingSSERespnse`. Additionaly added state to event generator

euri10:
I didnt check in details thechanges but one thing I'd like to see if you dont mind is a little addition in the docs for that feature: https://docs.litestar.dev/latest/usage/responses.html#server-sent-event-responses

something like that below this would do the job:
> In your iterator function you can yield integers, strings or bytes, the message sent in that case will have message as the event_type if the ServerSentEvent has no event_type set, otherwise it will use the event_type specified, and the data will be the yielded value.

If ping_interval is set to a positive value, blablabla. It is useful for instance on Telegram blablabla, (see the issue)

litestar/response/streaming.py
```diff
@@ -145,10 +145,127 @@ async def send_body(self, send: Send, receive: Receive) -> None:
...
+        Returns:
+            None
+        """
+
+        async with create_task_group() as task_group:
+            task_group.start_soon(partial(self._stream, send))
+            if self.ping_interval:
+                task_group.start_soon(partial(self._send_ping_event, send))
+            await self._listen_for_disconnect(cancel_scope=task_group.cancel_scope, receive=receive)
+
```
euri10, line 260: I have no real clear explanation but my gut feeling here would be to use an anyio Event instead of the boolean `content_exist`, since it is share among several coroutines but I'll defer to the async pro @provinzkraut 

otherwise this lgtm, thanks for sticking with us :)

RenameMe1: I'm changed boolean to  anyio.Event

Thank you all for your reviews and helps

tests/unit/test_response/test_sse.py
```diff
@@ -99,3 +99,29 @@ async def numbers() -> AsyncIterator[SSEData]:
...
     with pytest.raises(ImproperlyConfiguredException):
         ServerSentEvent(content=object())  # type: ignore[arg-type]
+
+
+async def test_sse_ping_events() -> None:
+    @get("/test_ping")
+    async def handler() -> ServerSentEvent:
+        async def slow_generator() -> AsyncIterator[SSEData]:
+            for i in range(1):
+                await anyio.sleep(1)
```
sobolevn, line 109: I would propose using smaller sleep intervals for tests. We can just decrease `ping_interval`, but tests would become slower because of this.

tests/examples/test_responses/test_sse_responses.py
```diff
@@ -9,3 +10,13 @@ async def test_sse_responses_example() -> None:
...
             events = [sse async for sse in event_source.aiter_sse()]
             assert len(events) == 50
+
+
+async def test_sse_responses_example_with_ping_events() -> None:
+    async with AsyncTestClient(app=app2) as client:
+        async with aconnect_sse(client, "GET", f"{client.base_url}/with_ping") as event_source:
+            events = [sse async for sse in event_source.aiter_sse()]
+            for i in range(9):
+                assert events[i].event == " ping"
```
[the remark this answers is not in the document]
RenameMe1: Thanks. I resolved this mistake

RenameMe1:
Hello @provinzkraut, I'm moved out SSE implementation into  `litestar/response/see.py`. Can you see, please? 

Thanks for you help, and so sorry for mistakes ><

RenameMe1:
@provinzkraut, friendly ping. Can you review, please? 

I hope, my work is right

RenameMe1:
@provinzkraut, friendly ping. Can you review, please?

Maybe it would be better to close this PR and let other people do it?

PR #4101, opened 2025-04-10
docs: add mention in guard docs from where connection.user comes from

### LINKED ISSUE #3189: Docs: Add guard example to JWT docs
### Summary

Currently in the [JWT docs](https://docs.litestar.dev/latest/usage/security/jwt.html) there are a few references regarding how to access the 'User' object during and after being authenticated. These boil down to:

a) queried using the token
b) directly from the request (which can be easily assumed to be attached via retrieve_user_handler prior to going to the api path)

However there are other instances where user details need to be extracted from the `connection` object (such as in role-based guards).
```py
def admin_guard(connection: ASGIConnection, _: BaseRouteHandler) -> None:
    if not connection.user.is_admin:
        raies NotAuthorizedException()
```

A gap in knowledge between the page on JWTs and guards is that it's not made entirely clear _how_  user gets attached to connection. I would like to suggest that an example guard is added to the JWT docs with a comment explaining that the Auth object automatically attaches it for you based on the object returned from `retrieve_user_handler`. 

It also isn't made abundantly clear that the TypeVar provided to the Auth object directly corresponds to the retrieve_user_handler. For a little while, I was actually setting the TypeVar based on my login response and wondering why it wasn't working. A silly mistake in hindsight, but I believe a simple comment could have saved me from it!

euri10:
## Description

- used literalinclude as well

## Closes

fixes #3189

### DIFF
--- a/docs/usage/security/guards.rst
+++ b/docs/usage/security/guards.rst
@@ -13,81 +13,31 @@ the example.
 
 We begin by creating an :class:`~enum.Enum` with two roles - ``consumer`` and ``admin``:
 
-.. code-block:: python
+.. literalinclude:: /examples/security/guards.py
+    :language: python
+    :lines: 12-14
     :caption: Defining the enum ``UserRole``
-
-    from enum import Enum
-
-
-    class UserRole(str, Enum):
-       CONSUMER = "consumer"
-       ADMIN = "admin"
 
 Our ``User`` model will now look like this:
 
-.. dropdown:: Click to toggle the model ``User``
-
-    .. code-block:: python
+.. literalinclude:: /examples/security/guards.py
+        :language: python
+        :lines: 17-24
         :caption: User model for role based authorization
-
-        from pydantic import BaseModel, UUID4
-        from enum import Enum
-
-
-        class UserRole(str, Enum):
-           CONSUMER = "consumer"
-           ADMIN = "admin"
-
-
-        class User(BaseModel):
-           id: UUID4
-           role: UserRole
-
-           @property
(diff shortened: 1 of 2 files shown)


docs/usage/security/guards.rst
```diff
@@ -13,81 +13,31 @@ the example.
...
 
-        def admin_user_guard(connection: ASGIConnection, _: BaseRouteHandler) -> None:
-           if not connection.user.is_admin:
-               raise NotAuthorizedException()
-
-
-        @post(path="/user", guards=[admin_user_guard])
-        def create_user(data: User) -> User: ...
+The connection has a `user` object attached to it thanks to the JWT middleware, see :doc:`authentication </usage/security/jwt>`
+and in particular the :meth:`JWTAuth.retrieve_user_handler` method.
```
euri10, line 40: this is what solves the issue, the rest is just codeblock > literalinclude

docs/usage/security/guards.rst
```diff
@@ -13,81 +13,31 @@ the example.
...
-        class User(BaseModel):
-           id: UUID4
-           role: UserRole
-
-           @property
-           def is_admin(self) -> bool:
-               """Determines whether the user is an admin user"""
-               return self.role == UserRole.ADMIN
-
+Here the ``admin_user_guard`` guard checks if the user is an admin.
```
provinzkraut, line 37:
```suggestion
Here, the ``admin_user_guard`` guard checks if the user is an admin.
```

provinzkraut (approved):

PR #4116, opened 2025-04-13
chore: updated failing channels test

cofin:
Corrects an issue with the channels backend tests

### DIFF
--- a/tests/unit/test_channels/test_backends.py
+++ b/tests/unit/test_channels/test_backends.py
@@ -77,10 +77,11 @@ async def test_pub_sub_unsubscribe(channels_backend: ChannelsBackend) -> None:
     await channels_backend.publish(b"something", ["foo"])
 
     event_generator = channels_backend.stream_events()
+    await asyncio.wait_for(async_next(event_generator), timeout=0.01)
     await channels_backend.unsubscribe(["foo"])
-    await channels_backend.publish(b"something", ["bar"])
+    await channels_backend.publish(b"something else", ["bar"])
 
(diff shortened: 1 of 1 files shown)


Harshal6927 (approved):

provinzkraut:
Why is this test failing now and wasn't before? :eyes:

tests/unit/test_channels/test_backends.py
```diff
@@ -77,10 +77,11 @@ async def test_pub_sub_unsubscribe(channels_backend: ChannelsBackend) -> None:
     await channels_backend.publish(b"something", ["foo"])
 
     event_generator = channels_backend.stream_events()
+    await asyncio.wait_for(async_next(event_generator), timeout=0.01)
     await channels_backend.unsubscribe(["foo"])
-    await channels_backend.publish(b"something", ["bar"])
+    await channels_backend.publish(b"something else", ["bar"])
 
-    assert await asyncio.wait_for(async_next(event_generator), timeout=0.01) == ("bar", b"something")
+    assert await asyncio.wait_for(async_next(event_generator), timeout=0.01) == ("bar", b"something else")
```
provinzkraut, line 84: I don't think this test is correct anymore.

It's supposed to show that unsubscribing works, but now it has consumed all events on the `foo` channel *before* unsubscribing, so even if unsubscribing does not work, the test would still be green.

provinzkraut (changes requested):

tests/unit/test_channels/test_backends.py
cofin: For some reason on the `psycopg` backend, it was returning (`foo`, `something`) instead of `bar`

PR #4120, opened 2025-04-15
docs: 📚 Fix return value for UploadFile

Sube-py:
## Description

- Correct write() method docstring to specify int return value (bytes written)
- Update seek() method docstring to indicate int return value (new position)


provinzkraut (approved):
Thanks :)

PR #4123, opened 2025-04-17
feat: add custom DTO factory support for mixed model types

abdulhaq-e:
## Description

This allows for nesting mixed models, e.g. a SQLAlchemy model in a dataclass or any other mixture of models.


provinzkraut (changes requested):
The feature itself looks good, the implementation approach isn't ideal imo. This generates nested DTOs, whereas I think it should instead generate *one* DTO that's nested. Especially for the codegen backend, this should be much more efficient.

PR #4126, opened 2025-04-17
fix(typing): Remove usage of private `_AnnotatedAlias`

provinzkraut:
See https://github.com/python/typing_extensions/issues/586.

We're not using this functionality anymore, so keeping this around just for backwards compatibility.


sobolevn (approved):
This is a correct thing to do ✅

PR #4128, opened 2025-04-19
docs: show exceptions raised aren't always JSON

### LINKED ISSUE #3062: Docs: Exceptions are not always JSON by default
### Summary

The [documentation states](https://docs.litestar.dev/latest/usage/exceptions.html#exception-handling) that
> Litestar handles all errors by default by transforming them into JSON responses.

Since #2131 this is not *entirely* true as exceptions will retain the `MediaType` of the route handler.

Consider the following endpoint:
```py
@get(sync_to_thread=False)
def handler(q: int) -> str:
    raise ValueError
```
The `MediaType` of this handler will be `MediaType.TEXT` by default.
When let's say a `ValidationException` is raised due to wrong/missing query param or an internal error is raised, etc. the resulting response will be a JSON error message but in the response the `content-type` header will be set to `text/plain`.

If a client expects JSON on errors --by for example checking the status code-- as declared in the documentation, then JSON parsing *might* fail as the response `content-type` header does not indicate JSON.

euri10:
## Description

-

## Closes

fixes #3062

### DIFF
--- a/docs/usage/exceptions.rst
+++ b/docs/usage/exceptions.rst
@@ -78,10 +78,8 @@ Otherwise, the responses will default to ``500 - "Internal Server Error"``.
 
 The following handler for instance will default to ``MediaType.TEXT`` so the exception will be raised as text.
 
-.. literalinclude:: ../../tests/e2e/test_exception_handlers/test_exception_handlers.py
+.. literalinclude:: /examples/exceptions/implicit_media_type.py
     :language: python
-    :caption: Example of a handler that returns a text response
-    :lines: 89-91
 
 You can customize exception handling by passing a dictionary, mapping either status codes
 or exception classes to callables. For example, if you would like to replace the default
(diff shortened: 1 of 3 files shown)


docs/usage/exceptions.rst
```diff
@@ -75,6 +75,14 @@ Litestar handles all errors by default by transforming them into **JSON response
 :class:`~litestar.exceptions.http_exceptions.HTTPException`, the responses will include the appropriate ``status_code``.
 Otherwise, the responses will default to ``500 - "Internal Server Error"``.
 
+
+The following handler for instance will default to ``MediaType.TEXT`` so the exception will be raised as text.
+
+.. literalinclude:: ../../tests/e2e/test_exception_handlers/test_exception_handlers.py
```
provinzkraut, line 81: Can we add this to `examples`? Otherwise we'll end up breaking the docs just by changing some tests 👀

provinzkraut (approved):

PR #4132, opened 2025-04-20
fix(openapi): support prefixItems for tuple schema

### LINKED ISSUE #4130: Bug: OpenAPI Schema Generation Incorrect for Tuple Type Hint
### Description

When using a fixed-length tuple type hint like `tuple[int, int, int]` within a data structure (e.g., `msgspec.Struct`) returned by a route handler, the generated OpenAPI schema for that field is incorrect. Instead of defining the type for each item position using the array form of items, it generates a schema with `"items": {"oneOf": [...]}` containing redundant type definitions.

Example:

```py
class Color(Struct):
    rgb: tuple[int, int, int]
```

Actual Behavior:

The generated OpenAPI schema within components.schemas.Color.properties.rgb is:

```json
{
  "rgb": {
    "items": {
      "oneOf": [
        {
          "type": "integer"
        },
        {
          "type": "integer"
        },
        {
          "type": "integer"
        }
      ]
    },
    "type": "array"
  }
}
```

It should be:

``` json5
{
  "rgb": {
    "type": "array",
    "items": [         // Array form to define types per position
      { "type": "integer" }, // Type for index 0
      { "type": "integer" }, // Type for index 1
      { "type": "integer" }  // Type for index 2
    ],
  }
}
```

_No response_

### MCVE

```python
# Assuming Struct is from msgspec, adjust if it's different
from msgspec import Struct

# Define the data structure
class Color(Struct):
    rgb: tuple[int, int, int]

# Define the route handler
@get("/", sync_to_thread=True)
def test() -> Color:
    # Return an instance matching the Color structure
    return Color(rgb=(255, 0, 0))

# Configure the Litestar app with OpenAPI
app = Litestar(
    [test],
    lifespan=[],
    debug=True,
    openapi_config=OpenAPIConfig(
        render_plugins=[ScalarRenderPlugin()],
        title="My API",
        version="1.0.0",
    ),
)
```

### Litestar Version

litestar[standard]>=2.15.2

### Platform

- [x] Linux
- [x] Mac
- [x] Windows

Jannchie:
- resolved #4130 

The previous implementation incorrectly treated Python tuples as arrays with items of a single, uniform type.

This change utilizes prefixItems (OpenAPI 3.1+) to correctly specify the type for each element at its specific position when the field is a tuple. This results in a more accurate and modern OpenAPI schema that precisely reflects the Python tuple structure.

Additionally:

- Adjusted assertions in `test_schema_tuple_with_union` as the original expectations seemed incorrect with the proper schema generation.
- Added a new test case `test_schema_tuple`.

### DIFF
--- a/tests/unit/test_openapi/test_schema.py
+++ b/tests/unit/test_openapi/test_schema.py
@@ -469,7 +469,23 @@ def test_schema_tuple_with_union() -> None:
 
 def test_schema_tuple() -> None:
     schema = get_schema_for_field_definition(FieldDefinition.from_annotation(Tuple[int, str]))
-    assert schema.prefix_items == [Schema(type=OpenAPIType.INTEGER), Schema(type=OpenAPIType.STRING)]
+    assert schema == Schema(
+        prefix_items=[Schema(type=OpenAPIType.INTEGER), Schema(type=OpenAPIType.STRING)],
+        type=OpenAPIType.ARRAY,
+    )
+
+
+def test_schema_optional_tuple() -> None:
+    schema = get_schema_for_field_definition(FieldDefinition.from_annotation(Optional[Tuple[int, str]]))
+    assert schema == Schema(
+        one_of=[
+            Schema(
+                prefix_items=[Schema(type=OpenAPIType.INTEGER), Schema(type=OpenAPIType.STRING)],
+                type=OpenAPIType.ARRAY,
+            ),
+            Schema(type=OpenAPIType.NULL),
+        ],
+    )
 
 
 def test_optional_enum() -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/test_openapi/test_schema.py
```diff
@@ -461,13 +461,17 @@ def test_schema_generation_with_ellipsis() -> None:
...
         Schema(type=OpenAPIType.INTEGER),
         Schema(one_of=[Schema(type=OpenAPIType.INTEGER), Schema(type=OpenAPIType.STRING)]),
     ]
 
 
+def test_schema_tuple() -> None:
+    schema = get_schema_for_field_definition(FieldDefinition.from_annotation(Tuple[int, str]))
+    assert schema.prefix_items == [Schema(type=OpenAPIType.INTEGER), Schema(type=OpenAPIType.STRING)]
+
+
```
provinzkraut, line 474: Can you add a test that shows that this works if the tuple if enclose in another generic, e.g. `Optional[Tuple[int, int]]`?

provinzkraut (approved):

PR #4133, opened 2025-04-20
feat(openapi): add custom example ids support

### LINKED ISSUE #4013: Bug: `Example` class is missing an ID field
### Description

The `litestar.openapi.spec.Example` class is missing an `id` field that would allow IDs to be customized when they are autogenerated when you provide multiple `Example`s to `ResponseSpec`.

As a workaround, I have had to set the `summary` field of `Example`s and then use the below custom code to monkey patch my OpenAPI specification:
```
import re

from litestar import Litestar


def _fix_autogenerated_example_ids(spec: dict) -> None:
    """Fix autogenerated example IDs in the provided OpenAPI schema in place,
    preferring to use the example's `summary` field if one exists, otherwise
    using `example` if there is only one example, and otherwise using `example #<index>`
    for each example where there are multiple examples.
    
    See https://github.com/litestar-org/litestar/issues/4013"""

    if isinstance(spec, dict):
        for k, v in spec.items():
            # Filter for example objects.
            if k == "examples" and isinstance(v, dict):
                # Filter for example IDs matching the format `<nonwhitespace_lowercase_alphabetic_string>-example-<digit/s>`.
                if all(re.search(r"[a-z]+-example-\d+", k2) for k2 in v.keys()):
                    # If every example has a `summary` field, use that as the new example ID.
                    if len({e["summary"] for e in v.values() if "summary" in e}) == len(v.values()):
                        new_examples = {}

                        for example in v.values():
                            new_examples[example["summary"]] = example
                    
                    else:
                        # If there's only one example, use the word "example" as the new example ID.
                        if len(v) == 1:
                            new_examples = {"example": list(v.values())[0]}

                        # Otherwise, use the format `example #<index>` as the new example ID.
                        else:
                            new_examples = {}

                            for idx, example in enumerate(v.values(), start=1):
                                new_examples[f"example #{idx}"] = example

                    spec[k] = new_examples
            
            # Recursively fix example IDs in nested objects.
            _fix_autogenerated_example_ids(v)
    
    # Recursively fix example IDs in nested lists.
    elif isinstance(spec, list):
        for elm in spec:
            _fix_autogenerated_example_ids(elm)


def gen_openapi(app: Litestar) -> dict:
    """Generate the OpenAPI schema for the provided Litestar application, applying necessary transformations to the schema."""
    
    # Generate the OpenAPI schema.
    schema = app.openapi_schema
    schema = schema.to_schema()

    # Fix example IDs autogenerated by Litestar.
    _fix_autogenerated_example_ids(schema)

    return schema
```

_No response_

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
```

### Logs

```bash

```

### Litestar Version

2.14.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Jannchie:
- resolved #4013

### DIFF
--- a/litestar/_openapi/schema_generation/utils.py
+++ b/litestar/_openapi/schema_generation/utils.py
@@ -34,13 +34,27 @@ def _should_create_literal_schema(field_definition: FieldDefinition) -> bool:
     )
 
 
+def get_example_id(example: Example, name: str, index: int) -> str:
+    """Get the example ID.
+
+    Args:
+        example: The example instance.
+        name: The name of the field.
+        index: The index of the example.
+
+    Returns:
+        The example ID.
+    """
+    return example.id or f"{name}-example-{index}"
+
+
 def get_formatted_examples(field_definition: FieldDefinition, examples: Sequence[Example]) -> Mapping[str, Example]:
     """Format the examples into the OpenAPI schema format."""
 
     name = field_definition.name or get_name(field_definition.type_)
     name = name.lower()
 
-    return {f"{name}-example-{i}": example for i, example in enumerate(examples, 1)}
+    return {get_example_id(example, name, i): example for i, example in enumerate(examples, 1)}
 
 
 def get_json_schema_formatted_examples(examples: Sequence[Example]) -> list[Any]:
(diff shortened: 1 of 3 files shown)


provinzkraut (approved):
That's a great addition, thanks!

provinzkraut:
@all-contributors add @Jannchie for code

PR #4142, opened 2025-04-24
fix: add a AsyncTestClientTransport for the AsyncTestClient

### LINKED ISSUE #1920: Bug: the app for test client is running in a different event loop
### Description

I've been trying to make rollbacks for database changes for the SQLAlchemy plugin. See the discussion:

https://github.com/orgs/litestar-org/discussions/1919

And I've encountered the issue described in the discussion above and in this issue:

https://github.com/encode/starlette/issues/1315

```
RuntimeError: Task <Task pending name='anyio.from_thread.BlockingPortal._call_func' coro=<BlockingPortal._call_func() running at /.../anyio/from_thread.py:219> cb=[TaskGroup._spawn.<locals>.task_done() at /.../anyio/_backends/_asyncio.py:726]> got Future <Future pending cb=[Protocol._on_waiter_completed()]> attached to a different loop
```

After diving into the issue deeper, I've discovered that `BaseTestClient.portal` always creates the portal in a separate thread and runs in this thread a new event loop. This behavior is described in the `anyio.from_thread.start_blocking_portal` function docstring:

> Start a new event loop in a new thread and run a blocking portal in its main task.

That means if I create in tests the database connection before calling `AsyncTestClient.__aenter__` (or `TestClient.__enter__`), it will be running in the main thread pytest-asyncio event loop and so it cannot be used as `SQLAlchemyAsyncConfig.session_maker`. And if you try, with the power of monkey patching, to create the database connection after entering the client context, you cannot use this database connection in the main thread in tests.

_No response_

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.0.0b2

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

euri10:
## Description

-

## Closes

fixes #1920

### DIFF
--- a/litestar/testing/transport.py
+++ b/litestar/testing/transport.py
@@ -194,9 +194,6 @@ class AsyncTestClientTransport(BaseTestClientTransport):
         else:
             return self._prepare_response(request, context, raw_kwargs)
 
-    async def aclose(self) -> None:
-        """Close the transport."""
-
 
 class TestClientTransport(BaseTestClientTransport):
     def handle_request(self, request: Request) -> Response:
--- a/litestar/testing/transport.py
+++ b/litestar/testing/transport.py
@@ -43,7 +43,7 @@ class SendReceiveContext(TypedDict):
     context: Any | None
 
 
-class AsyncTestClientTransport(Generic[T]):
+class BaseTestClientTransport(Generic[T]):
     def __init__(
         self,
         client: T,
@@ -138,13 +138,15 @@ class AsyncTestClientTransport(Generic[T]):
             "server": (host, port),
         }
 
+
+class AsyncTestClientTransport(BaseTestClientTransport):
     async def handle_async_request(self, request: Request) -> Response:
         scope = self.parse_request(request=request)
         if scope["type"] == "websocket":
             scope.update(
                 subprotocols=[value.strip() for value in request.headers.get("sec-websocket-protocol", "").split(",")]
             )
-            session = WebSocketTestSession(client=self.client, scope=cast("WebSocketScope", scope))  # type: ignore[arg-type]
+            session = WebSocketTestSession(client=self.client, scope=cast("WebSocketScope", scope))
             raise ConnectionUpgradeExceptionError(session)
 
         scope.update(method=request.method, http_version="1.1", extensions={"http.response.template": {}})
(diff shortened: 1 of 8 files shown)


euri10:
any idea how to cover this ? I dont understand why it pops now though

tests/unit/test_testing/test_lifespan_handler.py
```diff
@@ -35,3 +38,17 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+
+
+async def test_multiple_clients_event_loop() -> None:
+    @get("/")
+    def return_loop_id() -> dict:
+        return {"loop_id": id(get_event_loop())}
+
+    app = Litestar(route_handlers=[return_loop_id])
+
+    async with AsyncTestClient(app) as client_1, AsyncTestClient(app) as client_2:
```
provinzkraut, line 50: This should even be the same loop with different client instances, right?

euri10: i think so, this added test is just the mcve from the issue and it fails before the changes, is there something I should add ?

euri10: I realize the diff is overly complicate, the pr is simple: before AsyncTestClient used TestClientTransport hence the issue.
so I created a BaseClientTransport that essentially takes care of request/response and added a ASGITestClientTransport that inherits httpx ASGIBaseTransport and ovverrides handle_async_request while TestClientTransport overrides handle_request, the BaseClientTransport is what's common between te 2 transports.

provinzkraut: > i think so, this added test is just the mcve from the issue and it fails before the changes, is there something I should add ?

Yeah, maybe a test for the transport in particular that shows the client always uses the currently running loop?

euri10: actually if this test was using TestClient instead of AsyncTestClient it wouldn't pass as `start_blocking_portal` creates a new event loop.

> This should even be the same loop with different client instances, right?

by different client instances you meant a AsyncTestClient and a TestClient ?

provinzkraut: No I mean different instances `AsyncClient`. But actually, I think the test should probably just show that the client uses the running loop from the context it's created in.

litestar/testing/transport.py
provinzkraut, line 1: With this refactoring done, I think you actually don't even need 2 transports. The sync version could just do

```python
with self.client.portal() as portal:
    return portal.call(self.handle_async_request, request)
```

provinzkraut:
I don't think the behaviour is fully fixed yet, as the `LifespanHandler` still uses a portal.

```python
async def test_lifespan_loop() -> None:
    mock = MagicMock()

    @contextlib.asynccontextmanager
    async def lifespan(app: Litestar) -> AsyncGenerator[None, None]:
        mock(asyncio.get_running_loop())
        yield

    app = Litestar(lifespan=[lifespan])

    async with AsyncTestClient(app):
        pass

    mock.assert_called_once_with(asyncio.get_running_loop())
```

litestar/testing/transport.py
euri10: the benefit of having 2 transports is that they respectively inherit their httpx counterparts so I think it's a little bit more mypy friendly but I may be wrong

provinzkraut: Okay, that makes sense. Maybe could still move the request handling fn into the common base class?

euri10:
alright maybe we nee don top of this a version of LifespanHandler that "skips" the portal in case of the AsyncClient ?

provinzkraut:
> alright maybe we nee don top of this a version of LifespanHandler that "skips" the portal in case of the AsyncClient ?

I think so. Or, probably easier to do, we just have `LifespanHandler` be async by default, and wrap it with a portal call for the sync implementation?

euri10:
im kinda bocked on 2 failing tests during the lifespan shutdown though the new tests work fine, but I dont see a clear reason, any idea ?

provinzkraut:
> im kinda bocked on 2 failing tests during the lifespan shutdown though the new tests work fine, but I dont see a clear reason, any idea ?

Removing the `pytest.raises` gives us this:

```
FAILED tests/unit/test_testing/test_test_client.py::test_error_handling_on_shutdown[AsyncTestClient-asyncio] - exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
FAILED tests/unit/test_testing/test_test_client.py::test_error_handling_on_shutdown[AsyncTestClient-trio] - exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
FAILED tests/unit/test_testing/test_test_client.py::test_error_handling_on_shutdown[TestClient-asyncio] - RuntimeError
FAILED tests/unit/test_testing/test_test_client.py::test_error_handling_on_shutdown[TestClient-trio] - RuntimeError
```

so it seems that the issue is that the async client raises an exception group, where the sync client just raises a regular exception. The reason for that is that the async lifespan handler handles things in a task group. 

I think the cleanest solution here would be to have the implementation be in the async handler only, and just adapt that for the sync version with the portal calls. Then the behaviour should be the same. (I think)

euri10:
closing, this was fixed by https://github.com/litestar-org/litestar/pull/4291

Issue #4141, opened 2025-04-24
Bug: `sync_to_thread` warning shows on `AbstractAsyncOffsetPaginator`


ames0k0:
### Description

Code from the docs: [Offset Pagination With SQLAlchemy](https://docs.litestar.dev/2/usage/responses.html#offset-pagination-with-sqlalchemy)

### URL to code causing the issue

https://github.com/litestar-org/litestar/blob/main/litestar/di.py#L80

### MCVE

```python
# debug prints
print(f"{has_sync_callable=}")
print(f"{has_generator_dependency=}")
print(f"{is_class_dependency=}")
print(f"{is_async_callable(dependency)=}")
print(f"{dependency=}")
```

### Steps to reproduce

1. Copy & Paste

### Logs

```text
has_sync_callable=True
has_generator_dependency=False
is_class_dependency=True
is_async_callable(dependency)=True
dependency=<class '__main__.PersonOffsetPaginator'>
.../main2.py:44: LitestarWarning: Use of a synchronous callable <class '__main__.PersonOffsetPaginator'> without setting sync_to_thread is discouraged since synchronous callables can block the main thread if they perform blocking operations. If the callable is guaranteed to be non-blocking, you can set sync_to_thread=False to skip this warning, or set the environmentvariable LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD=0 to disable warnings of this type entirely.
  @get("/people", dependencies={"paginator": Provide(PersonOffsetPaginator)})
```

### Litestar Version

2.15.2

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

provinzkraut:
Not a bug. The docs are wrong. It should be `Provide(PersonOffsetPaginator, sync_to_thread=False)`, like the warning suggests.

ames0k0:
Sure, but do I really need to pass this flag for an asynchronous code ?

In the docs, with flag `sync_to_thread` highlighted only `synchronous function`:
- [Router Handler](https://docs.litestar.dev/2/usage/routing/handlers.html)
- [Dependency Injection](https://docs.litestar.dev/2/usage/dependency-injection.html)
- [Sync vs. Async](https://docs.litestar.dev/2/topics/sync-vs-async.html)


```
# litestar/di.py

sync_to_thread: Run sync code in an async thread. Defaults to False.
```

```
# litestar/handlers/http_handlers/base.py

[...]

PR #4144, opened 2025-04-24
docs: Update `Offset Pagination With SQLAlchemy` example

ames0k0:
Add missing `sync_to_thread=False` argument
Also fixed the `slice` method arguments

## Description

`slice` context
```python
# sqlalchemy/sql/util.py
# start=1, stop=3
limit_clause = _offset_or_limit_clause(stop - start)
```
```sql
# sqlite
select * from person limit -2 offset 3;
-- ...

# postgres
select * from person limit -2 offset 3;
-- LIMIT must not be negative
```

## Closes
https://github.com/litestar-org/litestar/issues/4141


provinzkraut (approved):
Looks good!

PR #4148, opened 2025-04-26
fix(DI): Ensure dependencies are cleaned up when exception occurrs during cleanup 

provinzkraut:
Fix a bug where a dependencies' cleanup was not guaranteed to be called would not be called if an exception was thrown:

- Late during ASGI response rendering
- During cleanup of another dependency when recovering from an exception

<hr>

Refactor the code such that:

- Exceptions are always caught and thrown into dependencies as long as the dependencies are alive
- If the handler raises an exception and during dependency cleanup other exceptions are raised, the sub-exceptions are grouped into an `ExceptionGroup`. This behaviour is now equivalent to regular dependency cleanup

### DIFF
--- a/docs/usage/dependency-injection.rst
+++ b/docs/usage/dependency-injection.rst
@@ -153,9 +153,9 @@ and committing otherwise.
 .. admonition:: Best Practice
     :class: tip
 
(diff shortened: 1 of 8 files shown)


cofin (approved):
LGTM

PR #4150, opened 2025-04-28
fix: possibly get httpexception.extra and httpexception.headers from class

### LINKED ISSUE #4006: Bug: Unable to set default `extra` on exception
### Description

Using the below code to define a custom exception does not let the `extra` field be returned in responses where the error is raised:
```
class InvalidApiKeyError(NotAuthorizedException):
    detail = "Invalid API key"
    extra = "The API key you provided is invalid, incorrect, expired, has been revoked, or your account is overdue."
```

The below code doesn't work either:
```
class InvalidApiKeyError(NotAuthorizedException):
    detail = "Invalid API key"

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.extra = "The API key you provided is invalid, incorrect, expired, has been revoked, or your account is overdue."
```

_No response_

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
```

### Logs

```bash

```

### Litestar Version

2.14.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

PokkaKiyo:
## Description
- For HTTPException, possibly get the value for extra and headers on the class

## Fixes 
- #4006


litestar/exceptions/http_exceptions.py
```diff
@@ -69,7 +70,7 @@ def __init__(
         """
         super().__init__(*args, detail=detail)
         self.status_code = status_code or self.status_code
-        self.extra = extra
+        self.extra = extra if extra is not Empty else getattr(self, "extra", None)
```
provinzkraut, line 73: Why not do the same as with `status_code` and set a default on the class?

PokkaKiyo: That was my first thought too.

But then I think it would be good to support the pattern where suppose a class with a default extra is defined, and if I don't actually want that extra, I don't want to define another class without that extra, I would simply set extra=None on the constructor, and to do that another sentinel value would be needed.

Suppose if OTPError is defined, I want to just do extra=None in my constructor, like
```python
class OTPError(HTTPException):
    status_code = HTTP_401_UNAUTHORIZED
    extra = {"action": "Generate another OTP"}

raise OTPError(extra=None)
```
Instead of
```python
class OTPError(HTTPException):
    status_code = HTTP_401_UNAUTHORIZED
    extra = {"action": "Generate another OTP"}

class OTPErrorNoExtra(OTPError):  # Naming is a pain
    extra = None

raise OTPErrorNoExtra()
```
Or refactoring
```python
class BaseOTPError(HTTPException):
    status_code = HTTP_401_UNAUTHORIZED

class OTPError(BaseOTPError):  # Refactoring is a pain
    # status_code = HTTP_401_UNAUTHORIZED
    extra = {"action": "Generate another OTP"}

raise BaseOTPError()
```

provinzkraut: > But then I think it would be good to support the pattern where suppose a class with a default extra is defined, and if I don't actually want that extra, I don't want to define another class without that extra, I would simply set extra=None on the constructor, and to do that another sentinel value would be needed.

That's what `Empty` is for :)

PokkaKiyo: Think I got it (hopefully)!

provinzkraut (commented):
This looks good but there's one thing I hadn't considered before (sorry for that): This is now a breaking change due to the fact that `extra` can now be `EmptyType` as well. 

So unless we find another version that doesn't alter the signatures, we'll have to wait for 3.0

provinzkraut:
> In the meantime, would it make sense, for v3, to apply the same changes to headers as well?

Yes, absolutely!

litestar/exceptions/http_exceptions.py
```diff
@@ -43,18 +44,18 @@ class HTTPException(LitestarException):
...
 
     def __init__(
         self,
         *args: Any,
         detail: str = "",
         status_code: int | None = None,
-        headers: dict[str, str] | None = None,
-        extra: dict[str, Any] | list[Any] | None = None,
+        headers: dict[str, str] | None | EmptyType = Empty,
+        extra: dict[str, Any] | list[Any] | None | EmptyType = Empty,
```
provinzkraut, line 58: This behaviour should be documented in the args docstring

PokkaKiyo: added!

provinzkraut (approved):
Looks good! Thanks!

PokkaKiyo:
@provinzkraut I've moved the changes for this into #4276. Feel free to take a look at your convenience 😊.

PR #4152, opened 2025-04-29
fix: misleading error message upon ImportError with --app argument to CLI

### LINKED ISSUE #4129: Bug: Misleading and unhelpful error message upon `ImportError` with `--app` argument to CLI
### Description

When using `litestar run` with application lookup through a directory (e.g. the CWD), and an `ImportError` happens, you get a normal traceback, showing exactly what is going on. However, if you instead import the app through the environment, as `litestar --app "app_module:app` run`, and any kind of `ImportError` happens, the only thing you get is this:
```
Using Litestar app from env: 'app_module:app'
 Invalid argument passed --app 'app_module:app': Module not found
```
In particular, this misleadingly seems to suggests that `app_module` cannot be found. In reality, you get the same error message no matter which module caused the `ImportError`. This issue is exacerbated by the fact that these errors typically happen due to a packaging problem, which often is only identified in CI testing or the staging environment, where direct analysis of the issue is not possible.

_No response_

### MCVE

```python

# the following module does not exist, highlighting an ImportError with a dependency.
import does_not_exist 

@get("/")
async def hello_world() -> dict[str, str]:
    return {"hello": "world"}


```

### Steps to reproduce

1. Put the above MCVE into a file `app_module.py`
2. `cd` into that directory. 
3. Run `litestar --app app_module:app run` (ensure that `litestar` is installed).

### Screenshots

_No response_

### Logs

```text
Using Litestar app from env: 'app_module:app'
 Invalid argument passed --app 'app_module:app': Module not found
```

### Litestar Version

2.15.2

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

cstarkers:
## Description

Fixes #4129 

Before this PR, if the `--app` CLI option is used, any `ImportErrors` that occur while instantiating the application are assumed to be caused by a bad `--app` argument.  This does not account for cases where the `--app` option is actually fine, but the `ImportError` arose from inside the applicaiton code, and this results in a misleading error message.

This PR adds a check on whether the exception resulted from a failure to import the module named in the `--app` option.  If the error arose while attempting some other import, it is not caught.  This is consistent with the behaviour that occurs when `--app` is not used.

### DIFF
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -283,10 +283,12 @@ def _validate_app_path(app_path: str) -> tuple[ModuleType, str]:
 
     try:
         module = importlib.import_module(module_path)
-    except ImportError:
-        console.print(f"[bold red] Invalid argument passed --app {app_path!r}: Module not found")
-        sys.exit(1)
-
+    except ImportError as e:
+        if e.name == module_path:
+            console.print(f"[bold red] Invalid argument passed --app {app_path!r}: Module not found")
+            sys.exit(1)
+        else:
+            raise (e)
     return module, app_name
 
 
--- a/tests/unit/test_cli/test_cli.py
+++ b/tests/unit/test_cli/test_cli.py
@@ -114,3 +114,20 @@ def test_incorrect_app_argument(
     assert result.exit_code == 1
 
     mock.assert_not_called()
+
+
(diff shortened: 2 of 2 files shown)


provinzkraut (approved):
That's a reasonable fix, thanks!

provinzkraut (approved):

PR #4158, opened 2025-05-01
fix(openapi): wrong annotation passed to FieldDefinition (DTO)

### LINKED ISSUE #3888: Bug: OpenAPI spec blank response schema when using Response object
### Description

See example below - using the Scalar docs the schema for the  `get_items` endpoint is blank (just says "No Body"). Ideally the schema would be populated exactly as if the response wasn't wrapped in a "Response" object, i.e. it should just use the type hinting info `List[Item]`.

http://127.0.0.1:8000/schema

### MCVE

```python
from typing import List

from litestar import Litestar, get, Response
from litestar.dto import DataclassDTO, DTOConfig


@dataclass
class Item:
    id: int
    name: str
    secret: str


class ItemReadDTO(DataclassDTO[Item]):
    config = DTOConfig(exclude=["secret"])


@get("/get_items", return_dto=ItemReadDTO)
async def get_items() -> Response[List[Item]]:
    return Response(
        headers={"X-SomeParam": "SomeValue"},
        content=[
            Item(id=1, name="Item 1", secret='123'),
            Item(id=2, name="Item 2", secret='456'),
        ]
    )


@get("/get_items_raw", return_dto=ItemReadDTO)
async def get_items_raw() -> List[Item]:
    return [
        Item(id=1, name="Item 1", secret='123'),
        Item(id=2, name="Item 2", secret='456'),
    ]


app = Litestar(
    route_handlers=[get_items, get_items_raw],
    openapi_config=OpenAPIConfig(
        title="Scalar Issue",
        version="0.1.0",
        render_plugins=[ScalarRenderPlugin()],
    ),
    debug=True,
)
```

### Steps to reproduce

```bash
run uvicorn main:app
```

### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.13.0

### Platform

- [X] Linux
- [ ] Mac
- [X] Windows

mmev:
## Description

Fix #3888

Fixed an issue involving incorrect annotations being passed to `FieldDefinition` when handling `Response` class. 
However, I'm not entirely satisfied with this solution. I'd be happy for any suggestions for alternative approaches.

## Closes
- litestar-org/litestar#3888

### DIFF
--- a/litestar/dto/base_dto.py
+++ b/litestar/dto/base_dto.py
@@ -15,6 +15,7 @@ from litestar.dto.data_structures import DTOData
 from litestar.dto.types import RenameStrategy
 from litestar.enums import RequestEncodingType
 from litestar.exceptions.dto_exceptions import InvalidAnnotationException
+from litestar.response.base import Response
 from litestar.types.builtin_types import NoneType
 from litestar.types.composite_types import TypeEncodersMap
 from litestar.typing import FieldDefinition
@@ -229,7 +230,7 @@ class AbstractDTO(Generic[T]):
         key = "data_backend" if field_definition.name == "data" else "return_backend"
         backend = cls._dto_backends[handler_id][key]  # type: ignore[literal-required]
 
-        if backend.wrapper_attribute_name:
+        if backend.wrapper_attribute_name and not field_definition.is_subclass_of(Response):
             # The DTO has been built for a handler that has a DTO supported type wrapped in a generic type.
             #
             # The backend doesn't receive the full annotation, only the type of the attribute on the outer type that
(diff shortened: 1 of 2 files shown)


litestar/dto/base_dto.py
```diff
@@ -227,7 +228,7 @@ def create_openapi_schema(
         key = "data_backend" if field_definition.name == "data" else "return_backend"
         backend = cls._dto_backends[handler_id][key]  # type: ignore[literal-required]
 
-        if backend.wrapper_attribute_name:
+        if backend.wrapper_attribute_name and not field_definition.is_subclass_of(Response):
```
provinzkraut, line 231: I'm not sure if this is the right part to fix, because, as this section indicates, we *do* supported wrapped return types, such as `Response`.

https://github.com/litestar-org/litestar/blob/af1af6ba0b528cba3e90c44b83421e75b681ce64/tests/unit/test_dto/test_factory/test_integration.py#L685-L714

If this is the same case and just a bug within that, could you amend the test case so the schema is checked within the test cases linked above? That makes it easier to track which behaviour belongs together logically

provinzkraut: I'm still not sure about the special casing of `Response`. This should work with all generic container types really. 

We do have `AbstractDTO.resolve_generic_wrapper_type`, which is supposed to do what you're special casing here, so I think we have a bug somewhere in that function, where it's not correctly detecting / unwrapping things

provinzkraut: Thanks for the thorough investigation @mmev! I think ultimately, we need to do some refactoring to get all the response special-casing out, but for now, this is a reasonable fix.

provinzkraut (approved):

PR #4160, opened 2025-05-01
feat(DTO): Add custom attribute accessor callable

provinzkraut:
Add a new `attribute_accessor` property to `AbstractDTO`, which can be used by both the codegen and functional backend to access attributes on an object.

This can be useful to implement DTOs for types that do not adhere to the plain protocol of `obj.data_field` or need to add additional checks, e.g. using `obj.related_field.all()` on a Django model to access related objects, while ensuring no I/O occurs.

### DIFF
--- a/tests/unit/test_dto/test_factory/test_backends/test_backends.py
+++ b/tests/unit/test_dto/test_factory/test_backends/test_backends.py
@@ -1,6 +1,7 @@
 # ruff: noqa: UP006,UP007
 from __future__ import annotations
 
+import inspect
 from dataclasses import dataclass, field
 from types import ModuleType
 from typing import TYPE_CHECKING, Callable, Dict, List, Optional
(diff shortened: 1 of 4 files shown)


tests/unit/test_dto/test_factory/test_backends/test_backends.py
```diff
@@ -351,6 +353,64 @@ class Factory(DataclassDTO):
...
+        model_type=Foo,
+        wrapper_attribute_name=None,
+        is_data_field=False,
+    )
+
+    backend.encode_data(Foo(id="1"))
+    mock.assert_called_once_with("id")
+
+
+@pytest.mark.xfail()
```
provinzkraut, line 387: This needs #4159 to work

cofin (approved):
This is really simple but adds quite a bit of flexibility!

PR #4162, opened 2025-05-02
feat: add scalar configuration

### LINKED ISSUE #3951: Enhancement: Support Scalar configuration
### Summary

While researching the Scalar project, I found that some custom config is avaliable, the full list can be found here: https://github.com/scalar/scalar/blob/main/documentation/configuration.md.

### Basic Example

There are an example from Scalar repo, for FastAPI, and support config: https://github.com/scalar/scalar/blob/main/packages/scalar_fastapi/scalar_fastapi/scalar_fastapi.py.

We need to change this file: https://github.com/litestar-org/litestar/blob/main/litestar/openapi/plugins.py#L351, `ScalarRenderPlugin`, to support this new feature. I also expect the config will be fully typed with Pydantic.

### Drawbacks and Impact

I don't think this enhancement will comes with any drawbacks to current users.

### Unresolved questions

Maybe we should also support to pass the config directly to Scalar in case of new config items added and we can use them without a Litestar update.

skv0zsneg:
## Description

Hello! I'm currently work on adding scalar configuration and during work got some questions:
 
- I start adding config params from [here](https://github.com/scalar/scalar/blob/main/documentation/configuration.md), but not sure about including every param. What should we add?
- Here is MRE for how to work with configuration:
```python
from litestar.openapi.datastructures import ScalarConfig

scalar_plugin = ScalarRenderPlugin(version="1.19.5", config=ScalarConfig(show_sidebar=False))

@get("/")
async def hello_world() -> dict[str, str]:
    """Keeping the tradition alive with hello world."""
    return {"hello": "world"}

app = Litestar(
    route_handlers=[hello_world],
    openapi_config=OpenAPIConfig(
        title="Litestar Example",
        description="Example of Litestar with Scalar OpenAPI docs",
        version="0.0.1",
        render_plugins=[scalar_plugin],
        path="/docs",
    ),
)
```
I'm not sure about config `dataclass` object location. Maybe it must be somewhere else?
- The config could change with changing Scalar version and we have `version` arg in `ScalarRenderPlugin`. So should we somehow detect difference?

## Closes
#3951

### DIFF
--- a/litestar/openapi/plugins.py
+++ b/litestar/openapi/plugins.py
@@ -1,24 +1,25 @@
 from __future__ import annotations
 
+import json
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, Sequence
 from dataclasses import asdict
+from typing import TYPE_CHECKING, Any, Sequence
 
 import msgspec
 import yaml
-import json
 
 from litestar.constants import OPENAPI_JSON_HANDLER_NAME
+from litestar.dto._backend import _camelize
 from litestar.enums import MediaType, OpenAPIMediaType
 from litestar.handlers import get
-from litestar.openapi.datastructures import ScalarConfig
 from litestar.serialization import encode_json, get_serializer
-from litestar.dto._backend import _camelize
 
 if TYPE_CHECKING:
     from litestar.config.csrf import CSRFConfig
     from litestar.connection import Request
+    from litestar.openapi.datastructures import ScalarConfig
     from litestar.router import Router
+
 
 __all__ = (
     "OpenAPIRenderPlugin",
@@ -380,6 +381,8 @@ class ScalarRenderPlugin(OpenAPIRenderPlugin):
             css_url: Download url for the Scalar CSS bundle.
                 If not provided, the Litestar-provided CSS will be used.
             path: Path to serve the OpenAPI UI at.
+            config: Scalar configuration.
+                If not provided the default Scalar configuration will be used.
             **kwargs: Additional arguments to pass to the base class.
         """
         self.js_url = js_url or f"https://cdn.jsdelivr.net/npm/@scalar/api-reference@{version}"
--- a/litestar/openapi/plugins.py
+++ b/litestar/openapi/plugins.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 from abc import ABC, abstractmethod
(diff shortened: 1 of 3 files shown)


litestar/openapi/plugins.py
```diff
@@ -423,6 +433,21 @@ def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
...
                 """.encode()
 
+    def render_config(self) -> str:
+        """Render Scalar configuration to JS object."""
+        if not self.config:
+            return ""
+
+        dict_config = {}
+        for param, value in asdict(self.config).items():
+            dict_config[_camelize(param, False)] = value
```
provinzkraut, line 443: If you make the config an `msgspec.Struct` you could just do `json_config = msgspec.json.encode(self.config)`

```python
class ScalarConfig(msgspec.Struct, rename="camel"):
 ...
```

provinzkraut:
Generally a good approach, but I'm not sure we need the config object at all. You could just add the options as arguments to the plugin directly, as it's already specialised for scalar.


> The config could change with changing Scalar version and we have version arg in ScalarRenderPlugin. So should we somehow detect difference?

One option to do this would be to just accept a `scalar_options: dict[str, Any]`. Then the user is responsible for providing the correct options corresponding to the version.

Actually, I think this is the right approach, because otherwise we'd always have to make Litestar release before a new scalar option could be used.

litestar/openapi/plugins.py
```diff
@@ -423,6 +430,16 @@ def render(self, request: Request, openapi_schema: dict[str, Any]) -> bytes:
                     </html>
                 """.encode()
 
+    def render_options(self) -> str:
+        """Render options to Scalar configuration."""
+        if not self.options:
+            return ""
+        return f"""
+                <script>
+                  document.getElementById('api-reference').dataset.configuration = '{json.dumps(self.options)}'
```
provinzkraut, line 439:
```suggestion
                  document.getElementById('api-reference').dataset.configuration = '{msgspec.json.encode(self.options).decode()}'
``` 
:)

provinzkraut (approved):

provinzkraut:
Looks good! If you address the missing coverage, this can go in!

a-wip0:
i see this PR already merged to main, is the documentation not exist? https://docs.litestar.dev/2/search.html?q=ScalarConfig

PR #4167, opened 2025-05-10
feat: Middleware constraints

provinzkraut:
## What?

Add a constraints framework for middlewares. These constraints can be defined on middleware classes / instances, and will be checked once the middleware stack for a handler have been resolved. 

Currently implemented checks are:

- Any 'A' must come before any 'B'
- Any 'A' must come after any 'B'
- Middleware must be at the top of the stack
- Middleware at the bottom of the stack
- Middleware must be unique in the stack

All checks take into account the class hierarchy, so `Any 'A' must come before any 'B'` actually means `Any type, subclass or instance of 'A' must come before any type, subclass or instance of 'B'`.

Setting constraints is only supported on `ASGIMiddleware`, but constraints referring to other middleware, may refer to any middleware type (class, instance or factory function). In addition, constraints referring to other middleware may to so by providing a string reference, which will be resolved at runtime. `ImportError`s can optionally be skipped here, making it possible for a constraint to only apply if e.g. a specific 3rd party library is installed.

## Why?

Even though middlewares ideally should be agnostic of the order they are in, for some, this cannot be the case, and may cause issues if in the wrong order. For example, applying a caching middleware *before* a compression middleware is likely not a good idea. 
However, since Litestar allows layering middlewares, it can be difficult to keep track of them all across more complex applications.

Another example would be two instrumentation middlewares, e.g. sentry and open telemetry, installed on the same app. They both *want* to be in the first spot, and both have their own monkey-patching utilities to try to achieve this. Currently, there is no guarantee about, or mechanisms to check what happens when both are installed.

Middleware constraints solve this, by allowing the middlewares themselves to express known constraints.

## Why not?

- When combining multiple constraints across many middlewares in a larger application, the resulting rules can become fairly complex. While the errors try to be helpful, it might still prove itself to be a detriment for the maintenance of an application
- It promotes the anti-pattern of not making middlewares agnostic regarding the rest of the stack. Ideally, middlewares should not care about the order they're applied in, and while for some this principle necessarily has to be broken, one could argue that this does not warrant a dedicated feature. The alternative then would be that users have to be careful and requirements about order clearly spelled out in the documentation

## Why not apply ordering constraints instead of just reporting violations?

While it would be possible to do this, there are some major downsides:

1. There may be multiple ways to order the middleware stack according to the constraints, making the order (for the user) non-deterministic
3. Too magical and not very explicit. The user doesn't have a lot of control over what's happening, and they do not get a final say in it
4. It breaks the "layered parameter" approach, by allowing e.g. a middleware defined on a handler to come before a middleware defined on the application. This goes against how all other layered parameters work, in that they either overwrite one another, or are additive and retain their order across layers

<hr>

Documentation and some cleanup is missing, I'm mainly looking for some early feedback.


litestar/_asgi/routing_trie/mapping.py
```diff
@@ -214,6 +214,34 @@ def build_route_middleware_stack(
...
+        # due to the way we're traversing over the app layers, the middleware stack is
+        # constructed in 'application > handler' order, which is the order we want the
+        # middleware to be called in.
+        #
+        # using this order however, since each middleware wraps the next callable, the
+        # *first* middleware in the stack would up being the *innermost* wrapper, i.e.
+        # the last one to receive the request and the first one to see the response.
+        #
+        # to achieve the intended call order, we perform the wrapping in reverse
+        # ('handler -> application').
```
euri10, line 226: i love this

euri10: you're a f*** ascii artist man

litestar/handlers/base.py
```diff
@@ -214,15 +215,6 @@ def _get_merge_opts(self, others: tuple[Router, ...]) -> dict[str, Any]:
...
 
-        # due to the way we're traversing over the app layers, the middleware stack is
-        # constructed in the wrong order (handler > application). reversing the order
-        # here is easier than handling it correctly at every intermediary step.
-        #
-        # we only call this if 'others' is non-empty, to ensure we don't change anything
-        # if no layers have been merged (happens in '._with_changes' for example)
-        if others:
-            merge_opts["middleware"] = tuple(reversed(merge_opts["middleware"]))
-
```
euri10, line 225: just for my own comprehension, is this removal "replaced" by the  `check_middleware_constraints(self.middleware)` addition below ?

tests/unit/test_middleware/test_constraints.py
```diff
@@ -0,0 +1,365 @@
...
+    with pytest.raises(ConstraintViolationError, match="MiddlewareTwo.*after.*MiddlewareOne"):
+        check_middleware_constraints(middlewares)
+
+
+def test_order_string_ref() -> None:
+    one = MiddlewareOne()
+    two = MiddlewareTwo()
+    three = MiddlewareThree()
+
+    two.constraints = MiddlewareConstraints().apply_before(fully_qualified_name(MiddlewareOne))
```
euri10, line 90: im not too sure I get the meaning of apply_before vs before.
```
two.constraints = MiddlewareConstraints(before=(MiddlewareOne,))
two.constraints = MiddlewareConstraints().apply_before(fully_qualified_name(MiddlewareOne))
```

tests/unit/test_middleware/test_constraints.py
```diff
@@ -0,0 +1,365 @@
...
+        check_middleware_constraints((one,))
+
+
+def test_first() -> None:
+    one = MiddlewareOne()
+    two = MiddlewareTwo()
+    one.constraints = MiddlewareConstraints(first=True)
+
+    with pytest.raises(
+        ConstraintViolationError, match="MiddlewareOne.*must be at the top of the stack. Found at index 1"
```
euri10, line 259: I'd rather see `first` instead of `top of the stack` but I guess it's just a matter of taste

euri10 (approved):
lgtm, I have just a few questions and some really minor comments,

I didnt check in detail the depth first search nor other logic and only read the tests. But I think it's ok since if tests pass then the logic should be good.

litestar/middleware/constraints.py
```diff
@@ -0,0 +1,328 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class MiddlewareConstraints:
+    before: tuple[MiddlewareForwardRef | Middleware | MiddlewareFactory, ...] = ()
+    after: tuple[MiddlewareForwardRef | Middleware | MiddlewareFactory, ...] = ()
+    first: bool | None = None
+    last: bool | None = None
+    unique: bool | None = None
+
+    def __post_init__(self) -> None:
```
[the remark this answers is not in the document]
provinzkraut: Why do you think it should acceptable? My reason for disallowing this outright was that it would give confusing error messages if e.g. you've accidentally set mutually exclusive options. With this, we can give informative ones.

tests/unit/test_middleware/test_constraints.py
provinzkraut: It's just a convenience method that allows other options, such as passing in string references and `ignore_not_found`

litestar/middleware/constraints.py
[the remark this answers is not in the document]
euri10: im just saying that I could decide to write a middleware whose position is last/first and unique at the same time, that doesnt look illogical to me, 
maybe im tripping and it is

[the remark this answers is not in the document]
euri10: this shouldnt be a blocker though, either im wrong and it is illogical either the use case is weird to ay the least anyways

tests/unit/test_middleware/test_constraints.py
euri10: mmm ok that's smart

litestar/middleware/constraints.py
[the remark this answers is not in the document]
provinzkraut: Well, that would definitely be a weird use case I'd say but.. Wouldn't this simply translate to "only this middleware"? 

`first=True` and `last=True` both imply `unique=True´, so `first=True` and `last=True` can only be satisfied if the position of the first and last middleware is the same so.. If it's the only one.

Okay so I agree that it's not illogical. Just nonsensical maybe? :sweat_smile: 

I guess whether or not we want to allow this comes down to:

1. Do we hold want to support this edge case in favour of silently allowing the (presumably) more common case of a misconfiguration?
2. Do we want to allow expressing this implicit length constraint on the middleware stack at all? If we want to explicitly support it, I'd say that having a separate constraint for that would be appropriate, but it doesn't fit the "per-middleware constraint" approach, as a "this handler can only have one middleware applied" constraint is very different from "this middleware can only appear once"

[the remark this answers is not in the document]
euri10: nonsensical is the best description, let's keep it this way ;)

PR #4170, opened 2025-05-12
fix: Default options handler when multiple methods are set

### LINKED ISSUE #4169: Bug: Default Options handler does not work correctly on paths with multiple methods
### Description

When multiple handlers are attached to the same path, the options handler will only report the first attached method.

_No response_

### MCVE

```python
from litestar import Litestar, get, patch


@get("/")
async def fn_get() -> None:
    return None


@patch("/")
async def fn_patch() -> None:
    return None


app = Litestar(
    route_handlers=[
        fn_get,
        fn_patch,
    ]
)
```

### Steps to reproduce

1. Run the Litestar app
2. curl -X OPTIONS http://127.0.0.1:8000/ -i
3. Observe the allow field, it will show either OPTIONS, PATCH or GET but not all 3.

### Screenshots

Image

### Logs

```text

```

### Litestar Version

2.16.0

### Platform

- [x] Linux
- [ ] Mac
- [x] Windows

PokkaKiyo:
## Description

- Add a new _DefaultHTTPOptionsRouteHandler for identifying the default options handler added by litestar
- If this handler is already defined in the route_handlers at the time, recreate it to ensure it references the latest router

## Fixes
#4169


PokkaKiyo:
I wonder if we could defer the creation of the options handler until all the routes are registered? Not quite familiar enough with the codebase to know where to do that though. 

This would probably solve another thing with custom option handlers...

provinzkraut:
> I wonder if we could defer the creation of the options handler until all the routes are registered? Not quite familiar enough with the codebase to know where to do that though.
> 
> This would probably solve another thing with custom option handlers...

That's what we do in v3 (sort of). Have you checked if the bug exists in v3 (you can check out the `v3` branch)? I'm not entirely sure, but I think it *might* incidentally be fixed there.

PokkaKiyo:
> > I wonder if we could defer the creation of the options handler until all the routes are registered? Not quite familiar enough with the codebase to know where to do that though.
> > This would probably solve another thing with custom option handlers...
> 
> That's what we do in v3. Have you checked if the bug exists in v3 (you can check out the `v3` branch)? I'm not entirely sure, but I think it _might_ incidentally be fixed there.

haha yep it's working as expected in v3.0!
it also solves the issue where a user's custom OPTIONS route must be the first defined route otherwise it will create multiple OPTIONS routes which will raise a unique-ness error later, great!

PokkaKiyo:
👍 Agreed

provinzkraut:
@PokkaKiyo Your tests are still valuable, as we seem to be missing this case so far. Would you mind re-opening a PR with just the added tests against the v3 branch?

Issue #4173, opened 2025-05-14
Enhancement: Exclude parameter from schema


JCHacking:
### Summary

Allow to exclude parameters from the autogenerated openapi

### Basic Example

```python
@get("/example")
def example(parameter: Annotation[str, Parameter(include_in_schema=False)]) -> None:
    return
```

### Drawbacks and Impact

It should not break anything if everything is included in openapi by default.

provinzkraut:
Closed in #4177

PR #4177, opened 2025-05-19
feat(openapi): add option to exclude parameter from schema

mmev:
## Description

I've added additional argument to `KwargsDefinition` that allows excluding a parameter from OpenAPI schema.
This change is backward-compatible and I've also verified that it should merge smoothly with v3.

## Closes
- litestar-org/litestar#4173

### DIFF
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -124,7 +124,7 @@ class KwargDefinition:
     Use as the key for the reference when creating a component for this type
     .. versionadded:: 2.12.0
     """
-    include_in_schema: bool = field(default=True)
+    include_in_schema: bool = True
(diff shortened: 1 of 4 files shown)


litestar/params.py
```diff
@@ -124,6 +124,8 @@ class KwargDefinition:
     Use as the key for the reference when creating a component for this type
     .. versionadded:: 2.12.0
     """
+    include_in_schema: bool = field(default=True)
```
provinzkraut, line 127:
```suggestion
    include_in_schema: bool = True
```
is enough :)

provinzkraut (approved):
Thanks @mmev!

PR #4182, opened 2025-05-26
Drop uvloop from direct dependencies

gsakkis:
## Description

`uvloop` is already a (conditional) dependency of `uvicorn[standard]` with a more constrained condition (excluded from cygwin and PyPy).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ Twitter = "https://twitter.com/LitestarAPI"
 annotated-types = ["annotated-types"]
 attrs = ["attrs"]
 brotli = ["brotli"]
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Yes, this looks correct: https://github.com/encode/uvicorn/blob/fa3d9d27dbe7ecbc68e0992309d0ba49ead5640b/pyproject.toml#L44

Please, update the lock file.

provinzkraut:
Closed because #4183 is no longer relevant

gsakkis:
#4183 was an irrelevant issue I came across trying to update the lock file; the PR is about removing the explicit uvloop dependency which is [still specified](https://github.com/litestar-org/litestar/blob/main/pyproject.toml#L59).

PR #4187, opened 2025-05-29
feat: allow weak references to Router and Litestar instances (#4186)

### LINKED ISSUE #4186: Enhancement: Allow weakrefs to be created to Litestar instances
### Summary

The Python standard library has facilities for creating weak references to objects, allowing them to be accessed without prohibiting garbage collection.

To be compatible with this facility, when a type defines `__slots__`, it must include `__weakref__` in the list of permitted slots. This is not presently the case for Litestar's types.

### Basic Example

As an example of a facility where weakrefs have value, consider a test suite wherein a variety of differently-configured application instances may be created, where one wishes to have cached httpx clients to use to invoke those instances. A `WeakKeyDictionary[Litestar, httpx.AsyncClient]` would be an appropriate tool to store clients for these applications, automatically deleting any client should an instance it's attached to be garbage collected, _if_ weakrefs could be created for Litestar objects.

Concretely:

```python
from weakref import WeakKeyDictionary
import httpx
import litestar

client_cache: WeakKeyDictionary[litestar.Litestar, httpx.AsyncClient] = WeakKeyDictionary()

def client_for_app(app: litestar.Litestar) -> httpx.AsyncClient:
    if (retval := client_cache.get(app, default=None)) is not None:
        return retval
    transport = httpx.ASGITransport(app=app)
    client_cache[app] = retval = httpx.AsyncClient(transport=transport)
    return retval
```

This is not currently possible, because `weakref.WeakKeyDictionary()[litestar.Litestar()] = None` fails with `TypeError: cannot create weak reference to 'Litestar' object`. Amending either `Litestar.__slots__` or `Router.__slots__` to contain `'__weakref__'` would resolve this.

### Drawbacks and Impact

Slightly increases memory usage. Would break compatibility with any preexisting code that already subclasses `Litestar` and adds a `__weakref__` slot without checking whether that slot already exists.

### Unresolved questions

_No response_

charles-dyfis-net:
## Description

Adds `__weakref__` slot to `Router`, and thus also to `Litestar`, allowing weak references to be created for objects of these types.

## Closes

Closes #4186.

### DIFF
--- a/litestar/router.py
+++ b/litestar/router.py
@@ -51,6 +51,7 @@ class Router:
     """
 
     __slots__ = (
+        "__weakref__",
         "after_request",
         "after_response",
         "before_request",


provinzkraut (changes requested):
As discussed in #4186

1. This would be considered a breaking change, so it can only happen in 3.0
2. We should just remove the `__slots__` as they don't provide real benefits on the app instance

PR #4189, opened 2025-05-31
docs: add `uv` examples to the docs

sobolevn:
Notes:
- I used `poetry` instead of `Poetry` for consistency with other names
- I added `'` where needed, because `python3 -m pip install litestar[cryptography]` generates `zsh: no matches found: litestar[cryptography]` on `bash` and `zsh` (not sure about others)
- `uv` examples are added, since `uv` is really popular now


sobolevn:
https://litestar-org.github.io/litestar-docs-preview/4189/usage/cli.html

Снимок экрана 2025-05-31 в 16 01 48

Looks nice :)

PR #4192, opened 2025-05-31
feat(jwt): [#4191] extend Token.encode() to support custom headers

### LINKED ISSUE #4191: Enhancement: Allow passing custom headers to Token.encode() for improved JWT flexibility
### Summary

The `Token.encode()` method in `litestar.security.jwt.Token` does not allow passing additional parameters supported by the `pyjwt` library, such as **headers**. This limits flexibility when generating JWTs in projects that require including extra data in the token header (e.g., kid for selecting the correct public key during validation on the client or in another service).
`Token.encode()` internally uses `pyjwt.encode()`, but its current signature does not allow forwarding these additional arguments.
I propose to enhance the `Token.encode()` method to allow providing custom headers.

### Basic Example

For example, in my project, token generation requires explicitly setting the **kid** in the header so that consumers can select the appropriate public key. 
I would prefer to use the Token class provided by the framework (`litestar.security.jwt.Token`) instead of implementing a custom token entity, as the existing functionality already satisfies most of my needs — except for header injection.

```python
from datetime import datetime, timedelta
from litestar.security.jwt import Token

token = Token(exp=datetime.now() + timedelta(hours=1), sub="user-id")

# Custom headers, e.g., setting 'kid' for key identification
custom_headers = {"kid": "my-key-id"}

encoded_token = token.encode(
    secret="my-secret",
    algorithm="HS256",
    headers=custom_headers,
)
```
```python
# Result: we can see that the custom header has been included
print(jwt.get_unverified_header(encoded_token))
{'alg': 'HS256', 'kid': 'my-key-id', 'typ': 'JWT'}
```
This allows consumers (e.g., client apps or services) to read the **kid** from the token header and select the appropriate public key for validation.

### Drawbacks and Impact

The change is backward-compatible and introduces minimal complexity. It improves flexibility without adding maintenance burden, as it only forwards arguments to `pyjwt.encode()`.

### Unresolved questions

_No response_

s-aleshin:
## Description
Allows passing custom JWT headers (e.g., kid) to Token.encode(). This improves flexibility in cases where custom JWT headers (like `kid`) are required.
-

## Closes
Fixes #4191

### DIFF
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -90,7 +90,7 @@ class Token:
     def decode_payload(
         cls,
         encoded_token: str,
-        secret: str,
+        secret: str | bytes,
         algorithms: list[str],
         issuer: list[str] | None = None,
         audience: str | Sequence[str] | None = None,
@@ -110,7 +110,7 @@ class Token:
     def decode(
         cls,
         encoded_token: str,
-        secret: str,
+        secret: str | bytes,
         algorithm: str,
         audience: str | Sequence[str] | None = None,
         issuer: str | Sequence[str] | None = None,
@@ -194,12 +194,18 @@ class Token:
         ) as e:
             raise NotAuthorizedException("Invalid token") from e
 
-    def encode(self, secret: str, algorithm: str) -> str:
+    def encode(
+        self,
+        secret: str | bytes,
+        algorithm: str,
+        headers: dict[str, Any] | None = None,
+    ) -> str:
         """Encode the token instance into a string.
 
         Args:
             secret: The secret with which the JWT is encoded.
             algorithm: The algorithm used to encode the JWT.
+            headers: Optional headers to include in the JWT (e.g., {"kid": "..."}).
 
         Returns:
             An encoded token string.
@@ -212,6 +218,7 @@ class Token:
                 payload={k: v for k, v in asdict(self).items() if v is not None},
                 key=secret,
                 algorithm=algorithm,
+                headers=headers,
             )
         except (jwt.DecodeError, NotImplementedError) as e:
             raise ImproperlyConfiguredException("Failed to encode token") from e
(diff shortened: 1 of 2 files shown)


provinzkraut (changes requested):
Two things:

1. This is a breaking change (for users that subclass `Token`), so it would have to be targeted at the `v3` branch
2. I don't think we should add `json_encoder` and `sort_headers`, as they are specific to `pyjwt`, and this is meant as an abstraction *over* `pyjwt`, not a drop-in replacement. If you do need to customise how `decode` interacts with `pyjwt`, you can simply subclass `Token`

s-aleshin:
> Two things:
> 
> 1. This is a breaking change (for users that subclass `Token`), so it would have to be targeted at the `v3` branch
> 2. I don't think we should add `json_encoder` and `sort_headers`, as they are specific to `pyjwt`, and this is meant as an abstraction _over_ `pyjwt`, not a drop-in replacement. If you do need to customise how `decode` interacts with `pyjwt`, you can simply subclass `Token`

1. I’ve rebased the branch onto v3.0 as you suggested — the PR now targets the correct base.
2. Ok, I agree. `json_encoder` and `sort_headers` are specific for `pyjwt` and use rarely. Deleted from the PR.

litestar/security/jwt/token.py
```diff
@@ -194,12 +195,18 @@ def decode(
         ) as e:
             raise NotAuthorizedException("Invalid token") from e
 
-    def encode(self, secret: str, algorithm: str) -> str:
+    def encode(
+        self,
+        secret: AllowedPrivateKeys | str | bytes,
```
provinzkraut, line 200: `str | bytes` is fine but we don't want to expose the pyjwt internals (`AllowedPrivateKeys`) here

s-aleshin: Got it, fixed.
Additionally, I updated the type of the secret parameter in two other methods of the Token class (`decode` and `decode_payload`) to ensure consistency across the API and avoid type mismatches during subclassing or static analysis.

provinzkraut (approved):
Thanks @s-aleshin!

PR #4209, opened 2025-06-16
Documentation - Fix non-existing function in "Accessing the OpenAPI schema in code" section

ctrl-Felix:
This is really a very very small PR (a one-liner, lol). But since I stumvled upon it I thought I'd provide the PR to fix it.

The docs currently call .dict() but that function is not existing. It is to_schema now. 

https://docs.litestar.dev/latest/reference/openapi/spec.html#litestar.openapi.spec.BaseSchemaObject


provinzkraut:
Thanks!

sobolevn:
I think we should consider removing coverage from "Required", because it stucks with no clear way of restarting it :(

PR #4218, opened 2025-06-19
feat: Generate correct OpenAPI type and format for pydantic.AwareDate…

### LINKED ISSUE #4217: Enhancement: Generate correct OpenAPI type and format for `pydantic.AwareDateTime`
### Summary

Litestar doesn't generate OpenAPI type and format for `pydantic.AwareDatetime`

### Basic Example

```python
import datetime

import litestar
import pydantic
from litestar.openapi.spec import OpenAPIFormat, OpenAPIType


class HelloResponse(pydantic.BaseModel):
    stdlib_datetime: datetime.datetime
    aware_datetime: pydantic.AwareDatetime


@litestar.get()
async def hello() -> HelloResponse:
    return HelloResponse(
        stdlib_datetime=datetime.datetime.now(),
        aware_datetime=datetime.datetime.now(datetime.UTC),
    )


app = litestar.Litestar([hello])
props = app.openapi_schema.components.schemas["HelloResponse"].properties

assert props["stdlib_datetime"].type == OpenAPIType.STRING
assert props["stdlib_datetime"].format == OpenAPIFormat.DATE_TIME

assert props["aware_datetime"].type is None  # Expected OpenAPIType.STRING
assert props["aware_datetime"].format is None  # Expected OpenAPIFormat.DATE_TIME

print("Assertions passed, aware datetime is not supported")
```

Run with uv:

```sh
uv run --python 3.13 --with litestar==2.16.0 --with pydantic==2.11.7 --no-project t.py
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

Kumzy:
Should fix #4217

It also includes other types listed there: https://docs.pydantic.dev/2.0/usage/types/datetime/

Had to modify the test serializer as these types of date are not in v1 of pydantic

### DIFF
--- a/tests/unit/test_plugins/test_pydantic/test_plugin_serialization.py
+++ b/tests/unit/test_plugins/test_pydantic/test_plugin_serialization.py
@@ -27,6 +27,12 @@ from litestar.serialization import (
 from . import PydanticVersion
 
 TODAY = datetime.date.today()
+YESTERDAY_DATE = datetime.date.today() - datetime.timedelta(days=1)
+TOMORROW_DATE = datetime.date.today() + datetime.timedelta(days=1)
+YESTERDAY_DATETIME = datetime.datetime.now() - datetime.timedelta(days=1)
+TOMORROW_DATETIME = datetime.datetime.now() + datetime.timedelta(days=1)
+AWARE_DATETIME = datetime.datetime.now(datetime.timezone.utc)
+NAIVE_DATETIME = datetime.datetime.now().replace(tzinfo=None)
 
 
 class CustomStr(str):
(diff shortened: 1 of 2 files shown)


tests/unit/test_plugins/test_pydantic/test_plugin_serialization.py
```diff
@@ -198,9 +217,26 @@ def model(pydantic_version: PydanticVersion) -> ModelV1 | ModelV2:
         ("conint", 1),
         ("url", "some://example.org/"),
         ("http_url", "http://example.org/"),
+        ("paste_date", YESTERDAY_DATE.isoformat()),
```
provinzkraut, line 220: You'll need to make these values constant, otherwise xdist will trip up because they're different on the different workers

Kumzy: Yes that's what I am trying to do at the moment, looking how to fix that

Kumzy:
Converted only the datetime to constants, kept the date dynamically generated but can change that too.

provinzkraut (approved):

PR #4222, opened 2025-06-23
feat: Add CLI `--env-file` option

### LINKED ISSUE #3217: rfc: remove or allow `dotenv` configuration in `from_env`
https://github.com/cofin/litestar/blob/af24784bbb3d584226ba39c323b021701fbb0678/litestar/cli/_utils.py#L105-L108

### LINKED ISSUE #4211: Enhancement: Command-line argument to load extra .env files
### Summary

I have a situation where I would like to load multiple .env files when the application starts, but as it stands now, litestar does not provide such a feature.

So far, I've solved the problem for myself by creating a [CLIPlugin](https://docs.litestar.dev/2/usage/cli.html#using-a-plugin):

```python
    """The purpose of this plugin is to add the --env-file option to the run command"""

    @override
    def on_cli_init(self, cli):
        from click import Option

        run_cmd = cli.commands.get("run")
        run_cmd.params.append(
            Option(
                ["--env-file", "env_files"], multiple=True, type=Path, help="Load environment variables from file(s)"
            )
        )

        original_callback = run_cmd.callback

        def wrapped_callback(*args: Any, env_files: tuple[Path] | None = (), **kwargs: Any):
            import dotenv

            from app.log import logger

            for env_file in env_files:
                if dotenv.load_dotenv(dotenv_path=env_file, override=True):
                    logger.info(f"Loaded environment variables from {env_file.absolute()}")
            return original_callback(*args, **kwargs)

        run_cmd.callback = wrapped_callback
```

However, I thimk it would be nice to have this feature as part of litestar.

### Basic Example

```sh
litestar run --env-file .env --env-file .env.prod
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

sakosha:
## Description

Add a `--env-file` flag to allow users to explicitly use .env environment variables:
- Provide custom names/paths for env files.
- Use multiple env files.
- Ignore `.env` files.

## Closes

Closes #3217, #4211

### DIFF
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -88,7 +88,9 @@ class LitestarEnv:
     is_app_factory: bool = False
 
     @classmethod
-    def from_env(cls, app_path: str | None, app_dir: Path | None = None, env_file: Path | None = None) -> LitestarEnv:
+    def from_env(
+        cls, app_path: str | None, app_dir: Path | None = None, env_files: tuple[Path, ...] | None = None
+    ) -> LitestarEnv:
         """Load environment variables.
 
         If ``python-dotenv`` is installed, use it to populate environment first
@@ -98,15 +100,18 @@ class LitestarEnv:
         if cwd_str_path not in sys.path:
             sys.path.append(cwd_str_path)
 
-        if env_file is not None and env_file.is_file():
+        if env_files:
             try:
                 import dotenv
-
-                dotenv.load_dotenv(env_file)
             except ImportError as err:
                 raise LitestarCLIException(
                     "Python-dotenv must be installed when using --env-file\nPlease install the `python-dotenv`"
                 ) from err
+            for env_file in env_files:
+                if env_file.is_dir():
+                    raise LitestarCLIException(f"{env_file} is a directory.\nPlease provide a file to --env-file")
+
+                dotenv.load_dotenv(env_file)
 
         app_path = app_path or getenv("LITESTAR_APP")
         app_name = getenv("LITESTAR_APP_NAME") or "Litestar"
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -627,7 +627,9 @@ def isatty() -> bool:
     return sys.stdout.isatty()
 
 
-def validate_mutually_exclusive_env_options(ctx: click.Context, param: click.Option, value: Sequence[Path]) -> Sequence[Path]:
+def validate_mutually_exclusive_env_options(
+    ctx: click.Context, param: click.Option, value: Sequence[Path]
+) -> Sequence[Path]:
     if ctx.params.get("env_files") and ctx.params.get("ignore_env_files"):
         raise click.BadParameter("The options '--env-file' and '--ignore-dotenv' are mutually exclusive.")
     return value
--- a/litestar/cli/_utils.py
+++ b/litestar/cli/_utils.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import contextlib
 import importlib
 import inspect
 import os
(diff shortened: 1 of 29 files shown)


litestar/cli/_utils.py
```diff
@@ -99,10 +98,16 @@ def from_env(cls, app_path: str | None, app_dir: Path | None = None) -> Litestar
...
-            import dotenv
+        if env_file is not None and env_file.is_file():
+            try:
+                import dotenv
+
+                dotenv.load_dotenv(env_file)
+            except ImportError as err:
+                raise LitestarCLIException(
+                    "Python-dotenv must be installed when using --env-file\nPlease install the `python-dotenv`"
+                ) from err
```
sobolevn:
```suggestion
            try:
                import dotenv
            except ImportError as err:
                raise LitestarCLIException(
                    "Python-dotenv must be installed when using `--env-file`\nPlease install the `python-dotenv` package"
                ) from err
            else:
                dotenv.load_dotenv(env_file)
```

pyproject.toml
```diff
@@ -63,7 +63,7 @@ maintainers = [
 ]
 name = "litestar"
 readme = "README.md"
-requires-python = ">=3.8,<4.0"
+requires-python = ">=3.9,<4.0"
```
sobolevn: This is a breaking change.

litestar/cli/main.py
```diff
@@ -24,8 +24,17 @@
     type=ClickPath(dir_okay=True, file_okay=False, path_type=Path),
     show_default=False,
 )
+@click.option(
+    "--env-file",
+    help="Path to a .env file to load environment variables from. If not provided, environment variables won't be loaded from any .env file.",
```
sobolevn: This is a breaking change. If no explicit `--env-file` is provided, then all existing apps will break, because no file will be loaded. This is not right.

sakosha: I think we can leave `.env` as the default, but how should we allow users to opt-out of loading .env file?

litestar/cli/_utils.py
sakosha: I changed it up to enable multiple `.env` files. Do you think we should still put the iteration into `else` block?

pyproject.toml
sakosha: fixed!

litestar/cli/main.py
sobolevn: Yes, opting out is a good idea

litestar/cli/_utils.py
```diff
@@ -99,10 +100,19 @@ def from_env(cls, app_path: str | None, app_dir: Path | None = None) -> Litestar
         if cwd_str_path not in sys.path:
             sys.path.append(cwd_str_path)
 
-        with contextlib.suppress(ImportError):
-            import dotenv
+        if env_files:
```
provinzkraut: This changes the behaviour in a breaking way. Previously, if `dotenv` was installed, we'd implicitly load the `.env` file, and do nothing if it wasn't installed. That behaviour should be kept as-is, to not break stuff

litestar/cli/_utils.py
```diff
@@ -99,10 +100,19 @@ def from_env(cls, app_path: str | None, app_dir: Path | None = None) -> Litestar
...
+            try:
+                import dotenv
+            except ImportError as err:
+                raise LitestarCLIException(
+                    "Python-dotenv must be installed when using --env-file\nPlease install the `python-dotenv`"
+                ) from err
+            for env_file in env_files:
+                if env_file.is_dir():
+                    raise LitestarCLIException(f"{env_file} is a directory.\nPlease provide a file to --env-file")
```
provinzkraut: I don't think we should should check that, and instead just let the errors from `dotenv` bubble up

sakosha:
I will write a new PR to target v3.0 branch.

PR #4239, opened 2025-07-16
fix: sort openapi type schema

### LINKED ISSUE #3646: Bug: order of types in openapi spec is not consistent in json rendering
### Description

We are seeing the order of types change in openapi generation, which makes comparing golden versions of the openapi spec problematic.  I think the specific problem we are seeing comes from https://github.com/litestar-org/litestar/blob/ffaf5616b19f6f0f4128209c8b49dbcb41568aa2/litestar/_openapi/schema_generation/schema.py#L160 where we use the `set` operation to uniquify the list of types.  The order doesn't matter to the correctness of the openapi spec, so perhaps the responsibility for ensuring a determistic spec file could also come from the serializer, but either way it would be helpful if we could always render the same openapi spec the same way.

_No response_

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.9.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

tjhoff:
## Description

- Sorts orders of types when using `Literal[*] | None`

Consider the following OpenAPI schema type:
`{'type': ['null', 'string'], 'enum': ['1', None]}`

Currently, there are scenarios (with environmental differences) in which this can instead be rendered as:
`{'type': ['string', 'null'], 'enum': ['1', None]}`

This can cause issues with continuous integration in monorepos where generated openapi json is checked into the source tree and validated for changes. 

Another discussion that this might spawn is why this is not in the OAPI 3.0 `oneOf` format instead - 
`{'oneOf': [{'type': 'string', 'const': '1'}, {'type': 'null'}]}`

This appears to be consistently sorted by something else in Litestar's openapi generation.

### DIFF
--- a/litestar/middleware/base.py
+++ b/litestar/middleware/base.py
@@ -190,7 +190,7 @@ class ASGIMiddleware(abc.ABC):
 
         class MyMiddleware(ASGIMiddleware):
             scopes = (ScopeType.HTTP,)
-            exclude = ("/not/this/path",)
+            exclude_path_pattern = ("/not/this/path",)
             exclude_opt_key = "exclude_my_middleware"
 
             def __init__(self, my_logger: Logger) -> None:


tjhoff:
@provinzkraut any chance you could take a look at this as you responded to the initial issue? Sorry for the tag!

provinzkraut (approved):
Looks good. I like the idea you suggested of using `oneOf` instead though. Would you want to give implementing that a try as well?

tjhoff:
Thanks! I'll take a look and see if it's straightforward.

tjhoff:
After some digging, this seems like it was explicit behavior added [here](https://github.com/litestar-org/litestar/pull/2550). I'm not quite clear on what the ramifications of changing this back would be; however, in my testing I did note that `Literal | <anything> | null` results in a `oneOf` type. 

I might leave this alone to prevent causing a regression to someone else! 😅

PR #4241, opened 2025-07-20
fix(jwt): Relax typing to allow `Sequence` for `Token.aud` 

### LINKED ISSUE #4240: Bug: JWT aud claim decoding fails when array is provided
## Describing the bug

When decoding a JWT that includes an `aud` (audience) claim or verification audience as an array (e.g. `"aud": ["service-a", "service-b"]`),:
```python
decoded = Token.decode(
    token_payload,
    pem_key,
    audience=["service-a", "service-b"],
    algorithm=jwt_algorithm,
    strict_audience=False,
)
```
the decoding fails with the following error
```
ValidationError(‘Expected str | null, got array - at $.aud’)
```

## Expected behavior

According to [RFC 7519 Section 4.1.3](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3), the `aud` claim **may be either a single string or an array of strings**.

The decoder (likely in `mspspec` or the JWT handling schema) should accept:
- `"aud": "some-audience"`
- `"aud": ["aud1", "aud2"]`

_No response_

### MCVE

```python
from litestar.security.jwt import Token

decoded = Token.decode(
    token_payload,
    pem_key,
    audience=["service-a", "service-b"],
    algorithm=jwt_algorithm,
    strict_audience=False,
)
```

### Steps to reproduce

Create a valid JWT and pem_key and run the snippet

### Screenshots

_No response_

### Logs

```text
cls = <class 'litestar.security.jwt.token.Token'>
encoded_token = '...MyToken...'
secret = '-----BEGIN PUBLIC KEY-----\n...MyPublicKey...\n-----END PUBLIC KEY-----\n'
algorithm = 'RS256', audience = ['my-realm'], issuer = None
require_claims = None, verify_exp = True, verify_nbf = True
strict_audience = False

    @classmethod
    def decode(
        cls,
        encoded_token: str,
        secret: str,
        algorithm: str,
        audience: str | Sequence[str] | None = None,
        issuer: str | Sequence[str] | None = None,
        require_claims: Sequence[str] | None = None,
        verify_exp: bool = True,
        verify_nbf: bool = True,
        strict_audience: bool = False,
    ) -> Self:
        """Decode a passed in token string and return a Token instance.
    
        Args:
            encoded_token: A base64 string containing an encoded JWT.
            secret: The secret with which the JWT is encoded.
            algorithm: The algorithm used to encode the JWT.
            audience: Verify the audience when decoding the token. If the audience in
                the token does not match any audience given, raise a
                :exc:`NotAuthorizedException`
            issuer: Verify the issuer when decoding the token. If the issuer in the
                token does not match any issuer given, raise a
                :exc:`NotAuthorizedException`
            require_claims: Verify that the given claims are present in the token
            verify_exp: Verify that the value of the ``exp`` (*expiration*) claim is in
                the future
            verify_nbf: Verify that the value of the ``nbf`` (*not before*) claim is in
                the past
            strict_audience: Verify that the value of the ``aud`` (*audience*) claim is
                a single value, and not a list of values, and matches ``audience``
                exactly. Requires the value passed to the ``audience`` to be a sequence
                of length 1
    
        Returns:
            A decoded Token instance.
    
        Raises:
            NotAuthorizedException: If the token is invalid.
        """
    
        options: JWTDecodeOptions = {
            "verify_aud": bool(audience),
            "verify_iss": bool(issuer),
        }
        if require_claims:
            options["require"] = list(require_claims)
        if verify_exp is False:
            options["verify_exp"] = False
        if verify_nbf is False:
            options["verify_nbf"] = False
        if strict_audience:
            if audience is None or (not isinstance(audience, str) and len(audience) != 1):
                raise ValueError("When using 'strict_audience=True', 'audience' must be a sequence of length 1")
            options["strict_aud"] = True
            # although not documented, pyjwt requires audience to be a string if
            # using the strict_aud option
            if not isinstance(audience, str):
                audience = audience[0]
    
        try:
            payload = cls.decode_payload(
                encoded_token=encoded_token,
                secret=secret,
                algorithms=[algorithm],
                audience=audience,
                issuer=list(issuer) if issuer else None,
                options=options,
            )
            # msgspec can do these conversions as well, but to keep backwards
            # compatibility, we do it ourselves, since the datetime parsing works a
            # little bit different there
            payload["exp"] = datetime.fromtimestamp(payload["exp"], tz=timezone.utc)
            payload["iat"] = datetime.fromtimestamp(payload["iat"], tz=timezone.utc)
            extra_fields = payload.keys() - {f.name for f in dataclasses.fields(cls)}
            extras = payload.setdefault("extras", {})
            for key in extra_fields:
                extras[key] = payload.pop(key)
            return msgspec.convert(payload, cls, strict=False)
        except (
            KeyError,
            jwt.exceptions.InvalidTokenError,
            ImproperlyConfiguredException,
            msgspec.ValidationError,
        ) as e:
>           raise NotAuthorizedException("Invalid token") from e
E           litestar.exceptions.http_exceptions.NotAuthorizedException: 401: Invalid token

.venv/lib/python3.12/site-packages/litestar/security/jwt/token.py:193: NotAuthorizedException
```

### Litestar Version

Version: 2.16.0

### Platform

- [x] Linux
- [x] Mac
- [ ] Windows

casabre:
## Description

Adding Sequence support for audience in JWT Token.

## Closes
#4240


igorcoding:
Hi! Is there any chance that this would be merged soon? Our team is currently experiencing an issue with the `aud` being an array. Or Maybe you know of any workaround to mitigate this issue without patching source code?

Harshal6927:
Hello @casabre , thanks for the PR! Could you please add a test case so we can go ahead and merge it?

tests/unit/test_security/test_jwt/test_token.py
```diff
@@ -18,17 +18,45 @@
 from litestar.security.jwt.token import JWTDecodeOptions
 
 
-@pytest.mark.parametrize("algorithm", ["HS256", "HS384", "HS512"])
-@pytest.mark.parametrize("token_issuer", [None, "e3d7d10edbbc28bfebd8861d39ae7587acde1e1fcefe2cbbec686d235d68f475"])
-@pytest.mark.parametrize("token_audience", [None, "627224198b4245ed91cf8353e4ccdf1650728c7ee92748f55fe1e9a9c4d961df"])
 @pytest.mark.parametrize(
-    "token_unique_jwt_id", [None, "10f5c6967783ddd6bb0c4e8262d7097caeae64705e45f83275e3c32eee5d30f2"]
+    "algorithm",
```
provinzkraut: This doubles the amount of tests run, due to the multiple `parametrize` decorators. Could you maybe just add one simple case for the sequence audience?

casabre:
@Harshal6927 @provinzkraut I added an own test for audience validation. Hope that works for you.

casabre:
@provinzkraut I didn't change the imports but now pytest is complaining about a missing `Sequence` import during runtime. Actually, I don't know if this is related to a pytest version change and if I should fix the imports or leave it as it is...

litestar/security/jwt/token.py
```diff
@@ -59,9 +59,9 @@ class Token:
     """Issued at - should always be current now."""
     iss: str | None = field(default=None)
     """Issuer - optional unique identifier for the issuer."""
-    aud: Union[str, Sequence[str], None] = field(default=None)  # noqa: UP007
+    aud: str | Sequence[str] | None = field(default=None)
     """Audience - intended audience(s)."""
-    jti: Optional[str] = field(default=None)  # noqa: UP007
+    jti: str | None = field(default=None)
```
provinzkraut, line 63: This was intentional. Since this is a public class that has its type annotations used during runtime, they need to be compatible with older Python versions. This union Syntax is only supported from 3.10 onwards

casabre: Yes, fair enough but at least `Optional` should be supported by `from __future__ import annotations`.

But it wasn't the main problem in the pipeline. I added the Union type hint.

casabre:
@provinzkraut In the end, a missing import makes the pytests fail
```
FAILED tests/unit/test_security/test_jwt/test_auth.py::test_jwt_auth - NameError: name 'Sequence' is not defined
```
e.g. at https://github.com/litestar-org/litestar/actions/runs/17854788852/job/50771227842?pr=4241#step:10:9032

litestar/security/jwt/token.py
provinzkraut: > Yes, fair enough but at least `Optional` should be supported by `from __future__ import annotations`.

No, it's not. `from __future__ import annotations` only works for static type checking, since it simply turns all annotations into strings. But if you try to *inspect that annotation at runtime*, it's still going to fail because the language version doesn't support it.

provinzkraut:
> @provinzkraut I didn't change the imports but now pytest is complaining about a missing `Sequence` import during runtime. Actually, I don't know if this is related to a pytest version change and if I should fix the imports or leave it as it is...

Well it's because you are using `Sequence` but don't import it anywhere :wink:

casabre:
> > @provinzkraut I didn't change the imports but now pytest is complaining about a missing `Sequence` import during runtime. Actually, I don't know if this is related to a pytest version change and if I should fix the imports or leave it as it is...
> 
> Well it's because you are using `Sequence` but don't import it anywhere 😉

Yes, you right 😅. But I was relying on the correctness of previous logic --> because `Sequence` was used before my change I thus, I thought, it is covered. E.g here https://github.com/litestar-org/litestar/blob/65b0bd2ea32730a3f5d24f2d6cb4b8f6be6f0df8/litestar/security/jwt/token.py#L116

It's not a problem for me to pull the collections import out of the type checking guard, but I am really wondering how it was working before 🤔

provinzkraut:
> It's not a problem for me to pull the collections import out of the type checking guard, but I am really wondering how it was working before 🤔

The function signature is not validated during runtime. The dataclass itself is.

casabre:
> > It's not a problem for me to pull the collections import out of the type checking guard, but I am really wondering how it was working before 🤔
> 
> The function signature is not validated during runtime. The dataclass itself is.

@provinzkraut Yes, I understand that but why was it failing now but before not, when Sequence was accessed too. Anyway... let't move ahead --> if you are satisfied with the change, can you just approve it 😊.

casabre:
@provinzkraut what do you say to the changes? Can we move forward and merge?

provinzkraut (approved):

PR #4243, opened 2025-07-23
fix: empty string being considered as the token if the token was coming from the cookies

iyad-f:
## Description
This PR fixes the issue where an empty string was being considered as the token when the token was coming from the cookies
-

### DIFF
--- a/litestar/security/jwt/middleware.py
+++ b/litestar/security/jwt/middleware.py
@@ -251,5 +251,6 @@ class JWTCookieAuthenticationMiddleware(JWTAuthenticationMiddleware):
         auth_header = connection.headers.get(self.auth_header) or connection.cookies.get(self.auth_cookie_key)
         if not auth_header:
             raise NotAuthorizedException("No JWT token found in request header or cookies")
-        encoded_token = auth_header.partition(" ")[-1]
+        parts = auth_header.strip().split(maxsplit=1)
+        encoded_token = parts[-1] if parts else ""
         return await self.authenticate_token(encoded_token=encoded_token, connection=connection)


provinzkraut (changes requested):
Thanks @iyad-f!

Could you please provide a test case and a more precise description of the issue this PR aims to fix?

iyad-f:
@provinzkraut basically if the encoded_token comes from the cookies then it will be simply a string and doing a partition will break it into three pieces ['', 'thetoken', ''] and doing [-1] on that will give us the empty string and not the token

First case is, the token comes from the header in the format of Bearer token
```py
auth_header = "Bearer mytoken"
print(auth_header.partition(" ")[-1]) # prints mytoken
 ```
 
 Second case is, the token comes from the cookies directly
 ```py
auth_header = "mytoken"

print(auth_header.partition(" ")[-1]) # prints an empty string 
```

provinzkraut:
Currently tests are failing for your branch, which means existing behaviour would be changed, and you need to add a test case that would fail *without* your fix, to prevent regressions.

iyad-f:
The tests are failing because the same method is used to get the encoded_token in the tests as well, so will have to change that as well

iyad-f:
from what i tested the test seems to fail for no jwt token found, this fix doesnt touch that part though, so not sure how to fix that

provinzkraut:
@iyad-f can you address the failing tests?

provinzkraut:
@iyad-f are you still working on this?

iyad-f:
@provinzkraut sorry didnt see the message, will take a look

disrupted:
could you check if #4552 fixed this issue?

PR #4251, opened 2025-08-07
Enum desc fix

### LINKED ISSUE #4250: Bug: Can't use `Parameter()` with `description` and `schema_extra` to change Enum's OpenAPI schema
### Description

Prior to v2.14.0, you could use `Parameter()` with `description` and `schema_extra` to modify the generated schema for an Enum. This was changed in #3525. Apart from messing with `app.openapi_schema`, it seems like users have to modify the description via the Enum's docstring:

https://github.com/litestar-org/litestar/blob/ad8ad7205c8e07692b3b05b01ab052f091949b0b/litestar/_openapi/schema_generation/schema.py#L595

Note that v2.13.0 did not let you modify the schema's description through the Enum's docstring and I don't see feature being mentioned, so I think this was unintentional.

I'm not aware of a way to modify the Enum's generated schema in v2.16.0 that is as easy as modifying `schema_extra` like you could in v2.13.0.

_No response_

### MCVE

```python
from enum import StrEnum

class TestEnum(StrEnum):
    ONE = "one"
    TWO = "two"

@get("/")
async def foo(q: TestEnum = Parameter(description="desc", schema_extra={"format": "byte"})) -> None: ...

app = Litestar([foo])
```

### Steps to reproduce

Run the MCVE on v2.13.0 and notice that the query parameter's OpenAPI schema has "description" and "format" fields:
```
{
  "info": {
    "title": "Litestar API",
    "version": "1.0.0"
  },
  "openapi": "3.1.0",
  "servers": [
    {
      "url": "/"
    }
  ],
  "paths": {
    "/": {
      "get": {
        "summary": "Foo",
        "operationId": "Foo",
        "parameters": [
          {
            "name": "q",
            "in": "query",
            "schema": {
              "type": "string",
              "enum": [
                "one",
                "two"
              ],
              "format": "byte",
              "description": "desc"
            },
            "description": "desc",
            "required": true,
            "deprecated": false,
            "allowEmptyValue": false,
            "allowReserved": false
          }
        ],
        "responses": {
          "200": {
            "headers": {}
          },
          "400": {
            "description": "Bad request syntax or unsupported method",
            "content": {
              "application/json": {
                "schema": {
                  "properties": {
                    "status_code": {
                      "type": "integer"
                    },
                    "detail": {
                      "type": "string"
                    },
                    "extra": {
                      "additionalProperties": {},
                      "type": [
                        "null",
                        "object",
                        "array"
                      ]
                    }
                  },
                  "type": "object",
                  "required": [
                    "detail",
                    "status_code"
                  ],
                  "description": "Validation Exception",
                  "examples": [
                    {
                      "status_code": 400,
                      "detail": "Bad Request",
                      "extra": {}
                    }
                  ]
                }
              }
            }
          }
        },
        "deprecated": false
      }
    }
  },
  "components": {
    "schemas": {}
  }
}
```

Then, run the MCVE on v2.16.0 and notice that there's no "description" or "format":
```
{
  "info": {
    "title": "Litestar API",
    "version": "1.0.0"
  },
  "openapi": "3.1.0",
  "servers": [
    {
      "url": "/"
    }
  ],
  "paths": {
    "/": {
      "get": {
        "summary": "Foo",
        "operationId": "Foo",
        "parameters": [
          {
            "name": "q",
            "in": "query",
            "schema": {
              "$ref": "#/components/schemas/TestEnum"
            },
            "required": true,
            "deprecated": false,
            "allowEmptyValue": false,
            "allowReserved": false
          }
        ],
        "responses": {
          "200": {
            "headers": {}
          },
          "400": {
            "description": "Bad request syntax or unsupported method",
            "content": {
              "application/json": {
                "schema": {
                  "properties": {
                    "status_code": {
                      "type": "integer"
                    },
                    "detail": {
                      "type": "string"
                    },
                    "extra": {
                      "additionalProperties": {},
                      "type": [
                        "null",
                        "object",
                        "array"
                      ]
                    }
                  },
                  "type": "object",
                  "required": [
                    "detail",
                    "status_code"
                  ],
                  "description": "Validation Exception",
                  "examples": [
                    {
                      "status_code": 400,
                      "detail": "Bad Request",
                      "extra": {}
                    }
                  ]
                }
              }
            }
          }
        },
        "deprecated": false
      }
    }
  },
  "components": {
    "schemas": {
      "TestEnum": {
        "type": "string",
        "enum": [
          "one",
          "two"
        ],
        "title": "TestEnum"
      }
    }
  }
}
```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.16.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

250MHz:
## Closes

Closes #4250

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -592,7 +592,7 @@ class SchemaCreator:
         schema.type = enum_type
         schema.enum = enum_values
         schema.title = get_name(field_definition.annotation)
-        schema.description = field_definition.annotation.__doc__
+        self.process_schema_result(field_definition, schema)
 
         return self.schema_registry.get_reference_for_field_definition(field_definition) or schema
 
--- a/tests/unit/test_openapi/test_parameters.py
+++ b/tests/unit/test_openapi/test_parameters.py
@@ -111,7 +111,7 @@ def test_create_parameters(person_controller: Type[Controller]) -> None:
             Schema(
                 type=OpenAPIType.ARRAY,
                 items=Reference(ref="#/components/schemas/tests_unit_test_openapi_utils_Gender"),
-                examples=[[Gender.MALE]],
+                examples=[[Gender.FEMALE]],
             ),
             Schema(type=OpenAPIType.NULL),
         ],
@@ -348,6 +348,7 @@ def test_parameter_schema_extra() -> None:
                 }
             ),
         ],
+        query2: Annotated[Gender, Parameter(description="gender description", schema_extra={"format": "foo"})],
     ) -> Any:
         return query1
 
@@ -364,6 +365,9 @@ def test_parameter_schema_extra() -> None:
             {"type": "string", "enum": ["denied", "values"]},
         ]
     }
+    assert schema["paths"]["/"]["get"]["parameters"][1]["schema"]["$ref"] == "#/components/schemas/Gender"
+    assert schema["paths"]["/"]["get"]["parameters"][1]["description"] == "gender description"
+    assert schema["components"]["schemas"]["Gender"]["format"] == "foo"
 
     # Attempt to pass invalid key
     app = Litestar([error_handler])
--- a/tests/unit/test_openapi/test_parameters.py
+++ b/tests/unit/test_openapi/test_parameters.py
@@ -51,7 +51,7 @@ def test_create_parameters(person_controller: Type[Controller]) -> None:
 
     parameters = _create_parameters(app=Litestar(route_handlers=[person_controller]), path="/{service_id}/person")
     assert len(parameters) == 10
-    page, name, service_id, page_size, from_date, to_date, gender, lucky_number, secret_header, cookie_value = tuple(
+    service_id, page, name, page_size, from_date, to_date, gender, lucky_number, secret_header, cookie_value = tuple(
         parameters
     )
 
--- a/tests/unit/test_openapi/conftest.py
+++ b/tests/unit/test_openapi/conftest.py
@@ -30,17 +30,17 @@ def create_person_controller() -> Type[Controller]:
             state: State,
             query: Dict[str, Any],
             cookies: Dict[str, Any],
+            # path parameter
+            service_id: int,
             # required query parameters below
             page: int,
             name: Optional[Union[str, List[str]]],  # intentionally without default
-            service_id: int,
             page_size: int = Parameter(
                 query="pageSize",
                 description="Page Size Description",
                 title="Page Size Title",
                 examples=[Example(description="example value", value=1)],
             ),
-            # path parameter
             # non-required query parameters below
             from_date: Optional[Union[int, datetime, date]] = None,
             to_date: Optional[Union[int, datetime, date]] = None,


tests/unit/test_openapi/test_parameters.py
```diff
@@ -111,7 +111,7 @@ def test_create_parameters(person_controller: Type[Controller]) -> None:
             Schema(
                 type=OpenAPIType.ARRAY,
                 items=Reference(ref="#/components/schemas/tests_unit_test_openapi_utils_Gender"),
-                examples=[[Gender.MALE]],
+                examples=[[Gender.FEMALE]],
```
provinzkraut, line 114: I feel like this shouldn't change?

litestar/_openapi/schema_generation/schema.py
```diff
@@ -592,7 +592,7 @@ def for_enum_field(
         schema.type = enum_type
         schema.enum = enum_values
         schema.title = get_name(field_definition.annotation)
-        schema.description = field_definition.annotation.__doc__
+        self.process_schema_result(field_definition, schema)
```
provinzkraut, line 575: Should add a test that this doens't change the default behaviour, and still uses the member docstring if nothing else is specified

tests/unit/test_openapi/test_parameters.py
250MHz: I don't understand Polyfactory well enough to debug this.

The `gender` parameter has the type `Optional[Union[Gender, List[Gender]]]`. So `for_field_definition` seems to be called like:
* `Union[Gender, List[Gender], NoneType]`
  * `Union[Gender, List[Gender]]`
    * `<enum 'Gender'>`
      * My changes cause `process_schema_result()` to be called. There's no examples for the Schema, so `create_examples_for_field()` is called which generates `Gender.MALE`
    * `List[Gender]`
      * `<enum 'Gender'>`
        * My change causes `process_schema_result()` to be called, but there's already an example `[Gender.MALE]` from earlier, so `create_examples_for_field()` is not called
      * `process_schema_result()` is called for `List[Gender]`. The Schema has no examples, so `create_examples_for_field()` is called
        * My change generates `[Gender.FEMALE]`
        * Current code generates `[Gender.MALE]`
    * `process_schema_result()` is called for `Union[Gender, List[Gender]]`. The Schema has no examples, so `create_examples_for_field()` is called
      * My change generates `[Gender.ANY]` (note: `_normalize_example_value()` returns `List[Gender]`)
      * Current code generates `Gender.ANY` (note: `_normalize_example_value()` returns `<enum 'Gender'>`)
  * `process_schema_result()` is called for `Union[Gender, List[Gender], NoneType]`. The Schema already has an explicitly set example `[Gender.MALE, [Gender.MALE, Gender.OTHER]]`, so `create_examples_for_field()` is not called

I don't understand why `create_examples_for_field()` being called for `List[Gender]` for the first time generates different values or why `create_examples_for_field()` for `Union[Gender, List[Gender]]` uses `List[Gender]` instead of `<enum 'Gender'>`. Do you have any ideas for how I could fix this?

litestar/_openapi/schema_generation/schema.py
250MHz: This change does remove the behavior of using the enum's docstring to set the Schema's description. We can add that back like:
 ```suggestion
         self.process_schema_result(field_definition, schema)
         if schema.description is None:
             schema.description = field_definition.annotation.__doc__
```
But could you confirm if this behavior is desired? You could not use the docstring like this in v2.13.0 before #3525, and I didn't see discussion about this behavior, so I thought that it was unintentionally added. In my own application, I was surprised to see my enum's docstring (with internal details irrelevant for API users) show up in my OpenAPI documentation since I don't think this behavior is documented.

Comment (provinzkraut) [reply litestar/_openapi/schema_generation/schema.py:575]: The behaviour is desired; It mirrors the functionality of how we process e.g. dataclass members.

File: tests/unit/test_openapi/test_parameters.py (hunk 1 above)
Comment (provinzkraut) [reply tests/unit/test_openapi/test_parameters.py:114]: Okay, thanks for the thorough investigation. Maybe @adhtruong has an idea why this is happening, but right now it seems to be acceptable in this test case.

**Comment by 250MHz** [comment]
Note that since only one Schema is made per Enum, if you use Parameter() multiple times with different descriptions, you would overwrite the description of a previous definition. The easy way around this is to use `schema_component_key` so that a separate Schema is made. (Should I add documentation for this somewhere if it doesn't already exist? Or is the docstring for `schema_component_key` sufficient for communicating this?)

**Comment by provinzkraut** [comment]
> Should I add documentation for this somewhere if it doesn't already exist? Or is the docstring for `schema_component_key` sufficient for communicating this?

I think it make sense to document this!

**Comment by 250MHz** [comment]
I'm not sure about the max line length to use for .rst files. I went with 120 characters since that seems like what most of parameters.rst uses, so I hope that's alright

File: docs/usage/routing/parameters.rst [hunk 3]
```diff
@@ -164,6 +164,28 @@ will be used for the value of the ``snake_case`` parameter.
 
 In this case, ``param`` is validated to be an *integer larger than 5*.
 
+Documentation for enum query parameters
```
provinzkraut:
```suggestion
Documenting enum query parameters
```

docs/usage/routing/parameters.rst
```diff
@@ -164,6 +164,28 @@ will be used for the value of the ``snake_case`` parameter.
...
+---------------------------------------
+
+Using :func:`~.params.Parameter` to modify the OpenAPI schema generated for query :term:`parameters <parameter>` that
+are enums can have surprising results. Since only one schema is generated per enum, calling :func:`~.params.Parameter`
```
provinzkraut:
```suggestion
Since only one schema is generated per enum, calling :func:`~.params.Parameter`
```
And maybe instead mention how the docstring is used for the description by default?

provinzkraut (approved):

provinzkraut (approved):

250MHz:
I use "can" to get around giving a more complete explanation. I'm not sure if I should talk about the following...

> The description can be changed with the description parameter of the the parameter function, but doing so can overwrite the descriptions [...]

`params.Parameter.description` _can_ change / overwrite the schema's description if `Schema.description` is not `None`. But if a schema was already generated for the enum with a description, then using `description` again has no effect. (E.g., in the example, if `q2` came before `q1`, then the schema's description is the docstring.)

> This can be avoided by using the schema_component_key [...]

`schema_component_key` is one way for multiple query parameters of the same enum to have different parameter descriptions, but you _can_ also use `schema_extra` like `Parameter(schema_extra={"description": "..."})`. Note that using `schema_extra` sets `Schema.description` regardless if it's already set. So using `schema_extra` multiple times will generate different `openapi.spec.Parameter` descriptions, but if `schema_component_key` isn't set, then there's just one `openapi.spec.Schema` description that's repeatedly overwritten. If you used another query parameter of the same enum (without a distinct `schema_component_key` or without using `schema_extra`) after the enum's schema's description was already generated, then the `openapi.spec.Parameter` description would be the previously set description.

Side note: if anyone who encounters problems with this in the future and wants to work on this, I think it's better for `params.Parameter.description` to only set `openapi.spec.Parameter.description` and not change the `openapi.spec.Schema.description` of the enum. I think the latter should stay as the enum's docstring or whatever the user sets through `params.Parameter.schema_extra`.

250MHz:
@provinzkraut Is it okay for me to resolve the comments since the PR was approved? Just wanna be sure

PR #4261, opened 2025-08-16
chore: Update pytest

provinzkraut:


tests/unit/test_concurrency.py
```diff
@@ -32,8 +33,16 @@ def func() -> int:
     return 1
 
 
-def test_sync_to_thread_asyncio() -> None:
+@pytest.fixture()
+def loop() -> Generator[AbstractEventLoop, None, None]:
```
sobolevn:
```suggestion
def loop() -> Iterator[AbstractEventLoop]:
```

tests/unit/test_concurrency.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from asyncio import AbstractEventLoop
 from collections.abc import Generator
```
sobolevn:
```suggestion
from collections.abc import Iterator
```

sobolevn (commented):
You can also change `def reset_state() -> Generator[None, None, None]:` to be `def reset_state() -> Iterator[None]:`, which is more correct :)

provinzkraut:
> You can also change `def reset_state() -> Generator[None, None, None]:` to be `def reset_state() -> Iterator[None]:`, which is more correct :)

Could you explain why it is *more correct*? Since `Generator` is a subtype of `Iterator`, and a generator is what's actually returned, `Generator` would be more specific. Why wouldn't that be considered more correct?

sobolevn:
In my opinion, we only need `Generator` when we expect to annotate `send` and `return` types. `Iterator` is good enough for this abstraction.

provinzkraut:
> In my opinion, we only need `Generator` when we expect to annotate `send` and `return` types. `Iterator` is good enough for this abstraction.

I see, so it's just that `Generator` is overspecialized for the use case. Thanks for the explanation.

euri10 (approved):

Issue #4265, opened 2025-08-19
Docs: Lifespan migration from Starlette / FastAPI needs update


danielniccoli:
### Summary

It seems the sample code at https://docs.litestar.dev/2/migration/fastapi.html#lifespan is no longer correct and in need of an update:

```py
@asynccontextmanager
async def lifespan(
    app_settings: AppSettings,
    _app: Litestar,
):
    # Setup code here
    yield
    # Teardown code here
```

PR #4268, opened 2025-08-20
docs: fix lifespan migration from fastapi

cclinet:
https://github.com/litestar-org/litestar/issues/4265

For reference, see the following two links:

https://docs.litestar.dev/2/usage/applications.html#lifespan-context-managers

https://fastapi.tiangolo.com/advanced/events/#lifespan

### DIFF
--- a/docs/migration/fastapi.rst
+++ b/docs/migration/fastapi.rst
@@ -177,7 +177,7 @@ and to easily access dependencies from higher levels.
 Lifespan
 ~~~~~~~~
 
-Litestar has the same async context manager syntax as FastAPI.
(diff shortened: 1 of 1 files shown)


docs/migration/fastapi.rst
```diff
@@ -177,7 +177,7 @@ and to easily access dependencies from higher levels.
 Lifespan
 ~~~~~~~~
 
-If you're using an async context manager and pass parameters to it, most likely the order of parameters is inversed between FastAPI and Litestar.
+Litestar has the same async context manager syntax as FastAPI.
```
provinzkraut:
```suggestion
Litestar uses the same async context manager style as FastAPI, so the code does not need to be changed:
```

PR #4271, opened 2025-08-24
fix: Preserve empty strings in multipart/form-data requests (#4204)

### LINKED ISSUE #4204: Bug: Empty strings in multipart/form-data requests are converted to None
### Description

Empty strings are parsed correctly when using `application/x-www-form-urlencoded`. However, when using `multipart/form-data`, they instead get converted to `None`.

This is due to `litestar/_multipart.py:112`:

```
if isinstance(data, UploadFile):
    await data.seek(0)
    fields[segment.name].append(data)
elif data:  # here, data is a bytearray
    fields[segment.name].append(data.decode(segment.charset or "utf-8"))
else:
    fields[segment.name].append(None)
```

Because `data` is tested using boolean conversion instead of explicit comparison,`''` (which converts to the empty bytearray) will fail the check and the `else` branch will be taken.

I'm willing to create a PR for this, but it would be a breaking change.

_No response_

### MCVE

```python
from typing import Annotated

from litestar import Litestar
from litestar import post


class Data(BaseModel):
    value: str | None


@post("/f", sync_to_thread=False)
def f(data: Annotated[Data, Body(media_type=RequestEncodingType.URL_ENCODED)]) -> bool:
    return data.value is not None


@post("/g", sync_to_thread=False)
def g(data: Annotated[Data, Body(media_type=RequestEncodingType.MULTI_PART)]) -> bool:
    return data.value is not None


app = Litestar(route_handlers=[f, g])
```

### Steps to reproduce

```
>>> import httpx
>>> httpx.post('http://localhost:8000/f', data={'value': ''}).text
'true'
>>> httpx.post('http://localhost:8000/g', data={'value': ''}, files={'file': ''}).text
'false'
```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.16.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

doubledare704:
## Description
- Fix issue #4204 where empty strings in multipart forms were converted to None
- Change boolean check to length check in _multipart.py line 114
- Empty bytearrays now decode to empty strings instead of becoming None
- Add comprehensive tests for empty string handling in multipart forms
- Raw multipart content parsing works correctly
- Existing functionality remains unaffected

### DIFF
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -37,7 +37,7 @@ from typing_extensions import (
 from litestar.constants import UNDEFINED_SENTINELS
 from litestar.types.builtin_types import NoneType, UnionTypes
 from litestar.utils.helpers import unwrap_partial
-from litestar.utils.typing import get_origin_or_inner_type, make_non_optional_union
+from litestar.utils.typing import get_origin_or_inner_type
 
 if TYPE_CHECKING:
     from litestar.types.protocols import DataclassProtocol
@@ -56,7 +56,6 @@ __all__ = (
     "is_non_string_iterable",
     "is_non_string_sequence",
     "is_optional_union",
-    "is_string_type",
     "is_undefined_sentinel",
     "is_union",
 )
@@ -295,30 +294,3 @@ def is_undefined_sentinel(value: Any) -> bool:
         A boolean.
     """
     return any(v is value for v in UNDEFINED_SENTINELS)
-
-
-def is_string_type(annotation: Any) -> bool:
-    """Check if a type annotation represents a string type.
-
-    This includes ``str``, subclasses of ``str``, and ``Optional`` variants of
-    these types. It also handles ``Annotated`` and ``ForwardRef`` types.
-
-    Args:
-        annotation: A type annotation to check.
-
-    Returns:
-        ``True`` if the annotation is a string type, ``False`` otherwise.
-    """
-    while get_origin(annotation) is Annotated:
-        annotation = get_args(annotation)[0]
-
-    if is_optional_union(annotation):
-        annotation = make_non_optional_union(annotation)
-        while get_origin(annotation) is Annotated:
-            annotation = get_args(annotation)[0]
-
-    # This handles cases like Union[str, int] which are not string types
-    if get_origin(annotation) in UnionTypes:
-        return False
-
-    return is_class_and_subclass(annotation, str)
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -37,7 +37,7 @@ from typing_extensions import (
 from litestar.constants import UNDEFINED_SENTINELS
 from litestar.types.builtin_types import NoneType, UnionTypes
 from litestar.utils.helpers import unwrap_partial
-from litestar.utils.typing import get_origin_or_inner_type
+from litestar.utils.typing import get_origin_or_inner_type, make_non_optional_union
 
 if TYPE_CHECKING:
     from litestar.types.protocols import DataclassProtocol
@@ -56,6 +56,7 @@ __all__ = (
     "is_non_string_iterable",
     "is_non_string_sequence",
     "is_optional_union",
+    "is_string_type",
     "is_undefined_sentinel",
     "is_union",
 )
@@ -294,3 +295,30 @@ def is_undefined_sentinel(value: Any) -> bool:
         A boolean.
     """
     return any(v is value for v in UNDEFINED_SENTINELS)
+
+
+def is_string_type(annotation: Any) -> bool:
+    """Check if a type annotation represents a string type.
+
+    This includes ``str``, subclasses of ``str``, and ``Optional`` variants of
+    these types. It also handles ``Annotated`` and ``ForwardRef`` types.
+
+    Args:
+        annotation: A type annotation to check.
+
+    Returns:
+        ``True`` if the annotation is a string type, ``False`` otherwise.
+    """
+    while get_origin(annotation) is Annotated:
+        annotation = get_args(annotation)[0]
+
+    if is_optional_union(annotation):
+        annotation = make_non_optional_union(annotation)
+        while get_origin(annotation) is Annotated:
+            annotation = get_args(annotation)[0]
+
+    # This handles cases like Union[str, int] which are not string types
+    if get_origin(annotation) in UnionTypes:
+        return False
+
+    return is_class_and_subclass(annotation, str)
--- a/tests/unit/test_kwargs/test_multipart_data.py
+++ b/tests/unit/test_kwargs/test_multipart_data.py
@@ -505,14 +505,27 @@ class ProductForm:
     optional_with_default: Optional[int] = None
 
 
-def test_multipart_handling_of_none_values() -> None:
+def test_multipart_handling_of_optional_values() -> None:
+    """Test that multipart forms handle optional fields correctly.
+
+    This test verifies that optional fields work properly with valid values,
+    and demonstrates the correct behavior after fixing issue #4204.
+    """
+
     @post("/", signature_types=[ProductForm])
     def handler(
         data: Annotated[ProductForm, Body(media_type=RequestEncodingType.MULTI_PART)],
-    ) -> None:
-        assert data
+    ) -> dict[str, Any]:
+        return {
+            "name": data.name,
+            "int_field": data.int_field,
+            "options": data.options,
+            "optional_without_default": data.optional_without_default,
+            "optional_with_default": data.optional_with_default,
+        }
 
     with create_test_client(route_handlers=[handler]) as client:
+        # Test with valid values for all fields
         response = client.post(
             "/",
             content=(
@@ -530,15 +543,23 @@ def test_multipart_handling_of_none_values() -> None:
                 b"[1,2,3,4]\r\n"
                 b"--1f35df74046888ceaa62d8a534a076dd\r\n"
                 b'Content-Disposition: form-data; name="optional_without_default"\r\n'
-                b"Content-Type: application/octet-stream\r\n\r\n\r\n"
+                b"Content-Type: application/octet-stream\r\n\r\n"
+                b"3.14\r\n"
                 b"--1f35df74046888ceaa62d8a534a076dd\r\n"
                 b'Content-Disposition: form-data; name="optional_with_default"\r\n'
-                b"Content-Type: application/octet-stream\r\n\r\n\r\n"
+                b"Content-Type: application/octet-stream\r\n\r\n"
+                b"42\r\n"
                 b"--1f35df74046888ceaa62d8a534a076dd--\r\n"
             ),
             headers={"Content-Type": "multipart/form-data; boundary=1f35df74046888ceaa62d8a534a076dd"},
         )
         assert response.status_code == HTTP_201_CREATED
+        result = response.json()
+        assert result["name"] == "moishe zuchmir"
+        assert result["int_field"] == 1
+        assert result["options"] == "[1,2,3,4]"
+        assert result["optional_without_default"] == 3.14
+        assert result["optional_with_default"] == 42
 
 
 class AddProductFormMsgspec(msgspec.Struct):
--- a/tests/unit/test_kwargs/test_multipart_data.py
+++ b/tests/unit/test_kwargs/test_multipart_data.py
@@ -621,98 +621,33 @@ def test_invalid_multipart_raises_client_error() -> None:
         assert response.status_code == HTTP_400_BAD_REQUEST
 
 
-# Test for GitHub issue #4204: Empty strings in multipart/form-data should be preserved
+# https://github.com/litestar-org/litestar/issues/4204
 def test_empty_strings_preserved_in_multipart_forms() -> None:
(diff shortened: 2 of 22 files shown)


tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +598,93 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+@post("/test-empty-url-encoded", sync_to_thread=False)
+def test_empty_url_encoded(
+    data: Annotated[EmptyStringTestData, Body(media_type=RequestEncodingType.URL_ENCODED)],
+) -> bool:
+    return data.value is not None
+
+
+@post("/test-empty-multipart", sync_to_thread=False)
+def test_empty_multipart(data: Annotated[EmptyStringTestData, Body(media_type=RequestEncodingType.MULTI_PART)]) -> bool:
+    return data.value is not None
```
provinzkraut: These should be included in the test case

litestar/utils/predicates.py
```diff
@@ -294,3 +295,30 @@ def is_undefined_sentinel(value: Any) -> bool:
         A boolean.
     """
     return any(v is value for v in UNDEFINED_SENTINELS)
+
+
+def is_string_type(annotation: Any) -> bool:
```
provinzkraut: Not necessary as this functionality already exists on `FieldDefinition`, e.g. using `is_subclass_of` / `has_inner_subclass_of`

doubledare704: removed

tests/unit/test_kwargs/test_multipart_data.py
doubledare704: moved to test case

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +619,93 @@ def test_invalid_multipart_raises_client_error() -> None:
             headers={"Content-Type": "multipart/form-data; charset=utf-8; boundary=20b303e711c4ab8c443184ac833ab00f"},
         )
         assert response.status_code == HTTP_400_BAD_REQUEST
+
+
+# Test for GitHub issue #4204: Empty strings in multipart/form-data should be preserved
+class EmptyStringTestData(BaseModel):
```
provinzkraut: Why is this testing a Pydantic model? This issue has nothing to do with Pydantic

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -504,14 +504,27 @@ class ProductForm:
     optional_with_default: Optional[int] = None
 
 
-def test_multipart_handling_of_none_values() -> None:
+def test_multipart_handling_of_optional_values() -> None:
+    """Test that multipart forms handle optional fields correctly.
```
provinzkraut, line 509:
```suggestion
    # https://github.com/litestar-org/litestar/issues/4204
    """Test that multipart forms handle optional fields correctly.
```

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+
+        # Test multipart form
+        response_multipart = client.post("/consistency-test", data={"value": ""}, files={"dummy": ""})
+        multipart_result = response_multipart.json()
+
+        # Both should return empty string, not None
+        assert url_result["value"] == ""
+        assert multipart_result["value"] == ""
+        assert url_result["value_type"] == "str"
+        assert multipart_result["value_type"] == "str"
```
provinzkraut: This is already covered by the other tests. You should try to keep these units tests minimal and focused; They ideally should only test *one* thing. What this test wants to show is that the multipart / url encoded form data decodes the same, so there should only be one assert needed for this.

This makes the tests easier to follow, since their intention is more clear.

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+
+        # Results should be identical
+        assert url_result == multipart_result
+
+
+def test_multipart_empty_string_with_raw_content() -> None:
```
provinzkraut: This test is unnecessary. It only proves that the manually generated multipart payload is equivalent to the one's generated by httpx. We already assume this is the case anyway, and we're not trying to prove that the parser is working correctly in this test case.

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
             headers={"Content-Type": "multipart/form-data; charset=utf-8; boundary=20b303e711c4ab8c443184ac833ab00f"},
         )
         assert response.status_code == HTTP_400_BAD_REQUEST
+
+
+# Test for GitHub issue #4204: Empty strings in multipart/form-data should be preserved
```
provinzkraut:
```suggestion
# https://github.com/litestar-org/litestar/issues/4204
```

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+def test_empty_strings_preserved_in_multipart_forms() -> None:
+    """Test that empty strings are preserved in multipart forms and not converted to None.
+
+    This test addresses GitHub issue #4204 where empty strings in multipart/form-data
+    requests were being converted to None instead of being preserved as empty strings.
+    The test focuses on core multipart parsing behavior, not serialization library specifics.
```
provinzkraut:
```suggestion
    This test addresses GitHub issue #4204 where empty strings in multipart/form-data
    requests were being converted to None instead of being preserved as empty strings.
    The test focuses on core multipart parsing behavior, not serialization library specifics.
```
this doesn't really add anything, a short comment and a link to the issue is fine

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+
+    @post("/test-form")
+    async def form_handler(request: Request) -> dict[str, Any]:
+        """Handler that directly tests form parsing without any serialization library."""
+        data = await request.form()
+        return {
+            "value": data.get("value"),
+            "value_type": type(data.get("value")).__name__,
+            "value_is_none": data.get("value") is None,
+        }
```
provinzkraut:
```suggestion
        assert data.get("value") == ""
```

tests/unit/test_kwargs/test_multipart_data.py
```diff
@@ -597,3 +618,100 @@ def test_invalid_multipart_raises_client_error() -> None:
...
+        # Both should preserve empty strings, not convert to None
+        assert url_result["value"] == ""
+        assert multipart_result["value"] == ""
+        assert url_result["value_type"] == "str"
+        assert multipart_result["value_type"] == "str"
+        assert url_result["value_is_none"] is False
+        assert multipart_result["value_is_none"] is False
+
+        # Results should be identical (consistency between form types)
+        assert url_result == multipart_result
```
provinzkraut: You already have an extra test case below that checks if multipart and url-encoded are the same. There is no need to test the same behaviour multiple times in multiple different tests. See my earlier explanation as to why it's better to keep tests small and focused.

provinzkraut (changes requested):
This already looks better. Still, a couple of suggestions for the tests

gsakkis:
After upgrading to 2.18 some of my tests broke due to this change. Ultimately it's an issue on my end but it took some time to figure it out, plus there are some difference between DTO (PydanticDTO specifically) and non-DTO (Pydantic BaseModel). 

In short:
- There is a `MyForm` model containing an UploadFile and a nullable field.
- In some tests I was passing `{field: None}` as form data and apparently http clients (or at least httpx) converts None to empty string.
- In 2.17 this works (i.e. the empty string is validated successfully as None) both when annotating the data as `MyForm` and as `DTOData[MyForm]`.
- In 2.18 it fails with
  - HTTP 400 (`litestar.ValidationException`) if annotated as `MyForm`
  - HTTP 500 (`msgspec.ValidationError`) if annotated as `DTOData[MyForm]` with `dto=PydanticDTO[MyForm]`
- Unrelated to this PR but while playing around I discovered that (both in 2.17 and 2.18) the string `"null"` is surprisingly converted to None in the DTO case (but not in the plain BaseModel case). Not sure if this is a bug, a feature or neither.
   - Update: it's even stranger, apparently it's case insensitive - `"NULL"` and `"Null"` are also converted to None.

I can post a MCVE if interested.

euri10:
> I can post a MCVE if interested.

i think you should

gsakkis:
- Server
```python
from datetime import date
from typing import Annotated, Any

from litestar.dto import DTOData
from litestar.plugins.pydantic import PydanticDTO


class MyForm(BaseModel, arbitrary_types_allowed=True):
    script: UploadFile
    updated_on: date | None = None


@post("/")
async def process(
    data: Annotated[MyForm, Body(media_type=RequestEncodingType.MULTI_PART)],
) -> dict[str, Any]:
    return {"filename": data.script.filename, "updated_on": data.updated_on}


@post("/dto", dto=PydanticDTO[MyForm])
async def process_dto(
    data: Annotated[DTOData[MyForm], Body(media_type=RequestEncodingType.MULTI_PART)],
) -> dict[str, Any]:
    builtins = data.as_builtins()
    return {
        "filename": builtins["script"].filename,
        "updated_on": builtins["updated_on"],
    }


app = Litestar(route_handlers=[process, process_dto])
```

- Client
```python
import httpx

with open("upload.py", "rb") as f:
    files = {"script": f}

    for data in {}, {"updated_on": None}, {"updated_on": "null"}:
        print(f"\n--- data={data} ---\n")
        resp = httpx.post("http://localhost:8000", files=files, data=data)
        print(resp.text)
        print("-" * 120)
        resp = httpx.post("http://localhost:8000/dto", files=files, data=data)
        print(resp.text)
```

- Output with 2.17
```
--- data={} ---

INFO - 2025-10-08 10:12:00,484 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:12:00,489 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}

--- data={'updated_on': None} ---

INFO - 2025-10-08 10:12:00,495 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:12:00,500 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}

--- data={'updated_on': 'null'} ---

INFO - 2025-10-08 10:12:00,509 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 400 Bad Request"
{"status_code":400,"detail":"Validation failed for POST /","extra":[{"message":"Input should be a valid date or datetime, input is too short","key":"updated_on"}]}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:12:00,515 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}
```

- Output with 2.18
```

--- data={} ---

INFO - 2025-10-08 10:11:23,666 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:11:23,674 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}

--- data={'updated_on': None} ---

INFO - 2025-10-08 10:11:23,681 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 400 Bad Request"
{"status_code":400,"detail":"Validation failed for POST /","extra":[{"message":"Input should be a valid date or datetime, input is too short","key":"updated_on"}]}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:11:23,688 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 500 Internal Server Error"
{"status_code":500,"detail":"Internal Server Error"}

--- data={'updated_on': 'null'} ---

INFO - 2025-10-08 10:11:23,694 - httpx - _client - HTTP Request: POST http://localhost:8000 "HTTP/1.1 400 Bad Request"
{"status_code":400,"detail":"Validation failed for POST /","extra":[{"message":"Input should be a valid date or datetime, input is too short","key":"updated_on"}]}
------------------------------------------------------------------------------------------------------------------------
INFO - 2025-10-08 10:11:23,699 - httpx - _client - HTTP Request: POST http://localhost:8000/dto "HTTP/1.1 201 Created"
{"filename":"upload.py","updated_on":null}
```

PR #4276, opened 2025-08-30
fix: possibly get httpexception.extra and httpexception.headers from class for v3

PokkaKiyo:
## Description

- For HTTPException, possibly get the value for extra and headers on the class

## Closes
- #4006


PokkaKiyo:
Happy to help!

PR #4286, opened 2025-09-06
feat: zstd middleware

### LINKED ISSUE #4168: Enhancement: Add zstd compression
### Summary

Add zstd compression middleware to Litestar.

With most modern browsers supporting zstd (chromium and firefox), I think it would be great to have the ability to use zstd compression in Litestar.

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

ceballots:
## Description

This PR introduces a Zstd compression middleware.

- `ZstdCompression` facade added to handle Zstandard compression.
- `final` parameter added to the write method of the facade to distinguish between streaming and non-streaming writes.

Reasoning for final

- When streaming, chunks are flushed incrementally using `zstd.FLUSH_BLOCK` for better performance and reduced memory usage.

- When the full response is available, `final=True` flushes the complete frame using `zstd.FLUSH_FRAME`, producing a valid Zstd stream for the client.

This approach keeps middleware generic and avoids Zstd-specific logic outside the facade.

## Closes
Closes #4168


pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
   "exceptiongroup>=1.2.2; python_version < \"3.11\"",
   # default litestar plugins
   "litestar-htmx>=0.4.0",
+  "zstandard>=0.24.0",
```
sobolevn: `compression.zstd` was added to 3.14, so this dep might not be required on further versions

litestar/middleware/compression/zstd_facade.py
```diff
@@ -0,0 +1,41 @@
...
+
+    encoding = CompressionEncoding("zstd")
+
+    def __init__(self, buffer: BytesIO, compression_encoding: Literal["zstd"] | str, config: CompressionConfig) -> None:
+        self.buffer = buffer
+        self.compression_encoding = compression_encoding
+        self.cctx = zstd.ZstdCompressor(level=config.zstd_compress_level)
+        self.compressor = self.cctx.stream_writer(buffer)
+
+    def write(self, body: bytes, final: bool = False) -> None:
```
sobolevn: `body` is allowed to be `bytearray` as well: https://github.com/indygreg/python-zstandard/blob/32a31f889e9f968320b76420e8cfd051beec35fc/zstandard/__init__.pyi#L258

litestar/config/compression.py
```diff
@@ -21,7 +21,7 @@ class CompressionConfig:
     using the ``compression_config`` key.
     """
 
-    backend: Literal["gzip", "brotli"] | str
+    backend: Literal["gzip", "brotli", "zstd"] | str
```
sobolevn, line 24: This is the same as just `str`

litestar/config/compression.py
```diff
@@ -81,3 +85,11 @@ def __post_init__(self) -> None:
 
             self.gzip_fallback = self.brotli_gzip_fallback
             self.compression_facade = BrotliCompression
+        elif self.backend == "zstd":
+            if self.zstd_compress_level < 1 or self.zstd_compress_level > 22:
+                raise ImproperlyConfiguredException("zstd_compress_level must be between 1 and 22")
```
sobolevn:
```suggestion
                raise ImproperlyConfiguredException(f"zstd_compress_level must be between 1 and 22, given: {self.zstd_compress_level}")
```

litestar/middleware/compression/brotli_facade.py
```diff
@@ -43,7 +43,7 @@ def __init__(
             lgblock=config.brotli_lgblock,
         )
 
-    def write(self, body: bytes) -> None:
+    def write(self, body: bytes, final: bool) -> None:
```
sobolevn: I think that this must have a default for backward compat.

litestar/middleware/compression/facade.py
```diff
@@ -29,11 +29,12 @@ def __init__(
         """
         ...
 
-    def write(self, body: bytes) -> None:
+    def write(self, body: bytes, final: bool) -> None:
```
sobolevn: the same here.

litestar/middleware/compression/middleware.py
```diff
@@ -82,10 +83,22 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
         await self.app(scope, receive, send)
 
+    def get_facade_cls(
+        self,
+        compression_encoding: Literal[CompressionEncoding.BROTLI, CompressionEncoding.GZIP, CompressionEncoding.ZSTD]
+        | str,
+    ) -> type[CompressionFacade]:
+        if compression_encoding == CompressionEncoding.GZIP:
+            return GzipCompression
+        if compression_encoding == CompressionEncoding.ZSTD:
```
sobolevn, line 93:
```suggestion
        elif compression_encoding == CompressionEncoding.ZSTD:
```

`elif` is better when working with `Literal` values. 

Is that correct that we don't use `assert_never()` here?

ceballots: I'd say no, to allow arbitrary user-supplied encodings

pyproject.toml
ceballots: fair, I can add the conditional dependency, but I'm afraid mypy and other checks might fail when trying to locate type stubs for the built-in compression.zstd module. Since 3.14 is not supported yet, I'd leave this out of this PR

litestar/config/compression.py
sobolevn: But, it can be left this way for extra context.

ceballots: thanks, gonna leave it as is then

pyproject.toml
sobolevn: let's add a comment maybe? :)

ceballots: good idea, added it!

ceballots:
@sobolevn I believe all comments were addressed. Let me know if there is anything left to do and what is the next step. Thanks!

pyproject.toml
```diff
@@ -50,6 +50,8 @@ dependencies = [
   "exceptiongroup>=1.2.2; python_version < \"3.11\"",
   # default litestar plugins
   "litestar-htmx>=0.4.0",
+  # starting from Python 3.14, `compression.zstd` is built-in and this will no longer be required
+  "zstandard>=0.24.0",
```
provinzkraut: Maybe add a version specifier right away? Then we don't have to remember adding it when 3.14 releases :)

provinzkraut: This should not be a required dependency, but an optional one, with a corresponding extra, like `brotli`

litestar/middleware/compression/facade.py
```diff
@@ -29,11 +29,13 @@ def __init__(
         """
         ...
 
-    def write(self, body: bytes) -> None:
-        """Write compressed bytes.
+    def write(self, body: bytes | bytearray, final: bool = False) -> None:
```
provinzkraut, line 32: Changing this to `bytearray` and introducing the `final` parameter makes this a breaking change. That's okay, it just means we can't release this in the `2.x` release line and it'll have to wait until `3.0`

provinzkraut (changes requested):

litestar/middleware/compression/facade.py
ceballots: True, then I should update the PR to be againts `v3.0`, right?

provinzkraut: No need, the `main` branch is on v3 already :)

pyproject.toml
ceballots: done

ceballots: done

ceballots:
@provinzkraut address your comments, let me know if you have any further suggestions! thank you

PR #4289, opened 2025-09-13
fix(routing): Ensure a `MethodNotAllowedException` exception during routing properly sets `Allow` header 

### LINKED ISSUE #4277: Bug: `405 Method Not Allowed` response should include the `Allow` header
### Description

According to the MDN Web Docs[^1]:

<blockquote>
<p>The HTTP <strong><code>Allow</code></strong> <a href="/en-US/docs/Glossary/Response_header">response header</a> lists the set of <a href="/en-US/docs/Web/HTTP/Reference/Methods">request methods</a> supported by a resource.
This header must be sent if the server responds with a <a href="/en-US/docs/Web/HTTP/Reference/Status/405"><code>405 Method Not Allowed</code></a> status code to indicate which request methods can be used instead.
An empty <code>Allow</code> value indicates that the resource allows no request methods, which might occur temporarily for a given resource.</p>
</blockquote>

Currently, when a router does not support an HTTP method, an `MethodNotAllowedException` is raised here

https://github.com/litestar-org/litestar/blob/26f20ac6c52de2b4bf81161f7560c8bb4af6f382/litestar/_asgi/routing_trie/traversal.py#L173-L174

and the corresponding response with a 405 status code is returned

https://github.com/litestar-org/litestar/blob/26f20ac6c52de2b4bf81161f7560c8bb4af6f382/litestar/plugins/problem_details.py#L111-L116

but the corresponding `Allow` header is never set.

The best place to do this AFAICT is in the ASGI Router itself, since it has access to the available methods, perhaps here:

https://github.com/litestar-org/litestar/blob/26f20ac6c52de2b4bf81161f7560c8bb4af6f382/litestar/_asgi/routing_trie/traversal.py#L83-L84

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Allow

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

provinzkraut:
`MethodNotAllowedException` exceptions raised during routing did not include the [`Allow`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Allow) header, which must be present in a `405 Method Not Allowed` response.

Fixes #4277.

### DIFF
--- a/litestar/_asgi/routing_trie/traversal.py
+++ b/litestar/_asgi/routing_trie/traversal.py
@@ -81,7 +81,10 @@ def parse_node_handlers(
     if node.is_asgi:
         return node.asgi_handlers["asgi"]
     if method:
-        return node.asgi_handlers[method]
+        try:
+            return node.asgi_handlers[method]
+        except KeyError as e:
+            raise MethodNotAllowedException(headers={"allowed": ", ".join(node.asgi_handlers.keys())}) from e
     return node.asgi_handlers["websocket"]
 
 
@@ -170,7 +173,5 @@ def parse_path_to_route(
             parsed_path_parameters,
             node.path_template,
         )
-    except KeyError as e:
-        raise MethodNotAllowedException() from e
     except ValueError as e:
         raise NotFoundException() from e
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
Hey team 👋

Is there a timeline for the release of this fix? Neither this PR nor the linked issue are part of any GitHub milestone.

provinzkraut:
This was released in `2.18.0`. The info is available on the issue: https://github.com/litestar-org/litestar/issues/4277#issuecomment-3369164140

edgarrmondragon:
> This was released in `2.18.0`. The info is available on the issue: [#4277 (comment)](https://github.com/litestar-org/litestar/issues/4277#issuecomment-3369164140)

I see what's going on:

- https://github.com/litestar-org/litestar/pull/4539

PR #4290, opened 2025-09-14
fix: docs[ReceiveRoutePlugin] documentation

Shekhrozx:
## Description

Fixed a documentation error: updated `route.http_methods` to the correct attribute `route.methods` when logging registered routes.

before:

```py
    from litestar.plugins import ReceiveRoutePlugin

    class RouteLoggerPlugin(ReceiveRoutePlugin):
        def receive_route(self, route: BaseRoute) -> None:
            print(f"Route registered: {route.path} [{', '.join(route.http_methods)}]")
```

after:

```py
print(f"Route registered: {route.path} [{', '.join(route.methods)}]")
```

### DIFF
--- a/docs/usage/plugins/index.rst
+++ b/docs/usage/plugins/index.rst
@@ -149,4 +149,4 @@ The following example shows a simple plugin that logs information about each rou
 
     class RouteLoggerPlugin(ReceiveRoutePlugin):
         def receive_route(self, route: BaseRoute) -> None:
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Indeed! It is `methods`, not `http_methods`: https://github.com/litestar-org/litestar/blob/262c776ef4ea540caef5e855e545aae062443115/litestar/routes/base.py#L79

Thanks!

PR #4291, opened 2025-09-14
feat(testing)!: Async native test client

### LINKED ISSUE #1920: Bug: the app for test client is running in a different event loop

provinzkraut:
Introduce new async-native testing utilities:

- `AsyncTestClient`
- `AsyncWebSocketTestSession`
- `LifespanHandler`
- `TestClientTransport`

And synchronhous equivalents:

- `TestClient`
- `WebSocketTestSession`
- `SyncTestClientTransport`

The test utilities are not async-first, meaning the implementation is async, and when using the synchronous versions, they will be bridged using `anyio`s [`BlockingPortal`](https://anyio.readthedocs.io/en/stable/api.html#anyio.from_thread.BlockingPortal), whereby they run the async code within a new thread + event loop.

This design removes various barriers and pitfalls, mainly dealing with multiple event loops when running tests asynchronously (e.g. avoiding nested event loops). It also simplifies the code quite a bit and brings a minor speedup when running async tests.

There is currently one issue stemming from a fundamental incompatibility between `pytest-asyncio` and `anyio.CancelScope`, https://github.com/pytest-dev/pytest-asyncio/issues/1191, but I'm working on a fix towards this, with a proof-of-concept already in place, so I think it's not a blocker.

Fixes #1920


litestar/testing/client/async_client.py
```diff
@@ -1,50 +1,61 @@
...
 
+class AsyncTestClient(AsyncClient, Generic[T]):
     def __init__(
         self,
         app: T,
         base_url: str = "http://testserver.local",
         raise_server_exceptions: bool = True,
         root_path: str = "",
-        backend: AnyIOBackend = "asyncio",
-        backend_options: Mapping[str, Any] | None = None,
```
euri10, line 44: why those 2 kwargs are removed from the async client while on the sync one they stay ?

tests/unit/test_connection/test_websocket.py
```diff
@@ -370,48 +393,33 @@ async def handler(socket: WebSocket[Any, Any, State]) -> None:
...
-def test_no_additional_headers() -> None:
-    @websocket("/")
-    async def handler(socket: WebSocket[Any, Any, State]) -> None:
-        await socket.accept()
-        await socket.close()
-
-    with create_test_client(handler).websocket_connect("/") as ws:
-        assert ws.extra_headers == []
-
-
```
euri10, line 396: those 2 tests are not needed anymore ?

euri10 (commented):
could we maybe add those 2 tests that were mentionned in the issue, they now pass successfully :
```
async def test_multiple_clients_event_loop() -> None:
    @get("/")
    def return_loop_id() -> dict:
        return {"loop_id": id(get_event_loop())}

    app = Litestar(route_handlers=[return_loop_id])

    async with AsyncTestClient(app) as client_1, AsyncTestClient(app) as client_2:
        response_1 = await client_1.get("/")
        response_2 = await client_2.get("/")

    assert response_1.json() == response_2.json()


async def test_lifespan_loop() -> None:
    mock = MagicMock()

    @asynccontextmanager
    async def lifespan(app: Litestar) -> AsyncGenerator[None, None]:
        mock(asyncio.get_running_loop())
        yield

    app = Litestar(lifespan=[lifespan])

    async with AsyncTestClient(app):
        pass

    mock.assert_called_once_with(asyncio.get_running_loop())
```

litestar/testing/client/async_client.py
provinzkraut: Well because they don't apply. Since the async test client is async-native, you can't dynamically select an async runner; If you're running your tests in asyncio, you're running your tests in asyncio 🤷‍♀️

tests/unit/test_connection/test_websocket.py
provinzkraut: They have been moved to elsewhere in the file!

euri10 (approved):

PR #4298, opened 2025-09-21
fix: register new CLI commands with `rich-click`

cofin:
After the most recent updates to `rich-click` and `click`, CLI plugins no longer show the command in the command help text.  

This is because the plugin was not loaded before rending the help text.  This change ensures that the plugins are loaded prior to formatting the help output.


dwreeves:
Sorry about that! 😓

PR #4333, opened 2025-09-27
feat(jwt): remove deprecate `litestar.contrib.jwt`

### LINKED ISSUE #4304: Task: Remove deprecated `litestar.contrib.jwt`
Remove deprecated `litestar.contrib.jwt`. See #4301 for details on the process

sobolevn:
Closes #4304


euri10:
just to be sure, can you confirm those "missing" bits are not needed for this module, it seems unlikely to me that the docs part doesnt need a change but I may be mistaken

- [x] Delete the litestar.contrib.<namespace> module
- [ ] Remove tests for the deprecation warnings for importing litestar.contrib.<namespace>
- [ ] If there is documentation in docs/usage referencing the removed module, update it to the new litestar.plugins.<namespace> references
- [ ] If there is a file docs/reference/contrib/<namespace.rst>, delete it. If no corresponding file in docs/reference/plugins/<namespace>.rst exists, move the file there instead of deleting it
- [x] Add an entry in docs/release-notes/changelog.rst (use the outline below)

sobolevn:
- [x] There are no tests for the deprecation :) 
- [x] There's an entry section about `security.jwt`, so no update is needed
- [x] `docs/reference/contrib/jwt.rst` delete :) No `.plugins.jwt` exists, so the move is not needed

PR #4347, opened 2025-09-29
fix: change Optional to NotRequired in OpenAPI pydantic plugin

### LINKED ISSUE #4294: Bug: OpenAPI Schema generator incorrect type for fields with default_factory
### Description

OpenAPI schema generator create incorrect type for fields with `default_factory`.
In MCVE `Foo.foo type` is `str`, but in generated openapi `(string | null)`, which is incorrect

_No response_

### MCVE

```python
from litestar import Litestar, get, post


def foo_factory() -> str:
    return 'foo'


class Foo(BaseModel):
    foo: str = Field(default_factory=foo_factory)
    bar: str = Field(default='bar')
    spam: str


@post('/')
def create_foo(data: Foo) -> None:
    pass


@get('/')
def foo() -> Foo:
    return Foo(spam='spam')


app = Litestar(route_handlers=[create_foo, foo])
```

### Screenshots

Image

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [ ] Linux
- [ ] Mac
- [x] Windows

raidzin:
## Description

Change wrapper from `Optional` to `NotRequired` for **pydantic** fields with _default_factory_

Now fields appear as `string` and not required instead of `(string | null)` 

## Closes
Fixes #4294


raidzin:
@provinzkraut, can i remove this test as not actual?

https://github.com/litestar-org/litestar/blob/2a8f19ec3d096bb3e0e18cd2078d434b1a1b91ef/tests/unit/test_plugins/test_pydantic/test_schema_plugin.py#L135-L150

provinzkraut:
> @provinzkraut, can i remove this test as not actual?

This is a regression test for the linked issue. Since we're still supporting Pydantic v1, that test should not break so no, it cannot be removed

provinzkraut:
@raidzin Can you add a changelog entry?

raidzin:
> @raidzin Can you add a changelog entry?

Done

docs/release-notes/changelog.rst
```diff
@@ -188,3 +188,11 @@
...
         Remove the deprecated ``litestar.contrib.minijinja.minijinja_from_state`` function.
         Instead use a callable that receives the minijinja ``State`` object as first argument.
+
+    .. change:: Change ``Optional`` to ``NotRequired`` for pydantic fields with ``default_factory``
+        :type: bugfix
+        :pr: 4347
+        :issue: 4294
+
+        Now ``pydantic`` fields with ``default_factory`` displayed as not nullable and not required.
+        Before it was nullable and not required.
```
provinzkraut: This description is quite vague "displayed as not nullable" doesn't really convey that this is regarding OpenAPI. You should try to be more specific here

Issue #4303, opened 2025-09-22
Task: Remove deprecated `litestar.contrib.htmx`


provinzkraut:
### Summary

Remove deprecated `litestar.contrib.htmx`

PR #4362, opened 2025-09-30
feat(deps)!: drop `litestar-htmx` from default dependencies

sobolevn:
Since now `litestar-htmx` is the default dependency, there's no need to duplicate this dep twice with the same version.


pyproject.toml
```diff
@@ -113,7 +113,9 @@ full = [
     zstd]; \
     python_version >= \"3.13\"""",
 ]
-htmx = ["litestar-htmx>=0.4.0"]
+# Needed for backward compat, does nothing,
+# because `litestar-hmtx` is a default dependency now:
+htmx = []
```
sobolevn: Not sure if `uv` allows to completely remove this?

provinzkraut:
Mhm. For 3.0 we want to remove this default dependency. For 2 this change makes sense. Can you point this at the `v2` branch?

sobolevn:
Ok, got it. I will change this PR to drop `litestar-htmx` from `dependencies` and install `[htmx]` instead. While I will create a new one to v2 to drop the dep :)

sobolevn:
`v2` does not have `htmx` extra :)

provinzkraut:
Maybe it makes sense to bundle this with #4303 then?

sobolevn:
Yes, sure :) Closing!

PR #4371, opened 2025-09-30
fix(OpenAPI): Fix `Stream` response being treated as `File` response in OpenAPI schema

provinzkraut:
Fix handler returning a `Stream` falsely indicating a file response in the OpenAPI schema:

```json
{
"description": "Stream Response",
"headers": {
  "content-length": {
    "schema": {
      "type": "string"
    },
    "description": "File size in bytes",
    "required": false,
    "deprecated": false
  },
  "last-modified": {
    "schema": {
      "type": "string",
      "format": "date-time"
    },
    "description": "Last modified data-time in RFC 2822 format",
    "required": false,
    "deprecated": false
  },
  "etag": {
    "schema": {
      "type": "string"
    },
    "description": "Entity tag",
    "required": false,
    "deprecated": false
  }
}
```

### DIFF
--- a/litestar/_openapi/responses.py
+++ b/litestar/_openapi/responses.py
@@ -124,7 +124,7 @@ class ResponseFactory:
             response = OpenAPIResponse(content=None, description=self.create_description())
         elif self.field_definition.is_subclass_of(Redirect):
             response = self.create_redirect_response()
-        elif self.field_definition.is_subclass_of((File, Stream)):
+        elif self.field_definition.is_subclass_of(File):
             response = self.create_file_response()
         else:
             media_type = self.route_handler.media_type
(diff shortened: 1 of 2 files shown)


robertlagrant:
How does this interact with the fact that (I believe) we can't return a File object, only an OctetStream, if the file is in memory? I've just updated and it's changed the response type of my application/octet-stream:
image

(Context is this is causing various typescript things to fail in the frontend code that uses the openapi spec, so it's not a non-breaking change in our case.)

provinzkraut:
> How does this interact with the fact that (I believe) we can't return a File object, only an OctetStream, if the file is in memory? I've just updated and it's changed the response type of my application/octet-stream

You can set the content media type manually:

```python
@get("/", content_media_type="application/octet-stream")
async def handler() -> Stream:
 ...
```

This should have the same result.

But now that I think about it, `application/octet-stream` is probably a more sensible default for `Stream`. 

Feel free to open a new issue for this.

PR #4372, opened 2025-09-30
feat(core)!: Add zero-cost middleware exclusion

provinzkraut:
Middlewares inheriting from `ASGIMiddleware` will now have zero runtime cost when they are excluded e.g. via the `scope` or `exclude_opt_key` options.

Previously, the base middleware was always being invoked for every request, evaluating the exclusion criteria, and then calling the user defined middleware functions. If a middleware had defined `scopes = (ScopeType.HTTP,)`, it would still be called for *every* request, regardless of the scope type. Only for requests with the type `HTTP`, it would then call the user's function.

With *zero cost exclusion*, the exclusion is being evaluated statically. At app creation time, when route handlers are registered and their middleware stacks are being built, a middleware that is to be excluded will simply not be included in the stack.

I've marked this change as breaking, even though no runtime behaviour difference is expected. Some test cases may break though if they relied on the fact that the middleware wrapper created by `ASGIMiddleware` was always being called, e.g. if using a mock or trying to observe side-effects of this behaviour.


litestar/middleware/base.py
```diff
@@ -215,27 +215,51 @@ async def handle(
         ScopeType.WEBSOCKET,
         ScopeType.ASGI,
     )
+    """Scope types this middleware should be applied to"""
     exclude_path_pattern: str | tuple[str, ...] | None = None
+    r"""
```
euri10, line 220:
```suggestion
    """
```

litestar/middleware/base.py
```diff
@@ -215,27 +215,51 @@ async def handle(
...
+    Exclude this middleware for all route handlers that have at least one path
+    matching the provided RegEx patterns.
+
+    The matching will be done against the path(s) defined on the *handler*, not the
+    resolved path of the request. This distinction is important, as they can differ for
+    dynamic paths. A handler with the path ``/user/{user_id:int}/`` would receive
+    requests with paths like ``/user/1234/``, so a pattern to exclude that handler would
+    have to be something like ``/user/.+?/``, and not ``/user/\d+/``, even if the latter
+    would match the path of an incoming request accepted by this handler.
```
euri10: mmm I find this hard to understand at first glance, maybe if we add what a ``/user/\d+/`` would lead to could be helpful,

pyproject.toml
```diff
@@ -429,6 +429,7 @@ lint.ignore = [
   "A005",    # Module shadows a built-in Python standard library module
   "PLC0415", # `import` should be at the top-level of a file
   "PLW1641", # Object does not implement `__hash__` method
+  "SIM103",  # 'Return the condition directly': This is a bit overzealous
```
euri10, line 432: unrelated ?

tests/unit/test_middleware/test_base_middleware.py
```diff
@@ -211,19 +213,63 @@ def handler() -> dict:
         assert response.status_code == HTTP_400_BAD_REQUEST
 
 
+@pytest.mark.parametrize(
+    "allowed_scopes,expected_calls",
+    [
+        ((ScopeType.HTTP,), ["/http"]),
+        ((ScopeType.HTTP, ScopeType.ASGI), ["/http", "/asgi"]),
```
euri10: not a crtic, but is there a reason we need those, I mean the scope tuples vs just the scope ?
if yes thenwhy we dont have (http, ws) ?

litestar/middleware/base.py
provinzkraut: This is actually necessary because of the regexes used in the docstring. Otherwise I'd have to double escape the backslashes, which is bad for readability.

pyproject.toml
provinzkraut: It wanted me to refactor a method I added but IMO the suggestion is bad, so I turned off the rule :woman_shrugging:

litestar/middleware/base.py
euri10: alright my bad ;)

tests/unit/test_middleware/test_base_middleware.py
provinzkraut: I just thought it's a quick way to cover all the different branches in the function. I've added the missing test cases for completeness' sake!

litestar/middleware/base.py
provinzkraut: I've tried rewording it

pyproject.toml
euri10: oh it's in the ignore, I misread this as an addition...

euri10 (approved):

litestar/middleware/base.py
```diff
@@ -215,27 +215,79 @@ async def handle(
...
+        exclude_path_pattern = "/user/{user_id:int}/profile"
+        exclude_path_pattern = "/user/\{.+?\}/"
+    """
     exclude_opt_key: str | None = None
     constraints: MiddlewareConstraints | None = None
 
+    def should_bypass_for_handler(self, handler: RouteHandlerType) -> bool:
+        """Return ``True`` if this middleware should be bypassed for ``handler``, according
+        to ``scopes``, ``exclude_path_pattern`` or ``exclude_opt_key``, otherwise `
+        `False``.
```
JacobCoffee:
```suggestion
        to ``scopes``, ``exclude_path_pattern`` or ``exclude_opt_key``, otherwise 
        ``False``.
```

JacobCoffee: I don’t know if this is github rendering weird or not

PR #4374, opened 2025-10-01
fix(channels/plugin): fix typo with pub_task cancelling

### LINKED ISSUE #4373: Bug: The `_on_shutdown` method of ChannelsPlugin has issues with task cancellation
### Description

**Problems:**
- Line 346: `self._sub_task = None` should be `self._pub_task = None`

https://github.com/litestar-org/litestar/blob/8837e8a989e810f80bccf4c59ed9ec0da1c37894/litestar/channels/plugin.py#L346
```python
        if self._sub_task:
            self._sub_task.cancel()
            with suppress(CancelledError):
                await self._sub_task
            self._sub_task = None

        if self._pub_task:
            self._pub_task.cancel()
            with suppress(CancelledError):
                await self._pub_task
            self._sub_task = None                  <-------------------- Here!!!
```

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

Peopl3s:
## Description

Fix the variable assignment bug in `_on_shutdown` ChannelsPlugin

## Closes #4373

### DIFF
--- a/litestar/channels/plugin.py
+++ b/litestar/channels/plugin.py
@@ -343,7 +343,7 @@ class ChannelsPlugin(InitPlugin, AbstractAsyncContextManager):
             self._pub_task.cancel()
             with suppress(CancelledError):
                 await self._pub_task
-            self._sub_task = None
+            self._pub_task = None
 
         await self._backend.on_shutdown()
 


Peopl3s:
> good catch, looks logical i'm curious how did you find that inconsistency ?

@euri10 I wanted to see how ChannelsPlugin is implemented under the hood and accidentally stumbled upon this

litestar/channels/plugin.py
```diff
@@ -343,7 +343,7 @@ async def _on_shutdown(self) -> None:
             self._pub_task.cancel()
             with suppress(CancelledError):
                 await self._pub_task
-            self._sub_task = None
+            self._pub_task = None
```
sobolevn, line 346: Can we test it?

Peopl3s: Yeah, sure. Added some tests

tests/unit/test_channels/test_plugin_shutdown.py
provinzkraut, line 1: There's a `test_plugin.py` module already that contains tests for shutdown. Having a separate file to test this one thing isn't really necessary, can you please move these tests into that file?

https://github.com/litestar-org/litestar/blob/9925defea31768fddabc6d04bb92f531a34defcf/tests/unit/test_channels/test_plugin.py#L415

tests/unit/test_channels/test_plugin_shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+async def test_shutdown_clears_task_references(memory_backend: MemoryChannelsBackend) -> None:
+    plugin = ChannelsPlugin(backend=memory_backend, arbitrary_channels_allowed=True)
+
+    await plugin._on_startup()
+
+    pub_task = plugin._pub_task
+    sub_task = plugin._sub_task
+    pub_queue = plugin._pub_queue
```
provinzkraut: Why this assignment? It only adds noise and makes it harder to follow the test; You now have to keep track of where this was assigned from

tests/unit/test_channels/test_plugin_shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+
+    assert pub_task is not None
+    assert sub_task is not None
+    assert pub_queue is not None
```
provinzkraut: These assertions aren't really necessary

tests/unit/test_channels/test_plugin_shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+
+    await plugin._on_shutdown()
+
+    assert pub_task.cancelled() or pub_task.done()
+    assert sub_task.cancelled() or sub_task.done()
```
provinzkraut: Why these assertions? According to the test name, task cancellation isn't what's being tested here

tests/unit/test_channels/test_plugin_shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+    assert plugin._pub_task is None
+    assert plugin._sub_task is None
+    assert plugin._pub_queue is None
+
+
+async def test_shutdown_idempotent_with_task_verification(memory_backend: MemoryChannelsBackend) -> None:
```
provinzkraut: This is partly duplicated with https://github.com/litestar-org/litestar/blob/9925defea31768fddabc6d04bb92f531a34defcf/tests/unit/test_channels/test_plugin.py#L415

tests/unit/test_channels/test_plugin_shutdown.py
provinzkraut, line 1: I suggest a different approach for these tests. *If* we want to test this behaviour, it should only show that repeatedly going through a start / stop cycle works as expected. Right now you are testing the implementation, not the behaviour, which means that any change in implementation will require to adjust all these tests as well. One test here that demonstrates a startup/shutdown loop should suffice.

provinzkraut (changes requested):

tests/unit/test_channels/test_plugin_shutdown.py
Peopl3s: Fair. Maybe something like [that](https://github.com/litestar-org/litestar/pull/4374/commits/027ac78fcbebf31b3048809545e47f19c72d4a94)?

tests/unit/test_channels/test_plugin.py
```diff
@@ -420,3 +420,33 @@ async def test_shutdown_idempotent(memory_backend: MemoryChannelsBackend) -> Non
...
+
+async def test_startup_shutdown_cycle(memory_backend: MemoryChannelsBackend) -> None:
+    plugin = ChannelsPlugin(backend=memory_backend, arbitrary_channels_allowed=True)
+
+    for _ in range(3):
+        await plugin._on_startup()
+
+        subscriber = await plugin.subscribe("test_channel")
+        plugin.publish(b"test_message", "test_channel")
+        await asyncio.sleep(0.1)
```
provinzkraut: This can become flaky. You should use `wait_published` instead

PR #4376, opened 2025-10-01
test: allow tests to be run regardless of root directory

euri10:
## Description

-


tests/unit/test_file_system.py
```diff
@@ -69,7 +69,7 @@ def setup_http_server(
     tmp_dir: pathlib.Path,
     file_path: pathlib.Path,
 ) -> Generator[None, None, None]:
-    server_file_path = pathlib.Path("tests/nginx_test_files").joinpath(file_path.name)
+    server_file_path = (pathlib.Path(__file__).parent.parent / "nginx_test_files").joinpath(file_path.name)
```
provinzkraut, line 72: Mhm. Maybe use `importlib.resources` instead, to make this less brittle?

euri10: it's brittle only if you think we may change the test dir structure no ?

we alredy use `Path(__file___)` in tests in some places : 

20251001_1311_661x625_1759317070

provinzkraut (approved):

PR #4389, opened 2025-10-03
fix(channels/backend/memory): add KeyError-safe delete channel from history

### LINKED ISSUE #4386: Bug: KeyError is not handled in MemoryChannelsBackend.unsubscribe
### Description

litestar/channels/backends/memory.py

https://github.com/litestar-org/litestar/blob/a733629e86a302a957d8ebb58df30ba4cf3ecf3d/litestar/channels/backends/memory.py#L63
```python
        """Unsubscribe from ``channels``"""
        self._channels -= set(channels)
        try:
            for channel in channels:
                del self._history[channel] <---------  There may be a KeyError at each iteration
        except KeyError:
            pass
```

The try-except block covers the entire cycle, but at the first KeyError, the cycle is interrupted, and the remaining channels are not deleted from the history.

Effects:
* Incomplete history cleanup
* Memory leak with frequent subscriptions/unsubscriptions
* Inconsistent state between _channels and _history

## Fix
```python
    self._channels -= set(channels)
    for channel in channels:
        self._history.pop(channel, None)  # safe
```

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

Peopl3s:
## Description

Add KeyError-safe delete channel from history)

Closes #4386

### DIFF
--- a/litestar/cli/commands/core.py
+++ b/litestar/cli/commands/core.py
@@ -87,6 +87,7 @@ def _run_uvicorn_in_subprocess(
         "port": port,
         "workers": workers,
         "factory": env.is_app_factory,
+        "app-dir": env.cwd,
     }
     if fd is not None:
         process_args["fd"] = fd
--- a/tests/unit/test_cli/test_core_commands.py
+++ b/tests/unit/test_cli/test_core_commands.py
@@ -176,6 +176,7 @@ def test_run_command(
             f"{path.stem}:app",
             f"--host={host}",
             f"--port={port}",
+            f"--app-dir={path.parent}",
         ]
         if fd is not None:
             expected_args.append(f"--fd={fd}")
--- a/tests/unit/test_cli/test_ssl.py
+++ b/tests/unit/test_cli/test_ssl.py
@@ -320,6 +320,7 @@ def test_arguments_passed(
             "--host=127.0.0.1",
             "--port=8000",
             "--workers=2",
+            f"--app-dir={path.parent}",
         ]
         if ssl_certfile is not None:
             expected_args.append(f"--ssl-certfile={project_path / ssl_certfile}")


provinzkraut:
Can you add a test that would fail with the old behaviour and passes with the new one? Ideally add the issue link as a comment as well

Peopl3s:
> Can you add a test that would fail with the old behaviour and passes with the new one? Ideally add the issue link as a comment as well

Sure

litestar/channels/backends/memory.py
```diff
@@ -59,7 +59,7 @@ async def unsubscribe(self, channels: Iterable[str]) -> None:
         """Unsubscribe from ``channels``"""
         self._channels -= set(channels)
         for channel in channels:
-            self._history.pop(channel, None)
+            self._history.pop(channel, None)  # https://github.com/litestar-org/litestar/issues/4386
```
provinzkraut:
```suggestion
            self._history.pop(channel, None)
```
Adding it to the test is good enough :)

tests/unit/test_channels/test_backends.py
```diff
@@ -219,3 +219,35 @@ async def test_memory_backend_stream_before_startup_raises() -> None:
...
+    # Verify that a channel without history doesn't exist in _history dict
+    assert "no_history" not in backend._history
+
+    # Unsubscribe from channels including one that doesn't exist in _history
+    # The order matters: "no_history" doesn't exist in _history, so old code with
+    # del would raise KeyError and stop, leaving "bar" and "baz" history intact
+    await backend.unsubscribe(["foo", "no_history", "bar", "baz"])
+
+    # With the fix using pop(channel, None), all channels should have their history cleared
+    # https://github.com/litestar-org/litestar/issues/4386
```
provinzkraut: Can you put this on top / above the test case? That's the convention we usually follow

Peopl3s: @provinzkraut Done

PR #4394, opened 2025-10-03
fix!: ensure that generator clean up is correctly propagated

### LINKED ISSUE #3772: Bug: No cleanup for SSE-endpoints if client tears connection down
### Description

This leads to resource leakage. Possibly relevant to #2958.

https://github.com/vrslev/litestar-sse-cleanup-issue

### MCVE

_No response_

### Steps to reproduce

```bash
1. Follow the instructions in README in the repository
2. See that logs do not contain `closed redis client`—which is expected.
```

### Screenshots

_No response_

### Logs

```bash
INFO:     Started server process [54771]
INFO:     127.0.0.1:54353 - "GET /sse HTTP/1.1" 200 OK
initialized redis client
caught an exception from blpop: Cancelled by cancel scope 1041246b0
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [54771]
```

### Litestar Version

2.12.1

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

winstxnhdw:
## Description

As discussed on Discord, we should ensure that synchronous generators are properly cleaned up. Async generators will  be cleaned up, so we should do the same for synchronous generators as well.

The following snippet will now properly print `"cleaned up inner"` on premature disconnects.

```python
from litestar import get
from litestar.response.sse import ServerSentEvent
from litestar.background_tasks import BackgroundTask
import time

@get('/test', sync_to_thread=True)
def test() -> ServerSentEvent:
    def inner_gen():
        try:
            for i in range(10):
                yield str(i)
                time.sleep(1)
        finally:
            print('cleaned up inner')

    def gen():
        igen = inner_gen()
        try:
            yield from igen
        except ClientDisconnectException:
            return 
        finally:
            igen.close()

    g = gen()
    return ServerSentEvent(g)
```

Closes #3772

### DIFF
--- a/litestar/utils/sync.py
+++ b/litestar/utils/sync.py
@@ -23,7 +23,7 @@ T = TypeVar("T")
 S = TypeVar("S", default=None)
 
 
-def iterable_to_generator(iterable: Iterable[T]) -> Generator[T, S, None]:
+def _iterable_to_generator(iterable: Iterable[T]) -> Generator[T, S, None]:
     """Convert an iterable to a generator.
 
     Args:
@@ -84,7 +84,7 @@ class AsyncIteratorWrapper(AsyncGenerator[T, S]):
         elif isinstance(iterator, AsyncIteratorWrapper):
             self._original_generator = iterator._original_generator
         else:
-            self._original_generator = iterable_to_generator(iterator)
+            self._original_generator = _iterable_to_generator(iterator)
 
         self.iterator = iter(iterator)
         self.generator = self._async_generator()
@@ -121,10 +121,9 @@ class AsyncIteratorWrapper(AsyncGenerator[T, S]):
         tb: TracebackType | None = None,
     ) -> T:
         try:
-            return (
-                await sync_to_thread(self._original_generator.throw, typ)
-                if isinstance(typ, BaseException)
-                else await sync_to_thread(self._original_generator.throw, typ, val, tb)
+            return await sync_to_thread(
+                self._original_generator.throw,
+                typ if isinstance(typ, BaseException) else typ(val, tb),
             )
         except StopIteration as e:
             raise StopAsyncIteration from e
--- a/tests/unit/test_utils/test_async_iterator_wrapper.py
+++ b/tests/unit/test_utils/test_async_iterator_wrapper.py
@@ -1,0 +1,108 @@
+from collections.abc import AsyncIterator, Generator
+from typing import TypeVar, Union
+
+from pytest import raises
+
+from litestar.response.streaming import ClientDisconnectError
+from litestar.utils.sync import AsyncIteratorWrapper
+
+T = TypeVar("T")
+
+
+class GeneratorException(Exception): ...
+
+
+async def anext(iterator: AsyncIterator[T]) -> T:
+    return await iterator.__anext__()
+
+
+def generator(state: list[int], iterations: int) -> Generator[int, Union[int, None]]:
+    try:
+        for _ in range(iterations):
+            received = yield state[0]
+            state[0] += 1
+
+            if received is not None:
+                state.append(received)
+
+    except ClientDisconnectError as error:
+        raise GeneratorException from error
+
+    finally:
+        state.append(-1)
+
+
+async def test_async_iterator_wrapper() -> None:
+    state = [0]
+    iterations = 10
+    async_iterator = AsyncIteratorWrapper(generator(state, iterations))
+
+    async for i in async_iterator:
+        assert state[0] == i
+
+    assert state[-1] == -1
+
+
+async def test_async_iterator_wrapper_aclose() -> None:
+    state = [0]
+    async_iterator = AsyncIteratorWrapper(generator(state, 10))
+
+    await anext(async_iterator)
+    await async_iterator.aclose()
+
+    assert state[-1] == -1
+
+
+async def test_async_iterator_wrapper_asend() -> None:
+    state = [0]
+    async_iterator: AsyncIteratorWrapper[int, int] = AsyncIteratorWrapper(generator(state, 10))
+    await anext(async_iterator)
+
+    assert await async_iterator.asend(10) == 1
+    assert state[-1] == 10
+
+
+async def test_async_iterator_wrapper_athrow() -> None:
+    state = [0]
+    async_iterator = AsyncIteratorWrapper(generator(state, 10))
+    await anext(async_iterator)
+
+    with raises(GeneratorException):
+        await async_iterator.athrow(ClientDisconnectError)
+
+    assert state[-1] == -1
+
+
+async def async_iterator_wrapper_from_iterable() -> None:
+    iterator = iter(range(3))
+    async_iterator = AsyncIteratorWrapper(iterator)
+
+    assert [i async for i in async_iterator] == [0, 1, 2]
+
+
+async def async_iterator_wrapper_from_iterable_aclose() -> None:
+    iterator = iter(range(3))
+    async_iterator = AsyncIteratorWrapper(iterator)
+
+    await anext(async_iterator)
+    await async_iterator.aclose()
+
+    with raises(StopAsyncIteration):
+        await anext(async_iterator)
+
+
+async def async_iterator_wrapper_from_iterable_asend() -> None:
+    iterator = iter(range(3))
+    async_iterator: AsyncIteratorWrapper[int, int] = AsyncIteratorWrapper(iterator)
+    await anext(async_iterator)
+
+    assert await async_iterator.asend(10) == 1
+
+
+async def async_iterator_wrapper_from_iterable_athrow() -> None:
+    iterator = iter(range(3))
+    async_iterator = AsyncIteratorWrapper(iterator)
+    await anext(async_iterator)
+
+    with raises(StopAsyncIteration):
+        await async_iterator.athrow(StopIteration)


litestar/utils/sync.py
provinzkraut, line 1: I don't think the wrapper should automatically close the generator. `AsyncIteratorWrapper` should be simply that; An async wrapper for a synchronous generator. If you want to close the generator manually, that should happen at the call site. 

The current implementation also introduces a discrepancy in behaviour: Sync generators will have their `close` called proactively, while async ones won't.

Also, the implementation should work without a lock. Acquiring a lock at every iteration step isn't great for performance. (and if a lock is used, it should be `anyio.Lock`)

provinzkraut (changes requested):

litestar/utils/sync.py
winstxnhdw: Now that I think about it, we probably don't need a lock. A boolean flag is sufficient.

> The current implementation also introduces a discrepancy in behaviour: Sync generators will have their close called proactively, while async ones won't.

I can add an `aclose` case but this is already handled by the GC.

> If you want to close the generator manually, that should happen at the call site.

You mean the route handler? Because the call site is certainly here, in this SSE. Or do you mean in `_ServerSentEventIterator`?

provinzkraut: > I can add an `aclose` case but this is already handled by the GC.

As I said, it's about explicitness. Maybe the GC does so for built-in generators, but a user could also supply their own thing that implements e.g. `aclose`, and is not automagically cleaned up :woman_shrugging: 

> You mean the route handler? Because the call site is certainly here, in this SSE.

Yes, the route handler. You could e.g. add an `aclose` method on the wrapper, that calls `close` if it's a generator.

litestar/utils/sync.py
```diff
@@ -54,29 +55,39 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
-                return
+        except ValueError:
+            return
+        finally:
+            while self.is_in_thread:
+                pass
+            if isinstance(self.old_iterator, Generator):
+                self.old_iterator.close()
+            elif isinstance(self.old_iterator, AsyncGenerator):
+                await self.old_iterator.aclose()
```
provinzkraut: This can never be an `AsyncGenerator`, as this wrapper only handles sync iterators / iterables

winstxnhdw:
I believe this is closer to what you want? We make `AsyncIteratorWrapper` into an actual async generator. Then handle the clean up within the SSE wrapper. This way the wrapper remains agnostic. We also proactively handle the clean up for both async and sync generators.

In the end, this should probably be considered a fix since this fixes a discrepancy in clean up between async and sync generators.

winstxnhdw:
In the end, this should probably be considered a fix since this fixes a discrepancy in clean up between async and sync generators.

litestar/response/sse.py
```diff
@@ -18,11 +18,28 @@
 _LINE_BREAK_RE = re.compile(r"\r\n|\r|\n")
 DEFAULT_SEPARATOR = "\r\n"
 
+T = TypeVar("T")
 
-class _ServerSentEventIterator(AsyncIteratorWrapper[bytes]):
+
+async def async_iterator_to_generator(
+    async_iterator: AsyncIterator[T] | AsyncIterable[T],
```
sobolevn:
```suggestion
    async_iterator: AsyncIterable[T],
```

`AsyncIterable` is the base class of `AsyncIterator`

litestar/utils/sync.py
```diff
@@ -51,25 +55,30 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
         """
-        self.iterator = iterator if isinstance(iterator, Iterator) else iter(iterator)
+        self.is_in_thread = False
+        self.original_iterator = iterator
+        self.iterator = iter(iterator)
         self.generator = self._async_generator()
 
     def _call_next(self) -> T:
         try:
+            self.is_in_thread = True
```
sobolevn: we can move `self.is_in_thread = True` from `try` body, can't we?

litestar/utils/sync.py
```diff
@@ -51,25 +55,30 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
+    __slots__ = ("generator", "is_in_thread", "iterator", "original_iterator")
 
     def __init__(self, iterator: Iterator[T] | Iterable[T]) -> None:
         """Take a sync iterator or iterable and yields values from it asynchronously.
 
         Args:
             iterator: A sync iterator or iterable.
         """
-        self.iterator = iterator if isinstance(iterator, Iterator) else iter(iterator)
+        self.is_in_thread = False
```
sobolevn: Do we want these new attributes to be public?

litestar/utils/sync.py
```diff
@@ -51,25 +55,30 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
         return sync_to_thread(self.func, *args, **kwargs)  # type: ignore[arg-type]
 
 
-class AsyncIteratorWrapper(Generic[T]):
+class AsyncGeneratorWrapper(AsyncGenerator[T, S]):
```
sobolevn: Adding type vars without a default is a breaking change for people who use strict typing :(

litestar/response/sse.py
winstxnhdw: You're right. Good catch.

litestar/utils/sync.py
winstxnhdw: I forgot I can just use `typing_extensions` for default `TypeVar` before `<3.13`

litestar/utils/sync.py
```diff
@@ -51,25 +54,30 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
         self.generator = self._async_generator()
 
     def _call_next(self) -> T:
+        self._is_in_thread = True
         try:
             return next(self.iterator)
         except StopIteration as e:
             raise ValueError from e
+        finally:
+            self._is_in_thread = False
```
winstxnhdw: I should also mention that this synchronisation flag is not strictly required. Without it, the generator still manages to clean up but an exception is thrown. If it were up to me, I would just hide the exception with a `try-except`.

tests/unit/test_response/test_sse.py
```diff
@@ -96,6 +96,30 @@ async def numbers() -> AsyncIterator[SSEData]:
                 assert events[i].retry == expected_events[i].retry
 
 
+async def test_sse_cleanup() -> None:
```
provinzkraut: This test does not work. It passes on `main` as well

provinzkraut (changes requested):
After thinking about it some more, I don't think `close`ing is the right thing to do here. Instead, we should `athrow`/`throw` into the underlying generators in any case instead of trying to cancel the iterating task by using `cancel_scope.cancel` while listening for disconnects. That would be the proper way to handle things, and also avoid the whole "terminating a thread" issue cleanly. 

If we `(a)throw` e.g. a `ClientDisconnect` into the (async)generators, users can also decide how they want to handle that cases specifically, or decide not to handle it at all.

It would also clean up the implementation I believe.

litestar/response/streaming.py
```diff
@@ -122,10 +153,11 @@ async def send_body(self, send: Send, receive: Receive) -> None:
         Returns:
             None
         """
+        disconnect_event = Event()
 
         async with create_task_group() as task_group:
-            task_group.start_soon(partial(self._stream, send))
-            await self._listen_for_disconnect(cancel_scope=task_group.cancel_scope, receive=receive)
+            task_group.start_soon(partial(self._listen_for_disconnect, disconnect_event, receive))
+            await self._stream(disconnect_event, send)
```
winstxnhdw: I decided to use `anyio.Event` here instead of `CancelScope` so that we can keep `athrow` and the iteration of the generator within the same thread. Otherwise, we would hit the typical runtime error of closing a generator while it is running in another thread.

litestar/response/streaming.py
```diff
@@ -70,39 +89,51 @@ def __init__(
...
 
         Returns:
             None
         """
         async for chunk in self.iterator:
+            if disconnect_event.is_set():
+                try:
+                    await self.iterator.athrow(ClientDisconnectException)
+                except (ClientDisconnectException, StopAsyncIteration):
+                    break
```
winstxnhdw: If user decides not to handle the disconnect, and it propagates back here, we ignore it.

litestar/exceptions/http_exceptions.py
```diff
@@ -97,6 +99,12 @@ class ClientException(HTTPException):
     status_code: int = HTTP_400_BAD_REQUEST
 
 
+class ClientDisconnectException(HTTPException):
```
provinzkraut: This is not a client exception. These exceptions are meant to be raised by library/user code and turned into HTTP error responses. Since this exception is raised after the connection is closed, it cannot be turned into an error response anymore

winstxnhdw (commented):
This was particularly tricky to do but now this PR handles streaming and SSE cases as well. Also no longer marked as breaking because by default, all behaviour should be the same.

litestar/response/streaming.py
provinzkraut: The `break` needs to happen unconditionally; Generators are not required to re-raise `ClientDisconnectException`, but we want to terminate in any case

tests/unit/test_response/test_sse.py
winstxnhdw: Do you have any suggestions for how we can even test this?

litestar/response/streaming.py
```diff
@@ -70,39 +89,51 @@ def __init__(
...
-            if message["type"] == "http.disconnect":
-                # despite the IDE warning, this is not a coroutine because anyio 3+ changed this.
-                # therefore make sure not to await this.
-                cancel_scope.cancel()
-            else:
-                await self._listen_for_disconnect(cancel_scope=cancel_scope, receive=receive)
-
-    async def _stream(self, send: Send) -> None:
+        while (message := await receive()) and not message["type"].endswith(".disconnect"):
+            pass
```
provinzkraut: I prefer the non-linlined version; It's more readable. With this, the intent isn't quite clear, as it looks just like a busy loop at first glance. Having a `while True`/`break` makes it more clear what's happening

litestar/response/streaming.py
```diff
@@ -70,39 +89,51 @@ def __init__(
...
+            self.iterator = AsyncGeneratorWrapper(iterator)
+        elif isinstance(iterator, (AsyncIterable, AsyncIterator)):
+            self.iterator = async_iterator_to_generator(iterator)
+
+        self._original_generator = (
+            iterator.content_async_iterator if isinstance(iterator, ServerSentEventIterator) else self.iterator
         )
 
-    async def _listen_for_disconnect(self, cancel_scope: CancelScope, receive: Receive) -> None:
+    async def _listen_for_disconnect(self, disconnect_event: Event, receive: Receive) -> None:
```
provinzkraut: Consider setting the event on the instance itself. Not really necessary to pass it around like this imo

litestar/response/streaming.py
```diff
@@ -70,39 +89,51 @@ def __init__(
...
+        if isinstance(iterator, (AsyncGeneratorWrapper, AsyncGenerator)):
+            self.iterator = iterator
+        elif isinstance(iterator, (Iterable, Iterator)):
+            self.iterator = AsyncGeneratorWrapper(iterator)
+        elif isinstance(iterator, (AsyncIterable, AsyncIterator)):
+            self.iterator = async_iterator_to_generator(iterator)
+
+        self._original_generator = (
+            iterator.content_async_iterator if isinstance(iterator, ServerSentEventIterator) else self.iterator
         )
```
winstxnhdw: There's now an issue with circular imports because I depend on `ServerSentEventIterator` here.

We can either split the files to prevent this issue, or pass `content_async_iterator` as an argument from `ServerSentEventIterator` to `ASGIStreamingResponse`.

litestar/utils/sync.py
```diff
@@ -51,25 +72,33 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
         return sync_to_thread(self.func, *args, **kwargs)  # type: ignore[arg-type]
 
 
-class AsyncIteratorWrapper(Generic[T]):
+class AsyncGeneratorWrapper(AsyncGenerator[T, S]):
```
provinzkraut: Renaming this is a breaking change

litestar/utils/sync.py
```diff
@@ -78,8 +107,39 @@ async def _async_generator(self) -> AsyncGenerator[T, None]:
...
+    def __aiter__(self) -> AsyncGeneratorWrapper[T, S]:
         return self
 
     async def __anext__(self) -> T:
         return await self.generator.__anext__()
+
+    async def aclose(self) -> None:
+        if isinstance(self._original_generator, Generator):
+            while self._is_in_thread:
+                pass
```
provinzkraut: Please remove this code and the other instances like it. I don't think we should implicitly try to make (async) generators thread safe. Instead, we should just handle them correctly

litestar/utils/sync.py
```diff
@@ -1,23 +1,44 @@
...
+        iterator: An iterable.
+
+    Returns:
+        A generator.
+    """
+    for item in iterator:
+        try:
+            yield item
+
+        except ClientDisconnectException:
```
provinzkraut: This should not be handled at this place. It's a general utility, so it should not be concerned with streaming semantics. 

Also, I'm not sure this is required at all? If the response gets passed an async iterator, there's nothing to `throw` into to begin with, and all this wrapper does is stop iterating, which is something the calling code does anyway on disconnect

winstxnhdw: Sorry, this is leftovers from testing. Will remove.

litestar/response/streaming.py
winstxnhdw: Is this instance thread-safe?

litestar/response/streaming.py
winstxnhdw: My thought process was that if there was an unexpected exception raised somewhere along the way, this would throw. But I don't feel too strongly about this.

litestar/response/streaming.py
winstxnhdw: Seems like it is since `ServerSentEvent` is only created once per handler. Alright, will change.

provinzkraut: Well no. But it's also never shared. There can only be one consumer of the response, which is the ASGI app calling it. This is the general convention being followed

litestar/response/streaming.py
provinzkraut: Right. But you are only calling `break` in the `except` part, which means that if a generator handles `ClientDisconnectException` and does not re-raise it, `break` never gets called. In most cases, this would probably then result in an `StopAsyncIteration` being raises, but that's also not guaranteed.

winstxnhdw: Hmm.. Is that so? In my tests, even if I just `pass` under the `ClientDisconnectException` handler, it still breaks fine. I think `ClientDisconnectException` will always be raised on `athrow` even if it was caught by the upstream generator.

```python
from asyncio import run


def gen():
    try:
        yield 1
        yield 2
        yield 3
    except ZeroDivisionError:
        pass
    finally:
        print('cleaned up outer finally')  # noqa: T201


async def test():
    g = gen()
    for item in g:
        try:
            g.throw(ZeroDivisionError)
        except (StopIteration, ZeroDivisionError):
            print('tada')  # this will print
            break


run(test())
```

EDIT: Oh nvm, it is indeed raising `StopIteration`

litestar/response/streaming.py
```diff
@@ -132,8 +132,8 @@ async def _stream(self, send: Send) -> None:
             if self.disconnect_event.is_set():
                 try:
                     await self.iterator.athrow(ClientDisconnectException)
-                except (ClientDisconnectException, StopAsyncIteration):
-                    break
+                finally:
+                    break  # noqa: B012
```
provinzkraut, line 117: This won't work if the generator re-raises `ClientDisconnectException`, as then there will be an unhandled exception

litestar/response/streaming.py
```diff
@@ -3,12 +3,14 @@
...
-from typing import TYPE_CHECKING, Any, Callable, Union
+from typing import TYPE_CHECKING, Any, Callable, TypeVar, Union
 
-from anyio import CancelScope, create_task_group
+from anyio import Event, create_task_group
 
 from litestar.enums import MediaType
+from litestar.exceptions import ClientDisconnectException
 from litestar.response.base import ASGIResponse, Response
+from litestar.response.sse import ServerSentEventIterator
```
provinzkraut: This module shouldn't need to know about `ServerSentEventIterator`. It should be compatible with all iterators / generators generally

litestar/response/streaming.py
```diff
@@ -103,6 +131,13 @@ async def _stream(self, send: Send) -> None:
             None
         """
         async for chunk in self.iterator:
+            if self.disconnect_event.is_set():
+                try:
+                    await self.iterator.athrow(ClientDisconnectError)
+                except BaseException:  # noqa: BLE001, S110
```
provinzkraut: This will now suppress *all* exceptions.

What we want is this:

- If a client disconnects, throw `ClientDisconnectError` into the generator
- If the generator re-raises that exception, ignore it
- Stop consuming the generator

tests/unit/test_response/test_sse.py
provinzkraut: You'd need to use the `subprocess_test_client` in order to consume a stream and abort it prematurely

litestar/response/streaming.py
```diff
@@ -70,28 +91,35 @@ def __init__(
...
+
+        self.disconnect_event = Event()
+        self.iterator: AsyncGenerator[str | bytes, None]
+
+        if isinstance(iterator, (AsyncIteratorWrapper, AsyncGenerator)):
+            self.iterator = iterator
+        elif isinstance(iterator, (Iterable, Iterator)):
+            self.iterator = AsyncIteratorWrapper(iterator)
+        elif isinstance(iterator, (AsyncIterable, AsyncIterator)):
+            self.iterator = async_iterator_to_generator(iterator)
```
provinzkraut: As I said before, this isn't doing anything useful. If it's an iterator, you're only `athrow`ing the exceptions into the wrapper created by `async_iterator_to_generator`, which will simply stop generating and raise the exception back. So except from adding some overhead, this has no effect. Instead, you should just check if `self.iterator` *is* a generator when reaching the point you want to `athrow`, and only then do so. This also has the benefit of not breaking the interface of the class by restricting `self.iterator` to be an `AsyncGenerator` only.

litestar/utils/sync.py
```diff
@@ -78,8 +102,29 @@ async def _async_generator(self) -> AsyncGenerator[T, None]:
...
 
-    def __aiter__(self) -> AsyncIteratorWrapper[T]:
+    def __aiter__(self) -> AsyncIteratorWrapper[T, S]:
         return self
 
     async def __anext__(self) -> T:
         return await self.generator.__anext__()
+
+    async def aclose(self) -> None:
+        self._original_generator.close()
```
provinzkraut: Needs to use `sync_to_thread`, otherwise that call might be blocking

litestar/utils/sync.py
```diff
@@ -78,8 +102,29 @@ async def _async_generator(self) -> AsyncGenerator[T, None]:
...
+
+    async def asend(self, value: S) -> T:
+        return self._original_generator.send(value)
```
provinzkraut: Needs to use `sync_to_thread`, otherwise that call might be blocking

litestar/utils/sync.py
```diff
@@ -78,8 +102,29 @@ async def _async_generator(self) -> AsyncGenerator[T, None]:
...
+        self,
+        typ: BaseException | type[BaseException],
+        val: BaseException | object = None,
+        tb: TracebackType | None = None,
+    ) -> T:
+        try:
+            return (
+                self._original_generator.throw(typ)
+                if isinstance(typ, BaseException)
+                else self._original_generator.throw(typ, val, tb)
```
provinzkraut: Needs to use `sync_to_thread`, otherwise that call might be blocking

litestar/utils/sync.py
```diff
@@ -51,18 +66,27 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
 
-class AsyncIteratorWrapper(Generic[T]):
+class AsyncIteratorWrapper(AsyncGenerator[T, S]):
     """Asynchronous generator, wrapping an iterable or iterator."""
 
-    __slots__ = ("generator", "iterator")
+    __slots__ = ("_original_generator", "generator", "iterator")
 
-    def __init__(self, iterator: Iterator[T] | Iterable[T]) -> None:
+    def __init__(self, iterable: Iterable[T]) -> None:
```
provinzkraut: Renaming the parameter is a breaking change

provinzkraut (changes requested):

litestar/response/streaming.py
winstxnhdw: What about `StopAsyncIteration`?

litestar/response/streaming.py
winstxnhdw: Sorry, I forgot to remove this. We are actually supposed to use `self._original_iterator` to get `content_async_iterator`. But perhaps it's better to defer everything to the `athrow` site.

litestar/response/streaming.py
provinzkraut: That should also be ignored

litestar/response/streaming.py
```diff
@@ -103,6 +120,16 @@ async def _stream(self, send: Send) -> None:
             None
         """
         async for chunk in self.iterator:
+            if self.disconnect_event.is_set():
+                try:
+                    if hasattr(self.iterator, "content_async_iterator"):
```
winstxnhdw: Having to use `hasattr` here couples this class with the SSE class. To fix this, we'll need to heavily refactor both classes. I'll leave this refactor to a follow-up PR since there's already so much going on in this one.

tests/unit/test_response/demo.py
```diff
@@ -20,12 +20,10 @@ async def get_notified(data: CleanupRequest) -> ServerSentEvent:
 
     async def generator() -> AsyncIterator[str]:
         try:
-            yield data.file_content
             while True:
+                yield data.file_content
                 await sleep(0.1)
```
winstxnhdw, line 25: Finally found the issue with the test. The SSE needs to continuously yield something otherwise the generator cannot handle the `athrow`. Is this something we should be worried about?

winstxnhdw:
Marking this as breaking again because async generators will no longer clean up by default as we no longer call `CancelScope`.

litestar/utils/sync.py
```diff
@@ -51,18 +66,27 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
         """
-        self.iterator = iterator if isinstance(iterator, Iterator) else iter(iterator)
+        self._original_generator: Generator[T, S, None]
+
+        if isinstance(iterator, Generator):
+            self._original_generator = iterator
+        elif isinstance(iterator, AsyncIteratorWrapper):
+            self._original_generator = iterator._original_generator
+        else:
+            self._original_generator = iterable_to_generator(iterator)
```
winstxnhdw: This line is still necessary for `asend` because `asend` needs a return value.

litestar/utils/sync.py
```diff
@@ -51,18 +66,27 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type:
...
-            iterator: A sync iterator or iterable.
+            iterator: A sync iterable.
         """
-        self.iterator = iterator if isinstance(iterator, Iterator) else iter(iterator)
+        self._original_generator: Generator[T, S, None]
+
+        if isinstance(iterator, Generator):
+            self._original_generator = iterator
+        elif isinstance(iterator, AsyncIteratorWrapper):
+            self._original_generator = iterator._original_generator
```
provinzkraut, line 85: This fundamentally changes the functionality of `AsyncIteratorWrapper`. It's supposed to wrap a synchronous iterable / iterator, and turn it into an async iterable. Now with the added functionality of also wrapping instances of itself (?), it gets really confusing.

litestar/response/streaming.py
```diff
@@ -103,6 +105,16 @@ async def _stream(self, send: Send) -> None:
             None
         """
         async for chunk in self.iterator:
+            if self.disconnect_event.is_set():
+                try:
+                    if isinstance(self.iterator.content_async_iterator, AsyncGenerator):  # type: ignore[attr-defined]
+                        await self.iterator.content_async_iterator.athrow(ClientDisconnectError)  # type: ignore[attr-defined]
```
provinzkraut, line 111: Again, I do not understand why you have to be so over specific here. It should be good enough to simply check:

- Is it an async-generator? `athrow`
- Is it an async-iterable? Just `break`

provinzkraut:
Hey @winstxnhdw, thanks for bringing up this issue and attempting to fix it! 

I know you've invested some time into this, so I'm sorry to say it, but I'll be closing this PR now though, since it doesn't seem to be getting into a state where I'd consider it mergeable, despite me sinking a lot of time into reviews, suggestions and conversations - time that could also have been spent fixing other (or this) bugs.

Should you want to contribute again in the future, I suggest you try to familiarise yourself with the material a bit more, and approach changes with a more holistic view, i.e. think deeper about the side effects your changes might have, and the intention behind them. To that end, I can also recommend simply reaching out on our discord server, where we'll be happy to answer all questions :)

litestar/response/streaming.py
winstxnhdw: The issue here is how `_ServerSentEventIterator` is designed. The content is kept in `self.content_async_iterator`, which means to close the original generator from the handler, we need to specifically throw `self.iterator.content_async_iterator` because `self.iterator` will only contain the header chunks like `event_id`, `event_type`, etc.

provinzkraut: `_ServerSentEventIterator` should handle that though, not the code that's calling it. It knows about its internal state, and is best suited to decide where to throw what exception. From the outside, it should be treated as any other async iterator / generator.

euri10:
you can also just put your PR as a draft and ask for anothers their pov if needed, I do that often when I try things and dont want to "pollute" other's workflow, which happens often for me, some PRs lead nowhere, as a draft it doesnt impact anyone else other than me.

PR #4405, opened 2025-10-06
fix: Use `inspect.iscoroutinefunction` instead of `asyncio.iscoroutinefunction` to address deprecation warning in Python 3.14+

edgarrmondragon:
## Description

Fixes a deprecation warning [introduced in Python 3.14][gh-122875].

I've tested this change with

```console
$ uv run pytest -q tests/unit/test_utils/test_predicates.py::test_is_async_callable
..........                                                                   [100%]
10 passed in 0.10s
```

## Closes

NA.

[gh-122875]: https://github.com/python/cpython/pull/122875

### DIFF
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -41,7 +41,7 @@ from litestar.utils.typing import get_origin_or_inner_type
 
 if sys.version_info >= (3, 10):
     from inspect import iscoroutinefunction
-else:
+else:  # pragma: no cover
     from asyncio import iscoroutinefunction
 
 if TYPE_CHECKING:
--- a/litestar/utils/predicates.py
+++ b/litestar/utils/predicates.py
@@ -42,6 +42,8 @@ from litestar.utils.typing import get_origin_or_inner_type
 if sys.version_info >= (3, 10):
     from inspect import iscoroutinefunction
 else:  # pragma: no cover
+    # In Python 3.9, AsyncMock is not detected
+    # as a coroutine function, so one test was failing.
     from asyncio import iscoroutinefunction
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
The error on Python 3.9 is legit, moving to draft while I figure it out...

litestar/utils/predicates.py
```diff
@@ -39,6 +39,11 @@
 from litestar.utils.helpers import unwrap_partial
 from litestar.utils.typing import get_origin_or_inner_type
 
+if sys.version_info >= (3, 10):
+    from inspect import iscoroutinefunction
+else:
+    from asyncio import iscoroutinefunction
```
edgarrmondragon, line 47: In Python 3.9, `AsyncMock` is not detected as a coroutine function, so one test was failing.

litestar/utils/predicates.py
```diff
@@ -39,6 +39,11 @@
 from litestar.utils.helpers import unwrap_partial
 from litestar.utils.typing import get_origin_or_inner_type
 
+if sys.version_info >= (3, 10):
+    from inspect import iscoroutinefunction
+else:
```
sobolevn: I think that the coverage is now lower, we need to add
```suggestion
else:  # pragma: no cover
```

here

litestar/utils/predicates.py
```diff
@@ -39,6 +39,11 @@
 from litestar.utils.helpers import unwrap_partial
 from litestar.utils.typing import get_origin_or_inner_type
 
+if sys.version_info >= (3, 10):
+    from inspect import iscoroutinefunction
+else:  # pragma: no cover
+    from asyncio import iscoroutinefunction
```
sobolevn, line 47:
```suggestion
    # In Python 3.9, AsyncMock is not detected
    # as a coroutine function, so one test was failing.
    from asyncio import iscoroutinefunction
```

Sorry, I missed this suggestion the first time :(

sobolevn (approved):

Issue #4401, opened 2025-10-05
Bug: The LITESTAR_QUIET_CONSOLE environment variable is treated as a truthy string


Peopl3s:
### Description

"0" or "false" behave like True.

https://github.com/search?q=repo%3Alitestar-org%2Flitestar%20LITESTAR_QUIET_CONSOLE&type=code

What to do: parse the boolean explicitly (in {"1","true","yes"}), and use it in the CLI in the same way.

### MCVE

```python

```

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [x] Linux
- [x] Mac
- [x] Windows
- [x] Other (Please specify in the description above)

Peopl3s:
Moved to https://github.com/litestar-org/litestar/issues/4422

PR #4407, opened 2025-10-06
fix(cli): correctly parse LITESTAR_QUIET_CONSOLE as boolean

7576457:
## Description

Previously, "0" and "false" were treated as truthy values.
Now we explicitly check for {"1", "true", "yes"}.

## Closes
* gh-4401


7576457:
I got a bit confused and shot myself in the foot,
hope you'll forgive me :)

Peopl3s:
@7576457 Maybe write a couple of tests?

litestar/cli/_utils.py
```diff
@@ -432,7 +432,7 @@ def show_app_info(app: Litestar) -> None:  # pragma: no cover
 def strtobool(value: str | None) -> bool:
     """Convert a string representation of truth to a boolean.
 
-    True values are: `1`, `true`, `yes`, `on` (case-insensitive).
```
7576457: I personally prefer having `on` there, so I initially wanted to include it. But I decided to stick with what was requested :)

7576457:
> @7576457 Maybe write a couple of tests?

Done!

sobolevn (commented):
Technically this is a breaking change. I am not sure about the deprecation process here, but I would want to raise this question :)

7576457:
> Technically this is a breaking change. I am not sure about the deprecation process here, but I would want to raise this question :)

I've been thinking about your words and couldn't reproduce a code-breaking situation.

I would say this is specifically a bug fix rather than a breaking change, because the current behavior is counter-intuitive:
LITESTAR_QUIET_CONSOLE=false enables quiet mode, which contradicts user expectations.

As stated in the original issue

Could you elaborate on this a bit more? :)

winstxnhdw:
It's a breaking change because someone could be depending on `LITESTAR_QUIET_CONSOLE="0"` to enable a quiet console.

provinzkraut:
> I would say this is specifically a bug fix rather than a breaking change, because the current behavior is counter-intuitive: LITESTAR_QUIET_CONSOLE=false enables quiet mode, which contradicts user expectations.

It may contradict *your* expectations, but for many, the expectation with env vars is that setting `"0"` means "disabled" and setting `"1"` means enabled. Everything else should be an exception though, and not treated as a truthy value by default. 

Now, I do agree with @sobolevn, that this is somewhat of a breaking change, and I'd like to make an alternative suggestion to the approach in this PR:

Since we do read booleans from env vars in a few places, I think we should ensure they behave the same in all places. Since we're *also* using `click` for our CLI to read some flags from the env, we should use its behaviour as the guide for how we interpret these as well.

This is how click does it: https://click.palletsprojects.com/en/stable/options/#values-from-environment-variables.

I'll close this in favour of #4422, which proposes a more hollistic approach. @7576457, if you want, you can continue there work on that ticket instead.

PR #4410, opened 2025-10-07
fix(litestar/config/app): fix Litestar init() and AppConfig defaults inconsistency

### LINKED ISSUE #4296: Bug: Litestar init() and AppConfig defaults aren't the same!
### Description

I was creating my config and app factories with plain dicts, then I started to look for some annotated option and found that AppConfig dataclass - cool!

But my single test started to crash with some error on request_max_body_size and without normal console logging.

So here we are:

```python
>>> from litestar.app import AppConfig
>>> from litestar import Litestar
>>> config = AppConfig()
... app1 = Litestar.from_config(config)
... app2 = Litestar()
>>> 
>>> app1.logging_config
>>> app2.logging_config
LoggingConfig(logging_module='logging', version=1,...)
>>>
>>> app1.request_max_body_size
<_EmptyEnum.EMPTY: 0>
>>> app2.request_max_body_size
10000000
```

So, with the help of AI, I identified the following discrepancies:

- logging_config: LoggingConfig() vs None
- response_cache_config: ResponseCacheConfig() vs None
- openapi_config: DEFAULT_OPENAPI_CONFIG vs None
- request_max_body_size: 10_000_000 vs Empty
- on_app_init: [] vs None

Additionally, plain init starts by performing some mutation of the given arguments and then passes to the config.
```python
if logging_config is Empty:
            logging_config = LoggingConfig()

        if debug is None:
            debug = os.getenv("LITESTAR_DEBUG", "0") == "1"

        if pdb_on_exception is None:
            pdb_on_exception = os.getenv("LITESTAR_PDB", "0") == "1"

config = AppConfig(
            after_exception=list(after_exception or []),
            after_request=after_request,
            ...)
```

_No response_

### MCVE

```python

```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.17.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Peopl3s:
## Description

Fixed inconsistency between :class:`Litestar` direct initialization and  :class:`~litestar.config.app.AppConfig` defaults. Previously, certain configuration parameters behaved differently when passed directly to ``Litestar()`` versus when using ``AppConfig`` with ``Litestar.from_config()``.

Key changes:
- ``logging_config`` with ``Empty`` now properly defaults to ``LoggingConfig`` in both initialization methods
- ``response_cache_config`` with ``None`` now properly defaults to ``ResponseCacheConfig`` in both methods
- ``request_max_body_size`` type annotation corrected to remove ``EmptyType`` since ``Empty`` is not allowed for this parameter
- All default values are now consistent between ``Litestar()`` and ``AppConfig`` initialization

This ensures that applications using ``AppConfig`` behave identically to those using direct ``Litestar`` initialization, maintaining backwards compatibility and predictable behavior.

Closes #4296

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -261,3 +261,23 @@
             difference is expected. Some test cases may break though if they relied on
             the fact that the middleware wrapper created by ``ASGIMiddleware`` was
             always being called
+
+    .. change:: Fix Litestar init() and AppConfig defaults inconsistency
+        :type: bugfix
+        :pr: TBD
+        :issue: 4296
+
+        Fixed inconsistency between :class:`Litestar` direct initialization and
+        :class:`~litestar.config.app.AppConfig` defaults. Previously, certain
+        configuration parameters behaved differently when passed directly to
+        ``Litestar()`` versus when using ``AppConfig`` with ``Litestar.from_config()``.
+
+        Key changes:
+        - ``logging_config`` with ``Empty`` now properly defaults to ``LoggingConfig`` in both initialization methods
+        - ``response_cache_config`` with ``None`` now properly defaults to ``ResponseCacheConfig`` in both methods
+        - ``request_max_body_size`` type annotation corrected to remove ``EmptyType`` since ``Empty`` is not allowed for this parameter
+        - All default values are now consistent between ``Litestar()`` and ``AppConfig`` initialization
+
+        This ensures that applications using ``AppConfig`` behave identically to those
+        using direct ``Litestar`` initialization, maintaining backwards compatibility
+        and predictable behavior.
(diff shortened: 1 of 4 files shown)


Peopl3s:
> Please, explain how this can affect end users? Will it be a breaking change for them? How can they get the old behaviour back if needed?

It looks like the existing code will continue to work as before, but now with consistent behavior.

Positive effects:

1. __Logging has been fixed__: Apps with AppConfig now have default logging
2. __API documentation has been fixed__: OpenAPI is now enabled by default for all
3. __Fixed execution errors__: Fixed crashes related to `request_max_body_size`
4. __Predictable behavior__: Both initialization methods work the same way

How to return the old behavior

```python
# Disable logging
config = AppConfig(logging_config=None)

# Disable OpenAPI
config = AppConfig(openapi_config=None)
```

Issue #4406, opened 2025-10-06
Bug: Typo in websocket_listener decorator argument passing


Peopl3s:
### Description

In the websocket_listener decorator function, located in litestar/handlers/websocket_handlers/listener.py, there appears to be a typo when passing the guards parameter to the WebsocketListenerRouteHandler constructor.

The websocket_listener decorator's signature correctly defines a parameter named guards:
```python
 guards: list[Guard] | None = None,
```

However, within the decorator function inside websocket_listener, when WebsocketListenerRouteHandler is instantiated, the guards parameter is passed as guard (singular):
```python
 guard=guards,
````

The WebsocketListenerRouteHandler constructor (which inherits from WebsocketRouteHandler, which in turn inherits from BaseRouteHandler) expects a parameter named guards (plural). This mismatch will likely lead to a runtime error, such as a TypeError if WebsocketListenerRouteHandler doesn't have a guard parameter, or the guards configuration will simply be ignored if guard is implicitly caught by **kwargs and then not processed correctly.

Impact:
* Any guards defined for a websocket_listener will not be correctly applied to the WebsocketListenerRouteHandler.
* The application might raise a TypeError during startup or when attempting to register the websocket listener, depending on how argument parsing is handled in the WebsocketListenerRouteHandler constructor and its parent classes.

## Solution:
Change guard=guards to guards=guards in the websocket_listener decorator function to correctly pass the guards parameter to the WebsocketListenerRouteHandler constructor.

### MCVE

```python

```

### Logs

```text

```

### Litestar Version

2.18.0

### Platform

- [x] Linux
- [x] Mac
- [x] Windows
- [x] Other (Please specify in the description above)

PR #4414, opened 2025-10-08
fix(websockets): Correctly pass `guards` to underlying handler

7576457:
## Description

Fixed a typo in the `websocket_listener` decorator where the `guards` parameter was incorrectly passed as `guard` (singular) to the `WebsocketListenerRouteHandler` constructor.

Additionally, fixed the test's context manager usage by changing it to `with client, client.websocket_connect("/") as ws:` instead of separate `with` blocks, which prevents tests from hanging due to improper resource cleanup order.

## Closes
* gh-4406


7576457:
Are tests needed for these changes? @Peopl3s

Peopl3s:
If these are not documentation fixes, then tests are always needed :)

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
```diff
@@ -441,3 +445,22 @@ async def lifespan() -> AsyncGenerator[None, None]:
...
+    async def custom_guard(connection: ASGIConnection, _: BaseRouteHandler) -> None:
+        nonlocal guard_called
+        guard_called = True
+
+    @websocket_listener("/", guards=[custom_guard])
+    async def handler(data: str) -> str:
+        return data
+
+    client = create_test_client([handler])
+    with client.websocket_connect("/") as ws:
```
provinzkraut:
```suggestion
    with create_test_client([handler]) as client, client.websocket_connect("/") as ws:
```
otherwise shutdown might not work properly, which can cause tests to hang

tests/unit/test_handlers/test_websocket_handlers/test_stream.py
```diff
@@ -150,3 +155,22 @@ def foo() -> bytes:
...
+        nonlocal guard_called
+        guard_called = True
+
+    @websocket_stream("/", guards=[custom_guard])
+    async def handler() -> AsyncGenerator[dict[str, str], None]:
+        yield {"Urfaust": "Gespinnst"}
+        yield {"Des": "Verderbens"}
+
+    client = create_test_client([handler])
+    with client.websocket_connect("/") as ws:
```
provinzkraut: Same issue as above

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
7576457: Thank you! I’ve fixed it.

This piece of code was copied from another similar test.
Since you mentioned it could cause tests to hang, I’ll probably fix the place I copied it from as well — if there are only one or two of them.
Otherwise, I think it would make sense to open an issue to address them all

7576457: They were in the same file, so I converted them to the same format. I think this PR is ready!

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
```diff
@@ -1,7 +1,9 @@
+from __future__ import annotations
```
provinzkraut:
```suggestion
```
This was intentionally not added to the tests to ensure backwards compatibility with older type annotations

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
provinzkraut: > This piece of code was copied from another similar test.

Sorry, I didn't notice that, otherwise I would have suggested fixing them in a separate PR. Fixing them now in this one is fine as well though, thanks :)

tests/unit/test_handlers/test_websocket_handlers/test_listeners.py
7576457: Hmm, considering that in a similar file [test_stream.py](https://github.com/litestar-org/litestar/blob/main/tests/unit/test_handlers/test_websocket_handlers/test_stream.py#L1C1-L1C35) this fragment was used, I assumed it was perfectly acceptable.

I originally did this to prevent mypy from complaining about this line:

```python
    async def custom_guard(connection: ASGIConnection, _: BaseRouteHandler) -> None:
```

In that case, I’ll remove this line, revert it back to how it was, and try to handle it in a different way

provinzkraut: > Hmm, considering that in a similar file 

Not all tests are created equal :)
That test in particular tests behaviour that isn't dependant on signature parsing the way this module is, so it's fine to use the new-style annotations there

7576457: Got it. It seems everything is working properly now. If any changes are needed, please let me know

Umpire2018:
@provinzkraut Maybe we could close this one?

Peopl3s:
@7576457 Great job!

PR #4416, opened 2025-10-10
feat(openapi): add custom Scalar render options example

### LINKED ISSUE #4381: Create example for customizing Scalar render options
Document this (more generally how options can be passed in according to https://guides.scalar.com/scalar/scalar-api-references/configuration and the `options` param):
```py
scalar_plugin = ScalarRenderPlugin(
    path="/scalar",
    options={
        "hideTestRequestButton": True,
    }
)
```

_Originally posted by @JacobCoffee in https://github.com/litestar-org/litestar/discussions/4252#discussioncomment-14575383_

Umpire2018:
## Description

This PR added a example that demonstrates how to customize Scalar render options using the `options` parameter in `ScalarRenderPlugin`.

Besides, [here](https://github.com/Umpire2018/litestar/pull/1) is an alternative provided by copilot. I found that there is enhancement after reviewing this PR and then came up with another solution.

Closes #4381 , thanks to @JacobCoffee BTW.


docs/usage/openapi/ui_plugins.rst
```diff
@@ -242,17 +242,6 @@ With that, you can preset your clientId or enable PKCE support.
 .. literalinclude:: /examples/openapi/plugins/swagger_ui_oauth.py
     :language: python
 
-Customizing the OpenAPI UI
```
JacobCoffee, line 251: should we have removed this section? it needs to be updated to not be so specific to scalar but i think `OpenAPI UI can be customized by overriding the default ``css_url`` and ``js_url`` attributes` is generic across all our renderers? can probably leave it alone and have another pr to update

JacobCoffee: we sortve duplicate it between https://docs.litestar.dev/main/usage/openapi/ui_plugins.html#cdn-and-offline-file-support / https://docs.litestar.dev/main/usage/openapi/ui_plugins.html#customizing-the-openapi-ui 😵‍💫

Umpire2018: > should we have removed this section?

1. I prefer to remove this section because i found that we already explained ```css_url``` and ```js_url``` in https://docs.litestar.dev/main/usage/openapi/ui_plugins.html#configuring-openapi-ui-plugins.

> it needs to be updated to not be so specific to scalar
2. I agree with you. As far as at the https://docs.litestar.dev/main/usage/openapi/ui_plugins.html#customizing-the-openapi-ui part , it would be better to provide different showcases.

> `css_url` and `js_url` attributes is generic across all our renderers?
3. Negative.  RapidDoc and Redoc don't supprt `css_url` base on [API Reference](https://docs.litestar.dev/main/reference/openapi/plugins.html#litestar.openapi.plugins.RapidocRenderPlugin) so far. But maybe it's time to leave some #TODO in code to make it generic.

> #cdn-and-offline-file-support part is duplicate.
4. Maybe we can combine it into #configuring-openapi-ui-plugins part?

Issue #4417, opened 2025-10-11
FEAT: Adding --json output option in /cli/core .py on routes command


MauricioReisdoefer:
### Summary

# Feature Request: Add `--json` Option to `routes` Command

## Description

Currently, the `routes` command outputs the list of application routes in a human-readable table format.  
It would be useful to add a `--json` flag to allow users to export the routes in **JSON format**, making it easier to programmatically consume the route information.

## Proposed Solution

- Add a `--json` option to the `routes` CLI command.
- When `--json` is provided, output the list of routes as a JSON array instead of the default table.
- Ensure that the JSON output includes all relevant route details (e.g., path, methods, name, handler, etc.).
- Update or add tests to validate JSON output.

## Benefits

- Programmatic access to route data for tooling or documentation generation.
- Improves automation and integration with other tools.

## Additional Context

This feature is especially useful for larger applications where manual inspection of routes is cumbersome, or when generating automated documentation.

## Example Usage

```bash
$ litestar routes --json
[
  {
    "name": h.name,
    "path": h.paths,
    "handler": h.handler_name,
    "methods": sorted(h.http_methods),
    "async": inspect.iscoroutinefunction(unwrap_partial(h.fn))  
  },
]
```

## Drawbacks

Some drawbacks only occurs in particular cases.

- Potential for sensitive information exposure
- Performance impact
- Maintenance overhead

PR #4418, opened 2025-10-11
feat: Adding new --format option in routes CLI command for JSON format (#4417)

MauricioReisdoefer:
## Description

This PR introduces a `--format` option for the `litestar routes` CLI command, allowing routes to be output in a structured JSON format. This is useful for automation, scripts, or integration with external tools.

## Usage Example

```bash
Example usage:

```bash
$ litestar routes --format=json
[
  {
    "name": h.name,
    "path": h.paths,
    "handler": h.handler_name,
    "methods": sorted(h.http_methods),
    "async": inspect.iscoroutinefunction(unwrap_partial(h.fn))  
  },
]
```
## Motivation and Context

Closing issue #4417. This change allows developers to easily consume route information programmatically, improving automation and tooling support.

Also i added new tests to cover this new feature.


provinzkraut (changes requested):
It would be better to add a `--format` option, with `json` as a choice, as that is more extendable.

MauricioReisdoefer:
I agree. But first i'll fix the unpassed tests.

litestar/repository/abc/_sync.py
provinzkraut, line 1: This seems to be an accidental change, and is what's making the CI fail

MauricioReisdoefer: That makes sense

MauricioReisdoefer: thanks for help. Sorry for creating another issue for it.

provinzkraut:
@MauricioReisdoefer if you want to get this merged, please address the changes I've requested regarding the flag.

MauricioReisdoefer:
Do you want any more changes in this pr? Like, creating a new format already? Because i want to start contributing in other parts after this pr.

litestar/cli/commands/core.py
```diff
@@ -319,7 +320,38 @@ def routes_command(app: Litestar, exclude: tuple[str, ...], schema: bool) -> Non
     if exclude is not None:
         sorted_routes = remove_routes_with_patterns(sorted_routes, exclude)
 
-    console.print(_RouteTree(sorted_routes))
+    if output_format == "json":
```
provinzkraut, line 323: Can you refactor this block into a separate function?

litestar/cli/commands/core.py
```diff
@@ -319,7 +320,38 @@ def routes_command(app: Litestar, exclude: tuple[str, ...], schema: bool) -> Non
...
+                route_info["handlers"] = [
+                    {
+                        "name": h.name,
+                        "path": h.paths,
+                        "handler": h.handler_name,
+                        "methods": sorted(h.http_methods),
+                        "async": inspect.iscoroutinefunction(unwrap_partial(h.fn)),
+                    }
+                    for h in route.route_handlers
+                ]
```
provinzkraut, line 345: This duplicates the logic already present in the `_RouteTree` class. I suggest extracting that logic into a common utility, which then both the tree output as well as the JSON output can use.

MauricioReisdoefer: Ok, i'll see it now

PR #4420, opened 2025-10-11
fix: `KeyError` swallowing in `default_serializer`

sobolevn:
It used to do several things that are not quite right:
1. `KeyError` raised in `encoder(value)` call was swallowed, it could potentially lead to a very hard to debug problems, when a `KeyError` bug in encoder would just select another one
2. `.get` + `if is None` is a bit tiny faster than `[]` with `except KeyError` in some cases

### DIFF
--- a/litestar/serialization/msgspec_hooks.py
+++ b/litestar/serialization/msgspec_hooks.py
@@ -85,10 +85,12 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
     type_encoders = {**DEFAULT_TYPE_ENCODERS, **(type_encoders or {})}
 
     for base in value.__class__.__mro__[:-1]:
-        encoder = type_encoders.get(base)
-        if encoder is None:
+        try:
+            encoder = type_encoders[base]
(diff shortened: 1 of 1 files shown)


provinzkraut:
> 2. `.get` + `if is None` is a bit tiny faster than `[]` with `except KeyError` in some cases

Really? Do you know why? I was under the impression that since 3.11 (I think it was this version?), `KeyError` should be faster in the happy path, i.e. if not exception is raised, because of the reduced overhead from exception handlers

sobolevn:
Yeap, you are right :) 

`KeyError` is faster. Sorry!

```
» pyperf timeit -s 'encoders = {str: str}' '          
. try:
.    encoders[str]
. except KeyError:
.    pass
. '
.....................
Mean +- std dev: 11.9 ns +- 0.1 ns
```

vs

```
» pyperf timeit -s 'encoders = {str: str}' '
enc = encoders.get(str)
if enc is not None: ...
'
.....................
Mean +- std dev: 17.4 ns +- 0.2 ns
```

I will update my PR.

Issue #4423, opened 2025-10-11
Enhancement: support `ReadOnly` type for `TypedDict`


sobolevn:
### Summary

Right now this happens for this code sample:

```python
from litestar import Litestar, get
from typing import TypedDict
from typing_extensions import ReadOnly

class Model(TypedDict):
    id: ReadOnly[str]

@get('/')
async def test_route() -> Model:
    return {'id': 'test'}

app = Litestar(
    route_handlers=[test_route],
)
```

`/schema` endpoint raises:

```
500: Parameter 'id' with type 'typing.ReadOnly[str]' could not be mapped to an Open API type. This can occur if a user-defined generic type is resolved as a parameter. If 'id' should not be documented as a parameter, annotate it using the `Dependency` function, e.g., `id: ... = Dependency(...)`.
```

Note that OpenAPI has `readOnly: true` property, see https://swagger.io/docs/specification/v3_0/data-models/data-types/#read-only-and-write-only-properties But! **It is not the same** `ReadOnly` as in Python. Our `ReadOnly` means that the key in `TypeDict` can't be changed. I think the proper solution here is to just unwrap `ReadOnly[T]` to `T`

I will send a PR.

PR #4424, opened 2025-10-11
feat(typing): support `typing.ReadOnly` for `TypedDict` schemas

sobolevn:
Closes https://github.com/litestar-org/litestar/issues/4423

### DIFF
--- a/litestar/utils/typing.py
+++ b/litestar/utils/typing.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 from collections import abc, defaultdict, deque
 from typing import (
     AbstractSet,
@@ -63,13 +62,6 @@ __all__ = (
 
 T = TypeVar("T")
 UnionT = TypeVar("UnionT", bound="Union")  # pyright: ignore
-
-tuple_types_regex = re.compile(
-    "^"
-    + "|".join(
-        [*[repr(x) for x in (List, Sequence, Iterable, Iterator, Tuple, Deque)], "tuple", "list", "collections.deque"]
-    )
-)
 
 instantiable_type_mapping = {
     AbstractSet: set,


tests/models.py
```diff
@@ -35,7 +35,7 @@ class DataclassPerson:
 class TypedDictPerson(TypedDict):
     first_name: Required[str]
     last_name: Required[str]
-    id: Required[str]
+    id: Required[ReadOnly[str]]
```
provinzkraut, line 38: Can you add a separate test case for this?

sobolevn: Should I 
1. keep this change and add a new test 
or should I
2. I revert this change and add a new test
?

provinzkraut: The change is fine, so just add a new test

provinzkraut:
We really need to do something about these CI timeouts.. https://github.com/litestar-org/litestar/actions/runs/18440384263/job/52539920154?pr=4424#step:10:118

provinzkraut:
@sobolevn I've cherry-picked this into the v2 branch, where it seems to fail under Python 3.8. Do you have any idea why?


https://github.com/litestar-org/litestar/actions/runs/20194318117/job/57976059886?pr=4524#step:11:505

sobolevn:
It might be a `typing_extensions` bug. We don't support 3.8 anymore: https://github.com/python/typing_extensions/blob/main/.github/workflows/ci.yml#L41

provinzkraut:
I'll check that out

provinzkraut:
Unfortunately that did not fix it

provinzkraut:
Got it. We're checking for [`__required_keys__`](https://docs.python.org/3/library/typing.html#typing.TypedDict.__required_keys__), which was only added in 3.9

Issue #4411, opened 2025-10-07
Bug: Behavior when using pydantic as a query dependency


madany01:
### Description

In the given example, there were 2 issues.
1- The fields types and metadata aren't reflected on swagger
2- When supplying wrong types (e.g. string abc instead of an int), a 500 Internal Server Error is raised instead of normal validation error.

### MCVE

```python
from typing import Any

from litestar import Litestar, get
from litestar.di import Provide
from pydantic import BaseModel, Field


class QueryValidator(BaseModel):
    skip: int = Field(
        default=0,
        title='this is skip title',
        description='this is skip description',
        examples=[1, 2, 3, 4],
    )
    limit: int = Field(
        default=10,
        title='this is limit title',
        description='this is limit description',
        examples=[11, 22, 33, 44],
    )


@get('/', dependencies={'query_params': Provide(QueryValidator)})
def route_handler(query_params: QueryValidator) -> Any:
    return query_params


app = Litestar([route_handler])
```

### Steps to reproduce

with the file named as `app.py`, run `litestar run -d --reload`

### Screenshots

<img width="1309" height="523" alt="Image" src="https://github.com/user-attachments/assets/382623fc-4a18-4ffd-82d3-09635f1c14a5" />

### Logs

```text
ERROR - 2025-10-07 19:57:51,403 - litestar - config - Uncaught exception (connection_type=http, path='/'):
Traceback (most recent call last):
  File "D:\example\litestar_projectdir\.venv\Lib\site-packages\litestar\middleware\_internal\exceptions\middleware.py", line 158, in __call__
    await self.app(scope, receive, capture_response_started)
  File "D:\example\litestar_projectdir\.venv\Lib\site-packages\litestar\_asgi\asgi_router.py", line 100, in __call__
    await asgi_app(scope, receive, send)
  File "D:\example\litestar_projectdir\.venv\Lib\site-packages\litestar\routes\http.py", line 81, in handle     
    response = await self._get_response_for_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        scope=scope, request=request, route_handler=route_handler, parameter_model=parameter_model
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #4426, opened 2025-10-11
Fix Pydantic field types in DI and proper validation errors

7576457:
## Description

-


provinzkraut:
I'm assuming this is meant to fix #4411?

If so, please read this comment: https://github.com/litestar-org/litestar/issues/4411#issuecomment-3393088271

The typing as `Any` is intentional, to ensure Litestar's validation is bypassed, and only Pydantic's validation is applied. The changes in this PR alter this behaviour, so the effect you're seeing is just a side-effect from the validation not being performed by Pydantic anymore.

That is to say, this behaviour *is not a bug*. It's intentional and expected behaviour.

PR #4427, opened 2025-10-11
refactor: type annotate `content` properly

winstxnhdw:
## Description

This PR is a series of small changes to reduce noise for an eventual generator propagation fix.

## Changes

- If we are expecting `content` to be a `Callable`, we should type it as such.
- Remove redundant `Any` as `ServerSentEventMessage` is typed `str | int | bytes | None`.
- Reuse `DEFAULT_HEADER` in SSE header builder

### DIFF
--- a/litestar/response/sse.py
+++ b/litestar/response/sse.py
@@ -37,17 +37,18 @@ class _ServerSentEventIterator(AsyncIteratorWrapper[bytes]):
         self.event_type = event_type
         self.retry_duration = retry_duration
         chunks: list[bytes] = []
+
         if comment_message is not None:
-            chunks.extend([f": {chunk}\r\n".encode() for chunk in _LINE_BREAK_RE.split(comment_message)])
+            chunks.extend(f": {chunk}{DEFAULT_SEPARATOR}".encode() for chunk in _LINE_BREAK_RE.split(comment_message))
 
         if event_id is not None:
-            chunks.append(f"id: {event_id}\r\n".encode())
+            chunks.append(f"id: {event_id}{DEFAULT_SEPARATOR}".encode())
 
         if event_type is not None:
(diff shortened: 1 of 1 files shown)


litestar/response/sse.py
```diff
@@ -37,33 +37,34 @@ def __init__(
         self.event_type = event_type
         self.retry_duration = retry_duration
         chunks: list[bytes] = []
+
         if comment_message is not None:
-            chunks.extend([f": {chunk}\r\n".encode() for chunk in _LINE_BREAK_RE.split(comment_message)])
```
euri10, line 41: im not sure we test this, i'm just surprised by the list comprehension removal here.

the DEFAULT_SEPARATOR use lgtm,

I dont feel comfortable commenting on types.

winstxnhdw: > im not sure we test this, i'm just surprised by the list comprehension removal here.

The only difference here is that we append to chunks as we encode  `comment_message`. There's somehow a slight performance improvement here.

provinzkraut: > There's somehow a slight performance improvement here.

It's because `list.extend` takes any iterable. When doing `list.extend([i for i in range(10)])`, you:

1. Build a list of the values
2. Pass the list to `list.extend`
3. `list.extend` iterates over all the values and adds them to the list

So essentially, you're creating a list that you throw away immediately, and you iterate over all items twice.

but if you instead do `list.extend(i for i in range(10))`, you create a generator expression, that produces the values lazily, so you only iterate over each value once.

winstxnhdw: Yeah, I thought it’d be something like this, I just didn’t expect it to be sufficient to overcome the generator overhead. I guess I’ve overestimated its cost all along.

PR #4429, opened 2025-10-12
refactor: simplify `encode_json` function

sobolevn:
After double checking type info: https://github.com/jcrist/msgspec/blob/bc60e96772c5e8a3babff967d86a9e7dfcdbfb1b/msgspec/json.pyi#L107

And implementation: https://github.com/jcrist/msgspec/blob/bc60e96772c5e8a3babff967d86a9e7dfcdbfb1b/msgspec/_core.c#L14223-L14224

And doing testing:

```python
>>> import msgspec
>>> msgspec.json.encode({}, enc_hook=None)
b'{}'
>>> msgspec.json.encode({})
b'{}'
```

I can be certain to say that this `if` is not required. It is better to remove it from the hot path.

### DIFF
--- a/litestar/serialization/msgspec_hooks.py
+++ b/litestar/serialization/msgspec_hooks.py
@@ -162,7 +162,7 @@ def encode_json(value: Any, serializer: Callable[[Any], Any] | None = None) -> b
         SerializationException: If error encoding ``obj``.
     """
     try:
-        return msgspec.json.encode(value, enc_hook=serializer) if serializer else _msgspec_json_encoder.encode(value)
+        return msgspec.json.encode(value, enc_hook=serializer)
     except (TypeError, msgspec.EncodeError) as msgspec_error:
         raise SerializationException(str(msgspec_error)) from msgspec_error
 


litestar/serialization/msgspec_hooks.py
```diff
@@ -162,7 +162,7 @@ def encode_json(value: Any, serializer: Callable[[Any], Any] | None = None) -> b
         SerializationException: If error encoding ``obj``.
     """
     try:
-        return msgspec.json.encode(value, enc_hook=serializer) if serializer else _msgspec_json_encoder.encode(value)
+        return msgspec.json.encode(value, enc_hook=serializer)
```
provinzkraut, line 165: I suspect there's some missing tests here, as there is quite a big difference in behaviour here.

If `serializer=None`, we use `_msgspec_json_encoder`, which is a `msgspec.json.Encoder`, with `enc_hook=default_serializer`. So it would be equivalent of calling `msgspec.json.encode(enc_hook=default_serializer)`. With the change, if `serializer=None`, our default serializer isn't being used.

You could of course change this to `msgspec.json.encode(enc_hook=serializer or default_serializer)`, but idk if this more performant than the `if` variant. Also, pre-constructing a `msgspec.json.Encoder`, with a `enc_hook` set *should* perform better, at least according to the docs.

sobolevn:
Yes, I missed this nuance! Thanks a lot!

PR #4430, opened 2025-10-12
refactor: re-enable `default_deserializer` coverage

sobolevn:
This function is rather important, we need to add more tests if it is not covered, not ignore it.

Let's see the results.

### DIFF
--- a/litestar/serialization/msgspec_hooks.py
+++ b/litestar/serialization/msgspec_hooks.py
@@ -95,9 +95,7 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
     raise TypeError(f"Unsupported type: {type(value)!r}")
 
 
-def default_deserializer(
-    target_type: Any, value: Any, type_decoders: TypeDecodersSequence | None = None
-) -> Any:  # pragma: no cover
+def default_deserializer(target_type: Any, value: Any, type_decoders: TypeDecodersSequence | None = None) -> Any:
     """Transform values non-natively supported by ``msgspec``
 
     Args:


sobolevn:
Here's the coverage result: https://app.codecov.io/gh/litestar-org/litestar/commit/39cec47ffa145a7ef8a12d78bd06fd2d2253489e/indirect-changes?dropdown=coverage&el=desc
Снимок экрана 2025-10-12 в 10 01 31

I will add tests for the new red paths.

sobolevn:
@provinzkraut I have a hard time figuring out how to actually test this 🤔 
Any ideas? :)

provinzkraut:
Hmm. For L119-L124 there exists a case that *should* cover this: https://github.com/litestar-org/litestar/blob/9a3bd38222b63ff4c5eaf4d23542c06cd2fd2a25/tests/unit/test_signature/test_validation.py#L294

At least that's the test added in #3519, which introduced these lines in the first place.

Apparently it's not testing what it's supposed to be testing. Maybe this is a test regression, and that case is getting caught somewhere upstream now? 

Would be a good starting point for investigating I think.

euri10:
```
def test_x() -> None:
    d = default_deserializer(list[int], [1,2,3,4])
    assert d == [1,2,3,4]
```
this takes the red path I think, or you're talking about the line 125-126 ?

sobolevn:
@provinzkraut this test does not trigger the red path :( 
I started with it.

@euri10 I guess that this test it so not quite correct, because we don't use `default_deserializer` directly, it is an internal callback. I have to somehow trigger it from a layer above.

PR #4431, opened 2025-10-12
docs: remove mention of nginx-unit which is officially unmaintained

euri10:
see https://github.com/nginx/unit/commit/76b39e928bd29f09278b60c3878b9f3d032b45b6

## Description

-


provinzkraut:
Huh, that was a short-lived thing

PR #4432, opened 2025-10-12
refactor: keep iterator on the main thread

winstxnhdw:
## Description

To prevent cross-thread reference errors with `Stream` iterators, we keep iterators on the main context and listen for a disconnect asynchronously instead of vice versa.

I have properly split the changes into 3 separate commits for better diffs.

## Other changes

- Reuse `stream_event` dictionary to avoid recreating the dictionary on every iteration.
- Remove redundant type checks, e.g. `AsyncIterable` covers `AsyncIterator` too.
- Some miscellaneous inlining and using terse expressions like `chain(self.cookies, cookies or ())`

### DIFF
--- a/litestar/response/streaming.py
+++ b/litestar/response/streaming.py
@@ -73,7 +73,7 @@ class ASGIStreamingResponse(ASGIResponse):
 
         self.disconnect_event = Event()
         self.iterator: AsyncIterable[str | bytes] | AsyncGenerator[str | bytes, None] = (
-            iterator if isinstance(iterator, (AsyncIterable, AsyncIterator)) else AsyncIteratorWrapper(iterator)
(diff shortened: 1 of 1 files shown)


provinzkraut:
Can you explain this change a bit more? Where is something handed off to a different thread?

winstxnhdw:
> Can you explain this change a bit more? Where is something handed off to a different thread?

Originally, we start `self._stream` in another task like so.

```python
    task_group.start_soon(partial(self._stream, send))
    await self._listen_for_disconnect(cancel_scope=task_group.cancel_scope, receive=receive)
```

If a user, who holds the reference to the iterator, tries to do something with it, Python raises a `ValueError: generator already executing` error. This only happens with a sync generator, presumably because we use `sync_to_thread` which actually hands it off to a different thread, and the combination of the stream being in a different context and iterating in a separate thread somehow leads to this issue.

PR #4435, opened 2025-10-13
docs: add note on how to add a middleware with a constraint through a…

euri10:
… plugin

## Description

- triggered by https://discord.com/channels/919193495116337154/1427238052882812929

### DIFF
--- a/docs/usage/middleware/creating-middleware.rst
+++ b/docs/usage/middleware/creating-middleware.rst
@@ -137,7 +137,7 @@ When using plugins that add middleware(s), it is important to understand that th
 middleware(s) are added *after* middlewares defined on the application and *before* middlewares defined on the other layers.
 
 Constraints are evaluated after all middlewares have been added though, so an order constraint on a middleware added by
-a plugin, has to take into account the position it is being added to.
+a plugin has to take into account the position it is being added to.
 
 Most of the time in a plugin you would do ``app_config.middleware.append(MyCustomMiddleware)`` which will be
 ok if it has no constraints attached.


docs/usage/middleware/creating-middleware.rst
```diff
@@ -130,6 +130,20 @@ be the *last* middleware on the *last* layer (i.e. the route handler).
...
     )
 
+Constraints and plugins
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using plugins that add middleware(s), it is important to understand that these
+middleware(s) are added *before* any user-defined middleware.
+
+This means that if you add in a plugin a middleware that has a constraint, the constraint has to be satisfied
+or an error will be raised.
```
provinzkraut: Not sure I understand. That's the case for all constraints, not just constraints on middlewares added by plugins?

docs/usage/middleware/creating-middleware.rst
```diff
@@ -130,6 +130,20 @@ be the *last* middleware on the *last* layer (i.e. the route handler).
         ]
     )
 
+Constraints and plugins
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using plugins that add middleware(s), it is important to understand that these
+middleware(s) are added *before* any user-defined middleware.
```
provinzkraut:
```suggestion
middleware(s) are added *after* middlewares defined on the application and *before* middlewares defined on the other layers.
```

I think this makes it more clear

docs/usage/middleware/creating-middleware.rst
euri10: yes you're right, i realize it is ambiguous. maybe i can rephrase saying something like:

This means that even for middlewares added through a plugin, the constraint has to be satisfied
or an error will be raised.

provinzkraut: I feel like that's still saying the same thing :sweat_smile: 

Maybe

> Constraints are evaluated after all middlewares have been added though, so an order constraint on a middleware added by a plugin, has to take into account the position it is being added to

?

euri10: no worries I added your suggestion, docs are hard to get "right"

docs/usage/middleware/creating-middleware.rst
```diff
@@ -130,6 +130,20 @@ be the *last* middleware on the *last* layer (i.e. the route handler).
...
     )
 
+Constraints and plugins
+~~~~~~~~~~~~~~~~~~~~~~~
+
+When using plugins that add middleware(s), it is important to understand that these
+middleware(s) are added *after* middlewares defined on the application and *before* middlewares defined on the other layers.
+
+Constraints are evaluated after all middlewares have been added though, so an order constraint on a middleware added by
+a plugin, has to take into account the position it is being added to.
```
provinzkraut:
```suggestion
a plugin has to take into account the position it is being added to.
```

PR #4436, opened 2025-10-14
fix(docs): ensure class member order follows source

### LINKED ISSUE #4393: Docs: Member order of classes is sometimes alphabetical instead of by source
### Summary

Looking at e.g. https://docs.litestar.dev/main/reference/middleware/index.html#litestar.middleware.ASGIMiddleware, we can see that the order the class members appear in is alphabetical, instead of by source. This is quite jarring, because it mixes e.g. methods with class variables, making it much harder to read.

Now, we *do* set `autodoc_member_order = "bysource"`, which *should* make it so the order is by source and *not* alphabetical, however, this seems to be disregarded in some cases.

There is a known quirk where placing something into a module's `__all__`  will mean the order of members therein is takes precedence over whatever `autodoc_member_order`, however, that should not apply to the class members itself.

It *seems* to have something to do with `__all__` though, as adding a class to e.g. `litestar.middleware.__init__` and adding that class to the module's `__all__` will result in its member appearing in alphabetical order, and removing it from `__all__` results in it being ordered as expected. However, I could not reproduce this behaviour in a new sphinx project (using the same sphinx version that is).

Umpire2018:
## Description

At first glance, other modules works fine. e.g. https://docs.litestar.dev/main/reference/middleware/logging.html#litestar.middleware.logging.LoggingMiddlewareConfig

Image

After digging in, i found that if a class is imported into `__init__.py` via `from ... import ...` and added to `__all__`, Sphinx may lose the original source order of its members so i changed to document classes directly from their original module, rather than through indirect imports in `__init__.py`.

| Before                                                                                                   | After                                                                                                    |
|----------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|
| image | image |

## Closes
close #4393

### DIFF
--- a/docs/reference/middleware/index.rst
+++ b/docs/reference/middleware/index.rst
@@ -1,8 +1,8 @@
 middleware
 ==========
 
-.. automodule:: litestar.middleware
-
+.. automodule:: litestar.middleware.base
+    :members:
 
 .. toctree::
     :maxdepth: 1
--- a/docs/release-notes/whats-new-2.rst
+++ b/docs/release-notes/whats-new-2.rst
@@ -187,11 +187,11 @@ Imports
 +----------------------------------------------------+------------------------------------------------------------------------+
 | ``starlite.AuthenticationResult``                  | :class:`.middleware.authentication.AuthenticationResult`               |
 +----------------------------------------------------+------------------------------------------------------------------------+
-| ``starlite.AbstractMiddleware``                    | :class:`.middleware.AbstractMiddleware`                                |
+| ``starlite.AbstractMiddleware``                    | :class:`.middleware.base.AbstractMiddleware`                           |
 +----------------------------------------------------+------------------------------------------------------------------------+
-| ``starlite.DefineMiddleware``                      | :class:`.middleware.DefineMiddleware`                                  |
+| ``starlite.DefineMiddleware``                      | :class:`.middleware.base.DefineMiddleware`                             |
 +----------------------------------------------------+------------------------------------------------------------------------+
-| ``starlite.MiddlewareProtocol``                    | :class:`.middleware.MiddlewareProtocol`                                |
+| ``starlite.MiddlewareProtocol``                    | :class:`.middleware.base.MiddlewareProtocol`                           |
 +----------------------------------------------------+------------------------------------------------------------------------+
 | **Security**                                                                                                                |
 +----------------------------------------------------+------------------------------------------------------------------------+
--- a/docs/release-notes/whats-new-3.rst
+++ b/docs/release-notes/whats-new-3.rst
@@ -337,7 +337,7 @@ parameter.
 Middleware configuration constraints
 -------------------------------------
 
-:class:`~litestar.middleware.ASGIMiddleware`\ s can now express constraints for how
+:class:`~litestar.middleware.base.ASGIMiddleware`\ s can now express constraints for how
 they are applied in the middleware stack, which are then validated on application
 startup.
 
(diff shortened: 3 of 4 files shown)


provinzkraut:
Interesting find! Does this replicate in a fresh project? If so, I think you should submit this as a bug to https://github.com/sphinx-doc/sphinx/.

As for the proposed fix on our side, I'm not entirely sure this is fine to do, since it would mean that only members of `litestar.middleware.base` get documented, and things that don't originate there but are imported into `litestar.middleware` are not. We could of course *explicitly* document `litestar.middleware.base`, but then I think it would be confusing if the import is `from litestar.middleware import ASGIMiddleware`, but the docs are for `litestar.middleware.base.ASGIMiddleware`. 

@euri10, got some thoughts on this? I know how much you like to consult documentation :)

euri10:
vey interesting finding yeah.

i agree with you it would be confusing, and if i had to choose i'd sacrifice the order for the "correct" import.

now the order is already weird to me in other cases, not just class member, if you look at https://docs.litestar.dev/main/reference/middleware/constraints.html for instance i find it weird to get exceptions before classes for instance,

provinzkraut (approved):

provinzkraut:
Alright, then let's merge this and see if can can sort out the imports again once the bug has been fixed upstream.

euri10:
> Alright, then let's merge this and see if can can sort out the imports again once the bug has been fixed upstream.

any issue to follow ?

Umpire2018:
> I think it would be confusing if the import is `from litestar.middleware import ASGIMiddleware`, but the docs are for `litestar.middleware.base.ASGIMiddleware`.

1. I agree with you on that. After investigating more on that, i think it may be better to leave it to what it is nowadays because i don't think upstream will fix this `autodoc` related issue considering there are lots of ongoing tickets right now.

2. It seems like class members are in alphabetical sequence elsewhere instead of `by source`. See [here](https://litestar-org.github.io/litestar-docs-preview/4436/reference/datastructures/index.html#litestar.datastructures.CacheControlHeader).

image

Maybe we can consider alternatives like https://docusaurus.io/ ?

euri10:
idk https://docusaurus.io/  to be fair so I cant comment on this
I had good success with the `autoapi` extension but I have to admitt I didnt dwelve into all the details, but overall I found it to be easier to use and have a nicer look than autodoc, but that is just a very uninformed impression

Umpire2018:
> now the order is already weird to me in other cases, not just class member, if you look at https://docs.litestar.dev/main/reference/middleware/constraints.html for instance i find it weird to get exceptions before classes for instance,

@euri10 Maybe this is expected behavior ? Because this is the sequence in source code and this PR haven't introduce relevant changes.

https://github.com/litestar-org/litestar/blob/94073d3447d9d9da81267819ccb2d5718248753c/litestar/middleware/constraints.py#L27-L40

euri10:
> > now the order is already weird to me in other cases, not just class member, if you look at https://docs.litestar.dev/main/reference/middleware/constraints.html for instance i find it weird to get exceptions before classes for instance,
> 
> @euri10 Maybe this is expected behavior ? Because this is the sequence in source code and this PR haven't introduce relevant changes.
> 
> https://github.com/litestar-org/litestar/blob/94073d3447d9d9da81267819ccb2d5718248753c/litestar/middleware/constraints.py#L27-L40

yep sorry for the confusion i was speaking in general, its expected behaviour as you rightully say, with autoapi this would be ordered differently I think without thinking about how it's organized in source,

provinzkraut:
> Maybe we can consider alternatives like https://docusaurus.io/ ?

No. We are currently implementing a lot of custom functionality in Sphinx, and I don't see enough of an upside to do the migration.

@Umpire2018 

> Interesting find! Does this replicate in a fresh project? If so, I think you should submit this as a bug to https://github.com/sphinx-doc/sphinx/.

?

provinzkraut (approved):

provinzkraut:
@Umpire2018 This is currently failing on docs build

PR #4438, opened 2025-10-14
fix: Converted all relative imports to in-package absolute

Peopl3s:
## Description

Converted all relative imports to in-package absolute

FYI: used @lyova24 pre-commit-hook [shrimport](https://github.com/lyova24/shrimport)


provinzkraut:
IMO, and according to PEP8, relative imports from siblings are fine. Having super long import chains to import a sibling module doesn't really help.

This is also already enforced by https://docs.astral.sh/ruff/rules/relative-imports/#relative-imports-tid252

lyova24:
@provinzkraut, `just in defense of my tool`, i checked litestar repo with shrimport before and it actually ignored one-dot relative imports on my machine, so there is no problem in [shrimport](https://github.com/lyova24/shrimport)

Issue #4422, opened 2025-10-11
Enhancement: Consistent behaviour of flags set via environment variables


provinzkraut:
### Summary

Litestar uses environment variables in a few places (mostly the CLI). When that env variable represents a flag, it needs to be converted to a bool. This logic however does not behave consistently in all cases.

Since we're using `click` for the CLI, where most of the env vars are used, we should model the rest of the flags after click's behaviour, which is documented here: https://click.palletsprojects.com/en/stable/options/#values-from-environment-variables

<hr>

- [ ] Use click's `envvar` for all flags within the CLI. Do not perform any conversion manually
- [ ] Introduce new functionality to convert flags passed via env-vars into booleans, modeled after click's behaviour, and apply it everywhere in the codebase where flags are read from the environment

PR #4443, opened 2025-10-17
feat: make environment flag parsing consistent with `click`

7576457:
## Description

As mentioned in the issue, handling environment variables is currently inconsistent. 
This PR aims to address this by using Click for a more consistent solution:

- Add a new `--quiet-console` flag to control `LITESTAR_QUIET_CONSOLE`.
- Introduce `envflag()`, which returns True for `1`, `true`, `t`, `yes`, `on`, `y`, and False otherwise.
- Apply `envflag()` wherever CLI or other code reads environment variables representing flags.

## Closes

* gh-4422

### DIFF
--- a/.github/workflows/cherry-pick-v2.yaml
+++ b/.github/workflows/cherry-pick-v2.yaml
@@ -18,7 +18,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
   validate:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
@@ -25,7 +25,7 @@ jobs:
 
       - name: Load cached Pre-Commit Dependencies
         id: cached-pre-commit-dependencies
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ~/.cache/pre-commit/
           key: pre-commit-4|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
@@ -36,7 +36,7 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
@@ -53,7 +53,7 @@ jobs:
   pyright:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
@@ -70,7 +70,7 @@ jobs:
   slotscheck:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - uses: actions/setup-python@v6
         with:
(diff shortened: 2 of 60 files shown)


litestar/utils/helpers.py
```diff
@@ -2,6 +2,7 @@
 
 from enum import Enum
 from functools import partial
+from os import getenv
```
provinzkraut:
```suggestion
import os
```

litestar/utils/helpers.py
```diff
@@ -102,3 +103,25 @@ def get_exception_group() -> type[BaseException]:
         from exceptiongroup import ExceptionGroup as _ExceptionGroup  # pyright: ignore
 
         return cast("type[BaseException]", _ExceptionGroup)
+
+
+def envflag(varname: str, default: bool = False) -> bool:
```
provinzkraut, line 109: This does not behave like click. It will consider all values it doesn't consider `True` to be `False`, so e.g. a typo like `Tru` or `2` would also be considered `False`.

It should raise in that case instead. Please consult the click docs in the original issue I've linked on the actual behaviour

litestar/utils/helpers.py
```diff
@@ -102,3 +103,25 @@ def get_exception_group() -> type[BaseException]:
...
+        varname (str): The name of the environment variable to check.
+        default (bool): Value to return if the variable is not set OR empty.
+
+    Returns:
+        bool: True if the environment variable is set to a truthy value,
+        otherwise False.
+    """
+    envvar = getenv(varname)
+    if not envvar:
+        return default
```
provinzkraut: This will return `default` if the value is set to `""`, which is not what we want.

tests/unit/test_utils/test_helpers.py
```diff
@@ -45,3 +46,40 @@ def test_unique_name_for_scope() -> None:
     assert unique_name_for_scope("a", ["a", "a_0", "b"]) == "a_1"
 
     assert unique_name_for_scope("b", ["a", "a_0", "b"]) == "b_0"
+
+
+def test_envflag_truthy_values() -> None:
+    for value in ("1", "true", "TRUE", "t", "T", "yes", "YES", "on", "ON", "y", "Y"):
+        os.environ["TEST_FLAG"] = value
+        assert envflag("TEST_FLAG") is True
+        del os.environ["TEST_FLAG"]
```
provinzkraut: You should use the `monkeypatch` fixture for this instead.

https://docs.pytest.org/en/stable/how-to/monkeypatch.html

provinzkraut (changes requested):

litestar/utils/helpers.py
7576457: Thanks for the feedback! I’ve addressed your comments.

According to the click documentation you linked: "Any other value is interpreted as deactivating the flag". This means invalid values like `Tru` or `2` should return False, not raise an error. Could you clarify what behavior you're expecting?

7576457:
@provinzkraut 
I’ve applied your feedback and made the changes. Please review them.
Also, I have a question regarding raising an error.

I’m currently using `LitestarException`. Would it be better to create a more specific error, for example `LitestarEnvironmentError`, in `litestar/exceptions/base_exceptions.py`?

litestar/utils/helpers.py
```diff
@@ -102,3 +104,33 @@ def get_exception_group() -> type[BaseException]:
         from exceptiongroup import ExceptionGroup as _ExceptionGroup  # pyright: ignore
 
         return cast("type[BaseException]", _ExceptionGroup)
+
+
+def envflag(varname: str) -> bool | None:
```
provinzkraut: This should accept an optional `default` that is returned if the flag is not set. This would also resolve the current type issues

provinzkraut (changes requested):

litestar/utils/helpers.py
7576457: Done

provinzkraut:
Looks like you need to adjust the code to reflect the new default behaviour

7576457:
> Looks like you need to adjust the code to reflect the new default behaviour

Yes, you’re right. I’ve made the changes.
I only ended up using the `default` parameter in `litestar/utils/warnings.py`.
In the other places, it didn’t seem necessary

PR #4445, opened 2025-10-18
fix(typing): `DataContainerType` should contain types, not instances

sobolevn:
Since `mypy` and `pyright` both see a lot of `except ImportError:` above, they use `Any` and `Any` disables the type checking :(


sobolevn:
Ok, I give up. It is incorrect, but it somehow works.

PR #4459, opened 2025-11-05
fix: Enable provide with stateful callable and cleanup

### LINKED ISSUE #4457: Bug: `Provide` fails to detect generator methods on callable instances
## Description

The `Provide` class cannot properly detect when a callable instance's `__call__` method is a generator function. This breaks the [dependency injection with cleanup pattern](https://docs.litestar.dev/main/usage/dependency-injection.html#dependencies-with-yield-cleanup-step) when using stateful callable instances.

## Reproduction

```python
from typing import Generator, Any
from uuid import UUID, uuid4

class SessionSchedulerFactory:
    """A callable factory that maintains state across calls."""
    def __init__(self):
        self._sessions: dict[UUID, object] = {}
    
    def __call__(self) -> Generator[object, Any, None]:
        """Generator method that should be detected by Provide."""
        try:
            yield uuid4()
        finally:
            pass

def session_scheduler_factory() -> Generator[object, Any, None]: 
    try:
        yield uuid4()
    finally:
        pass

# Function works correctly
Provide(session_scheduler_factory, sync_to_thread=False).has_sync_generator_dependency  # True

# Callable instance fails to be detected
Provide(SessionSchedulerFactory(), sync_to_thread=False).has_sync_generator_dependency  # False
```

**Note:** This example is simplified for reproducibility. The real use case involves maintaining a cache of sessions across multiple dependency injections, which requires instance state.

## Root Cause

In `Provide.__init__`, `isclass(instance)` returns `False` for callable instances, so the code checks `isgeneratorfunction(instance)` instead of `isgeneratorfunction(instance.__call__)`:

```python
is_class_dependency = isclass(dependency)
self.has_sync_generator_dependency = isgeneratorfunction(
    dependency if not is_class_dependency else dependency.__call__
)
```

## Proposed Fix

Check callable instances the same way as classes:

```python
is_class_dependency = isclass(dependency)
is_callable_instance = not is_class_dependency and hasattr(dependency, '__call__')

check_target = dependency.__call__ if (is_class_dependency or is_callable_instance) else dependency
self.has_sync_generator_dependency = isgeneratorfunction(check_target)
self.has_async_generator_dependency = isasyncgenfunction(check_target)
```

### Litestar Version

litestar==2.18.0

enrico-stauss:
## Description
Fix the Provide class for callable class instances returning an (Async) Generator.
-

### DIFF
--- a/litestar/di.py
+++ b/litestar/di.py
@@ -59,7 +59,7 @@ class Provide:
 
         is_class_dependency = isclass(dependency)
         is_function = isfunction(dependency)
-        is_callable_instance = not is_class_dependency and not is_function and callable(dependency)
+        is_callable_instance = not is_class_dependency and not is_function
         if is_class_dependency or is_callable_instance:
             check_target = dependency.__call__  # type: ignore[operator]
         else:


litestar/di.py
```diff
@@ -58,12 +58,15 @@ def __init__(
...
 
         is_class_dependency = isclass(dependency)
-        self.has_sync_generator_dependency = isgeneratorfunction(
-            dependency if not is_class_dependency else dependency.__call__  # type: ignore[operator]
-        )
-        self.has_async_generator_dependency = isasyncgenfunction(
-            dependency if not is_class_dependency else dependency.__call__  # type: ignore[operator]
-        )
+        is_function = isfunction(dependency)
+        is_callable_instance = not is_class_dependency and not is_function and callable(dependency)
```
provinzkraut:
```suggestion
        is_callable_instance = not is_class_dependency and not is_function
```
`dependency` is already validated to be a callable just above.

enrico-stauss:
Hi @provinzkraut , thanks for your review, that was an obvious oversight on my end. I commited as per your suggestion and the PR should be ready to merge :)

enrico-stauss:
Hey @provinzkraut , any chance we can move on with this PR? :)

PR #4469, opened 2025-11-12
fix: duplicate query parameters fail validation if one of the paramaters is a list

### LINKED ISSUE #4242: Bug: multiple parameters with duplicate query works inconsistently
### Description

When using `Parameter(query=...` to rename the parameter like [this](https://docs.litestar.dev/2/usage/routing/parameters.html#alternative-names-and-constraints), it is possible to have the same query parameter map to multiple different Python parameters by setting `query` to an identical value.

Sometimes this works, and the provided HTTP query parameter gets copied into multiple Python parameters of different types.  Sometimes it does not work, and a validation error is returned. Behaviour seems to depend on what types it is being coerced into.

I would expect that using the same query value to assign a HTTP query param to multiple Python params this is either always allowed, or never allowed (and if it is not allowed, the current error message could be improved).

_No response_

### MCVE

```python
from typing import Annotated
from datetime import datetime

@get("/broken")
async def broken(
    some_date: Annotated[datetime, Parameter(query="someDate", required=True)],
    some_list: Annotated[
        list[str],
        Parameter(query="someDate", required=True),
    ],
) -> str:
    return f"{some_date=}, {some_list=}"

@get("/works")
async def works(
    some_date: Annotated[datetime, Parameter(query="someDate", required=True)],
    some_string: Annotated[
        str,
        Parameter(query="someDate", required=True),
    ],
) -> str:

    return f"{some_date=}, {some_string=}"

app = Litestar([broken, works])
```

### Steps to reproduce

1. Run the MCVE 
2. `curl 'localhost:8000/works?someDate=2025-01-01T00:00:00Z'`  - this works.
3. `curl 'localhost:8000/broken?someDate=2025-01-01T00:00:00Z'` - this does not work, and returns a validation error.

### Screenshots

_No response_

### Logs

```text
# curl 'localhost:8000/broken?someDate=2025-01-01T00:00:00Z'
{"status_code":400,"detail":"Validation failed for GET /broken?someDate=2025-01-01T00:00:00Z","extra":[{"message":"Expected `datetime`, got `array`","key":"some_date","source":"query"}]}%   

# curl 'localhost:8000/works?someDate=2025-01-01T00:00:00Z' 
some_date=datetime.datetime(2025, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), some_string='2025-01-01T00:00:00Z'%
```

### Litestar Version

2.16.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

cstarkers:
## Description

Allow a single query parameter to be aliased to multiple Python parameters, even when one of the Python parameters is a list.

Previously, if a query paramater was remapped to multiple different Python paramenters by using `Parameter(query="alias")` and one of the Python parameters was a list, a validation error was raised.  This was inconsisent with how non-list query parameters worked.

Fixes #4242

### DIFF
--- a/litestar/_kwargs/extractors.py
+++ b/litestar/_kwargs/extractors.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-from collections import defaultdict
-from functools import lru_cache, partial
+from functools import partial
 from typing import TYPE_CHECKING, Any, Callable, NamedTuple, cast
 
 from multidict import MultiDict, MultiDictProxy
@@ -39,7 +38,6 @@ __all__ = (
     "create_connection_value_extractor",
     "create_data_extractor",
     "create_multipart_extractor",
-    "create_query_default_dict",
     "create_url_encoded_data_extractor",
     "headers_extractor",
     "json_extractor",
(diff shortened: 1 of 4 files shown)


provinzkraut (commented):
Looks good in general, but I think this should take into account header parameters as well. The same semantics should apply there as well

cstarkers:
> Looks good in general, but I think this should take into account header parameters as well. The same semantics should apply there as well

Cool, done now

I also refactored some stuff as MultiDict, which I can imagine having a perf impact (probably positive, Multidict *should* be fast, but who knows, it could be negative)

Is there a performance testing suite/metrics for Litestar?

PR #4475, opened 2025-11-15
feat: Allows Rate Limit Config to be passed in as a list of [DurationUnit, int]

### LINKED ISSUE #1262: Enhancement: Rate Limit Config

thecyberbutler:
## Description

- Added the ability to pass the rate_limit in as a list.  Checks are performed to ensure that the types are correct, and then the list is cast into the tuple expected by the rest of the code.

Fixes #1262

### DIFF
--- a/litestar/middleware/rate_limit.py
+++ b/litestar/middleware/rate_limit.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from time import time
-from typing import TYPE_CHECKING, Any, Callable, Literal, cast, Union, get_args
+from typing import TYPE_CHECKING, Any, Callable, Literal, cast, get_args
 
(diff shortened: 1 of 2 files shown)


provinzkraut:
I think you have misunderstood the issue. The goal isn't to simply allow passing the same things in a list that you can already pass as a tuple (which is what your PR is doing). The goal is to allow specifying *multiple* rate limit conditions.

PR #4483, opened 2025-11-18
feat(otel): instrument Litestar specific components

cofin:
Since we rely on the upstream ASGI open telemetry integration, some of the Litestar specific concepts did not always get wrapped inside the correct span.  

This change allows the Open Telemetry plugin to instrument guards, events, and the CLI commands in Litestar.

### DIFF
--- a/litestar/plugins/opentelemetry/middleware.py
+++ b/litestar/plugins/opentelemetry/middleware.py
@@ -15,12 +15,12 @@ try:
 except ImportError as e:
     raise MissingDependencyException("opentelemetry") from e
 
+from opentelemetry import trace
 from opentelemetry.instrumentation.asgi import OpenTelemetryMiddleware
+from opentelemetry.trace import Status, StatusCode
 from opentelemetry.util.http import get_excluded_urls
 
 if TYPE_CHECKING:
-    from opentelemetry.trace import Status  # noqa: F401
-
     from litestar.plugins.opentelemetry import OpenTelemetryConfig
     from litestar.types import ASGIApp, Receive, Scope, Send
 
@@ -33,7 +33,6 @@ class OpenTelemetryInstrumentationMiddleware(AbstractMiddleware):
     - Enhanced error status tracking on spans
     """
 
-    __slots__ = ("config", "open_telemetry_middleware")
     __singleton_middleware__: ClassVar[OpenTelemetryMiddleware | None] = None
 
     def __init__(self, app: ASGIApp, config: OpenTelemetryConfig) -> None:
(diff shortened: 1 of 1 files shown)


litestar/plugins/opentelemetry/instrumentation.py
```diff
@@ -0,0 +1,253 @@
...
+        return _OTEL_AVAILABLE
+
+    try:
+        import opentelemetry  # noqa: F401
+
+        _OTEL_AVAILABLE = True
+    except ImportError:
+        _OTEL_AVAILABLE = False
+
+    return _OTEL_AVAILABLE
```
provinzkraut, line 58: I don't understand this part. Importing the otel module doesn't make any sense if otel isn't available

litestar/plugins/opentelemetry/instrumentation.py
```diff
@@ -0,0 +1,253 @@
...
+            yield span
+        except Exception as exc:
+            if span.is_recording():
+                span.record_exception(exc)
+                span.set_status(Status(StatusCode.ERROR, str(exc)))
+            raise
+
+
+@asynccontextmanager
+async def create_span_async(name: str, attributes: dict[str, Any] | None = None) -> AsyncGenerator[Any, None]:
```
provinzkraut, line 93: What's the purpose of this? It doesn't do anything async as far as I can tell?

litestar/plugins/opentelemetry/instrumentation.py
```diff
@@ -0,0 +1,253 @@
...
+        attributes: Optional attributes to add to the span
+
+    Yields:
+        The span object if OTEL is available, otherwise None
+    """
+    if not _is_otel_available():
+        yield None
+        return
+
+    from opentelemetry.trace import Status, StatusCode
```
provinzkraut, line 76: There really shouldn't be an import in a hot path

litestar/plugins/opentelemetry/instrumentation.py
```diff
@@ -0,0 +1,253 @@
...
+
+    tracer = _get_tracer(__name__)
+    with tracer.start_as_current_span(name) as span:
+        if attributes:
+            span.set_attributes(attributes)
+
+        try:
+            yield span
+        except Exception as exc:
```
provinzkraut, line 85: Is this necessary? I was under the impression that the active span already records exceptions

litestar/plugins/opentelemetry/instrumentation.py
```diff
@@ -0,0 +1,253 @@
...
+                "litestar.dependency.provider": provider_name,
+            },
+        ) as span:
+            result = provider_func(*args, **kwargs)
+            if inspect.isawaitable(result):
+                result = await result
+
+            # Add result type as attribute
+            if span and span.is_recording():
+                span.set_attribute("litestar.dependency.result_type", type(result).__name__)
```
provinzkraut, line 218: I think this might break for generator dependencies

PR #4488, opened 2025-11-22
fix(openapi): generate requestBody for body parameter

### LINKED ISSUE #4482: Bug: OpenAPI generation with binary request bodies
### Description

When using the `body` reserved keyword parameter in the route definition, no `requestBody` is generated in the OpenAPI spec.

Related: https://github.com/orgs/litestar-org/discussions/3038

### Root Cause

https://github.com/litestar-org/litestar/blob/bcea856542f095ef5c78a6fff489a232c3078240/litestar/_openapi/path_item.py#L64-L68

The code above only checks `data`.

Please let me know whether not checking for `body` is intentional; if it isn't, I'm happy to submit a PR with a fix.

_No response_

### MCVE

```python
from litestar.response import Response
from litestar.status_codes import HTTP_201_CREATED


@post(
    path="/upload/",
    operation_id="upload_binary",
    status_code=HTTP_201_CREATED,
)
async def upload_binary(
    body: bytes,
    # body: bytes = Body(media_type="application/octet-stream"), # <-- does not work as well
) -> Response[dict[str, str]]:
    return Response(
        content={
            "message": "Data uploaded successfully",
            "size": len(body),
        },
        status_code=HTTP_201_CREATED,
    )


app = Litestar(route_handlers=[upload_binary])
```

### Steps to reproduce

1. Create the above file and save to `app.py`
2. `litestar run`
3.  `curl -O http://localhost:8000/schema/openapi.yaml`
4. `cat openapi.yaml`

```yaml
components:
  schemas: {}
info:
  title: Litestar API
  version: 1.0.0
openapi: 3.1.0
paths:
  /upload:
    post:
      deprecated: false
      operationId: upload_binary
      responses:
        '201':
          content:
            application/json:
              schema:
                additionalProperties:
                  type: string
                type: object
          description: Document created, URL follows
          headers: {}
      summary: UploadBinary
servers:
- url: /
```

I believe the output should look like this:

```yaml
paths:
  /upload:
    post:
      operationId: upload_binary
      requestBody:  # ← Should be present
        required: true
        content:
          application/octet-stream:
            schema:
              type: string
              format: binary
```

Naturally, this results in incorrect OpenAPI client generation.

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.18.0

### Platform

- [x] Linux
- [x] Mac
- [ ] Windows

burnash:
## Description

OpenAPI generation now correctly creates requestBody when using the `body: bytes` parameter. Previously only `data` parameter was checked.

- Check both `data` and `body` fields in `path_item.py`
- Default to `application/octet-stream` for `bytes` type
- Add tests for all body parameter syntaxes

## Closes

Fixes #4482

### DIFF
--- a/litestar/_openapi/path_item.py
+++ b/litestar/_openapi/path_item.py
@@ -62,12 +62,17 @@ class PathItemFactory:
         signature_fields = route_handler.parsed_fn_signature.parameters
 
         request_body = None
+        request_body_field = None
         if data_field := signature_fields.get("data"):
             request_body = create_request_body(
                 self.context, route_handler.handler_id, route_handler.data_dto, data_field
             )
+            request_body_field = data_field
+        elif body_field := signature_fields.get("body"):
+            request_body = create_request_body(self.context, route_handler.handler_id, None, body_field)
+            request_body_field = body_field
 
-        raises_validation_error = bool(data_field or self._path_item.parameters or parameters)
+        raises_validation_error = bool(request_body_field or self._path_item.parameters or parameters)
         responses = create_responses_for_handler(
             self.context, route_handler, raises_validation_error=raises_validation_error
         )
--- a/litestar/_openapi/request_body.py
+++ b/litestar/_openapi/request_body.py
@@ -34,7 +34,9 @@ def create_request_body(
     Returns:
         A RequestBody instance.
     """
-    media_type: RequestEncodingType | str = RequestEncodingType.JSON
+    media_type: RequestEncodingType | str = (
+        "application/octet-stream" if data_field.is_subclass_of(bytes) else RequestEncodingType.JSON
+    )
     schema_creator = SchemaCreator.from_openapi_context(context, prefer_alias=True)
     if isinstance(data_field.kwarg_definition, BodyKwarg) and data_field.kwarg_definition.media_type:
         media_type = data_field.kwarg_definition.media_type
--- a/tests/unit/test_openapi/test_request_body.py
+++ b/tests/unit/test_openapi/test_request_body.py
@@ -217,3 +217,69 @@ def test_unwrap_read_only() -> None:
             }
         }
     }
+
+
+def test_body_parameter_binary_request() -> None:
+    @post("/upload/")
+    async def handle_binary_upload(body: bytes) -> None:
+        return None
+
+    app = Litestar([handle_binary_upload])
+    schema = app.openapi_schema.to_schema()
+
+    assert "requestBody" in schema["paths"]["/upload"]["post"]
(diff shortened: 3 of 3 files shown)


litestar/_openapi/path_item.py
```diff
@@ -62,12 +62,17 @@ def create_operation_for_handler_method(
...
             request_body = create_request_body(
                 self.context, route_handler.handler_id, route_handler.data_dto, data_field
             )
+            request_body_field = data_field
+        elif body_field := signature_fields.get("body"):
+            request_body = create_request_body(self.context, route_handler.handler_id, None, body_field)
+            request_body_field = body_field
 
-        raises_validation_error = bool(data_field or self._path_item.parameters or parameters)
+        raises_validation_error = bool(request_body_field or self._path_item.parameters or parameters)
```
provinzkraut, line 75: I don't think we should set `raises_validation_error=True` in this case, as the `body` parameter is only allowed to be `bytes` and will always receive the raw request body

litestar/_openapi/request_body.py
```diff
@@ -34,7 +34,9 @@ def create_request_body(
     Returns:
         A RequestBody instance.
     """
-    media_type: RequestEncodingType | str = RequestEncodingType.JSON
+    media_type: RequestEncodingType | str = (
```
provinzkraut, line 37: This is incorrect; `body: bytes` is just the raw request body, it does not signify a binary payload. That would be `data: bytes`. So no changes should be made here

tests/unit/test_openapi/test_request_body.py
```diff
@@ -217,3 +217,69 @@ async def handler(
             }
         }
     }
+
```
provinzkraut, line 220: I think these tests would be more readable if you used `pytest.mark.parametrize`

litestar/_openapi/request_body.py
burnash: I'm new to the framework and not fully confident about the correct place to enforce the expected behavior.
Could you outline where in the existing code (e.g. specific modules or functions) the fix should be implemented, so I can update the PR accordingly?

provinzkraut: I think there might be a slight misunderstanding! 

There's no need to change anything here, you just need to revert the changes you've made in this section. 

Currently, (with your changes implemented - this line here) the proper way to achieve what you want would be to do something like:

```python
@post("/")
async def handler(body: bytes = Body(content_encoding="application/octet-stream")) -> None:
    pass
```

Am I correct in assuming that what you're asking is to be able to omit the `Body` param, and default to `application/octet-stream` for a bare `body: bytes`?

provinzkraut:
@burnash Are you still working on this?

PR #4495, opened 2025-11-26
feat: add plain text traceback to HTML debug response context

JacobCoffee:
## Description

When working with django its nice to have a copypasta button to show people or feed a model.
I'd like to upstream this thing ive been using if it is seen as useful.

<details>
<summary>Example App</summary>

also see:
<details>
<summary>Images</summary>

image
image

</details>

MCVEish

```py
"""MCVE demonstrating the Copy Traceback feature in Litestar debug pages.

Run with: `uv run litestar --app example_exceptions_copy_app:app run --debug`
    
View:
- Single:  http://127.0.0.1:9000/
- Chained: http://127.0.0.1:8000/chained

Then visit http://localhost:8000/ in your browser to see the debug page.
Click "Switch to copy-and-paste view" to see the plain text traceback,
then click "Copy to clipboard" to copy it.
"""

def calculate_value(x: int) -> float:
    """A function that will fail with division by zero."""
    return 100 / x


def process_data(data: dict) -> float:
    """Process data and calculate a value."""
    value = data.get("value", 0)
    return calculate_value(value)


@get("/")
async def index() -> dict:
    """Endpoint that triggers an exception for demonstration."""
    data = {"name": "test", "value": 0}
    result = process_data(data)
    return {"result": result}


@get("/chained")
async def chained_exception() -> dict:
    """Endpoint that demonstrates chained exceptions."""
    try:
        data = {"value": 0}
        process_data(data)
    except ZeroDivisionError as e:
        raise ValueError("Failed to process data") from e
    return {"status": "ok"}


@get("/healthy")
async def healthy() -> dict:
    """A working endpoint for comparison."""
    return {"status": "healthy"}


app = Litestar(
    route_handlers=[index, chained_exception, healthy],
    debug=True,
)
```
</details>


JacobCoffee:
assuming its something we want to merge, could this be backported

provinzkraut:
Nice, I like it! One suggestion though: Could this be two separate buttons that you can click straight trough?

So you'd have:
- A "copy traceback to clipboard"
- A "plaintext view" / "interactive view" toggle

provinzkraut (approved):

PR #4506, opened 2025-11-30
feat: make route reverse take handlers

### LINKED ISSUE #4498: Enhancement: Litestar.route_reverse should be able to take handler objects
### Summary

In my code, I have this pattern:

```
    @get("/")
    def base_route(self) -> Response[bytes]:
        ...  # actual logic

    @get("/notifications")
    def notifications_route(self, request: HTMXRequest) -> Response[Any]:
        if not request.htmx:
            return Redirect(request.app.route_reverse(str(self.base_route)))`

        ...  # actual logic
```

The user visits the root route, which returns a rendered template that, among other things, lazy-loads a few sub-templates that are served through routes like `notifications_route`. These routes are meant to be accessed only through HTMX, not directly by the user. In the event that that happens, the user should be redirected back to the root route.

I want to do this in a way that is robust to changes in both the (method) name and path of the base route. I am also not really a fan of explicitly specifying a handler name, since that doesn't always play well with code analysis and automated refactoring. However, I understand (purely as an implementation detail, I believe) that `HTTPRouteHandler.__str__` generates a unique name, which leads to what I have done above. It'd be nice if I could just write `Redirect(request.app.route_reverse(self.base_route))`, which feels more intuitive to me.

Alternatively, might it make sense to set `name` in `HTTPRouteHandler.__init__` if not provided by the user...? That would be simpler, but also backwards incompatible.

### Basic Example

```
    @get("/notifications")
    def notifications_route(self, request: HTMXRequest) -> Response[Any]:
        if not request.htmx:
            return Redirect(request.app.route_reverse(self.base_route))`
```

### Drawbacks and Impact

idk tbh...doesn't feel that much more complex than the existing impl

### Unresolved questions

_No response_

marcuslimdw:
## Description

- Modify `route_reverse` and downstream consumers to take `BaseRouteHandler` instances.
- Standardise the docstrings for route name parameters.

Closes #4498.


docs/migration/fastapi.rst
```diff
@@ -334,38 +334,6 @@ Also FastAPI let you pass a dictionary while in Litestar you need to explicitly
                     name="uploads.html", context={"debug": app_settings.debug}
                 )
 
-Default handler names
-~~~~~~~~~~~~~~~~~~~~~~~
-
-In FastAPI, the handler name defaults to the local name of the function. In Litestar, you need to explicitly declare the `name` parameter in the route decorator. This is important when using e.g. `url_for`.
```
marcuslimdw, line 340: is there a reason for this, other than usage of `url_for` and `route_reverse` (which will be made unnecessary by this PR)?

litestar/connection/base.py
```diff
@@ -307,11 +309,11 @@ def clear_session(self) -> None:
     def get_session_id(self) -> str | None:
         return value_or_default(value=self._connection_state.session_id, default=None)
 
-    def url_for(self, name: str, **path_parameters: Any) -> str:
+    def url_for(self, name: str | BaseRouteHandler, **path_parameters: Any) -> str:
         """Return the url for a given route handler name.
```
provinzkraut:
```suggestion
        """Return the url for a given route handler / handler name.
```

docs/migration/fastapi.rst
provinzkraut: Nope, it can go with these changes

provinzkraut:
Whoops, looks like I forgot to hit merge on this one :)

PR #4509, opened 2025-12-02
Fix: Optional field regex validation in MsgspecDTO (#4195) 

### LINKED ISSUE #4195: Bug: WriteDTO ignores msgspec constraints when marked as Optional
### Description

When `URL` is marked as `Optional`, the Regex constraint never triggers allowing invalid input to be used.

_No response_

### MCVE

```python

from typing import Annotated, Optional

import msgspec
from litestar.dto import DTOConfig, MsgspecDTO

URL = Annotated[
    str,
    msgspec.Meta(
        pattern=r"(https?:\/\/)([\w\-])+\.{1}([a-zA-Z]{2,63})([\/\w-]*)*\/?\??([^#\n\r]*)?#?([^\n\r]*)",
    ),
]


class Base(msgspec.Struct):
    url: Optional[URL] = None
    foo: bool = True


class WriteDTO(MsgspecDTO[Base]):
    config = DTOConfig(exclude={"foo"})


@post("/", dto=WriteDTO)
async def test(data: Base) -> Base:
    return data


if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app)
```

### Steps to reproduce

1. POST http://localhost:8000/ {"url": "htp:/gogle.com"}
2. Response will return 201.
3. Remove WriteDTO from / endpoint
4. Response will error correctly with a ValidationException

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.16.0

### Platform

- [ ] Linux
- [ ] Mac
- [x] Windows

rsamanth1:
## Problem

When a field is marked as `Optional[Annotated[str, msgspec.Meta(pattern=...)]]`, the regex constraint validation was bypassed, allowing invalid input to pass through.

**Before:**
```python
POST / {"url": "htp:/gogle.com"}  # Returns 201 (should be 400)
```

**After:**
```python
POST / {"url": "htp:/gogle.com"}  # Returns 400 
```

## Solution

Added post-decoding validation that:
1. Extracts `Meta` constraints from `Optional[Annotated[T, Meta(...)]]` types
2. Validates non-None values against these constraints  
3. Raises `ValidationException` for invalid patterns

## Changes

- Modified `msgspec_dto.py`:
  - Added `_extract_meta_from_annotation()` to unwrap Optional/Annotated types
  - Override `decode_builtins()` and `decode_bytes()` to add validation
  - Added `_validate_meta_constraints()` and `_validate_instance()` methods
  
- Added comprehensive tests in `test_dto_optional_validation.py`:
  - Invalid URLs rejected with 400
  - Valid URLs accepted with 201
  - None values allowed for optional fields
  - Required fields still validate correctly
  - Parametrized tests for various URL patterns

## Testing
```bash
pytest tests/unit/dto/test_dto_optional_validation.py -v
```

### DIFF
--- a/litestar/dto/msgspec_dto.py
+++ b/litestar/dto/msgspec_dto.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import dataclasses
 from dataclasses import replace
-from typing import TYPE_CHECKING, Generic, TypeVar
+from typing import TYPE_CHECKING, Generic, TypeVar, Optional, Callable, Any
 
 import msgspec.inspect
-from msgspec import NODEFAULT, Struct, structs
+from msgspec import NODEFAULT, Struct, structs, Meta
 
 from litestar.dto.base_dto import AbstractDTO
 from litestar.dto.data_structures import DTOFieldDefinition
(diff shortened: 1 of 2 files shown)


provinzkraut:
I believe this is following msgspec's logic. The field annotation should be `Annotated[Optional[str], msgspec.Meta(pattern=...)]` not `Optional[Annotated[str, msgspec.Meta(pattern=...)]]`. Have you verified that it works as intended with the correct annotation?

litestar/dto/msgspec_dto.py
```diff
@@ -15,23 +15,32 @@
...
+    def validator(value: str | None) -> str | None:
+        if value is None:
+            return None
+        if not re.fullmatch(pattern, value):
+            raise ValueError(f"Invalid value: {value}")
+        return value
+    return validator
+
+# Example usage:
+# URL = Annotated[Optional[str], optional_pattern(r"regex_here")]
```
provinzkraut, line 42: This does not work. There is no code here using that functionality.

PR #4515, opened 2025-12-06
fix(docs): proper delete methods in repository tutorial

vladpi:
## Description
- In the repository tutorial, incorrect deletion methods are specified in the text – `remove` instead of `delete`.


provinzkraut:
@cofin

cofin:
Thanks @provinzkraut `delete_many` and `delete` is correct for the service, but this is a good catch on the repository.  

This might be something to considering unifying for the Advanced Alchemy 2.0 release though.

@vladpi Thanks for the change.

cofin (approved):

PR #4524, opened 2025-12-13
feat(v2): Support Python 3.14

provinzkraut:
Add tests and apply necessary package updates to properly support 3.14.


litestar/utils/sync.py
```diff
@@ -41,7 +41,7 @@ class AsyncCallable:
     def __init__(self, fn: Callable[P, T]) -> None:  # pyright: ignore
         self.func = fn
 
-    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # pyright: ignore
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> Awaitable[T]:  # type: ignore[valid-type] # pyright: ignore
```
sobolevn: This new type ignore seems sus 🤔

provinzkraut: Agree. I'm also not sure about this. It's reporting a `ParamSpec "P" is unbound`. It seems to work fine when defining the param spec using parameter list, but when using `ParamSpec` directly, it's breaking. Not sure how to fix this

PR #4525, opened 2025-12-14
fix(v2):  Fix `TypedDict` required field on 3.8

provinzkraut:
Ensure required fields are detected properly, by backporting the `__required_keys__` logic introduced in Python 3.9


provinzkraut:
cc @sobolevn

PR #4526, opened 2025-12-14
ci: add manual docs things with selective input (#4258)

JacobCoffee:
## Description

- tries to fix no version swpper in prod docs

Open https://docs.litestar.dev/latest/index.html and in console:
```
// current
console.log(DOCUMENTATION_OPTIONS.URL_ROOT);  // undefined
console.log(DOCUMENTATION_OPTIONS.URL_ROOT + "_static/versions.json");  // "undefined_static/versions.json"

// w/ fix
const urlRoot = DOCUMENTATION_OPTIONS.URL_ROOT ?? "";
fetch(urlRoot + "_static/versions.json").then(r => r.json()).then(console.log);
```
image

### DIFF
--- a/docs/_static/versioning.js
+++ b/docs/_static/versioning.js
@@ -1,7 +1,7 @@
+const urlRoot = DOCUMENTATION_OPTIONS.URL_ROOT ?? "";
+
 const loadVersions = async () => {
-  const res = await fetch(
-    DOCUMENTATION_OPTIONS.URL_ROOT + "_static/versions.json",
-  );
+  const res = await fetch(urlRoot + "_static/versions.json");
   if (res.status !== 200) {
     return null;
   }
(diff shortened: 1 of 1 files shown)


docs/_static/versioning.js
```diff
@@ -1,7 +1,7 @@
+const urlRoot = DOCUMENTATION_OPTIONS.URL_ROOT ?? "";
```
provinzkraut, line 1: Why does this value not exist sometimes?

JacobCoffee: i dont know, thats the better question :\ i was trying to look through changelogs on sphinx7/8 but im not finding anything to point why it all of a sudden broke some time ago for us. i can keep poking

provinzkraut: Ah. Well, let's keep it for now :woman_shrugging:

PR #4534, opened 2025-12-23
feat: Added exclude_spans option for otel config

### LINKED ISSUE #4533: Enhancement: Add option to otel plugin to exclude_spans
### Summary

Enable passing the exclude_spans kwarg to OpenTelemetryMiddleware to allow excluding send receive events in order to reduce noise and improve performance

### Basic Example

```python
OpenTelemetryConfig(
    exclude_spans=["send", "receive"],
)
```

### Drawbacks and Impact

No drawbacks this is just an option

### Unresolved questions

No unresolved questions

vikigenius:
## Description

- Adds a config option to exclude_spans for the Otel middleware
- This allows excluding the send receive spans to improve performance and reduce noise for example in streaming endpoints where every stream event is logged as an ASGI event

### DIFF
--- a/litestar/contrib/opentelemetry/config.py
+++ b/litestar/contrib/opentelemetry/config.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, Literal
 
 from litestar.contrib.opentelemetry._utils import get_route_details_from_scope
 from litestar.contrib.opentelemetry.middleware import (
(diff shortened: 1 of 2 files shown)


litestar/contrib/opentelemetry/config.py
```diff
@@ -77,6 +77,8 @@ class OpenTelemetryConfig:
     OpenTelemetry supports excluding urls by passing an env in the format '{exclude_urls_env_key}_EXCLUDED_URLS'. With
     the default being ``LITESTAR_EXCLUDED_URLS``.
     """
+    exclude_spans: list[Literal["receive", "send"]] | None = None
```
provinzkraut, line 80: I think it would make sense to not restrict this to ASGI spans, as it's global config for the integration that can be used to exclude other spans as well

vikigenius: Huh, I am not sure how to make this apply to non ASGI spans, i just copy pasted the docstring from the OTEL docs

https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/asgi/asgi.html#opentelemetry.instrumentation.asgi.OpenTelemetryMiddleware

I think by default it's just send receive ASGI spans.

Would it make more sense to call it something different on the litestar side

```exclude_asgi_spans```

vikigenius: We can even do it similar to how logfire does it for FastAPI `record_send_receive` and default it to False 

https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.Logfire.instrument_fastapi(request_attributes_mapper)

provinzkraut: Ah, guess I'm mistaken then. Alright, let's keep this as-is

PR #4539, opened 2026-01-04
fix(routing): Fix header name when raising `MethodNotAllowedException`

edgarrmondragon:
## Description

https://github.com/litestar-org/litestar/pull/4289 used the wrong header name.

## Closes

Closes https://github.com/litestar-org/litestar/issues/4277 (for good).


sobolevn (approved):
Oups :)

LGTM

PR #4542, opened 2026-01-06
fix(security): Fix JWT cookie authentication to handle tokens without Bearer prefix

### LINKED ISSUE #4541: Bug: JWTCookieAuthenticationMiddleware fails to authenticate tokens from cookies
### Description

**Bug Summary**

  JWTCookieAuthenticationMiddleware.authenticate_request fails to correctly parse JWT tokens stored in cookies because it unconditionally applies partition(" ")[-1] to extract the token, assuming all tokens have a "Bearer " prefix.

  **Current Behavior**

  When a JWT token is provided via cookie (without "Bearer " prefix), authentication fails with a 401 Unauthorized error, even though the token itself is valid.

 **Expected Behavior**

  JWT tokens stored in cookies should be used directly without expecting a "Bearer " prefix, since cookies typically store raw JWT values.

  **Root Cause**

  The current implementation at litestar/security/jwt/middleware.py:265-268:

```
  auth_header = connection.headers.get(self.auth_header) or connection.cookies.get(self.auth_cookie_key)
  if not auth_header:
      raise NotAuthorizedException("No JWT token found in request header or cookies")
  encoded_token = auth_header.partition(" ")[-1]  # BUG: fails for cookies
```

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.19.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

0asys:
## Description

  - Fix JWTCookieAuthenticationMiddleware.authenticate_request to correctly handle JWT tokens from cookies that don't have a "Bearer " prefix
  - Add regression test to ensure cookie-based authentication works with raw JWT tokens

  Handle header and cookie tokens separately:
  1. Try Authorization header first → apply partition(" ")[-1] to strip "Bearer " prefix
  2. Fall back to cookie → use token directly (no prefix expected)

  - Added test_jwt_cookie_auth_without_bearer_prefix regression test that verifies:
    - Raw JWT token in cookie authenticates successfully
    - Bearer-prefixed token in header still works

## Closes

fixes #4541

### DIFF
--- a/tests/unit/test_security/test_jwt/test_auth.py
+++ b/tests/unit/test_security/test_jwt/test_auth.py
@@ -366,6 +366,54 @@ async def test_jwt_cookie_auth(
             assert response.json()["message"] == f"can't logout, jti is {decoded_token.jti}"
 
 
+async def test_jwt_cookie_auth_without_bearer_prefix() -> None:
+    """Test that JWTCookieAuth correctly handles raw JWT tokens in cookies without 'Bearer ' prefix.
+
+    This is a regression test for a bug where cookie tokens were incorrectly parsed.
+    The original code used `partition(" ")[-1]` on all tokens, which works for header
+    tokens with "Bearer " prefix but returns an empty string for raw cookie tokens
+    (since str.partition(" ") on a string without spaces returns (original, "", "")).
+    """
+    user = UserFactory.build()
+    token_secret = secrets.token_hex()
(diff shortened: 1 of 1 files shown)


0asys:
I realised that litestar adds the Bearer prefix.

PR #4552, opened 2026-01-10
fix(jwt): store raw JWT token in cookies without Bearer prefix

disrupted:
## Description

when switching to `JWTCookieAuth` I noticed how it incorrectly stores the JWT token with a "Bearer " prefix in the cookie value.

AFAICT this violates:

- RFC 6265 (HTTP cookies): cookies should contain just the token value
- RFC 6750 (Bearer token usage): "Bearer" is an Authorization header scheme, bearer tokens should not be stored in cookies

I also updated the middleware accordingly. `OAuth2PasswordBearerAuth` was also affected.

## Todo

- [x] update tests

### DIFF
--- a/litestar/security/jwt/middleware.py
+++ b/litestar/security/jwt/middleware.py
@@ -250,8 +250,8 @@ class JWTCookieAuthenticationMiddleware(JWTAuthenticationMiddleware):
         self.auth_cookie_key = auth_cookie_key
 
     async def authenticate_request(self, connection: ASGIConnection[Any, Any, Any, Any]) -> AuthenticationResult:
-        """Given an HTTP Connection, parse the JWT api key stored in the header or cookie and retrieve the user correlating
-        to the token from the DB.
+        """Given an HTTP Connection, parse the JWT api key stored in the header and retrieve the user correlating to the
+        token from the DB.
 
         Args:
             connection: An Litestar HTTPConnection instance.


disrupted:
is it really expected behavior that the `JWTCookieAuthenticationMiddleware` reads from Authorization header? I certainly did not expect it and was surprised to find this ambiguity in the security layer. imo the next major release (3.0) would be a good opportunity to make such breaking changes if you agree that the behavior should be more predictable.
I am mostly interested in fixing the cookie auth which is currently broken. I can partially rollback the middleware changes but I believe it would still make sense to discuss this further.

provinzkraut:
> is it really expected behavior that the `JWTCookieAuthenticationMiddleware` reads from Authorization header? 

No, it doesn't. That's why I said it's questionable :)

I just wanted to point out that, if this change is being made, it would need to be an intentional breaking change. That's a separate change from the one that this PR initially suggested to make:

> when switching to JWTCookieAuth I noticed how it incorrectly stores the JWT token with a `"Bearer "` prefix in the cookie value.

Should be separate PRs IMO, the individual changes are fine.

disrupted:
> Should be separate PRs IMO

agreed

disrupted:
not sure why the `build-docs` job is failing

euri10:
> not sure why the `build-docs` job is failing

nothing woring on your side there are issues reaching the alchemy docs

provinzkraut:
Thanks @disrupted!

PR #4561, opened 2026-01-26
fix(openapi): generate parameters for model-typed query/headers/cookies kwargs 

### LINKED ISSUE #2015: Enhancement: Support for Typing the `query` keyword (and other reserved kwargs) to generate OpenAPI Spec

anishgirianish:
<!--
By submitting this pull request, you agree to:
- follow [Litestar's Code of Conduct](https://github.com/litestar-org/.github/blob/main/CODE_OF_CONDUCT.md)
- follow [Litestar's contribution guidelines](https://github.com/litestar-org/.github/blob/main/CONTRIBUTING.md)
- follow the [PSFs's Code of Conduct](https://www.python.org/psf/conduct/)
-->
## Description
 
## Summary                                                                                                                             
                                                                                                                                      
- Decompose model-typed query, headers, and cookies reserved kwargs into individual OpenAPI parameters                              
- Supports dataclasses, Pydantic, msgspec Struct, TypedDict, and attrs                                                              
- Backward compatible — non-model types like dict are unaffected                                                                    
                                                                                                                                                                                                                                                                            
## Test plan                                                                                                                           
                                                                                                                                      
- Parametrized tests across all 5 model types                                                                                       
- Backward compat, optional/required fields, headers/cookies support                                                                
- E2E and issue #2015 regression tests  


<!--
Please add in issue numbers this pull request will close, if applicable
Examples: Fixes #4321 or Closes #1234

Ensure you are using a supported keyword to properly link an issue:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
-->
## Closes #2015

### DIFF
--- a/litestar/_openapi/parameters.py
+++ b/litestar/_openapi/parameters.py
@@ -36,7 +36,7 @@ def _is_struct_type(annotation: Any) -> bool:
         from msgspec import Struct
 
         return isinstance(annotation, type) and issubclass(annotation, Struct)
-    except ImportError:
+    except ImportError:  # pragma: no cover
         return False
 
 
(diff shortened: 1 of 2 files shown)


provinzkraut:
This does not fit in with Litestar's existing design, it re-implements all the model parsing logic, is not generic on model types as it hard-codes support for pydantic/attrs/msgspec/dataclasses, and does not address the issues raised in the first attempt / linked issue, or the ones it states it solves.

I'll be closing this. 

@anishgirianish I appreciate your attempt to contribute, but I'd ask you to refrain from opening any more PRs unless they're done with a little more care, and you've either made yourself more familiar with the code base, or asked for guidance beforehand.

anishgirianish:
Hi @provinzkraut 

Thanks for the feedback, and apologies for the noise. I should have studied the plugin architecture more carefully before opening this.                                    
                                                                                                                                                                             
If I understand correctly, the right approach would be to reuse what OpenAPISchemaPlugin.to_openapi_schema() already produces, which returns a schema with required and properties resolved, and decompose that into individual parameters, rather than re-implementing per-model-type field extraction in parameters.py.                         
                                                                                                                                                                             
is that the right direction, or would you suggest a different approach? Happy to take any pointers before trying again.

PR #4562, opened 2026-01-29
fix(stores): validate mutually exclusive expires_in and keep_ttl in RedisStore.set

agarwalkunal2707:
Problem
The Redis SET command does not allow both EX (expiration time) and KEEPTTL options to be specified together—they are mutually exclusive. Previously, if a user passed both parameters, the call would be forwarded to Redis which would return a protocol error.

Solution
Added an early validation check that raises a clear ImproperlyConfiguredException with an actionable error message before the Redis call is made.

Changes
Added parameter validation in RedisStore.set()
Updated docstring to document the new exception

### DIFF
--- a/tests/unit/test_stores.py
+++ b/tests/unit/test_stores.py
@@ -236,6 +236,34 @@ def test_redis_with_non_default(connection_pool_from_url_mock: Mock, mock_redis:
     assert backend._redis is mock_redis.return_value
 
 
+@patch("litestar.stores.redis.Redis")
+@patch("litestar.stores.redis.ConnectionPool.from_url")
+async def test_redis_set_with_expires_in_and_keep_ttl_raises(
+    connection_pool_from_url_mock: Mock, mock_redis: Mock
+) -> None:
+    """Test that setting both expires_in and keep_ttl raises ValueError."""
+    store = RedisStore.with_client()
+
+    with pytest.raises(ValueError, match="Cannot set both 'expires_in' and 'keep_ttl'"):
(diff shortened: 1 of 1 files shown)


litestar/stores/redis.py
```diff
@@ -140,20 +140,30 @@ def _make_key(self, key: str) -> str:
...
+            keep_ttl: If ``True``, the TTL of the key will not be changed. If ``False``, the TTL of the key will be set to the value of ``expires_in``
+
+        Raises:
+            ImproperlyConfiguredException: If both ``expires_in`` and ``keep_ttl`` are set, as these options are mutually exclusive
 
         Returns:
             ``None``
         """
+        if expires_in is not None and keep_ttl:
+            raise ImproperlyConfiguredException(
```
provinzkraut: I think a `ValueError` is more appropriate here

litestar/stores/redis.py
```diff
@@ -140,20 +140,30 @@ def _make_key(self, key: str) -> str:
         prefix = f"{self.namespace}:" if self.namespace else ""
         return prefix + key
 
-    async def set(self, key: str, value: str | bytes, expires_in: int | timedelta | None = None) -> None:
+    async def set(
```
provinzkraut, line 144: Can you add `typing.overload`s, so this issue can be caught early by the type checker?

agarwalkunal2707: done

litestar/stores/redis.py
agarwalkunal2707: done

agarwalkunal2707:
added test cases

PR #4579, opened 2026-02-05
docs(aa): fix the url of advanced-alchemy project

Kumzy:
This PR fixes the incorrect AA intersphinx link meanwhile we fix the issue with the DNS


provinzkraut:
@Kumzy seems that some other stuff is breaking now

provinzkraut (approved):

PR #4581, opened 2026-02-06
fix(tests): `test_encode_validation` did not use `algorithm` param

sobolevn:
I am going through tests, because I am writting my own ones for the same feature.


sobolevn:
Ha! :) 

I found one more case. I will open a new PR, because test is different.

PR #4583, opened 2026-02-06
fix(jwt): issuer validation in jwt token

sobolevn:
According to the docs, `issuer` can be `str | Container[str] | None` https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode

We incorrectly transformed `str` into `list`, basically we had something like this:

```python
>>> issuer = 'my-awesome-company'
>>> list(issuer)
['m', 'y', '-', 'a', 'w', 'e', 's', 'o', 'm', 'e', '-', 'c', 'o', 'm', 'p', 'a', 'n', 'y']
```

Now it is fixed, test was added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ litestar run
   - [Swagger-UI](https://swagger.io/tools/swagger-ui/)
 - [Trio](https://trio.readthedocs.io/en/stable/) support (built-in, via [AnyIO](https://anyio.readthedocs.io/))
 - Ultra-fast validation, serialization and deserialization using [msgspec](https://github.com/jcrist/msgspec)
-- [SQLAlchemy integration](https://docs.advanced-alchemy.litestar.dev/latest/)
+- [SQLAlchemy integration](https://advanced-alchemy.litestar.dev/latest/)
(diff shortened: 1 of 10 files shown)


tests/unit/test_security/test_jwt/test_token.py
```diff
@@ -262,3 +262,23 @@ def test_token_encode_includes_custom_headers() -> None:
     assert header["alg"] == "HS256"
     assert "kid" in header
     assert header["kid"] == custom_headers["kid"]
+
+
+@pytest.mark.parametrize('issuer', [None, 'text', ['list', 'of', 'values']])
```
sobolevn:
```suggestion
@pytest.mark.parametrize("issuer", [None, "text", ["list", "of", "values"]])
```

sobolevn:
@provinzkraut sorry, but I can't merge :(
Because [Tests And Linting / build-docs (pull_request)](https://github.com/litestar-org/litestar/actions/runs/21763904668/job/62794414967?pr=4583) is required and I don't have permissions to bypass it.

PR #4588, opened 2026-02-09
fix(logging): emit log without stack trace when exception is in disable_stack_trace

### LINKED ISSUE #4510: disable_stack_trace suppresses exception logging entirely instead of only stack traces
### Description

When using `StructLoggingConfig.disable_stack_trace` with the Structlog plugin, exceptions matching the configured HTTP status codes / exception classes are not logged at all — neither message nor stack trace.

From the docs I expected that only the stack trace would be suppressed, while a one‑line error log (e.g. with message, level, path) would still be emitted.

The relevant code path seems to be in [ExceptionHandlerMiddleware.handle_exception_logging](https://github.com/litestar-org/litestar/blob/b4fad371bb991aa456ef8fe6dd28137fb44c1c6b/litestar/middleware/_internal/exceptions/middleware.py#L231)

If `disable_stack_trace` contains 404 or NotFoundException,
`exc_detail.isdisjoint(disable_stack_trace) is False`,
so exception_logging_handler is never called.

This effectively disables logging for these exceptions, not just their stack traces.

Is the current behavior (no log at all when status/exception is listed in disable_stack_trace) intentional, or is this a bug?

_No response_

### MCVE

```python
from litestar.exceptions import NotFoundException
from litestar.logging import StructLoggingConfig
from litestar.logging.config import LoggingConfig, default_logger_factory
from litestar.middleware.logging import LoggingMiddlewareConfig
from litestar.plugins.structlog import StructlogConfig, StructlogPlugin


@get("/404")
async def trigger_404() -> None:
    raise NotFoundException()


structlog_config = StructlogConfig(
    structlog_logging_config=StructLoggingConfig(
        log_exceptions="always",
        disable_stack_trace={404, NotFoundException},
        standard_lib_logging_config=LoggingConfig(
            # root/stdlib config here, e.g.
            root={"level": "INFO", "handlers": ["queue_listener"]},
        ),
        logger_factory=default_logger_factory(as_json=False),
    ),
    middleware_logging_config=LoggingMiddlewareConfig(),
)

app = Litestar(
    route_handlers=[trigger_404],
    plugins=[StructlogPlugin(config=structlog_config)],
)
```

skylarkoo7:
## Summary

`disable_stack_trace` suppresses exception logging entirely instead of only suppressing the stack trace. When an exception's status code or class matches an entry in `disable_stack_trace`, the `handle_exception_logging` method short-circuits completely — no log line is emitted at all, not even a one-line error message.

The fix separates the "should we log?" decision from the "should we include the stack trace?" decision:

- The handler is now always invoked when logging is enabled, regardless of `disable_stack_trace`
- When the exception matches `disable_stack_trace`, the handler receives an empty traceback list (`[]`) instead of the formatted exception
- When it doesn't match, the handler receives the full traceback as before

## Changes

- `litestar/middleware/_internal/exceptions/middleware.py` — split the `handle_exception_logging` conditional into a "should log" check and a separate "include stack trace" check; pass `[]` when stack trace is suppressed
- `litestar/logging/config.py` — default exception logging handlers (both standard-lib and structlog) now check `tb`: use `logger.exception()` when traceback is present, `logger.error()` when it is empty
- `tests/unit/test_logging/test_logging_config.py` — updated `test_disable_stack_trace` to assert the handler IS always called, and that `tb` is empty when stack trace is suppressed
- `tests/unit/test_logging/test_structlog_config.py` — same update for `test_structlog_disable_stack_trace`

## Test plan

- All 10 parametrized cases in both test files updated to verify: handler is always called, `tb` is non-empty when stack trace is expected, `tb` is `[]` when stack trace is suppressed
- Custom handlers that inspect the `tb` argument will see `[]` for suppressed exceptions and can decide their own behavior
- Backward compatible: no change in behavior when `disable_stack_trace` is empty (the default)

Fixes #4510

### DIFF
--- a/litestar/logging/config.py
+++ b/litestar/logging/config.py
@@ -120,20 +120,34 @@ def _default_exception_logging_handler_factory(is_struct_logger: bool) -> Except
     if is_struct_logger:
 
         def _default_exception_logging_handler(logger: Logger, scope: Scope, tb: list[str]) -> None:
-            logger.exception(
-                "Uncaught exception",
-                connection_type=scope["type"],
-                path=scope["path"],
-            )
+            if tb:
+                logger.exception(
+                    "Uncaught exception",
+                    connection_type=scope["type"],
+                    path=scope["path"],
+                )
+            else:
+                logger.error(
+                    "Uncaught exception",
+                    connection_type=scope["type"],
+                    path=scope["path"],
+                )
 
     else:
 
         def _default_exception_logging_handler(logger: Logger, scope: Scope, tb: list[str]) -> None:
-            logger.exception(
-                "Uncaught exception (connection_type=%s, path=%r):",
-                scope["type"],
-                scope["path"],
-            )
+            if tb:
+                logger.exception(
+                    "Uncaught exception (connection_type=%s, path=%r):",
+                    scope["type"],
+                    scope["path"],
+                )
+            else:
+                logger.error(
+                    "Uncaught exception (connection_type=%s, path=%r):",
+                    scope["type"],
+                    scope["path"],
+                )
 
     return _default_exception_logging_handler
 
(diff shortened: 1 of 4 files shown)


cofin:
#4593 is probably a better place for this update.

PR #4593, opened 2026-02-11
feat(v3)!: New logging experience

### LINKED ISSUE #2858: Better Logging Interface

### LINKED ISSUE #3455: Enhancement: allow finer tuning `LoggingConfig` for exception logging
### Summary

Setting `LoggingConfig.log_exceptions` to `"always"` is good to get more data about uncaught exceptions in a `debug=False` environment but can get pretty noisy, since we'll also log some common errors for schema validation, auth, etc when the developer might really only be interested in uncaught internal server errors.

It would be helpful if we could configure this according to HTTP status codes (or perhaps `Union[int, type[Exception]]` - similar to the keys of `ExceptionHandlersMap`?) and only log when raising for these status codes/types

### Basic Example

Something like
`LoggingConfig(exceptions={422, 500, MyCustomException})`

### Drawbacks and Impact

N/A

### Unresolved questions

_No response_

### LINKED ISSUE #3913: Bug: Unexpected logger behaviour. Unable to use build-in default logger in Litestar
### Description

Hi,

I am currently using Litestar for my new services. And after FastAPI I can't get the logger to work correctly.

First. During development we use std readable custom logs. And on the server we use json logs. I didn't find this fix in the documentation in sours, I guess it's not there. Additional mention of this [issue](https://github.com/litestar-org/litestar/issues/3827).

Second. After several tests with Litestar, I could not jam your default logger. Using a custom logger based on the default logger, where clear formatters, logging levels, handlers, and interceptors are defined, it doesn't want to work as I expect it to work in litestar as it does in other parts of the service(s).

For example, when doing local development during an exception I expect to get an expanded log of my pattern, but instead I get the default Litestar log.

P.S. I'm not sure if it's a bug or a feature, but in that case I'd suggest considering stopping your default logger initialization or accepting a pre-created and prepared users' logger, as is done in the [Faststream](https://faststream.airt.ai/latest/getting-started/logging/?h=logger#using-your-own-loggers) framework.

_No response_

### MCVE

```python
from fastapi import APIRouter, FastAPI
from whatever import configure_logging


configure_logging("dev", "DEBUG")  # !My custom logger dict config definition


@get("/")
async def failed_endpoint_litestar() -> dict[str, str]:
    _ = 1 / 0  # ! Nothing in stdout :'(

    return {"message": "Hello World"}


router = APIRouter()

@router.get("/")
async def failed_endpoint_fastapi() -> dict[str, str]:
    _ = 1 / 0  # ! I've got my excepted log in stdout here

    return {"message": "Hello World"}


litestar_app = Litestar(route_handlers=[failed_endpoint_litestar])

fastapi_app = FastAPI()
fastapi_app.include_router(router)
```

### Steps to reproduce

```bash
Litestar - `uv run python -m uvicorn src.report:litestar_app --host 0.0.0.0 --port 80 --reload`
FastAPI - `uv run python -m uvicorn src.report:fastapi_app --host 0.0.0.0 --port 80 --reload`
```

### Screenshots

```bash
Below
```

### Logs

```bash
Below
```

### Litestar Version

"litestar==2.13.0",
Litestar - api
Litestar - console
FastAPI - api
FastAPI - console

### Platform

- [X] Linux
- [X] Mac
- [ ] Windows

provinzkraut:
Still need to clean up some things to reduce unnecessary breaking changes, but feature wise, everything is there.

<hr>

- Remove global Litestar logging config; Litestar now only configures its own logging
- Simplify structlog integration
- Remove structlog plugin (not necessary anymore)
- Don't configure structlog globalls
- Remove picologging
- Simplify logging configuration
- Simplify logging middleware configuration
- Unify logging config / logging middleware config
- Improve testing integration (behave nicely with caplog / structlogs `capture_logs`)
- Better docs

<hr>

Closes #2858
Closes #3913
Closes #3455

### DIFF
--- a/tests/unit/test_middleware/test_exception_handler_middleware.py
+++ b/tests/unit/test_middleware/test_exception_handler_middleware.py
@@ -1,4 +1,3 @@
-import sys
 from collections.abc import Generator
 from inspect import getinnerframes
 from typing import TYPE_CHECKING, Any, Callable, Optional
@@ -242,7 +241,6 @@ def test_exception_handler_middleware_calls_app_level_after_exception_hook() -> 
         (False, None, True),
     ],
 )
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_exception_handler_default_logging(
     caplog: "LogCaptureFixture",
     is_debug: bool,
@@ -364,7 +362,6 @@ def test_pdb_on_exception(mocker: MockerFixture) -> None:
     mock_post_mortem.assert_called_once()
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_get_debug_from_scope(caplog: "LogCaptureFixture") -> None:
     @get("/test")
     def handler() -> None:
--- a/tests/unit/test_middleware/test_logging_middleware.py
+++ b/tests/unit/test_middleware/test_logging_middleware.py
@@ -1,4 +1,3 @@
-import sys
 from collections.abc import Generator
 from logging import INFO
 from typing import TYPE_CHECKING, Annotated, Any
@@ -49,7 +48,6 @@ def handler() -> HTTPRouteHandler:
     return handler_fn
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_middleware_regular_logger(caplog: "LogCaptureFixture", handler: HTTPRouteHandler) -> None:
     with (
         create_test_client(route_handlers=[handler], middleware=[LoggingMiddleware()]) as client,
@@ -126,7 +124,6 @@ def test_logging_middleware_struct_logger(handler: HTTPRouteHandler) -> None:
         }
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_middleware_exclude_pattern(caplog: "LogCaptureFixture", handler: HTTPRouteHandler) -> None:
     @get("/exclude")
     def handler2() -> None:
@@ -151,7 +148,6 @@ def test_logging_middleware_exclude_pattern(caplog: "LogCaptureFixture", handler
         assert len(caplog.messages) == 2
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_middleware_exclude_opt_key(caplog: "LogCaptureFixture", handler: HTTPRouteHandler) -> None:
     @get("/exclude", skip_logging=True)
     def handler2() -> None:
@@ -176,7 +172,6 @@ def test_logging_middleware_exclude_opt_key(caplog: "LogCaptureFixture", handler
 
 
 @pytest.mark.parametrize("include", [True, False])
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_middleware_compressed_response_body(
     include: bool, caplog: "LogCaptureFixture", handler: HTTPRouteHandler
 ) -> None:
@@ -235,7 +230,6 @@ async def test_logging_middleware_post_binary_file_without_structlog(monkeypatch
 
 
 @pytest.mark.parametrize("logger_name", ("litestar", "other"))
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_messages_are_not_doubled(logger_name: str, caplog: "LogCaptureFixture") -> None:
     # https://github.com/litestar-org/litestar/issues/896
 
@@ -258,7 +252,6 @@ def test_logging_messages_are_not_doubled(logger_name: str, caplog: "LogCaptureF
         assert len(caplog.messages) == 2
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 13), reason="Broken. Skip because of pending removal in v3")
 def test_logging_middleware_log_fields(caplog: "LogCaptureFixture", handler: HTTPRouteHandler) -> None:
     with (
         create_test_client(


docs/usage/logging.rst
```diff
@@ -62,111 +39,156 @@ or status codes that should not generate stack traces in logs:
...
+Litestar offers a built-in `structLog <https://www.structlog.org/en/stable/>`
+integration, which enables Litestar to use structlog. To set it up, use
+:class:`~litestar.logging.structlog.StructLoggingConfig` instead of
+:class:`~litestar.logging.config.LoggingConfig`:
 
 .. code-block:: python
 
-    import logging
+   from litestar import Litestar, Request, get
+   from litestar.plugins.structlog import StructLoggingConfig
```
Kumzy: This import no longer exists right? Its litestar.logging.structlog ?

docs/usage/logging.rst
```diff
@@ -62,111 +39,156 @@ or status codes that should not generate stack traces in logs:
...
+
+You can pass custom structlog configuration to
+:class:`~litestar.logging.structlog.StructLoggingConfig`, which will get passed on to
+the wrapped logger:
+
+.. code-block:: python
 
     from litestar import Litestar, Request, get
-    from litestar.logging.config import LoggingConfig
+    from litestar.plugins.structlog import StructLoggingConfig
```
Kumzy: Same here the import no longer exists right? Its litestar.logging.structlog ?

litestar/config/app.py
```diff
@@ -134,8 +134,8 @@ class AppConfig:
     """A list of callables returning async context managers, wrapping the lifespan of the ASGI application"""
     listeners: list[EventListener] = field(default_factory=list)
     """A list of :class:`EventListener <.events.listener.EventListener>`."""
-    logging_config: BaseLoggingConfig | None = field(default=None)
-    """An instance of :class:`BaseLoggingConfig <.logging.config.BaseLoggingConfig>` subclass."""
+    logging_config: LoggingConfig = field(default_factory=LoggingConfig)
```
Kumzy: We can no longer pass None to this to disable the logging config?

litestar/logging/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .config import BaseLoggingConfig, LoggingConfig, StructLoggingConfig
+from .config import LoggingConfig
 
-__all__ = ("BaseLoggingConfig", "LoggingConfig", "StructLoggingConfig")
+__all__ = ("LoggingConfig",)
```
Kumzy, line 3: Why is ``StructLoggingConfig`` removed from the export there?

litestar/app.py
```diff
@@ -503,12 +496,16 @@ def __init__(
...
-            self.get_logger = self.logging_config.configure()
-            self.logger = self.get_logger("litestar")
+        logger_shutdown = self.logging_config._configure_logger()
+        if logger_shutdown:
+            self.on_shutdown.append(logger_shutdown)
+        self.logger = self.logging_config.get_litestar_logger()
 
         self.asgi_handler = self._create_asgi_handler()
 
+    def get_litestar_logger(self, name: str | None = None) -> Logger:
```
Kumzy, line 502: I would like to know why rename/change ``get_logger`` to ``get_litestar_logger``?

litestar/app.py
```diff
@@ -570,20 +567,6 @@ def _get_default_plugins(plugins: list[PluginProtocol]) -> list[PluginProtocol]:
     def debug(self) -> bool:
         return self._debug
 
-    @debug.setter
```
Kumzy, line 573: Is it intended so we can not change the debug status while the app is running or it was never working before?

litestar/config/app.py
euri10: my app on this branch uses Litestar(logging_config=None) and I didnt have any errors

litestar/app.py
provinzkraut: It has never properly worked before

litestar/app.py
provinzkraut: Because there is also a `get_logger`, which is just the function for how to retrieve a logger from the logging implementation, and `get_litestar_logger`, which, specifically, gets a logger named `litestar`, or a child logger of that logger. The naming here is intended to make it very explicit that this shouldn't be used as a general purpose logging facility

litestar/config/app.py
provinzkraut: Mhm. I think that's just an oversight in the typing. I'll check on that

litestar/logging/__init__.py
provinzkraut: Just an oversight

litestar/app.py
provinzkraut: It's also very tricky to get right, because not all logging libraries work the same in regards of setting levels dynamically.

litestar/logging/structlog.py
```diff
@@ -0,0 +1,153 @@
...
+
+    _structlog_config: dict[str, Any] | None = None
+
+    @property
+    def should_log_as_json(self) -> bool:
+        if sys.stderr.isatty():
+            return not self.pretty_print_tty
+        return True
+
+    def _get_structlog_config(self) -> dict[str, None]:
```
Kumzy, line 110: Litestar mostly use ``dict[str, Any]`` for returns type. It's the only line in the code where the returns is ``dict[str, None]``. Is it intended?

provinzkraut: Nope. Any I'm not sure why mypy isn't complaining about this 😅

litestar/logging/__init__.py
provinzkraut: Actually, no. Since the `StructloggingConfig` now depends on `structlog`, this would crash with no structlog installed

litestar/config/app.py
cofin: I think there are some performance implications here.  IIRC, using `None` here can improve performance in some cases.

We should probably ensure the `None` remains an option here.

litestar/logging/config.py
```diff
@@ -1,535 +1,207 @@
...
+    This is a convenience function and functionally equivalent to passing
+    :class:`~litestar.middleware.logging.LoggingMiddleware` to the root Litestar app.
+    To configure request logging more precisely, set ``log_requests=False``, and pass
+    the middleware manually.
+    """
 
-    exception_logging_handler: ExceptionLoggingHandler | None
-    """Handler function for logging exceptions."""
-    disable_stack_trace: set[Union[int, type[Exception]]]  # noqa: UP007
+    disable_stack_trace: set[int | type[Exception]] = dataclasses.field(default_factory=set)
```
cofin, line 83: It would be nice to use this opportunity to set a few default exceptions to exclude from the stack trace. 

I tend to always add the following default set to every application to cleanup logging output: `{404, 401, 403, NotAuthorizedException, PermissionDeniedException}`

litestar/logging/config.py
```diff
@@ -1,535 +1,207 @@
...
-    root: dict[str, dict[str, Any] | list[Any] | str] = field(
-        default_factory=lambda: {
-            "handlers": ["queue_listener"],
-            "level": "INFO",
-        }
-    )
-    """This will be the configuration for the root logger.
 
-    Processing of the configuration will be as for any logger, except that the propagate setting will not be applicable.
+    always_propagate_on_pytest: bool = True
```
cofin, line 113: Nice addition!

Makefile
```diff
@@ -136,16 +135,16 @@ docs-clean: 										## Dump the existing built docs
 
 docs-serve:  								## Serve the docs locally
 	@echo "=> Serving documentation"
-	uv run --isolated --python 3.12 sphinx-autobuild docs docs/_build/ -j auto --watch litestar --watch docs --watch tests --watch CONTRIBUTING.rst --open-browser --port=0
+	uv run --isolated sphinx-autobuild docs docs/_build/ -j auto --watch litestar --watch docs --watch tests --watch CONTRIBUTING.rst --open-browser --port=0
```
cofin, line 138: I think we should consider leaving these pins in.

When i install the dev env, I typically use the lowest supported version for the library (3.10 in this case), but the latest versions of Sphinx have a minimum version of 3.12.  We have been running into similar problems with the Advanced Alchemy docs.

provinzkraut: I had a reason for unpinning this but I can't remember. I'll find out

provinzkraut:
Superseded by #4615

PR #4595, opened 2026-02-12
feat: Add after exception hook for otel config

### LINKED ISSUE #4594: Enhancement: expose after exception hook for OTEL config
### Summary

Allow setting after exception hooks for OTEL config. Would allow to change behavior when exception is raised for a span.

### Basic Example

Currently I do something like this without which the exceptions in internal spans are simply swallowed

```python
async def otel_after_exception(exc: Exception, _scope: Scope) -> None:
    span = get_current_span()
    span.record_exception(exc)
    span.set_status(Status(StatusCode.ERROR, str(exc)))


@dataclass
class OpenTelemetryConfig(BaseOpenTelemetryConfig):
    after_exception: list[Callable[[Exception, Scope], Awaitable[None]]] = field(
        default_factory=lambda: [otel_after_exception]
    )
```

```python
class OpenTelemetryPlugin(BaseOpenTelemetryPlugin):
    """Opentelemetry Plugin."""

    @override
        app_config = super().on_app_init(app_config)
        if isinstance(self.config, OpenTelemetryConfig):
            app_config.after_exception.extend(self.config.after_exception)
        return app_config
```

### Drawbacks and Impact

Extra parameter, should not affect default behavior

### Unresolved questions

Should we also include a default change in behavior as in my example?

vikigenius:
## Description

- Expose after exception hook for OTEL config

### DIFF
--- a/tests/unit/test_contrib/test_opentelemetry.py
+++ b/tests/unit/test_contrib/test_opentelemetry.py
@@ -17,6 +17,7 @@ from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanE
 from litestar import WebSocket, get, websocket
 from litestar.config.app import AppConfig
 from litestar.contrib.opentelemetry import OpenTelemetryConfig, OpenTelemetryPlugin
+from litestar.enums import ScopeType
 from litestar.exceptions import http_exceptions
 from litestar.status_codes import HTTP_200_OK
 from litestar.testing import create_test_client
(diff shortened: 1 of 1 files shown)


vikigenius:
@provinzkraut just added some tests to improve the patch coverage, but the test just got randomly cancelled

provinzkraut:
> @provinzkraut just added some tests to improve the patch coverage, but the test just got randomly cancelled

I re-ran the tests. We're currently experiencing issues with the test suite in GH Actions, as it will intermittently stall and run into a timeout. Haven't been able to track down the source of that yet.

PR #4597, opened 2026-02-13
fix: bound methods incorrectly handled in DI leading to spurious warn…

### LINKED ISSUE #4596: Bug: Async Generator Bound Methods Not Detected in Litestar 2.20+
### Description

After PR #4459 (released in Litestar 2.20), bound methods that are async generators are no longer correctly detected. Instead, they are treated as sync callables, triggering the `warn_implicit_sync_to_thread` warning.

### MCVE

```python
import warnings
from collections.abc import AsyncGenerator

warnings.simplefilter('always')

class Config:
    async def provide_data(self) -> AsyncGenerator[str, None]:
        """Async generator method."""
        yield "data"


config = Config()

# This should NOT trigger a warning since provide_data is an async generator
# But in Litestar 2.20+ it incorrectly triggers warn_implicit_sync_to_thread
provide = Provide(dependency=config.provide_data)

print(f"has_async_generator_dependency: {provide.has_async_generator_dependency}")
print(f"has_sync_callable: {provide.has_sync_callable}")
```

### Steps to reproduce

Run the MCVE on 2.19 or any commit before #4459 

Verify no warnings

Run the MCVE on 2.20 or after the PR got merged

There will be warnings

### Logs

```text
/home/void/Projects/Alexi/lexalr/bug.py:19: LitestarWarning: Use of a synchronous callable <bound method Config.provide_data of <__main__.Config object at 0x7f2fdcf0b050>> without setting sync_to_thread is discouraged since synchronous callables can block the main thread if they perform blocking operations. If the callable is guaranteed to be non-blocking, you can set sync_to_thread=False to skip this warning, or set the environmentvariable LITESTAR_WARN_IMPLICIT_SYNC_TO_THREAD=0 to disable warnings of this type entirely.
  provide = Provide(dependency=config.provide_data)
has_async_generator_dependency: False
has_sync_callable: True
```

### Litestar Version

2.20.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

vikigenius:
## Description

- #4459 introduced a regression causing spurious warnings when providing bound methods to di
- This PR has a separate check for methods so we can correctly identify bound asyncgens

closes #4596

### DIFF
--- a/litestar/di.py
+++ b/litestar/di.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from inspect import isasyncgenfunction, isclass, isfunction, isgeneratorfunction
+from inspect import isasyncgenfunction, isclass, isfunction, isgeneratorfunction, ismethod
 from typing import TYPE_CHECKING, Any
 
 from litestar._signature import SignatureModel
@@ -59,7 +59,8 @@ class Provide:
 
         is_class_dependency = isclass(dependency)
         is_function = isfunction(dependency)
-        is_callable_instance = not is_class_dependency and not is_function
+        is_method = ismethod(dependency)
+        is_callable_instance = not is_class_dependency and not is_function and not is_method
         if is_class_dependency or is_callable_instance:
             check_target = dependency.__call__  # type: ignore[operator]
         else:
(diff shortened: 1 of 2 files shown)


provinzkraut:
Thanks!

PR #4602, opened 2026-02-16
fix(openapi): fix specifications for responses without content

### LINKED ISSUE #4598: Bug: `Response[None]` results in a different OpenAPI response specificiation than `None`
### Description

Following up after a long time on my other two issues about `Response[None]` return type annotation (https://github.com/litestar-org/litestar/issues/2914, https://github.com/litestar-org/litestar/issues/3640), I have found another issue.

If you use `None` as the return type, then the OpenAPI response specification will not include `content`, which is the proper thing to do in case of responses with no content, which is usually what you want to achieve when using `None` as the return type.

But when you use `Response[None]` as the return type, then the OpenAPI response specification will include `content` with a JSON schema describing `null` value. Which is simply not true, as the actual behaviour is not to respond with `null`, but respond with no data.

I believe simply this piece of code didn't take using `Response[None]` into consideration:

https://github.com/litestar-org/litestar/blob/v2.20.0/litestar/_openapi/responses.py#L121-L122

_No response_

### MCVE

```python
import json

@get("/foo", status_code=204)
async def foo() -> None:
    return None


@get("/bar", status_code=204)
async def bar() -> Response[None]:
    return Response(None)


app = Litestar(
    [foo, bar],
    on_startup=[
        lambda app: print(json.dumps(app.openapi_schema.to_schema(), indent=2))
    ],
)
```

### Steps to reproduce

Run the above and observe the output (shortened here):

```json
{
  "openapi": "3.1.0",
  "paths": {
    "/foo": {
      "get": {
        "responses": {
          "204": {
            "description": "Request fulfilled, nothing follows"
          }
        }
      }
    },
    "/bar": {
      "get": {
        "responses": {
          "204": {
            "description": "Request fulfilled, nothing follows",
            "content": {
              "application/json": {
                "schema": {
                  "type": "null"
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.20.0

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

spietras:
## Description

OpenAPI response specifications for responses without content were different from the actual behaviour in some cases.

This change aligns OpenAPI response specifications with the actual behaviour to not include any content when:
- status code is below `200`
- status code is `204`
- status code is `304`
- method is `HEAD`

## Closes

Closes #4598

### DIFF
--- a/litestar/_openapi/responses.py
+++ b/litestar/_openapi/responses.py
@@ -27,7 +27,6 @@ from litestar.response import (
     Response as LitestarResponse,
 )
 from litestar.response.base import ASGIResponse
-from litestar.types.builtin_types import NoneType
 from litestar.typing import FieldDefinition
 from litestar.utils import get_enum_string_value, get_name
 
@@ -120,12 +119,12 @@ class ResponseFactory:
 
     def create_success_response(self) -> OpenAPIResponse:
         """Create the schema for a success response."""
-        if self.field_definition.is_subclass_of((NoneType, ASGIResponse)):
-            response = OpenAPIResponse(content=None, description=self.create_description())
-        elif self.field_definition.is_subclass_of(Redirect):
+        if self.field_definition.is_subclass_of(Redirect):
             response = self.create_redirect_response()
         elif self.field_definition.is_subclass_of(File):
             response = self.create_file_response()
+        elif self.field_definition.is_subclass_of(ASGIResponse) or not self.route_handler.returns_content:
+            response = self.create_empty_response()
         else:
             media_type = self.route_handler.media_type
 
@@ -162,6 +161,13 @@ class ResponseFactory:
             )
         self.set_success_response_headers(response)
         return response
+
+    def create_empty_response(self) -> OpenAPIResponse:
+        """Create the schema for a response with no content."""
+        return OpenAPIResponse(
+            content=None,
+            description=self.create_description(),
+        )
 
     def create_redirect_response(self) -> OpenAPIResponse:
         """Create the schema for a redirect response."""
--- a/litestar/handlers/http_handlers/base.py
+++ b/litestar/handlers/http_handlers/base.py
@@ -51,11 +51,10 @@ from litestar.types import (
     Send,
     TypeEncodersMap,
 )
-from litestar.types.builtin_types import NoneType
 from litestar.utils import deprecated as litestar_deprecated
 from litestar.utils import ensure_async_callable
 from litestar.utils.empty import value_or_default
-from litestar.utils.predicates import is_async_callable, is_class_and_subclass
+from litestar.utils.predicates import is_async_callable
 from litestar.utils.scope.state import ScopeState
 from litestar.utils.warnings import warn_implicit_sync_to_thread, warn_sync_to_thread_with_async_callable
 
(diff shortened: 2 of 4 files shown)


litestar/_openapi/responses.py
```diff
@@ -120,12 +119,12 @@ def create_description(self) -> str:
...
         """Create the schema for a success response."""
-        if self.field_definition.is_subclass_of((NoneType, ASGIResponse)):
-            response = OpenAPIResponse(content=None, description=self.create_description())
-        elif self.field_definition.is_subclass_of(Redirect):
+        if self.field_definition.is_subclass_of(Redirect):
             response = self.create_redirect_response()
         elif self.field_definition.is_subclass_of(File):
             response = self.create_file_response()
+        elif self.field_definition.is_subclass_of(ASGIResponse) or not self.route_handler.returns_content:
+            response = self.create_empty_response()
```
spietras, line 127: `Redirect`, `File` are checked first, because they also might fall into `returns_content` being `False`, but they need their own special handling.

Checking `ASGIResponse` is still there, because I guess the assumption is that you can't infer any meaningful content schema for it, so you fall back to specifying there is no content. And it works like that regardless of whether there will be actual content in the response or not.

Checking `NoneType` is removed, because `None` doesn't necessarily mean there is no content. It also depends on the status code and method. `GET` with `200` returning `None` will include `null` in the actual response, and now it will also include content in OpenAPI specification.

litestar/handlers/http_handlers/base.py
```diff
@@ -531,6 +530,15 @@ def resolve_request_max_body_size(self) -> int | None:
...
         return value_or_default(self._request_max_body_size, None)  # pyright: ignore
 
+    @property
+    def returns_content(self) -> bool:
+        """Whether the route handler returns any content in the response body."""
+        return not (
+            self.status_code < 200
+            or self.status_code in {HTTP_204_NO_CONTENT, HTTP_304_NOT_MODIFIED}
+            or self.http_methods == {HttpMethod.HEAD}
+        )
```
spietras, line 540: I don't know what the best place for this logic is. For now, I assumed that the handler is the owner of this information, because below it also validates the annotation based on the same logic.

litestar/handlers/http_handlers/base.py
```diff
@@ -591,23 +608,6 @@ def _validate_handler_function(self) -> None:
...
-            field_definition = self.parsed_fn_signature.return_type
-            if not (
-                is_empty_response_annotation(field_definition)
-                or is_class_and_subclass(field_definition.annotation, File)
-                or is_class_and_subclass(field_definition.annotation, ASGIFileResponse)
-            ):
-                raise ImproperlyConfiguredException(
-                    f"{self}: Handlers for 'HEAD' requests must not return a value. Either return 'None' or a response type without a body."
-                )
-
```
spietras: I don't know why this check was so complicated before. I simplified it, and I think it does the same as before? Correct me if I'm missing something.

I also moved it be closer to the other check with status codes, as both of them are about empty annotations.

tests/unit/test_openapi/test_config.py
```diff
@@ -55,24 +55,15 @@ def handler_2() -> None:
...
-        },
-        "/2": {
-            "get": {
-                "deprecated": False,
-                "operationId": "id_y",
-                "responses": {"200": {"description": "Request fulfilled, document follows", "headers": {}}},
-                "summary": "Handler2",
-            }
-        },
-    }
```
spietras: These handlers are annotated with `None`, but they actually return `null`. So now the OpenAPI specification correctly includes a schema for the response with `null`. So I would need to add this here.

But I think it's pointless to compare the whole thing anyway. This test only cares about customising operation IDs, so I restricted it to check only that.

And the test whether handlers like this include `null` content schema in response specifications is added in another place.

PR #4605, opened 2026-02-22
fix: OpenAPI schema missing minItems/maxItems for msgspec Meta on collections

### LINKED ISSUE #4514: Bug: OpenAPI schema missing `minItems` for list fields with `msgspec.Meta(min_length=N)`
## Description

When using `msgspec.Struct` with `Annotated[list[T], Meta(min_length=N)]`, Litestar's OpenAPI schema generation does not include the `minItems` constraint, even though:

1. Runtime validation works correctly (msgspec rejects arrays shorter than N)
2. msgspec's own `msgspec.json.schema()` correctly generates `minItems`

### Minimal Example

```python
from typing import Annotated
from uuid import UUID

from msgspec import Meta, Struct


class ArticleCreateSchema(Struct, frozen=True):
    title: Annotated[str, Meta(min_length=1, max_length=255)]
    category_ids: Annotated[list[UUID], Meta(min_length=1)]  # <-- This should have minItems=1


@post("/articles")
async def create_article(data: ArticleCreateSchema) -> dict:
    return {"status": "ok"}


app = Litestar([create_article])

if __name__ == "__main__":
    import json
    schema = app.openapi_schema.to_schema()
    article_schema = schema["components"]["schemas"]["ArticleCreateSchema"]
    print(json.dumps(article_schema, indent=2))
```

### Output (Actual)

```json
{
  "properties": {
    "title": {
      "type": "string",
      "maxLength": 255,
      "minLength": 1
    },
    "category_ids": {
      "items": {
        "type": "string",
        "format": "uuid"
      },
      "type": "array"
    }
  },
  "required": ["category_ids", "title"],
  "title": "ArticleCreateSchema",
  "type": "object"
}
```

### Expected Output

```json
{
  "properties": {
    "title": {
      "type": "string",
      "maxLength": 255,
      "minLength": 1
    },
    "category_ids": {
      "items": {
        "type": "string",
        "format": "uuid"
      },
      "type": "array",
      "minItems": 1
    }
  },
  "required": ["category_ids", "title"],
  "title": "ArticleCreateSchema",
  "type": "object"
}
```

## Proof that msgspec handles this correctly

```python
import msgspec
from msgspec import Meta, Struct
from typing import Annotated
from uuid import UUID

class TestSchema(Struct):
    category_ids: Annotated[list[UUID], Meta(min_length=1)]

schema = msgspec.json.schema(TestSchema)
print(schema)
# Output: {'$ref': '#/$defs/TestSchema', '$defs': {'TestSchema': {'title': 'TestSchema', 'type': 'object', 'properties': {'category_ids': {'type': 'array', 'items': {'type': 'string', 'format': 'uuid'}, 'minItems': 1}}, 'required': ['category_ids']}}}
```

Note: msgspec correctly outputs `'minItems': 1` in its schema.

## Environment

- Python: 3.11.8
- Litestar: 2.18.0
- msgspec: 0.20.0
- OS: Windows 11

## Analysis

Looking at the Litestar source code, there is logic in `litestar/typing.py` that should handle this:

```python
kwargs["min_items"] = meta.min_length
```

However, this conversion is not being triggered when processing msgspec Struct fields with `Meta(min_length=N)` on list types.

## Impact

This causes issues for frontend code generation tools (like `@hey-api/openapi-ts` with Zod plugin) that rely on the OpenAPI schema to generate validation rules. The generated Zod schemas are missing `.min(N)` constraints for array fields.

## Workaround

Currently, we're using a post-processing script to patch the generated OpenAPI schema:

```python
def fix_array_min_items(schema_dict: dict) -> None:
    schemas = schema_dict.get("components", {}).get("schemas", {})
    for schema_name, schema in schemas.items():
        for prop_name, prop in schema.get("properties", {}).items():
            if prop.get("type") == "array" and "minItems" not in prop:
                # Manually add minItems based on known constraints
                pass
```

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.18

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

worksbyfriday:
## Summary

Fixes #4514.

When using `Annotated[list[T], Meta(min_length=N)]` on a msgspec Struct field, the OpenAPI schema generation did not include `minItems`/`maxItems`, even though:
- Runtime validation worked correctly (msgspec rejects short arrays)
- `msgspec.json.schema()` correctly generated `minItems`

The constraint metadata was lost at two boundaries:

1. **`litestar/plugins/core/_msgspec.py`**: `kwarg_definition_from_field()` extracted `min_length`/`max_length` for string/bytes types (lines 65-77) but not for collection types (`ListType`, `SetType`, `FrozenSetType`, `VarTupleType`). Added collection type handling that maps `min_length` → `min_items` and `max_length` → `max_items`.

2. **`litestar/typing.py`**: `_annotated_types_extractor()` handled `annotated_types.MinLen`/`MaxLen` but not `msgspec.Meta` (which is a separate class, not an `annotated_types` subclass). Added a branch for `msgspec.Meta` that extracts all constraint attributes, correctly mapping length constraints to items constraints for sequence containers.

### Before
```json
"category_ids": {
  "items": {"type": "string", "format": "uuid"},
  "type": "array"
}
```

### After
```json
"category_ids": {
  "items": {"type": "string", "format": "uuid"},
  "type": "array",
  "minItems": 1
}
```

## Test plan

- Added `test_create_schema_from_msgspec_annotated_collection_type` verifying `minItems`/`maxItems` for list and set fields
- Verified reporter's exact reproduction case produces correct schema
- Verified existing `test_create_schema_from_msgspec_annotated_type` still passes (no regression on int/str/bytes fields)

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4605

### DIFF
--- a/tests/unit/test_openapi/test_schema.py
+++ b/tests/unit/test_openapi/test_schema.py
@@ -307,9 +307,9 @@ def test_create_schema_from_msgspec_annotated_type() -> None:
             msgspec.Meta(max_length=16, examples=["example"], description="description", title="title", pattern=r"\w+"),
         ]
         bytes_field: Annotated[bytes, msgspec.Meta(max_length=2, min_length=1)]
-        default_field: Annotated[str, msgspec.Meta(min_length=1)] = "a"
         list_field: Annotated[list[UUID], msgspec.Meta(min_length=1, max_length=10)]
         set_field: Annotated[set[int], msgspec.Meta(min_length=2)]
+        default_field: Annotated[str, msgspec.Meta(min_length=1)] = "a"
 
     schema = get_schema_for_field_definition(FieldDefinition.from_kwarg(name="Lookup", annotation=Lookup))
 


litestar/typing.py
```diff
@@ -100,6 +101,29 @@ def _annotated_types_extractor(meta: Any, is_sequence_container: bool) -> dict[s
             kwargs["pattern"] = "[[:ascii:]]"
         elif meta.func == str.isdigit:  # pragma: no cover  # coverage quirk: It expects a jump here for branch coverage
             kwargs["pattern"] = "[[:digit:]]"
+    elif isinstance(meta, msgspec.Meta):
```
provinzkraut: This should not be handled here, since it's an implementation specific thing. It should be handled in the plugin. You can check how it's handled for other msgspec constraints / Pydantic, within their respective plugins.

worksbyfriday: You're right. Removed the `msgspec.Meta` handling from `_annotated_types_extractor`. The plugin's `kwarg_definition_from_field` now handles collection types — `msgspec.inspect` already unwraps `Meta` into `ListType(min_length=...)`, so the plugin just needs to extract those fields.

tests/unit/test_openapi/test_schema.py
```diff
@@ -321,6 +321,22 @@ class Lookup(msgspec.Struct):
     }
 
 
+def test_create_schema_from_msgspec_annotated_collection_type() -> None:
```
provinzkraut: This should just be merged with the test case directly above

worksbyfriday:
Both review items addressed:
1. Moved msgspec.Meta handling to the plugin (commit bde4c46)
2. Merged the collection test into the existing test case (commit 777b4f6)

Ready for another look.

tests/unit/test_openapi/test_schema.py
worksbyfriday: Done — merged into the existing test.

PR #4606, opened 2026-02-22
fix(dto): apply __schema_name__ only to root transfer model

### LINKED ISSUE #4599: Bug: DTO __schema_name__ doesn't work with relationship model
### Description

Following: feat(dto-schema-name): add __schema_name__ to dto base #4131
@provinzkraut Here is an exemple of the issue with __schema_name__

Using __schema_name__ on a DTO with a model that has a child model (either dataclass like the below example or sqlalchemy model) create a wrong $ref on the OpenApi schema

_No response_

### MCVE

```python

import json

from litestar.dto import DataclassDTO, DTOConfig


@dataclass
class Child:
    name: str


@dataclass
class Person:
    name: str
    age: int
    email: str
    child: Child


class ReadDTO(DataclassDTO[Person]):
    __schema_name__ = "DTO_NAME"
    config = DTOConfig(exclude={"email"})


@get("/person/{name:str}", return_dto=ReadDTO, sync_to_thread=False)
def get_person(name: str) -> Person:
    # Your logic to retrieve the person goes here
    # For demonstration purposes, a placeholder Person instance is returned
    return Person(
        name=name, age=30, email=f"email_of_{name}@example.com", child=Child("test")
    )


app = Litestar(
    route_handlers=[get_person],
    on_startup=[
        lambda app: print(json.dumps(app.openapi_schema.to_schema(), indent=2))
    ],
)
```

### Steps to reproduce

See below copy-paste of the relevant part of the output of the schema.

Expected results (without __schema_name__):
You can see that GetPersonPersonResponseBody has a property child, that create a $ref to "#/components/schemas/GetPersonPersonChildResponseBody"

```
"components": {
    "schemas": {
      "GetPersonPersonChildResponseBody": {
        "properties": {
          "name": {
            "type": "string"
          }
        },
        "type": "object",
        "required": [
          "name"
        ],
        "title": "GetPersonPersonChildResponseBody"
      },
      "GetPersonPersonResponseBody": {
        "properties": {
          "name": {
            "type": "string"
          },
          "age": {
            "type": "integer"
          },
          "child": {
            "$ref": "#/components/schemas/GetPersonPersonChildResponseBody"
          }
        },
        "type": "object",
        "required": [
          "age",
          "child",
          "name"
        ],
        "title": "GetPersonPersonResponseBody"
      }
    }
}
```

Results with __schema_name__="DTO_NAME":
property child is created with a $ref to the parent object "#/components/schemas/DTO_NAME"
```
  "components": {
    "schemas": {
      "DTO_NAME": {
        "properties": {
          "name": {
            "type": "string"
          },
          "age": {
            "type": "integer"
          },
          "child": {
            "$ref": "#/components/schemas/DTO_NAME"
          }
        },
        "type": "object",
        "required": [
          "age",
          "child",
          "name"
        ],
        "title": "DTO_NAME"
      }
    }
  }
```

### Screenshots

Schema generated without __schema_name__:
Image

Schema generated with __schema_name__:
Image

### Logs

```text

```

### Litestar Version

2.20.0

### Platform

- [ ] Linux
- [x] Mac
- [x] Windows

navalprakhar:
## Description

When a DTO has `__schema_name__` set and the model has nested children, the child's `$ref` wrongly pointed to the parent's custom schema name instead of getting its own schema.

Added an `is_root` flag to `create_transfer_model_type` so the custom name only applies to the top-level model.

## Closes

Fixes #4599

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -211,6 +211,9 @@ class DTOBackend:
         # Only apply the custom __schema_name__ to the root transfer model.
         # Nested models get their own generated name to avoid child $refs
         # pointing to the parent's schema.
+        # model_name only equals model_type.__name__ for the root model.
+        # Nested models have a different name, so __schema_name__ won't
+        # override their generated schema name.
         if model_name == self.model_type.__name__ and self.dto_factory.__schema_name__:
             struct_name = self.dto_factory.__schema_name__
         else:


provinzkraut:
Mhm. This addition would make it a breaking change. Would be better to fix this in a backwards compatible way.

navalprakhar:
> Mhm. This addition would make it a breaking change. Would be better to fix this in a backwards compatible way.

Fixed: `__schema_name__` is now detected as root vs nested internally by comparing `model_name`, no signature change.

litestar/dto/_backend.py
```diff
@@ -208,7 +208,10 @@ def create_transfer_model_type(
         Returns:
             A ``BackendT`` class.
         """
-        struct_name = self.dto_factory.__schema_name__ or self._create_transfer_model_name(model_name)
+        if model_name == self.model_type.__name__ and self.dto_factory.__schema_name__:
```
provinzkraut, line 217: Clever. Can put a comment above this, explaining what it does and why, as it's not immediately obvious to a future reader?

navalprakhar: Done, added a comment above the check.

provinzkraut: Comment doesn't really say why that checks prevents this

navalprakhar: How about:

model_name only equals model_type.__name__ for the root model. Nested models have a different name, so __schema_name__ won't override their generated schema name.

provinzkraut: That + the existing comment sounds good

navalprakhar: Done, combined both comments.

PR #4615, opened 2026-02-27
feat(logging)!: Radically new logging experience

### LINKED ISSUE #2858: Better Logging Interface

### LINKED ISSUE #3455: Enhancement: allow finer tuning `LoggingConfig` for exception logging

### LINKED ISSUE #3913: Bug: Unexpected logger behaviour. Unable to use build-in default logger in Litestar

provinzkraut:
Building upon #4610 and the draft in #4593, this a more radical approach to restructure logging. 

#4610 made it so exception logging isn't necessary anymore from within Litestar, which made the only place where `LoggingConfig` was needed the request logging middleware, which then in turn only needed a small subset of the functionality.

This here change:

- Removes `litestar.logging`
- Removes `Litestar.logger`
- Removes `Request.logger`
- Alters `LoggingMiddleware` to accept a logger name, instance or factory
- Adds a flag to `LoggingMiddleware` to switch between plain and structured logging

The main idea is that, since we don't need to know anything about any logging internals anymore, we can push the whole logging configuration out into the user's application, where it is much better suited to deal with all eventualities. This greatly simplifies both Litestar's logging related code, as well as application's logging setups, since they can now easily configure logging as they please, and simply pass Litestar a handle.

Really, this is all that's needed to set up logging now, and I'd argue, this is how it always should have been :)

<hr>

Closes #2858
Closes #3913
Closes #3455

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4615

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -4,7 +4,39 @@
 =============
 
 .. changelog:: 3.0.0
-    :date: 2024-08-30
+    :date: 2364-01-27
+
+    .. change:: Remove logging config and related constructs
+        :type: feature
+        :breaking:
+        :pr: 4615
+        :issue: 2858, 3913, 3455
+
+        Remove Litestar's logging configuration, and related modules. In particular,
+        the following have been removed:
+
+        - ``Litestar.logging_config`` parameter
+        - ``Litestar.logger`` attribute
+        - ``Litestar.get_logger`` method
+        - ``AppConfig.logging_config`` attribute
+        - ``ASGIConnection.logger`` property
+        - ``litestar.logging`` module
+        - ``litestar.plugins.structlog`` module
+        - ``litestar.middleware.lgging.LoggingMiddlewareConfig`` class
+
+        Aside from debugging, there is only one place where Litestar logs, which is
+        :class:`~litestar.middleware.logging.LoggingMiddleware`, which has been changed
+        to accept a logger directly.
+        ``litestar.middleware.lgging.LoggingMiddlewareConfig`` has also been removed,
+        with ``LoggingMiddleware`` now accepting all configuration arguments directly.
+
+        Users that have previously relied on Litestar to configure logging, can simply
+        use the standard logging configuration provided by their logging library of
+        choice.
+
+        .. seealso::
+            :doc:`/usage/logging`
+
 
     .. change:: Remove default exception handler for all unhandled exceptions
         :type: feature


RoTorEx:
Nice job, thanks!

PR #4619, opened 2026-03-01
feat: auto-detect raised HTTPExceptions for OpenAPI docs

### LINKED ISSUE #2416: Enhancement: Automatically document exceptions
### Summary

Look for `litestar.HttpException` exceptions in the code (+ in guards & middlewares applied?) of route handlers and add them to `raises` (or somewhere else), similarly what https://github.com/Kludex/fastapi-responses does.

### Basic Example

```py
from litestar.exceptions import PermissionDeniedException

@get(sync_to_thread=False)
def test(force_perm_fail: bool = False) -> None:
    if force_perm_fail:
        raise PermissionDeniedException

app = Litestar([test])
```
image

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

Br1an67:
## Description

When a route handler raises `HTTPException` subclasses but does not explicitly declare them via the `raises` parameter, those error responses are currently missing from the generated OpenAPI schema.

This PR adds AST-based auto-detection of raised `HTTPException` subclasses in route handlers. The OpenAPI response generator now automatically scans handler source code for `raise` statements targeting `HTTPException` subclasses and includes the corresponding error responses in the schema.

**Behavior:**
- When `raises` is **not set** on a handler → automatically detect raised exceptions via AST inspection
- When `raises` is **explicitly set** → use only the declared exceptions (no auto-detection), preserving full developer control
- Detection gracefully falls back to an empty list if source inspection fails (e.g., dynamically generated handlers)

**Changes:**
- New module `litestar/_openapi/exception_detection.py` with `detect_exceptions_from_handler()` 
- Modified `litestar/_openapi/responses.py` to call auto-detection in `create_responses()`
- Added unit tests in `tests/unit/test_openapi/test_exception_detection.py`

## Closes

Closes #2416

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4619

### DIFF
--- a/litestar/_openapi/exception_detection.py
+++ b/litestar/_openapi/exception_detection.py
@@ -1,0 +1,67 @@
+"""Automatic detection of HTTPException subclasses raised in route handlers."""
+
+from __future__ import annotations
+
+import ast
+import inspect
+import textwrap
+from typing import TYPE_CHECKING, Any, Callable
+
+if TYPE_CHECKING:
+    from litestar.exceptions.http_exceptions import HTTPException
+
+__all__ = ("detect_exceptions_from_handler",)
+
+
+def detect_exceptions_from_handler(fn: Callable[..., Any]) -> list[type[HTTPException]]:
+    """Detect HTTPException subclasses raised in a handler function via AST inspection.
+
+    Scans the handler's source code for ``raise`` statements and resolves exception
+    class names against the handler's global namespace. Only direct subclasses of
+    :class:`HTTPException <litestar.exceptions.http_exceptions.HTTPException>` are returned.
+
+    Args:
+        fn: The route handler callable to inspect.
+
+    Returns:
+        A list of HTTPException subclass types found in ``raise`` statements.
+    """
+    from litestar.exceptions.http_exceptions import HTTPException
+
+    try:
+        source = inspect.getsource(fn)
+        source = textwrap.dedent(source)
+        tree = ast.parse(source)
+    except (OSError, TypeError, IndentationError, SyntaxError):
+        return []
+
+    exception_names: set[str] = set()
+
+    for node in ast.walk(tree):
+        if not isinstance(node, ast.Raise) or node.exc is None:
+            continue
+
+        exc_node = node.exc
+        # Handle: raise SomeException(...)
+        if isinstance(exc_node, ast.Call):
+            func = exc_node.func
+            if isinstance(func, ast.Name):
+                exception_names.add(func.id)
+            elif isinstance(func, ast.Attribute):
+                exception_names.add(func.attr)
+        # Handle: raise SomeException (no call)
+        elif isinstance(exc_node, ast.Name):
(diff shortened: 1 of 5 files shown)


litestar/_openapi/exception_detection.py
```diff
@@ -0,0 +1,67 @@
...
+    class names against the handler's global namespace. Only direct subclasses of
+    :class:`HTTPException <litestar.exceptions.http_exceptions.HTTPException>` are returned.
+
+    Args:
+        fn: The route handler callable to inspect.
+
+    Returns:
+        A list of HTTPException subclass types found in ``raise`` statements.
+    """
+    from litestar.exceptions.http_exceptions import HTTPException
```
provinzkraut: This import should not be inline

litestar/_openapi/exception_detection.py
```diff
@@ -0,0 +1,67 @@
...
+import textwrap
+from typing import TYPE_CHECKING, Any, Callable
+
+if TYPE_CHECKING:
+    from litestar.exceptions.http_exceptions import HTTPException
+
+__all__ = ("detect_exceptions_from_handler",)
+
+
+def detect_exceptions_from_handler(fn: Callable[..., Any]) -> list[type[HTTPException]]:
```
provinzkraut, line 31: What happens if this is called with a callable that's not a function, e.g. a class, bound / unbound method, classmethod?

provinzkraut (commented):
Thanks for this submission. I do have a few concerns:

**Performance impact**
Doing this for *every* handler can get very costly for bigger applications, as it could increase startup time drastically. One approach to mitigating this might be using `tokenize` instead of `ast`

**User expectations**
If we add a feature that does something automatically, users will expect to work everywhere. `HTTPException`s *can* be raised at every point during the request processing chain, for example in dependencies and guards. As a user, I would expect guards in particular to work, since they're part of the authorisation chain.

However, making this work everywhere is going to be quite tricky.

**Hidden magic**
In general, Litestar tries not to do too much magic, and be explicit about things. This design goes quite a bit against that. 

<hr>

Assuming the goal is "Ensure that error responses are properly documented", I think this might be better suited for a linter. We've talked about adding a `check` command to Litestar's CLI before, where something like this would be a natural fit.

Br1an67:
Thanks for the thoughtful review — all valid concerns.

I've pushed an update that addresses them:

1. **Performance** — Switched from `ast.parse` to `tokenize`, which is significantly cheaper (no full tree construction).
2. **Hidden magic** — The feature is now **opt-in** via `OpenAPIConfig(auto_detect_exceptions=True)`. Off by default, so no surprise behavior.
3. **Guards/dependencies** — Acknowledged limitation: only the handler body is scanned. This is documented in the config field's docstring.

If you'd still prefer this to live as a CLI `check` command rather than a config option, I'm happy to pivot. Let me know which direction you'd prefer.

PR #4625, opened 2026-03-10
feat(openapi): use root_path for default server URL

### LINKED ISSUE #2077: Enhancement: Add `root_path` value to OpenAPI servers if present
### Summary

Currently the OpenAPI config servers URLs default to `"/"` even if a `root_path` is present, even though for example debug errors display the `root_path`. (`ValueError on GET /api/` if a `ValueError` is raised in `test_handler` in the code below.)

It would be nice to have Litestar automatically add the `root_path` value --if present-- as the default instead.

### Basic Example

Code:
```py


@get(path="/", sync_to_thread=False)
def test_handler(request: Request) -> dict[str, str | list]:
    return {
        "request.scope['root_path']": request.scope["root_path"],
        "app.openapi_config.servers": request.app.openapi_config.servers,
    }


if __name__ == "__main__":
    import uvicorn

    uvicorn.run(app, port=5000, root_path="/api")
```
Current response (Litestar 2.0.0b4):
```json
{
	"request.scope['root_path']": "/api",
	"app.openapi_config.servers": [
		{
			"url": "/",
			"description": null,
			"variables": null
		}
	]
}
```
Proposed:
```json
{
	"request.scope['root_path']": "/api",
	"app.openapi_config.servers": [
		{
			"url": "/api",
			"description": null,
			"variables": null
		}
	]
}
```

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

ndpvt-web:
## Description

When no servers are explicitly configured in `OpenAPIConfig`, this change uses the ASGI `root_path` from the request scope as the default server URL. This makes the OpenAPI schema work correctly when the app is deployed behind a reverse proxy with a path prefix.

Previously, the default server URL was always `"/"` regardless of the `root_path` setting. Now, if `root_path` is present in the scope and no custom servers are configured, the schema endpoint dynamically sets the server URL to the `root_path` value.

This follows the approach suggested by @peterschutt in the issue discussion.

Closes #2077

## Changes

- Modified `litestar/_openapi/plugin.py` to check for `root_path` in the request scope when serving the OpenAPI schema
- If no explicit servers are configured and `root_path` is present, uses it as the default server URL

## AI Disclosure

This PR was created with AI assistance (Claude by Anthropic). The implementation follows the approach outlined by maintainers in the issue discussion.

Co-Authored-By: AI Assistant (Claude) <ai-assistant@contributor-bot.dev>

### DIFF
--- a/litestar/_openapi/plugin.py
+++ b/litestar/_openapi/plugin.py
@@ -158,7 +158,15 @@ class OpenAPIPlugin(InitPlugin, ReceiveRoutePlugin):
 
             @get(paths, media_type=plugin_.media_type, sync_to_thread=False, name=handler_name)
             def _handler(request: Request) -> bytes:
-                return plugin_.render(request, self.provide_openapi_schema())
+                schema = self.provide_openapi_schema()
+                # If servers haven't been explicitly set and root_path is present, use it
+                if (
+                    schema.get("servers") == [{"url": "/"}]
+                    and (root_path := request.scope.get("root_path"))
+                    and root_path
+                ):
+                    schema = {**schema, "servers": [{"url": root_path}]}
+                return plugin_.render(request, schema)
 
             return _handler
 


themavik (commented):
curl test

themavik:
Ignore the review titled curl test — accidental API ping.

themavik (commented):
OpenAPI `_handler` only swaps `servers` when the cached schema still equals the default `[{"url": "/"}]`; good guard, but anyone who intentionally sets that exact default behind a proxy will still get `root_path` injected.

mohsinsurani (commented):
Hi @ndpvt-web! Thanks for tackling this highly requested feature.

I noticed an architectural edge case here regarding how the schema updates are isolated. By checking `request.scope.get("root_path")` exclusively inside the HTTP route handler, this implementation successfully updates the schema during a live web request, but it misses the other primary access boundaries discussed earlier in the main issue tracker:

1. **CLI Generation:** If a user runs the CLI utility (e.g., `litestar --root-path=/api schema openapi`) to dump a static schema definition file, this route handler is never triggered, so the exported schema won't contain the correct server path.
2. **Direct App Properties:** Accessing `app.openapi_schema` directly on the application instance outside of a live request cycle will still fallback to the root path defaults.

To make this completely robust, we should probably intercept the configuration at the `OpenAPIConfig` initialization layer or explicitly check if a root path variable was passed down through the application's global configuration options rather than relying solely on the live ASGI HTTP request scope.

Happy to collaborate on an updated strategy or help add the test coverage for the CLI execution paths if you'd like!

PR #4630, opened 2026-03-14
perf(docs): set a timeout on async resource test issue HTTP calls

tejasae-afk:
Add a timeout to the HTTP call in docs/examples/testing/async_resource_test_issue.py to avoid indefinite hangs.

Reason: Network calls without a timeout can hang a worker indefinitely.

Validation: `/Users/tejasattarde/Desktop/gh-patchbot/.venv/bin/python -m py_compile docs/examples/testing/async_resource_test_issue.py`.

Context: py-http-timeout at docs/examples/testing/async_resource_test_issue.py:38.

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4630

### DIFF
--- a/docs/examples/testing/async_resource_test_issue.py
+++ b/docs/examples/testing/async_resource_test_issue.py
@@ -35,5 +35,5 @@ def app(http_test_client: httpx.AsyncClient) -> Litestar:
 
 def test_handler(app: Litestar) -> None:
     with TestClient(app) as client:
-        response = client.get("/")
+        response = client.get("/", timeout=10.0)
         assert response.json() == {"status": 200}


docs/examples/testing/async_resource_test_issue.py
```diff
@@ -35,5 +35,5 @@ def app(http_test_client: httpx.AsyncClient) -> Litestar:
 
 def test_handler(app: Litestar) -> None:
     with TestClient(app) as client:
-        response = client.get("/")
+        response = client.get("/", timeout=10.0)
```
sobolevn, line 38: my 2c: I don't think that test client calls really need a timeout, it is better managed by external tools like `pytest-timeout` in my opinion.

provinzkraut: Agree with @sobolevn here

tejas-ae: Agreed. Thanks

PR #4639, opened 2026-03-19
fix: isolate type decoders from shared signature annotation state

maksimzayats:
## Description

This removes the shared `_decoder` mutation from signature parsing.

Previously, `SignatureModel.create()` could attach a decoder directly to the annotated type, which made decoder behavior process-global. That meant one handler, app, or DI provider could accidentally affect another if they used the same annotation class.

Now, signature parsing no longer mutates user-defined or stdlib classes. Request parsing stays scoped to the current handler or provider through the existing `default_deserializer(type_decoders=...)` path.

```python
class UserId:
    def __init__(self, value: str) -> None:
        self.value = value


@get("/decoded", type_decoders=[(lambda t: t is UserId, lambda t, v: t(f"decoded:{v}"))])
def decoded(user_id: UserId) -> str:
    return user_id.value


@get("/plain")
def plain(user_id: UserId) -> str:
    return user_id.value
```

Old behavior:
```python
# Building /decoded mutated the shared UserId type:
UserId._decoder = lambda t, v: t(f"decoded:{v}")

# So /plain could start decoding too, even without its own decoder.
# /decoded?user_id=1 -> "decoded:1"
# /plain?user_id=1   -> "decoded:1"   # leaked
```

New behavior:
```python
# No mutation on UserId.
# Decoding stays scoped to the current handler.

# /decoded?user_id=1 -> "decoded:1"
# /plain?user_id=1   -> 400 validation error
```

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4639

### DIFF
--- a/litestar/_signature/model.py
+++ b/litestar/_signature/model.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import re
 from functools import partial
-from pathlib import Path, PurePath
 from typing import (
     TYPE_CHECKING,
     Annotated,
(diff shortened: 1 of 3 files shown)


sobolevn (commented):
I would categorize this as a tricky optimization instead of just a bug. Current behavior looks more correct to me. However, we also have to measure the perf impact of this change.

maksimzayats:
Hi @sobolevn ! Hmm, the old behavior looks quite strange to me tbh... and it looks like more a bug, but I'd love to hear other maintainers opinion on this

sobolevn:
Context: https://github.com/jcrist/msgspec/issues/497 and https://github.com/jcrist/msgspec/issues/483

provinzkraut:
Agree with @sobolevn. Yes, it's a bit iffy to store things globally like this, but it's an optimisation we deemed worth it, as the lookup process is quite expensive. If you want to remove this, I'd need comprehensive benchmarks to show there's no negative impact on performance

themavik (commented):
Dropping the `_decoder` setattr hack in `SignatureModel._create_annotation` removes annotation mutation; the new parsing tests cover cross-handler order, split `Litestar` apps, and `Provide` without touching `UserId`.

PR #4643, opened 2026-03-25
feat(jwt): add leeway support for token expiration verification

### LINKED ISSUE #4584: Enhancement: support `leeway` in `jwt`
### Summary

Leeway allows to give `exp` some extra slack to fix potential clock errors.
It is a standard jwt feature, docs: https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode

I implemented it in `django-modern-rest` that borrowed a lot of code / inspiration from `litestar`'s jwt implementation: https://github.com/wemake-services/django-modern-rest/pull/445

Exact places: 
- https://github.com/wemake-services/django-modern-rest/blob/840ec2b6a2706fd6b6f7abde4ca2bdebab744ded/django_modern_rest/security/jwt/token.py#L53-L54
- https://github.com/wemake-services/django-modern-rest/blob/840ec2b6a2706fd6b6f7abde4ca2bdebab744ded/django_modern_rest/security/jwt/token.py#L65

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

Krishnachaitanyakc:
## Summary

Adds support for the `leeway` parameter in JWT token decoding, as requested in #4584.

- Added a `leeway` field (type `float | timedelta`, default `0`) to `BaseJWTAuth`, `JWTAuth`, `JWTCookieAuth`, and `OAuth2PasswordBearerAuth`
- Threaded `leeway` through `JWTAuthenticationMiddleware` and `JWTCookieAuthenticationMiddleware` to `Token.decode` and `Token.decode_payload`
- `Token.decode_payload` passes `leeway` directly to PyJWT's `jwt.decode()`, which natively supports it
- Maintained backward compatibility: if a subclass overrides `decode_payload` without the `leeway` parameter, the call falls back to the old signature via a `TypeError` catch
- When `leeway` is set, `Token.decode` bypasses `__post_init__` validation (which rejects expired `exp` values) since PyJWT already validated expiry with leeway applied

Closes #4584

## Test plan

- [x] Added `test_decode_with_leeway_allows_recently_expired_token` -- verifies a token expired by 5 seconds is accepted with 10 seconds of leeway (parametrized over `int`, `float`, and `timedelta`)
- [x] Added `test_decode_with_leeway_does_not_allow_long_expired_token` -- verifies a token expired by 1 hour is still rejected with 10 seconds of leeway
- [x] All 164 existing JWT tests pass (including `test_custom_decode_payload` for backward compatibility)

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4643

### DIFF
--- a/litestar/security/jwt/auth.py
+++ b/litestar/security/jwt/auth.py
@@ -94,6 +94,11 @@ class BaseJWTAuth(Generic[UserType, TokenT], AbstractSecurityConfig[UserType, To
     not a list of values, and matches ``audience`` exactly. Requires that
     ``accepted_audiences`` is a sequence of length 1
     """
+    leeway: float | timedelta = 0
+    """A time margin in seconds (or as a :class:`~datetime.timedelta`) to account for
+    clock skew when verifying the ``exp`` (*expiration*) and ``nbf`` (*not before*)
+    claims. Defaults to ``0``.
+    """
 
     @property
     def openapi_components(self) -> Components:
@@ -153,6 +158,7 @@ class BaseJWTAuth(Generic[UserType, TokenT], AbstractSecurityConfig[UserType, To
             verify_expiry=self.verify_expiry,
             verify_not_before=self.verify_not_before,
             strict_audience=self.strict_audience,
+            leeway=self.leeway,
         )
 
     def login(
@@ -347,6 +353,11 @@ class JWTAuth(Generic[UserType, TokenT], BaseJWTAuth[UserType, TokenT]):
     not a list of values, and matches ``audience`` exactly. Requires that
     ``accepted_audiences`` is a sequence of length 1
     """
+    leeway: float | timedelta = 0
+    """A time margin in seconds (or as a :class:`~datetime.timedelta`) to account for
+    clock skew when verifying the ``exp`` (*expiration*) and ``nbf`` (*not before*)
+    claims. Defaults to ``0``.
+    """
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


sobolevn (commented):
I don't think that this implementation is correct, please take a look at https://github.com/wemake-services/django-modern-rest/blob/8033c9ed0d871ff413875350a4dc664bd6f174ba/dmr/security/jwt/token.py#L75

Krishnachaitanyakc:
@sobolevn Thanks for the pointer to the `InitVar` approach! I've reworked the implementation following your suggestion:

**What changed:**
- `leeway` is now an `InitVar[float]` on `Token`, so it participates in `__post_init__` validation but is not stored as an instance attribute
- `__post_init__` applies the leeway tolerance when checking `exp`, allowing recently-expired tokens within the leeway window to pass validation
- `decode()` converts `timedelta` leeway to float seconds and passes it directly to `cls(**payload, leeway=leeway_seconds)`
- Removed `msgspec.convert()` — the Token is now constructed via its own `__init__`, which naturally triggers `__post_init__` with the leeway
- Removed `_construct_without_validation()` — no longer needed since `__post_init__` itself handles leeway
- Removed the `TypeError` catch around `decode_payload` — that backward-compat path is no longer necessary
- Dropped the `msgspec` import from this module entirely

JacobCoffee (changes requested):
need to get CI all passing as well :)

PR #4645, opened 2026-03-25
chore!: Drop pydantic 1

provinzkraut:
Drop support for Pydantic 1. 

I've left in testing constructs to help test against multiple Pydantic version, as they may help us in the future, if another major release drops with this sort of backwards compatibility layer that we want to test against.

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4645


tests/unit/test_plugins/test_pydantic/test_integration.py
Kumzy: This needs to be removed too ?

Kumzy: There is still the ``pydantic_1_app.py`` in test_apps, it needs to be removed too?

litestar/plugins/pydantic/utils.py
```diff
@@ -239,7 +163,7 @@ def is_pydantic_root_model(annotation: Any) -> bool:
 class PydanticModelInfo:
     pydantic_version: Literal["1", "2"]
```
Kumzy, line 149: This needs to be adjusted to remove 1 ?

litestar/plugins/pydantic/__init__.py
```diff
@@ -13,7 +13,7 @@
     from pydantic.v1 import BaseModel as BaseModelV1
```
Kumzy: Remove?

litestar/plugins/pydantic/plugins/schema.py
Kumzy: To be removed ?

Kumzy: Same there

litestar/plugins/pydantic/plugins/schema.py
```diff
@@ -258,7 +257,7 @@ def is_undefined_sentinel(value: Any) -> bool:
 
     @staticmethod
     def is_constrained_field(field_definition: FieldDefinition) -> bool:
```
Kumzy: Is this method still needed as it returns always false?

tests/unit/test_plugins/test_pydantic/test_integration.py
provinzkraut: Removed

litestar/plugins/pydantic/plugins/schema.py
provinzkraut: Ah, base method also already return `False`. The override isn't needed!

Issue #4648, opened 2026-03-26
Enhancement: Support stdlib `compression.zstd` as a backend in `CompressionMiddleware`


provinzkraut:
### Summary

We currently only support the `zstandard` backend, but Python 3.14 has built-in zsdt support. The APIs differ slightly though, so it's not a drop-in replacement, and the `ZstdCompression` facade class needs to be adjusted

PR #4649, opened 2026-03-26
feat: Officially support Python 3.14

provinzkraut:
A bit late to the party, but better than never :)

Most changes are from an updated ruff version

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4649

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
         python-version: ["3.11", "3.12", "3.13", "3.14"]
     uses: ./.github/workflows/test.yml
     with:
-      coverage: ${{ (matrix.python-version == '3.14' || matrix.python-version == '3.11') }}
+      coverage: true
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -53,8 +53,7 @@ annotated-types = ["annotated-types"]
 attrs = ["attrs"]
 brotli = ["brotli"]
 zstd = [
-    # Only needed for <3.14, since >=3.14 ships with stdlib support
-    "zstandard>=0.24.0; python_version < '3.14'",
+    "zstandard>=0.24.0",
```
sobolevn, line 56: question: why do change this? maybe we should change the import instead?

provinzkraut: See #4648. The comment was wrong: `zstandard` is not a drop-in replacement for `compression.zstd`. The APIs differ slightly. This will need some adjustment that I think is out of the scope of this PR

PR #4659, opened 2026-04-01
fix: Optional[UploadFile] should be None when no file submitted in multipart form

### LINKED ISSUE #4204: Bug: Empty strings in multipart/form-data requests are converted to None

### LINKED ISSUE #4647: Bug: Optional[UploadFile] with multipart/form-data broken since 2.18.0
### Description

I've ran into a bug in my app where an optional file upload stopped working, it seems that bug was introduced with 2.18.0

_No response_

### MCVE

```python
import msgspec
from typing import Optional, Annotated
from litestar import Litestar, get, post, Response


# @dataclass also triggers the bug
class Foo(msgspec.Struct):
    file: Optional[UploadFile] = None

    
@get("/")
async def foo() -> Response:
    return Response("""\
<!DOCTYPE html>
<html>
    <form action="/" method="post" enctype="multipart/form-data">
    <input type="string" name="text"/><br/>
    <input type="file" name="file"/><br/>
    <input type="submit"/>
    </form>
</html>
""", media_type="text/html")

@post("/")
async def bar(data: Annotated[Foo, Body(media_type=RequestEncodingType.MULTI_PART)]) -> str:
    return f"ok, file is {data.file}"

app = Litestar(route_handlers=[foo, bar])
```

### Steps to reproduce

1. Install litestar >= 2.18.0
2. Run the program `litestar --app main:app run`
3. Visit the webpage and select a random file, then press submit
4. This works and results in `ok file is ...`
5. Visit the webpage again but don't select a file and submit
6. Get the error `{"status_code":400,"detail":"Validation failed for POST /","extra":[{"message":"Unsupported type: <class 'str'>","key":"file","source":"body"}]}`

Downgrading to 2.17.0 cause the program to work as expected, file is None.

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.18.0+ have the issue, 2.17.0 does not

### Platform

- [x] Linux
- [ ] Mac
- [x] Windows

infraAnchor:
## Description

- When a browser submits a multipart form without selecting a file, the file field
  arrives as an empty string (`""`). Since 2.18.0 (PR #4271 which fixed #4204),
  this empty string is preserved instead of being converted to `None`, causing a
  `400` validation error for `Optional[UploadFile]` fields.

- Fix: in `_extract_multipart` (`litestar/_kwargs/extractors.py`), inside the
  type-hints loop, detect when a field value is `""` and the annotated type is
  `Optional[UploadFile]` (or a subclass). Replace the empty string with `None`
  before it reaches the conversion layer — an empty file input means "no file
  selected", which matches the semantics of an `Optional` upload field.

- Regression test added in `tests/unit/test_kwargs/test_multipart_data.py`.

## Closes

Closes #4647

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4659


brakhane:
>  detect when a field value is "" and the annotated type is Optional[UploadFile] (or a subclass).

Wouldn't this result in a weird exception if the field is annotated as (required) UploadFile and no file is uploaded nonetheless? I think getting a None in those cases is better than an empty string for consistency reasons

PR #4660, opened 2026-04-01
feat(contrib): add background_interval plugin

jcpunk:
## Description

- Adds a plugin for setting up background scheduled functions


jcpunk:
My specific use case is for pruning data for GDPR.  My application can set lifecycles for automatic expiration as well as manual cleanups.

This type of background task scheduling is something I thought might be of wider value.

provinzkraut:
Thank you for the contribution, but this is somewhat outside the scope of Litestar itself. If you want to publish this, I'd recommend you do it as an independent project.

PR #4662, opened 2026-04-02
fix(opentelemetry): update tests for otel sdk 1.40.0 compatibility

giorgiovilardo:
## Description

Referring to [opentelemetry-python#5009](https://github.com/open-telemetry/opentelemetry-python/issues/5009), the issue is a bit nuanced: there is one change to make upstream - long story very short, there is an unsafe attribute access that is triggered by passing no boundaries to `ExplicitBucketHistogramAggregation` i.e. `aggregation_last_value = {Counter: ExplicitBucketHistogramAggregation()}` and then pass this as `preferred_aggregation` to the `InMemoryMetricReader`,  in `tests/unit/test_contrib/test_opentelemetry.py:33-35`.

This will be fixed upstream but also made more robust on Litestar side with the proposed change:

`aggregation_last_value = {Counter: ExplicitBucketHistogramAggregation(boundaries=[])}`

By fixing this, we would expose a subsequent problem on `test_open_telemetry_middleware_with_http_route`, more specifically when navigating `scope_metrics` via index with `scope_metrics = resource_metrics.scope_metrics[0]`; this is due to 1.40.0 `TracerProvider`s having a new `TracerMetrics` class that instruments them. This changes the ordering on the `scope_metrics`, making indexed access fail.

The other proposed change will just get the correct scope skipping the otel tagged one.

This will allow, once upstream is done, to unpin (or in general update) `opentelemetry-sdk`

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4662

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,7 +146,7 @@ dev = [
   "psutil>=5.9.8",
   "hypercorn>=0.16.0",
   "daphne>=4.0.0",
-  "opentelemetry-sdk<1.40.0",
+  "opentelemetry-sdk",
   "httpx-sse",
   "structlog",
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -2056,7 +2056,7 @@ dev = [
     { name = "hypercorn", specifier = ">=0.16.0" },
     { name = "hypothesis" },
     { name = "litestar", extras = ["full"] },
-    { name = "opentelemetry-sdk", specifier = "<1.40.0" },
+    { name = "opentelemetry-sdk" },
     { name = "psutil", specifier = ">=5.9.8" },
     { name = "psycopg", extras = ["pool"], marker = "python_full_version >= '3.13'", specifier = "<3.2.4" },
     { name = "psycopg", extras = ["pool", "binary"], marker = "python_full_version < '3.13'", specifier = ">=3.1.10,<3.2" },


giorgiovilardo:
@xrmx @provinzkraut

tests/unit/test_contrib/test_opentelemetry.py
```diff
@@ -31,7 +31,7 @@ def resource() -> Resource:
 
 @pytest.fixture(scope="session")
 def reader() -> InMemoryMetricReader:
-    aggregation_last_value = {Counter: ExplicitBucketHistogramAggregation()}
+    aggregation_last_value = {Counter: ExplicitBucketHistogramAggregation(boundaries=[])}
```
xrmx, line 34: Nice catch! The sdk code did not expect to be used with a different instrument 😅 , given the name of the variable maybe `LastValueAggregation` is what you are after?

giorgiovilardo:
hey @xrmx, yep already done that but not doing the unpin in this pr - leaving it to the more established contributors

relevant diffs / pip show

image

make test-all output

image


so yeah technically this is already able to receive otel 1.40.0 I guess, but still better to also fix upstream the unsafe access

provinzkraut:
Thanks for this @giorgiovilardo! I agree with @xrmx though, we should remove the version constraint in this same PR

giorgiovilardo:
@provinzkraut done ;)

PR #3742, opened 2024-09-15
fix(OpenAPI): Correctly handle `msgspec.Struct` tagged unions

provinzkraut:
Fix a bug where we would not include the struct fields implicitly generated by msgspec for its [tagged union](https://jcristharif.com/msgspec/structs.html#tagged-unions) support.

The change consists of an addition to the OpenAPI plugin where the tagged field will be added as a `const` to the schema.

Fix #3659.

PR #4671, opened 2026-04-04
feat(msgspec_dto): include msgspec tagged union field

abdulhaq-e:
When using `msgspec` tagged union, it is sometimes desirable to include the tag field in the DTO. This wasn't supported.

The current implementation unconditionally includes this field. I believe it's a good default option, however I think we should introduce a custom `MsgspecDTOConfig` (similar to what we had for `SQLAlchemy`) which can control this behaviour. 

This code was also generated using AI but this sometimes it's an area of the code base I'm more familiar with.

I've created it as draft to discuss having `MsgspecDTOConfig` and the flag and for any further edits.

### DIFF
--- a/litestar/dto/msgspec_dto.py
+++ b/litestar/dto/msgspec_dto.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 from dataclasses import replace
-from typing import TYPE_CHECKING, Generic, TypeVar
+from typing import TYPE_CHECKING, Generic, Literal, TypeVar
 
 import msgspec.inspect
 from msgspec import NODEFAULT, Struct, structs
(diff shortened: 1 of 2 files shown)


abdulhaq-e:
This is similar to https://github.com/litestar-org/litestar/pull/3742 which added the tag field to open api. Perhaps since it's always added to Open api and since this is the new v3, it might be okay to just introduce it unconditionally. One can then exclude it if needed using `exclude` in `DTOConfig`

litestar/dto/msgspec_dto.py
```diff
@@ -33,16 +32,36 @@ def _default_or_none(value: Any) -> Any:
     return None if value is NODEFAULT else value
 
 
+def _msgspec_attribute_accessor(obj: object, name: str) -> Any:
```
provinzkraut, line 35: I'm a bit concerned about the performance aspect of this, since this function will now be called for every attribute access, whereas without it, access would be done via direct access (`<obj>.<attribute name>`)

PR #4679, opened 2026-04-09
fix: do not use `time.time` for rate limiting

sobolevn:
Use `time.monotonic` instead.
Difference: https://docs.python.org/3/library/time.html#time.monotonic


sobolevn:
Ok, this change is not compatible with the current public API :/
And using time math is even worth.

PR #4685, opened 2026-04-11
fix(schema_generation/schema):  Fix OpenAPI schema type for Decimal to match serialization

### LINKED ISSUE #4682: Bug: inconsistent type of Decimal in JSON and OpenAPI schema
### Description

When using `msgpsec` and a `decimal.Decimal`, the default serialization format is supposed to be string. (This can be adjusted via a parameter in the JSON encoder https://jcristharif.com/msgspec/api.html#msgspec.json.Encoder but is not recommended because of precision losses.) This matches the behavior observed in the JSON return value of a route in Litestar, which correctly returns a string. However, the generated OpenAPI schema assigns `number` as the type of the `Decimal` the value.

If I understand the codebase correctly, this inconsistency is created during the OpenAPI schema generation. `Decimal`s are handled by the same function as `float` and `int`, which does not match the assumption of `msgspec`.

https://github.com/litestar-org/litestar/blob/32c595ba13873c3992e35c33686d63f5ccc49f7d/litestar/_openapi/schema_generation/schema.py#L514

### MCVE

```python
from decimal import Decimal

import msgspec

class HelloWorld(msgspec.Struct):
    value: Decimal

@get("/")
async def hello_world() -> HelloWorld:
    return HelloWorld(value=Decimal("1.001"))


```

### Steps to reproduce

1. Run the MCVE
2. Send `curl http://127.0.0.1:8000/` and receive `{"value":"1.001"}`
3. Open `http://127.0.0.1:8000/schema` in browser and see `value: number`

### Litestar Version

2.21.1

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

Peopl3s:
## Description

The OpenAPI schema generation now correctly assigns ``string`` type to ``decimal.Decimal``  fields instead of ``number``, matching the default string serialization used by msgspec.

Closes #4682 

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4685


docs/release-notes/changelog.rst
```diff
@@ -363,6 +363,15 @@
         A callable which takes in a :class:`~litestar.types.Scope` and returns a boolean
         to indicate whether to bypass the middleware for the current request.
 
+
+    .. change:: Fix OpenAPI schema type for Decimal to match serialization
```
provinzkraut: Why is this added in the middle of the changelog? You can just drop this, as it will be released as part of a 2.x release

provinzkraut (approved):

PR #4687, opened 2026-04-12
fix(openapi): correctly mark nullable required fields in schema

### LINKED ISSUE #4673: OpenAPI schema generation incorrectly marks nullable required fields as not required
## Summary

Fields declared as `field: int | None` (nullable, no default) are not included in the OpenAPI schema's `required` array. Litestar conflates "nullable" (can be `None`) with "optional" (has a default), causing the generated schema to diverge from both Pydantic's own `model_json_schema()` output and the OpenAPI 3.1.0 specification.

## Reproduction

```python


class TestSchema(BaseModel):
    required_non_nullable_field: int
    required_nullable_field: int | None
    optional_nullable_field: int | None = None


@get("/")
def get_test() -> TestSchema:
    return TestSchema(required_non_nullable_field=2, required_nullable_field=1, optional_nullable_field=3)


app = Litestar(route_handlers=[get_test])
```

**Expected** `required` array (matches Pydantic's `model_json_schema()`):
```json
"required": ["required_non_nullable_field", "required_nullable_field"]
```

**Actual** `required` array from Litestar's `/schema/openapi.json`:
```json
"required": ["required_non_nullable_field"]
```

## Root Cause

The bug exists in four locations that each independently conflate nullability with optionality:

1. **`litestar/typing.py:242`** — `FieldDefinition.is_required` returns `not self.is_optional and ...`, but `is_optional` (line 289) returns `True` whenever `NoneType` is in the type args — i.e., whenever the type is nullable, regardless of whether a default exists.

2. **`litestar/_openapi/schema_generation/plugins/dataclass.py:32`** — The required-field comprehension includes `not is_optional_union(type_hints[field.name])`, which excludes nullable fields even when they have no default.

3. **`litestar/_openapi/schema_generation/plugins/struct.py:48`** — Same `not is_optional_union(...)` pattern as the dataclass plugin.

4. **`litestar/plugins/pydantic/utils.py:251`** — Only wraps fields in `NotRequired` when `default_factory` is set, but the downstream `FieldDefinition.is_required` check (used at `plugins/pydantic/plugins/schema.py:175`) still fails due to issue #1.

## Affected Model Types

All model types are affected: Pydantic (v1/v2), dataclasses, and msgspec Structs.

## Versions

Confirmed on 2.16.0 and 2.21.0.

cofin:
## Summary

Fixes #4673 — Fields typed as `T | None` without a default were incorrectly excluded
from the OpenAPI schema's `required` array. Litestar conflated "nullable" (type contains
`None`) with "optional" (has a default value), causing the generated schema to diverge
from the OpenAPI 3.1.0 specification and Pydantic's own `model_json_schema()` output.

Affects all model types: dataclasses, msgspec Structs, attrs, and Pydantic.

## Fix

Four one-line changes removing the incorrect nullability checks from required-field computation:

- **`typing.py`** — `FieldDefinition.is_required` no longer uses `is_optional` (root cause)
- **`dataclass.py`**, **`struct.py`**, **`attrs.py`** — Removed redundant `is_optional_union()` 
  checks from their OpenAPI schema plugins; each already had correct default-checking logic

Runtime request validation behavior is unchanged — `_kwargs/parameter_definition.py` has
its own independent override, and nullable params without defaults were already enforced
as required at the validation layer.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4687">https://litestar-org.github.io/litestar-docs-preview/4687</a>


docs/release-notes/changelog.rst
Harshal6927, line 1: should revert this and add it on release

cofin: @provinzkraut is this the approach we are taking to the 3.0 change log? For some reason I thought we were doing it this way specifically for the 3.0 release.

provinzkraut: Changes only targeting 3.0 go into the 3.0 changelog. Changes that are also backported to 2.x do not need to be added there, we'll be using the old flow for that.

We could start adding release notes directly into the 2.x changelog, but in any case, they don't in the 3.0 changelog if they end up in the 2.x changelog.

cofin: Reverted.

tests/unit/test_openapi/test_nullable_required_fields.py
provinzkraut, line 1: This should be integrated with existing tests. It also breaks the pattern we usually follow to have non-native library tests in their own test suites; This has tests for `attrs`, `pydantic` and `msgspec`

tests/unit/test_openapi/test_nullable_required_fields.py
```diff
@@ -0,0 +1,237 @@
...
+    def handler(data: PydanticModel) -> PydanticModel:
+        return data
+
+    _assert_handler_schema_required(create_test_client(handler), "PydanticModel")
+
+
+# ─── FieldDefinition.is_required Tests ───────────────────────────────
+
+
+def test_field_definition_is_required_nullable_no_default() -> None:
```
provinzkraut: All these tests could be combined with `pytest.mark.parametrize`

tests/unit/test_plugins/test_attrs/test_schema_spec_generation.py
```diff
@@ -48,6 +48,6 @@ def handler(data: Person) -> Person:
                 },
             },
             "type": "object",
-            "required": ["complex", "first_name", "id", "last_name"],
+            "required": ["complex", "first_name", "id", "last_name", "optional", "pets"],
```
provinzkraut: Having the `optional` field be required doesn't make a lot of sense

tests/unit/test_typing.py
```diff
@@ -349,13 +349,15 @@ class Bar(msgspec.Struct):
         is True
     )
 
-    assert FieldDefinition.from_annotation(Optional[str]).is_required is False
+    # Nullable type without a default is required (nullability != optionality)
+    assert FieldDefinition.from_annotation(Optional[str]).is_required is True
```
provinzkraut: This should have an assertion that it's not required, but nullable

litestar/typing.py
```diff
@@ -241,7 +241,7 @@ def is_required(self) -> bool:
         if isinstance(self.kwarg_definition, ParameterKwarg) and self.kwarg_definition.required is not None:
             return self.kwarg_definition.required
 
-        return not self.is_optional and not self.is_any and (not self.has_default or self.default is None)
+        return not self.is_any and not self.has_default
```
provinzkraut: Why is `Any` never required?

provinzkraut (changes requested):

litestar/typing.py
cofin: It should not have been.  Removed it.

provinzkraut (commented):
Can you add a 3.0 changelog entry for this? Otherwise looking good!

cofin:
@provinzkraut changelog entry added back.  This should be ready to merge now.

provinzkraut (approved):

PR #4690, opened 2026-04-13
fix(request): content-length check incompatible with decompression middleware

### LINKED ISSUE #4125: Bug: Content Length Check on the Request Body Incompatible with Decompression Middleware
### Description

The `connection.request.Request` class has a `stream` method that is used to populate the request's body. It will raise a `ClientException("Malformed request")` when it detects `total_bytes_streamed > announced_content_length`.

This does not account for the situation where the ASGI receive function includes decompression middleware in the asgi app chain. When decompressed by middleware the body would naturally be larger than the content length of the network message. 

The comments in request.py suggest that the check is somewhat redundant because the ASGI server, such as uvicorn, should check this. As `total_bytes_streamed > announced_content_length` could be true before the body has finished streaming in and decompressing, would it be better to just remove the check if there is `Content-Encoding` header present that is set to anything other than `identity`? Alternatively, if the decompression middleware mutates the content-length header as it goes we could reload the `announced_content_length` for another check before raising the `ClientException`.

A version of [decompression middleware without gzip support](https://gist.github.com/enerqi/7062bc6aae17ba930700ae168f81ad94)

At this point I'm unclear if this is really a bug or if there's some other approach to avoid the issue.

_No response_

### MCVE

```python

```

### Steps to reproduce

Send a network request to a Litestar app with a compressed body and a content-encoding header that decompression middleware can use to choose how to decompress that body.

Without the decompression middleware the types won't match the API, it'll just be a byte stream and not e.g json.

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.15.2

### Platform

- [ ] Linux
- [ ] Mac
- [x] Windows

cofin:
## Summary

- Remove the content-length integrity check in `Request.stream()` that raised `ClientException("Malformed request")` when streamed bytes exceeded the `Content-Length` header
- Replace with unconditional `request_max_body_size` enforcement on all requests
- Update docs to reflect that content-length validation is delegated to the ASGI server

Closes #4125

## Detail

`Request.stream()` previously compared `total_bytes_streamed` against the `Content-Length` header and raised a 400 if the body exceeded it. This breaks when request decompression middleware wraps the ASGI `receive` function — the decompressed body is legitimately larger than the wire-format `Content-Length`.

The fix removes this check entirely. This is safe because:

1. **ASGI servers already enforce this** — uvicorn, granian, and hypercorn validate content-length at the transport layer ([uvicorn docs](https://github.com/encode/uvicorn/blob/fe3910083e3990695bc19c2ef671dd447262ae18/docs/server-behavior.md?plain=1#L11))
2. **`request_max_body_size` still enforces body limits** — the 413 response for oversized bodies is unchanged
3. **The old code had a subtle bug** — the `if/elif` structure meant that when `Content-Length` was present, `request_max_body_size` enforcement was unreachable (only the content-length check ran)

### Behavioral change

| Scenario | Before | After |
|----------|--------|-------|
| Body > Content-Length, within `request_max_body_size` | 400 Bad Request | **Request succeeds** |
| Body > `request_max_body_size` | 400 Bad Request (if Content-Length set) | **413 Request Entity Too Large** |
| Body > `request_max_body_size`, no Content-Length | 413 | 413 (unchanged) |

This is intentional — the previous behavior was incorrect when decompression middleware was in use, and the content-length check was redundant with ASGI server enforcement.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4690">https://litestar-org.github.io/litestar-docs-preview/4690</a>

### DIFF
--- a/tests/unit/test_connection/test_request.py
+++ b/tests/unit/test_connection/test_request.py
@@ -538,15 +538,34 @@ def test_state() -> None:
         assert response.json() == {"state": 2}
 
 
-def test_request_body_exceeds_content_length() -> None:
-    @post("/")
-    def handler(body: bytes) -> None:
-        pass
+def test_request_body_exceeding_content_length_is_allowed() -> None:
+    """Body larger than Content-Length but within request_max_body_size succeeds.
+
+    Decompression middleware can legitimately produce a body larger than the
+    wire-format Content-Length. Content-length enforcement is delegated to the
+    ASGI server (e.g. uvicorn).
+    """
+
+    @post("/", request_max_body_size=10)
+    def handler(body: bytes) -> bytes:
+        return body
 
     with create_test_client([handler]) as client:
         response = client.post("/", headers={"content-length": "1"}, content=b"ab")
-        assert response.status_code == HTTP_400_BAD_REQUEST
-        assert response.json() == {"status_code": 400, "detail": "Malformed request"}
+        assert response.status_code == 201
+        assert response.content == b"ab"
+
+
+def test_request_body_exceeding_content_length_still_enforces_max_body_size() -> None:
+    """Content-length mismatch is allowed, but request_max_body_size is still enforced."""
+
+    @post("/", request_max_body_size=1)
+    def handler(body: bytes) -> bytes:
+        return body
+
+    with create_test_client([handler]) as client:
+        response = client.post("/", headers={"content-length": "1"}, content=b"ab")
+        assert response.status_code == HTTP_413_REQUEST_ENTITY_TOO_LARGE
 
 
 def test_request_body_exceeds_max_request_body_size() -> None:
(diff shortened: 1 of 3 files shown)


tests/unit/test_connection/test_request.py
```diff
@@ -538,15 +538,21 @@ async def get_state(request: Request[Any, Any, State]) -> dict[str, str]:
         assert response.json() == {"state": 2}
 
 
-def test_request_body_exceeds_content_length() -> None:
+def test_request_body_exceeding_content_length_is_allowed() -> None:
```
provinzkraut, line 541: Can you explicitly set `request_max_body_size` here, and maybe add a failing variant, so it's a bit clearer what's being tested here?

provinzkraut (commented):
I'm still iffy about this, since [my proposal](https://github.com/django/asgiref/pull/510) to make this check on ASGI servers mandatory instead of an implementation detail was rejected, but I think it's an okay tradeoff here.

tests/unit/test_connection/test_request.py
cofin: yep - updated.

provinzkraut (approved):

PR #4691, opened 2026-04-13
feat(otel): move away from contrib module

### LINKED ISSUE #4466: Bug: OpenTelemetryHookHandler has the wrong type
### Description

The opentelemetry [docs](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/asgi/asgi.html#opentelemetry.instrumentation.asgi.OpenTelemetryMiddleware)

indicate that the callable has 3 arguments, Span, Scope and Event

The OpenTelemetryHook handler just indicates 2 arguments

https://github.com/litestar-org/litestar/blob/9bcae1e42dabce7a7d4308cebec0615d4c1c93c4/litestar/contrib/opentelemetry/config.py#L29

### Litestar Version

master

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

### LINKED ISSUE #4468: Bug: KeyError: 'route_handler' in all routes when setting exclude_opt_key in OpenTelemetryConfig
### Description

When setting `exclude_opt_key` in OpenTelemetryConfig all the routes fail with `KeyError: route_handler`

### MCVE

```python
from litestar.contrib.opentelemetry import OpenTelemetryConfig, OpenTelemetryPlugin

# --- Toggle this flag to compare behavior ---
USE_EXCLUDE_OPT_KEY = True

otel_config = OpenTelemetryConfig(
    # The single line below triggers the KeyError in routing middleware:
    exclude_opt_key="skip_otel" if USE_EXCLUDE_OPT_KEY else None,
)

@get("/")
async def root() -> dict[str, str]:
    return {"ok": "root"}

# Mark this route with the opt that should be excluded from OTEL
@get("/skip", opt={"skip_otel": True})
async def skip() -> dict[str, bool]:
    return {"skipped": True}

# A normal route (no opt)
@get("/health")
async def health() -> dict[str, str]:
    return {"status": "ok"}

app = Litestar(
    route_handlers=[root, skip, health],
    plugins=[OpenTelemetryPlugin(config=otel_config)],
)
```

### Steps to reproduce

1. Put the mcve in app.py
2. `uv add 'litestar[opentelemetry]'`
3. `uv run uvicorn app:app --reload`

Just try to open the app

### Logs

```text
ERROR:    Exception in ASGI application
  File "/home/void/Projects/Alexi/litestarbug/.venv/lib/python3.13/site-packages/uvicorn/protocols/http/httptools_impl.py", line 409, in run_asgi
        self.scope, self.receive, self.send
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/void/Projects/Alexi/litestarbug/.venv/lib/python3.13/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
  File "/home/void/Projects/Alexi/litestarbug/.venv/lib/python3.13/site-packages/litestar/app.py", line 624, in __call__
  File "/home/void/Projects/Alexi/litestarbug/.venv/lib/python3.13/site-packages/litestar/middleware/base.py", line 139, in wrapped_call
    if should_bypass_middleware(
       ~~~~~~~~~~~~~~~~~~~~~~~~^
        scope=scope,
        ^^^^^^^^^^^^
    ...<2 lines>...
        exclude_opt_key=self.exclude_opt_key,
    ):
    ^
  File "/home/void/Projects/Alexi/litestarbug/.venv/lib/python3.13/site-packages/litestar/middleware/_utils.py", line 70, in should_bypass_middleware
    if exclude_opt_key and scope["route_handler"].opt.get(exclude_opt_key):
                           ~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'route_handler'
```

### Litestar Version

master

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

cofin:
Moves the OpenTelemetry plugin from `litestar.contrib.opentelemetry` to its permanent home at `litestar.plugins.opentelemetry`. The `contrib` directory is removed entirely on this branch — there's no shim here.

A few things ride along:

- **#4468** — guards `scope["route_handler"]` access in `litestar/middleware/_utils.py` so the OTel middleware doesn't `KeyError` when it runs as the outermost ASGI layer (which it does in practice, since it sits before route resolution).
- **#4466** — corrects the type signatures on the client-request and client-response hooks to match `opentelemetry-instrumentation-asgi` (3 parameters, not 2), so user-supplied hooks actually type-check.
- **Upstream parity** — exposes `tracer` and `http_capture_headers_*` config params that were missing from `OpenTelemetryConfig` but supported by the underlying instrumentation library.

The deprecation shim at `litestar.contrib.opentelemetry` (with its `DeprecationWarning`) has been split out to a follow-up PR targeting `v2` — that's where 2.x users actually need the migration window before 3.0 GA. This PR is purely the move plus the bug fixes and parity work.

Fixes #4468
Fixes #4466

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4691">https://litestar-org.github.io/litestar-docs-preview/4691</a>


litestar/middleware/_utils.py
```diff
@@ -81,7 +81,11 @@ def should_bypass_middleware(
     if scope["type"] not in scopes:
         return True
 
-    if exclude_opt_key and scope["route_handler"].opt.get(exclude_opt_key):
+    if (
+        exclude_opt_key
+        and (route_handler := scope.get("route_handler")) is not None
```
JacobCoffee, line 86: why the addition of this None check? will routing not be resolved sometimes when this is caled?

litestar/plugins/opentelemetry/_utils.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
JacobCoffee, line 1: did this break and warrant swapping to stringized types or something else?

JacobCoffee (commented):
lgtm , @provinzkraut ?

litestar/plugins/opentelemetry/_utils.py
cofin: Nothing broke.  It's purely a preference (and a rule I've got set locally for libraries). I think, in general, we need to use stringized types to avoid the issues we have with them.  End users can use it when they use litestar in their apps, but I try to avoid using it in the actual codebase.  This has prevented issues with things like `from datetime import datetime` and `import datetime` type resolution issues.

litestar/middleware/_utils.py
cofin: This fixed a pre-existing issue that I inadvertently found.  OpenTelemetry middleware runs as a global ASGI middleware, wrapping the entire app to capture the full request span lifecycle. 

That means it executes before the router resolves the handler, so `scope["route_handler"]` simply isn't there yet. Without the None check, calling `should_bypass_middleware` from the otel layer would raise `KeyError`.

provinzkraut:
This seems to be targeting the wrong branch. It's deprecating things to be removed in `3.0`, but it's also targeting the `3.0` branch. Should be targeting the `v2` branch, and then a follow-up PR made into `main` that does the removal

cofin:
@provinzkraut deprecation notices removed and will be bundled in a separate PR.

provinzkraut:
@cofin if this is now intentionally targeting `main` (i.e. `v3.0`), it should include a changelog entry

cofin:
@provinzkraut merge conflicts resolved and changelog entry added.

provinzkraut (approved):

PR #4696, opened 2026-04-15
docs(usage/excluding-and-including-endpoints): fix misleading regex examples in exclude docs, use anchored patterns for prefix matching

### LINKED ISSUE #4695: Docs: Misleading regex example in security middleware docs
### Summary

This example given in the docs for excluding routes from the security middleware uses misleading regex examples: https://github.com/litestar-org/litestar/blob/f416d1ecf2d953f2ab6ff6992b516af84b450e6b/docs/usage/security/excluding-and-including-endpoints.rst?plain=1#L16-L17

It *says* that the patterns will be prefixed-matched, but they are not (which is intentional). It should instead use correct example patterns to prefix-match, e.g. `^/login` instead of `/login`, and explicitly mention this important detail (as we do in other parts of the documentation)

Peopl3s:
## Description

Fix misleading regex examples in exclude docs, use anchored patterns for prefix matching

Closes #4695 

<hr>
📚 Documentation preview 📚: https://litestar-org.github.io/litestar-docs-preview/4696

### DIFF
--- a/docs/usage/security/excluding-and-including-endpoints.rst
+++ b/docs/usage/security/excluding-and-including-endpoints.rst
@@ -13,9 +13,11 @@ Excluding routes
 --------------------
 
 The ``exclude`` argument takes a :class:`string <str>` or :class:`list` of :class:`strings <str>` that are interpreted
-as regex patterns. For example, the configuration below would apply authentication to all endpoints except those where
-the route starts with ``/login``, ``/signup``, or ``/schema``. Thus, one does not have to exclude ``/schema/swagger``
-as well - it is included in the ``/schema`` pattern.
+as regex patterns matched against the full path. Because the patterns are not implicitly anchored, a pattern like
+``/schema`` would match *any* path containing ``/schema``, not just paths that start with it. To match only paths that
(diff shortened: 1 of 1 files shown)


docs/usage/security/excluding-and-including-endpoints.rst
```diff
@@ -31,7 +33,7 @@ as well - it is included in the ``/schema`` pattern.
     session_backend_config=ServerSideSessionConfig(),
     # exclude any URLs that should not have authentication.
     # We exclude the documentation URLs, signup and login.
-    exclude=["/login", "/signup", "/schema"],
+    exclude=[r"^/login", r"^/signup", r"^/schema"],
```
sobolevn: While we are at it, can you please fix the formatting of the whole block? `retrieve_user_handler=retrieve_user_handler,` and other need 4 spaces

sobolevn (approved):
Otherwise patterns built here: https://github.com/litestar-org/litestar/blob/f416d1ecf2d953f2ab6ff6992b516af84b450e6b/litestar/middleware/_utils.py#L39 can indeed do unexpected things. Great find!

PR #4708, opened 2026-04-18
feat: Replace zstandard with backports.zstd for better compatibility with Python3.14+

### LINKED ISSUE #4648: Enhancement: Support stdlib `compression.zstd` as a backend in `CompressionMiddleware`

PokkaKiyo:
Replaces the zstandard depedency with backports.zstd instead.
This is similar to the strategy aiohttp uses for zstandard support.

If this PR is alright, I think we can also improve the way the API of the CompressionConfig in a separate PR.

Note:
The zstandard dependency cannot be fully removed because the tests require it (because httpx), I have therefore moved it to become a test dependency.

Closes #4648

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4708">https://litestar-org.github.io/litestar-docs-preview/4708</a>


sobolevn (approved):
I like that we can now swap the dependency easily. 

Question: there should be no user-facing changes, right?

PokkaKiyo:
> Question: there should be no user-facing changes, right?

There isn't any user-facing changes in this PR, because user interact via the CompressionConfig and the API hasn't changed there (the change to the default value shouldn't matter in this case).

That being said, the zstd library provides many more configuration knobs than what we expose via the CompressionConfig API, and I think we can certainly expose more of these knobs in a separate PR.

PR #4736, opened 2026-04-26
fix: robust PydanticDTO validation for nested collections and datetimes

zzzzkkkk7866:
This PR provides a consolidated and robust fix for PydanticDTO validation issues, specifically addressing #4530 (nested collections) and #4481 (datetime formats).

### Key Fixes:
1. **Pydantic v2 Compatibility**: Bypasses the problematic automatic model instantiation in DTO backends by directly parsing to `msgspec.Struct` and then using `msgspec.to_builtins()` before Pydantic's `model_validate()`. This ensures that Pydantic receives compatible native Python types (dicts/lists), resolving the `model_type` errors in nested structures (#4530).
2. **Enhanced DTOBackend**: Adds support for `CollectionType` within nested unions in `_transfer_nested_union_type_data`, preventing `RuntimeError` when handling structures like `list[Model] | None`.
3. **Flexible Datetime Validation**: Maps `datetime`, `date`, and `time` to `str` in `PydanticDTO`'s `_down_types`. This allows the underlying Pydantic model to use its flexible ISO8601 parsing instead of `msgspec`'s strict RFC3339 enforcement (#4481).

### Validation:
- Surgically tested on `v2.15.1` (local env) with regression tests for both #4530 and #4481.
- Syntax-verified for Python 3.11+ compatibility on the `main` branch.
- Clean implementation with no duplicate imports or unrelated changes.

This approach maintains the high performance of `msgspec` for initial parsing while ensuring 100% compatibility with Pydantic's validation logic.

### DIFF
--- a/litestar/plugins/pydantic/dto.py
+++ b/litestar/plugins/pydantic/dto.py
@@ -2,6 +2,9 @@
 
 from __future__ import annotations
 
+import datetime
+import msgspec
+
 import dataclasses
 from dataclasses import replace
 from typing import TYPE_CHECKING, Annotated, Any, Generic, TypeAlias, TypeVar
@@ -40,6 +43,9 @@ T = TypeVar("T", bound="ModelType | Collection[ModelType]")
 __all__ = ("PydanticDTO",)
 
 _down_types: dict[Any, Any] = {
+    datetime.datetime: str,
+    datetime.date: str,
+    datetime.time: str,
     pydantic.EmailStr: str,
     pydantic.IPvAnyAddress: str,
     pydantic.IPvAnyInterface: str,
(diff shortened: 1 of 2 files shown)


litestar/plugins/pydantic/dto.py
```diff
@@ -71,14 +77,18 @@ class PydanticDTO(AbstractDTO[T], Generic[T]):
     @override
     def decode_builtins(self, value: dict[str, Any]) -> Any:
         try:
-            return super().decode_builtins(value)
+            backend = self._dto_backends[self.asgi_connection.route_handler.handler_id]["data_backend"]
+            data = backend.parse_builtins(value, self.asgi_connection)
+            return self.model_type.model_validate(msgspec.to_builtins(data))
```
provinzkraut, line 82: This is a huge performance regression, as data will be passed through msgspec twice now

provinzkraut (changes requested):

provinzkraut:
Closing as the approach is not viable. See my comment above.

PR #4737, opened 2026-04-26
chore: harden github parameters for docs

Harshal6927:
## Description

- Update workflow permissions for preview docs

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4737">https://litestar-org.github.io/litestar-docs-preview/4737</a>


.github/workflows/docs-preview.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
           owner: litestar-org
           repositories: litestar-docs-preview
+          permission-contents: write
```
sobolevn, line 47: Why is this needed?

Harshal6927: `actions/create-github-app-token` tokens [inherit installation permissions by default](https://github.com/actions/create-github-app-token#permission-permission-name). If the app has more than contents write on that preview repo, the deploy step gets more than it needs.

provinzkraut (approved):

PR #4750, opened 2026-05-06
feat(core): Introduce explicit parameter declarations with generics

provinzkraut:
Replace implicit parameter declarations with explicit ones.

## Motivation

There are two main points:

- Providing clarity in handler and DI function definitions where parameters come from
- Avoiding implicit, accidental definition of parameters

Just looking at the definition of `handler`, it's impossible to know where `one` and `two` come from. `one` is implicitly being created as a query parameter, while `two` is a path parameter declared on the router. Even worse, when including `handler` in a different context *without* that path parameter, it'll be implicitly treated as a query parameter as well.

```python
@get("/")
async def handler(one: str, two: str) -> None:
 pass

router = Router("/{two:str}", route_handlers=[handler])
```

However, using explicit markers, it's immediately clear what the source of each parameter is, and errors can be caught much earlier:

```python
@get("/")
async def handler(one: FromQuery[str], two: FromPath[str]) -> None:
 pass

router = Router("/{two:str}", route_handlers=[handler])
```

## Changes

- Introduce new parameter markers:
  - `FromQuery`
  - `FromPath`
  - `FromCookie`
  - `FromHeader`
- Introduce new parameter specifiers (these follow the OpenAPI naming scheme):
  - `QueryParameter(name="something")` instead of `Parameter(query="something")`
  - `HeaderParameter(name="something")` instead of `Parameter(header="something")`
  - `CookieParameter(name="something")` instead of `Parameter(cookie="something")`
  - `PathParameter(name="something")` 
- Introduce deprecation warnings for:
  - Raw `Parameter()` calls without specifying a source
  - `Parameter` calls
  - Implicitly declared query parameters

## Migrating

The migration path is rather simple:

- Replace `Parameter` calls:
  - `Parameter(query=...)` -> `QueryParameter(name=...)` 
  - `Parameter(header=...)` -> `HeaderParameter(namer=...)` 
  - `Parameter(cookie=...)` -> `CookieParameter(name=...)` 
  - `Parameter(...)` without a source -> `QueryParameter()` / `HeaderParameter()` / `CookieParameter()` / `PathParameter()`
  - Hoist defaults `param: int = Parameter(default="something")` -> `param: FromQuery[int] = "something"`
- Explicitly declare unmarked parameters:
  - `FromQuery[<type>]` for query parameters
  - `FromPath[<type>]` for path parameters
  - `FromHeader[<type>]` for header parameters
  - `FromCookie[<type>]` for cookie parameters

### `Query[<type>]` style simple form

Use `Query[<type>]` instead of `FromQuery[<type>]` for the simple form.
While this is more succinct and mirrors APIs of other frameworks more closely, it also bears less context. `FromQuery` tells you both the origin *and* the direction; "This comes *from* a query parameter".

### `Query(...)` style for verbose form

Use `Query(...)` instead of `QueryParameter(...)` for the verbose form. `QueryParamter` is more verbose, and does not carry a lot of extra information. However, it has 2 advantages: 

1. It mirrors the OpenAPI naming scheme of "query parameters", "header parameters", "path parameter" and "cookie parameters"
2. It leaves room for a future API expansion / simplification. Right now, it's not possible to e.g. use `Query` as both the short and verbose form, which would have been the best compromise, however, with future updates to Python's typing specification / type checkers, this might become possible

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4750">https://litestar-org.github.io/litestar-docs-preview/4750</a>


docs/usage/routing/handlers.rst
```diff
@@ -410,12 +416,12 @@ containing the route handler instance and paths. It can also be used to build a
...
 
     @get("/def/{param:int}", name="three")
-    def handler_three(param: int) -> None:
+    def handler_three(param: FromPath[int]) -> None:
         pass
 
 
     @get("/{handler_name:str}", name="four")
-    def handler_four(request: Request, name: str) -> Redirect:
+    def handler_four(request: Request, name: FromQuery[str]) -> Redirect:
```
Kumzy: Should be ``handler_name`` instead of ``name`` no?

litestar/_kwargs/kwargs_model.py
```diff
@@ -484,3 +519,59 @@ def _validate_raw_kwargs(
...
+                f"{param_type_name} parameter {field_name!r} declared using deprecated default "
+                f"'param: <type> = Parameter(...)' style. Use '{short_alternative}[<type>]' "
+                f"or 'Annotated[<type>, {param_type_name.title()}Parameter(...)]' instead"
+            )
+        elif style == "annotated":
+            msg = (
+                f"{param_type_name} parameter {field_name!r} declared using deprecated annotated "
+                f"'param: Annotated[<type>, Parameter(...)]' style. Use "
+                f"'{short_alternative}[<type>]' or "
+                f"'Annotated[<type>, {param_type_name.title()}Param(...)]' instead"
```
Kumzy: ``Parameter`` instead of ``Param``?

docs/usage/routing/handlers.rst
provinzkraut: Yeah. Should also be `FromPath` :)

sobolevn (commented):
I haven't reviewed all 3000 changed lines 🙈, but I like the design.

I have a pretty similar one in DMR.
Thanks a lot for working on this!

Question: 

```python
@put("/{item_title:str}")
async def update_item(item_title: FromPath[str]): ...
```

we still define the type twice. What do others think: is that a problem? Can we solve this with `FromPath` now?

provinzkraut:
> I haven't reviewed all 3000 changed lines 🙈, but I like the design.
Yeah it's a lot, I know :sweat_smile: 
But most of it is quite trivial (about 90% of the code changes were made with a few `sed` calls).


> Question:
> 
> ```python
> @put("/{item_title:str}")
> async def update_item(item_title: FromPath[str]): ...
> ```
> 
> we still define the type twice. What do others think: is that a problem? Can we solve this with `FromPath` now?

Nothing really changed in that regard. We *could* make that typing optional, but then you'd have less consistency. The issue is, what happens in cases where the path parameter isn't consumed? We still need to generate an OpenAPI schema for it, so then you would have to specify it there.

sobolevn:
@provinzkraut maybe we can relax the path rule to be optional when `FromPath` exists? This will make

```python
@put("/{item_title}")
async def update_item(item_title: FromPath[str]): ...
```

less verbose :)

provinzkraut:
As mentioned [here](https://github.com/litestar-org/litestar/pull/4750#issuecomment-4466950836), this *would* be possible, but there's quite a few cases where it will not work, and I feel like it's better to have one way to do it, so it's consistent.

PR #4755, opened 2026-05-09
fix(openapi): set security=[] for routes excluded from auth

### LINKED ISSUE #3013: Bug: `AbstractSecurityConfig` sets `security` for all paths, even those `exclude`d
### Description

---

_No response_

### MCVE

```python
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.4.5

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

kimjune01:
Fixes #3013.

Routes excluded from auth via `exclude_from_auth` still had root-level
security applied in the OpenAPI spec. Per OpenAPI 3.1, setting `security: []`
on an operation overrides the root-level requirement.

Check `route_handler.opt.get("exclude_from_auth")` in path item generation
and set `security=[]` on the operation when true.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4755">https://litestar-org.github.io/litestar-docs-preview/4755</a>

### DIFF
--- a/litestar/_openapi/path_item.py
+++ b/litestar/_openapi/path_item.py
@@ -16,6 +16,7 @@ from litestar.utils.helpers import unwrap_partial
 if TYPE_CHECKING:
     from litestar._openapi.datastructures import OpenAPIContext
     from litestar.handlers.http_handlers import HTTPRouteHandler
+    from litestar.openapi.spec import SecurityRequirement
     from litestar.routes import HTTPRoute
 
 __all__ = ("create_path_item_for_route", "merge_path_item_operations")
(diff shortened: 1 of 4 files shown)


litestar/_openapi/path_item.py
```diff
@@ -72,6 +73,18 @@ def create_operation_for_handler_method(
             self.context, route_handler, raises_validation_error=raises_validation_error
         )
 
+        security: list[SecurityRequirement] | None
+        if route_handler.opt.get("exclude_from_auth"):
```
provinzkraut: The opt key is not static. It can be configured dynamically. This needs to be taken into account here

provinzkraut (changes requested):

litestar/_openapi/path_item.py
```diff
@@ -22,6 +22,28 @@
...
+
+    Args:
+        middleware: List of middleware configurations.
+
+    Returns:
+        A set of exclude_opt_key strings.
+    """
+    exclude_keys: set[str] = set()
+
+    for mw in middleware:
```
provinzkraut, line 36: this will break for subclassed middlewares / ones that use custom factories instead of `DefineMiddleware`

provinzkraut (changes requested):

provinzkraut:
Closing as per our [AI policy](https://github.com/litestar-org/.github/blob/main/AI_POLICY.md)

kimjune01:
Sorry for the noise. If you'd like to automatically block and ban AI PRs before they reach your review queue, here's a GitHub Action that catches all the common patterns: https://github.com/kimjune01/sweep/blob/master/action.yml

sobolevn:
@kimjune01 are you suggesting to use AI to detect AI PRs? :)

kimjune01:
@sobolevn AI is optional. Most of the work is happening as a shell script. give it a skim!

Issue #4523, opened 2025-12-13
Enhancement: allow customizing error OpenAPI schemas


Yolley:
### Summary

Method `create_error_responses` creates OpenAPI schemas assuming that generated responses from HTTPException conform to the default [schema](https://github.com/litestar-org/litestar/blob/c77d63bc8a1ed55a7ebfb762d5847dbfe2990a49/litestar/_openapi/responses.py#L295). But given that error responses can be customized via exception handlers or plugins (i.e. [problem_details](https://docs.litestar.dev/2/reference/plugins/problem_details.html#module-litestar.plugins.problem_details)).

May be it makes sense to either expose `ResponseFactory` customization or at least be able to provide a custom `create_error_responses` handler to OpenAPIConfig or the app itself?

<details>
<summary>For now I am patching the method like so</summary>

```python
def create_problem_details_responses(exceptions: list[type[HTTPException]]) -> Iterator[tuple[str, OpenAPIResponse]]:
    """Create the schema for error responses, if any."""
    grouped_exceptions: dict[int, list[type[HTTPException]]] = {}
    for exc in exceptions:
        if not grouped_exceptions.get(exc.status_code):
            grouped_exceptions[exc.status_code] = []
        grouped_exceptions[exc.status_code].append(exc)
    for status_code, exception_group in grouped_exceptions.items():
        exceptions_schemas = []
        group_description: str = ""
        for exc in exception_group:
            example_title = ""
            example_detail = ""
            if hasattr(exc, "detail") and exc.detail:
                group_description = exc.detail
                example_detail = exc.detail

            if not example_title:
                with contextlib.suppress(Exception):
                    example_title = HTTPStatus(status_code).phrase

            exceptions_schemas.append(
                Schema(
                    type=OpenAPIType.OBJECT,
                    required=["title", "status"],
                    properties={
                        "status": Schema(type=OpenAPIType.INTEGER),
                        "title": Schema(type=OpenAPIType.STRING),
                        "detail": Schema(type=OpenAPIType.STRING),
                        "extra": Schema(
                            type=[OpenAPIType.NULL, OpenAPIType.OBJECT, OpenAPIType.ARRAY],
                            additional_properties=Schema(),
                        ),
                    },
                    description=pascal_case_to_text(get_name(exc)),
                    examples=[{"status": status_code, "title": example_title, "detail": example_detail, "extra": {}}],
                )
            )
[...]

Issue #4003, opened 2025-02-15
Bug: Very difficult to customize exceptions outputted to OpenAPI


umarbutler:
### Description

Lets say I have the below custom exception:
```
class ChunkingTimeoutError(HTTPException):
    status_code = 504
    detail = "Chunking timed out. Did you try to chunk a very large text with a very low chunk size or very little variation in levels of whitespace?"
```

There is no way that I can see for me to make the description of the status code itself "Chunking timed out" or "Timeout error" while at the same time having the description of the error itself (as defined in `detail`) be the longer message "Chunking timed out. Did you try to chunk a very large text with a very low chunk size or very little variation in levels of whitespace?".

Additionally, if I try this:
```
class ChunkingTimeoutError(HTTPException):
    status_code = 504
    detail = "Chunking timed out"
    extra = "Did you try to chunk a very large text with a very low chunk size or very little variation in levels of whitespace?"
```

The `extra` will not appear in the OpenAPI spec. Instead it will still be set to `{}`.

Annotating `ChunkingTimeoutError` with a `Body` does something quite weird as well.
```
class ChunkingTimeoutError(HTTPException):
    status_code = 504
    detail = "Chunking timed out. Did you try to chunk a very large text with a very low chunk size or very little variation in levels of whitespace?"

ChunkingTimeoutError = Annotated[ChunkingTimeoutError, Body(description="Chunking timed out")]
```

The above code will cause the description of the error to literally be set to the word "Annotated".

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Logs

```bash

```

### Litestar Version

2.14.0 final

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

umarbutler:
I'm also unable to customize the title of exceptions. `__schema_name__` did not help.

Umpire2018:
```python
from litestar import Litestar, get
from litestar.exceptions import HTTPException

class ChunkingTimeoutError(HTTPException):
    status_code = 504
    detail = "Chunking timed out. Did you try to chunk a very large text with a very low chunk size or very little variation in levels of whitespace?"

@get(
    "/greet",
    raises=[ChunkingTimeoutError],
)
[...]

Issue #3021, opened 2024-01-24
Enhancement: Generate `raises` `HTTPException`s as OpenAPI components


tuukkamustonen:
### Summary

```py
@get(raises=[NotFoundException])
def endpoint(): ...
```

...seems to inject each declared `raises` entry as duplication into the generated OpenAPI spec:

```json
          "404": {
            "description": "Nothing matches the given URI",
            "content": {
              "application/json": {
                "schema": {
                  "properties": {
                    "status_code": {
                      "type": "integer"
                    },
                    "detail": {
                      "type": "string"
                    },
                    "extra": {
                      "additionalProperties": {},
                      "type": [
                        "null",
                        "object",
                        "array"
                      ]
                    }
                  },
                  "type": "object",
                  "required": [
                    "detail",
                    "status_code"
                  ],
                  "description": "Not Found Exception",
                  "examples": {
                    "NotFoundException": {
                      "value": {
                        "status_code": 404,
                        "detail": "Not Found",
                        "extra": {}
                      }
                    }
                  }
                }
              }
            }
          },
```

That's quite lengthy, consider 100 endpoints and it's going to be there 100 times.

Could these be injected into the components section, rather:

```json
    "components": {
        "schemas": {
            "NotFoundException": {
              "description": "Not Found Exception",
              "type": "object",
              "properties": {
                "status_code": {
                  "type": "integer"
                },
                "detail": {
                  "type": "string"
                },
                "extra": {
                  "additionalProperties": {},
                  "type": [
                    "null",
                    "object",
                    "array"
                  ]
                }
              },
              "required": [
                "detail",
                "status_code"
              ],
              "examples": {
                "NotFoundException": {
                  "value": {
                    "status_code": 404,
                    "detail": "Not Found",
                    "extra": {}
                  }
                }
              }
```

Then the references much become much shorter:

```json
                    "404": {
                        "description": "Nothing matches the given URI",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/NotFoundException"
                                }
                            }
                        }
                    },
```

provinzkraut:
@guacs

guacs:
I think this seems reasonable.

I think there was some discussion related to this with @peterschutt. I think his idea was to move everything into components (i.e. all complex types including user defined types) and only have references to them within the schema.

peterschutt:
[...]

Issue #3020, opened 2024-01-24
Enhancement: Allow customizing `HTTPException` content in OpenAPI spec


tuukkamustonen:
### Summary

How `HTTPException`s get into OpenAPI spec is done is in https://github.com/litestar-org/litestar/blob/14759558cf231c61049fdd5b65a39eccb8841616/litestar/_openapi/responses.py#L280-L285. The fields written (`status_code`, `detail`, `extra`) is hard-coded there.

How about asking this from the `HTTPException` class instead, so that if adds fields or customizes `HTTPException` in general, it would be also possible to reflect that in the OpenAPI spec?

tuukkamustonen:
Thinking out loud, for the rationale here, we often add `error` or similar field to the responses:

```json
{
    "status_code": 409,
    "detail": "User already has thing X",
    "error": "AFP00146"
}
{
    "status_code": 409,
    "detail": "Some other case which leads to conflict",
    "error": "AFP00147"
}
```

This will allow the caller to identify the error, due to the more granular `error` compared to 409 status code.

Two things here:

1. Need to add that `error` to the OpenAPI spec (what this PR is about)
2. Would need to somehow render this in OpenAPI GUIs

Regarding (2), when you configure `raises=[BadRequest]

```py
class UserAlreadyHasThingX(HTTPException):
    status_code: int = HTTP_409_CONFLICT
    error: str = "AFP00146"

class SomeOtherConflict(HTTPException):
    status_code: int = HTTP_409_CONFLICT
    error: str = AFP00147"  # or this could be `some_other_conflict`, as long as it's identifier

@get(..., raises=[UserAlreadyHasThingX, SomeOtherConflict])
def endpoint(): ...
```

This will generate `oneOf` array in the OpenAPI schema, which is nice. All GUIs know how to show them, e.g..

![image](https://github.com/litestar-org/litestar/assets/94327/5facb6aa-f71a-426f-ab75-23538737b8e4)

The problem is, that both schemas are similar (except their description), so there's no benefit. We'd need to generate literal/enum `AFP00146`/`AFP00147` for the `error` fields, so that reader would understand that despite the similar fields (in all error responses) the values (and descriptions) for certain errors are different.

PR #4757, opened 2026-05-09
feat(openapi): OpenAPISpecPlugin for plugin-contributed document fragments

cofin:
This PR adds `OpenAPISpecPlugin`, a small public extension point that lets any plugin contribute fragments to the generated OpenAPI document. The hook is generic, not security-shaped — a plugin returns a sparsely-populated `Components` for the document and a sparsely-populated `Operation` for each route handler, and Litestar's schema constructor merges those fragments in with explicit, audited per-field rules.

The class lives at `litestar.plugins.OpenAPISpecPlugin` with two no-op default methods: `get_openapi_components()` and `get_openapi_operation(route_handler)`. Plugins are global — there is no `include`/`exclude` constructor knob. Per-route decisions are made inside the plugin's own contribution method, where it already has the route handler in hand. If a path-pattern filter is wanted later, it can mirror `ASGIMiddleware`'s exclusion knobs verbatim instead of inventing a third dialect.

Components and Operation merges raise `ImproperlyConfiguredException` on dict-key collisions instead of silently overwriting. `merge_openapi_components` walks an explicit allowlist of `Components` dict fields rather than introspecting via `dataclasses.fields()`, and names the field, the colliding key, and the contributing plugin label in the error message. `merge_openapi_operation` lands alongside as the per-operation merger: `security` extends, `tags` extend-and-dedup, `parameters` and `callbacks` raise on collision (parameters key on `(name, param_in)`), and `responses` merge by status code with collisions raising. Fields owned by the route handler and the schema constructor — `operation_id`, `summary`, `description`, `request_body`, `deprecated`, `external_docs` — raise at merge time if a plugin tries to set them.

The mergeable `responses` field is the upstream-facing answer to the exception-customization cluster (#2416, #3020, #3021, #4003, #4523): an exception-response plugin can return `Operation(responses={...})` and have those entries land in the operation without monkey-patching `OpenAPIConfig`. To make that work without colliding on every default-emitted error response, `PathItemFactory.create_operation_for_handler_method` runs a three-phase orchestration: collect each plugin fragment exactly once and union the `fragment.responses` keys into a `plugin_owned_status_codes` set; build the operation, threading that set into `create_responses_for_handler` so the default `ResponseFactory.create_responses` skips matching exceptions from default emission; merge the cached fragments. The returned fragment IS the plugin's claim — no separate discriminator method is needed.

`create_responses_for_handler` and `ResponseFactory.create_responses` gain an optional `plugin_owned_status_codes: set[str] | None = None` parameter. The default `None` preserves today's behavior byte-for-byte; only apps that register an `OpenAPISpecPlugin` returning `Operation(responses={...})` see the filter take effect.

No security semantics, no auth changes — just the document plumbing. Apps without an `OpenAPISpecPlugin` produce byte-identical OpenAPI output to before.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4757">https://litestar-org.github.io/litestar-docs-preview/4757</a>


litestar/plugins/base.py
```diff
@@ -276,10 +280,85 @@ def is_constrained_field(field_definition: FieldDefinition) -> bool:
...
+            ``exclude`` takes precedence over ``include``.
+    """
+
+    __slots__ = ("exclude", "include")
+
+    def __init__(
+        self,
+        *,
+        include: str | Iterable[str] | None = None,
+        exclude: str | Iterable[str] | None = None,
```
provinzkraut: What's the use case for these? Our current plugin mechanism is globally, i.e. they either don't apply to routes directly or to all routes.

litestar/plugins/base.py
```diff
@@ -276,10 +280,85 @@ def is_constrained_field(field_definition: FieldDefinition) -> bool:
...
+        self.include = include
+        self.exclude = exclude
+
+    def applies_to(self, route_handler: HTTPRouteHandler) -> bool:
+        """Return ``True`` if this plugin should contribute for ``route_handler``.
+
+        The default implementation honors :attr:`include` and :attr:`exclude` path patterns.
+        Subclasses can override this method for custom filtering logic.
+        """
+        paths = route_handler.paths
```
provinzkraut: If this is implemented, I think it should match (and probably re-use) the middleware exclusion logic (the one found in `ASGIMiddleware`)

litestar/plugins/base.py
```diff
@@ -276,10 +280,85 @@ def is_constrained_field(field_definition: FieldDefinition) -> bool:
...
+
+    def get_openapi_components(self) -> Components | None:
+        """Return OpenAPI components contributed by this plugin.
+
+        Returns:
+            Components to merge into the generated OpenAPI document, or ``None``.
+        """
+        return None
+
+    def get_openapi_security_requirements(
```
provinzkraut: This seems a bit too specific. Maybe it would make sense to have this be more general so that it can return arbitrary components for the route, and our schema constructor will figure out what to make of them?

litestar/_openapi/plugin.py
```diff
@@ -23,10 +24,21 @@
...
     from litestar.openapi.plugins import OpenAPIRenderPlugin
-    from litestar.openapi.spec import OpenAPI, PathItem
+    from litestar.openapi.spec import Components, OpenAPI, PathItem
     from litestar.routes import BaseRoute
 
 
+def merge_openapi_components(target: Components, source: Components) -> None:
+    """Merge contributed OpenAPI components into a generated components object."""
+    for field in fields(source):
+        if source_value := getattr(source, field.name, None):
```
provinzkraut: This is a dangerous pattern we've been bitten a *lot* by in the past. We need to make this more sound (less dynamic), and think about merge scenarios / conflicts, similar to what we're doing with the route handler merges

tests/unit/test_openapi/test_responses.py
```diff
@@ -162,6 +164,111 @@ class HouseNotFoundError(HTTPException):
     assert house_not_found_exc_response[1].description == HouseNotFoundError.detail
 
 
+# ---------------------------------------------------------------------------
+# T5b — plugin_owned_status_codes pre-pass parameter
```
cofin:
```suggestion
# plugin_owned_status_codes pre-pass parameter
```

PR #4759, opened 2026-05-09
feat(otel): backport litestar.plugins.opentelemetry + deprecate contrib namespace

cofin:
This moves the OpenTelemetry integration to `litestar.plugins.opentelemetry` and turns `litestar.contrib.opentelemetry` into a deprecation shim. The same move landed on main as #4691; this is the v2 side, so 2.x users get both a real migration target on their current version and a `DeprecationWarning` pointing them at it. They can update imports while still on 2.x and upgrade cleanly to 3.0.0.

Three symbols are covered: `OpenTelemetryConfig`, `OpenTelemetryInstrumentationMiddleware`, and `OpenTelemetryPlugin`. The contrib location — which was the real implementation on v2 until now — becomes a thin PEP 562 `__getattr__` shim that re-exports from the plugin namespace.

```python
# before
from litestar.contrib.opentelemetry import OpenTelemetryPlugin

# after
from litestar.plugins.opentelemetry import OpenTelemetryPlugin
```

Two upstream bug fixes ride along with the move:

- A `route_handler is None` guard in the middleware helper, fixing a `KeyError` when the OTel middleware runs as the outermost layer before route resolution (#4468)
- Corrected client request/response hook signatures — three params instead of two, matching `opentelemetry-instrumentation-asgi` (#4466)

The config also exposes `tracer` and `http_capture_headers_*` fields that #4691 added on main.

Tests split cleanly: `tests/unit/test_contrib/test_opentelemetry.py` is now the deprecation suite (each access fires `DeprecationWarning` and resolves to the same object as the plugin namespace), and `tests/unit/test_plugins/test_opentelemetry.py` carries the behavior tests for the new namespace.

Deprecated in 2.22.0, removed in 3.0.0. Changelog entry added under the new 2.22.0 block.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4759">https://litestar-org.github.io/litestar-docs-preview/4759</a>


docs/release-notes/changelog.rst
provinzkraut, line 1: No manually created changelog for `v2` branches

docs/usage/middleware/builtin-middleware.rst
```diff
@@ -318,7 +318,7 @@ offering client- and server-side sessions. Server-side sessions are backed by Li
 - File based sessions
 - Redis based sessions
 - Valkey based sessions
-- Database based :ref:`advanced-alchemy:usage/frameworks/litestar:Session Middleware`
+- Database based session middleware (via ``advanced_alchemy.extensions.litestar``)
```
provinzkraut: This seems unrelated?

provinzkraut (changes requested):

docs/usage/middleware/builtin-middleware.rst
cofin: It's there, but there is an issue with the autodocs on AA.  I've opened an issue to fix this (and several other similar issues) at AA.

provinzkraut (approved):

PR #4767, opened 2026-05-10
fix(openapi): set security=  [ ] for routes not in from auth

### LINKED ISSUE #3013: Bug: `AbstractSecurityConfig` sets `security` for all paths, even those `exclude`d

TanayK07:
Closes  #3013

## Description

We use both FastAPI and Litestar at work and noticed routes with `exclude_from_auth=True` were still showing auth in OpenAPI docs becuase security was being set at the root level and nothing was overriding it.

Fix was to pass `exclude_opt_key` into `OpenAPIConfig` during `on_app_init`, then while generating path items if that opt key is set on a handler it adds `security: []` to override the global one.

Also works for custom `exclude_opt_key` values not only `"exclude_from_auth"`.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4767">https://litestar-org.github.io/litestar-docs-preview/4767</a>

### DIFF
--- a/litestar/openapi/config.py
+++ b/litestar/openapi/config.py
@@ -118,6 +118,7 @@ class OpenAPIConfig:
 
     def __post_init__(self) -> None:
         self.path = normalize_path(self.path)
+        self.security_exclude_opt_keys: set[str] = set()
 
         self.default_plugin: OpenAPIRenderPlugin | None = None
         for plugin in self.render_plugins:
--- a/litestar/_openapi/path_item.py
+++ b/litestar/_openapi/path_item.py
@@ -72,6 +72,14 @@ class PathItemFactory:
             self.context, route_handler, raises_validation_error=raises_validation_error
         )
 
+        exclude_keys = self.context.openapi_config.security_exclude_opt_keys
+        if exclude_keys and any(route_handler.opt.get(key) for key in exclude_keys):
+            security = []
+        elif route_handler.security:
+            security = list(route_handler.security)
+        else:
+            security = None
+
         return route_handler.operation_class(
             operation_id=operation_id,
             tags=sorted(route_handler.tags) if route_handler.tags else None,
@@ -81,7 +89,7 @@ class PathItemFactory:
             responses=responses,
             request_body=request_body,
             parameters=parameters or None,  # type: ignore[arg-type]
-            security=list(route_handler.security) if route_handler.security else None,
+            security=security,
         )
 
     def create_operation_id(self, route_handler: HTTPRouteHandler, http_method: HttpMethod) -> str:
--- a/litestar/security/base.py
+++ b/litestar/security/base.py
@@ -99,6 +99,8 @@ class AbstractSecurityConfig(ABC, Generic[UserType, AuthType]):
             else:
                 app_config.openapi_config.security = [self.security_requirement]
 
+            app_config.openapi_config.security_exclude_opt_keys.add(self.exclude_opt_key)
+
         if self.guards:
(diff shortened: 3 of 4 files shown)


provinzkraut:
I just closed #4755 yesterday with almost the same content. That PR is linked on the issue this aims to close. Please at least take a look at the closed PRs and why they were closed. Closing this for the same reasons + low effort

TanayK07:
@provinzkraut apologies but I don't understand. I did take a look at the PR which was closed, at the same time, my code wasn't AI slop , it works was tested for my org, I just spun up a slightly different one from my org's fork. Also both the comments in the previous PR don't apply to me. could you kindly PTAL and LMK what is wrong with the approach. Thanks

TanayK07:
> @provinzkraut apologies but I don't understand. I did take a look at the PR which was closed, at the same time, my code wasn't AI slop , it works was tested for my org, I just spun up a slightly different one from my org's fork. Also both the comments in the previous PR don't apply to me. could you kindly PTAL and LMK what is wrong with the approach. Thanks

The two issues raised there dont really apply to my approach. I'm not touching middleware at all, so middleware introspection breaking for subclasses isnt a problem here.

provinzkraut:
Sorry @TanayK07, I only did a cursory check of your changes, and they seemed very similar to the closed PR. Unfortunately I only have so much time to review PRs, and with the recent influx of low-effort contributions, if something seems like it fits that, I'm more eager to just close it than I usually would. You got caught in the crossfire here.

litestar/openapi/config.py
```diff
@@ -118,6 +118,7 @@ class OpenAPIConfig:
 
     def __post_init__(self) -> None:
         self.path = normalize_path(self.path)
+        self.security_exclude_opt_keys: set[str] = set()
```
provinzkraut, line 121: Dataclass fields should be defined as class members, not here

provinzkraut (commented):
Thanks @TanayK07, but I still don't think this is the right way to go about this. The issue is that it doesn't scale; It adds special casing just for the security middlewares and their configs, so the OpenAPI constructor now is directly tied to those. I think the better approach would be to use what #4757 proposes, and then have the plugin exclude routes based on the patterns / opt keys, etc.

TanayK07:
> Sorry @TanayK07, I only did a cursory check of your changes, and they seemed very similar to the closed PR. Unfortunately I only have so much time to review PRs, and with the recent influx of low-effort contributions, if something seems like it fits that, I'm more eager to just close it than I usually would. You got caught in the crossfire here.

I understand. No issues. I have already actually contributed once to litestar  before and I run a repo of my own, small one with some (670 odd stars) I understand your issue.

TanayK07:
> Thanks @TanayK07, but I still don't think this is the right way to go about this. The issue is that it doesn't scale; It adds special casing just for the security middlewares and their configs, so the OpenAPI constructor now is directly tied to those. I think the better approach would be to use what #4757 proposes, and then have the plugin exclude routes based on the patterns / opt keys, etc.

That makes sense.  I see how wiring exclude_opt_key into OpenAPIConfig is too specific when #4757 gives a generic way to do it. I will rebase it on top once that gets merged in a couple of weeks then.

PR #4786, opened 2026-05-21
fix: change code to reflect changes of psycopg library

Kumzy:
Explanation for the changes of psycopg explained there: https://github.com/psycopg/psycopg/pull/1092#issuecomment-3263261593

`notifies()` now holds the connection lock across every `yield`. While a consumer is suspended on a yielded notification,any concurrent `execute()` on that connection (e.g. the `UNLISTEN` issued by `unsubscribe`) blocks until the consumer iterates again.

Initial issue raised [psycopg/psycopg#1128]

Replaced the direct `notifies()` loop in `stream_events` with a background pump task feeding an `asyncio.Queue`. `stream_events` now reads from the queue and filters by `_subscribed_channels`. This is the same pattern `AsyncPgChannelsBackend` already uses, the connection lock is no longer held across user-visible suspensions, and the filter handles the post`UNLISTEN` backlog.

The pump uses `notifies(timeout=0.1)` like ``psycopg``is doing.
https://github.com/psycopg/psycopg/blob/3.3.4/psycopg/psycopg/connection_async.py#L45

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4786">https://litestar-org.github.io/litestar-docs-preview/4786</a>


litestar/channels/backends/psycopg.py
```diff
@@ -46,9 +58,22 @@ async def unsubscribe(self, channels: Iterable[str]) -> None:
 
         self._subscribed_channels = self._subscribed_channels - set(channels)
 
+    async def _pump_notifications(self) -> None:
+        # psycopg 3.2+ holds the connection lock across notifies(); cycling on a
```
provinzkraut, line 62: Mhm. This doesn't feel right. Seems like it might considerably impact performance. Is this really the only way to implement this properly? Maybe we raise the issue in the upstream repo again / ask for clarification on how to deal with this?

Kumzy: Right I am not sure about performance but it can only worsen performances. I can try to ask for clarification, will try to create an issue for that

Kumzy: https://github.com/psycopg/psycopg/issues/1310

PR #4801, opened 2026-05-23
feat(core): Introduce generic request body markers

provinzkraut:
Introduce generic request body markers, following the pattern established in #4750.

Add:

- `JSONBody`
- `MsgPackBody`
- `MultipartBody`
- `URLEncodedBody`

They can be used just like request parameter markers, and replace the more clunky `Annotated[T, Body(media_type=RequestEncodingType.MULTI_PART))]` form.

For example

```python
@post("/")
async def handler(data: Annotated[dict, Body(media_type=RequestEncodingType.MULTI_PART))]) -> None:
    pass
``` 
can be replaced with

```python
@post("/")
async def handler(data: MultipartData[dict]) -> None:
    pass
```

## Migrating

Replace

- `Annotated[T, Body(media_type=RequestEncodingType.JSON))]` with `JSONBody[T]`
- `Annotated[T, Body(media_type=RequestEncodingType.MESSAGEPACK))]` with `MsgPackBody[T]`
- `Annotated[T, Body(media_type=RequestEncodingType.MULTI_PART))]` with `MultipartBody[T]`
- `Annotated[T, Body(media_type=RequestEncodingType.URL_ENCODED))]` with `URLEncodedBody[T]`

## Open questions

Currently not implemented is *required* `data` marking. A bare `data:` parameter will be inferred to be JSON. Do we want to change this, for the sake of being more explicit, or do we allow this affordance? 

My personal feeling is that, if we want to move away from the magic parameters like `data`, we should require explicit marking in all cases. That is, the request body will always be injected into a parameter marked with `Body()`.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4801">https://litestar-org.github.io/litestar-docs-preview/4801</a>

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -16,23 +16,23 @@ jobs:
         with:
           script: |
             const title = context.payload.pull_request.title;
-            
+
             const types = [
                 "build",
                 "chore",
-                "ci", 
-                "docs", 
-                "feat", 
-                "fix", 
-                "perf", 
-                "refactor", 
-                "revert", 
-                "style", 
+                "ci",
+                "docs",
+                "feat",
+                "fix",
+                "perf",
+                "refactor",
+                "revert",
+                "style",
                 "test",
             ];
-            
+
             const pattern = new RegExp(`^(${types.join("|")})(\\([^)]+\\))?!?: .+`);
-            
+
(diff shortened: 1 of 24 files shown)


cofin (approved):
Looks good!

PR #4805, opened 2026-05-25
fix(core): Fix `Annotated` parmeters implicitly treated as query parameters (#4804)

### LINKED ISSUE #4804: Bug: `Annotated` dependency parameters causing `Kwarg resolution ambiguity detected`
### Description

Since `2.22.0`, wrapping any dependency parameter in `Annotated` results in 

```text
ImproperlyConfiguredException: 500: Kwarg resolution ambiguity detected for the following keys: foo. Make sure to use distinct keys for your dependencies, path parameters, and aliased parameters.
```

```python
from typing import Annotated

from litestar import get, Litestar


async def provide_foo() -> str:
    return ""


@get("/", dependencies={"foo": provide_foo})
async def handler(foo: Annotated[int, "aaa"]) -> None:
    pass


app = Litestar([handler])
```

provinzkraut:
Fix a bug introduced in #4783, that would cause bare handler parameters wrapped in `Annotated` and not explicitly marked with the new parameter markers (`FromQuery[]` etc.) to be treated as a query parameter, causing validation issues.

Closes #4804.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4805">https://litestar-org.github.io/litestar-docs-preview/4805</a>

### DIFF
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -519,7 +519,7 @@ class FieldDefinition:
             # if not, create a new KwargDefinition
             else:
                 if (name := kwargs.get("name")) == "data":
-                    model = BodyKwarg
+                    model: type[KwargDefinition] = BodyKwarg
                 elif name is None:
                     model = KwargDefinition
                 else:
--- a/litestar/typing.py
+++ b/litestar/typing.py
@@ -518,7 +518,12 @@ class FieldDefinition:
                 )
             # if not, create a new KwargDefinition
             else:
-                model = BodyKwarg if kwargs.get("name") == "data" else ParameterKwarg
+                if (name := kwargs.get("name")) == "data":
+                    model = BodyKwarg
+                elif name is None:
+                    model = KwargDefinition
+                else:
+                    model = ParameterKwarg
                 kwargs["kwarg_definition"] = model(**kwarg_definition_merge_args)
 
         kwargs.setdefault("annotation", unwrapped)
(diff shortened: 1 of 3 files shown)


litestar/typing.py
```diff
@@ -518,7 +518,12 @@ def from_annotation(cls, annotation: Any, **kwargs: Any) -> FieldDefinition:  #
                 )
             # if not, create a new KwargDefinition
             else:
-                model = BodyKwarg if kwargs.get("name") == "data" else ParameterKwarg
+                if (name := kwargs.get("name")) == "data":
```
sobolevn, line 521: this can be unwrapped into `elif` with one less indentation

provinzkraut: How does unwrapping this into `elif` remove a level of indentation? I don't think I follow :(

sobolevn:
```python
else:
    if (name := kwargs.get("name")) == "data": 
        model = BodyKwarg
```

->

```python
elif (name := kwargs.get("name")) == "data": 
    model = BodyKwarg
```

:)

provinzkraut: Ah, *that* branch. Yes, that's better!

provinzkraut: Although I must say that I find the non-elif version easier to read than the collapsed branches.

PR #4641, opened 2026-03-23
docs: add section on default OpenAPI schema file endpoints

tysoncung:
## Problem

The documentation for OpenAPI didn't explicitly document the availability of the default schema file endpoints (`openapi.json`, `openapi.yaml`, `openapi.yml`). Users had to discover these paths by reading the source code or stumbling upon brief mentions in other sections.

## Solution

Added a new "Default OpenAPI Schema Files" section to the OpenAPI UI Plugins documentation that clearly documents:

- `/schema/openapi.json` — always available (auto-registered even without explicit config)
- `/schema/openapi.yaml` and `/schema/openapi.yml` — available with `YamlRenderPlugin`
- Note about PyYAML dependency requirement
- Example URLs for common use
- How these paths change when the root path is customized
- Mention of external tools (Scalar API Client, Postman) that can consume these endpoints

## Files Changed

- `docs/usage/openapi/ui_plugins.rst`: Added new documentation section before the "Configuring the OpenAPI Root Path" section

Closes #3941

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4641">https://litestar-org.github.io/litestar-docs-preview/4641</a>

PR #4807, opened 2026-05-25
docs(openapi): document default schema file endpoints (#3941)

### LINKED ISSUE #3941: Docs: Add information on the default OpenAPI files available
### Summary

In https://docs.litestar.dev/latest/usage/openapi we don't really *specifically* say anything about the availability of `$openapi_path/openapi.{json,yaml,yml}` except in 2 places, 1 of which says it is deprecated: 
- https://docs.litestar.dev/latest/usage/openapi/ui_plugins.html#providing-a-subclass-of-openapicontroller (but just glossing over)
- https://docs.litestar.dev/latest/usage/openapi/ui_plugins.html#configuring-the-openapi-root-path (but just glossing over)

It would be nice to document this when people need them e.g., for Scalar API client's collection import feature

vinicq:
## Summary

Closes #3941.

Adds a dedicated section to `docs/usage/openapi/ui_plugins.rst` covering where the OpenAPI schema is actually served, since the current docs only gloss over it.

What the section now documents:

- The auto-registered `/<openapi_path>/openapi.json` endpoint. It's always available even when `render_plugins` does not explicitly include `JsonRenderPlugin`, because UI plugins need a JSON document to point at. This behavior lives in `litestar/_openapi/plugin.py` and was not previously called out.
- How to enable `/<openapi_path>/openapi.yaml` and `/<openapi_path>/openapi.yml` by adding `YamlRenderPlugin` to `render_plugins`, plus the `litestar[yaml]` extra requirement.
- Concrete guidance for users importing the schema into external tools (Scalar API Client, Postman, Insomnia, Bruno, `openapi-typescript`, `oasdiff`), which is the use case the issue specifically called out.

The section sits right after "Configuring the OpenAPI Root Path" so the path resolution context flows naturally into "and here is what lives under that path."

## Test plan

- [x] Section renders as valid RST (verified against `docutils` parser, no new errors introduced)
- [x] Confirmed against source: `litestar/openapi/plugins.py` for the JSON/YAML default paths and `litestar/_openapi/plugin.py:178-179` for the auto-registration of `JsonRenderPlugin` when no other plugin claims `/openapi.json`
- [x] No code changes, so no existing tests are affected

### DIFF
--- a/docs/usage/openapi/ui_plugins.rst
+++ b/docs/usage/openapi/ui_plugins.rst
@@ -151,6 +151,56 @@ In the following example, we configure the OpenAPI root path to be ``/docs``:
 This will result in any of the OpenAPI endpoints being served at ``/docs`` instead of ``/schema``, e.g.,
 ``/docs/openapi.json``.
 
+Serving the OpenAPI schema as a downloadable file
+-------------------------------------------------
+
+In addition to the UI rendered at the OpenAPI root path, Litestar can also serve the raw OpenAPI schema document at
+predictable URLs underneath that root path. These endpoints are what tools like Scalar API Client, Postman, Insomnia,
+and CI validators consume when they import or validate an API contract.
+
+Default endpoints
+~~~~~~~~~~~~~~~~~
+
+With the default ``OpenAPIConfig`` (no custom ``render_plugins``), Litestar serves:
+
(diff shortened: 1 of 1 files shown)


provinzkraut:
There already is an open PR for this: #4641

PR #4808, opened 2026-05-25
feat(di): Deprecate `params.Dependency`, introduce `di.NamedDependency` and `params.SkipValidation`

provinzkraut:
Add a new `NamedDependency` marker, building upon the work of #4750 and #4801. This completes the move to generic parameter markers to indicate the value source.

`NamedDependency` is also a preparation for the overhauled di that will be introduced in version 3, which supports typed-based DI alongside the current name-based DI, and will feature its counterpart `TypeDependency`.

<hr>

- Add `di.NamedDependency` generic, replacing `Annotated[<type>, Depenency()]` as a marker
- Deprecate `params.Dependency`
- Introduce `params.SkipValidation` generic. `Dependency(skip_validation=True)` can be cleanly migrated to this. In addition, `SkipValidation` works on all parameters, not just dependencies

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4808">https://litestar-org.github.io/litestar-docs-preview/4808</a>

### DIFF
--- a/litestar/params.py
+++ b/litestar/params.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, ClassVar, Hashable, Sequence, TypeVar
 from typing_extensions import Annotated, TypeAlias
 
 from litestar.enums import ParamType, RequestEncodingType
+from litestar.exceptions import LitestarDeprecationWarning
 from litestar.types import Empty
 
 __all__ = (
@@ -28,8 +29,12 @@ __all__ = (
     "ParameterKwarg",
     "PathParameter",
     "QueryParameter",
+    "SkipValidation",
+    "SkipValidationMarker",
     "URLEncodedBody",
 )
+
+from litestar.utils import deprecated
 
 if TYPE_CHECKING:
     from litestar.openapi.spec.example import Example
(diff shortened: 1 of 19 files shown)


cofin (approved):
LGTM

litestar/typing.py
```diff
@@ -11,6 +11,7 @@
 
 from typing_extensions import Annotated as te_Annotated
 
+from litestar import di
```
Kumzy, line 14: Don't you want to specify the import there instead of importing the whole module?
``from litestar.di import Dependency`` or use ``as``

litestar/params.py
```diff
@@ -529,7 +534,29 @@ def __hash__(self) -> int:
...
+        if self.skip_validation:
+            warnings.warn(
+                "Deprecated parameter 'skip_validation'. This will be removed in "
+                "Litestar 3.0. To skip validation of a dependency parameter, annotate "
+                "it as 'SkipValidation[<type>]' instead",
+                category=LitestarDeprecationWarning,
+                stacklevel=2,
+            )
+        else:
+            object.__setattr__(self, "skip_validation", False)
```
Kumzy: Is that really updating ? 
I mean we have a default ``skip_validation: bool = field(default=False)`` and the class is using a frozen dataclass.

Kumzy (commented):
My 2 comments + a general comment: 
As ``params.Dependency`` is getting deprecated, is ``params.DependencyKwarg`` going to be deprecated in v3 too?

litestar/params.py
provinzkraut: Ah no, that's not necessary anymore. I had the default as `Empty` in an earlier version

litestar/typing.py
provinzkraut: Kinda intentional, because we also have `_kwargs.dependencies.Dependency` and `params.Dependency`. It's a bit confusing for now until we can get rid of the first two in v3.

provinzkraut:
> As `params.Dependency` is getting deprecated, is `params.DependencyKwarg` going to be deprecated in v3 too?

It is. Currently there's no way to use it without getting a deprecation warning, but I'll add an explicit deprecation warning as well

litestar/params.py
provinzkraut: Removed

PR #4814, opened 2026-05-28
fix(openapi): handle Annotated metadata in Optional[Literal[...]] schemas

pavdwest:
`Optional[Annotated[Literal[...], Meta(...)]]` raised a `KeyError` during
schema generation because `make_non_optional_union` preserves the
`Annotated` wrapper, and `_iter_flat_literal_args` then iterated the PEP
593 metadata as if it were a literal value.

Strip wrapper annotations at the top of `_iter_flat_literal_args` so the
function is robust regardless of where it is entered from. Add a
parametrized regression test covering both `msgspec.Meta` and
`annotated_types` metadata to confirm the fix is shape-agnostic.

## Description

Minimal example detailing the issue:

```python
from typing import Annotated, Literal, Optional
from msgspec import Meta, Struct

Flag = Annotated[Literal["A", "B"], Meta(description="legacy")]

class Body(Struct):
    flag: Optional[Flag] = None   # <-- Optional is what exposes the bug

@post("/x")
def x(data: Body) -> Body: return data

app = Litestar(route_handlers=[x])
```

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4814">https://litestar-org.github.io/litestar-docs-preview/4814</a>

### DIFF
--- a/litestar/_openapi/schema_generation/schema.py
+++ b/litestar/_openapi/schema_generation/schema.py
@@ -49,6 +49,7 @@ from litestar.utils.predicates import (
 from litestar.utils.typing import (
     get_origin_or_inner_type,
     make_non_optional_union,
+    unwrap_annotation,
     unwrap_new_type,
 )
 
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):

provinzkraut:
Thanks!

PR #4710, opened 2026-04-20
fix(serialization/msgspec_hooks): avoid dict allocation in default_serializer when no custom type_encoders

Peopl3s:
## Description

Avoid dict allocation in default_serializer when no custom type_encoders

Closes #4701

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4710">https://litestar-org.github.io/litestar-docs-preview/4710</a>

PR #4826, opened 2026-05-31
perf(serialization): optimize default_serializer to avoid unnecessary…

### LINKED ISSUE #4701: Bug: default_serializer creates a full copy of DEFAULT_TYPE_ENCODERS on every call
### Description

default_serializer is registered as the enc_hook for msgspec encoders and is called for every value that is not natively supported by msgspec during serialization. On every invocation it unconditionally merges DEFAULT_TYPE_ENCODERS with the caller-supplied type_encoders

https://github.com/litestar-org/litestar/blob/main/litestar/serialization/msgspec_hooks.py#L85
```python
def default_serializer(value: Any, type_encoders: Mapping[Any, Callable] | None = None) -> Any:
      type_encoders = {**DEFAULT_TYPE_ENCODERS, **(type_encoders or {})}  # new dict every call                                                                                                             
      for base in value.__class__.__mro__[:-1]:
          try:                                                                                                                                                                                              
              encoder = type_encoders[base]             
          except KeyError:                                                                                                                                                                                  
              continue                                                                                                                                                                                      
          else:
              return encoder(value)                                                                                                                                                                         
      raise TypeError(...)
```

What does it mean:
* Unnecessary heap allocation and GC pressure on every serialization of a non-native type
* The cost compounds in responses containing many such values
* MRO iteration also has no result cache, the same type lookup is repeated for every serialized instance   

*Suggested fix *
```python                                                                                                                                                                                            
                                                                                                                                                                                                            def default_serializer(value: Any, type_encoders: Mapping[Any, Callable] | None = None) -> Any:
    # Avoid copying DEFAULT_TYPE_ENCODERS when no custom encoders are provided
    encoders = DEFAULT_TYPE_ENCODERS if not type_encoders else {**DEFAULT_TYPE_ENCODERS, **type_encoders}        
                                                                                         
    for base in value.__class__.__mro__[:-1]:                                                                                                                                                             
        if (encoder := encoders.get(base)) is not None:                                                                                                                                                   
            return encoder(value)                                                                                                                                                                         
```

_No response_

### MCVE

```python
import cProfile                                       
from datetime import datetime
from litestar.serialization import encode_json
                                                                                                                                                                                                            @dataclass
class Record:                                                                                                                                                                                             
    id: int                                           
    created_at: datetime
    name: str
                                                                                                                                                                                                            
records = [Record(id=i, created_at=datetime.now(), name=f"item-{i}") for i in range(10_000)] 
                                                                                                             
payload = [{"id": r.id, "created_at": r.created_at, "name": r.name} for r in records]                                                                                                                     
                                                                                                                                                                                                            
# Profile encode_json with 10k datetime values        
cProfile.run("for _ in range(100): encode_json(payload)", sort="cumulative")
```

### Steps to reproduce

1. Create a Litestar handler that returns a large list of objects containing non-native types (datetime, Decimal, UUID, etc.)
2. Profile the serialization path under load using cProfile or py-spy                                                                                                                                     
3. Observe that default_serializer appears near the top of the profile with high cumulative time and allocation count 

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

v2.21.1

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

longtsing:
perf(serialization): optimize default_serializer to avoid unnecessary dict copy

Avoid creating a full copy of DEFAULT_TYPE_ENCODERS on every call when no custom type_encoders are provided. This reduces heap allocation and GC pressure during serialization of non-native types.

Also replace try/except KeyError with dict.get() for cleaner code.

Fixes #4701

 

## Description 

 

Fixes #4701

This PR optimizes the `default_serializer` function in `litestar/serialization/msgspec_hooks.py` to avoid unnecessary dictionary copying on every serialization call.

### Changes

- **Avoid unnecessary dict copy**: When no custom `type_encoders` are provided, use `DEFAULT_TYPE_ENCODERS` directly instead of creating a merged copy
- **Cleaner code**: Replace `try/except KeyError` pattern with `dict.get()` for better readability

### Performance Impact

Before this change, every call to `default_serializer` would create a new dictionary by merging `DEFAULT_TYPE_ENCODERS` with the provided `type_encoders` (or an empty dict). This caused unnecessary heap allocation and GC pressure, especially when serializing responses containing many non-native types (datetime, Decimal, UUID, etc.).

After this change, when no custom type encoders are provided, the function reuses `DEFAULT_TYPE_ENCODERS` directly, eliminating the dictionary copy overhead.

### Testing

- All existing tests pass
- The optimization is backward compatible and does not change any behavior

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4826">https://litestar-org.github.io/litestar-docs-preview/4826</a>

### DIFF
--- a/litestar/serialization/msgspec_hooks.py
+++ b/litestar/serialization/msgspec_hooks.py
@@ -82,14 +82,10 @@ def default_serializer(value: Any, type_encoders: Mapping[Any, Callable[[Any], A
     Raises:
         TypeError: if value is not supported
     """
-    type_encoders = {**DEFAULT_TYPE_ENCODERS, **(type_encoders or {})}
+    encoders = DEFAULT_TYPE_ENCODERS if not type_encoders else {**DEFAULT_TYPE_ENCODERS, **type_encoders}
 
     for base in value.__class__.__mro__[:-1]:
-        try:
-            encoder = type_encoders[base]
-        except KeyError:
-            continue
-        else:
+        if (encoder := encoders.get(base)) is not None:
             return encoder(value)
 
     raise TypeError(f"Unsupported type: {type(value)!r}")


sobolevn:
There's an existing PR for this https://github.com/litestar-org/litestar/pull/4710/changes

PR #4839, opened 2026-06-08
fix: Prevent passing an empty boundary to the parser by returning a 400

EmmanuelNiyonshuti:
### Description

Fix https://github.com/litestar-org/litestar/issues/4638

Also got some inspiration from https://github.com/Kludex/starlette/issues/1588
Django and Flask do the same. 

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4839">https://litestar-org.github.io/litestar-docs-preview/4839</a>

### DIFF
--- a/litestar/connection/request.py
+++ b/litestar/connection/request.py
@@ -24,6 +24,7 @@ from litestar.exceptions import (
     InternalServerException,
     LitestarException,
     LitestarWarning,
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):
Thanks!

PR #4846, opened 2026-06-12
fix(openapi): fix RecursionError for recursive PEP 695 type aliases in schema generation

### LINKED ISSUE #4843: Bug: RecursionError generating OpenAPI for a recursive PEP 695 `type` alias
### Description

A recursive PEP 695 `type` alias used as a handler annotation crashes OpenAPI schema generation with `RecursionError`.

`SchemaCreator.for_type_alias_type` unwraps the alias's `__value__` and recurses into it with no cycle guard, so a self-referential alias (e.g. a `JSON` type) expands forever:

https://github.com/litestar-org/litestar/blob/v2.23.0/litestar/_openapi/schema_generation/schema.py — `for_type_alias_type` calls `for_field_definition(... annotation=field_definition.annotation.__value__ ...)`.

Litestar already guards self-referential **models** (#2429 / #2869, via a registered `$ref`), but that guard was never extended to the type-alias path added in #3715. So recursive *models* work, while recursive *type aliases* do not.

Downstream I've worked around it by patching `for_type_alias_type` to track in-progress aliases and return a free-form `Schema()` on revisit, which breaks the cycle. I can open a PR upstream if a maintainer confirms the preferred shape (free-form vs. `$ref`).

_No response_

### MCVE

```python

```

### Steps to reproduce

```python

type JSON = None | bool | str | float | int | list[JSON] | dict[str, JSON]


@get("/")
async def handler() -> JSON:
    return {}


app = Litestar([handler])
app.openapi_schema.to_schema()  # RecursionError
```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

- litestar 2.23.0
- Python 3.14

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

Sujit-1509:
Fixes #4843.

## Summary

Recursive PEP 695 type aliases can currently trigger a `RecursionError` during OpenAPI schema generation.

For example:

`python
type JSON = None | bool | str | float | int | list[JSON] | dict[str, JSON]
`

When Litestar resolves the alias through `for_type_alias_type()`, it repeatedly follows `__value__` and eventually exceeds Python's recursion limit.

This change adds cycle detection for type alias resolution inside `SchemaCreator`.

When an alias is encountered while it is already being processed, the generator returns a free-form `Schema()` instead of recursively resolving the alias again. This matches Litestar's existing fallback strategy for types that lack a concrete OpenAPI representation.

## Changes

- Added `processing_type_aliases` to `SchemaCreator.__slots__` and initialized it as an empty set in `__init__`.
- Added recursion guard in `for_type_alias_type()` using `try/finally` to ensure the set is cleaned up even on exceptions.
- Added regression test `test_recursive_type_alias_type_keyword` that reproduces the exact MCVE from the issue and asserts the schema is generated correctly without a `RecursionError`.

## Testing

- Regression test: **PASS**
- Full `tests/unit/test_openapi/test_schema.py`: **51 passed**

## Why `Schema()` is correct here

A recursive type alias has no finite structural representation in OpenAPI. Returning a free-form `Schema()` (no `type` key) is interpreted as ''any valid JSON value'', which is exactly what recursive aliases like `JSON` aim to model. This matches the existing fallback used in `create_schema_for_annotation()` for unmapped types.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4846">https://litestar-org.github.io/litestar-docs-preview/4846</a>


tests/unit/test_openapi/test_schema.py
```diff
@@ -842,3 +842,31 @@ def test_decimal_schema_type() -> None:
 
     schema = create_schema_for_annotation(Decimal)
     assert schema.type == OpenAPIType.STRING
+
+
+@pytest.mark.skipif(sys.version_info < (3, 12), reason="type keyword not available before 3.12")
```
sobolevn:
```suggestion
@pytest.mark.skipif(sys.version_info < (3, 12), reason="PEP-695 aliases are not available before 3.12")
```

tests/unit/test_openapi/test_schema.py
```diff
@@ -842,3 +842,31 @@ def test_decimal_schema_type() -> None:
...
+
+    @get("/")
+    def handler() -> annotation:  # type: ignore[valid-type]
+        return None
+
+    app = Litestar([handler])
+    schema = app.openapi_schema.to_schema()
+    assert schema
+    # Verify the nested lists/dicts are resolved correctly without RecursionError
+    response_schema = schema["paths"]["/"]["get"]["responses"]["200"]["content"]["application/json"]["schema"]
```
sobolevn, line 866: Maybe you can just assert the full schema?

litestar/_openapi/schema_generation/schema.py
```diff
@@ -343,14 +344,22 @@ def for_new_type(self, field_definition: FieldDefinition) -> Schema | Reference:
...
-            FieldDefinition.from_kwarg(
-                annotation=field_definition.annotation.__value__,
-                name=field_definition.name,
-                default=field_definition.default,
-                kwarg_definition=field_definition.kwarg_definition,
+        alias_type = field_definition.annotation
+        if alias_type in self.processing_type_aliases:
+            return Schema()
+
+        self.processing_type_aliases.add(alias_type)
```
sobolevn, line 351: Can you please check an annotation like `Annotated[JSON, {}]`? Some annotations might be unhashable, so we need to check that they also work.

Sujit-1509:
@sobolevn Thanks for the review!

I've addressed the first two points:

* Updated the skip reason to `"PEP-695 aliases are not available before 3.12"`.
* Switched the test to assert the full generated schema structure instead of only checking the nested array/object cases.

For the `Annotated[JSON, {}]` suggestion, I investigated that path and added a regression test covering it.

From what I found, `FieldDefinition.from_kwarg()` unwraps the `Annotated` metadata before schema generation reaches `for_type_alias_type()`, so the recursion guard ends up tracking the underlying `TypeAliasType` rather than the metadata object itself. Because of that, I wasn't able to reproduce any hashability issues with `Annotated[JSON, {}]`.

I've pushed the additional test to document that behavior and verify it continues working.

Please let me know if you'd like me to explore any other edge cases around recursive aliases.

tests/unit/test_openapi/test_schema.py
Sujit-1509: Done!

PR #4861, opened 2026-06-16
docs(responses.headers): fix ETag example validity

Vitaly312:
### Fix ETag header examples validity:
- add `sync_to_thread=False` for non-async handlers
- register the missing `get_population_count_dynamic` handler in the App constructor
- wrap the  ETag header value in `""` in the `get_population_count_dynamic` to comply with the HTTP specification
### Add tests for docs examples:
- fully cover by tests ETag examples
- test_response_cookies.py: change type of a returned value of the mocked `randint` from str to int to match the real signature of this function

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4861">https://litestar-org.github.io/litestar-docs-preview/4861</a>


provinzkraut (approved):
Thanks!

PR #4863, opened 2026-06-16
fix(response): deduplicate cookies before returning ASGIResponse

### LINKED ISSUE #4862: Bug: Response.set_cookie does not deduplicate cookies
### Current behavior
The set_cookie method does not upsert the new cookie, just append.
```bash
❯ curl -i http://localhost:8000/
HTTP/1.1 200 OK
...
set-cookie: foo=bar; Path=/; SameSite=lax
set-cookie: foo=foo; Path=/; SameSite=lax
```
### Expected behavior
Cookies should be deduplicated to avoid redundant `set-cookie` headers in the response.
```bash
❯ curl -i http://localhost:8000/
HTTP/1.1 200 OK
...
set-cookie: foo=foo; Path=/; SameSite=lax
```

_No response_

### MCVE

```python


@get("/", sync_to_thread=False)
def test_handler() -> Response:
    return Response({}, cookies={"foo": "bar"})


async def after_request(response: Response) -> Response:
    response.set_cookie(key="foo", value="foo")
    return response


app = Litestar(
    route_handlers=[test_handler],
    after_request=after_request,
)
```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

litestar: the main branch
python: 3.12

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Vitaly312:
## Description
Fixes #4862.

Deduplicate cookies in the `Response.set_cookie` and in the `ASGIResponse` constructor to handle both local and other-scoped cookies.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4863">https://litestar-org.github.io/litestar-docs-preview/4863</a>


litestar/response/base.py
```diff
@@ -351,9 +351,7 @@ def delete_cookie(
         Returns:
             None.
         """
-        cookie = Cookie(key=key, path=path, domain=domain, expires=0, max_age=0)
-        self.cookies = [c for c in self.cookies if c != cookie]
-        self.cookies.append(cookie)
+        self.cookies.append(Cookie(key=key, path=path, domain=domain, expires=0, max_age=0))
```
provinzkraut: Why is this touching cookie deletion?

litestar/response/base.py
```diff
@@ -427,7 +425,7 @@ def to_asgi_response(
         return ASGIResponse(
             background=self.background or background,
             body=self.render(self.content, media_type, get_serializer(type_encoders)),
-            cookies=cookies,
+            cookies={c: c for c in cookies}.values(),
```
provinzkraut: I don't think this is the right place for this fix. It would better be fixed in `set_cookie`. It also doesn't handle case sensitivity, and should deduplicate cookies by name, not hash.

provinzkraut (changes requested):

litestar/response/base.py
Vitaly312: If cookie duplicates are filtered in the `to_asgi_response`, which is called after all `delete_cookie` methods, it is unnecessary to filter them before deleting.

But if you prefer to keep the public Response.cookies clean during its lifetime, I will revert this change.

litestar/response/base.py
Vitaly312: I will fix case sensitivity and deduplicate cookies by name, thanks.

Location: I chose this to filter all cookies once at the end instead of adding these checks in the constructor(which passes cookies directly), `set_cookie`, `delete_cookie` and runs it after each insertion. Besides the `Response` class provides public self.cookies which can be modified by users directly.
Additionally, collisions like 
```python
@get("/", response_cookies=[Cookie(key="my-cookie", value="1")])
def handler() -> Response:
    return Response(content=None, cookies=[Cookie(key="my-cookie", value="2")])
```
can be resolved only at the `to_asgi_response` level (by comparing self.cookies and the `cookies` param which contains cookies from the other scope). so checking in the `to_asgi_response` and deduplicating cookies in this place looks like the only acceptable option to provide uniqueness.

The downside of that approach is that during the lifecycle of Response the Response.cookies can contain duplicates. Should I change this to use the second approach?

litestar/response/base.py
Vitaly312: Done. Reverted and refactored to use the `_upsert_cookie` helper to keep the public `self.cookies` clean during its lifetime.

litestar/response/base.py
Vitaly312: I've refactored this to comply with your feedback and combined both approaches.
- both `set_cookie` and `delete_cookie` delete duplicates to keep the `Response.cookies` clean during its lifetime
- Additionally, the merged list of cookies (that combines local and other-scoped cookies) is deduplicated in the `ASGIResponse` constructor (Since filtering `documentation_only` cookies is implemented on this level and deduplication must be after that to prevent overriding real cookie by documentation one), to prevent collisions with local and other-scoped cookies as I mentioned.
- case insensitivity is implemented in Cookie's `__eq__` and `__hash__` methods.

litestar/datastructures/cookie.py
provinzkraut, line 1: These changes are are breaking

litestar/response/base.py
```diff
@@ -116,7 +116,8 @@ def __init__(
         self.body = body
         self.content_length = content_length
         self._encoded_cookies = tuple(
-            cookie.to_encoded_header() for cookie in (cookies or ()) if not cookie.documentation_only
+            cookie.to_encoded_header()
+            for cookie in {c: c for c in (cookies or ()) if not c.documentation_only}.values()
```
provinzkraut:
```suggestion
for cookie in {c.key.lower(): c for c in (cookies or ()) if not c.documentation_only}.values()
```
this would take care of de-duplicating cookies by key

provinzkraut (changes requested):

litestar/response/base.py
Vitaly312: thanks! but I think cookies with different pathes shouldn't be overridden. maybe use `(c.key.lower(), c.domain, c.path)` as a dictionary key?

Vitaly312:
@provinzkraut I've addressed your feedback. Let me know if some changes are needed

PR #4866, opened 2026-06-17
fix(response): fix cookie priority order

### LINKED ISSUE #4864: Bug: response_cookies of any scope overrides Response.cookies with the same name
### Description

Response.cookies have higher priority than response_cookies, but the `Response` class merges self.cookies behind other-scoped cookies, e.g. `itertools.chain(self.cookies, cookies)` and it leads to overriding local cookie by other-scoped one.

This bug is in the `Response` class and all derived classes that override `to_asgi_response`, e.g. `Redirect`, `Stream`, etc.

_No response_

### MCVE

```python
from litestar.datastructures import Cookie


@get("/", response_cookies=[Cookie(key="my-cookie", value="response-wrapper")], sync_to_thread=False)
def handler() -> Response:
    return Response(content=None, cookies=[Cookie(key="my-cookie", value="response")])

app = Litestar(route_handlers=[handler])

# ❯ curl -i http://localhost:8000/
# HTTP/1.1 200 OK
# ...
# set-cookie: my-cookie=response; Path=/; SameSite=lax
# set-cookie: my-cookie=response-wrapper; Path=/; SameSite=lax

# So the overridden value is 'response-wrapper'
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

litestar: main branch

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Vitaly312:
## Description
Fixes #4864 

Since the last cookie is overriding the previous one with the same name, `self.cookies` should be the last in the cookies list (after cookies from the other scope, e.g. `response_cookies`) because they have higher priority than the `to_asgi_response` `cookie` parameter which contains cookies from the other scope.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4866">https://litestar-org.github.io/litestar-docs-preview/4866</a>

### DIFF
--- a/litestar/response/base.py
+++ b/litestar/response/base.py
@@ -386,6 +386,9 @@ class Response(Generic[T]):
         except (AttributeError, ValueError, TypeError) as e:
             raise ImproperlyConfiguredException("Unable to serialize response content") from e
 
+    def _merge_cookies(self, cookies: Iterable[Cookie] | None) -> Iterable[Cookie] | None:
+        return self.cookies if cookies is None else itertools.chain(cookies, self.cookies)
+
     def to_asgi_response(
         self,
         request: Request,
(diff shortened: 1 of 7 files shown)


sobolevn (approved):

sobolevn:
Thanks! I guess that the coverage is failing because you did not test template / etc kinds of responses.

PR #4868, opened 2026-06-19
fix(channels): remove empty channel entries on unsubscribe to prevent memory leak

### LINKED ISSUE #4867: Bug: ChannelsPlugin.unsubscribe leaks empty channel entries with arbitrary_channels_allowed=True
### Summary

### Relevant code

```python
for channel in channels:
    try:
    except KeyError:
        continue
    if not channel_subscribers:
```

### Reproduction

```python
import asyncio


async def main() -> None:
    async with plugin:
        for i in range(1000):
            channel = f"user_{i}"


asyncio.run(main())
```

### Expected behavior

### Actual behavior

### Suggested fix

### Note

### Environment

- Python 3.13 / 3.14

lesnik512:
### Description

`ChannelsPlugin._channels` grows without bound when the plugin is used with
`arbitrary_channels_allowed=True` and dynamic channel names (e.g. one channel per user),
because `unsubscribe()` removed the subscriber and unsubscribed the backend but never deleted
the now-empty `set()` nor its key from `self._channels`. For long-lived applications this is a
memory leak that only a process restart clears.

This PR removes a dynamically created channel from `_channels` once its last subscriber leaves.
Channels declared in the constructor (`channels=[...]`) are preserved, since they back the
generated WebSocket route handlers and the startup backend subscription.

### Reproduction (before this PR)

```python
import asyncio

async def main() -> None:
    async with plugin:
        for i in range(1000):
            s = await plugin.subscribe([f"user_{i}"])
            await plugin.unsubscribe(s, [f"user_{i}"])
        assert len(plugin._channels) == 0  # fails: 1000

asyncio.run(main())
```

### Changes

- `ChannelsPlugin` records constructor-declared channels and, in `unsubscribe()`, deletes a
  channel key once its subscriber set is empty **only** for dynamically created channels.
- Added tests: arbitrary channel pruned after last subscriber, declared channel kept, and a
  1000-cycle no-leak check. Existing unsubscribe tests use declared channels and are unaffected.

### Notes

`RedisChannelsStreamBackend.stream_events()` keeps a `stream_ids` cursor map that exhibits the
same unbounded-growth pattern; happy to address it here in a follow-up commit or a separate PR —
let me know your preference.

### Closes

Closes #4867

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4868">https://litestar-org.github.io/litestar-docs-preview/4868</a>


litestar/channels/plugin.py
Vitaly312: This will cause KeyError if the last subscriber unsubsribes twice since key is deleted.


```suggestion
            channel_subscribers = self._channels.get(channel, set())
```

Vitaly312 (changes requested):

litestar/channels/plugin.py
lesnik512: Confirmed and fixed in c096e58d4. Reproduced with a TDD test: after the last subscriber of an arbitrary channel unsubscribes, its `_channels` entry is deleted, so a second `unsubscribe` for that channel raised `KeyError`. Switched to `self._channels.get(channel, set())` so the existing "not subscribed" branch handles the empty set.

Vitaly312:
Thanks! The code is fine, just a little nit: comments look a bit verbose, I think we can trim them since the code isn't complicated enough. What do you think?

lesnik512:
> Thanks! The code is fine, just a little nit: comments look a bit verbose, I think we can trim them since the code isn't complicated enough. What do you think?

I've compacted and removed some comments, please take a loog

litestar/channels/plugin.py
```diff
@@ -103,6 +103,9 @@ def __init__(
         self._subscriber_class = subscriber_class
 
         self._channels: dict[str, set[Subscriber]] = {channel: set() for channel in channels or []}
+        # Declared channels are kept when empty (they back route handlers); arbitrary channels
+        # are dropped on unsubscribe to avoid unbounded growth.
+        self._declared_channels: set[str] = set(channels or [])
```
Vitaly312: If channels is a generator, `_declared_channels` will be empty since it was exhausted during defining `_channels`
```suggestion
        self._channels: dict[str, set[Subscriber]] = {channel: set() for channel in channels or []}
        # Declared channels are kept when empty (they back route handlers); arbitrary channels
        # are dropped on unsubscribe to avoid unbounded growth.
        self._declared_channels: set[str] = set(self._channels)
```

Vitaly312 (changes requested):

Vitaly312 (approved):
Thanks. CI failure is unrelated to the changes(looks like some flaky test)

Vitaly312 (approved):
Approved (again?). Anyway, I don't have write access :)

PR #4872, opened 2026-06-20
fix(channels): make subscribe cancellation-safe to prevent subscriber leak

### LINKED ISSUE #4871: ChannelsPlugin.subscribe is not cancellation-safe: a subscriber is leaked into _channels if the request is cancelled mid-subscribe
### Summary

```python
# litestar/channels/plugin.py
for channel in channels:
    ...
if channels_to_subscribe:
if history:
    )
return subscriber
```

### MRE

```python
import asyncio

        self.in_history.set()


async def main() -> None:
    channel = "user:1"

asyncio.run(main())
```

Output:

```
```

### Expected vs actual

### Proposed fix

```python
if history:
channels_to_subscribe = set()
for channel in channels:
    ...
if channels_to_subscribe:
return subscriber
```

Happy to open a PR.

### Environment

- Litestar 2.21.1
- Python 3.13 / 3.14

lesnik512:
### Description

`ChannelsPlugin.subscribe` registers the new subscriber into `self._channels` **synchronously, before** it `await`s `backend.subscribe` and `put_subscriber_history`. The history fetch performs real I/O (e.g. Redis `xrevrange`), so a cancellation there raises `CancelledError` without returning the subscriber — it stays registered in `_channels`, unreachable by the caller (which never received it) and so never `unsubscribe`d, while the backend stays subscribed.

It surfaces with SSE / long-lived subscriptions on ASGI servers that cancel the request task when the client disconnects. A client that connects and disconnects while `subscribe` is still fetching history leaks a subscriber every time; under bursty connect/disconnect (reconnect storms) `_channels` and the backend's subscribed-channel set grow without bound — rising memory and, for the Redis stream backend, rising CPU as `XREAD` keeps scanning orphaned channels. Cleared only by a restart.

`start_subscription` inherited the flaw: it does `subscriber = await self.subscribe(...)` and only enters its `try/finally` (which unsubscribes) *after* `subscribe` returns, so a cancellation inside `subscribe` skipped the cleanup.

### Fix

Fetch history **before** registering the subscriber. `put_subscriber_history` is the only real cancellation window here, so doing it first means a cancellation discards an unregistered subscriber instead of leaking it. `start_subscription` needs no separate change — `subscribe` now leaves nothing behind on cancel, and there is no `await` between `subscribe` returning and the `try`, so the context manager is safe too.

### Behavior note

History is now enqueued strictly before any live event (previously live events could interleave ahead of history). The trade-off is a small window: events published between the history snapshot and `backend.subscribe` completing are captured by neither — comparable to the existing history-vs-live ordering races.

### Tests

Added `test_subscribe_cancellation_does_not_leak` and `test_start_subscription_cancellation_does_not_leak`, using a `MemoryChannelsBackend` subclass whose `get_history` suspends (standing in for any backend whose history fetch awaits I/O). Each cancels `subscribe`/`start_subscription` mid-history and asserts no subscriber is left in `_channels` and the backend is not left subscribed. Both fail on `main` and pass with the fix; the full `test_plugin.py` suite remains green. No external services required.

### Closes

Closes #4871

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4872">https://litestar-org.github.io/litestar-docs-preview/4872</a>


tests/unit/test_channels/test_plugin.py
```diff
@@ -279,6 +279,61 @@ async def test_subscribe_non_existent_channel_raises(memory_backend: MemoryChann
...
+
+    async def consume() -> None:
+        async with plugin.start_subscription([channel], history=1):
+            pass
+
+    task = asyncio.create_task(consume())
+    await backend.in_history.wait()  # entering the context manager, suspended in get_history
+
+    task.cancel()  # client disconnects mid-subscribe
+    await asyncio.gather(task, return_exceptions=True)
```
sobolevn: I suggest asserting the result here. We must have at least one exception, right?

sobolevn (commented):
(not a full review)

lesnik512:
@sobolevn Hi! Is it OK now?

PR #4883, opened 2026-06-25
docs: fix and enable type checkers for examples

Vitaly312:
## Description

Closes #3816 

1 - All errors are fixed, and corresponding settings are included in the config file to add checks into the CI.
2, 3 - seem unnecessary at the moment since all errors were fixed without any additional configurations
4 - The latest version of documentation already describes `type: ignore`. Anyway this PR reduces quantity of these comments.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4883">https://litestar-org.github.io/litestar-docs-preview/4883</a>

PR #4884, opened 2026-06-26
refactor(types): widen PrometheusConfig.buckets from list to Sequence

Vitaly312:
## Description
Part of #4883 
This PR widens the type of `PrometheusConfig.buckets` from list to Sequence to match the type of `prometheus_client.metrics.Histogram` constructor's buckets argument, which requires common `Sequence[float | str]`. 

Since list[T] is a Sequence[T], this change is backward compatible.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4884">https://litestar-org.github.io/litestar-docs-preview/4884</a>


sobolevn (approved):
Thanks!

PR #4885, opened 2026-06-26
fix(jwt): auto-generate jti when token_unique_jwt_id is not provided 

### LINKED ISSUE #2118: Revoke jwt token
### Summary

Provide a mechanism to revoke a jwt token on /logout similar to `flask-jwt-extended` (https://flask-jwt-extended.readthedocs.io/en/stable/blocklist_and_token_revoking.html).

### Basic Example

_No response_

### Drawbacks and Impact

_No response_

### Unresolved questions

_No response_

Yadidiah-k:
Fixes #2118

Previously, create_token() only generated a jti when token_unique_jwt_id was explicitly passed. If callers omitted it, jti was None, silently breaking token revocation.

Changes:
- auth.py: Changed token_unique_jwt_id if token_unique_jwt_id is not None else uuid4().hex to token_unique_jwt_id or uuid4().hex so empty strings and None both trigger auto-generation
- test_auth.py: Updated assertions to use a truthiness check instead of is not None, matching the new behavior

Testing: Updated existing hypothesis tests — 85 tests pass.

### DIFF
--- a/litestar/security/jwt/auth.py
+++ b/litestar/security/jwt/auth.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from dataclasses import asdict, dataclass, field
 from datetime import UTC, datetime, timedelta
 from typing import TYPE_CHECKING, Any, Generic, Literal, cast
+from uuid import uuid4
 
 from typing_extensions import TypeVar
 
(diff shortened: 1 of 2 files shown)


vineoy:
Verified this on current `main`.
`create_token(identifier="u1")` produces `jti=None`, so revocation by `jti` cannot work unless a `token_unique_jwt_id` is explicitly supplied. The direction in this PR is correct.

Also, using:

```python
token_unique_jwt_id or uuid4().hex
```

is preferable to `is not None` because an empty-string `token_unique_jwt_id` should also result in a generated ID.

A few things that look worth addressing before merge:

* rebase onto current `main` — the PR is currently 52 commits behind and `auth.py`, `middleware.py`, and `token.py` have changed since the PR base
* clean up the trailing whitespace / extra blank line in the test diff
* consider an explicit uniqueness regression test: two calls to `create_token()` should produce different `jti` values
* consider documenting that `jti` is now generated automatically when omitted

I reproduced the baseline locally and got 94 tests passing in the JWT auth test suite.

Happy to help with testing/docs follow-up rather than duplicating the core implementation.

PR #4887, opened 2026-06-30
fix: KeyError in LoggingMiddleware when a request fails before a response body is sent

### LINKED ISSUE #4855: Bug: Logging middelware bug when a middleware raises an exception.
### Description

I'm using the main branch in production in several projects and I've been noticing an exception appearing in my logs that trace back to the logging middleware. This is probably a regression because I was on older commit and things were working fine.

Perhaps I'm writing middelwares incorrectly and such a design shouldn't be supported but here is what's happening: if an inner middleware (e.g. a custom auth middleware) raises an exception then the logging middleware will raise `KeyError` at line https://github.com/litestar-org/litestar/blob/af445635e0d812deacc5196cdf8710c47f87abc3/litestar/middleware/logging.py#L226 . This happens because in the normal flow of the request/response, the logging context dictionary doesn't get populated with a field called 'http.response.body' (a.k.a HTTP_RESPONSE_BODY).

The following test has been generated with the help of an LLM:

```python
def test_logging_middleware_no_keyerror_when_auth_middleware_raises_before_send(caplog: "LogCaptureFixture") -> None:
    """Regression test: LoggingMiddleware must not raise KeyError when an HTTPException
    is raised from an inner middleware before any ASGI send call (e.g. an auth middleware
    that rejects the request outright without sending a response body).

    Before the fix, the except-HTTPException branch in LoggingMiddleware.handle() set
    HTTP_RESPONSE_START in log_context but never set HTTP_RESPONSE_BODY, so
    extract_response_data() would KeyError on the .pop() and the whole request would
    surface as 500 instead of the intended 401.
    """

    class AlwaysRejectAuthMiddleware(AbstractAuthenticationMiddleware):
        async def authenticate_request(self, connection: ASGIConnection) -> AuthenticationResult:
            raise NotAuthorizedException("Token expired")

    @get("/protected")
    def protected_handler() -> dict:
        return {"secret": "data"}

    with (
        create_test_client(
            route_handlers=[protected_handler],
            middleware=[
                LoggingMiddleware(
                    "litestar.test",
                    response_log_fields=["status_code"],
                    request_log_fields=["path"],
                ),
                DefineMiddleware(AlwaysRejectAuthMiddleware),
            ],
        ) as client,
        caplog.at_level(INFO),
    ):
        response = client.get("/protected")
        # Before the fix this was 500 because KeyError propagated through the middleware stack
        assert response.status_code == 401
        assert any("status_code=401" in msg for msg in caplog.messages)
```

The LLM also suggested a fix by setting a default value for the field in after lines https://github.com/litestar-org/litestar/blob/af445635e0d812deacc5196cdf8710c47f87abc3/litestar/middleware/logging.py#L133 and https://github.com/litestar-org/litestar/blob/af445635e0d812deacc5196cdf8710c47f87abc3/litestar/middleware/logging.py#L140, happy to create a PR

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

main branch

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

Sreekant13:
## Description

Fixes #4855.

When an inner middleware (e.g. an auth middleware) raises an HTTPException before any response body is sent, `LoggingMiddleware`'s exception branches set `HTTP_RESPONSE_START` but never `HTTP_RESPONSE_BODY`. `extract_response_data()` then does `log_context.pop(HTTP_RESPONSE_BODY)`, raising `KeyError` — so the intended 401 surfaces
as a 500.

This synthesizes the missing response messages (an empty body, plus empty headers on the start message) in both exception branches so the extractor has what it expects. Added a regression test that reproduces the KeyError-as-500 and confirms the correct status after the fix.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4887">https://litestar-org.github.io/litestar-docs-preview/4887</a>

### DIFF
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -130,14 +130,21 @@ class LoggingMiddleware(ASGIMiddleware):
             # Log response with the correct status code from the exception
             if self.response_log_fields:
                 scope_state = ScopeState.from_scope(scope)
-                scope_state.log_context[HTTP_RESPONSE_START] = {"status": exc.status_code}
+                # The request failed before a response was sent, so synthesize the
+                # response messages the extractor expects. Without a HTTP_RESPONSE_BODY
+                # entry, extract_response_data() would raise a KeyError. See #4855.
+                scope_state.log_context[HTTP_RESPONSE_START] = {"status": exc.status_code, "headers": []}
+                scope_state.log_context[HTTP_RESPONSE_BODY] = {"body": b""}
                 self.log_response(scope=scope)
             raise
         except Exception:
             # Log response with 500 status for unhandled exceptions
             if self.response_log_fields:
                 scope_state = ScopeState.from_scope(scope)
-                scope_state.log_context[HTTP_RESPONSE_START] = {"status": HTTP_500_INTERNAL_SERVER_ERROR}
+                # See the HTTPException branch above: synthesize the response messages so
+                # extract_response_data() doesn't raise a KeyError. See #4855.
+                scope_state.log_context[HTTP_RESPONSE_START] = {"status": HTTP_500_INTERNAL_SERVER_ERROR, "headers": []}
+                scope_state.log_context[HTTP_RESPONSE_BODY] = {"body": b""}
                 self.log_response(scope=scope)
             raise
 
--- a/litestar/middleware/logging.py
+++ b/litestar/middleware/logging.py
@@ -23,6 +23,11 @@ from litestar.utils.scope.state import ScopeState
 
 __all__ = ("LoggingMiddleware",)
 
+# Key under which ``LoggingMiddleware`` records, in ``ScopeState.log_context``, the status code of an
+# exception it caught. Its presence tells ``extract_response_data()`` that the request ended early, so
+# any response message the send wrapper never got to record has to be filled in.
+CAUGHT_EXCEPTION_STATUS_CODE = "caught_exception_status_code"
+
 
 if TYPE_CHECKING:
     import logging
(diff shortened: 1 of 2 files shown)


Sreekant13:
Gentle nudge on this whenever the team has capacity. It fixes a KeyError that turns an intended 401 into a 500, with a regression test. Thanks for your work on Litestar.

litestar/middleware/logging.py
```diff
@@ -130,14 +130,21 @@ async def handle(self, scope: Scope, receive: Receive, send: Send, next_app: ASG
             # Log response with the correct status code from the exception
             if self.response_log_fields:
                 scope_state = ScopeState.from_scope(scope)
-                scope_state.log_context[HTTP_RESPONSE_START] = {"status": exc.status_code}
+                # The request failed before a response was sent, so synthesize the
+                # response messages the extractor expects. Without a HTTP_RESPONSE_BODY
+                # entry, extract_response_data() would raise a KeyError. See #4855.
+                scope_state.log_context[HTTP_RESPONSE_START] = {"status": exc.status_code, "headers": []}
+                scope_state.log_context[HTTP_RESPONSE_BODY] = {"body": b""}
```
provinzkraut: It's not guaranteed that no request body has been set, so this could silently swallow things. Maybe a better fix would be to:

1. Note in `ScopeState.log_context` that an exception occurred and was caught by the logging middleware
2. Catch the missing body in `extract_response_data` *if* we know an exception occurred and was caught by the logging middleware

Sreekant13:
Thanks for the review @provinzkraut, and apologies for the slow turnaround. You were right that the original approach could swallow real data, so I have reworked it along the lines you suggested.

`LoggingMiddleware` now records the caught exception's status code in `ScopeState.log_context` under a `caught_exception_status_code` key, and `extract_response_data()` reads that marker. When it is present, the missing response messages are filled in with `setdefault`, so anything the send wrapper already recorded is kept as it was. Nothing is overwritten any more.

The concrete difference: a stream that sends a 200 and then fails partway through used to be logged as a 500, because both messages were replaced unconditionally. It is now logged as the 200 that actually reached the client.

I added `test_logging_middleware_exception_does_not_overwrite_a_started_response` to cover exactly that, and it fails against the previous version of this branch, so it guards the behaviour you flagged. The original early-failure test now also asserts the logged status code rather than only the response status.

The branch is rebased onto current main, so it is no longer behind.

Tests run, on Python 3.12:

- `pytest tests/unit/test_middleware/test_logging_middleware.py`: 17 passed
- `pytest tests/unit/test_middleware/`: 782 passed
- `ruff check` and `ruff format --check` on both changed files: clean

I could not run mypy, since it is not installed in my environment.

Sreekant13:
Thank you, and apologies for the slow reply. This deserved an answer three weeks ago.

You are right, and the objection lands. `extract_response_data` swallowing a missing body cannot tell "no body was ever set" apart from "a body was set and something went wrong", so it hides the second case. That is worse than the bug it fixes.

Your shape is better, and I would like to implement it as you described: record on `ScopeState.log_context` that an exception was raised and caught by the logging middleware, then have `extract_response_data` handle the missing body only when that flag is set. Narrow, and it cannot mask an unrelated failure.

Two things I would rather get right than guess at:

1. Where the flag should live on `log_context`, whether that is a dedicated key or something you already set there.
2. Whether an exception that the middleware catches and then raises again should still set it. I would say yes, since the body is equally absent either way, but you know the intent behind that path.

If you confirm those two, I will rework the PR. If you would rather this were closed and reopened as a fresh change against your design, that is fine as well.

Sreekant13:
Note on the red check, in case it is masking this PR's state: the failing `Test server integration` job is a repo-wide collection error, not something this change introduces. Click 9 now raises on `click.utils.get_binary_stream`, and with `filterwarnings=error` that fails collection of `tests/unit/test_cli`, `tests/e2e/test_advanced_alchemy.py` and `tests/unit/test_repository/test_generic_mock_repository.py` before any test runs. This PR only touches `LoggingMiddleware` and its own tests, none of which are in that set.

The rework from the earlier review has been in since c58cb8e (records `CAUGHT_EXCEPTION_STATUS_CODE` in `log_context` and lets `extract_response_data()` `setdefault` only what is missing, so a partially sent response keeps its real status). Glad to rebase if the base has moved; it is ready whenever there is a cycle for another look.

Sreekant13:
@provinzkraut an update rather than a ping: I synced this branch to `main`, and the full CI is green now. The `Test server integration` job that was red on my earlier note was a repo-wide collection error, not something this change introduced, and it resolves on current main. The fix and its regression tests are unchanged from your last review round. No rush, just flagging that it is ready whenever you have a window.

PR #4896, opened 2026-07-05
fix(signature): resolve litestar.di names in TYPE_CHECKING handler annotations

### LINKED ISSUE #4870: Bug: Signature types issue, NameError
### Description

`signature_types` does not make a `TYPE_CHECKING`-only name resolvable, even for
a bare-name annotation (in some cases?) and will result in a `NameError`.

I noticed this while upgrading Litestar and porting a codebase to Python 3.14.

### MCVE

requirements for this example: `uv pip install litestar[sqlalchemy,standard] aiosqlite`

`main.py`
```python
import uvicorn
from litestar import Litestar
from litestar.di import NamedDependency
from litestar.plugins.sqlalchemy import SQLAlchemyAsyncConfig, SQLAlchemyInitPlugin

from handler import handler

config = SQLAlchemyAsyncConfig(connection_string="sqlite+aiosqlite:///async.sqlite")
plugin = SQLAlchemyInitPlugin(config=config)
app = Litestar(
    route_handlers=[handler],
    plugins=[plugin],
    signature_types=[
        AsyncSession,
        NamedDependency,
    ],
)

uvicorn.run(app)
```

`handler.py`
```python

from litestar import get

if TYPE_CHECKING:
    from litestar.di import NamedDependency


@get("/")
async def handler(
    db_session: NamedDependency[AsyncSession],  # noqa: ARG001
) -> None:
    return
```

### Steps to reproduce

1. Run main.py
2. See NameError: `NameError: name 'NamedDependency' is not defined`

### Litestar Version

2.24.0

ritsth:
## Description

Fixes #4870

`NamedDependency` used in a handler annotation behind `if TYPE_CHECKING:` raises `NameError`. Litestar's forward-ref namespace (in `litestar/utils/signature.py`) is built from `params`, `datastructures`, `connection` and `types`, but not `di`, so the name never resolves. `signature_types` doesn't help either, since it keys by `__name__` and `NamedDependency` is `Annotated[T, Dependency(...)]` whose `__name__` is `"Annotated"`.

`di` was left out because it imports from `signature.py` (circular import). I added a small lazily-imported, cached helper that merges `di`'s public names into the namespace, so `NamedDependency` resolves like `params`/`datastructures` already do.

Added a regression test that fails on `main` with the reported `NameError` and passes with the fix. Tested on Python 3.12 and 3.14.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4896">https://litestar-org.github.io/litestar-docs-preview/4896</a>

### DIFF
--- a/litestar/utils/signature.py
+++ b/litestar/utils/signature.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import sys
 import typing
 from dataclasses import dataclass, replace
-from functools import lru_cache
 from inspect import Signature, getmembers, isclass, ismethod
 from itertools import chain
 from typing import TYPE_CHECKING, Annotated, Any, Self, Union, get_args, get_origin, get_type_hints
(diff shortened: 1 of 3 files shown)


litestar/utils/signature.py
```diff
@@ -151,6 +155,17 @@ def get_fn_type_hints(fn: Any, namespace: dict[str, Any] | None = None) -> dict[
     # inspect the underlying function for methods
     if hasattr(fn_to_inspect, "__func__"):
         fn_to_inspect = fn_to_inspect.__func__  # pyright: ignore[reportFunctionMemberAccess]
+
+    if "NamedDependency" not in _GLOBAL_NAMES:
```
provinzkraut: Mhm. Is there no other way to avoid the circular import, perhaps by making a different import lazy, one that's not inside a function that gets called that frequently?

ritsth: Good point!

PR #4899, opened 2026-07-07
fix(channels): prevent subscriber leak on cancellation, empty channel entry leak, and concurrent subscribe/unsubscribe race (Fixes #4871, #4867, #4894)

### LINKED ISSUE #4867: Bug: ChannelsPlugin.unsubscribe leaks empty channel entries with arbitrary_channels_allowed=True

### LINKED ISSUE #4871: ChannelsPlugin.subscribe is not cancellation-safe: a subscriber is leaked into _channels if the request is cancelled mid-subscribe

### LINKED ISSUE #4894: Bug(channels): concurrent subscribe/unsubscribe race condition causes silent event loss
### Description

Concurrent subscribing and unsubscribing of the last subscriber for the same channel leads to unsubscribing the plugin from the backend for this channel and events are suppressed for all subscribers.

_No response_

### MCVE

```python
import asyncio
from asyncio import Event

from anyio import fail_after

        network_lag.set()
        await second_start_subscription.wait()
        return await super().unsubscribe(channels)


network_lag = Event()
second_start_subscription = Event()


async def test_phantom_channel():
    channels = ChannelsPlugin(arbitrary_channels_allowed=True, backend=StubMemoryChannelBackend())

    async def concurrent_subscribe():
        await network_lag.wait()
        second_start_subscription.set()
        subscriber = await channels.subscribe("42")
        await channels.wait_published("42", "42")
        try:
            with fail_after(0.1):
                await anext(subscriber.iter_events())
        except TimeoutError as e:
            raise ValueError("Expected event not received") from e

    async with channels:
        first_subscriber = await channels.subscribe("42")
        await asyncio.gather(concurrent_subscribe(), channels.unsubscribe(first_subscriber, "42"))


asyncio.run(test_phantom_channel())  # ValueError: Expected event not received
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

litestar: main branch
python: 3.12

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

isheng-eqi:
## Summary

Three related fixes in ChannelsPlugin covering concurrency safety and resource cleanup:

1. **#4894 — Concurrent subscribe/unsubscribe race**: Added an asyncio.Lock to make channel state inspection + backend calls atomic
2. **#4871 — Cancellation-safe subscribe**: Reordered subscribe() so history fetch happens before registration; if cancelled during the fetch, the subscriber is discarded without leaking
3. **#4867 — Empty channel entry leak**: unsubscribe() now deletes the channel key from self._channels when the last subscriber leaves, preventing unbounded memory growth

## Changes

- **litestar/channels/plugin.py**: Initialize self._lock, protect critical sections, reorder subscribe for cancellation safety, delete empty channel entries
- **tests/unit/test_channels_concurrent.py**: 3 regression tests using MCVEs from the issues

## Test Plan

- [x] test_concurrent_subscribe_unsubscribe_does_not_lose_events (#4894)
- [x] test_unsubscribe_cleans_empty_channel_entries (#4867)
- [x] test_subscribe_cancellation_does_not_leak_subscriber (#4871)
- [x] ruff check + ruff format pass

Closes #4871, Closes #4867, Closes #4894


provinzkraut:
@isheng-eqi there is already an open PR for this, which is linked in the issue, same with the other PR you've opened. Since this PR appears to be fully automated, I'm closing this per our AI policy and banning you from future contributions.

PR #4917, opened 2026-07-12
fix(dto): use declared dto as return_dto fallback over serialization plugin

### LINKED ISSUE #4906: Bug: Controller `dto` is not used as fallback for response serialization when `return_dto` is omitted
### Summary

According to the documentation, when `return_dto` is not provided, the `dto` should also be used to serialize the handler's return value.

However, this does not seem to happen when using `SQLAlchemyDTO` from Advanced Alchemy.

Documentation:

https://docs.litestar.dev/2/usage/dto/0-basic-use.html#dto-layer-parameters

> If no `return_dto` is declared on the handler, `dto` is also used to encode the return data.

---

### Versions

- Litestar: 2.24.0
- Advanced Alchemy: 1.11.0
- SQLAlchemy: 2.0.51
- Python: 3.14.6

---

### Minimal example

```python
from litestar import Controller, get
from advanced_alchemy.extensions.litestar import (
    SQLAlchemyDTO,
    SQLAlchemyDTOConfig,
)


class UserDTO(SQLAlchemyDTO[User]):
    config = SQLAlchemyDTOConfig(
        exclude={
            User.id,
            User.code,
        }
    )


    path = "/users"
    dto = UserDTO

    @get()
    async def list_users(self) -> list[User]:
        return await service.list()
```

---

### Expected behavior

The response should be serialized using UserDTO, excluding the configured fields, because no return_dto is explicitly provided.

### Actual behavior

The response is serialized with all model fields. If I explicitly set:

```python
return_dto = UserDTO
```

everything works as expected.

### Additional information

Printing the controller attributes shows:

```python
print(UserController.dto)
print(UserController.return_dto)
```

Output:

```python
<class '__main__.UserDTO'>
<member 'return_dto' of 'Controller' objects>
```

---

This makes me wonder whether the documented fallback from dto to return_dto is not being applied.
Could this be a regression, or is the documentation outdated for SQLAlchemyDTO?

If this behavior belongs to Advanced Alchemy rather than Litestar, I'd appreciate it if you could point me to the appropriate repository.

Thank you!

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.24.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

isaka1022:
## Description

Fixes #4906.

The docs state that when a handler declares `dto` but no `return_dto`, the `dto`
is also used to serialize the return value:

> Additionally, if no `return_dto` is declared on the handler, this [the `dto`] will also be used to encode the return data for the handler.

This fallback silently stops working once a serialization plugin is registered
for the return type. The reporter hit it with `SQLAlchemyDTO` (Advanced
Alchemy), whose `SQLAlchemySerializationPlugin` supports SQLAlchemy models: the
configured `dto` (with its `exclude`) was ignored and every field was
serialized.

### Root cause

`BaseRouteHandler._resolve_return_dto` (`litestar/handlers/base.py`) resolved the
return DTO in this order:

1. explicit `return_dto` on the handler,
2. a DTO auto-generated by a serialization plugin that supports the return type,
3. fall back to the declared `dto` (`data_dto`).

Step 2 preempted step 3. When a serialization plugin matched the return type, it
generated a fresh, unconfigured DTO and the declared `dto` was never consulted.
This is why the fallback works for plain DTOs (no serialization plugin matches,
so the `else` branch reaches `data_dto`) but not for `SQLAlchemyDTO` (the
serialization plugin matches and shadows `data_dto`).

### Fix

Consult the declared `dto` before serialization plugins when it supports the
return type, so an explicitly configured DTO takes precedence over a
plugin-generated one. Plugin generation is preserved for the cases it is meant
to cover: when no `dto` is declared, or when the declared `dto` does not support
the return type.

The change is two lines in `_resolve_return_dto` (a new `elif` branch); the final
`is_supported_model_type_field` guard is unchanged.

## How was this tested

Reproduction mirroring the issue (Advanced Alchemy `SQLAlchemyDTO` +
`SQLAlchemySerializationPlugin`, `dto` declared on the controller,
`return_dto` omitted, `exclude={"id", "code"}`):

- Before: `GET /users` -> `[{"id": 1, "code": "secret", "name": "alice"}]`
  (excluded fields leaked).
- After: `GET /users` -> `[{"name": "alice"}]` (fallback applied).

Added a regression test in `tests/unit/test_dto/test_integration.py`
(`test_data_dto_used_as_return_dto_fallback_over_serialization_plugin`) that uses
an in-repo serialization plugin returning an unconfigured DTO for the return
type, with a declared `dto` that excludes a field. It asserts the excluded field
is absent from the response. The test fails on `main` (leaks the excluded field)
and passes with this change.

- New test passes; the existing `tests/unit/test_dto`, `tests/unit/test_plugins`,
  and `tests/unit/test_handlers` suites remain green.
- `pre-commit run --files ...`, `mypy`, and `pyright` pass on the changed files.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4917">https://litestar-org.github.io/litestar-docs-preview/4917</a>


provinzkraut (approved):
Thanks!

Issue #4869, opened 2026-06-19
Bug: `AllowedHostsMiddleware` rejects valid hosts when the `Host` header includes a port


fmoga:
### Description

`AllowedHostsMiddleware` used to strip the port from the `Host` header before matching it against the allowed-hosts regex. As of 2.22.0 it matches the **full, un-stripped** header, so a request with `Host: localhost:8000` no longer matches an allowed host of `localhost` and is rejected with `400 - invalid host header`.

The host check works correctly in production behind a proxy (where `Host` is the bare domain with no port), which is likely why this has gone unnoticed — it only manifests when the `Host` header carries a port.

### Root cause

The port-strip was lost in [`c4189610` ("fix: fix typing after merge")](https://github.com/litestar-org/litestar/commit/c4189610), a follow-up to the intentional `x-forwarded-host` removal. The walrus now binds `host` to the **un-stripped** header value; `host.split(":")[0]` survives only as the truthiness guard on the left of the `and`, while `fullmatch(host)` receives the value with the port still attached:

```python
# before (2.21.x) — port stripped, then matched
if host := headers.get("host").split(":")[0]:
    if self.allowed_hosts_regex.fullmatch(host):

# after (2.22.0+) — `host` keeps the port; the strip only gates the `if`
if (host := headers.get("host")) is not None and host.split(":")[0]:
    if self.allowed_hosts_regex.fullmatch(host):   # matches "localhost:8000"
```

### Suggested fix

Match against the port-stripped value while keeping the `None` guard, e.g.:

```python
if (raw_host := headers.get("host")) is not None and (host := raw_host.split(":")[0]):
    if self.allowed_hosts_regex.fullmatch(host):
```

### MCVE

```python
from litestar import Litestar, get
from litestar.config.allowed_hosts import AllowedHostsConfig
from litestar.testing import TestClient


@get("/")
async def handler() -> str:
    return "ok"


app = Litestar(
    route_handlers=[handler],
    allowed_hosts=AllowedHostsConfig(allowed_hosts=["localhost"]),
)


def test_host_with_port_is_allowed() -> None:
    # base_url drives the Host header -> "localhost:8000"
    with TestClient(app=app, base_url="http://localhost:8000") as client:
        response = client.get("/")
    assert response.status_code == 200  # passes on 2.21.x, fails (400) on >= 2.22.0
```

### Expected behavior
[...]

PR #4922, opened 2026-07-14
fix(middleware): restore host-port stripping in AllowedHostsMiddleware

chuenchen309:
## What

A request whose `Host` header includes a port (e.g. `example.com:8080`) is incorrectly rejected with 400 by `AllowedHostsMiddleware`, even when the bare hostname (`example.com`) is present in `allowed_hosts`.

## Root cause

Commit c4189610b ("fix: fix typing after merge") changed:
```python
```
into:
```python
```
The walrus assignment now binds `host` to the **full** header value (including any port), and `host.split(":")[0]` is only used for a truthiness check — its result is discarded. So `self.allowed_hosts_regex.fullmatch(host)` downstream still receives the port-suffixed value, which never matches a bare-hostname regex pattern.

## Fix

Rebind `host` via a second walrus assignment so the stripped (port-free) value is what's actually used for matching:
```python
if (host := headers.get("host")) is not None and (host := host.split(":")[0]):
```

## How I verified this

- Traced the regression to c4189610b via `git log -p --follow -- litestar/middleware/allowed_hosts.py`.
- Reproduced standalone with a minimal ASGI script directly invoking `AllowedHostsMiddleware`.
- Added a parametrize case to the existing `test_middleware_allowed_hosts` test (`("http://x.example.com:8080", None, HTTP_200_OK)`); confirmed it fails on the current code (`assert 400 == 200`) before the fix.
- Applied the one-line fix, confirmed all 14 tests in `test_allowed_hosts_middleware.py` pass.
- Ran the full `tests/unit/test_middleware/` suite: 776 passed, no regressions.
- `mypy litestar/middleware/allowed_hosts.py` — clean.
- `pre-commit run` on changed files — clean (ruff check/format, unasyncd, typos).

## Disclosure

I used AI assistance (Claude) to help investigate and draft this fix, but I personally reviewed the root-cause trace, ran the repro and full test suite myself, and reviewed the final diff before submitting. Happy to answer any questions about the change.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4922">https://litestar-org.github.io/litestar-docs-preview/4922</a>

### DIFF
--- a/litestar/middleware/allowed_hosts.py
+++ b/litestar/middleware/allowed_hosts.py
@@ -64,7 +64,7 @@ class AllowedHostsMiddleware(AbstractMiddleware):
             return
 
         headers = MutableScopeHeaders(scope=scope)
-        if (host := headers.get("host")) is not None and host.split(":")[0]:
+        if (host := headers.get("host")) is not None and (host := host.split(":")[0]):
             if self.allowed_hosts_regex.fullmatch(host):
                 await self.app(scope, receive, send)
                 return
--- a/tests/unit/test_middleware/test_allowed_hosts_middleware.py
+++ b/tests/unit/test_middleware/test_allowed_hosts_middleware.py
@@ -74,6 +74,7 @@ def test_allowed_hosts_middleware_redirect_regex() -> None:
         ("http://moisheAzuchmir.com", None, HTTP_400_BAD_REQUEST),
         ("http://x.moishe.zuchmir.com", None, HTTP_400_BAD_REQUEST),
         (None, "x.example.com", HTTP_400_BAD_REQUEST),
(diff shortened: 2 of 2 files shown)


chuenchen309:
Thanks for the review time on this — I found while doing some further digging that #4873 already fixes this exact regression (referencing #4869), opened back on 2026-06-20, well before my independent discovery of it today. Since it's the same root cause and fix, I'll close this one in favor of the earlier PR rather than duplicate the review effort. Apologies for not catching the existing PR before submitting — will be more careful to cross-check open PRs against the exact file path before starting work in the future.

provinzkraut:
> Thanks for the review time on this

There was no review

PR #4926, opened 2026-07-14
fix(stores): MemoryStore/FileStore.exists() ignores expiry

chuenchen309:
## What

`exists()` checked only key/file presence, unlike `get()` which correctly treats an expired entry as gone (and lazily removes it). Since expired entries are only cleared on access rather than eagerly, a key that has passed its `expires_in` remains "present" in the backing dict/filesystem until something touches it — so `exists()` reported `True` for a key that `get()` would return `None` for, an inconsistency between two methods of the same `Store` interface:

```python
store = MemoryStore()
await store.set("foo", b"bar", expires_in=timedelta(milliseconds=10))
await asyncio.sleep(0.05)
await store.exists("foo")  # True  <- wrong, the key has expired
await store.get("foo")     # None  <- correct
```

Same bug in `FileStore.exists()` (checks `path.exists()` without loading/checking the stored object's expiry).

## Fix

Give `exists()` the same load-check-expire logic `get()` already uses (minus the `renew_for` handling, which only makes sense for `get()`).

## Testing

Added `test_exists_expired` to `tests/unit/test_stores.py`, using the existing parametrized `store` fixture (covers all four backends) and the same `frozen_datetime`/Redis-`expire`-call pattern as the existing `test_expires` test. Confirmed via `git stash` that it fails against the pre-fix code (`assert True is False`) for `memory_store` and `file_store`, and passes after.

- `pytest tests/unit/test_stores.py -k "memory_store or file_store"` — 96 passed, no regressions. (`redis_store`/`valkey_store` variants error on setup/teardown in my sandbox — no Docker daemon available — identically for the pre-existing `test_exists[redis_store]` test, confirming this is an environment limitation, not something introduced by this change.)
- `ruff check` clean, `mypy` clean on the changed files.

## AI disclosure

Developed with AI assistance (Claude Code), which located the inconsistency while comparing `Store` implementations' `exists()` against their `get()` counterparts. I reviewed the diff, independently reproduced the stale-`True` bug and the fix by executing both against the actual classes, and ran the tests/linters/type-checker myself before opening this PR.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4926">https://litestar-org.github.io/litestar-docs-preview/4926</a>


hasansezertasan (approved):

litestar/stores/memory.py
```diff
@@ -104,7 +104,17 @@ async def delete_expired(self) -> None:
 
     async def exists(self, key: str) -> bool:
         """Check if a given ``key`` exists."""
-        return key in self._store
+        async with self._lock:
```
provinzkraut: Why not just call `.get` here, instead of replicating all the `.get` logic?

litestar/stores/file.py
```diff
@@ -177,7 +177,16 @@ async def delete_expired(self) -> None:
     async def exists(self, key: str) -> bool:
         """Check if a given ``key`` exists."""
         path = self._path_from_key(key)
-        return await path.exists()
+        storage_obj = await self._load_from_path(path)
```
provinzkraut: Same comment: Why not just use `.get`?

provinzkraut (changes requested):

litestar/stores/memory.py
chuenchen309: You're right — done in 44acf11c, for both this and `file.py`. `exists()` is now just `await self.get(key) is not None` in each, and all the duplicated expiry/unlink logic is gone.

One detail worth flagging since it's the only place the two aren't equivalent: the code I removed used `if not storage_obj` while `get()` returns the value itself, so truth-testing it would make a stored empty value (`b""`) report as missing. Hence `is not None` rather than a plain boolean check — with a test (`test_exists_empty_value`) pinning that, otherwise it's an easy thing to "simplify" back later.

Sorry for the slow turnaround on this one.

provinzkraut (approved):

litestar/stores/memory.py
provinzkraut: @chuenchen309 Please do not use AI to write review comments on your behalf. It is in direct violation of our AI policy.

PR #4927, opened 2026-07-15
fix(datastructures): extend_header_value only reads the first duplicate header entry, dropping the rest

chuenchen309:
## Description

`extend_header_value()` used `self.get(key)` to read the existing value before joining in the new one. `MutableMapping.get()` is backed by `__getitem__`, which returns only the **first** raw header entry matching `key` — but a header can legitimately appear as multiple separate raw entries in `self.headers` (see `add()`'s own docstring: "This method keeps duplicates"), not just as one comma-joined value.

**Concrete failure**: add two raw `vary` entries (`"accept"`, `"accept-encoding"`), then `extend_header_value("vary", "user-agent")` — the result silently drops `"accept-encoding"` entirely, leaving only `"accept,user-agent"`.

## Fix

Use `getall()` instead of `get()` to collect ALL duplicate entries, flattening each entry's own comma-separated list (matching the existing single-entry behavior) before joining in the new value. `getall()` currently raises `KeyError` when the header isn't present at all (a separate falsy-default bug also present in this method, which I've already reported/fixed independently in #4924) — catching that `KeyError` here keeps this fix correct regardless of merge order between the two PRs.

## Testing

- Added `test_mutable_scope_headers_extend_header_value_preserves_duplicate_entries`, asserting all raw duplicate entries survive `extend_header_value`.
- Confirmed red→green: reverting only `headers.py` reproduces `['bar,qux'] != ['bar,baz,qux']` (the middle duplicate entry silently dropped); reapplying passes.
- Full `tests/unit/test_datastructures/`: 148 passed.
- `ruff check` and `mypy` — clean.
- xref: no open PR touches `datastructures/headers.py`.

## AI disclosure

Developed with AI assistance (Claude Code), which located the `get()`-vs-`getall()` inconsistency while comparing `extend_header_value` against the multi-entry header-add contract documented on `add()`. I reviewed the diff, independently reproduced the data-loss bug and verified the fix by executing both against the actual class, and ran the tests/linter/type-checker myself before opening this PR.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4927">https://litestar-org.github.io/litestar-docs-preview/4927</a>

### DIFF
--- a/litestar/datastructures/headers.py
+++ b/litestar/datastructures/headers.py
@@ -188,9 +188,13 @@ class MutableScopeHeaders(MutableMapping):
         Returns:
             None
         """
-        existing = self.get(key)
-        if existing is not None:
-            value = ",".join([*existing.split(","), value])
+        try:
+            existing_entries = self.getall(key)
+        except KeyError:
+            existing_entries = []
+        existing_values = [v for existing in existing_entries for v in existing.split(",")]
+        if existing_values:
+            value = ",".join([*existing_values, value])
         self[key] = value
 
     def __getitem__(self, key: str) -> str:
--- a/tests/unit/test_datastructures/test_headers.py
+++ b/tests/unit/test_datastructures/test_headers.py
@@ -195,6 +195,14 @@ def test_mutable_scope_headers_extend_header_value_preserves_duplicate_entries(
     assert mutable_headers.getall("foo") == ["bar,baz,qux"]
 
 
+def test_mutable_scope_headers_from_tuple_extend_header_value_preserves_duplicate_entries(
+    mutable_headers_from_tuple: MutableScopeHeaders,
+) -> None:
+    mutable_headers_from_tuple.add("foo", "baz")
+    mutable_headers_from_tuple.extend_header_value("foo", "qux")
+    assert mutable_headers_from_tuple.getall("foo") == ["bar,baz,qux"]
+
+
 def test_mutable_scope_headers_getitem(mutable_headers: MutableScopeHeaders, existing_headers_key: str) -> None:
     assert mutable_headers[existing_headers_key] == "bar"
 
--- a/tests/unit/test_datastructures/test_headers.py
+++ b/tests/unit/test_datastructures/test_headers.py
@@ -184,6 +184,17 @@ def test_mutable_scope_headers_from_tuple_extend_header_value_new_header(
     assert list(raw_headers_tuple) == [(b"foo", b"bar"), (b"bar", b"baz")]
 
 
+def test_mutable_scope_headers_extend_header_value_preserves_duplicate_entries(
+    mutable_headers: MutableScopeHeaders,
+) -> None:
+    # A header can legitimately appear as multiple separate raw entries (not just a
+    # single comma-joined value). extend_header_value must not silently drop any of
+    # them.
+    mutable_headers.add("foo", "baz")
+    mutable_headers.extend_header_value("foo", "qux")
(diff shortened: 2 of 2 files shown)


tests/unit/test_datastructures/test_headers.py
```diff
@@ -184,6 +184,17 @@ def test_mutable_scope_headers_from_tuple_extend_header_value_new_header(
     assert list(raw_headers_tuple) == [(b"foo", b"bar"), (b"bar", b"baz")]
 
 
+def test_mutable_scope_headers_extend_header_value_preserves_duplicate_entries(
+    mutable_headers: MutableScopeHeaders,
```
sobolevn, line 188: Any reason, why you don't test `mutable_headers_from_tuple`?

chuenchen309: Good catch, no reason — just missed pairing it with the tuple-backed fixture like the other tests in this file do. Added `test_mutable_scope_headers_from_tuple_extend_header_value_preserves_duplicate_entries` in 1a3726fde.

litestar/datastructures/headers.py
```diff
@@ -188,9 +188,13 @@ def extend_header_value(self, key: str, value: str) -> None:
         Returns:
             None
         """
-        existing = self.get(key)
-        if existing is not None:
-            value = ",".join([*existing.split(","), value])
+        try:
+            existing_entries = self.getall(key)
```
sobolevn, line 192:
```suggestion
            existing_entries = self.getall(key, [])
```

why not using a default?

tests/unit/test_datastructures/test_headers.py
chuenchen309: Good call — I added exactly that in the follow-up commit `test(datastructures): add tuple-backed counterpart`: `test_mutable_scope_headers_from_tuple_extend_header_value_preserves_duplicate_entries` runs the same duplicate-preservation assertion through the `mutable_headers_from_tuple` fixture, so both the dict-backed and tuple-backed construction paths are now covered.

litestar/datastructures/headers.py
chuenchen309: I tried this, but `self.getall(key, [])` doesn't behave as expected here: `getall` gates its default with `if default:` (truthy) rather than `if default is not None:`, so a falsy `[]` is ignored and it still raises `KeyError` on an absent header — the exact case `extend_header_value` needs to survive. Quick repro on this branch:

```python
>>> MutableScopeHeaders().getall("absent", [])
KeyError
```

That is why I kept the explicit `try/except`. Happy to instead tighten `getall` to `if default is not None:` (so an empty-list default works the way one would reasonably expect) and then use your one-liner — just let me know which direction you prefer.

sobolevn: Fixed in https://github.com/litestar-org/litestar/pull/4932

chuenchen309:
Thanks — and `KeyError(key)` is a good catch on top of it.

Once #4932 lands, the `try`/`except KeyError` here collapses into exactly what you originally suggested:

```python
existing_entries = self.getall(key, [])
```

I applied both locally to check rather than assume: with #4932's `if default is not None`, the one-liner passes the full `test_headers.py` (74 passed), identical to the `try`/`except` form.

I've left this PR on the `try`/`except` so it doesn't depend on an unmerged branch. Happy either way on ordering — merge #4932 first and I'll push the simplification here, or take this as-is and I'll send the cleanup separately. Just say which you prefer.

PR #4933, opened 2026-07-17
fix(openapi): order path parameters to match the url in generated schema

### LINKED ISSUE #3644: Bug: openapi parameter order depends on whether a dependency is consumed
### Description

Unconsumed parameters [appear to be added last to the openapi spec](https://github.com/litestar-org/litestar/blob/ffaf5616b19f6f0f4128209c8b49dbcb41568aa2/litestar/_openapi/parameters.py#L226), which causes strange parameter ordering in other generated code.  It looks like some effort was made to account for this at https://github.com/litestar-org/litestar/blame/ffaf5616b19f6f0f4128209c8b49dbcb41568aa2/tests/unit/test_openapi/test_schema.py#L613, it even looks like some effort was made to allow for this.   I'm wondering if this is really the desired behavior.  I would have expected the order in the openapi spec to match the order in the path.  As it stands, it means that consuming a dependency will change the openapi spec and code generated from it.

_No response_

### MCVE

_No response_

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.9.1

### Platform

- [ ] Linux
- [ ] Mac
- [ ] Windows

Bishwas-py:
## Description

The generated OpenAPI schema orders path parameters by how the handler happens to consume them, not by their position in the url. A path parameter consumed through a dependency, or not consumed at all, gets pushed out of its natural url position, so refactoring code that doesn't touch the url can still reorder the public schema (and any client code generated from it).

Fixes #3644

### Root cause

`ParameterFactory.create_parameters_for_handler` builds the parameter list by iterating the handler's function signature, recursing into dependency providers when a field is a dependency, and appending unconsumed path parameters at the very end. None of these three insertion paths follow the url's actual left-to-right order, even though that order is already available via `route.path_parameters`.

### Fix

Added `ParameterFactory._order_path_parameters`, which reorders only the path-type parameters to match their position in `self.path_parameters` (the url's real order) before the list is returned. Query and header parameters keep their existing position; this is scoped to fixing path-parameter ordering specifically.

## How was this tested

Added three tests to `tests/unit/test_openapi/test_parameters.py`:
- a path parameter consumed via a dependency
- a path parameter left unconsumed entirely
- a path parameter interleaved with a query parameter, to confirm the reorder doesn't disturb query parameter position

Before the fix, for `/users/{user_id}/posts/{post_id}/comments/{comment_id}` with `post_id` consumed via a dependency, the generated schema ordered parameters as `user_id, comment_id, post_id`. After the fix, it orders them as `user_id, post_id, comment_id`, matching the url.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4933">https://litestar-org.github.io/litestar-docs-preview/4933</a>


litestar/_openapi/parameters.py
```diff
@@ -246,7 +246,25 @@ def create_parameters_for_handler(self) -> list[Parameter]:
...
+
+        Returns:
+            The parameters with path parameters ordered to match the URL path.
+        """
+        path_order = {name: index for index, name in enumerate(self.path_parameters)}
+        path_slots = [i for i, parameter in enumerate(parameters) if parameter.param_in == ParamType.PATH]
+        ordered_path_parameters = sorted(
+            (parameters[i] for i in path_slots), key=lambda parameter: path_order[parameter.name]
+        )
+        for slot, parameter in zip(path_slots, ordered_path_parameters, strict=True):
```
sobolevn: question: can we order all parameters? it would be a bigger change, but I guess it makes sense to have a strong and consistent order based on: type and internal logic for each type.

Bishwas-py: scoped this to path parameters, matching #3644. ordering all parameter types is a bigger change, it would need a canonical cross-type order and would break `test_unwrap_new_type`'s current assumption that a query param can precede a path param. am happy to open a follow-up for that if you're interested.

sobolevn: I don't think that this is a good idea to split ordering in PRs. Let's focus on this problem in a single PR :)

provinzkraut: I agree with @sobolevn. Let's try to find a general solution for this problem that provides a canonical ordering for parameters of all types.

Bishwas-py: pushed a general solution in 7f7273dd2. parameters are now ordered by type: path, query, cookie, header (matching ParamType's own order), then by an internal rule per type as url position for path and alphabetical for the rest.

litestar/_openapi/parameters.py
```diff
@@ -246,7 +248,30 @@ def create_parameters_for_handler(self) -> list[Parameter]:
...
+
+        Args:
+            parameters: The parameters to order.
+
+        Returns:
+            The parameters, ordered by type and then by url position or name.
+        """
+        path_order = {name: index for index, name in enumerate(self.path_parameters)}
+
+        def sort_key(parameter: Parameter) -> tuple[int, int | str]:
```
sobolevn: why do we put this function inside the method?

Bishwas-py: it closed over path_order, which is local there. inlined it as a lambda in 9f6737bc0, no nested def needed.

litestar/_openapi/parameters.py
```diff
@@ -246,7 +248,30 @@ def create_parameters_for_handler(self) -> list[Parameter]:
...
+        Args:
+            parameters: The parameters to order.
+
+        Returns:
+            The parameters, ordered by type and then by url position or name.
+        """
+        path_order = {name: index for index, name in enumerate(self.path_parameters)}
+        return sorted(
+            parameters,
+            key=lambda parameter: (
```
sobolevn: Let's make this a module level protected named function please :) 

There's not need to compile it in-place

Bishwas-py: moved it to a module level protected function in 7bcd4eac1, bound path_order with functools.partial.

litestar/_openapi/parameters.py
```diff
@@ -14,13 +15,23 @@
...
     from litestar.handlers.base import BaseRouteHandler
     from litestar.openapi.spec import Reference
     from litestar.types.internal_types import PathParameterDefinition
 
 __all__ = ("create_parameters_for_handler",)
 
+_PARAM_TYPE_ORDER = {"path": 0, "query": 1, "cookie": 2, "header": 3}
+
+
+def _parameter_sort_key(parameter: Parameter, path_order: Mapping[str, int]) -> tuple[int, int | str]:
```
sobolevn: let's put it in very bottom :)

Bishwas-py: yup, moved it

litestar/_openapi/parameters.py
```diff
@@ -246,7 +249,24 @@ def create_parameters_for_handler(self) -> list[Parameter]:
         )
 
         self.create_parameters_for_field_definitions(handler_fields)
-        return self.parameters.list()
+        return self._order_parameters(self.parameters.list())
```
provinzkraut, line 252: I think it would be good to change `ParameterCollection.list` so it doesn't convert the parameters to a list, but return the dict view directly; Since the list is thrown away immediately and only consumed by `sorted` to iterate over it, any iterable will do, and would be more efficient

Bishwas-py: done in c85cec069, list returns the dict view directly now.

litestar/_openapi/parameters.py
```diff
@@ -69,9 +72,9 @@ def add(self, parameter: Parameter) -> None:
             f"'{parameter.name}' with different types."
         )
 
-    def list(self) -> list[Parameter]:
-        """Return a list of all ``Parameter``'s in the collection."""
-        return list(self._parameters.values())
+    def list(self) -> ValuesView[Parameter]:
```
provinzkraut:
```suggestion
    def list(self) -> Iterable[Parameter]:
```
`ValuesView` is unnecessarily specific

Bishwas-py: @provinzkraut any rereview?

Bishwas-py:
please review

PR #4936, opened 2026-07-19
fix(routing): return Not Found for websocket upgrade to non-websocket route

### LINKED ISSUE #4935: Bug: Unhandled websocket upgrade request returns 500
### Description

A websocket upgrade request for a path that has no websocket handlers throws an error with a logged stack trace, inclusive of:

```
File ".../litestar/_asgi/routing_trie/traversal.py", line 86, in parse_node_handlers
    return node.asgi_handlers["websocket"]
KeyError: 'websocket'
```

As a user with a logging configuration that downgrades litestar.exception types that are used deliberately to be logged at INFO, this means I'm getting ERROR-level logs going through paths meant for unexpected circumstances.

This can be fixed with:

```patch
diff --git a/litestar/_asgi/routing_trie/traversal.py b/litestar/_asgi/routing_trie/traversal.py
index 1164d0cdb..6742e5c23 100644
--- a/litestar/_asgi/routing_trie/traversal.py
+++ b/litestar/_asgi/routing_trie/traversal.py
@@ -83,7 +83,10 @@ def parse_node_handlers(
             return node.asgi_handlers[method]
         except KeyError as e:
             raise MethodNotAllowedException(headers={"allow": ", ".join(node.asgi_handlers.keys())}) from e
-    return node.asgi_handlers["websocket"]
+    try:
+        return node.asgi_handlers["websocket"]
+    except KeyError as e:
+        raise NotFoundException() from e


 @lru_cache(1024)
```

_No response_

### MCVE

```python
import pytest

from litestar import get
from litestar.exceptions import WebSocketDisconnect


@get("/")
async def index() -> str:
    return "hello"


def test_websocket_upgrade_to_http_only_route_is_not_found() -> None:
    # litestar's default LoggingConfig dumps a full traceback for every exception it
    # catches, expected or not -- disable it so only a genuinely unexpected failure
    # (the AssertionError below) produces output.
    with create_test_client([index], logging_config=None) as client:
        with pytest.raises(
            WebSocketDisconnect
        ) as unknown_path, client.websocket_connect("/nope"):
            pass
        with pytest.raises(
            WebSocketDisconnect
        ) as http_only_route, client.websocket_connect("/"):
            pass

        assert (
            http_only_route.value.detail == unknown_path.value.detail == "Not Found"
        ), f"{http_only_route.value.detail!r} == {unknown_path.value.detail!r} == 'Not Found'"


if __name__ == "__main__":
    test_websocket_upgrade_to_http_only_route_is_not_found()
    print("PASS")
```

### Steps to reproduce

If the error is present, the MCVE shown above will emit:

```
AssertionError: 'Internal Server Error' == 'Not Found' == 'Not Found'
```

Otherwise, it will emit:

```
PASS
```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.24.0

### Platform

- [x] Linux
- [x] Mac
- [ ] Windows

vidigoat:
## Description

Fixes #4935.

If you send a websocket upgrade to a path that only has HTTP handlers (e.g. a route defined with `@get("/")`), the server blows up with a raw `KeyError: 'websocket'` instead of just rejecting the connection.

The cause is in `parse_node_handlers`: for a websocket (no method) it does `node.asgi_handlers["websocket"]` directly, with no guard. If the node exists but has no websocket handler, that lookup raises `KeyError`, which bubbles up as an unhandled 500. The method branch right above it already guards its lookup and raises `MethodNotAllowedException`, and a completely unknown path already returns Not Found — so the HTTP-only-route case was the only one left unguarded.

The fix wraps that lookup and raises `NotFoundException`, so a websocket upgrade to a non-websocket route closes cleanly with 4500 "Not Found", same as an unknown path.

I added a regression test that connects a websocket to both an unknown path and an HTTP-only route and checks both are rejected as Not Found. The HTTP-only case fails with the old code (raises `KeyError`) and passes with the fix.

## Closes

Closes #4935

### DIFF
--- a/litestar/_asgi/routing_trie/traversal.py
+++ b/litestar/_asgi/routing_trie/traversal.py
@@ -85,7 +85,10 @@ def parse_node_handlers(
             return node.asgi_handlers[method]
         except KeyError as e:
             raise MethodNotAllowedException(headers={"allow": ", ".join(node.asgi_handlers.keys())}) from e
-    return node.asgi_handlers["websocket"]
+    try:
+        return node.asgi_handlers["websocket"]
+    except KeyError as e:
+        raise NotFoundException() from e
 
 
 @lru_cache(1024)
--- a/tests/unit/test_asgi/test_routing_trie/test_traversal.py
+++ b/tests/unit/test_asgi/test_routing_trie/test_traversal.py
@@ -113,4 +113,6 @@ def test_websocket_upgrade_without_matching_handler_is_not_found(path: str) -> N
     ):
         pass
 
-    assert exc.value.exceptions[0].detail == "Not Found"
+    disconnect = exc.value.exceptions[0]
+    assert isinstance(disconnect, WebSocketDisconnect)
+    assert disconnect.detail == "Not Found"
--- a/tests/unit/test_asgi/test_routing_trie/test_traversal.py
+++ b/tests/unit/test_asgi/test_routing_trie/test_traversal.py
@@ -1,6 +1,9 @@
 from typing import Any
 
+import pytest
+
 from litestar import Router, asgi, get
+from litestar.exceptions import WebSocketDisconnect
 from litestar.response.base import ASGIResponse
 from litestar.status_codes import HTTP_404_NOT_FOUND
 from litestar.testing import create_test_client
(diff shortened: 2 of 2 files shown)


provinzkraut:
Closing as per our AI policy. Further violations will result in being blocked from contributing to this organisation.

provinzkraut:
Added clarity for others seeing this: This PR is wrong. It has been discussed why this is the wrong approach. The contributor chose to ignore this and submit an LLM generated PR based on the initial, wrong, proposed fix. Since *they* did not bother to invest a few minutes of effort to follow the discussion, I do not see any value investing *my* time to get this PR into a mergeable state.

PR #4938, opened 2026-07-20
fix(routing): unsupported websocket upgrade attempt should be 400, not 500

### LINKED ISSUE #4935: Bug: Unhandled websocket upgrade request returns 500

charles-dyfis-net:
## Description

Presently, a websocket upgrade attempt on a URL that doesn't support it throws a `KeyError`, which becomes a HTTP 500. A result code that reflects client rather than server misbehavior is more appropriate.

Closes #4935

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4938">https://litestar-org.github.io/litestar-docs-preview/4938</a>

### DIFF
--- a/litestar/_asgi/routing_trie/traversal.py
+++ b/litestar/_asgi/routing_trie/traversal.py
@@ -4,7 +4,7 @@ from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
 from litestar._asgi.routing_trie.types import PathParameterSentinel
-from litestar.exceptions import MethodNotAllowedException, NotFoundException
+from litestar.exceptions import ClientException, MethodNotAllowedException, NotFoundException
 from litestar.utils import normalize_path
 
 __all__ = ("parse_node_handlers", "parse_path_params", "parse_path_to_route", "traverse_route_map")
@@ -85,7 +85,10 @@ def parse_node_handlers(
             return node.asgi_handlers[method]
         except KeyError as e:
             raise MethodNotAllowedException(headers={"allow": ", ".join(node.asgi_handlers.keys())}) from e
-    return node.asgi_handlers["websocket"]
+    try:
+        return node.asgi_handlers["websocket"]
+    except KeyError as e:
+        raise ClientException() from e
 
 
 @lru_cache(1024)
--- a/tests/unit/test_asgi/test_routing_trie/test_traversal.py
+++ b/tests/unit/test_asgi/test_routing_trie/test_traversal.py
@@ -1,6 +1,9 @@
 from typing import Any
 
+import pytest
+
 from litestar import Router, asgi, get
+from litestar.exceptions import WebSocketDisconnect
 from litestar.response.base import ASGIResponse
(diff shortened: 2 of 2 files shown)


charles-dyfis-net:
Filed as draft because unit tests have only been run against Python 3.14 -- will only remove if other versions pass CI.

provinzkraut (approved):

PR #4955, opened 2026-07-31
fix(openapi): reflect root_path in OpenAPI servers (#2077)

### LINKED ISSUE #2077: Enhancement: Add `root_path` value to OpenAPI servers if present

shrutz2:
## Description
Fixes #2077

When an app is served under a `root_path` (e.g. `uvicorn app --root-path /api`,
or behind a proxy), the generated OpenAPI schema still listed its servers as
`[{"url": "/"}]`, so the base URL shown in the docs/spec was wrong.

This makes the OpenAPI plugin substitute the request's `root_path` into `servers`
when the user has not configured a custom `servers` value. A custom `servers`
value is left untouched, and the cached schema is not mutated (a per-request copy
is returned). It follows the same `app_root_path` / `root_path` precedence that
`Request.base_url` already uses.

Tests cover three cases: the root_path is reflected in `servers`, a custom
`servers` value is not overridden, and the default `/` is preserved when no
root_path is set.

> This PR was created with AI assistance (Claude by Anthropic). The implementation follows the approach outlined by maintainers in the issue discussion.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4955">https://litestar-org.github.io/litestar-docs-preview/4955</a>

### DIFF
--- a/litestar/_openapi/plugin.py
+++ b/litestar/_openapi/plugin.py
@@ -100,6 +100,24 @@ class OpenAPIPlugin(InitPlugin, ReceiveRoutePlugin):
             self._openapi_schema = self.provide_openapi().to_schema()
         return self._openapi_schema
 
+    def schema_for_request(self, request: Request) -> dict[str, Any]:
+        """Return the OpenAPI schema, reflecting the ASGI ``root_path`` in ``servers``.
+
+        When the application is mounted under a sub-path (e.g. ``root_path="/api"``)
+        the schema's default ``servers`` value of ``[{"url": "/"}]`` no longer points
+        at the real base URL. If the user has not customised ``servers``, substitute
+        the request's ``root_path`` so generated URLs are correct (issue #2077). A
+        customised ``servers`` value is left untouched, and the cached schema is not
+        mutated.
+        """
+        schema = self.provide_openapi_schema()
+        # Mirror Request.base_url precedence: app_root_path (set by e.g. uvicorn
+        # --root-path) wins over the ASGI root_path.
+        root_path = (request.scope.get("app_root_path") or request.scope.get("root_path", "")).rstrip("/")
+        if root_path and schema.get("servers") == [{"url": "/"}]:
+            schema = {**schema, "servers": [{"url": root_path}]}
+        return schema
+
     def create_openapi_router(self) -> Router:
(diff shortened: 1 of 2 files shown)


litestar/_openapi/plugin.py
```diff
@@ -100,6 +100,24 @@ def provide_openapi_schema(self) -> dict[str, Any]:
...
+
+        When the application is mounted under a sub-path (e.g. ``root_path="/api"``)
+        the schema's default ``servers`` value of ``[{"url": "/"}]`` no longer points
+        at the real base URL. If the user has not customised ``servers``, substitute
+        the request's ``root_path`` so generated URLs are correct (issue #2077). A
+        customised ``servers`` value is left untouched, and the cached schema is not
+        mutated.
+        """
+        schema = self.provide_openapi_schema()
+        # Mirror Request.base_url precedence: app_root_path (set by e.g. uvicorn
```
jaens, line 114: does this partially reimplement `Request.base_url` then?

PR #4961, opened 2026-08-06
fix(docs): wrong postgres volume path

suspiciousRaccoon:
## Description
Changes the path of the postgres volume in the docker compose example from `/var/lib/postgresql/data` to `/var/lib/postgresql` since it uses latest and it [changed](https://github.com/docker-library/postgres/blob/62a714f93cc32220de46fd12235c9d509e3b1ad6/18/bookworm/Dockerfile#L188-L193) on version 18+. 

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4961">https://litestar-org.github.io/litestar-docs-preview/4961</a>

### DIFF
--- a/docs/topics/deployment/docker.rst
+++ b/docs/topics/deployment/docker.rst
@@ -159,7 +159,7 @@ If you want to run the container as part of a Docker Compose setup then you can 
         ports:
           - "5432:5432"
         volumes:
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
Thanks!

Issue #4489, opened 2025-11-26
Enhancement(CI): Add Zizmor


JacobCoffee:
### Summary

- Add [Zizmor](https://github.com/zizmorcore/zizmor)
- Also enable [immutable releases](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases) (only maintainer's can do this, not a code change)
- Maybe enable sigstore signing? (though maybe [gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) does that for us i cant remember)

abhisek343:
Hi! I'd like to work on the code portion of this issue: add Zizmor to the repository's CI/pre-commit workflow, address or explicitly document any findings, and add focused validation for the configuration.

I understand that enabling immutable releases is a maintainer-only repository setting, so I would leave that out of the PR. I would also keep Sigstore signing separate unless maintainers confirm it belongs in the same change. Is this issue available for me to take?

PR #4962, opened 2026-08-06
feat(ci): Zizmor baseline policy + Dependabot/uv cooldown (follow-up to #4489) (#4947)

### LINKED ISSUE #4947: Enhancement(CI): Zizmor baseline policy + Dependabot/uv cooldown (follow-up to #4489)
### Summary

Follow-up to #4489 to agree on a baseline for adding [Zizmor](https://github.com/zizmorcore/zizmor) (GitHub Actions static analysis), plus a couple of related Dependabot/pre-commit hardening items. Opening this per @cofin's request so the direction gets a review before I put up the PR.

Everything here is a recommendation — happy to go whichever way maintainers prefer. The actual config and workflow fixes land in a follow-up PR.

### Basic Example

**1. How to run it**
- `pre-commit` hook ([`zizmor-pre-commit`](https://github.com/zizmorcore/zizmor-pre-commit)) so findings show up locally.
- For CI, two ways to run it:
  1. **CLI directly** — `uvx zizmor` (we already use `uv`), then upload the SARIF ourselves. More control, one more action to pin.
  2. **[`zizmorcore/zizmor-action`](https://github.com/zizmorcore/zizmor-action)** — wraps the run + SARIF upload for us. Less boilerplate, but it's another third-party action to SHA-pin and trust.
  Either way results land in the Security tab and as PR annotations. Slight lean toward the CLI for control, but no strong opinion — maintainer's call.
- I'd make it **blocking** on PRs. Most of the findings below are mechanical to fix, so it's a one-time cleanup rather than an ongoing tax.
- I'd stick with the default (`Regular`) persona — the full run below came from it, so `pedantic` doesn't add anything for the baseline. Easy to tighten later.

**2. Action pinning (`unpinned-uses`)**
Actions are pinned to tags today (e.g. `actions/checkout@v7.0.1`), and Zizmor flags 52 of those as High. Given the recent run of supply-chain incidents (`tj-actions/changed-files`, etc.), I'd pin everything to a full commit SHA with a version comment and let Dependabot keep the SHAs fresh:

```yaml
# .github/zizmor.yml
rules:
  unpinned-uses:
    config:
      policies:
        "*": hash-pin
```

Third-party actions in the tree: `astral-sh/setup-uv`, `JamesIves/github-pages-deploy-action`, `codecov/codecov-action`, `dawidd6/action-download-artifact`, `pypa/gh-action-pypi-publish`, `ikalnytskyi/action-setup-postgres`. No `litestar-org/*` actions are used. If you'd rather keep GitHub's own actions on tags, `"actions/*": ref-pin` carves those out — the rule is per-org, so the "some like hashes, some don't" thing is fully configurable.

**3. What a real run turns up**
88 findings total (58 High / 17 Medium / 13 Low), all at the default persona:

| Rule | Count | Severity | Where | Fix |
|---|---|---|---|---|
| `unpinned-uses` | 52 | High | all workflows | SHA-pin (see above) |
| `excessive-permissions` | 17 | Medium | ci, publish, pr-merged, pr-title | add least-privilege `permissions:` blocks |
| `artipacked` | 13 | Low | ci, docs, notify-released-issues, publish, test | `persist-credentials: false` on checkouts that don't push |
| `cache-poisoning` | 2 | High | docs.yml, publish.yml | guard `actions/cache` on the release paths |
| `template-injection` | 2 | High | notify-released-issues.yml, test.yml | move `${{ github.* }}` into `env:` |
| `dangerous-triggers` | 1 | High | docs-preview.yml (`workflow_run`) | audit (already partly hardened) |
| `github-env` | 1 | High | docs-preview.yml | avoid untrusted writes to `$GITHUB_ENV`/`$GITHUB_PATH` |

Goal is a small `.github/zizmor.yml` where any suppression carries an inline `# reason`, so the policy explains itself.

**4. Dependabot (not Zizmor, but same theme)**
`dependabot.yml` only watches `github-actions` today (7-day cooldown). Two gaps:
- Add a `uv` ecosystem — our Python deps (`uv.lock`) aren't watched at all.
- Put a cooldown on third-party updates but let litestar-org's own packages (Advanced Alchemy, Polyfactory, …) through without the delay, via `cooldown.exclude:`:

```yaml
- package-ecosystem: "uv"
  directory: "/"
  schedule: { interval: "daily" }
  cooldown:
    default-days: 7
    exclude: ["advanced-alchemy", "polyfactory"]  # + rest of litestar-org packages
```

Same exclusion could apply to the existing `github-actions` cooldown.

**5. `.pre-commit-config.yaml`**
Zizmor doesn't scan this, but it's the same pinning story — hooks are tag-pinned (`rev: v5.0.0`, …), and a bad hook runs on every contributor's machine and in CI. I'd pin `rev:` to SHAs to match the actions policy. One catch: Dependabot has no pre-commit ecosystem, so keeping them fresh means either [`pre-commit.ci`](https://pre-commit.ci/) or Renovate — worth deciding who owns those bumps.

### Drawbacks and Impact

- SHA-pinning is less readable and adds Dependabot PRs (cooldown + grouping keep that in check).
- Turning CI blocking means one upfront cleanup pass: 52 hash-pins (mostly mechanical), 17 permission blocks, 13 `persist-credentials`, and ~6 real-logic findings.
- One extra tool in the chain, but Zizmor is self-contained and low-maintenance.
- Net win for a widely-depended-on library.

Optional extras worth a look:
- [OpenSSF Scorecard](https://github.com/ossf/scorecard-action) — pairs well, feeds the same code-scanning UI.
- Set the repo's default `GITHUB_TOKEN` to read-only (a settings toggle) — pairs with the permissions work and the immutable-releases item in #4489.
- Dependabot grouped updates + an `open-pull-requests-limit` once `uv` is watched.

### Unresolved questions

- Hash-pin all actions, or keep `actions/*` on tags?
- Run it via the CLI (`uvx zizmor`) or [`zizmorcore/zizmor-action`](https://github.com/zizmorcore/zizmor-action)?
- Blocking vs advisory to start, and do we want the SARIF upload?
- Fix the 6 real-logic High findings in the same PR, or split them out?
- Which litestar-org packages should be excluded from the cooldown?
- SHA-pin the pre-commit hooks too — and if so, `pre-commit.ci` or Renovate for the bumps?
- Any appetite for the optional extras above?

namdamdoi68-oss:
Follow-up to #4489.

- Adds `.github/workflows/zizmor.yml`: runs zizmor on `pull_request` and `push` to `main`, with hash-pinned actions and least-privilege permissions.
- Adds `.github/zizmor.yml`: sets the `dependabot-cooldown` baseline to 7 days.

Closes #4947

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4962">https://litestar-org.github.io/litestar-docs-preview/4962</a>

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,23 @@
+name: Zizmor Security Audit
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  zizmor:
+    name: zizmor
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,5 @@
+# Zizmor security audit baseline policy configuration
+rules:
+  dependabot-cooldown:
+    config:
+      days: 7


namdamdoi68-oss:
Hi @Goldziher @provinzkraut — fix for #4947 is ready for review. All 23 CI checks are 100% GREEN. Added Zizmor baseline policy and Dependabot/uv cooldown.

.github/dependabot.yml
```diff
@@ -6,3 +6,10 @@ updates:
       interval: "daily"
     cooldown:
       default-days: 7
+
+  - package-ecosystem: "pip"
```
sobolevn: This does not seem correct, we use `uv`

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@v0.6.2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
sobolevn: https://github.com/zizmorcore/zizmor-action does not list `token` as required. Why is it needed?

tests/unit/test_ci_workflow_config.py
```diff
@@ -0,0 +1,51 @@
+import unittest
+from pathlib import Path
+
+import yaml
+
+
+class TestCIWorkflowConfig(unittest.TestCase):
```
sobolevn: Tests must not use CI workflows.

sobolevn (changes requested):
If you are using AI, please double check its output.

namdamdoi68-oss:
Thanks for the thorough review @sobolevn! I've addressed all your feedback:
1. Removed `pip` ecosystem from `.github/dependabot.yml` (keeping only `github-actions` as Litestar uses `uv`).
2. Removed the unnecessary `token` input from `.github/workflows/zizmor.yml`.
3. Pinned action references to full commit SHAs (`actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683` and `zizmorcore/zizmor-action@07c082875150937b4ee74a88f72df9e5ee9f7832`) to satisfy security scanning.
4. Removed `tests/unit/test_ci_workflow_config.py`.

Squashed into a single clean commit with DCO sign-off (`Signed-off-by: namdamdoi68-oss`). Ready for review!

PR #4963, opened 2026-08-06
fix(cors): never emit a wildcard Access-Control-Allow-Origin alongside credentials

sridhar-3009:
## What

`CORSConfig(allow_credentials=True)` combined with the default `allow_origins=["*"]` produces a response carrying both:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```

Per the [Fetch spec](https://fetch.spec.whatwg.org/#http-access-control-allow-origin), a literal `*` is invalid alongside credentials — when a request's credentials mode is `include`, the server has to echo back a specific origin, not the wildcard. Browsers enforce this by rejecting the response outright (treating it as a CORS failure), so any app that sets `allow_credentials=True` without also narrowing `allow_origins` away from the default wildcard gets credentialed cross-origin requests that silently never succeed in the browser, even though the server-side response looks fine.

Repro against current `main`:

```python
from litestar.config.cors import CORSConfig

@get("/endpoint")
async def handler() -> str: return "ok"

app = Litestar(route_handlers=[handler], cors_config=CORSConfig(allow_credentials=True))

    r = client.get("/endpoint", headers={"Origin": "https://example.com"})
    print(r.headers["access-control-allow-origin"])       # "*"
    print(r.headers["access-control-allow-credentials"])  # "true"  <- invalid combination
```

I found this doing a read through `config/cors.py` / `middleware/_internal/cors.py` — it isn't covered by any existing test; `test_cors_credentials.py` only exercises credentials with an explicit, narrow `allow_origins`, never the wildcard-plus-credentials combination.

## Fix

- `simple_headers` / `preflight_headers` (`config/cors.py`) now omit the wildcard `Access-Control-Allow-Origin` entirely when `allow_credentials` is set, since these are cached and origin-agnostic — they can't know the request's actual origin.
- The middleware (`middleware/_internal/cors.py`) always echoes the validated request origin back in that case, on both the preflight response and the actual response.
- This also fixes an existing correctness gap in `send_wrapper`: origin-echoing was previously gated on `has_cookie` (whether the current request happens to carry a `Cookie` header) rather than on `self.config.allow_credentials`. That heuristic misses the very first credentialed request of a session — the one establishing the cookie in the first place — which doesn't carry one yet. Swapped it for the config flag, which is both correct and simpler (removed the now-unused `has_cookie` parameter entirely).

## Tests

- Updated the existing `test_cors_simple_response` parametrized test — it was asserting the old (invalid) `"*"` behavior for every `allow_credentials=True` case; now asserts the origin is echoed instead when credentials are on.
- Added `test_cors_credentials_with_wildcard_origin_echoes_specific_origin`, exercising both the preflight and actual-response path with the default wildcard `allow_origins`, including the no-cookie-yet case.

Full suite (5536 tests), `pre-commit`, and `mypy` all pass locally.

### DIFF
--- a/litestar/config/cors.py
+++ b/litestar/config/cors.py
@@ -110,7 +110,11 @@ class CORSConfig:
             A dictionary of headers to set on the response object.
         """
         headers: dict[str, str] = {"Access-Control-Max-Age": str(self.max_age)}
-        if self.is_allow_all_origins:
+        # A literal "*" is invalid alongside credentials per the Fetch spec - browsers
+        # reject the response outright. When both are configured, the specific request
+        # origin has to be echoed back instead, which middleware does per-request since
+        # this property is cached and origin-agnostic.
+        if self.is_allow_all_origins and not self.allow_credentials:
             headers["Access-Control-Allow-Origin"] = "*"
         else:
             headers["Vary"] = "Origin"
@@ -138,7 +142,7 @@ class CORSConfig:
             A dictionary of headers to set on the response object.
         """
         simple_headers = {}
-        if self.is_allow_all_origins:
+        if self.is_allow_all_origins and not self.allow_credentials:
             simple_headers["Access-Control-Allow-Origin"] = "*"
         if self.allow_credentials:
             simple_headers["Access-Control-Allow-Credentials"] = "true"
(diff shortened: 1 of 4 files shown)


provinzkraut:
If it's invalid to set both, we should raise an exception at startup / config time, instead of silently adapting.

provinzkraut:
@sridhar-3009 are you still working on this?

Issue #4110, opened 2025-04-13
Bug: no event publish for psycopg and asyncpg channels backend


Harshal6927:
### Description

When publishing events using `PsycoPgChannelsBackend` or `AsyncPgChannelsBackend`, they are not being published to subscribers.

Note: This example will work with `MemoryChannelsBackend`

### MCVE

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "litestar[standard]>=2.15.2",
#   "psycopg[binary,pool]>=3.2.6",
# ]
# ///

from litestar import Litestar, get
from litestar.channels import ChannelsPlugin
import time
from litestar.channels.backends.memory import MemoryChannelsBackend
# from litestar.channels.backends.psycopg import PsycoPgChannelsBackend

import uvicorn

channels_plugin = ChannelsPlugin(
    # backend=PsycoPgChannelsBackend(
    #     "postgresql://myuser:mypassword@127.0.0.1:8931/mydb"
    # ),
    backend=MemoryChannelsBackend(),
    arbitrary_channels_allowed=True,
    create_ws_route_handlers=True,
    ws_handler_base_path="/ws",
)


@get("/")
async def index() -> str:
    return "Hello, world!"


@get("/create-event")
async def create_event() -> None:
    channels_plugin.publish(time.time_ns(), "test-event-channel")


app = Litestar(
    route_handlers=[index, create_event],
    plugins=[channels_plugin],
    debug=True,
)


if __name__ == "__main__":
    uvicorn.run(app=app, port=8621)



# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "websockets",
# ]
# ///

import asyncio
import websockets
import json


async def listen_to_channel():
    uri = "ws://127.0.0.1:8621/ws/test-event-channel"

    print(f"Connecting to {uri}...")
    try:
        async with websockets.connect(uri) as websocket:
            print("Connection established! Waiting for messages...")

            while True:
                message = await websocket.recv()
                try:
                    data = json.loads(message)
                    print(f"Received: {data}")
                except json.JSONDecodeError:
                    print(f"Received: {message}")
    except websockets.exceptions.ConnectionClosedError as e:
        print(f"Connection closed unexpectedly: {e}")
    except Exception as e:
        print(f"Error: {e}")


if __name__ == "__main__":
    asyncio.run(listen_to_channel())
```

### Steps to reproduce

[...]

PR #4967, opened 2026-08-08
fix(channels): harden psycopg listener lifecycle

cofin:
## Description

`PsycoPgChannelsBackend` consumes notifications on the same connection used for `LISTEN` and `UNLISTEN`. The active notification generator can therefore block dynamic subscription changes.

This change:

- moves notification consumption to a background pump and queue
- serializes subscription changes
- safely restarts the pump around `LISTEN` and `UNLISTEN`
- filters queued events after unsubscribe
- propagates listener failures
- stops the pump before closing the connection
- removes redundant commits on the autocommit connection
- supports backend reuse across lifecycles

Development and documentation environments now require `psycopg>=3.2.4`. Lockfile changes are limited to psycopg packages.

This replaces #4786 while retaining and crediting its pump, queue, and filtering approach. It addresses the PsycoPg-specific part of #4110.

### Performance

The queue adds approximately 0.35 μs per event in a synthetic benchmark. Subscription transitions add approximately 16 μs before database I/O. The queue is unbounded, so sustained consumer lag can increase memory usage.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4967">https://litestar-org.github.io/litestar-docs-preview/4967</a>


provinzkraut (approved):
Not a fan of the design (or rather that this is what's required), but I don't have any better ideas, and it seems to be the best we can do, so let's get this in.

PR #4969, opened 2026-08-08
fix(channels): make subscriptions lifecycle-safe

### LINKED ISSUE #4867: Bug: ChannelsPlugin.unsubscribe leaks empty channel entries with arbitrary_channels_allowed=True

### LINKED ISSUE #4871: ChannelsPlugin.subscribe is not cancellation-safe: a subscriber is leaked into _channels if the request is cancelled mid-subscribe

### LINKED ISSUE #4894: Bug(channels): concurrent subscribe/unsubscribe race condition causes silent event loss

cofin:
## Description

`ChannelsPlugin` updates subscriber state and backend subscriptions without shared lifecycle synchronization. Concurrent transitions can unsubscribe active subscribers, while failures can leak subscribers and dynamic channel entries.

This change:

- validates and materializes channel inputs before mutation
- serializes subscriber bookkeeping and backend transitions
- registers subscribers before backend subscription and history retrieval
- rolls back state after errors or cancellation
- keeps history retrieval and subscriber shutdown outside the lock
- removes empty dynamic channels while retaining configured channels
- makes repeated unsubscribe operations safe
- subscribes generated WebSockets before accepting the connection

This consolidates the applicable fixes from #4895, #4872, and #4868 while crediting their authors.

Closes #4894  
Closes #4871  
Closes #4867

The Redis stream cursor behavior noted in #4867 remains unchanged.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4969">https://litestar-org.github.io/litestar-docs-preview/4969</a>


litestar/channels/plugin.py
```diff
@@ -180,33 +183,43 @@ async def subscribe(self, channels: str | Iterable[str], history: int | None = N
...
-
-        if history:
-            await self.put_subscriber_history(subscriber=subscriber, limit=history, channels=channels)
+        subscriber = self._subscriber_class(
+            plugin=self,
+            max_backlog=self._max_backlog,
+            backlog_strategy=self._backlog_strategy,
+        )
+        try:
+            async with self._lifecycle_lock:
```
provinzkraut: Are you sure these locks are necessary? The code they guard that mutates things has no suspension points, so I don't see how there could be data cases. The only suspension points are in `subscribe`, which should be idempotent. This goes for the other locked sections as well

Vitaly312: I authored the original PR (#4895) with this race fix and added this lock to make state mutation and backend call atomic to avoid a race condition between concurrent subscribe/unsubscribe (yield point is the I/O on backend.subscribe/unsubscribe) which otherwise leads to reading dirty local state

provinzkraut: > yield point is the I/O on backend.subscribe/unsubscribe) which otherwise leads to reading dirty local state

How? As I said, that operation itself should be idempotent, and it's at the end of the block of actions that mutate internal state. There's no suspension point in between those actions.

Vitaly312: For example, coroutine 1 unsubscribes the last subscriber, modifies local state(makes a list empty), passes control flow to coro2 on an ‘await backend‘, coro2 reads local state(list is empty now) and subscribes the plugin to the channel(as first subscription) and then coro1 unsubscribes the plugin from the backend

litestar/channels/plugin.py
```diff
@@ -180,33 +183,43 @@ async def subscribe(self, channels: str | Iterable[str], history: int | None = N
...
+
+            if history:
+                await self.put_subscriber_history(subscriber=subscriber, limit=history, channels=channels)
+        except BaseException:
+            cleanup_task = create_task(self.unsubscribe(subscriber, channels))
+            try:
+                await asyncio.gather(asyncio.shield(cleanup_task), return_exceptions=True)
+            except CancelledError:
+                await asyncio.gather(cleanup_task, return_exceptions=True)
+            raise
```
Vitaly312: If the backend is unreachable (e.g. network failure):
1. `subscribe` for the first subscriber fails, its except block triggers `unsubscribe`, then `_remove_subscriber`
2. `_remove_subscriber` fails, `self._channels.setdefault(channel, set()).add(subscriber)` in its except block returns the subscriber into the local state
3. So this call of `subscribe` will end by an exception but a subscriber stays in the local state, and next subscriptions will not trigger backend subscription. So, all subsequent events will not be published since the plugin is unsubscribed

I think it's better to just move backend calls before state mutating instead of compensating state mutation on exceptions, it will simplify code.

cofin: @Vitaly312 I agree.  I'll make the necessary adjustments.

litestar/channels/plugin.py
cofin: I spent some time re-reviewing the code as well as the discussion here.  So, just to state it slightly differently, the lock is there to prevent conflicting `subscribe`/`unsubscribe` from incorrectly mutating state?  Multiple calls to `subscribe` are idempotent already and are not really the reason for adding the lock.

litestar/channels/plugin.py
Vitaly312: @cofin Thanks! Actually I opened a PR https://github.com/cofin/litestar/pull/8 into this branch with these adjustments (since it's too big to paste as code suggestions). Feel free to merge or adapt it if helpful!

litestar/channels/plugin.py
Vitaly312: Almost. The lock prevents overwriting the backend status on interleaving of two coroutines where one reads local state which is ahead of the actual backend state.

provinzkraut: > For example, coroutine 1 unsubscribes the last subscriber, modifies local state(makes a list empty), passes control flow to coro2 on an ‘await backend‘, coro2 reads local state(list is empty now) and subscribes the plugin to the channel(as first subscription) and then coro1 unsubscribes the plugin from the backend

The block "mutates local state" is fully synchronous, with no suspension in between. The only time it can suspend is before and after the local state mutation. 

```python
...
if history:
    # this can suspend, but no state is modified before and the call does 
    # not mutate internal state
async with self._lifecycle_lock:
    # this mutates state but there cannot be any suspension
    affected_channels = set()
    new_channels = set()
    for channel in channels:
        channel_subscribers = self._channels.get(channel, set())
        if subscriber not in channel_subscribers:
            if not channel_subscribers:
                new_channels.add(channel)
            affected_channels.add(channel)

    if new_channels:
        # this suspends, but all the state it depends on is synchronous, and
        # the call should be idempotent
        await self._backend.subscribe(new_channels)

    for channel in affected_channels:
        self._channels.setdefault(channel, set()).add(subscriber)
```

my point is: two coroutines cannot interleave during the state mutation, as there's no coro call while that is happening.

Can you show with a test case that this has a race condition without the lock?

Vitaly312: I wrote an MRE in corresponding issue https://github.com/litestar-org/litestar/issues/4894. It isn't adapted for this PR where state mutates after backend call but I hope it shows the problem.

Also my point is that interleaving happens between state mutation and backend call in first coroutine.
So this is how it breaks, even if state mutates after backend call:
1. Initial state: 1 subscriber on the channel
2. coro1 unsubscribes: reads state, decides that it needs to unsubscribe backend. On the line `await self._backend.subscribe(new_channels)` it yields control to the event loop
3. The event loop runs coro2(it subscribes to the same channel). coro2 reads state(there is still one subscriber in this!), decides that it doesn't needs to subscribe backend(since it isn't the new channel, returns.
4. The event loop resumes coro1. It mutates `self._channels`, pops the first subscriber, but there is a second subscriber(added by coro2) at this moment, returns
5. So now backend is unsubscribed, but there is a second subscriber in the `self._channels`, and subsequent subscriptions will not trigger `self._backend.subscribe` and all subscribers will not receive any events.

docs/release-notes/changelog.rst
```diff
@@ -6,6 +6,13 @@
 .. changelog:: 3.0.0
     :date: 2364-01-27
 
+    .. change:: Make channel subscriptions lifecycle-safe
```
provinzkraut, line 9: Can we get a more detailed title and description here? It says what was done mechanically, but doesn't really explain the bug that was fixed

litestar/channels/plugin.py
```diff
@@ -180,33 +183,39 @@ async def subscribe(self, channels: str | Iterable[str], history: int | None = N
...
-        if channels_to_subscribe:
-            await self._backend.subscribe(channels_to_subscribe)
 
+        subscriber = self._subscriber_class(
+            plugin=self,
+            max_backlog=self._max_backlog,
+            backlog_strategy=self._backlog_strategy,
+        )
         if history:
             await self.put_subscriber_history(subscriber=subscriber, limit=history, channels=channels)
```
Vitaly312, line 203: History should be fetched after the state mutation. Otherwise, if the event was published after history fetching but before adding a corresponding subscriber in the `self._channels` (e.g. when backend.subscribe suspends), this event will be lost since it isn't in the history and _sub_worker will ignore it because the subscriber isn't in the `self._channels`. I think history should be fetched in the end like it was before (missed that in my PR), but it should roll back if failed:
```python
...
if history:
    try:
    except:
        await self.unsubscribe(subscriber=subscriber, channels=channels)
        raise
return subscriber
```
But this approach requires some deduplication. I mean, if some event was published after the plugin puts a subscriber in the state, and before history fetching, this event will be consumed twice. I guess this requires changing abstract `ChannelsBackend.stream_events` signature to provide some timestamp/ID for each event to distinguish between identical events and duplicates (or just change `subscribe` signature to not handle history fetching at all/require that consumers must be idempotent).

Issue #4719, opened 2026-04-25
feat(middleware): W3C Trace Context correlation middleware


cofin:
## Summary
New ASGI middleware for distributed-trace correlation. Reads correlation/trace IDs from a configurable W3C + cloud-vendor header fallback list (`x-request-id`, `traceparent`, `x-cloud-trace-context`, `x-amzn-trace-id`, etc.), generates a UUID when no header matches, propagates via `contextvars.ContextVar` so handlers/loggers can read it, restores prior context on unwind. Opt-in via `OpenTelemetryConfig.enable_correlation_middleware`. Usable standalone without OTEL installed.

Reference implementation: sqlspec at `sqlspec/extensions/litestar/plugin.py:169` and `sqlspec/utils/correlation.py`. Pattern is ported verbatim, swapping sqlspec-specific types for Litestar-native ones.

**This is its own PR — explicitly NOT folded into PR #4483.**

## Target branch
`feat/correlation-middleware` (cut from `main`)

## Acceptance criteria
- [ ] `litestar/middleware/correlation.py` exists with `CorrelationContext`, `CorrelationMiddleware`, `TRACE_CONTEXT_FALLBACK_HEADERS`
- [ ] Token-based contextvar reset (idiomatic; handles "no prior value")
- [ ] Defensive `traceparent` parse: malformed → raw header, never raises
- [ ] UUID fallback when no headers match
- [ ] `OpenTelemetryConfig` extended with `enable_correlation_middleware`, `correlation_header`, `correlation_headers`, `auto_trace_headers`
- [ ] Plugin auto-injects middleware when opt-in enabled
- [ ] Works standalone without OTEL installed
- [ ] Works alongside OTEL when both enabled
- [ ] Unit tests at `tests/unit/test_middleware/test_correlation.py`, ≥90% coverage
- [ ] Integration tests via TestClient + AsyncTestClient (sequential AND concurrent isolation)
- [ ] Performance test: <5% overhead vs baseline
- [ ] Docs: `docs/usage/middleware/correlation.rst`, `docs/reference/middleware/correlation.rst`, examples, OTEL docs cross-link, changelog entry
- [ ] **CI on GitHub MUST be 100% green before merge** (hard blocker)
- [ ] `make lint && make test && make test-examples && make docs` clean

## Out of scope
- OTEL deep instrumentation (PR #4483)
- Logging middleware integration (follow-up)

PRD: `tmp/specs/active/correlation-middleware/prd.md` (gitignored)

[...]

PR #4970, opened 2026-08-08
feat(middleware): add W3C Trace Context correlation middleware

cofin:
## Description

Applications need a consistent correlation identifier across handlers, logging, and other request-scoped work without requiring an observability integration. This adds standalone correlation middleware for HTTP and WebSocket connections.

`CorrelationMiddleware` selects the first safe value from a configurable list of request headers and generates a UUID when no suitable value is present. W3C `traceparent` values are validated before extracting the trace ID. The selected identifier is available through `CorrelationContext` and `scope["state"]["correlation_id"]`, and can be written to a configurable response header.

Header names are normalized and deduplicated in priority order. Empty values and values containing control characters are skipped. Existing response correlation headers are replaced, and request context and scope state are restored when the middleware unwinds.

The middleware reads directly from the ASGI scope without additional dependencies or blocking work. It uses `__slots__`, performs bounded header processing, and does not add application-level state.

Refs #4719

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4970">https://litestar-org.github.io/litestar-docs-preview/4970</a>

### DIFF
--- a/litestar/middleware/correlation.py
+++ b/litestar/middleware/correlation.py
@@ -1,18 +1,20 @@
-from collections.abc import Generator, Sequence
-from contextlib import contextmanager
-from contextvars import ContextVar, Token
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Final
 from uuid import uuid4
 
-from litestar.datastructures.headers import MutableScopeHeaders
+from litestar.datastructures.headers import Headers, MutableScopeHeaders
+from litestar.enums import ScopeType
+from litestar.middleware.base import ASGIMiddleware
+from litestar.types import Empty
+from litestar.utils.scope.state import ScopeState
 
 if TYPE_CHECKING:
     from litestar.types import ASGIApp, Message, Receive, Scope, Send
 
 __all__ = (
     "TRACE_CONTEXT_FALLBACK_HEADERS",
-    "CorrelationContext",
     "CorrelationMiddleware",
+    "get_correlation_id",
 )
 
 TRACE_CONTEXT_FALLBACK_HEADERS: Final[tuple[str, ...]] = (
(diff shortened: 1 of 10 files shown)


litestar/middleware/correlation.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def _strip_safe_value(value: str) -> str | None:
+    value = value.strip()
+    if not value or any(ord(character) < 32 or ord(character) == 127 for character in value):
+        return None
+    return value
+
+
+class CorrelationContext:
```
provinzkraut: Is this necessary? It diverges from our usual design quite a bit, and context vars are easy to break e.g. by not copying context properly (which is quite common, and I'm not even sure we're always doing it correctly everywhere).

IMO we should put this in `ScopeState`, since it only exists within a request context, and whenever you're in a request context, you'll also have access to the `scope`, and therefore the `ScopeState`. You could still add a convenience wrapper, just one that stores it in `scope` instead of a context var.

cofin: yep - we can definitely go this route.  Give me some time to rework this, but I'd much prefer to keep it simple if there's a better way.

litestar/middleware/correlation.py
```diff
@@ -0,0 +1,239 @@
...
+    "x-client-trace-id",
+)
+
+_correlation_id_var: ContextVar[str | None] = ContextVar("litestar_correlation_id", default=None)
+_LOWERCASE_HEX = frozenset("0123456789abcdef")
+_MISSING = object()
+
+
+def _is_lowercase_hex(value: str, length: int) -> bool:
+    return len(value) == length and all(character in _LOWERCASE_HEX for character in value)
```
provinzkraut:
```suggestion
    return len(value) == length and not _LOWERCASE_HEX.difference(value)
```

litestar/middleware/correlation.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Args:
+            scope: The ASGI scope.
+
+        Returns:
+            Extracted or generated correlation ID.
+        """
+        for name in self.header_names:
+            name_bytes = name.encode("latin-1")
+            for raw_name, raw_value in scope.get("headers", ()):
```
provinzkraut: Why not use our built in headers utils for this?

litestar/middleware/correlation.py
```diff
@@ -0,0 +1,239 @@
...
+        async def send_wrapper(message: "Message") -> None:
+            if message["type"] == "http.response.start" and self.response_header_name:
+                headers = MutableScopeHeaders.from_message(message)
+                headers[self.response_header_name] = correlation_id
+            await send(message)
+
+        try:
+            await self.app(scope, receive, send_wrapper)
+        finally:
+            CorrelationContext.reset(token)
```
provinzkraut: This means the correlation ID is not available in all middlewares. Is that an intentional limitation?

cofin: In the way I had been using, there wasn't a need for this.  However, now that I'm thinking about it, it's probably best to just share through the scope so other middleware can get the ID.  This should be addressed by the change to use `ScopeState` right?

provinzkraut: Yes, it should

litestar/middleware/correlation.py
```diff
@@ -0,0 +1,173 @@
...
+    "grpc-trace-bin",
+    "x-amzn-trace-id",
+    "x-b3-traceid",
+    "x-client-trace-id",
+)
+
+_LOWERCASE_HEX = frozenset("0123456789abcdef")
+
+
+def get_correlation_id(scope: "Scope") -> str | None:
```
provinzkraut: How about
```suggestion
def get_correlation_id(source: "ASGIConnection | Scope") -> str | None:
```
?

Then you could easily use it anywhere

litestar/middleware/correlation.py
```diff
@@ -0,0 +1,173 @@
...
+)
+
+TRACE_CONTEXT_FALLBACK_HEADERS: Final[tuple[str, ...]] = (
+    "x-request-id",
+    "x-correlation-id",
+    "traceparent",
+    "x-cloud-trace-context",
+    "grpc-trace-bin",
+    "x-amzn-trace-id",
+    "x-b3-traceid",
```
provinzkraut: I think these proprietary ones should not be enabled by default? Could lead to confusing results otherwise

cofin: To me, I think auto-handling the cloud provider's injection would be the least confusing right?  If someone is using this on a cloud console, it would "just work".  Otherwise, they'd have a different correlation ID than the one pushed in by the vendor.  

I think most of the cloud vendors honor one of these, so I don't think it's a massive list we have to maintain, but I can double check this.

Would this be a configuration item we'd need to add otherwise?

litestar/middleware/correlation.py
cofin: nice, I like this change.  I'll make the updates.

tests/unit/test_middleware/test_correlation.py
```diff
@@ -29,6 +35,10 @@ async def _send(_: Message) -> None:
     return None
 
 
+async def _await_middleware(awaitable: Awaitable[None]) -> None:
+    await awaitable
```
provinzkraut: What's the purpose of this?

litestar/middleware/correlation.py
cofin: I thought about this since we last discussed.  I do agree with taking them out. I'll remove the cloud specific ones and provide a way to override/append additional correlation headers.

PR #4972, opened 2026-08-09
fix(openapi): `components` field cannot be a list

sobolevn:
Check the spec definition: https://spec.openapis.org/oas/v3.2.0.html#components-object It is true for both 3.0 and 3.1 as well. Oldest version: https://spec.openapis.org/oas/v3.0.0.html#fixed-fields


sobolevn:
Ok, this is a custom API.

PR #4973, opened 2026-08-10
feat(openapi): document structured query kwargs as query parameters

### LINKED ISSUE #2015: Enhancement: Support for Typing the `query` keyword (and other reserved kwargs) to generate OpenAPI Spec

amantu-qbit:
## What

Closes #2015.

The reserved `query` kwarg receives the whole query mapping, so annotating it with a model makes every field of that model a query parameter, validated as one at runtime. Schema generation skipped all reserved kwargs unconditionally, so those routes came out with no `parameters` entry at all and consumers of the spec couldn't see the parameters the handler actually enforces.

## How

`ParameterFactory.create_parameters_for_query_kwarg` decomposes a structured `query` annotation into one `Parameter` per field.

It goes through the existing `SchemaCreator` rather than introspecting each model library directly, so pydantic, msgspec, attrs, dataclasses and TypedDicts are all handled by the code that already knows how, and nested models keep resolving to component references.

An unannotated `query`, or one annotated as a mapping, returns early — that stays a no-op, and returning before touching the schema creator keeps it side-effect free for those annotations.

## One thing to flag

Because the shared schema registry is used, the annotated model itself is also registered as a component. For `query: Filters` you get parameters for each field **and** a `Filters` entry under `components/schemas` that nothing references.

I went this way deliberately: a throwaway registry would avoid the unreferenced component but would break `$ref`s for nested models, which seemed like the worse trade. If you'd rather not emit it, I'm happy to look at pruning it after generation — just say which you prefer.

## Tests

Seven tests in `tests/unit/test_openapi/test_parameters.py`: per-field parameters with correct required-ness and defaults, nested models keeping their references, no-ops for `Any` / `dict` / `dict[str, str]` / `MultiDict`, the existing collision error still raised when a decomposed field clashes with an explicit parameter of a different type, and a runtime test asserting the documented parameters are the ones actually validated.

### DIFF
--- a/litestar/_openapi/parameters.py
+++ b/litestar/_openapi/parameters.py
@@ -170,12 +170,56 @@ class ParameterFactory:
         )
         return self.create_parameter(field_definition=field_definition, parameter_name=parameter_name)
 
+    def create_parameters_for_query_kwarg(self, field_definition: FieldDefinition) -> None:
+        """Add a Parameter model for each member of a structured ``query`` kwarg.
+
+        The reserved ``query`` kwarg receives the full mapping of query parameters, so when it is
+        annotated with a structured type every field of that type is itself a query parameter and
+        is validated as one at runtime. Documenting the annotation as a single parameter would be
+        wrong, so it is decomposed into one parameter per field instead.
+
+        Args:
+            field_definition: The field definition of the ``query`` kwarg.
+        """
+        if field_definition.is_any or field_definition.is_mapping:
+            # ``query`` is unannotated or annotated as a plain mapping, so there are no known
+            # parameter names to document. Returning before touching the schema creator keeps
+            # this the no-op it has always been for those annotations.
+            return
+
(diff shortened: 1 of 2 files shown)


provinzkraut:
This adds documentation for a feature that does not exist. Closing and blocking @amantu-qbit for submitting AI generated nonsense

PR #4978, opened 2026-08-11
fix(channels): stop the psycopg listener without cancelling it

Kumzy:
Cancelling a task parked in psycopg's `wait()` is caught by its interrupt handling, which issues a server-side cancel and then waits again while the task is already cancelling. When that does not resolve, the task stays in the cancelling state and awaiting it never returns, so `unsubscribe()` and `on_shutdown()` hang.

This should fix the issue with the 3.11 tests suite getting time out after 10 minutes. 

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4978">https://litestar-org.github.io/litestar-docs-preview/4978</a>

### DIFF
--- a/litestar/channels/backends/psycopg.py
+++ b/litestar/channels/backends/psycopg.py
@@ -12,6 +12,12 @@ from litestar.channels.backends.base import ChannelsBackend
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, Iterable
 
+_LISTEN_POLL_INTERVAL = 0.1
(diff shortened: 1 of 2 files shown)


provinzkraut (approved):
Let's give it a try. Can you also backport this to `v2`?

Kumzy:
> Let's give it a try. Can you also backport this to `v2`?

I will need to see how to do that once it's validated but sure

Issue #3459, opened 2024-05-02
Bug: [OpenAPI] Upload file is not possible when defined as Optional


vkuznetsovgn:
### Description

When defining an `UploadFile` as `Optional` the upload file button disappears.
Removing the `Optional` makes the button to be present but there's no way not to specify the file anymore as there's no `send empty value` checkbox in this case.

![Screenshot from 2024-05-02 11-21-23](https://github.com/litestar-org/litestar/assets/117752877/f741cbf0-8b72-4a07-9b51-7c9b1b4f1f3d)

### MCVE

```python
from typing import Annotated, Optional
from pydantic import BaseModel, ConfigDict
from litestar.datastructures import UploadFile
from litestar.enums import RequestEncodingType
from litestar.params import Body

class FormData(BaseModel):
    model_config = ConfigDict(arbitrary_types_allowed=True)
    
    file: Optional[UploadFile]
    some_params: dict


@post("/test")
async def handle_file(data: Annotated[FormData, Body(media_type=RequestEncodingType.MULTI_PART)]) -> None:
    print(data.file)


app = Litestar(route_handlers=[handle_file])
```

### Litestar Version

2.8.2

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

peterschutt:
I believe a dup of this: https://github.com/litestar-org/litestar/issues/3411

vkuznetsovgn:
@peterschutt Possibly but I'm not sure as the OpenAPI UI looks differently in your case. Maybe we could merge two issues.

Issue #3407, opened 2024-04-19
Bug: cant upload one file only if type is list[UploadFile] | None


euri10:
### Description

I though I had the same issue as #2939 but it's slightly different, here is a test case that fails for one file and not for 2 files:
```
@dataclass
 class OFiles:
    file_list: Optional[List[UploadFile]]


@pytest.mark.parametrize("file_count", (1, 2))
def test_upload_multiple_ofiles_in_model(file_count: int) -> None:
    @post("/")
    async def handler(data: OFiles = Body(media_type=RequestEncodingType.MULTI_PART)) -> None:
        assert len(data.file_list) == file_count

        for file in data.file_list:
            assert await file.read() == b"1"

    with create_test_client([handler]) as client:
        files_to_upload = [("file_list", b"1") for _ in range(file_count)]
        response = client.post("/", files=files_to_upload)

        assert response.status_code == HTTP_201_CREATED

```

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
"![SCREENSHOT_DESCRIPTION](SCREENSHOT_LINK.png)"
```

### Litestar Version

2.8.2final0

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows
- [ ] Other (Please specify in the description above)

---
> [!NOTE]  
> While we are open for sponsoring on [GitHub Sponsors](https://github.com/sponsors/litestar-org/) and 
> [OpenCollective](https://opencollective.com/litestar), we also utilize [Polar.sh](https://polar.sh/) to engage in pledge-based sponsorship.
>
> Check out all issues funded or available for funding [on our Polar.sh dashboard](https://polar.sh/litestar-org)
> * If you would like to see an issue prioritized, make a pledge towards it!
> * We receive the pledge once the issue is completed & verified
> * This, along with engagement in the community, helps us know which features are a priority to our users.

<a href="https://polar.sh/litestar-org/litestar/issues/3407">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/litestar-org/litestar/issues/3407/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/litestar-org/litestar/issues/3407/pledge.svg">
</picture>
</a>

JacobCoffee:
[...]

PR #4982, opened 2026-08-13
fix(openapi): preserve user metadata in schema generation

### LINKED ISSUE #3411: Bug: `| None` removes the `Upload File` button in OpenAPI schema docs
May need to be a separate issue, but `| None` also removes the `Upload File` button in OpenAPI schema docs :(

```py
from litestar import Controller, Litestar, post


class Files(Controller):
    @post(path="/files/upload")
    async def upload_files(
        self,
        data: UploadFile | None = Body(media_type=RequestEncodingType.MULTI_PART),
    ) -> str:
        return data


app = Litestar(route_handlers=[Files])
```

Without `| None`
image

With `| None`

image

_Originally posted by @JacobCoffee in https://github.com/litestar-org/litestar/issues/3407#issuecomment-2067336772_

### LINKED ISSUE #4002: Bug: Setting the title of an Enum (+ msgspec Struct) does not work (in terms of OpenAPI gen)
### Description

Setting the title of an Enum with `Body()` or `msgspec.Meta` as shown below will not allow it to be shown in the autogenerated OpenAPI spec. **Even if you try setting __schema_name__**, it will not work.
```
class UniversalClassificationScoringMethod(StrEnum):
    AUTO = "auto"
    CHUNK_MAX = "chunk_max"
    CHUNK_AVG = "chunk_avg"
    CHUNK_MIN = "chunk_min"


UniversalClassificationScoringMethod.__schema_name__ = "Universal classification scoring method"

UniversalClassificationScoringMethod = Annotated[
    UniversalClassificationScoringMethod,
    Meta(
        title="Universal classification scoring method",
    ),
    Body(
        title="Universal classification scoring method",
        description="A method for producing an overall confidence score for a universal classification.",
        examples=[Example(value=UniversalClassificationScoringMethod.AUTO)],
    ),
]
```

**This also applies to msgspec Structs, however, __schema_name__ does at least still work.**

_No response_

### MCVE

```python
# Your MCVE code here
```

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Screenshots

```bash
```

### Logs

```bash

```

### Litestar Version

2.14.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Kumzy:
Optional file uploads (`UploadFile | None`, `list[UploadFile] | None`, …) were losing their multipart object schema because the optional branch dropped the field's kwarg context, leaving a `oneOf: [..., null]` that Swagger UI can't render as a file input. They now keep the normal multipart schema, and optionality is expressed through `requestBody.required: false` instead — which also means optional `data` parameters are no longer documented as required.

Also fixes enum schemas ignoring user-supplied metadata: `Parameter(title=..., description=...)` on an enum was being overwritten by the auto-generated class-name title. User values now win, with the class name kept as the default.

Fixes #3411, fixes #4002. Also covers the schema side of #3407 and the handler-level part of #3459 (optional uploads inside model fields still need the nullable-encoding rework).

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4982">https://litestar-org.github.io/litestar-docs-preview/4982</a>


litestar/_openapi/schema_generation/schema.py
provinzkraut: Couldn't this be folded into this line?

provinzkraut (approved):

PR #4984, opened 2026-08-13
fix(openapi): fix dangling schema refs and stale schema caches

Kumzy:
`SchemaCreator.not_generating_examples` created the derived creator without passing `schema_registry`, so component schemas created through it (pagination cursor types, constrained collection items) registered into a throwaway registry, leaving dangling `$refs` in the document when `create_examples=True`.

`OpenAPIPlugin.receive_route` reset only the OpenAPI model cache, not the serialized dict cache, so `/schema/openapi.json` kept serving a stale document after runtime route registration while `app.openapi_schema` was fresh.

`Litestar.update_openapi_schema()` built a new document and discarded the result, making it a no-op. It now invalidates both caches so the next access rebuilds lazily through `provide_openapi()`, matching how `receive_route` behaves; it still raises `ImproperlyConfiguredException` when no OpenAPI config is set.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4984">https://litestar-org.github.io/litestar-docs-preview/4984</a>


litestar/app.py
```diff
@@ -983,7 +983,10 @@ def update_openapi_schema(self) -> None:
         Returns:
             None
         """
-        self.plugins.get(OpenAPIPlugin)._build_openapi()
+        plugin = self.plugins.get(OpenAPIPlugin)
```
Kumzy, line 995: Removed the `_build_openapi` and invalidated `_openapi` and  `_openapi_schema` because there are rebuilds on the next access automatically

litestar/_openapi/schema_generation/schema.py
```diff
@@ -253,7 +253,9 @@ def not_generating_examples(self) -> SchemaCreator:
         """Return a SchemaCreator with generate_examples set to False."""
         if not self.generate_examples:
             return self
-        return type(self)(generate_examples=False, plugins=self.plugins, prefer_alias=False)
+        return type(self)(
```
Kumzy, line 260: if you remove `schema_registry` there and pass `generate_examples=True`, you got `$refs$` broken in the final document

litestar/_openapi/plugin.py
```diff
@@ -90,6 +90,12 @@ def _build_openapi(self) -> OpenAPI:
         openapi.components.schemas = context.schema_registry.generate_components_schemas()
         return openapi
 
+    def invalidate_schema_cache(self) -> None:
+        if self._openapi_config is None:
+            raise ImproperlyConfiguredException("OpenAPIConfig not initialized")
+        self._openapi = None
+        self._openapi_schema = None
```
Kumzy, line 97: I added there the invalidation of the `_openapi_schema` too, there was just `_open_api` invalidation and created a method that we can reuse if needed

litestar/app.py
```diff
@@ -983,7 +983,7 @@ def update_openapi_schema(self) -> None:
         Returns:
             None
         """
-        self.plugins.get(OpenAPIPlugin)._build_openapi()
+        self.plugins.get(OpenAPIPlugin).invalidate_schema_cache()
```
provinzkraut: Slight change in behaviour: If there's a bug in the schema, the old behaviour would have raises here. IMO this method *should* rebuild the schema, as that's also what it says it does

Kumzy: Ok changed so it rebuild it and raises if there is an error, can you tell me if it's better like that ? @provinzkraut

provinzkraut (approved):

Kumzy (commented):
Good to merge

PR #4988, opened 2026-08-14
fix(dto): apply msgspec constraints to the constrained union member

### LINKED ISSUE #4181: Bug: CodegenBackend rewrites field type incorrectly, causing msgspec error
### Description

I encountered an error when trying to create a custom DTO that deserializes a msgspec.Struct and combines it with additional information from the connection.

Here’s my struct:
```
class UpsertUser(msgspec.Struct, kw_only=True):
    username: Annotated[
        str, msgspec.Meta(min_length=3, max_length=32, pattern=r"^[a-zA-Z0-9_]+$")
    ]
    password: Union[
        Annotated[str, msgspec.Meta(min_length=8, max_length=40)], UnsetType
    ] = UNSET
```
Note that the msgspec.Meta must be associated with `Annotated[str, ...]`. It cannot be applied to something like `Union[str, UnsetType, msgspec.Meta(...)]`, otherwise msgspec will complain.

However, when using `MsgspecDTO[UpsertUser]`, the codegen backend rewrites the password type as:

`typing.Annotated[typing.Union[str, msgspec.UnsetType], msgspec.Meta(min_length=8, max_length=40, examples=[])]`

This transformation causes an error in msgspec, since Meta is not allowed in that position.

_No response_

### MCVE

```python

```

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.16.0

### Platform

- [ ] Linux
- [x] Mac
- [ ] Windows

NoiceHax:
Transfer models wrap optional and partial fields in a union with `None` or `UnsetType`, but constraints copied from the source model were attached to the union itself, giving annotations like `Annotated[Union[str, UnsetType], Meta(min_length=8)]`. msgspec only accepts type specific constraints on the type they apply to and raises a TypeError for that, so any DTO with a constrained optional field or `partial=True` failed to build.

The meta now goes on the single constrainable member of the union. If a union has more than one member that could take constraints there is nowhere obvious to put it, so it stays on the union as before. Partial DTOs no longer need to drop `min_length` and `max_length`, which was a workaround for the same problem.

The old workaround test is replaced by a parametrized one covering each constraint on both backends, and I added tests for `Union[Annotated[str, Meta], UnsetType]` and `Optional[Annotated[str, Meta]]`. They are in tests/unit/test_dto/test_factory/test_integration.py and fail without the change.

Closes #4181

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4988">https://litestar-org.github.io/litestar-docs-preview/4988</a>

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -15,7 +15,6 @@ from typing import (
     Protocol,
     Union,
     cast,
-    get_args,
 )
 
 import msgspec
@@ -38,9 +37,8 @@ from litestar.enums import RequestEncodingType
 from litestar.params import KwargDefinition
 from litestar.serialization import decode_json, decode_msgpack
 from litestar.types import Empty
-from litestar.types.builtin_types import NoneType
 from litestar.typing import FieldDefinition
-from litestar.utils import is_union, unique_name_for_scope
+from litestar.utils import unique_name_for_scope
 
 if TYPE_CHECKING:
     from litestar.connection import ASGIConnection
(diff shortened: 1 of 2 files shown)


litestar/dto/_backend.py
```diff
@@ -837,7 +855,7 @@ def _create_struct_for_field_definitions(
 
         if field_definition.passthrough_constraints:
             if (field_meta := _create_struct_field_meta_for_field_definition(field_definition)) is not None:
-                field_type = Annotated[field_type, field_meta]
+                field_type = _annotate_with_meta(field_type, field_meta)
```
provinzkraut: Wouldn't it be easier to check `field_definition.is_partial` than do the type unwrapping? iiuc this case only happens for partial fields where we *add* an outer `Union`. In other cases, we do not alter the type, so a wrongly attached `Meta` is a user error, which we should not silently fix.

PR #4989, opened 2026-08-14
fix(routing): raise on paths that only differ in their path parameters

### LINKED ISSUE #3622: Bug: Routing based on path parameter data type is not using first matching route handler
### Description

Was told Litestar is intended to route differently based on the type of the path parameter:
> Q: Is Litestar intended to route differently based on the type of the path parameter
> A:  ... this only works if the parameters you pass in are actually compatible and distinct, e.g. using a str is basically a catch-all, so if you provide that last, it will always match

However, when providing route handlers in the order of `int path param handler` followed by `str path param handler`, the `str path param handler` is the only one that is called in the example code below.

_No response_

### MCVE

```python


@get("/{id:int}")
async def hello_world_int(id: int) -> str:
    return f"Hello, world (int)! {id}"


@get("/{id:str}")
async def hello_world_str(id: str) -> str:
    return f"Hello, world (str)! {id}"

# Expect that path "/42" would route to hello_world_int
# and path "/abc" would go to hello_world_str
app = Litestar([hello_world_int, hello_world_str])
```

### Steps to reproduce

```bash
1. Run app: `litestar run`
2. In browser access: 127.0.0.1:8000/42
3. Result: Hello, world (str)! 42  
  Error: (expected to see "(int)" rather than "(str)"
4. In browser access: 127.0.0.1:8000/ABC
5. Result (as expected): Hello, world (str)! ABC
```

### Screenshots

_No response_

### Logs

_No response_

### Litestar Version

2.9.1

### Platform

- [X] Linux
- [ ] Mac
- [ ] Windows

NoiceHax:
The routing trie collapses every path parameter at a given position into one node, so `/{id:int}` and `/{id:str}` resolve to the same node. Whichever route got registered last silently overwrote the earlier handler and made it unreachable. The issue discussion settled on raising an error instead of routing by parameter type, so registering paths like that now raises `ImproperlyConfiguredException`.

The check runs per method, so two paths that share a node but register different methods still work. OPTIONS is skipped. Litestar generates an OPTIONS handler for every path that does not declare one, so those handlers collide on a shared node even though no user handler is being shadowed. Re-registering the exact same route is still fine, which matters when the trie gets rebuilt after handlers are added to an app that already exists.

Tests are in tests/unit/test_asgi/test_routing_trie/test_mapping.py. I also added a note to the path parameter docs.

Closes #3622

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4989">https://litestar-org.github.io/litestar-docs-preview/4989</a>

### DIFF
--- a/docs/usage/routing/parameters.rst
+++ b/docs/usage/routing/parameters.rst
@@ -175,6 +175,15 @@ The following types are supported in the ``{name:type}`` slot:
 * ``timedelta``: Accepts duration strings compatible with the standard (Pydantic/Msgspec) timedelta formats.
 * ``uuid``: Accepts all uuid values.
 
+.. important::
+    Routing does not take the name or the type of a path parameter into account, so
+    ``/{id:int}`` and ``/{id:str}`` are the same route as far as routing is concerned;
+    the declared type only determines how the captured value is coerced. Registering
+    handlers for both would leave one of them unreachable and therefore raises an
+    :exc:`~.exceptions.ImproperlyConfiguredException`. To serve several value types
+    under the same path, use a single handler with the widest slot type and branch on
+    the value inside it.
+
 Query parameters
 ~~~~~~~~~~~~~~~~
 
--- a/litestar/_asgi/routing_trie/mapping.py
+++ b/litestar/_asgi/routing_trie/mapping.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Literal, cast
 
 from litestar._asgi.routing_trie.types import (
     ASGIHandlerTuple,
@@ -9,6 +9,7 @@ from litestar._asgi.routing_trie.types import (
     create_node,
 )
 from litestar._asgi.utils import wrap_in_exception_handler
+from litestar.exceptions import ImproperlyConfiguredException
 from litestar.types.internal_types import PathParameterDefinition
 
 __all__ = ("add_mount_route", "add_route_to_trie", "build_route_middleware_stack", "configure_node")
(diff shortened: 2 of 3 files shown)


provinzkraut:
Thinking about this again, this does introduce some inconsistencies: You normally do not get an exception when you override a path, but in this special case you'll now get one. IMO we should either always raise / warn there or never.

*If* we were to do this, my vote would be to emit a warning for every handler which overrides another, unless the handler explicitly states that it intends to override another handler (e.g. by setting `overrides="other_handler_name"` to the handler decorator).

@litestar-org/members thoughts on this?

NoiceHax:
Agreed, and thanks for pushing back on this — the inconsistency you describe is real and it is a good reason not to merge this as written.

To restate it so we are working from the same thing: overriding a path is normally silent, and this PR would make one particular shape of override loud. That is a worse rule than either of the consistent alternatives.

Of your two options I think the warning-with-opt-out is clearly the better one, for the reason you gave: the silent case is the one that actually bites people, and a path that differs only in its parameter *names* is the shape most likely to be accidental rather than deliberate. Raising unconditionally would break working applications on upgrade, which seems out of proportion.

So: happy to rework this into a warning on every override, with an explicit opt-out on the handler, if the team lands there. Two things I would want settled first, since they change the implementation rather than just the message:

1. Whether the opt-out is `overrides="other_handler_name"` as you sketched, or a plain boolean. The named form is better documentation and lets the framework verify the override actually hit the handler you meant, but it needs handler names to be reliably resolvable at registration time.
2. Whether this warns at registration or at first resolution — registration is more useful, but it is also where partial application state is least settled.

I am in no rush on this one, and I would rather it waited for a decision than merged inconsistent. If the team prefers to close it and track the broader override-warning behaviour as its own issue, that is a perfectly good outcome too — say the word and I will close it and open that issue instead.

This PR is also behind `main` now; I will rebase once there is a direction to rebase toward.

provinzkraut:
@NoiceHax please do no communicate using AI in the review process. This is clearly stated in our AI policy. 

There is no value added by having you as a proxy between a maintainer an the LLM you are using, I could be having that conversation with the AI myself just as well.

PR #4996, opened 2026-08-16
feat(stores/redis): use GETEX on Redis >= 6.2 for get+renew

### LINKED ISSUE #4993: Enhancement: Use redis `GETEX` in `RedisStore.get` for `renew_for` functionality on redis 6.2+
### Summary

`RedisStore.get` predates Redi 6.2, which introduced [`GETEX`](https://redis.io/docs/latest/commands/getex/) to get a value and set its expiry in one step. On supported redis versions, `GETEX` should be used instead of the custom script.

Karunasagar12:
## Summary

Closes #4993 — on Redis 6.2+, `GETEX` atomically gets a value and sets its expiry in a single command, eliminating the Lua script overhead.

## Changes

**`litestar/stores/redis.py`:**

1. Added `_check_getex_support()` — lazily detects Redis server version via `INFO server` on first `get()` call with `renew_for`. Result is cached for the lifetime of the store instance.
2. `get()` now uses `GETEX key EX <seconds>` when Redis >= 6.2, falling back to the existing Lua script on older versions.
3. **Bug fix:** Changed `renew_for.seconds` to `int(renew_for.total_seconds())`. The old code used `timedelta.seconds` which only returns the seconds *component*, not total seconds — `timedelta(days=1, seconds=5).seconds == 5`, not 86405.

**`tests/unit/test_stores.py`:**

| Test | What it verifies |
|------|-----------------|
| `test_get_and_renew_redis_uses_getex` | Version detection works, GETEX path is taken, TTL is renewed |
| `test_get_and_renew_redis_lua_fallback` | Lua script still works when GETEX is unsupported |
| `test_get_renew_for_timedelta_uses_total_seconds` | timedelta conversion uses total_seconds() |

## Backward Compatibility

- Redis < 6.2: falls back to existing Lua script (no behavior change)
- Redis >= 6.2: same semantics, fewer round-trips
- Version check uses a single cached `INFO server` call (negligible cost)

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4996">https://litestar-org.github.io/litestar-docs-preview/4996</a>

### DIFF
--- a/litestar/stores/redis.py
+++ b/litestar/stores/redis.py
@@ -27,6 +27,7 @@ class RedisStore(NamespacedStore):
         "_delete_all_script",
         "_get_and_renew_script",
         "_redis",
+        "_supports_getex",
         "handle_client_shutdown",
     )
 
@@ -45,8 +46,9 @@ class RedisStore(NamespacedStore):
         self._redis = redis
         self.namespace: str | None = value_or_default(namespace, "LITESTAR")
         self.handle_client_shutdown = handle_client_shutdown
+        self._supports_getex: bool | None = None  # lazy-detected on first use
 
-        # script to get and renew a key in one atomic step
+        # script to get and renew a key in one atomic step (fallback for Redis < 6.2)
         self._get_and_renew_script = self._redis.register_script(
             b"""
         local key = KEYS[1]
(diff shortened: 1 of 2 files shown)


sobolevn (commented):
I think 3.0.0 is the right time to drop older redis versions support.

6.0 https://redis.io/docs/latest/operate/rs/release-notes/rs-6-0-may-2020/ was released 6 years ago.
and 6.1 never existed.

Karunasagar12:
Good call — updated. Dropped the Lua script and version detection entirely. `get()` now just calls `GETEX` directly.

Also fixed a pre-existing bug while I was in there: the old code used `timedelta.seconds` (which only returns the seconds *component*) instead of `int(total_seconds())`. So `renew_for=timedelta(days=1)` was silently renewing for 0 seconds.

sobolevn:
This would require a deprecation message in 2.x branch

provinzkraut:
> I think 3.0.0 is the right time to drop older redis versions support.
> 
> 6.0 https://redis.io/docs/latest/operate/rs/release-notes/rs-6-0-may-2020/ was released 6 years ago. and 6.1 never existed.

I'm still stuck at 6.0 at work.. :grimacing: 
And I'm sure I'm not the only one. I'd rather we keep it unless there's a good reason against dropping it?

Karunasagar12:
Restored the Lua fallback — GETEX is now used only when Redis >= 6.2 (probed once per store instance, result cached). Older servers fall back to the existing Lua script unchanged.

Also kept the `total_seconds()` fix: the old `.seconds` attribute on a `timedelta(days=1)` returns 0, not 86400, which would silently renew for 0 s.

Added unit tests for both paths and the timedelta edge case (no live Redis needed).

Issue #4009, opened 2025-02-16
Litestar v3 roadmap


provinzkraut:
### Litestar v3 is coming soon*! 🚀

We’re thrilled to announce that Litestar v3 is on the horizon, bringing a host of exciting new features, improvements, and some breaking changes to make your development experience even better. Here’s what you can expect

### Things to do for v3

- [x] Remove `StaticFileConfig`
- [x] Drop support for implicit Optional Default Parameters
- [x] Replace OpenAPI Controller with Plugins
  + Elimination of `OpenAPIController` Subclassing
  + Changes to Endpoint Configuration
  + Modification to `root_schema_site` Handling
- [x] Remove deprecated `app` parameter for `Response.to_asgi_response`
- [x] Remove deprecated scope state utilities
- [x] Remove deprecated utility function `is_sync_or_async_generator`
- [x] Remove of semantic HTTP route handler classes in favour of functional style decorators
- [x] Remove of deprecated `litestar.middleware.exceptions` module and `ExceptionHandlerMiddleware`
- [x] Overhaul route handler configuration (https://github.com/litestar-org/litestar/pull/3900)
- [ ] Remove `litestar.contrib` namespace
- [ ] Improve "file system" abstractions
- [ ] Remove all deprecated items scheduled for removal in 3.0
- [ ] Remove pagination 
- [ ] Don't use `__future__.annotations, only use stringized annotations where strictly necessary / clean up signature namsepaces
- [ ] Deprecate plugin protocols in favour of abstract base classes
- [ ] Remove deprecated plugin protocols
- [ ] Use native event loop in `AsyncTestClient`
- [ ] Drop Python 3.8 support
- [ ] Add CLI `shell` command
- [ ] Improve and simplify logging interface (https://github.com/litestar-org/litestar/issues/2858)
- [ ] Update sentry integration
- [ ] Add migration guide :)

<hr>

We will update this issue with information about the release, how to prepare, upcoming pre-releases etc.

If you've got any questions, remarks or additions, this is the place to leave them :)

marcuslimdw:
would it be possible to have more standardisation of acronym capitalisation rules?

`@HTTPRouteHandler(..., http_method=[HttpMethod.GET, HttpMethod.POST])` kind of grinds my gears.

(my preference is for only initial caps i.e. `HttpMethod` because that's more easily scannable/parsable)

provinzkraut:
> would it be possible to have more standardisation of acronym capitalisation rules?
> 
[...]

PR #4998, opened 2026-08-17
refactor(middleware): migrate `CSRFMiddleware` to `ASGIMiddleware`

Kumzy:
Part of #4009 

Moving `CSRFMiddleware` from `MiddlewareProtocol` to `ASGIMiddleware` and into `litestar.middleware._internal`, removing the `litestar.middleware.csrf` module from the public API.

Constructor takes keywords arguments now instead of a `CSRFConfig` object.

Exclusion (`exclude` / `exclude_from_csrf`) is now resolved at startup via `should_bypass_for_handler`, so `CSRFConfig.exclude` patterns match the handler's path template instead of the request path.

Added some tests:
- One to test that the default are the same as the `CSRFConfig`
- One that checks the cookie settings from the config (path, domain, secure, etc.) actually show up on the cookie the app sends back
- One that checks that changing which HTTP methods are "safe" in the config really changes what gets blocked
- One checks that excluding a route from CSRF works with the route pattern like `/user/{user_id:int}`, not the full URL like `/user/1`. It was failing before.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4998">https://litestar-org.github.io/litestar-docs-preview/4998</a>


litestar/_asgi/routing_trie/mapping.py
```diff
@@ -203,7 +203,21 @@ def build_route_middleware_stack(
         asgi_handler = wrap_in_exception_handler(app=asgi_handler)
 
         if app.csrf_config:
-            asgi_handler = CSRFMiddleware(app=asgi_handler, config=app.csrf_config)
+            csrf_middleware = CSRFMiddleware(
```
provinzkraut: This gets a bit noisy with all the config passing in here. Maybe a `from_config` classmethod on the middleware would make it easier? Or passing the config to the middleware would also be easier IMO for these internal only middlewares

Kumzy: Implemented it, will look to implement it also in the other middleware (case by case) because in the compression one, we kept the ``config`` in the ``__init__``

PR #4999, opened 2026-08-17
refactor(middleware): migrate `CompressionMiddleware` to `ASGIMiddleware`

Kumzy:
Part of #4009, trying stacked PR.

Moving `CompressionMiddleware` from `AbstractMiddleware` to `ASGIMiddleware`.

The constructor takes keyword arguments now instead of an `app` and a `CompressionConfig` object.
I changed also the `CompressionFacade` since it consumed the config object too so the protocol changes too: facades now receive `backend_config` directly instead of the whole `CompressionConfig`. Custom facades just rename their `config` parameter to `backend_config`. For the built-in backends, `CompressionConfig` derives `backend_config` from its existing fields, so configuring compression through `CompressionConfig` is unchanged.

Same breaking change as #4998: `exclude` patterns now match the handler's path template at startup instead of the request path (documented in the changelog and whats-new-3).

Added some tests:
- One that checks `exclude_opt_key` really disables compression for a route.
- One that checks `minimum_size` from the config is really applied.
- One that checks the gzip/brotli compression levels from the config really change the output
- One that checks `backend_config` is used when gzip is the main backend. It was failing before
- One that checks mounted ASGI routes still get compressed
- One that checks the gzip fallback uses the configured gzip level

Not sure about the `CompressionFacade` changes (that it was the best way to do it) so any feedback is welcome

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/4999">https://litestar-org.github.io/litestar-docs-preview/4999</a>


docs/release-notes/changelog.rst
```diff
@@ -36,6 +36,62 @@
...
+
+            # after
+            middleware = CompressionMiddleware(
+                facade=config.compression_facade,
+                backend_config=config.backend_config,
+                gzip_fallback=config.gzip_fallback,
+                gzip_backend_config={"compress_level": config.gzip_compress_level},
+                minimum_size=config.minimum_size,
+                exclude=config.exclude,
+                exclude_opt_key=config.exclude_opt_key,
```
provinzkraut: I think this should just set the values directly, instead of using the config, as that's the intended purpose

provinzkraut (commented):
Yeah, I think we might not want to do the config refactor for this one, as we expose it to the facade, which users might depend on. No need to make this more breaking

Kumzy:
> Yeah, I think we might not want to do the config refactor for this one, as we expose it to the facade, which users might depend on. No need to make this more breaking

So just moving it to ASGIMiddleware without touching the rest?

Kumzy:
I did not implement the ``from_config``there because ``config`` is still in the ``__init__``. What should we do?

PR #5001, opened 2026-08-17
refactor(middleware): migrate `AllowedHostsMiddleware` to `ASGIMiddleware`

Kumzy:
Part of #4009.

Moving `AllowedHostsMiddleware` from `AbstractMiddleware` to `ASGIMiddleware`.

The constructor takes keyword arguments now instead of an `app` and an `AllowedHostsConfig` object. Configuring allowed hosts through `AllowedHostsConfig` is unchanged.

One subtlety this middleware has that the others don't: `AllowedHostsConfig.scopes` is user-configurable, and mounted ASGI apps receive both `http` and `websocket` connections at runtime, so a startup-only check can't fully replace the old per-request scope check. Configured scopes are therefore also enforced per connection via `should_bypass_for_scope`, without this `scopes={"websocket"}` would have silently stopped host-checking websocket connections through ASGI mounts.

Added some tests:
- One that checks websocket connections through mounted ASGI apps are still host-checked with `scopes={"websocket"}`. It was failing before
- One that checks websocket connections through mounted ASGI apps are NOT host-checked with `scopes={"http"}`. It was failing before (just other direction than the previous test)
- One that checks the `scopes` setting from the config is really applied
- One that checks an empty `scopes` set falls back to the defaults instead of disabling the check
- One that checks `exclude_opt_key` really disables the host check for a route
- One that checks ASGI route handlers get host-checked like before

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5001">https://litestar-org.github.io/litestar-docs-preview/5001</a>


Kumzy:
Implemented the ``from_config`` to reflect the changes done in ``CSRFMiddleware``

PR #4987, opened 2026-08-13
fix(channels): psycopg listener lifecycle + graceful stop (backport `v2`)

Kumzy:
Backports #4967 and #4978. The second builds on the listener-task lifecycle the first introduces, so they come as a pair.

The CI test were not passing already on the `v2` branch so not sure what to do

PR #5002, opened 2026-08-17
fix(tests): restore optional gender and lucky_number openapi params (`v2` CI)

Kumzy:
Stacked on #4987. Fixes the two failures keeping `v2` CI red.

The typescript converter test: the #4750 cherry-pick dropped `required=False` from the `gender` and `lucky_number` query params in the openapi test conftest, and adjusted the assertions to match instead of restoring them.

The psycopg channels tests on 3.8/3.9: the backend creates its `asyncio.Lock` in `__init__`, and on ≤3.9 a lock binds to the loop active at creation (lazy binding only came in 3.10). The tests create the backend on one loop and run the app on the `TestClient` portal loop, so a contended acquire during shutdown blows up with "got Future attached to a different loop". Now recreated in `on_startup` on the running loop, like the event queue already is.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5002">https://litestar-org.github.io/litestar-docs-preview/5002</a>

### DIFF
--- a/litestar/channels/backends/psycopg.py
+++ b/litestar/channels/backends/psycopg.py
@@ -35,8 +35,6 @@ class PsycoPgChannelsBackend(ChannelsBackend):
     async def on_startup(self) -> None:
         self._exit_stack = AsyncExitStack()
         self._event_queue = asyncio.Queue()
-        # Created here rather than in __init__ because on Python <= 3.9 it binds to the event loop
-        # active at creation time
         self._listener_lock = asyncio.Lock()
         self._shutting_down = False
         self._listener_conn = await AsyncConnection[Any].connect(self._pg_dsn, autocommit=True)
--- a/litestar/channels/backends/psycopg.py
+++ b/litestar/channels/backends/psycopg.py
@@ -21,12 +21,12 @@ _STOP_LISTENER_TIMEOUT = 5.0
 
 class PsycoPgChannelsBackend(ChannelsBackend):
     _listener_conn: AsyncConnection[Any]
+    _listener_lock: asyncio.Lock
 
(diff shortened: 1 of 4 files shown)


litestar/channels/backends/psycopg.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, pg_dsn: str) -> None:
     async def on_startup(self) -> None:
         self._exit_stack = AsyncExitStack()
         self._event_queue = asyncio.Queue()
+        # Recreate the lock because on Python <= 3.9 it binds to the event loop active at creation time
```
provinzkraut: Should just remove the lock from `__init__` then

Kumzy: Done

tests/unit/test_channels/test_psycopg_backend.py
```diff
@@ -47,6 +47,7 @@ async def test_subscription_mutations_are_serialized() -> None:
     backend = PsycoPgChannelsBackend("postgresql://unused")
     connection = _ConcurrentConnection()
     backend._listener_conn = connection  # type: ignore[assignment]
+    backend._listener_lock = asyncio.Lock()  # normally created by on_startup, which needs a real connection
```
provinzkraut: This should call `on_startup` then

Kumzy: Removed that and instead dded a fixture that run on_startup and on_shutdown to emulate the real lifecycle instead of assigning a lock by hand

provinzkraut (approved):

PR #5005, opened 2026-08-18
refactor(middleware): migrate `RateLimitMiddleware` to `ASGIMiddleware`

Kumzy:
Part of #4009 (Stacked PR to avoid conflicts in the changelog and whats-new-3 docs)

Moving `RateLimitMiddleware` from `AbstractMiddleware` to `ASGIMiddleware`.

The constructor takes keyword arguments now instead of an `app` and a `RateLimitConfig` object, with each argument defaulting to the matching config field. `RateLimitConfig` itself is unchanged and stays the way to configure this: its `middleware` property now returns a configured `RateLimitMiddleware` instance instead of a `DefineMiddleware`, so `middleware=[config.middleware]` keeps working as before.

One subtlety: `scopes` now filters by handler type at startup, and rate limiting must keep applying to mounted ASGI apps (that's what the `::mount` cache key bucket is for). So the middleware declares `scopes = (HTTP, ASGI)` with a runtime scope-type guard, meaning mounts stay rate limited while websocket connections through them pass through, like before.

Added some tests, each verified to fail when its kwarg is dropped from the `middleware` property:
- One that checks `set_rate_limit_headers=False` really disables the headers. It was not covered before.
- One that checks the four custom header key settings are applied. Dropping them was undetected before.
- One that checks mounted ASGI apps are really rate limited (429 on the second request). The existing mount test couldn't tell "own bucket" apart from "not wrapped at all".
- Extended the custom identity test so a different identity gets its own quota, without this, dropping `identifier_for_request` from the property was undetected.

Also updated the `test_stores` example test that reached into `app.middleware[0].kwargs["config"]`, which no longer exists now that the property returns an instance.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5005">https://litestar-org.github.io/litestar-docs-preview/5005</a>

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -38,6 +38,7 @@
 
     .. change:: Migrate ``CompressionMiddleware`` to ``ASGIMiddleware``
         :type: feature
+        :pr: 4999
         :issue: 4009
         :breaking:
 
@@ -47,8 +48,9 @@
 
         Applications that configure compression through
         :class:`~litestar.config.compression.CompressionConfig` are unaffected, since
-        Litestar constructs the middleware itself. The constructor takes keyword
-        arguments now instead of an ``app`` and a ``CompressionConfig``:
+        Litestar constructs the middleware itself. The middleware still takes the
+        ``CompressionConfig`` - which the ``CompressionFacade`` protocol continues to
+        receive unchanged - but no longer wraps an ``app``:
 
         .. code-block:: python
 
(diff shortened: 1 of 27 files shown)


Kumzy:
Also implemented the ``from_config`` to reflect the changes done in ``CSRFMiddleware``

docs/release-notes/changelog.rst
```diff
@@ -6,6 +6,54 @@
...
+        :pr: 5005
+        :issue: 4009
+        :breaking:
+
+        ``RateLimitMiddleware`` has been moved from the legacy ``AbstractMiddleware``
+        base to :class:`~litestar.middleware.ASGIMiddleware`, as part of migrating all
+        built-in middleware off the legacy bases.
+
+        Applications that configure rate limiting through
+        :class:`~litestar.middleware.rate_limit.RateLimitConfig` and its ``middleware``
```
provinzkraut: I don't think that's the right migration path.

`middleware=[rate_limit_config.middleware]`

should become `middleware=[RateLimitMiddleware(...)]`

docs/release-notes/changelog.rst
```diff
@@ -6,6 +6,54 @@
...
+
+            # after
+            middleware = RateLimitMiddleware(
+                rate_limit=("minute", 10),
+                store=config.store,
+                exclude=config.exclude,
+            )
+            asgi_app = middleware(next_app)
+
+        Two behavioural changes for excluded routes, matching the other migrated
```
provinzkraut:
```suggestion
        Two behavioural changes for excluded routes
```
not relevant

litestar/middleware/rate_limit.py
```diff
@@ -57,40 +57,99 @@ def get_remote_address(request: Request[Any, Any, Any]) -> str:
...
+        self.rate_limit_remaining_header_key = rate_limit_remaining_header_key
+        self.rate_limit_reset_header_key = rate_limit_reset_header_key
+        self.exclude_path_pattern = tuple(exclude) if isinstance(exclude, list) else exclude
+        self.exclude_opt_key = exclude_opt_key
         self._lock = anyio.Lock()
 
-    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        """ASGI callable.
+    @classmethod
+    def from_config(cls, config: RateLimitConfig) -> RateLimitMiddleware:
```
provinzkraut: Don't really think we need this here?

provinzkraut:
Since this is one of the cases where the config becomes obsolete, we should deprecate the `config.middleware` property

litestar/middleware/rate_limit.py
Kumzy: Yes I did not look enough and tried to apply the same thing without looking at the same use

Kumzy:
> Since this is one of the cases where the config becomes obsolete, we should deprecate the `config.middleware` property

I create a deprecation in ``v2`` branch, or is it now for ``4.X`` ? like `handlers/base.py` @provinzkraut

PR #5006, opened 2026-08-18
refactor(middleware): migrate `PrometheusMiddleware` to `ASGIMiddleware`

Kumzy:
Part of #4009. Stacked PR to avoid conflicts in the changelog and whats-new docs.

Moving `PrometheusMiddleware` from `AbstractMiddleware` to `ASGIMiddleware`.

The constructor takes keyword arguments now instead of an `app` and a `PrometheusConfig` object, with each argument defaulting to the matching config field. `PrometheusConfig` itself is unchanged and stays the way to configure this: its `middleware` property now returns a configured `PrometheusMiddleware` instance instead of a `DefineMiddleware`, so `middleware=[config.middleware]` keeps working as before. The `excluded_http_methods` check stays inside `handle()`.

Same subtlety as allowed-hosts: `PrometheusConfig.scopes` is user-configurable and mounted ASGI apps receive both `http` and `websocket` connections at runtime, so configured scopes are also enforced per connection via `should_bypass_for_scope`. Without this, restricting to `scopes={"websocket"}` would have silently kept instrumenting http connections through ASGI mounts.

Added some tests, covering the prometheus-specific behavior of this migration:
- One that checks mounted ASGI apps are still instrumented. Dropping `ASGI` from the scopes tuple was undetected before.
- One that checks configured scopes hold for connections through mounts (http through a mount is NOT recorded with `scopes={"websocket"}`). It was failing.
- One that checks `group_path=False` uses the request path instead of the template. The `False` branch was not covered before.
- One that checks a subclass set via `middleware_class` works with the new `handle` contract, since this PR changes that hook (no more `self._config`).
- Extended the buckets assertion so default bucket boundaries must be absent - without this, dropping `buckets` from the property was undetected.

Two pre-existing quirks found while porting, unchanged here (possible follow-ups): `request.method` raises `KeyError` for websocket connections when `excluded_http_methods` is set, and `PrometheusConfig.exclude_unhandled_paths` is a dead field nothing reads.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5006">https://litestar-org.github.io/litestar-docs-preview/5006</a>

### DIFF
--- a/docs/examples/plugins/prometheus/using_prometheus_exporter.py
+++ b/docs/examples/plugins/prometheus/using_prometheus_exporter.py
@@ -1,13 +1,13 @@
 from litestar import Litestar
-from litestar.plugins.prometheus import PrometheusConfig, PrometheusController
+from litestar.plugins.prometheus import PrometheusController, PrometheusMiddleware
 
 
 def create_app(group_path: bool = True) -> Litestar:
     # Default app name and prefix is litestar.
-    prometheus_config = PrometheusConfig(group_path=group_path)
+    prometheus_middleware = PrometheusMiddleware(group_path=group_path)
 
     # By default the metrics are available in prometheus format and the path is set to '/metrics'.
     # If you want to change the path and format you can do it by subclassing the PrometheusController class.
 
-    # Creating the litestar app instance with our custom PrometheusConfig and PrometheusController.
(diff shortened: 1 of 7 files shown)


Kumzy:
Here also implemented the ``from_config`` to reflect the changes done in ``CSRFMiddleware``

PR #5009, opened 2026-08-19
fix(channels): handle exceptions in _pub_worker to prevent crash

### LINKED ISSUE #5008: Bug(channels): single exception in backend.publish breaks all subsequent events publishing
### Description

If single exception(e.g. network failure) happens in the `backend.publish`, `backend._pub_worker` dies, causing ignoring all subsequent events that will be published + deadlock on shutdown(since the plugin tries to empty the queue before shutdown)

_No response_

### MCVE

```python
import asyncio
from unittest.mock import AsyncMock

from anyio import fail_after

    def __init__(self, history: int = 0) -> None:
        super().__init__(history)
        self._counter = 0

    async def publish(self, data: bytes, channels: Iterable[str]) -> None:
        if self._counter == 0:
            raise RuntimeError("Network lag")
        self._counter += 1
        await super().publish(data, channels)


async def test_deadlock_on_shutdown():
    plugin = ChannelsPlugin(arbitrary_channels_allowed=True, backend=StubMemoryChannelBackend())
    with fail_after(1):
        async with plugin:
            plugin.publish("test", "test")
    # TimeoutError (during __aexit__ --> _on_shutdown)


async def test_single_exception_suppress_all_publishes():
    plugin = ChannelsPlugin(arbitrary_channels_allowed=True, backend=StubMemoryChannelBackend())
    async with plugin:
        plugin.publish("test", "test")
        sub = await plugin.subscribe("42")
        plugin.publish("42", "42")  # Stub backend does not break that, but ...
        try:
            await asyncio.wait_for(sub._queue.get(), timeout=1)  # TimeoutError
        except TimeoutError:
            print(
                "Backend available now but event isn't published "
                "because _pub_worker is broken since it doesn't handle exceptions"
            )
            plugin._on_shutdown = AsyncMock() # just to exit from context manager without the deadlock


asyncio.run(test_deadlock_on_shutdown())
asyncio.run(test_single_exception_suppress_all_publishes())
```

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

the main branch(v3)

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Vitaly312:
## Description
Fixes #5008 
Wraps `self._backend.publish` with try-except block to prevent crashing background publishing worker.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5009">https://litestar-org.github.io/litestar-docs-preview/5009</a>

### DIFF
--- a/litestar/channels/plugin.py
+++ b/litestar/channels/plugin.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 from asyncio import CancelledError, Queue, Task, create_task
 from contextlib import AbstractAsyncContextManager, asynccontextmanager, suppress
 from functools import partial
@@ -28,6 +29,9 @@ if TYPE_CHECKING:
     from litestar.params import FromPath
     from litestar.types import LitestarEncodableType, TypeEncodersMap
     from litestar.types.asgi_types import WebSocketMode
+
+
+logger = logging.getLogger(__name__)
 
 
 class ChannelsException(LitestarException):
@@ -306,7 +310,10 @@ class ChannelsPlugin(InitPlugin, AbstractAsyncContextManager):
     async def _pub_worker(self) -> None:
         while self._pub_queue:
             data, channels = await self._pub_queue.get()
-            await self._backend.publish(data, channels)
+            try:
+                await self._backend.publish(data, channels)
+            except Exception as exc:
+                logger.exception("Error while publishing event: %s", exc)
             self._pub_queue.task_done()
 
     async def _sub_worker(self) -> None:
--- a/tests/unit/test_channels/test_plugin.py
+++ b/tests/unit/test_channels/test_plugin.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 
 import asyncio
 from secrets import token_hex
-from typing import cast
-from unittest.mock import AsyncMock, MagicMock
+from typing import TYPE_CHECKING, cast
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from _pytest.fixtures import FixtureRequest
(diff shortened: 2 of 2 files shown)


litestar/channels/plugin.py
```diff
@@ -306,7 +310,10 @@ async def ws_handler_func(socket: WebSocket) -> None:
     async def _pub_worker(self) -> None:
         while self._pub_queue:
             data, channels = await self._pub_queue.get()
-            await self._backend.publish(data, channels)
+            try:
+                await self._backend.publish(data, channels)
+            except Exception as exc:
+                logger.exception("Error while publishing event: %s", exc)
```
provinzkraut, line 316: This should use the Litestar logger if anything

Vitaly312: I followed the pattern [litestar/events/listener.py](https://github.com/litestar-org/litestar/blob/64cd7da34afc9934c5438eaa1c6f297e4ef16f32/litestar/events/listener.py#L74-L75), since I didn't find any global logging module in the main branch. So could you clarify how it should be done? Should I add `logger` param in the channels plugin constructor like in the [`LoggingMiddleware`](https://github.com/litestar-org/litestar/blob/64cd7da34afc9934c5438eaa1c6f297e4ef16f32/litestar/middleware/logging.py#L50)?

provinzkraut: mhm. well, it shouldn't be done like this there either :sweat_smile: 
the issue is that this will log in a blocking way, unless the user specifically configures something else. logging is a bit of a difficult story, we've completely reworked this for v3.

I see 2 options:

1. We (re)introduce some light logging config for litestar (something like a non-blocking queue listener as a fallback for litestar's logging handlers)
2. We handle this the same way exceptions are handled in the main request loop: Silently drop tracebacks, unless a logger has been explicitly provided. The one issue here is that channels aren't part of the request/response lifecycle, so we can't let exceptions bubble up to the ASGI server to handle

There's actually 2 other blind spots here: 

1. We don't call exception hooks (and currently can't, because they expect `Exception, Scope`)
2. We don't call `Litestar.debugger_module.post_mortem()`

I need to think about this some more

PR #5019, opened 2026-08-26
fix: do not suggest FromQuery for path parameter defaults

### LINKED ISSUE #5018: Bug: Error message suggests FromQuery for path parameter defaults
### Description

When using old parameter syntax like `slug: str = Parameter(...)` on a path parameter during v2-to-v3 migration, Litestar raises an error to guide developers toward the new `Annotated` syntax. However, the error message incorrectly recommends `FromQuery` instead of `FromPath`.

### Cause
* `ParameterKwarg.param_type` defaults to `ParamType.QUERY` and only updates if `header=`, `cookie=`, or `query=` is explicitly passed.
* `raise_for_kwarg_as_default` is reached from `FieldDefinition.from_annotation`, which inspects a single parameter in isolation and holds no reference to the route. The same function is also called during DTO field generation (`litestar/dto/dataclass_dto.py`), where no route exists at all. Because path-ness comes from the route path rather than the parameter default, Litestar cannot know it is a path parameter at this stage.
* As a result, the `ParamType.PATH: "FromPath"` mapping inside `raise_for_kwarg_as_default` is unreachable for default `Parameter()` usages, falling back to `FromQuery`.

### Impact
This is purely a Developer Experience / messaging issue. Following the bad suggestion and using `FromQuery` still works because Litestar re-derives path parameters from the route path later at startup, generating the correct `in: path` OpenAPI schema without breaking runtime execution. However, the error message itself remains semantically misleading during migration.

I have a fix ready with test coverage and can submit a PR.

_No response_

### MCVE

```python

@get("/{slug:str}")
async def handler(slug: str = Parameter(description="a slug")) -> str:
    return slug

app = Litestar([handler])
# ImproperlyConfiguredException: 500: Usage of parameter defaults to declare
# metadata is no longer supported. Use 'FromQuery' instead
```

### Steps to reproduce

1. Declare a path parameter using the 2.x style default: `slug: str = Parameter(...)
2. Start the app
3. The error names `FromQuery`, but the parameter is a path parameter

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

3.0.0b0 (main, cbc0c51a)

### Platform

- [ ] Linux
- [ ] Mac
- [x] Windows

Sharawey74:
Closes #5018

## Description

`raise_for_kwarg_as_default` maps `ParameterKwarg.param_type` to the replacement syntax it recommends. For a plain `Parameter()` that mapping is unreliable: `param_type` defaults to `ParamType.QUERY` and only changes when `header=`, `cookie=` or `query=` is passed, so a path parameter is told to use `FromQuery`.

This adds a check for whether `param_type` was actually chosen, and only uses the alias mapping when it was. Otherwise the message falls back to `Annotated[<type>, Parameter(...)]`, which is valid for every parameter kind.

`param_type` counts as chosen when either the default is one of the dedicated subclasses (`QueryParameter`, `PathParameter`, `HeaderParameter`, `CookieParameter`), which set it themselves, or one of `header=` / `cookie=` / `query=` was passed.

The route path is not usable at this point. `FieldDefinition.from_annotation` inspects one parameter in isolation, and `raise_for_kwarg_as_default` is also called from DTO field generation where no route exists at all.

Side effect worth noting: the `ParamType.PATH: "FromPath"` entry was previously unreachable, since nothing could set `param_type` to `PATH` on a value reaching this function. It becomes reachable through `PathParameter()`.

## Tests

Three tests added to `tests/unit/test_typing.py`:

* `test_kwarg_definition_as_default_without_explicit_source_suggests_parameter` — covers the reported bug, and fails without the fix.
* `test_kwarg_definition_as_default_suggests_dedicated_alias` — the four dedicated subclasses still name their alias, including `FromPath`.
* `test_kwarg_definition_as_default_with_explicit_source_suggests_dedicated_alias` — the deprecated `header=` / `cookie=` / `query=` kwargs still name their alias.

608 tests pass across `test_typing`, `test_kwargs`, `test_signature` and `test_openapi`. ruff, mypy, pyright and slotscheck are clean.

### DIFF
--- a/litestar/utils/errors.py
+++ b/litestar/utils/errors.py
@@ -5,16 +5,34 @@ from litestar.exceptions import ImproperlyConfiguredException
 from litestar.params import BodyKwarg, KwargDefinition, ParameterKwarg
 
 
+def _has_explicit_param_type(default: ParameterKwarg) -> bool:
+    """Whether ``default.param_type`` reflects an explicit choice by the user.
+
+    ``ParameterKwarg.param_type`` defaults to :attr:`ParamType.QUERY <litestar.enums.ParamType>` and
+    is only reassigned when one of the ``header``/``cookie``/``query`` arguments is given. A plain
+    ``Parameter()`` therefore reports ``QUERY`` even when it annotates a path parameter, since
+    whether a parameter is a path parameter is decided by the route path rather than the annotation.
+    The dedicated subclasses (:class:`~litestar.params.PathParameter` and friends) each set
+    ``param_type`` themselves, so for those the value is meaningful.
+    """
+    return type(default) is not ParameterKwarg or any(
+        value is not None for value in (default.header, default.cookie, default.query)
+    )
+
+
 def raise_for_kwarg_as_default(default: KwargDefinition) -> NoReturn:
     if isinstance(default, BodyKwarg):
         alternative = "Annotated[<type>, Body(...)]"
     elif isinstance(default, ParameterKwarg) and not default.is_constrained:
-        alternative = {
-            ParamType.QUERY: "FromQuery",
-            ParamType.HEADER: "FromHeader",
-            ParamType.COOKIE: "FromCookie",
-            ParamType.PATH: "FromPath",
-        }[default.param_type]
+        if _has_explicit_param_type(default):
+            alternative = {
+                ParamType.QUERY: "FromQuery",
+                ParamType.HEADER: "FromHeader",
+                ParamType.COOKIE: "FromCookie",
+                ParamType.PATH: "FromPath",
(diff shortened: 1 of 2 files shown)


Sharawey74:
The two failing checks here are unrelated to this change:

* `Test server integration` - `click` 8.5.0 was released a few hours before this run and breaks collection via `rich-click`. It affects all PRs, not just this one. Reported in #5020.
* `codecov/patch` - not a required check. The added lines are covered by the three new tests in `tests/unit/test_typing.py`.

All six required checks pass, and the test matrix is green on 3.11 through 3.14.

Issue #5020, opened 2026-08-26
Bug: click 8.5.0 breaks test collection via rich-click get_binary_stream deprecation


Sharawey74:
### Description

`click` 8.5.0 was released on 2026-08-26 at 13:33 UTC. Since then, the `Test server integration` job fails during test collection on every PR.

`rich-click` 1.9.8 (its latest release, from 2026-05-28) imports an attribute that `click` 8.5.0 deprecates:

    rich_click/__init__.py:65   from click.utils import get_binary_stream as get_binary_stream
    click/utils.py:681          warnings.warn("'click.utils.get_binary_stream' is deprecated ...")

Litestar imports `rich_click` at `litestar/cli/__init__.py:8`, and `pyproject.toml` sets `filterwarnings = ["error"]`, so the DeprecationWarning becomes fatal at import time. Three test modules then fail to import and pytest aborts before running anything.

Only the `Test server integration` job is affected, because it escapes the lockfile:

    run: |
      uv sync
      uv pip install -U "${{ matrix.uvicorn-version }}"

`uv sync` installs the locked versions, then `uv pip install -U` upgrades uvicorn and its dependencies (including `click`) to the newest available, picking up 8.5.0.

**Evidence this is time-based rather than PR-specific:**

* Passing, 2026-08-26 08:10 UTC (before the click release): https://github.com/litestar-org/litestar/actions/runs/32946384544
* Failing, 2026-08-26 16:16 UTC (after the click release): https://github.com/litestar-org/litestar/actions/runs/32987489461

Same job and same code path, opposite results either side of the click 8.5.0 upload time.

### URL to code causing the issue

https://github.com/litestar-org/litestar/actions/runs/32987489461

### MCVE

```python
uv sync
uv pip install -U "uvicorn>=0.27.0"
uv run --no-sync pytest tests -m server_integration
```

### Steps to reproduce

1. Open any pull request, or run the MCVE commands locally with click 8.5.0 available
2. The `Test server integration` job fails during collection
3. No tests run - the failure happens while importing test modules, so all 12 selected tests are never reached

### Logs

```text
collected 4624 items / 3 errors / 4612 deselected / 12 selected
==================================== ERRORS ====================================
_____________________ ERROR collecting tests/unit/test_cli _____________________
tests/unit/test_cli/conftest.py:16: in <module>
    from litestar.cli._utils import _path_to_dotted_path
litestar/cli/__init__.py:8: in <module>
    import rich_click
[...]

PR #5022, opened 2026-08-27
fix: correct session cookie chunking and reassembly

### LINKED ISSUE #5021: Bug: chunked client-side session cookies exceed RFC 6265's 4096-byte limit
### Description

`CookieBackendConfig` splits a large session across `session-0`, `session-1`, ...
using `CHUNK_SIZE = 4096 - 64` ([link](https://github.com/litestar-org/litestar/blob/cbc0c51abd59b11e7b429be0aba6639e38488e2c/litestar/middleware/session/client_side.py#L41)), reserving 64 bytes for the cookie name and attributes. The real overhead is larger than 64 in ordinary configurations, so every full chunk it emits is over the 4096-byte limit RFC 6265 sets for a cookie's name, value and attributes combined.

With `secure=True` - which any production deployment sets - this happens with the **default** cookie key. A longer configured key adds to the overrun.

This is easy to miss because nothing local enforces the limit: `httpx`, the Litestar test client, and browsers talking directly to the app all accept the oversized cookie. It surfaces only behind a proxy that enforces the RFC, and there the failure is silent and total (details below). An example of this is the Google Front-End (GFE) servers in front of Google Cloud Run. If a cookie exceeds 4096 bytes the entire cookie set is silently dropped.  The response keeps its normal status code, and nothing is logged.  Yes you can argue that's a Google problem but good luck getting it fixed :)

Measured against a live Cloud Run service by varying the session size:

```
cookie 4084 bytes -> forwarded
cookie 4097 bytes -> forwarded
cookie ~4104 bytes -> ALL Set-Cookie headers on the response dropped
```

Chunking therefore cannot work at all behind such a proxy: the moment a session
needs more than one cookie, every cookie on that response is discarded.

** Suggested fix **

Compute the available value length per cookie from the actual serialized overhead rather than a fixed constant — the chunk name, the quoting, and the attribute string the config will produce — instead of `4096 - 64`. Since `_create_session_cookies` already builds `Cookie` objects with known parameters, the overhead can be measured once from a zero-length-value cookie and subtracted.

Failing loudly would also be an improvement over the current silence: if a cookie would exceed 4096 bytes, a warning naming the cookie would have turned a multi-hour diagnosis into a log line.

** Workaround for others hitting this **

Keep the session small enough to fit in a single unchunked cookie. Splitting is not a viable fallback behind a conforming proxy.

_No response_

### MCVE

```python
"""Chunked session cookies exceed RFC 6265's 4096 bytes."""

from litestar import Request, get
from litestar.middleware.session.client_side import CookieBackendConfig

LIMIT = 4096


@get("/set")
async def set_session(request: Request) -> str:
    # Large enough that the backend splits the session across several cookies.
    request.session["payload"] = "x" * 6000
    return "ok"


def emitted_cookies(session_key: str, **kwargs) -> list[tuple[str, int]]:
    config = CookieBackendConfig(key=session_key, secret=b"0" * 32, max_age=3600, **kwargs)
    with create_test_client([set_session], middleware=[config.middleware]) as client:
        response = client.get("/set")
        return [
            (raw.split("=", 1)[0], len(raw))
            for raw in response.headers.get_list("set-cookie")
        ]


for key in ("session", "session_cdp"):
    for kwargs in ({}, {"secure": True}):
        print(f"key={key!r} {kwargs}")
        for name, size in emitted_cookies(key, **kwargs):
            verdict = "OK" if size <= LIMIT else f"OVER LIMIT by {size - LIMIT}"
            print(f"    {name:20} {size:5} bytes   {verdict}")
```

### Actual output

```
key='session' {}
    session-0             4090 bytes   OK
    session-1             4090 bytes   OK
    session-2              126 bytes   OK
key='session' {'secure': True}
    session-0             4098 bytes   OVER LIMIT by 2
    session-1             4098 bytes   OVER LIMIT by 2
    session-2              134 bytes   OK
key='session_cdp' {}
    session_cdp-0         4094 bytes   OK
    session_cdp-1         4094 bytes   OK
    session_cdp-2          130 bytes   OK
key='session_cdp' {'secure': True}
    session_cdp-0         4102 bytes   OVER LIMIT by 6
    session_cdp-1         4102 bytes   OVER LIMIT by 6
    session_cdp-2          138 bytes   OK
```

### Expected

Every emitted cookie is at most 4096 bytes, for any configured key and any
combination of cookie attributes.

## Where the 64 bytes go

For `session-0` with `secure=True`:

| Part | Bytes |
| --- | --- |
| Name `session-0` | 9 |
| `=` | 1 |
| Quoting of the base64 value (`"..."`) | 2 |
| Value (`CHUNK_SIZE`) | 4032 |
| `; HttpOnly; Max-Age=3600; Path=/; SameSite=lax; Secure` | 54 |
| **Total** | **4098** |

Non-value overhead is 66 bytes against the 64 reserved. Two contributors look
easy to overlook:

- The base64 value contains `=` padding, so `Cookie.to_header` wraps it in
  double quotes, adding 2 bytes that `CHUNK_SIZE` does not account for.
- `Secure` adds 8 bytes and is normally absent in development and present in
  production, so the same code can stay under the limit locally and exceed it
  once deployed.

Overhead also grows with the configured `key`, and with `domain` or a longer
`path` if set — none of which the fixed 64-byte allowance can accommodate.

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

Litestar 2.24.0 (Python 3,12,13)

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

anuanas2007:
## Description

Fixes two bugs in the client-side session backend. They interact, so both are fixed here.

Fixes #5021

**1. Chunks are reassembled in the wrong order (not previously reported)**

I ran into this while verifying #5021. Happy to file it as a separate issue if you would rather track it on its own.

A large session is split across `session-0`, `session-1`, ... and reassembled with a plain `sorted()`, which orders the names alphabetically. Alphabetically `session-10` comes before `session-2`, so once a session needs ten or more cookies the payload is concatenated out of order and the session is lost. Below ten chunks the two orders agree, which is why this has gone unnoticed.

Usually the session decodes to empty and the user is silently logged out. Sometimes the misordered payload yields malformed JSON in the associated-data segment, raising `SerializationException`, which `load_from_connection` does not suppress, so the request returns a 500.

**2. Chunks exceed the 4096-byte cookie limit (#5021)**

`CHUNK_SIZE = 4096 - 64` reserves a fixed 64 bytes for the cookie name and attributes. The real overhead is 66 with the default key and `secure=True`, and grows with the key, `domain` and `path`, so every full chunk goes over the limit. Some proxies then silently drop every `Set-Cookie` header on the response.

Fixing #5021 produces smaller, and therefore more, chunks, which makes the ordering bug easier to hit. Hence one PR.

## The fix

- `get_cookie_keys` sorts by the numeric chunk index instead of the cookie name. The index is read from the suffix after the configured key rather than by splitting on `-`, so keys that contain a hyphen (`my-session`) stay correct.
- The per-cookie budget is measured rather than assumed, as suggested in #5021. A zero-length value is itself emitted quoted (`session=""`), so the quoting that base64 values incur is already accounted for.
- If `key`, `path` and `domain` fill the 4096 bytes on their own, that measurement goes negative and no cookie was emitted at all, a silent logout with a 200 response. It now raises `ImproperlyConfiguredException`.

## Verification

The MCVE from #5021, run unchanged. The "before" column matches the output posted in the issue byte-for-byte:

```
BEFORE                                AFTER
session-0      4098  OVER LIMIT by 2  session-0      4096  OK
session_cdp-0  4102  OVER LIMIT by 6  session_cdp-0  4096  OK
```

Sessions of 40,000 B (14 chunks), 200,000 B (67) and 900,000 B (298) were all lost before and all round-trip now.

Across 4 key lengths x 3 attribute combinations x 5 session sizes:

```
before: 1908 cookies | over limit: 1540 | worst overrun: 289 bytes
after:  1953 cookies | over limit:    0 | worst overrun:   0 bytes
```

## Tests

Four tests added to `tests/unit/test_middleware/test_session/test_client_side_backend.py`:

- `test_get_cookie_keys_are_ordered_by_chunk_index` asserts the ordering directly.
- `test_large_session_round_trips` checks that a session needing more than ten cookies survives a round trip, over both `session` and `my-session`.
- `test_session_cookies_do_not_exceed_max_cookie_size` checks that every emitted `Set-Cookie` is within 4096 bytes, over key lengths 7, 60 and 250 and three attribute combinations.
- `test_dump_data_raises_when_cookie_attributes_leave_no_room` covers the new guard.

`test_dump_and_load_data` previously asserted `len(value) <= CHUNK_SIZE`. It now checks the emitted header against `MAX_COOKIE_SIZE`, since the fixed constant no longer determines the limit.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5022">https://litestar-org.github.io/litestar-docs-preview/5022</a>

### DIFF
--- a/litestar/middleware/session/client_side.py
+++ b/litestar/middleware/session/client_side.py
@@ -38,7 +38,11 @@ if TYPE_CHECKING:
     from litestar.types import Message, Scope, ScopeSession
 
 NONCE_SIZE = 12
+# Retained for backwards compatibility; no longer used. The space available for a cookie's value is now
+# measured from the cookie itself, see ``ClientSideSessionBackend._chunk_size``.
 CHUNK_SIZE = 4096 - 64
+MAX_COOKIE_SIZE: Final = 4096
+"""The maximum size of a single cookie, as mandated by :rfc:`6265`."""
 AAD = b"additional_authenticated_data="
 SET_COOKIE_INCLUDE = {f.name for f in fields(Cookie) if f.name not in {"key", "secret"}}
 CLEAR_COOKIE_INCLUDE = {f.name for f in fields(Cookie) if f.name not in {"key", "secret", "max_age"}}
@@ -70,6 +74,22 @@ class ClientSideSessionBackend(BaseSessionBackend["CookieBackendConfig"]):
             extract_dataclass_items(config, exclude_none=True, include=CLEAR_COOKIE_INCLUDE)
         )
 
+    def _chunk_size(self, name: str) -> int:
+        """Return the number of bytes available for the value of a cookie named ``name``.
+
+        The budget is measured from a cookie with a zero-length value, which is itself emitted quoted
+        (``session=""``). The two quote bytes are therefore already accounted for, which matters because
+        base64 values contain ``+``, ``/`` and ``=`` and are consequently quoted as well.
+
+        Args:
+            name: Key the cookie will be emitted under.
+
+        Returns:
+            The maximum length of that cookie's value.
+        """
+        overhead = len(Cookie(value="", key=name, **self._set_cookie_params).to_header(header=""))
+        return MAX_COOKIE_SIZE - overhead
+
     def dump_data(self, data: Any, scope: Scope | None = None) -> list[bytes]:
         """Given serializable data, including pydantic models and numpy types, dump it into a bytes string, encrypt,
         encode and split it into chunks of the desirable size.
@@ -83,14 +103,27 @@ class ClientSideSessionBackend(BaseSessionBackend["CookieBackendConfig"]):
               string that is encrypted using AES-CGM.
 
         Returns:
-            List of encoded bytes string of a maximum length equal to the ``CHUNK_SIZE`` constant.
+            List of encoded bytes strings, each short enough that the cookie carrying it stays within the
+            :rfc:`6265` size limit of ``MAX_COOKIE_SIZE`` bytes.
         """
         serialized = self.serialize_data(data, scope)
         associated_data = encode_json({"expires_at": round(time.time()) + self.config.max_age})
         nonce = urandom(NONCE_SIZE)
         encrypted = self.aesgcm.encrypt(nonce, serialized, associated_data=associated_data)
         encoded = b64encode(nonce + encrypted + AAD + associated_data)
-        return [encoded[i : i + CHUNK_SIZE] for i in range(0, len(encoded), CHUNK_SIZE)]
+        if len(encoded) <= self._chunk_size(self.config.key):
+            return [encoded]
+
+        # Chunks are stored under ``<key>-<index>``, so the space left for the value depends on how many
+        # digits the largest index takes. Widen the allowance until the resulting chunk count no longer
+        # needs more digits than the size it was computed for.
+        digits = 1
+        while True:
+            size = self._chunk_size(f"{self.config.key}-{'9' * digits}")
+            count = -(-len(encoded) // size)
+            if len(str(count - 1)) <= digits:
+                return [encoded[i : i + size] for i in range(0, len(encoded), size)]
+            digits += 1
 
     def load_data(self, data: list[bytes]) -> dict[str, Any]:
         """Given a list of strings, decodes them into the session object.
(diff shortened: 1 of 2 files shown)


anuanas2007:
The failing Test server integration check is unrelated to this change.
click 8.5.0 breaks test collection via rich-click, tracked in #5020.
All required checks pass.

PR #4815, opened 2026-05-28
fix(openapi): propagate msgspec.Meta on Optional[Annotated[...]] struct fields

pavdwest:
# Summary

`msgspec.Meta` placed inside an `Optional` on a `Struct` field - `Optional[Annotated[T, Meta(...)]]` - was silently dropped from the generated OpenAPI schema. `description`, `examples`, `title`, and every numeric/string constraint (`gt`, `ge`, `lt`, `le`, `min_length`, `max_length`, `pattern`, …) went missing. The non-Optional form worked, which made this look like a serialization quirk, but the cause was a missing case in `kwarg_definition_from_field`: it checks `isinstance(field.type, Metadata)`, while msgspec exposes the optional shape as `UnionType(types=(Metadata(...), NoneType()))` - one level deeper than the check.

# Root cause

`msgspec.inspect` presents `Optional[Annotated[T, Meta(...)]]` as `UnionType((<inner>, NoneType))` where `<inner>` is one of two shapes depending on what `Meta` carries:

- a `Metadata` wrapper, when `Meta` carries description/examples/title, or
- the bare constraint-bearing type (e.g. `IntType(gt=1, ...)`) - msgspec hoists constraint metadata directly onto the type.

Both shapes are buried inside the `UnionType`, so the existing `isinstance(field.type, Metadata)` check misses them and the Optional path falls through without producing a `ParameterKwarg`.

# Fix

A small helper at the top of `kwarg_definition_from_field` peeks inside a single-non-None `UnionType` and returns the inner arm, letting the existing isinstance ladder handle both shapes. Heterogeneous unions (multiple non-None arms) are left alone - no single `Meta` semantically applies to the whole union.

Note:
- Applies to both struct OpenAPI generation (`StructSchemaPlugin`) and msgspec DTOs (`MsgspecDTO._generate_field_definitions`) - same helper, same effect.
- Not a request hot path: runs once per struct field at OpenAPI build / DTO class creation.
- Heterogeneous unions (`Union[Annotated[str, Meta], int]`) intentionally left as-is and pinned by a regression test.

# Minimal reproduction

```python
from typing import Annotated, Optional
import msgspec
from litestar import Litestar, post

class Body(msgspec.Struct):
    name:  Annotated[str, msgspec.Meta(description="works", min_length=2)]
    alias: Optional[Annotated[str, msgspec.Meta(description="dropped", min_length=2)]] = None

@post("/")
def handler(data: Body) -> Body:
    return data

[...]

PR #5028, opened 2026-08-31
fix(dto): push Meta constraints into Optional's non-None arm

### LINKED ISSUE #5025: Bug: Regression - Setting `min_length` on Optional `str` types causes 500 error
### Description

I just upgraded from 2.22.0 to 2.24.0 and started seeing this error: `TypeError: Can only set `min_length` on a str, bytes, or collection type - type `typing.Annotated[typing.Optional[str], msgspec.Meta(min_length=1, examples=[])]` is invalid`

This was surprising, because applying meta annotations like this to optional parameters is a core thing we do in many places throughout our projects.

At first I thought this might be a msgspec problem, but it runs just fine outside of litestar, and as far as I can tell the issue only appears when a request is made to a handler that wraps a `Struct` in litestar's `MsgspecDTO`.

Instead of checking the exact types for this sort of thing maybe it should be checking whether the sub type is inherited from a valid base type. Or maybe explicitly allow for optional versions of these types.

This worked previously, so I consider it a regression. If this was a purposeful change then I feel like it should wait until a major version bump like 3.0 as it breaks existing code. And if that's the case: what is the canonical way to set a minimum length on an optional string now?

_No response_

### MCVE

```python
from typing import Annotated

class TestStruct(Struct):
    text: Annotated[str, Meta(min_length=1)] | None


@post('/test', dto=MsgspecDTO[TestStruct])
async def test(data: TestStruct) -> None:
    return


# then just send a valid POST request, like
# curl --header "Content-Type: application/json" --request POST --data '{"text": "any"}' http://localhost:8000/test
```

### Steps to reproduce

Set things like `min_length` or `max_length` on optional `str`, `bytes`, etc. in a `Struct` in a `MsgspecDTO` and then make a valid request to a handler that uses that.

### Screenshots

_No response_

### Logs

```text
  File "litestar/middleware/_internal/exceptions/middleware.py", line 158, in __call__
  File "litestar/routes/http.py", line 81, in handle
  File "litestar/routes/http.py", line 133, in _get_response_for_request
  File "litestar/routes/http.py", line 163, in _call_handler_function
    kwargs = await parameter_model.to_kwargs(connection=request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "litestar/_kwargs/kwargs_model.py", line 456, in to_kwargs
    await extractor(output, connection)
  File "litestar/_kwargs/extractors.py", line 491, in extractor
    values["data"] = await data_extractor(connection)
  File "litestar/_kwargs/extractors.py", line 368, in _extract_multipart
    return data_dto(connection).decode_builtins(form_values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "litestar/dto/base_dto.py", line 115, in decode_builtins
    return backend.populate_data_from_builtins(value, self.asgi_connection)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "litestar/dto/_codegen_backend.py", line 109, in populate_data_from_builtins
    return self.transfer_data_from_builtins(self.parse_builtins(builtins, asgi_connection))
  File "litestar/dto/_backend.py", line 267, in parse_builtins
    return convert(
           ^^^^^^^^
TypeError: Can only set `min_length` on a str, bytes, or collection type - type `typing.Annotated[typing.Optional[str], msgspec.Meta(min_length=1, examples=[])]` is invalid
```

### Litestar Version

I confirmed this first appeared in 2.23.0 and still exists in 2.24.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

mokashang:
### Summary

Building the msgspec transfer model for a struct field like
`Optional[Annotated[str, msgspec.Meta(min_length=1)]]` produced
`Annotated[Optional[str], Meta(min_length=1)]` — with the constraint on the outer
`Union` rather than the `str` arm. `msgspec` rejects that shape
(`min_length`/`max_length`/`pattern`/`gt`/`ge`/`lt`/`le`/`multiple_of` may only
sit on the base type they constrain), so decoding raised

```
TypeError: Can only set `min_length` on a str, bytes, or collection type - type
`typing.Annotated[typing.Optional[str], msgspec.Meta(min_length=1, examples=[])]` is invalid
```

at request time in `_backend.py::parse_raw` / `parse_builtins`, and the handler
returned HTTP 500.

### Root cause

`kwarg_definition_from_field` (extended in #4815 to recognise
`Optional[Annotated[T, Meta(...)]]`) correctly lifts the inner `Meta` into a
`ParameterKwarg`. But when `_create_struct_for_field_definitions` rebuilds the
transfer struct, it does:

```python
        field_type = Annotated[field_type, field_meta]
```

`field_type` here is `Union[str, None]` (from
`_create_transfer_model_type_annotation`) — so the `Meta` lands on the outer
`Union`, not on the `str` arm the user originally annotated.

### Fix

When the transfer `field_type` is `Union[X, None]` with a single non-`None`
arm, wrap `X` with the `Meta` instead and rebuild the union, mirroring
`Optional[Annotated[X, ...]]` which `msgspec` accepts. Other shapes are
wrapped directly, as before. The same helper is used for the
non-`passthrough_constraints` branch so a `ParameterKwarg` foreign annotation
lands in the same place — no behaviour change for msgspec there, but keeps the
two branches consistent.

### Testing

- Added `test_msgspec_dto_optional_annotated_meta_constraint` parametrised
  over the seven `msgspec.Meta` constraints that trip the outer-union check
  (`min_length`, `max_length`, `pattern`, `gt`, `ge`, `lt`, `le`), each
  asserting that
  - `None` decodes cleanly (Optional stays honoured), and
  - a value violating the constraint raises `msgspec.ValidationError`
    (pinning that the constraint is enforced on the inner arm, not silently
    dropped).
- Full `tests/unit/test_dto` + `tests/unit/test_openapi` — 552 passed, 1 skipped, 1 xfailed.
- The pre-existing pyright ignore on the `struct_fields.append(...)` tuple
  became unnecessary once the wrapping goes through a helper returning `Any`,
  so it is removed. `uv run pyright litestar/dto/_backend.py` and
  `uv run mypy litestar/dto/_backend.py` are both clean.

### Reproducer

Before this change:

```python
from typing import Annotated, Optional


class Body(Struct):
    text: Optional[Annotated[str, Meta(min_length=1)]] = None


@post("/", dto=MsgspecDTO[Body])
async def handler(data: Body) -> None: ...


with TestClient(Litestar(route_handlers=[handler])) as c:
    print(c.post("/", json={"text": "any"}).status_code)  # 500 before, 201 after
    print(c.post("/", json={"text": ""}).status_code)     # 400 (constraint enforced)
    print(c.post("/", json={"text": None}).status_code)   # 201 (Optional honoured)
```

Fixes #5025

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -6,6 +6,20 @@
 .. changelog:: 3.0.0
     :date: 2364-01-27
 
+    .. change:: Fix ``TypeError`` from ``MsgspecDTO`` on ``Optional[Annotated[T, Meta(...)]]`` struct fields
+        :type: bugfix
+        :issue: 5025
+
+        A request to a handler using ``MsgspecDTO`` for a struct with a field like
+        ``Optional[Annotated[str, msgspec.Meta(min_length=1)]]`` raised
+        ``TypeError: Can only set `min_length` on a str, bytes, or collection type``
+        during request decoding, returning HTTP 500. The DTO transfer model was
+        built as ``Annotated[Optional[str], Meta(min_length=1)]`` — with the
+        constraint on the outer ``Union`` rather than the ``str`` arm — which
+        ``msgspec`` rejects. Meta constraints are now pushed into the non-``None``
+        arm of an ``Optional`` field when the transfer model is generated,
+        mirroring ``Optional[Annotated[T, Meta(...)]]`` which ``msgspec`` accepts.
+
     .. change:: Fix ``TypeError`` when generating a schema for a union of enums
         :type: bugfix
         :pr: 4997
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -15,6 +15,8 @@ from typing import (
     Protocol,
     Union,
     cast,
+    get_args,
+    get_origin,
 )
 
 import msgspec
@@ -818,6 +820,23 @@ def _create_struct_field_meta_for_field_definition(field_definition: TransferDTO
     )
 
 
+def _wrap_field_type_with_metadata(field_type: Any, metadata: Any) -> Any:
+    """Attach ``metadata`` to ``field_type``, pushing it into the non-``None`` arm of an ``Optional``.
+
+    ``msgspec`` rejects ``Annotated[Optional[str], Meta(min_length=1)]`` because constraints
+    like ``min_length`` may only be applied to the base type they constrain, not to the
+    outer union. When ``field_type`` is ``Union[X, None]`` with a single non-``None`` arm,
+    wrap ``X`` instead and rebuild the union, mirroring ``Optional[Annotated[X, ...]]`` which
+    ``msgspec`` accepts. Other shapes are wrapped directly.
+    """
+    if get_origin(field_type) is Union:
+        args = get_args(field_type)
+        non_none = tuple(a for a in args if a is not type(None))
+        if len(non_none) == 1 and len(non_none) < len(args):
+            return Union[Annotated[non_none[0], metadata], None]
+    return Annotated[field_type, metadata]
+
+
 def _create_struct_for_field_definitions(
     *,
     model_name: str,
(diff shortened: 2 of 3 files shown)


provinzkraut:
Closing due to our AI policy

PR #5029, opened 2026-08-31
feat(redis): support hash namespace strategy

### LINKED ISSUE #4992: Enhancement: Support redis hashes with field expiration for namespaced `RedisStore`s on redis 7.4+
### Summary

Redis `7.4` introduced [field expiration](https://redis.io/docs/latest/develop/data-types/hashes/#field-expiration) for hashes. This allows us to more easily implement namespaced stores for redis, also giving a big performance improvement for bulk deletions (because the hash key can simply be deleted instead of iterating over every individual key in the namespace).

This feature should be opt-in, possibly with a `None` default, detecting the redis version on startup and using the optimal strategy. 

On redis 8+, [`HGETEX`](https://redis.io/docs/latest/commands/hgetex/) can be used for "get and renew" actions

ningmao-hlyz:
## Description

Closes #4992

This adds an opt-in hash-based namespace strategy for `RedisStore`, using Redis hash field expiration introduced in Redis 7.4.

### Changes

- Add `RedisStoreNamespaceStrategy` with `keys`, `hash`, and `auto` options.
- Keep `keys` as the default so existing users retain the current `namespace:key` layout.
- Store hash-strategy fields in one Redis hash named directly after the namespace; no additional hash-key prefix is introduced.
- Use atomic Lua scripts with `HPEXPIRE`/`HPTTL` on Redis 7.4–7.x.
- Use native `HSETEX` and conditional `HGETEX` on Redis 8+.
- Preserve `keep_ttl`, renew-only-existing-expiry, delete, and parent-child namespace behavior.
- Preserve millisecond precision for `timedelta` expiry values and reject non-positive hash expirations before writing.
- Require a namespace for the explicit `hash` strategy; `auto` falls back to `keys` when Redis version detection is unavailable or unsupported.
- Document that the two layouts are not migrated automatically.

### Compatibility

- Existing users keep the `keys` strategy by default.
- `hash` requires Redis 7.4 or newer and `redis-py` 6.1.0 or newer.
- `auto` selects `hash` on Redis 7.4+ and `keys` on older or unavailable servers.
- Data written with one strategy is not visible through the other strategy.

### Validation

- Redis 7.4.7: 127 Redis-focused tests passed, 4 skipped.
- Redis 8.10.1: 127 Redis-focused tests passed, 4 skipped.
- Ruff, mypy, pyright, slotscheck, `uv lock --check`, and `git diff --check` passed.

### delete_all() Benchmark
Test environment: Windows x64, local loopback TCP against Redis 7.4.7 / 8.10.1; persistence disabled. Each scenario was run seven times. The table reports minimum, median, and maximum durations for <code>RedisStore.delete_all()</code>. Data pre‑population time is excluded from measurements.</p>

Redis | Values | KEYS layout | HASH layout | Median improvement
-- | -- | -- | -- | --
7.4.7 | 1 000 | 3.984 ms | 0.438 ms | 9.1x
7.4.7 | 10 000 | 31.487 ms | 0.434 ms | 72.6x
8.10.1 | 1 000 | 3.856 ms | 0.397 ms | 9.7x
8.10.1 | 10 000 | 29.149 ms | 0.390 ms | 74.7x

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5029">https://litestar-org.github.io/litestar-docs-preview/5029</a>


docs/usage/stores.rst
```diff
@@ -161,13 +161,31 @@ When using the :class:`RedisStore <.redis.RedisStore>`, this allows to reuse the
...
+bulk deletion scale with the number of namespaces instead of the number of stored values. Enable this layout explicitly
+with :attr:`RedisStoreNamespaceStrategy.HASH <.redis.RedisStoreNamespaceStrategy.HASH>`:
+
+.. code-block:: python
+
+    from litestar.stores.redis import RedisStore, RedisStoreNamespaceStrategy
+
+    store = RedisStore.with_client(namespace_strategy=RedisStoreNamespaceStrategy.HASH)
+
+Existing prefixed keys are not migrated or visible through the hash layout, so the default remains
```
provinzkraut: this belongs in the release notes, not the documentation

litestar/stores/redis.py
```diff
@@ -13,11 +14,30 @@
...
+    from collections.abc import Awaitable, Callable
     from types import TracebackType
 
     from redis.asyncio.connection import Connection
 
-__all__ = ("RedisStore",)
+__all__ = ("RedisStore", "RedisStoreNamespaceStrategy")
+
+
+class RedisStoreNamespaceStrategy(StrEnum):
```
provinzkraut: We generally use literals instead of enums

litestar/stores/redis.py
```diff
@@ -13,11 +14,30 @@
...
+    """Storage strategies for namespaced values in :class:`RedisStore`."""
+
+    KEYS = "keys"
+    HASH = "hash"
+    AUTO = "auto"
+
+
+_HASH_FIELD_EXPIRATION_MIN_VERSION = (7, 4)
+_HASH_FIELD_EXPIRATION_COMMANDS_VERSION = (8, 0)
+_HASH_KEY_PREFIX = "__litestar_redis_store_hash__:"
```
provinzkraut, line 29: why does this introduce a new prefix?

litestar/stores/redis.py
```diff
@@ -13,11 +14,30 @@
...
+_HASH_NAMESPACE_SEPARATOR = b"_".hex()
```
provinzkraut: what's the purpose of this?

litestar/stores/redis.py
```diff
@@ -123,23 +228,102 @@ def with_client(
...
+        if self._redis_version is not None:
+            store._redis_version = self._redis_version
+            store._resolved_namespace_strategy = self._resolved_namespace_strategy
+        return store
 
     def _make_key(self, key: str) -> str:
         prefix = f"{self.namespace}:" if self.namespace else ""
         return prefix + key
 
+    def _make_hash_key(self) -> str:
```
provinzkraut, line 248: why is this different than `_make_key`?

litestar/stores/redis.py
```diff
@@ -123,23 +228,102 @@ def with_client(
...
+        return int(value.total_seconds()) if isinstance(value, timedelta) else value
+
+    async def _set_hash_value(
+        self,
+        key: str,
+        value: bytes,
+        expires_in: int | timedelta | None,
+        keep_ttl: bool,
+    ) -> None:
+        """Set a value using Redis hash field expiration commands."""
```
provinzkraut, line 296: Why can't this just use `hsetex`?

litestar/stores/redis.py
```diff
@@ -63,9 +103,68 @@ def __init__(
...
+            b"""
+        local hash = KEYS[1]
+        local field = ARGV[1]
+        local renew = tonumber(ARGV[2])
+        local ttl = redis.call('HTTL', hash, 'FIELDS', 1, field)[1]
+
+        if ttl > 0 then
+            return redis.call('HGETEX', hash, 'EX', renew, 'FIELDS', 1, field)[1]
+        end
+        return redis.call('HGET', hash, field)
```
provinzkraut, line 150: What's the purpose of this? Why can't it just call `HGETEX`?

litestar/stores/redis.py
```diff
@@ -201,6 +388,19 @@ async def get(self, key: str, renew_for: int | timedelta | None = None) -> bytes
             The value associated with ``key`` if it exists and is not expired, else
             ``None``
         """
+        if await self._resolve_namespace_strategy() is RedisStoreNamespaceStrategy.HASH:
+            hash_key = self._make_hash_key()
```
provinzkraut, line 393: Why not just always use `hgetex` if available?

docs/usage/stores.rst
ningmao-hlyz: Okay , done

litestar/stores/redis.py
ningmao-hlyz: Okay,done

litestar/stores/redis.py
ningmao-hlyz: The prefix isolates hash keys from the legacy namespace:key layout, preventing cross‑layout impacts from strategy changes or namespace deletion.

litestar/stores/redis.py
ningmao-hlyz: - The separate separator variable is no longer needed and has been removed.
- We now build deletion‑matching patterns as `namespace_*`.
- This targets keys like `FOO_child` while skipping `FOOBAR`.
- It also stops characters inside the namespace from being treated as Redis glob wildcards.
- Testing confirmed that the previous KEYS‑based deletion strategy carried a risk of accidental key removal.

litestar/stores/redis.py
ningmao-hlyz: `_make_hash_key` is kept intentionally separate: `_make_key` generates one Redis key per value, while `_make_hash_key` returns the shared namespace hash key with encoded namespace formatting applied.

litestar/stores/redis.py
ningmao-hlyz: Redis 7.4‑7.x supports hash‑field expiration but lacks `HSETEX`, so a Lua script atomically runs `HSET`+`HEXPIRE`. Redis 8+ uses native `HSETEX`.

litestar/stores/redis.py
ningmao-hlyz: The Lua‑scripted renewal checks the field’s existing TTL first to preserve persistence semantics: persistent fields remain unchanged, while expiring fields get their TTL refreshed. The full check‑and‑renew workflow runs atomically.

litestar/stores/redis.py
ningmao-hlyz: Unconditionally using `HGETEX` would add a TTL to persistent fields during renewal.
The conditional path maintains legacy semantics: `HGETEX` for expiring fields, plain `HGET` for persistent values on Redis 8+

litestar/stores/redis.py
provinzkraut: Yeah but why?

litestar/stores/redis.py
provinzkraut: Yes but why is it manually executing a redis command instead of using the [`hsetex`](https://redis.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.hsetex) method?

litestar/stores/redis.py
ningmao-hlyz: The main reason for separation is safety and isolation.  

`_make_key` serves the legacy one‑key‑per‑value layout. We want to avoid any risk that future adjustments to `_make_key` could break legacy user data if it were also used for the hash namespace. Separate helpers let both storage formats evolve independently.

litestar/stores/redis.py
ningmao-hlyz: The project maintains compatibility with `redis‑py >= 4.4.4`.  

The `hsetex()` helper method was only added in later redis‑py releases. Older supported client versions do **not** expose a `.hsetex()` method on the client object at all, so calling `client.hsetex()` directly would raise an `AttributeError` and crash for users on those SDK versions, even when talking to a Redis 8 server.  

`execute_command()` lets us issue the raw server command conditionally, benefiting from native HSETEX on Redis 8+ without bumping our redis‑py minimum requirement.

Alternatively, could we consider adjusting the redis‑py version constraint in `pyproject.toml`?

ningmao-hlyz:
Hi, @provinzkraut . Could you please take another look at these changes? Thank you.

PR #5033, opened 2026-09-01
fix(dto): 500 error when Optional field has msgspec.Meta constraint

### LINKED ISSUE #5025: Bug: Regression - Setting `min_length` on Optional `str` types causes 500 error

happykust:
## Description

Previously, attempting to use the Optional data type in a `msgspec.Struct` with a msgspec constraint resulted in a `TypeError`. This occurred because the `msgspec.Meta` annotation was applied to the entire union, not the original data type. `msgspec` does not allow such an annotation.

Now, only the non-None data type is annotated in the structure field, and msgspec correctly handles constraints.
And None is still accepted.

This also affected partial DTOs with an `Optional` field, which previously also returned a 500, now - 400.

Closes #5025 
Re-open #5032 - Sorry, I'm confused about whether to create a PR for main or v2, so I will open a PR with a fix for v2 if needed.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5033">https://litestar-org.github.io/litestar-docs-preview/5033</a>

### DIFF
--- a/litestar/dto/_backend.py
+++ b/litestar/dto/_backend.py
@@ -6,16 +6,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Collection, Mapping, Set
 from dataclasses import replace
-from typing import (
-    TYPE_CHECKING,
-    Annotated,
-    Any,
-    ClassVar,
-    Final,
-    Protocol,
-    Union,
-    cast,
-)
+from typing import TYPE_CHECKING, Annotated, Any, ClassVar, Final, Protocol, Union, cast, get_args, get_origin
 
 import msgspec
 from msgspec import UNSET, Struct, UnsetType, convert, defstruct, field
@@ -37,6 +28,7 @@ from litestar.enums import RequestEncodingType
 from litestar.params import KwargDefinition
 from litestar.serialization import decode_json, decode_msgpack
 from litestar.types import Empty
+from litestar.types.builtin_types import NoneType, UnionTypes
 from litestar.typing import FieldDefinition
 from litestar.utils import unique_name_for_scope
 
@@ -837,7 +829,13 @@ def _create_struct_for_field_definitions(
 
         if field_definition.passthrough_constraints:
             if (field_meta := _create_struct_field_meta_for_field_definition(field_definition)) is not None:
-                field_type = Annotated[field_type, field_meta]
+                field_types = get_args(field_type)
+                non_none = [t for t in field_types if t not in (NoneType, msgspec.UnsetType)]
+                if get_origin(field_type) in UnionTypes and len(non_none) == 1:
+                    constrained = non_none[0]
+                    field_type = Union[tuple(Annotated[t, field_meta] if t is constrained else t for t in field_types)]
+                else:
+                    field_type = Annotated[field_type, field_meta]
         elif field_definition.kwarg_definition:
             field_type = Annotated[field_type, field_definition.kwarg_definition]
 
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -8,6 +8,7 @@
 
     .. change:: Fix ``TypeError`` when Optional field has msgspec.Meta constraint
         :type: bugfix
+        :pr: 5032
         :issue: 5025
 
         Previously, attempting to use the Optional data type in a ``msgspec.Struct`` with a
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -6,9 +6,9 @@
 .. changelog:: 3.0.0
     :date: 2364-01-27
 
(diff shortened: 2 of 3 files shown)


happykust:
Like in #5032 Codecov showing [uncovered lines at litestar/dto/_backend.py](https://app.codecov.io/gh/litestar-org/litestar/commit/12e17a6349256072e5a7c891bfb82860bffaca4f). The same results on the main branch and this is about import lines:
```
litestar/dto/_backend.py     269     59    124      3    84%   5-81, 126, 161-162, 179, 196, 230, 254, 272, 296, 314, 345, 382, 398, 426, 445, 468, 503, 530, 558, 565, 570, 579, 637, 690, 757, 765, 783, 800, 821, 861, 887, 904, 930, 956, 960, 964, 969, 974
```

litestar/dto/_backend.py
```diff
@@ -837,7 +829,13 @@ def _create_struct_for_field_definitions(
 
         if field_definition.passthrough_constraints:
             if (field_meta := _create_struct_field_meta_for_field_definition(field_definition)) is not None:
-                field_type = Annotated[field_type, field_meta]
+                field_types = get_args(field_type)
+                non_none = [t for t in field_types if t not in (NoneType, msgspec.UnsetType)]
+                if get_origin(field_type) in UnionTypes and len(non_none) == 1:
+                    constrained = non_none[0]
```
provinzkraut, line 835: This should use the already parsed type annotations

PR #5034, opened 2026-09-02
perf(di): build canonical dependency graph

stupidprogrammer4:
<html><head></head><body><h2>Description</h2><p>The previous dependency-graph construction recursively expanded every dependency path into a separate tree. When multiple dependencies referenced the same sub-dependency, the corresponding <code inline="">DependencyContainer</code> and signature metadata were recreated for every reachable path.</p><p>For dense or highly convergent dependency DAGs, this caused the amount of bootstrap work to grow exponentially with the number of dependencies.</p><p>This PR:</p><ul><li><p>Builds a canonical graph containing exactly one <code inline="">DependencyContainer</code> per reachable dependency key.</p></li><li><p>Uses iterative traversal to avoid consuming the Python call stack for deeply nested dependency chains.</p></li><li><p>Extracts and validates signature metadata once per dependency, then merges it into the handler's <code inline="">KwargsModel</code>.</p></li><li><p>Preserves shared nodes across multiple dependency branches.</p></li><li><p>Detects dependency cycles and raises an <code inline="">ImproperlyConfiguredException</code> containing the cycle path.</p></li><li><p>Adds tests covering shared-node reuse, single-pass metadata creation, and cycle detection.</p></li></ul><p>For example, a dense graph with 16 dependencies previously expanded into 32,768 duplicated nodes. The canonical graph contains only 16 nodes.</p><h3>Benchmarks</h3><p>Benchmark environment:</p><ul><li><p>Python 3.13.7</p></li><li><p><code inline="">Litestar(..., openapi_config=None)</code></p></li><li><p>Dense dependency DAG:</p><ul><li><p><code inline="">d0</code> has no dependencies.</p></li><li><p>Each <code inline="">dN</code> depends on every preceding dependency.</p></li><li><p>The route handler requests only the final dependency.</p></li></ul></li><li><p>Results are median application initialization times.</p></li><li><p>Before: 3 repetitions.</p></li><li><p>After: 5 repetitions.</p></li></ul>
Dependencies | Before | After | Speedup
-- | -- | -- | --
12 | 64.731 ms | 10.068 ms | 6.43×
14 | 329.805 ms | 10.883 ms | 30.30×
16 | 1,255.295 ms | 12.982 ms | 96.69×

<p>These are local synthetic microbenchmarks of application startup and do not measure request throughput.</p><h3>Testing</h3><p>Added tests verifying that:</p><ul><li><p>Shared dependencies reuse the same canonical graph node.</p></li><li><p>Signature metadata is created exactly once per reachable dependency.</p></li><li><p>Dependency cycles produce a configuration error containing the detected cycle path.</p></li></ul><p>The targeted dependency-batch test module passes successfully:</p><pre><code class="language-text">12 passed
</code></pre></body></html>

### DIFF
--- a/litestar/_kwargs/dependencies.py
+++ b/litestar/_kwargs/dependencies.py
@@ -2,10 +2,21 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
-__all__ = ("DependencyContainer", "create_dependency_batches", "map_dependencies_recursively", "resolve_dependency")
+from litestar.exceptions import ImproperlyConfiguredException
+
+__all__ = (
+    "DependencyContainer",
+    "DependencyGraph",
+    "create_dependency_batches",
+    "create_dependency_graph",
+    "map_dependencies_recursively",
+    "resolve_dependency",
+)
 
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from litestar._kwargs.cleanup import DependencyCleanupGroup
     from litestar.connection import ASGIConnection
     from litestar.di import Provide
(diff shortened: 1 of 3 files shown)


provinzkraut:
DI is already in the process of a rewrite for v3

PR #5035, opened 2026-09-02
perf: precompute CORS Access-Control-Allow-Headers/Methods values

Kumzy:
Build both header values once in ``CORSMiddleware.__init__`` instead of sorting and joining the configured lists on every response. Breaking only for code mutating allow_headers/allow_methods after construction, matching the rest of the CORS state that is already frozen there.

### DIFF
--- a/docs/release-notes/changelog.rst
+++ b/docs/release-notes/changelog.rst
@@ -6,23 +6,6 @@
 .. changelog:: 3.0.0
     :date: 2364-01-27
 
-    .. change:: Precompute CORS ``Access-Control-Allow-Headers`` and ``Access-Control-Allow-Methods`` values
-        :type: feature
-        :breaking:
-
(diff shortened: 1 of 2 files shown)


docs/release-notes/changelog.rst
```diff
@@ -6,6 +6,23 @@
 .. changelog:: 3.0.0
     :date: 2364-01-27
 
+    .. change:: Precompute CORS ``Access-Control-Allow-Headers`` and ``Access-Control-Allow-Methods`` values
```
provinzkraut: Don't think this needs a changelog entry

Kumzy: Done, I removed it

Kumzy:
I removed the breaking changes from the PR title if we remove the changelog entry, also because the API is internal. I can add it back if it's wrong @provinzkraut

PR #4702, opened 2026-04-16
feat: generate OpenAPI params from typed reserved kwargs

hamza-mobeen:
## Description

This PR implements core framework support for generating OpenAPI parameter schemas directly from strongly-typed reserved kwargs like `query`, `headers`, and `cookies`, significantly improving the developer experience (DX) and OpenAPI introspectability. 
Previously, when a developer annotated a reserved kwarg such as `query` with a `TypedDict`, the framework's OpenAPI generator skipped it entirely due to a blanket exclusion of `RESERVED_KWARGS` during schema parameter resolution in `parameters.py`.

### Key Technical Enhancements
**1. Semantic Parameter Parsing & Introspection**
Refactored the `litestar._openapi.parameters.ParameterFactory` to support introspection of `TypedDict` definitions applied to `DOCUMENTABLE_RESERVED_KWARGS` (`query`, `headers`, `cookies`). 
By deeply inspecting `__required_keys__` and utilizing Litestar's robust `SchemaCreator`, the OpenAPI spec now automatically generates exhaustive, type-safe schema attributes for each item present in the specified type dictionary.
*Before this PR (OpenAPI Schema omitted completely):*
```python
class SearchQuery(TypedDict, total=False):
    page: int
    search: str
@get("/items")
async def get_items(query: SearchQuery) -> None:
    ...
```
*After this PR (Schema effectively renders distinct parameters):*
```json
"parameters": [
  {
    "name": "page",
    "in": "query",
    "required": false,
    "schema": {"type": "integer"}
  },
  {
    "name": "search",
    "in": "query",
    "required": false,
    "schema": {"type": "string"}
  }
]
```
**2. Constant Refactoring Architecture (`litestar/constants.py`)**
Introduced an isolated `DOCUMENTABLE_RESERVED_KWARGS` constant namespace. This preserves backwards compatibility with `RESERVED_KWARGS` exclusions — effectively preventing internal kwargs (`state`, `request`, `scope`, etc) from incorrectly exposing into public OpenAPI outputs, whilst explicitly authorizing `query`, `headers`, and `cookies` to map their documentation properties.
**3. Extensive Validation & Unit Testing**
Includes a comprehensive test suite (`test_reserved_kwargs_openapi.py`) of 11 new granular unit tests covering:
* Positive assertion checks over `headers`, `query`, and `cookies` generation.
* Extractive verification mapping properties precisely to `ParamType.QUERY`, `ParamType.HEADER`, and `ParamType.COOKIE`.
[...]

PR #5036, opened 2026-09-02
fix(openapi): document model-typed reserved kwargs

### LINKED ISSUE #2015: Enhancement: Support for Typing the `query` keyword (and other reserved kwargs) to generate OpenAPI Spec

icecold009:
## Summary

- Document model-typed `query`, `headers`, and `cookies` kwargs as individual OpenAPI parameters.
- Support optional and `Annotated[...]` model wrappers while preserving nested schema references.
- Add regression coverage for issue #2015.

## Validation

- `pytest tests/unit/test_openapi`: 288 passed, 1 xfailed
- `pytest tests/unit/test_kwargs`: 253 passed, 1 xfailed
- Targeted mypy: passed
- Ruff check and format: passed
- `git diff --check`: passed
- Pre-commit selected hooks passed except `unasyncd`, which crashes in its isolated hook environment with `TypeError: File.__init__() got an unexpected keyword argument 'limiter'` before checking files.

Fixes #2015

### DIFF
--- a/litestar/_openapi/parameters.py
+++ b/litestar/_openapi/parameters.py
@@ -21,6 +21,12 @@ if TYPE_CHECKING:
 
 __all__ = ("create_parameters_for_handler",)
 
+_DOCUMENTABLE_RESERVED_KWARGS: dict[str, ParamType] = {
+    "query": ParamType.QUERY,
+    "headers": ParamType.HEADER,
+    "cookies": ParamType.COOKIE,
+}
+
 
 class ParameterCollection:
     """Facilitates conditional deduplication of parameters.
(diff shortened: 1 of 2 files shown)


provinzkraut:
As explained in https://github.com/litestar-org/litestar/pull/4702#issuecomment-4260762632, this is not a OpenAPI gap. The feature does not exist. Closing this PR.

PR #5037, opened 2026-09-02
feat(jwt): add support for leeway param in jwt security backends

### LINKED ISSUE #4584: Enhancement: support `leeway` in `jwt`

happykust:
## Description

Add `leeway` parameter support to set allowed clock error margin in second for token validation.
I mirrored `jwt.decode` logic so `leeway` parameter also respected by `iat` checks, not only `exp`.

### Why changes are breaking

Because of I added `leeway` parameter in `decode` and `decode_payload` methods' signatures, users who redefined either method in their code should add this parameter in signature, otherwise they will get TypeError. 

Closes #4584 

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5037">https://litestar-org.github.io/litestar-docs-preview/5037</a>

### DIFF
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -199,9 +199,7 @@ class Token:
                 raise ImproperlyConfiguredException(f"{_LEEWAY_EXTRA_KEY} is a reserved key and cannot be set")
 
             extras[_LEEWAY_EXTRA_KEY] = leeway
-            token = msgspec.convert(payload, cls, strict=False)
-            token.extras.pop(_LEEWAY_EXTRA_KEY, None)
-            return token
+            return msgspec.convert(payload, cls, strict=False)
         except (
             KeyError,
             jwt.exceptions.InvalidTokenError,
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -4,7 +4,7 @@ import dataclasses
 from collections.abc import Sequence  # noqa: TC003
 from dataclasses import asdict, dataclass, field
 from datetime import UTC, datetime, timedelta
-from typing import TYPE_CHECKING, Any, TypedDict
+from typing import TYPE_CHECKING, Any, Final, TypedDict
 
 import jwt
 import msgspec
@@ -19,7 +19,7 @@ __all__ = (
     "Token",
 )
 
-_LEEWAY_EXTRA_KEY = "__leeway__"
+_LEEWAY_EXTRA_KEY: Final = "__leeway__"
 
 
 def _normalize_datetime(value: datetime) -> datetime:
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -83,11 +83,12 @@ class Token:
             raise ImproperlyConfiguredException(f"exp value must be a datetime in the future, leeway is {leeway}")
 
         if isinstance(self.iat, datetime) and (
-            (iat := _normalize_datetime(self.iat)).timestamp() <= _normalize_datetime(datetime.now(UTC)).timestamp()
+            ((iat := _normalize_datetime(self.iat)) - timedelta(seconds=leeway)).timestamp()
+            <= _normalize_datetime(datetime.now(UTC)).timestamp()
         ):
             self.iat = iat
         else:
-            raise ImproperlyConfiguredException("iat must be a current or past time")
+            raise ImproperlyConfiguredException(f"iat must be a current or past time, leeway is {leeway}")
 
     @classmethod
     def decode_payload(
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -68,7 +68,6 @@ class Token:
     """Extra fields that were found on the JWT token."""
 
     def __post_init__(self) -> None:
-        # using extras because of msgspec.convert in decode method below doesn't support InitVar
         leeway = self.extras.get(_LEEWAY_EXTRA_KEY, 0)
 
         if len(self.sub) < 1:
--- a/litestar/security/jwt/token.py
+++ b/litestar/security/jwt/token.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import dataclasses
 from collections.abc import Sequence  # noqa: TC003
 from dataclasses import asdict, dataclass, field
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from typing import TYPE_CHECKING, Any, TypedDict
 
 import jwt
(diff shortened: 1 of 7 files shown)


litestar/security/jwt/token.py
```diff
@@ -19,6 +19,8 @@
     "Token",
 )
 
+_LEEWAY_EXTRA_KEY = "__leeway__"
```
sobolevn:
```suggestion
_LEEWAY_EXTRA_KEY: Final = "__leeway__"
```

litestar/security/jwt/token.py
```diff
@@ -183,7 +195,14 @@ def decode(
             extras = payload.setdefault("extras", {})
             for key in extra_fields:
                 extras[key] = payload.pop(key)
-            return msgspec.convert(payload, cls, strict=False)
+
+            if _LEEWAY_EXTRA_KEY in extras:
+                raise ImproperlyConfiguredException(f"{_LEEWAY_EXTRA_KEY} is a reserved key and cannot be set")
+
+            extras[_LEEWAY_EXTRA_KEY] = leeway
+            token = msgspec.convert(payload, cls, strict=False)
```
sobolevn: There's a big problem here (not related to your PR, just a general bug in litestar's JWT implementation): after `jwt` just checked the token payload, `__post_init__` will validate it the second time. 

And since this check is __time__ based, it can fail on the second validation 🙈 

In `django-modern-rest` which used this design from Litestar initially, we changed the logic to never validate `JWToken` at all in `__post_init__` and only validate it on `.encode` and `.decode` which keeps a single source of truth.

I would say that this should be done here as well.

I think that we can know accept this PR as-is and ask you if you want to send a second one with the proper validation fix :)

tests/unit/test_security/test_jwt/test_auth.py
```diff
@@ -895,3 +895,33 @@ class CustomToken(Token):
 
     response = client.get("/", headers={"Authorization": header})
     assert response.status_code == expected_status_code
+
+
+@pytest.mark.parametrize(
+    "leeway, expected_status_code",
+    [
+        pytest.param(2, 200, id="with_leeway"),
```
sobolevn: Please, use `time-machine` for this test.

We need to test more values: `1` second (which is less then `leeway`), exact boundry of `2` seconds, `3` seconds (which is bigger than `leeway`).

sobolevn (commented):
Thank you!

litestar/security/jwt/token.py
happykust: Yeah, I see.  It seems I found correct implementation at [dmr/security/jwt/token.py#L82-L153](https://github.com/wemake-services/django-modern-rest/blob/f0cfdcff98214641345cf6f02cbfb83325cc8c80/dmr/security/jwt/token.py#L82-L153)

If is it, I think I can try to move the same behavior to Litestar)

litestar/security/jwt/token.py
```diff
@@ -66,22 +68,27 @@ class Token:
     """Extra fields that were found on the JWT token."""
 
     def __post_init__(self) -> None:
+        # using extras because of msgspec.convert in decode method below doesn't support InitVar
```
sobolevn:
```suggestion
```

tests/unit/test_security/test_jwt/test_auth.py
```diff
@@ -895,3 +897,35 @@ class CustomToken(Token):
...
+
+
+_LEEWAY: Final = 2
+
+
+@pytest.mark.parametrize(
+    "seconds, expected_status_code",
+    [
+        pytest.param(_LEEWAY - 1, 200, id="with_leeway"),
+        pytest.param(_LEEWAY, 401, id="at_leeway"),
```
sobolevn, line 910: what does the spec say? shoud it be `(..., leeway)` or `(..., leeway]`?

happykust: It should be `(..., leeway)` according to [RFC 7519](https://www.rfc-editor.org/info/rfc7519/#section-4.1.4) and [PyJWT implementation](https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/api_jwt.py#L510-L511), isn't it?

When current datetime reaches token's `exp` value + leeway we must not accept token

sobolevn: please, add an explicit comment about it near the test case

sobolevn (approved):
Thanks!

Btw, if you want me to merge this, you need to fix the CI as the side quest 😅 
You can send another PR for that.

litestar/security/jwt/token.py
```diff
@@ -66,22 +68,26 @@ class Token:
     """Extra fields that were found on the JWT token."""
 
     def __post_init__(self) -> None:
+        leeway = self.extras.get(_LEEWAY_EXTRA_KEY, 0)
```
sobolevn: Oups, nope there's one more thing. Right now 
```suggestion
        leeway = self.extras.pop(_LEEWAY_EXTRA_KEY, 0)
```

this will keep `leeway` in extras. So, creating a token with extra `leeway` and encoding it - must not produce the leeway extra. Please, add a test case for that.

litestar/security/jwt/token.py
sobolevn: now this `.pop` is not needed

happykust:
We need restart CI or just merge #5043 and we're all done :)

sobolevn (approved):
Again, except the part that now we can have douple validation (not related to the PR directly, this bug was already there), LGTM.

In the next PR, please, fix the double validation bug.

happykust:
Okay, got it

litestar/security/jwt/token.py
```diff
@@ -183,6 +194,11 @@ def decode(
             extras = payload.setdefault("extras", {})
             for key in extra_fields:
                 extras[key] = payload.pop(key)
+
+            if _LEEWAY_EXTRA_KEY in extras:
```
sobolevn, line 198: Technically, this is a breaking change, but storing `__leeway__` key would be rather rare :)

sobolevn (approved):

PR #5040, opened 2026-09-04
perf: change how the deprecated resolves per request

Kumzy:
Currently ``_call_handler_function`` call the ``@litestar_deprecated`` on every request, paying a little cost and emitting a unwanted ``DeprecationWarning`` from internal code. Switeched to ``before_request`` attribute directly. The deprecation still fires for user code calling the method explicitly.

Important to note, there is no test for that and in ``pyproject.toml``, we mark that so I am not sure how to add a test for that https://github.com/litestar-org/litestar/blob/cbc0c51abd59b11e7b429be0aba6639e38488e2c/pyproject.toml#L227

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5040">https://litestar-org.github.io/litestar-docs-preview/5040</a>

### DIFF
--- a/litestar/handlers/http_handlers/base.py
+++ b/litestar/handlers/http_handlers/base.py
@@ -720,7 +720,7 @@ class HTTPRouteHandler(BaseRouteHandler):
         response_data: Any = None
         parameter_model = self._get_kwargs_model_for_route(request.scope["path_params"].keys())
 
-        if before_request_handler := self.resolve_before_request():
+        if before_request_handler := self.before_request:
             response_data = await before_request_handler(request)
 
         # create and enter an AsyncExit stack as we may or may not have a
--- a/litestar/handlers/http_handlers/base.py
+++ b/litestar/handlers/http_handlers/base.py
@@ -99,6 +99,7 @@ class HTTPRouteHandler(BaseRouteHandler):
         "_kwargs_models",
         "_request_class",
         "_request_max_body_size",
+        "_resolved_before_request",
         "_response_class",
         "_response_type_handler",
         "_sync_to_thread",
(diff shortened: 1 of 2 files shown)


provinzkraut (changes requested):
This makes it a breaking change, which the deprecation was added to avoid. We still have to *call* `resolve_before_request`, so if it's overwritten in a subclass, we actually use that value.

Kumzy:
> This makes it a breaking change, which the deprecation was added to avoid. We still have to _call_ `resolve_before_request`, so if it's overwritten in a subclass, we actually use that value.

You are right but what about checking that once at init whether the method is actually overwritten? And if so, we call it only in that case?

provinzkraut:
> > This makes it a breaking change, which the deprecation was added to avoid. We still have to _call_ `resolve_before_request`, so if it's overwritten in a subclass, we actually use that value.
> 
> You are right but what about checking that once at init whether the method is actually overwritten? And if so, we call it only in that case?

We could always call at `__init__` time, yeah

Kumzy:
> We could always call at `__init__` time, yeah

I did some modification on the PR but I saw that an other method ``resolve_after_response`` behaves differently.

1. Should I implement the same thing in a different PR?
2. In the changes, the hook is now cached at registration time, so modifying that  on a running app has no effect. Should I add a changelog if we agee on that? Because before it was per request

Also now when calling the super() emits the `DeprecationWarning` once at startup instead of on every request.

litestar/handlers/http_handlers/base.py
```diff
@@ -536,6 +538,9 @@ def request_max_body_size(self) -> int | None:
     def on_registration(self, route: BaseRoute, app: Litestar) -> None:
         super().on_registration(route=route, app=app)
 
+        if type(self).resolve_before_request is not HTTPRouteHandler.resolve_before_request:
+            self._resolved_before_request = self.resolve_before_request()
+
```
provinzkraut, line 543: Are we sure with this check the diff in performance isn't non negligible?

Kumzy: This code only executes on registration of the route (not per request) and the performance is negligible, but there is a performance gain on the per request where the resolve_before_request was before, the warning deprecation was always running on the request while now its set on the registration. But maybe I missed something

Kumzy: I can try to do a benchmark if you want to have real numbers

provinzkraut (approved):

PR #5048, opened 2026-09-06
fix: forward `name` argument in `LitestarGroup.add_command()`

### LINKED ISSUE #5047: Bug: LitestarGroup.add_command() silently ignores the name argument
### Description

While poking around the CLI internals, I noticed that LitestarGroup.add_command() takes a name parameter (matching Click's Group.add_command signature) but never actually passes it through to super().add_command(). So if you call it with a custom name, it's silently dropped and the command just gets registered under cmd.name instead.

This also seems to affect LitestarExtensionGroup, since it calls self.add_command(command, entry_point.name) when loading CLI plugins from entry points,  meaning a plugin's entry point name won't actually be respected if it differs from the command's own .name.

----

**Expected behavior** :  The command should be registered as `custom-name`.

**Actual behavior** :  The command is registered as `original-name` the `name` argument is accepted but has no effect.

## Suggested fix

```python
def add_command(self, cmd: Command, name: str | None = None) -> None:
    if cmd.callback:
        cmd.callback = _inject_args(cmd.callback)
    super().add_command(cmd, name=name)
```

https://github.com/litestar-org/litestar/blob/main/litestar/cli/_utils.py

### Litestar Version

3.0.0b0

aviralsaxena16:
## Description

Fixes #5047 

`add_command()` was ignoring the `name` arg and always falling back to `cmd.name`. This just forwards it properly to Click's `super().add_command()`.

Also fixes the same thing for `LitestarExtensionGroup`, since it uses `name` when loading plugins from entry points.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5048">https://litestar-org.github.io/litestar-docs-preview/5048</a>


provinzkraut (approved):
Thanks!

PR #5055, opened 2026-09-07
fix(middleware): strip port from Host header before allowed-hosts match

### LINKED ISSUE #4869: Bug: `AllowedHostsMiddleware` rejects valid hosts when the `Host` header includes a port

CyberneticX-Tech:
### Description

`AllowedHostsMiddleware` matched the `Host` header **with the port still attached**, so a request carrying `Host: localhost:8000` did not match an allowed host of `localhost` and was rejected with `400 - invalid host header`.

The port-strip was lost in `c4189610`. The walrus binds `host` to the raw header, so `host.split(":")[0]` survives only as the truthiness guard on the left of the `and`, while `fullmatch()` still receives the port-bearing value:

```python
# current - `host` keeps the port, the strip only gates the `if`
    if self.allowed_hosts_regex.fullmatch(host):   # matches against "localhost:8000"
```

This only shows up when the `Host` header carries a port, which is why it goes unnoticed behind a proxy that forwards a bare domain.

### Fix

Bind the port-stripped value to `host` and keep the `None` guard on the raw header:

```python
if (raw_host := headers.get("host")) is not None and (host := raw_host.split(":")[0]):
```

One line, no behavioural change beyond restoring the port-strip.

### Verification

Reproduced locally against the MCVE from the issue. Before the fix `Host: localhost:8000` returned `400`; after it returns `200`, while disallowed hosts stay rejected **with and without** a port, so the check is not weakened:

```
Host: localhost:8000  -> 200 ok
Host: localhost       -> 200 ok
Host: evil.com        -> 400 {"message":"invalid host header"}
Host: evil.com:8000   -> 400 {"message":"invalid host header"}
```

### Tests

Extended the existing `test_middleware_allowed_hosts` parametrisation with three cases: two allowed hosts carrying a port must pass, and a disallowed host carrying a port must still be rejected so the negative path is covered too.

Confirmed these are genuine regression tests. With the fix reverted the two positive cases fail and pass again once it is restored:

```
FAILED test_middleware_allowed_hosts[http://x.example.com:8000-None-200]
FAILED test_middleware_allowed_hosts[http://moishe.zuchmir.com:8000-None-200]
2 failed, 14 passed
```

`tests/unit/test_middleware` is green at **779 passed**. Four failures in `test_compression_middleware` (zstd) and `test_logging_middleware` are pre-existing and reproduce identically on an unmodified checkout, unrelated to this change. `ruff check`, `ruff format --check`, and `mypy` all pass on the touched files.

Closes #4869

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5055">https://litestar-org.github.io/litestar-docs-preview/5055</a>


provinzkraut:
Duplicate of #4873

PR #5058, opened 2026-09-08
fix(routing): backtrack from path-type catch-all to a more specific sibling route

### LINKED ISSUE #4991: `{slug:str}/hello` shadowed by `{path:path}` catch-all — no backtracking in routing trie
### Description

Description

When a `{path:path}` catch-all route and a more specific param-first route with a static suffix (e.g. `{slug:str}/hello`) are both registered, the catch-all always wins and the specific route is never matched, regardless of registration order.

This affects CMS-style applications where a catch-all permalink resolver (`{path:path}`) needs to coexist with specific param-first routes.

Static-first routes (e.g. `/hello/{slug}` or `/t/{slug}`) work fine alongside `{path:path}` because static children are checked before the path-parameter sentinel in the trie. The problem is specific to **param-first + static-suffix** patterns.

_No response_

### MCVE

```python

@get("/{slug:str}/hello", name="specific")
async def specific(slug: str) -> str:
    return f"specific: {slug}"

@get("/{path:path}", name="catch_all")
async def catch_all(path: str) -> str:
    return f"catch_all: {path}"

app = Litestar([specific, catch_all])
```

### Steps to reproduce

Run the app
2. `GET /foo/hello` → returns `catch_all: foo/hello`
3. `GET /foo/bar` → returns `catch_all: foo/bar` (correct — catch-all fallback)

### Screenshots

_No response_

### Logs

```text

```

### Litestar Version

2.24.0

### Platform

- [x] Linux
- [ ] Mac
- [ ] Windows

Yasser-Ameur:
## Description

A `{path:path}` catch-all and a more specific route with a static suffix such as `{slug:str}/hello` cannot coexist today: the catch-all wins for every request, whatever the registration order, so the specific route is unreachable. This is the CMS-style case from #4991, where a permalink resolver on `{path:path}` has to sit beside a few fixed routes.

**Cause.** All dynamic segments at one trie position share a single `PathParameterSentinel` child (`litestar/_asgi/routing_trie/mapping.py`, where `is_path_type` is set on that node). `traverse_route_map` consumed the rest of the path the moment it reached a node with `is_path_type=True`, without trying that node's static children, so `hello` was never looked at.

**Change.** `traverse_route_map` now remembers the catch-all node when it also has static children and keeps matching the specific branch. It falls back to consuming the remainder as the catch-all only when the specific branch dead-ends, ends on a node without handlers, or the request ends on the shared node itself (so a one-segment request still yields `/foo`, as before). A path-type node with no static children keeps the previous one-line fast path, so routes that already worked take the same code as before, and no allocation is added to the hot path for them.

**Proof.** Three new tests in `tests/e2e/test_routing/test_path_resolution.py`, all failing on `main`:

- `test_specific_route_matches_before_path_type_catchall[False|True]`: both registration orders; `/foo/hello` hits the slug route, `/foo/bar/baz` and `/foo` hit the catch-all with values `/foo/bar/baz` and `/foo`.
- `test_specific_route_with_deeper_static_suffix_matches_before_path_type_catchall`: `/{slug:str}/hello/world` beside the catch-all, including the partial-match fallback `/foo/hello/other`.
- `test_path_type_node_without_handlers_is_not_found`: `/{path:path}/foo` alone still returns 404 for `/a/b`, covering the fallback's own not-found branch.

`pytest tests/e2e/test_routing tests/unit/test_asgi/test_routing_trie`: 78 passed. `ruff check`, `ruff format --check` and `mypy` clean on the touched files. Coverage on `traversal.py`: every changed line covered; the remaining gaps predate this change.

Disclosure per the Litestar AI policy: the change was prepared with AI assistance under my direction, and I am the human in the loop responsible for it.

Fixes #4991

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5058">https://litestar-org.github.io/litestar-docs-preview/5058</a>


Yasser-Ameur:
Reopening to rerun CI: the 3.12 failure was test_forbid_unknown_fields in test_dto, which does not touch this change and passed on the previous run of the same tree.

PR #5065, opened 2026-09-09
feat(redis): support NX and XX in RedisStore.set

Shekhrozx:
## Description

Add `NX` and `XX` options to `RedisStore.set()` to support Redis atomic conditional writes.

### Problem

Currently, `RedisStore.set()` does not expose Redis `NX`/`XX` options.

This makes implementing idempotency safely difficult. For example:

```py
if not await store.get(idempotency_key):
    await store.set(idempotency_key, "processing") 
    await create_record()
```

Two concurrent requests can both execute `GET`, receive `None`, and then both create records.

Redis `SET NX` solves this atomically:

```py
created = await store.set(
    idempotency_key,
    "processing",
    expires_in=3600,
    nx=True,
)

if not created:
    # Already exists / being processed
    ...
```

Only one concurrent request can successfully create the key.

### Proposed Change

Expose `nx` and `xx` in `RedisStore.set()`:

```py 
await store.set(
    key,
    value,
    expires_in=3600,
    nx=True,
)
```

and:

```py
await store.set(
    key,
    value,
    xx=True,
)
```

The method should return whether the conditional `SET` succeeded, allowing callers to determine whether they acquired the key.

`NX` and `XX` should be mutually exclusive.

### Motivation

This enables common Redis patterns directly through `RedisStore`, including:

- Idempotency keys
- Distributed locks
- Atomic "create if not exists"
- Conditional cache updates

It also avoids requiring developers to bypass `RedisStore` and access the underlying Redis client directly.

### Backward Compatibility

Existing calls without `nx` or `xx` retain the current behavior.

### Tests

Add tests covering:

- `NX` with an existing/non-existing key
- `XX` with an existing/non-existing key
- `NX` + `XX` validation

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5065">https://litestar-org.github.io/litestar-docs-preview/5065</a>

### DIFF
--- a/litestar/stores/redis.py
+++ b/litestar/stores/redis.py
@@ -147,6 +147,8 @@ class RedisStore(NamespacedStore):
         value: str | bytes,
         expires_in: int | timedelta | None = ...,
         keep_ttl: Literal[False] = ...,
+        nx: bool = ...,
+        xx: bool = ...,
     ) -> None: ...
 
     @overload
@@ -157,6 +159,8 @@ class RedisStore(NamespacedStore):
         expires_in: None = ...,
         *,
         keep_ttl: Literal[True],
+        nx: bool = ...,
+        xx: bool = ...,
     ) -> None: ...
 
     async def set(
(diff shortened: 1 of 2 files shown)


provinzkraut:
Stores are supposed to be general purpose abstractions, not e.g. an alternative interface to the underlying storage engine. If you need redis specific functionality, you should use redis.

PR #5069, opened 2026-09-09
fix(dto): keep nested DTOData collections as builtins, until create_instance() (#3620)

### LINKED ISSUE #3620: Bug: DTOData class not behaving as expected and PydanticDTO class error validation
### Description

The DTOData class works as expected when there isn't a nested model. Based on the documentation, the idea of DTOData class is to delay the validation until `create_instance` method is called.

This is working as expected for the first level model but calls Pydantic validation on nested models.

**Second issue:**  I am facing an issue when overriding the `ValidationException` class. I am using an extra field in the error response which is not serialised properly. An internal error is raised:

```py
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/contrib/pydantic/pydantic_dto_factory.py", line 101, in decode_bytes
    return super().decode_bytes(value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/dto/base_dto.py", line 97, in decode_bytes
    return backend.populate_data_from_raw(value, self.asgi_connection)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/dto/_codegen_backend.py", line 143, in populate_data_from_raw
    data_as_builtins=self._transfer_to_dict(self.parse_raw(raw, asgi_connection)),
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 22, in func
  File "<string>", line 22, in <genexpr>
  File "<string>", line 25, in func
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/pydantic/main.py", line 176, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for B
nested_c
  List should have at most 1 item after validation, not 2 [type=too_long, input_value=[C(i_am_c=None), C(i_am_c=None)], input_type=list]
    For further information visit https://errors.pydantic.dev/2.7/v/too_long

  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/middleware/_internal/exceptions/middleware.py", line 158, in __call__
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/_asgi/asgi_router.py", line 99, in __call__
    await asgi_app(scope, receive, send)
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/routes/http.py", line 80, in handle
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/routes/http.py", line 132, in _get_response_for_request
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/routes/http.py", line 152, in _call_handler_function
    response_data, cleanup_group = await self._get_response_data(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/routes/http.py", line 176, in _get_response_data
    data = await kwargs["data"]
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/_kwargs/extractors.py", line 501, in dto_extractor
    return data_dto(connection).decode_bytes(body)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/contrib/pydantic/pydantic_dto_factory.py", line 103, in decode_bytes
    raise ValidationException(extra=convert_validation_error(ex)) from ex
litestar.exceptions.http_exceptions.ValidationException: 400: Bad Request

  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/serialization/msgspec_hooks.py", line 162, in encode_json
    return msgspec.json.encode(value, enc_hook=serializer) if serializer else _msgspec_json_encoder.encode(value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/serialization/msgspec_hooks.py", line 92, in default_serializer
TypeError: Unsupported type: <class 'app.C'>

  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/uvicorn/protocols/http/httptools_impl.py", line 399, in run_asgi
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 70, in __call__
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/app.py", line 591, in __call__
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/middleware/_internal/exceptions/middleware.py", line 175, in __call__
    await self.handle_request_exception(
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/middleware/_internal/exceptions/middleware.py", line 205, in handle_request_exception
    await response.to_asgi_response(app=None, request=request)(scope=scope, receive=receive, send=send)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/response/base.py", line 451, in to_asgi_response
    body=self.render(self.content, media_type, get_serializer(type_encoders)),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/response/base.py", line 392, in render
    return encode_json(content, enc_hook)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/redacted/Library/Caches/pypoetry/virtualenvs/redacted/lib/python3.12/site-packages/litestar/serialization/msgspec_hooks.py", line 164, in encode_json
    raise SerializationException(str(msgspec_error)) from msgspec_error
litestar.exceptions.base_exceptions.SerializationException: Unsupported type: <class 'app.C'>
```

_No response_

### MCVE

```python

from litestar import Controller, Litestar, Request, Response, post
from litestar.dto import DTOConfig, DTOData
from litestar.exceptions import ValidationException

IdVar = TypeVar("IdVar")


def router_handler_exception_handler(
    request: Request, exc: ValidationException
) -> Response:
    return Response(
        content={
            "error": "validation error",
            "path": request.url.path,
            "extra": exc.extra,
        },
        status_code=400,
    )


class BaseA(BaseModel, Generic[IdVar]):
    id_var: IdVar = None


class C(BaseModel):
    i_am_c: str | None = None


class B(BaseA[int], Generic[IdVar]):
    i_am_b: str | None
    nested_c: list[C] = Field(default=[], max_length=1)


class A(BaseA[str], Generic[IdVar]):
    i_am_a: str | None
    nested_b: list[B] = Field(default=[], max_length=1)


class WriteADto(PydanticDTO[A]):
    config = DTOConfig(
        rename_strategy="camel",
        max_nested_depth=3,
        exclude={"id_var", "nested_b.0.id_var"},
    )


class Test(Controller):
    @post(path="/test", dto=WriteADto)
    async def test(self, data: DTOData[A]) -> None:
        print(data.create_instance())


app = Litestar(
    route_handlers=[Test],
    exception_handlers={ValidationException: router_handler_exception_handler},
)
```

### Steps to reproduce

```bash
For Issue: 1
1. Call the /test endpoint with following data:

{
  "iAmA": null,
  "nestedB": [
    {
      "iAmB": null,
      "nestedC": [
        {
          "iAmC": null
        }
      ]
    },
    {
      "iAmB": null,
      "nestedC": [
        {
          "iAmC": null
        }
      ]
    }
  ]
}

2. Works without raising an error.
3. Comment out `exception_handlers` from app.
3. Call /test endpoint with following data:
```json
{
  "iAmA": null,
  "nestedB": [
    {
      "iAmB": null,
      "nestedC": [
        {
          "iAmC": null
        },
        {
          "iAmC": null
        }
      ]
    }

  ]
}
```
4. Raises an internal server error and ValidationException.

For Issue 2:
1. Enable `ValidationException` in the app.
2. Call /test endpoint with following data:
```json
{
  "iAmA": null,
  "nestedB": [
    {
      "iAmB": null,
      "nestedC": [
        {
          "iAmC": null
        },
        {
          "iAmC": null
        }
      ]
    }

  ]
}
```
2. Raises an internal server error and ValidationException and SerializationException.
```


### Screenshots

```bash
```

### Logs

_No response_

### Litestar Version

2.9.0final0

### Platform

- [ ] Linux
- [X] Mac
- [ ] Windows

bishnusyangja:
## Description

Fixes #3620

For the first issue:
It seems the pydantic has solved the problem

For second issue:
`DTOData` is documented to keep payload data as builtins and only construct the domain model when `create_instance()` is called. That already worked for top-level nested models, but nested models inside collections and mappings were still instantiated during decode.

That meant Pydantic constraints on nested types ran before the handler, and a failed nested validation could raise `SerializationException` when the exception extra still held model instances.

Nested values now remain dictionaries until `create_instance()`, matching top-level `DTOData` behaviour. Applies to both the default and experimental codegen backends.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5069">https://litestar-org.github.io/litestar-docs-preview/5069</a>


bishnusyangja:
@itzzdev09 I would be grateful if you have some time to review this PR

itzzdev09:
@bishnusyangja happy to. I checked this out and ran it against `main` rather than reading the diff alone. **The fix is real and does what it claims** — but there's one regression I'd want resolved before merge.

## The fix works

I confirmed your regression test genuinely fails on `main` (HTTP 500) and passes here, and that the behaviour change is exactly what #3620 asks for. Comparing `as_builtins()` output across both backends:

| shape | `main` | this PR |
|---|---|---|
| `list[Leaf]` | `[Leaf(name='a')]` | `[{'name': 'a'}]` ✅ |
| `dict[str, Leaf]` | `{'k': Leaf(name='a')}` | `{'k': {'name': 'a'}}` ✅ |
| `list[list[Leaf]]` | `[[Leaf(name='a')]]` | `[[{'name': 'a'}]]` ✅ |
| `tuple[Leaf, ...]` | `(Leaf(name='a'),)` | `({'name': 'a'},)` ✅ |
| nested obj → list (`max_nested_depth=2`) | `{'items': [Leaf(name='a')]}` | `{'items': [{'name': 'a'}]}` ✅ |

The return/encode path is untouched (`{"items":[{"name":"a"},{"name":"b"}]}` before and after), and `tests/unit/test_dto` + `tests/unit/test_plugins/test_pydantic` give **415 passed** on this branch. The only 2 failures are `test_openapi_schema_for_type_with_custom_generic_type`, which fail identically on `main` — missing `advanced_alchemy` locally, not yours.

`_nested_destination_type` is a nice call, by the way — it kills three copies of that branch and is what lets the codegen backend stay in sync.

## Blocking: `create_instance()` no longer rebuilds models behind a union

`as_builtins()` now correctly defers, but for a **union-typed nested field** `create_instance()` never converts the dict back, so the caller gets a raw `dict` where a model is declared:

```
                          main                          this PR
Optional[Leaf]            TopHolder(inner=Leaf('a'))    TopHolder(inner={'name': 'a'})   ❌
list[Optional[Leaf]]      [Leaf('a'), None]             [{'name': 'a'}, None]            ❌
```

Both backends, identically. A handler doing `data.create_instance().inner.name` gets `AttributeError: 'dict' object has no attribute 'name'`. Note this hits the **plain top-level `Optional[Leaf]` field** too, not just collections.

### Root cause

`_transfer_nested_union_type_data` in `litestar/dto/_backend.py` picks the union member by isinstance:

```python
if inner_type.nested_field_info and isinstance(
    source_value,
    inner_type.nested_field_info.model if is_data_field else inner_type.field_definition.annotation,
):
```

On the `create_instance()` path (`transfer_data_from_builtins`) `is_data_field` is `True`, so it tests `isinstance(source_value, Leaf)`. Before this PR `source_value` *was* a `Leaf`, so it matched. Now it's a `dict`, nothing matches, the loop falls through to `return source_value`, and the dict is handed back untouched.

The codegen backend has the same shape at `_handle_transfer_instance` — `constraint_type = simple_type_.nested_field_info.model` feeding `isinstance({source_value_name}, {constraint_type_name})` — which is why both backends fail together.

The non-union path is fine because it goes straight to `_transfer_instance_data`, which already handles a `Mapping` source.

### Repro

Four tests, **all pass on `main`, all fail here**:

```python
@dataclass
class UnionLeaf:
    name: str

@dataclass
class UnionTopHolder:
    inner: Optional[UnionLeaf]

@dataclass
class UnionListHolder:
    items: list[Optional[UnionLeaf]]

@pytest.mark.parametrize("codegen", [False, True])
def test_create_instance_rebuilds_optional_nested_model(codegen: bool) -> None:
    @post(
        dto=DataclassDTO[Annotated[UnionTopHolder, DTOConfig(experimental_codegen_backend=codegen)]],
        return_dto=None,
        sync_to_thread=False,
    )
    def handler(data: DTOData[UnionTopHolder]) -> dict[str, Any]:
        instance = data.create_instance()
        assert isinstance(instance.inner, UnionLeaf), f"got {type(instance.inner).__name__}"
        return {"ok": True}

    with create_test_client(route_handlers=[handler]) as client:
        assert client.post("/", json={"inner": {"name": "a"}}).status_code == 201
```

(plus the `UnionListHolder` equivalent asserting `isinstance(instance.items[0], UnionLeaf)` and `instance.items[1] is None`.)

### Direction

The selector needs to accept a `Mapping` when the source is builtins, e.g. matching on `Mapping` in addition to the model type. One caveat worth thinking through rather than patching past: for `Union[ModelA, ModelB]` a bare dict is genuinely ambiguous and the first nested member would win. `Optional[Model]` — by far the common case, and the one that regressed — has only one nested member, so it's unambiguous there. It may be worth restricting the Mapping fallback to that case and leaving multi-model unions on the existing behaviour.

## Minor

1. **Mappings are fixed but untested.** `dict[str, Leaf]` works (table above), and you did change `transfer_type.value_type` to propagate `nested_as_dict` — but both new tests use lists. A mapping case would lock that in.
2. **`destination_type is dict` early return** is correct and slightly faster (`dict(**unstructured_data)` built a second dict). `unstructured_data` is freshly constructed per call, so returning it directly aliases nothing.
3. Changelog entry reads well and the `:issue:` reference is right.

Happy to re-check once the union path is sorted.

bishnusyangja:
@itzzdev09 I have updated the PR, and addressed your comments above. I would be happy if you review it again.

itzzdev09:
A few updates from my side:

The earlier blocker is fixed. create_instance() now correctly rebuilds a real model for Optional[Model] and list[Optional[Model]], while preserving None as None. This works on both the default and codegen backends.
as_builtins() now keeps nested values as dicts ({'inner': {'name': 'a'}} instead of {'inner': A(name='a')} on main).
Union[Model, str] works for both cases: a dict payload is converted to the model, while a string remains a string.
Tests: tests/unit/test_dto and the Pydantic plugin tests pass with 423 passed. The remaining 2 failures are test_openapi_schema_for_type_with_custom_generic_type, and they fail identically on main. main also fails the PR's new test_dto_data_nested_collection_as_builtins, as expected.
No regression here: Union[ModelA, ModelB] returns 500 on both main and this PR. This is because msgspec rejects untagged unions containing multiple Struct types (all Struct types must be tagged).
Non-blocking suggestions
In litestar/dto/_backend.py, _transfer_nested_union_type_data: nested_member_count is recomputed on every call/request, even though it only depends on transfer_type. It could potentially be computed once and reused.
In litestar/dto/_codegen_backend.py, the f-string passed to self._start_block(...) is a little difficult to read. Assigning the condition to a variable first would make this easier to follow.
The comment mentioning that Union[ModelA, ModelB] "stays on the isinstance(model) path" is a bit misleading, since DTOs cannot currently decode these untagged unions. I'd suggest rewording the comment or adding a test that explicitly pins the current error.
Missing test: It would be good to cover Union[Model, str] with a dict payload—the case where the other union member is not None. I checked this behaviour with a small script on the PR branch and it looks correct, although I haven't run this exact test yet.

Here's a test you can use:

```
@dataclass
class UnionStrHolder:
    inner: Union[UnionLeaf, str]

def test_dto_data_create_instance_rebuilds_model_in_union_with_scalar(
    use_experimental_dto_backend: bool,
) -> None:
    @post(
        dto=DataclassDTO[
            Annotated[
                UnionStrHolder,
                DTOConfig(experimental_codegen_backend=use_experimental_dto_backend),
            ]
        ],
        return_dto=None,
        sync_to_thread=False,
    )
    def handler(data: DTOData[UnionStrHolder]) -> dict[str, Any]:
        assert data.as_builtins() == {"inner": {"name": "a"}}
        assert isinstance(data.create_instance().inner, UnionLeaf)
        return {"ok": True}

    with create_test_client(route_handlers=[handler]) as client:
        assert client.post("/", json={"inner": {"name": "a"}}).status_code == 201
```

If the test file doesn't already import Union, it will also need:

`from typing import Union`

bishnusyangja:
@itzzdev09 I added the missing tests for Union and all other required fixes. I hope its ready for merge now.

itzzdev09:
Thanks for the updates — the changes look right to me. Just to set expectations: I'm not a maintainer and have no merge access, so my review doesn't move this forward on its own. It needs a litestar maintainer; @provinzkraut  has been merging most DTO changes lately.

PR #5078, opened 2026-09-17
feat(stores): use Redis GETEX for renew_for on 6.2+

### LINKED ISSUE #4993: Enhancement: Use redis `GETEX` in `RedisStore.get` for `renew_for` functionality on redis 6.2+

TianHengZhuang:
## Description

Closes #4993

`RedisStore.get(..., renew_for=...)` now uses the native Redis `GETEX` command when the server supports it (Redis 6.2+), instead of the custom Lua script.

The Lua script remains as a fallback for older Redis servers. Support is probed once per store instance via a no-op `GETEX` on a missing key.

## Notes

- GETEX is atomic get+set-expiry, which is the point of the issue.
- Semantic note documented in the docstring: on Redis 6.2+, `GETEX` will set an expiry even if the key previously had none; the Lua path only renews when `TTL > 0`.
- Added unit tests that mock the client to pin path selection (getex vs Lua fallback). Integration renew tests are unchanged.

<hr>
📚 Documentation preview 📚: <a href="https://litestar-org.github.io/litestar-docs-preview/5078">https://litestar-org.github.io/litestar-docs-preview/5078</a>


TianHengZhuang:
Pushed a mypy fix: the support probe now uses the typed `getex` client method instead of untyped `execute_command` (CI had `no-untyped-call` at redis.py:88). Unit tests updated for the probe-based path selection.

PR #5088, opened 2026-09-21
perf(pydantic): serialize models to JSON via `msgspec.Raw`

Kumzy:
The pydantic `BaseModel` type encoder now returns `msgspec.Raw(model.__pydantic_serializer__.to_json(...))` instead of `model.model_dump(mode="json")`, so models are no longer converted into a dict that msgspec has to traverse a second time.


litestar/plugins/pydantic/plugins/init.py
```diff
@@ -123,15 +124,17 @@ def _create_pydantic_v2_encoders(
...
-                by_alias=prefer_alias,
-                exclude=exclude,
-                exclude_defaults=exclude_defaults,
-                exclude_none=exclude_none,
-                exclude_unset=exclude_unset,
-                include=include,
-                mode="json",
-                round_trip=round_trip,
+            pydantic.BaseModel: lambda model: msgspec.Raw(  # pyright: ignore[reportOptionalMemberAccess]
+                model.__pydantic_serializer__.to_json(
```
provinzkraut, line 128: Why not `model_dump_json`?

https://pydantic.dev/docs/validation/dev/api/pydantic/base_model/#pydantic.BaseModel.model_dump_json

sobolevn: One Python frame less :)

provinzkraut: Haha, I guess! Don't think it makes much of a difference here though? And you trade that for having to use a private method 🤷

Kumzy: We can not use as is `model_dump_json` because we need to return a dict there for msgspec. On the use of `model_dump_json`, it use behind the same private function + a `decode()` function.
As for the private method, I agree with you on that but its documented  [there](https://github.com/pydantic/pydantic/blob/v2.12.5/pydantic/main.py#L139). 
If we want to use `model_dump_json`, we can but we will have to use `Raw` of msgspec anyway and it owuld do bytes -> str -> bytes

provinzkraut: > We can not use as is model_dump_json because we need to return a dict there for msgspec. 

Why? That's why we use `msgspec.Raw`, so we can hand an already serialised string to it.

Kumzy: > Why? That's why we use `msgspec.Raw`, so we can hand an already serialised string to it

We can as stated in the last part of the message but that would add a convertion instead of using the pydantic private method whjere we do not have that. My message was not so clear

provinzkraut: I don't think I understand. With the private method, we're getting a dict, that we're then passing to msgspec? How is that better than an already encoded JSON string that we're passing to msgspec?

sobolevn: Source: https://github.com/pydantic/pydantic/blob/915896d163835a57fd7987180087409a3229bd71/pydantic/main.py#L538-L604

sobolevn: In my opinion calling the "private" method is better: one less frame and one less `str` creation.

provinzkraut: Ah okay. It literally just calls that thing internally. In that case: Yes, seems like the correct choice!

Kumzy: Yes maybe it was not clear, I will fix the test

